Issue #1288, opened 2023-06-08
nimble-rich-text-editor and nimble-rich-text-viewer Components


vivinkrishna-ni:
## 😯 Problem to Solve

Similar issues: #793 #794 
Requirement for `Comments`  Feature to create, view, and edit rich text content.
[UI Mockup For Comment](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045)

## 💁 Proposed Solution

- FAST component: N/A
- Other reference components: N/A

1. `nimble-rich-text-editor` to create and edit the rich text and get the `markdown` output.
2. `nimble-rich-text-viewer` to view the rich text.

## 📋 Tasks

- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` spec
- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` custom element
- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` unit tests
- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` Storybook tests
- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` Storybook docs
- [ ] `nimble-rich-text-editor` and `nimble-rich-text-viewer` Angular wrapper
- [ ] Add to Angular example app
- [ ] Update Component Status table in README

m-akinc:
These components now exist in an "incubating" state. This issue now tracks any work necessary to bring it out of incubating.

jattasNI:
We discussed the long term plan to move this out of incubating as part of #2406. The proposal there is that this component belongs in the Spright package within the Nimble repo. It's not a reflection of the quality, completeness, or usefulness of the component, just that it is more special-purpose than typical Nimble components like text field and table.

The work involved would be to move the code from packages/nimble* to packages/spright* and move the stories to the Spright section of storybook. This would be a breaking change for clients as they would now import from spright-angular, SprightBlazor, or spright-components packages. We'd need to either update client apps or notify them to make the updates themselves.

@vivinkrishna-ni What do you think of this proposal? If you like the direction, any chance your team would be interested in implementing that migration?

vivinkrishna-ni:
@jattasNI We have discussed this internally within the team and there are a couple of questions we would like to clarify ourselves before taking up the work of migration for rich text components to Spright. I agree that certain criteria mentioned in the description of this issue have to be met to move it out of incubating. 

[...]

PR #1272, opened 2023-05-26
HLD: Accessible labels and localization in Nimble

msmithNI:
# Pull Request

## 🤨 Rationale

#1090

## 👩‍💻 Implementation

N/A (HLD and doc updates)

## 🧪 Testing

N/A

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1289, opened 2023-06-08
Nimble Rich Text Editor and Viewer Spec Document

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Creating spec for `nimble-rich-text-editor` and `nimble-rich-text-viewer` components.
Issue: #1288 

## 👩‍💻 Implementation

This PR contains the nimble spec document for the new nimble component addition. The spec document consists of a design discussion for the `nimble-rich-text-editor` and `nimble-rich-text-viewer` components and explains the external libraries used for the development of a rich text editor.  

## 🧪 Testing

N/A

## ✅ Checklist

N/A


packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
+# Nimble Rich Text Editor and Nimble Rich Text Viewer
+
+## Overview
+
+-   The `nimble-rich-text-editor` component is a text editing tool that offers a comprehensive range of functionalities for
+    incorporating rich text content. It provides users with the ability to apply different text styles and formats to their content.
+-   The `nimble-rich-text-viewer` component is way to view all the rich text content.
+
+### Background
+
```
vikisekarNI, line 10: We can write a brief two liner on the potential use cases that has driven the idea of building these components. The background or need for these components are obviously driven by
1. Comments
2. Description

Get some inspirations from [Comments Epic Description](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/ASW%20SystemLink%20Platform/Initiatives/?workitem=2205215)

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+[Nimble issue #1288](https://github.com/ni/nimble/issues/1288)
+
+[Comments UI mockup](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045)
+
+### Non-goals
+
+-   The visual and interaction design for `nimble-rich-text-editor` and `nimble-rich-text-viewer` is out of scope for this spec component.
```
vikisekarNI, line 17: Let's elaborate this more specifically. 
1. ViD and IxD are out of scope for these components in nimble design system  
2. However we have an IxD workflow for SLE application (the figma mockup you linked above) that we will rely on.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+2. Using the keyboard shortcuts to modify the text formatting in the editor.
+3. Support to get markdown output from the editor.
+4. Support to convert the markdown to HTML to render it back in the `nimble-rich-text-viewer` component.
+
+#### _Additional features out of scope of this spec_
+
+-   Allowing the user to tag or mention by entering `@` in the editor and selecting the user name from the drop-down list.
+-   Support for adding images to the editor either by uploading or by pasting it.
+-   Support for adding hyperlinks to the existing words in the editor. However, adding raw links by pasting to the editor is supported.
+-   Support for underlining the text.
```
vikisekarNI, line 46: Since we are putting underline support out of scope which is typically done using `<u>` tag of HTML in any markdown, this eliminates our dependency to use html within markdown for current set of features. 

If there are options in tip tap to switch off HTML tags in markdown, it will be worthy to use it and mention it in security section which will completely disallow injection attacks using `<script>` tags into our component, until we need HTML tags

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+-   Provide APIs to make the buttons in the toolbar located below the editor to be hidden or disable.
```
vikisekarNI, line 47:
```suggestion
-   Provide APIs to configure visibility/disabled state of toolbar buttons for rich text editor
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Risks and Challenges
+
+-   Visual design and interaction design for both the components are out of scope for the initial release. We will start with the existing
+    [mockup screens](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045) which is designed for
+    comments feature. At this moment, we cannot afford to cover all the corner case of visual design specs but we managed to refer from
+    other visual designs and try to incorporate with the nimble standards.
```
vikisekarNI, line 54:
```suggestion
-   Visual design and interaction design for both the components are out of scope for the initial release. We will start with the existing
    [mockup screens](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045) which is designed for
    comments feature considering customer requirements and deadlines. At this moment, we cannot afford to cover all the corner cases of visual design specs. However we are organically referring to existing nimble components like `button`, `text-area` to provide a consistent visual design
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+-   Due to the immediate requirement for GM (`Comments Feature`), we are not adding any additional enhancements or feature requirements
+    apart from the mentioned features in the initial release.
```
vikisekarNI, line 56: Be careful on mentioning customer names in open source libraries. 

```suggestion
-   Due to immediate requirements for comments feature from a business customer, any additional enhancements or requirements apart from whatever is mentioned in this spec are deferred to future scope. 
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+-   The mobile view of the component has not been designed yet, and we are actively collaborating with the design team to create basic mockup screens. We will
+    update this spec accordingly as progress is made.
```
vikisekarNI, line 58:
```suggestion
    update this spec accordingly based on the progress.
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+Sample Mockup Screenshots
+
+_`nimble-rich-text-editor`_
+
+![Editor](./spec-images/editor-sample.png)
+
+_`nimble-rich-text-viewer`_
+
+![Viewer](./spec-images/viewer-sample.png)
```
vikisekarNI, line 70: Let's crop the top header division with username, date and edit/delete menu as per latest discussion of scoping it into application workflow.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+_Props/Attrs_
+
+-   none
+
+_Methods_
+
+-   `setMarkdown()` - sets the markdown input from the consumer component to preload the content in the editor.
+-   `getMarkdown()` - gets the processed markdown output from the component.
+-   [`destroy()`](https://tiptap.dev/api/editor#destroy) - stops the editor instance and unbinds all events.
```
vikisekarNI, line 128: Any need for exposing `destroy`? Do we need to handle it internally on detaching this custom element?

vikisekarNI (changes requested):

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Added a brief point to elaborate on the background for this component addition.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+#### `nimble-rich-text-editor`
+
+_Props/Attrs_
+
+-   none
+
+_Methods_
+
+-   `setMarkdown()` - sets the markdown input from the consumer component to preload the content in the editor.
```
vikisekarNI, line 132:
```suggestion
-   `setMarkdown()` - sets the markdown input from the consumer component which can be used for preloading the content
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Yes, it can be handled internally inside the custom element by destroying the editor in `disconnectedCallback` lifecycle hooks.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,335 @@
...
+-   `getMarkdown()` - gets the processed markdown output from the component.
+-   [`destroy()`](https://tiptap.dev/api/editor#destroy) - stops the editor instance and unbinds all events.
+-   [`isEmpty()`](https://tiptap.dev/api/editor#is-empty) - obtains a boolean value indicating whether the editor is empty or not.
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, we can emit the `isEmpty` status through the `onChange` event. I have removed the same in the document.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+_Note_: All other keyboard interaction determined by the slotted element will not be defined in this document.
+
+### Globalization
+
+Currently, this component does not have any string data to be localized. In the footer menu for text formatting, only icons will be displayed. However,
+strings may be used for tooltips to enable localization, which will be managed through the `nimble-tooltip` component.
+
+### Security
+
+-   none
```
vikisekarNI, line 300: Consider having the below security items: 
1. Prose mirror uses markdown-it for converting markdown to HTML and HTML to markdown. We will follow the [security guidelines of markdown-it](https://github.com/markdown-it/markdown-it/blob/master/docs/security.md#security) to turn off HTML at source as given in the [API docs](https://markdown-it.github.io/markdown-it/#MarkdownIt.new)
2. For additional safety, use [sanitize-html](https://www.npmjs.com/package/sanitize-html) package to whitelist only specific html tags on need basis or completely disallow any html tags

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Test Plan
+
+The current testing plan involves writing unit tests to verify the core functionality of the component. All tests for this component will
+follow the standards set by the nimble repository, which include unit tests to assess component logic and Chromatic tests to cover all
+visual states of the component.
+
+### Tooling
+
+_Are there any special considerations for tooling? Will tooling changes need to be made? Is there a special way to light up this component
```
vikisekarNI, line 326: Update this section as applicable - NA

vikisekarNI (changes requested):

vikisekarNI (changes requested):
Address suggestions to important sections

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,339 @@
...
+2. `footer` section consists of `nimble-button` to control each text formatting functionalities like bold, italic etc,. and a
+   `footer-actions` slot element which is typically used to add action buttons to the right bottom of the component.
+
+Here is the basic high level template used in this component:
+
+```html
+<template>
+    <section>
+        <div id="editor"></div>
+    </section>
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated to section to wrap the entire editor into it.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,339 @@
...
+    <footer>
+        <div id="toolbar">
+            <nimble-button></nimble-button>
+        </div>
+        <div>
+            <slot name="footer-actions"></slot>
+        </div>
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated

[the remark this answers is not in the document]
vivinkrishna-ni: Yes, we can use section instead of div. I have updated the same.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Deferring this comment as it is captured in the security section of this doc.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+-   `setMarkdown()` - sets the markdown input from the consumer component to preload the content in the editor.
+-   `getMarkdown()` - gets the processed markdown output from the component.
+-   [`destroy()`](https://tiptap.dev/api/editor#destroy) - stops the editor instance and unbinds all events.
+-   [`isEmpty()`](https://tiptap.dev/api/editor#is-empty) - obtains a boolean value indicating whether the editor is empty or not.
+
+_Events_
+
+-   `onChange` - event emitted when there is a change in the the editor. This can be achieved through tiptap's
+    [transaction event](https://tiptap.dev/api/events#transaction).
```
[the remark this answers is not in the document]
vivinkrishna-ni: For `change` event, we will be using the update event exposed by tiptap. I have updated the same in the doc.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+5. Includes support to retrieve the content of the editor as HTML and Markdown output (using
+   [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown)).
+6. Includes extensions to support [@mention](https://tiptap.dev/api/nodes/mention) functionality (Future scope).
+
+The `nimble-rich-text-editor` is initialized by creating an instance of the [Editor](https://tiptap.dev/api/editor#introduction) class from
+the Tiptap's core library. With that we have access to all the APIs exposed, by utilizing some of their extensions like
+[StarterKit](https://tiptap.dev/api/extensions/starter-kit) which is a collection of most popular Tiptap extensions includes bold, italics
+and all other basic rich text formatting options. All these options([marks](https://tiptap.dev/api/marks)) can also be accessed individually.
+
+The rich text content entered in the editor is converted to markdown output using `prosemirror-markdown` serializer. Here is the reference for basic formatting
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated the doc on the same.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+| Ctrl + Shift + 8        | To enable the focused paragraph a bulleted list          |
+| Tab (For lists)         | To create a sub point within a bulleted or numbered list |
+| Shift + Tab (For lists) | To remove a sub point from a bulleted or numbered lists  |
+
+_Keyboard navigation with toolbar buttons focused_
+
+| Key          | Behavior                                                                       |
+| ------------ | ------------------------------------------------------------------------------ |
+| Space, Enter | Enable the selected text formatting feature and return the focus to the editor |
+| Shift + Tab  | To reach the editor back to focus                                              |
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, we will make sure to capture it in the development.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Added the points as suggested.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+A Blazor wrapper will not be created for the component.
+
+### Visual Appearance
+
+NA
+
+## Implementation
+
+We have chosen to utilize the [`Tiptap`](https://tiptap.dev/) rich text editor as the underlying third-party library for developing the
+`nimble-rich-text-editor`. This decision was made due to its extensive range of customization options over other third party libraries for
```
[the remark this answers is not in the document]
vivinkrishna-ni: We have added the same content here as well for your information.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,337 @@
...
+rich text editing and it is the ideal choice for meeting our specific use cases and some are mentioned below:
+
+1. Includes all [basic functionalities](https://tiptap.dev/) like bold, italics, numbered and bulleted lists etc.
+2. Includes support to work seamlessly inside the `shadow root`.
+3. [Tiptap](https://tiptap.dev/introduction) is built on top of ProseMirror and so have access to all the powerful APIs provided by
+   ProseMirror through [`@tiptap/pm`](https://tiptap.dev/guide/prosemirror).
+4. Includes support to render the content of the editor with markdown input.
+5. Includes support to retrieve the content of the editor as HTML and Markdown output (using
```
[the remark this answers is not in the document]
vivinkrishna-ni: HTML output is not part of the initial scope and is also not in the roadmap. We aligned to have markdown as the only output. If there is any requirements on this we will update the same here.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,373 @@
...
+4. Support to convert the markdown to HTML to render it back in the `nimble-rich-text-viewer` component.
+
+#### _Additional features out of scope of this spec_
+
+-   Allowing the user to tag or mention by entering `@` in the editor and selecting the user name from the drop-down list.
+-   Support for adding images to the editor either by uploading or by pasting it.
+-   Support for adding hyperlinks to the existing words in the editor. However, adding raw links by pasting to the editor is supported.
+-   Support for striking out and underlining text. The reason for including these features in future scope that we utilize the
+    [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown) serializer and parser to convert the text into markdown format and vice
+    versa. However, the supported functionality of prosemirror-markdown, as mentioned in their
```
vikisekarNI, line 54:
```suggestion
-   Support for striking out and underlining text. We use
    [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown) serializer and parser to convert the text into markdown format and vice
    versa. However, the supported functionality of prosemirror-markdown, as mentioned in their
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,373 @@
...
+    [documentation](https://github.com/ProseMirror/prosemirror-markdown#documentation), does not include support for strikeout and underline. To
+    address this limitation, we need to extend the class and add the necessary nodes to both the serializer and parser which will be added in the
```
vikisekarNI, line 56: Is there any tiptap/prose mirror documentation that instructs on applying these extensions for strikeout and underline behaviors? If yes, it will be good to link it here.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,373 @@
...
+_Methods_
+
+-   `setMarkdown()` - sets the markdown input from the consumer component which can be used for preloading the content.
+-   `getMarkdown()` - gets the processed markdown output from the component.
+
+_Events_
+
+-   `change` - event emitted when there is a change in the the editor. This can be achieved through tiptap's
+    [update event](https://tiptap.dev/api/events#update). Below is a scenarios of event triggers for the update event, indicating when they do and do not
+    occur:
```
vikisekarNI, line 142:
```suggestion
    [update event](https://tiptap.dev/api/events#update). Below are few scenarios to understand when an update event will trigger or not trigger:
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,373 @@
...
+strings may be used for tooltips to enable localization, which will be managed through the `nimble-tooltip` component.
+
+### Security
+
+-   Prose mirror uses markdown-it for converting markdown to HTML and HTML to markdown. We will follow the
+    [security guidelines of markdown-it](https://github.com/markdown-it/markdown-it/blob/master/docs/security.md#security) to turn
+    off HTML at source as given in the [API docs](https://markdown-it.github.io/markdown-it/#MarkdownIt.new). Prosemirror-markdown follows the same as shown in this
+    [specific line of code](https://github.com/ProseMirror/prosemirror-markdown/blob/26e58302399b7d9a9b3bc8fc3bf5807627ca29e5/src/from_markdown.ts#L245).
+-   For additional safety, using [sanitize-html](https://www.npmjs.com/package/sanitize-html) package to whitelist only specific html
+    tags on need basis or completely disallow any html tags.
```
vikisekarNI, line 337:
```suggestion
-   For additional safety we may use [sanitize-html](https://www.npmjs.com/package/sanitize-html) package on need basis if HTML is turned on. This will whitelist only specific html
    tags needed for rich text markdown. 
```

vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+| Tab                     | To reach the editor                                      |
+| Shift + Arrow keys      | To select the group of texts in the editor               |
+| Ctrl + B                | To make the selected text bold                           |
+| Ctrl + I                | To make the selected text italic                         |
+| Ctrl + Shift + 7        | To enable the focused paragraph a numbered list          |
+| Ctrl + Shift + 8        | To enable the focused paragraph a bulleted list          |
+| Tab (For lists)         | To create a sub point within a bulleted or numbered list |
+| Shift + Tab (For lists) | To remove a sub point from a bulleted or numbered lists  |
+
+_Keyboard navigation with toolbar buttons focused_
```
mollykreis, line 318: Switching between the buttons in the toolbar is missing here. For example, moving from the "bold" button to the "italic" button. Is that accomplished by pressing "Tab" or by pressing the "Right Arrow" key? I think it should be using the arrow keys, which means if you use the `nimble-toolbar` as I mentioned in another comment, you'll get that behavior for free.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+| Key          | Behavior                                                                       |
+| ------------ | ------------------------------------------------------------------------------ |
+| Space, Enter | Enable the selected text formatting feature and return the focus to the editor |
+| Shift + Tab  | To reach the editor back to focus                                              |
+
+_Note_: All other keyboard interaction determined by the slotted element will not be defined in this document.
+
+### Globalization
+
+Currently, this component does not have any string data to be localized. In the footer menu for text formatting, only icons will be displayed. However,
```
mollykreis, line 329: The `nimble-tooltip` is an experimental component, so I wouldn't recommend using it here.

Also, I think it should be a requirement to have tooltips (we've been using the `title` attribute to accomplish this) and accessible text for each icon button. There is a spec in progress for how to configure localizable strings on a component, so once the infrastructure work is complete, I'd expect this component to plug into the shared mechanism for providing localized strings. See #1272

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -17,32 +17,30 @@ including Comments and other instances that necessitate rich text capabilities.
...
 -   Blazor integration will be out of scope for this component as we have less or no experience in Blazor component development or other
     related technology. Blazor example app addition will also be not covered in this spec.
--   Adding an example Angular client application for this component may not be included in the initial releases. However, the integration
-    of the rich-text-editor with Angular applications, achieved through a directive, will be developed following the completion of this
-    component.
 
 ### Features
 
-The `nimble-rich-text-editor` supports the basic text formatting options that includes:
+Both the components provides support for the basic text formatting options as follows:
```
mollykreis, line 29: nit:
```suggestion
Both components provide support for the following basic text formatting options:
```

change/@ni-nimble-components-02d33e01-7578-482d-aa7f-8f592617a882.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add spec document for nimble rich text editor and nimble rich text viewer",
+  "packageName": "@ni/nimble-components",
+  "email": "vivin.krishna.r.n@ni.com",
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks, Jesse! I have updated my email in this change file.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -17,32 +17,30 @@ including Comments and other instances that necessitate rich text capabilities.
...
 
 1. Bold
 2. Italics
 3. Numbered List
 4. Bulleted List
 
-This component will also offer APIs and interactive methods to format texts in the following ways:
+The `nimble-rich-text-editor` component will also offer APIs and interactive methods to format texts in the following ways:
```
mollykreis, line 36: nit:
```suggestion
The `nimble-rich-text-editor` component will also offer APIs and interactive methods to format text in the following ways:
```

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+[Comments UI mockup](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045)
+
+### Non-goals
+
+-   The visual design and interaction design for these components are out of scope for the initial release. However we have an
+    [IxD workflow](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045) for SLE application that we will rely
+    on.
+-   Blazor integration will be out of scope for this component as we have less or no experience in Blazor component development or other
+    related technology. Blazor example app addition will also be not covered in this spec.
+-   Adding an example Angular client application for this component may not be included in the initial releases. However, the integration
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for letting us know. We will add these components to the example client app once the Angular integration is completed. I have now removed this point from the Non-Goals.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
+# Nimble Rich Text Editor and Nimble Rich Text Viewer
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the suggestion. We went up with option 2 by having the actual spec in `rich-text-editor` and `rich-text-viewer` spec have the link to the actual spec.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+
+Collaboration is a key requirement in order for users to adopt a tool at scale and would help improve overall customer satisfaction with the product.
+The Comment Feature, designed for adding and viewing comments with rich text content, should be generic enough to be reused in several workflows.
+Therefore, it is essential to develop web components that allow users to view and create rich text content, enabling their use in various scenarios,
+including Comments and other instances that necessitate rich text capabilities.
+
+[Nimble issue #1288](https://github.com/ni/nimble/issues/1288)
+
+[Comments UI mockup](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045)
+
```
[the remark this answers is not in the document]
vivinkrishna-ni: Added AzDo Comments Epic link here.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+_nimble-rich-text-editor_
+
+The `nimble-rich-text-editor` will be divided into two sections namely an `editor` section and a `footer` section.
+
+1. `editor` section is the actual text area to add or update rich text contents.
+2. `footer` section consists of `nimble-button` to control each text formatting functionalities like bold, italic etc,. and a
+   `footer-actions` slot element which is typically used to add action buttons to the right bottom of the component.
+
+Here is the shadow DOM template used in `nimble-rich-text-editor` component:
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have moved this to the Anatomy section.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+-   The visual design and interaction design for these components are out of scope for the initial release. However we have an
+    [IxD workflow](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-94045) for SLE application that we will rely
+    on.
+-   Blazor integration will be out of scope for this component as we have less or no experience in Blazor component development or other
+    related technology. Blazor example app addition will also be not covered in this spec.
+-   Adding an example Angular client application for this component may not be included in the initial releases. However, the integration
+    of the rich-text-editor with Angular applications, achieved through a directive, will be developed following the completion of this
+    component.
+
+### Features
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have added this point in the feature section and linked the markdown flavor we support.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Thanks for the information on localization. I have updated the same in the spec.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+-   none
+
+_Methods_
+
+-   `setMarkdown()` - sets the markdown input from the consumer component which can be used for preloading the content.
+-   `getMarkdown()` - gets the processed markdown output from the component.
+
+_Events_
+
+-   `change` - event emitted when there is a change in the the editor. This can be achieved through tiptap's
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have updated this event name to `input` which makes more sense compared to the `change` event as it fires for every input when typing. I have updated this accordingly in the spec.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+6. `Top layer` support - Hyperlink using `nimble-dialog` which renders the popup in top layer.
+
+**_Note_**: This prototype is not made in a responsive fashion for smaller screens and is intended only for testing the specific functionalities listed
+above.
+
+### States
+
+_Button State_
+
+The buttons in the toolbar will indicate the formatting status of the selected text in the editor. For example, if the text in the editor
```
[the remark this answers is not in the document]
vivinkrishna-ni: @mollykreis, Initially, when developing the prototype for this component, we tried using the `nimble-toggle-button` but in some cases it does not behave as expected. We wanted the button to toggle the state only based on the condition we set in `checked` Boolean attribute of the component not changing the state by clicking on the button. In one of the scenario, when we were having focus on the sub point of numbered or bulleted list (image 1), clicking the button should not toggle its state as condition attached to `?checked` attribute is still `true`. But it changes its state to `false` (Image 2). We used `nimble-toggle-button` for numbered list in the below snapshots.
image
Not expected:
image

Whereas, when trying with the `nimble-button` by toggling the `appearance` we can achieve it as it does not toggle the appearance on click. We used `nimble-button` for bulleted list in the below snapshots.
image
Expected:
image

Is this expected for `nimble-toggle-button` to change its state on click and not based on the condition? If yes, can we go with the `nimble-button` approach?

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: @mollykreis, We have tried using the `nimble-toolbar` in our prototype and here are some of the problems faced in the keyboard accessibility,
1. Consider having the focus in the editor text area and by pressing "Tab" will move the focus to icon buttons in the toolbar. Now by pressing "Arrow keys" to move the focus between formatting buttons. We also have `footer-actions` slot element in the same toolbar at the right end, which we specifically added for the comments feature ([mockup](https://www.figma.com/proto/Q5SU1OwrnD08keon3zObRX/SystemLink?type=design&node-id=6280-95369&scaling=min-zoom&page-id=2428%3A32954&starting-point-node-id=6280%3A94118&show-proto-sidebar=1)) to have `Cancel` and `Add comment` button. To reach the buttons in the slot, we can only use "Tab" because both the buttons are slotted into `slot = footer-actions` and the `footer-actions` is slotted into `slot = end` of nimble-toolbar. Is this making inconsistent in moving the focus by using "Arrow keys" for formatting options and "Tab" for slotted buttons which in the same toolbar?
2. From the comments feature point of view, even after we change the focus from formatting options (icon buttons) to `footer-actions` buttons by pressing "Tab", we couldn't move forward once after that. For an instance, consider the focus in the formatting options, pressing "Tab" moves the focus to `cancel` button and pressing "Tab" again, moves the focus back to the formatting options and not the `Add comment` button. 

Please let us know if there are any alternative design that we should consider to resolve these issue or is it fine to use the section or div tag to achieve the keyboard accessibility through "Tab" in all places.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
```diff
@@ -0,0 +1,377 @@
...
+   `footer-actions` slot element which is typically used to add action buttons to the right bottom of the component.
+
+Here is the shadow DOM template used in `nimble-rich-text-editor` component:
+
+```html
+<template>
+    <section id="editor"></section>
+    <footer>
+        <section id="toolbar">
+            <nimble-button></nimble-button>
```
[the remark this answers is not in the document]
vivinkrishna-ni: Resolving this comment as I have replied to the other comment on the usage of `nimble-toolbar`.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for sharing the `group-row` component as an example for resolving this issue we were able to quickly check it in our end and it worked. So we can go with the `nimble-toggle-button` for the state changes in the toolbar. I have updated the spec accordingly. Adding here the link for the reference:

- template: [nimble/packages/nimble-components/src/table/components/group-row/template.ts at main · ni/nimble · GitHub](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/components/group-row/template.ts)
- class: [nimble/packages/nimble-components/src/table/components/group-row/index.ts at main · ni/nimble · GitHub](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/components/group-row/index.ts)

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
vivinkrishna-ni: Thanks for the detailed information, we will use `nimble-toolbar` to wrap all the formatting options and wrap this toolbar and a slot element into another section to maintain it in a flexbox to stick with the mockups. I have updated the same in the [shadow DOM template](https://github.com/ni/nimble/blob/rich-text-editor-viewer-spec/packages/nimble-components/src/rich-text-editor/specs/README.md#anatomy) section of this spec. I have also updated the key accessibility which is to have "arrow keys" for the toolbar and "Tab" to switch to footer-action buttons.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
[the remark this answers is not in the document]
vivinkrishna-ni: @jattasNI and Team, we have added the `isEmpty` attribute as part of the event details to dynamically update the enable state of the "Add comment" button. 

We have referred to the `nimble-banner` [toggle](https://github.com/ni/nimble/blob/rich-text-editor-viewer-spec/packages/nimble-components/src/banner/index.ts#L66) event on how to set the details object in the event that is fired and in which they were using something like `newState`, `oldState` to check whether it is open or not. 

Please let us know your thoughts on this approach.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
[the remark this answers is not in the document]
vivinkrishna-ni: Added the `nimble-toolbar` for formatting options as suggested.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+1. `editor` section is the actual text area to add or update rich text content.
+2. `footer` section consists of `nimble-button` to control each text formatting functionalities like bold, italic, etc, and a
+   `footer-actions` slot element which is typically used to add action buttons to the right bottom of the component.
+
+Example usage of the `nimble-rich-text-editor` in the application layer is as follows:
+
+```html
+<nimble-rich-text-editor>
+    <div slot="footer-actions">
```
mollykreis, line 97: Is it expected that a client would provide a `div` containing multiple buttons like this, or would they do something more like:
```
<nimble-rich-text-editor>
    <nimble-button slot="footer-actions>Cancel</nimble-button>
    <nimble-button slot="footer-actions>Add Comment><nimble-button>
</nimble-rich-text-editor>
```

In the example you provide, it would be the responsibility of a client to space the buttons appropriately. However, in the example I provided, the nimble component could add the appropriate spacing between each component placed in the `footer-actions` slot.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+-   none
+
+_Events_
+
+-   none
+
+_CSS Classes and CSS Custom Properties that affect the component_
+
+-   The height of the component will always grow to fit the content.
```
jattasNI, line 208: It would be good to clarify: clients can still set the width? And this setting could affect the height (a smaller width will cause the content to reflow and take up more height)?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+An Angular directive will be created for both components. The components will not have form association, so a `ControlValueAccessor` will not be created.
+
+### Blazor integration
+
+A Blazor wrapper is initially out of scope for these components.
+
+### Visual Appearance
+
+NA
```
jattasNI, line 239:
```suggestion
Visual design has been requested is currently still pending.
```

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+-   Bold - `**Bold**`
+-   Italics - `*Italics*`
+-   Numbered list - `1. Numbered list`
+-   Bulleted list - `* Bulleted list`
+
+The `nimble-rich-text-viewer` will be responsible for converting the input markdown string to HTML Fragments with the help of
+`prosemirror-markdown` parser, which is then converted to HTML string and rendered into the component to view all rich text content.
+
+### Prototype
```
jattasNI, line 280: It'll be good to have a discussion sometime about the process for converting this prototype into production code. I don't think we want one big PR for the entire feature at once. Rather it would be good to start small with a vertical slice and incrementally add functionality. This can help ensure better code quality and test coverage.

No spec updates are needed for this, just want to put it on your radar as a future discussion point.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+-   Focus state of the editor will be the same as the `nimble-text-area`.
+-   Focus state of the buttons will be the same as the `nimble-button`.
+
+_Keyboard accessibility and shortcuts for text formatting_
+
+All the basic keyboard accessibility for the tiptap rich text editor is mentioned in their
+[keyboard shortcut](https://tiptap.dev/api/keyboard-shortcuts#predefined-keyboard-shortcuts) page. To facilitate easy access, listed down
+the supported text formatting keyboard shortcuts in the tiptap editor.
+
+| Key                     | Behavior                                                 |
```
jattasNI, line 327: Double checking: does TipTap automatically support Command instead of Ctrl for the shortcuts on macOS?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Performance
+
+Performance of tiptap explained in an example of having [long texts](https://tiptap.dev/examples/book).
+
+### Dependencies
+
+This component is dependent on the [`tiptap`](https://tiptap.dev/) third party library. As mentioned,
+[tiptap](https://tiptap.dev/introduction#why-should-i-use-tiptap) is built on top of prosemirror, we might also use some of the prosemirror's
+library. For the currently supported features, we will include the following libraries that will be added to the packages.json
```
jattasNI, line 374:
```suggestion
library. For the currently supported features, we will include the following libraries that will be added to the package.json
```

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: I have updated the spec as per the example you have showed. The standard spacing between the slotted elements will be handled by the component in `footer-actions` sections and other styling will be handled by the client.

packages/nimble-components/src/rich-text-editor-viewer/specs/README.md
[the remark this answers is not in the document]
vivinkrishna-ni: Resolving this comment as we have the same conversation in the other thread.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+    1. This event will fired for every input in the content of the editor, including text inputs, text formatting changes, and text removals.
+    2. This event will not fire when there are no changes made to the content of the editor. For example, all mouse events, selecting the texts, state
+       changes, etc,
+
+    The emitted event data is in the format of a `CustomEvent` and consists of a `detail` object with an `isEmpty` boolean flag indicating whether the editor
+    is empty or not.
+
+_CSS Classes and CSS Custom Properties that affect the component_
+
+-   The minimum and maximum height of the `editor section` will be fixed value. If the content exceeds the maximum height, we will show the vertical scrollbar to
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have updated it accordingly in the spec to have the minimum height as per the suggestion from Jesse in the other comment and the minimum width based on to accommodate all buttons being visible in the footer. 

The maximum width and height will not be fixed in the component and will be configurable by the client.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+    view the content of the editor.
+-   The minimum and maximum width of the `editor section` will be fixed value considering the mobile and desktop view of the component respectively. However, if the
+    screen width is reduced
+    below the specified minimum width, the layout may not handle or adapt to the content.
+-   The `footer section` will be a flexbox container and have the same width as the editor section.
+-   The `formatting option toolbar` in the footer section will be enclosed within a flexbox container, enabling the buttons to wrap and occupy roughly seventy
+    percent of the entire footer. If the buttons exceed the seventy percent limit, they will be positioned below in the same container, aligning from left
```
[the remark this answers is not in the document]
vivinkrishna-ni: This will be applicable only if we set the maximum height in the component and not in the client. Wrapping the buttons and adding them to the other row will increase the height of the component. So, we thought of having a minimum width for the entire component based on the buttons available in the footer (formatting options + slot elements). For the initial release, we will consider having the four buttons space for the formatting options and the remaining space for the slot as mentioned in the spec.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,401 @@
...
+_Events_
+
+-   `input` - event emitted when there is a change in the editor. This can be achieved through Tiptap's [update event](https://tiptap.dev/api/events#update).
+    Below are a few scenarios to understand when this event will fire:
+
+    1. This event will fired for every input in the content of the editor, including text inputs, text formatting changes, and text removals.
+    2. This event will not fire when there are no changes made to the content of the editor. For example, all mouse events, selecting the texts, state
+       changes, etc,
+
+    The emitted event data is in the format of a `CustomEvent` and consists of a `detail` object with an `isEmpty` boolean flag indicating whether the editor
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for bringing up this use case. I have updated `isEmpty` as a read-only property as suggested and removed it from the event data.

packages/nimble-components/src/rich-text-editor/specs/README.md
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the detailed information. I have updated the spec as per the suggestion. We will get started with this approach for the initial release and as said as per the visual design from Rick, we will incorporate the changes during the implementation or in the subsequent releases.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Sure, we will plan to discuss this once we start with the implementation and make sure to have a more granular PR for the entire component development.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Yes, it does support `Cmd` in macOS, and here is the official page of Tiptap for [keyboard shortcuts](https://tiptap.dev/api/keyboard-shortcuts#predefined-keyboard-shortcuts). I have added this point as well in the spec.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Yes, it will increase the height if the width is small. I have updated the same in the spec.

mollykreis (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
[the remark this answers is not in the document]
jattasNI: Looking good, thanks for updating. A couple last minor points:

>The client will determine the default and maximum height and width of the component.

Nimble should provide a default height and width for the case where the element is placed on the page with no size styling. This can just be a reasonable fixed pixel size, as long as the client can override it.

Also, a possible nice-to-have feature for the editor component would be to use the [built-in resize attribute](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) to allow the user to dynamically grow the height of the text input box (like GitHub allows). I bet it would be trivial to implement, but it's just an idea so no need to capture it if you're worried about scope.

image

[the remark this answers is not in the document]
suseendran-ni: Thanks for suggesting this @jattasNI. This new feature looks great. And from your reference, as you mentioned it seems easy to integrate. Can we please consider this after completing all the required features? As I am not sure of the development effort to write test scripts for this feature, it would be easy for us to analyze this once after making good progress on the development.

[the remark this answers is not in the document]
jattasNI: Definitely, that makes sense.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,405 @@
...
+       changes, etc,
+
+_CSS Classes and CSS Custom Properties that affect the component_
+
+-   The minimum height of the component will be set to show at least one line in the text area along with a footer section. If the content exceeds
+    the current height (set by the client), a vertical scrollbar will appear, allowing users to view the hidden content.
+-   The minimum width of the component will be determined based on ensuring all buttons in the footer are visible. We have arrived to this decision
+    to align with the minimal requirement for the initial release. However, adjustments will be made later based on the visual design for mobile view is complete.
+-   The content in the text area will adjust its layout based on the width, potentially increasing the height of the editor. But in case of having
+    a long nested lists that is beyond the current width (set by the client), a horizontal scrollbar will appear within the text area to view the content.
```
rajsite, line 134: Is this saying the control will always grow vertically to fit the content that is being edited without making a vertical scrollbar?

I think we should be using the native textarea as an example of the default behavior we should follow. If I understand correctly that the intention is the control will grow to fit the content I think that should be an optional configuration. Maybe a `<nimble-rich-text-editor auto-size>` or a `<nimble-rich-text-editor fit-to-content>` option

This applies to the viewer element as-well

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,405 @@
...
+
+-   none
+
+_Events_
+
+-   none
+
+_CSS Classes and CSS Custom Properties that affect the component_
+
+-   The height of the component will always grow to fit the content.
```
rajsite, line 208: Yea as mentioned above I think growing the height should be opt-in. It should have the same sizing behavior as the editor.

rajsite (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -0,0 +1,410 @@
...
+### API
+
+_Props/Attrs_
+
+-   `markdownValue` - is an accessor used to get and set the markdown value.
+    -   `getter` - this will serialize the content by extracting the Node from the editor and convert it into a markdown string using
+        [prosemirror-markdown serializer](https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/to_markdown.ts#L30).
+    -   `setter` - this will parse the markdown string into a Node and load it back into the editor using
+        [prosemirror-markdown parser](https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/from_markdown.ts#L199).
+-   `isEmpty` - is a read-only property that indicates whether the editor is empty or not. This will be achieved through Tiptap's
```
atmgrifter00, line 111: Not sure how critical it is to capture Angular implementation details in this spec, but I think it's at least worth mentioning how we would accomplish a bindable, read-only property in the Angular API. We would likely need to follow the pattern of the [`openChange` `Output` of the `NimbleBannerDirective`](https://github.com/ni/nimble/blob/1d5ab5f6f23e11e7b7c30600c196a04ea087c5af/angular-workspace/projects/ni/nimble-angular/src/directives/banner/nimble-banner.directive.ts#L25), though without the need for the `Change` suffix, as there is no two-way binding concern. Doing this requires handling an event from the underlying component in order to emit the event from the directive (as done [here](https://github.com/ni/nimble/blob/1d5ab5f6f23e11e7b7c30600c196a04ea087c5af/angular-workspace/projects/ni/nimble-angular/src/directives/banner/nimble-banner.directive.ts#L67) by the `NimbleBannerDirective`).

atmgrifter00 (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: For `isEmpty`, we have referred to the [read-only property `open`](https://github.com/ni/nimble/blob/1d5ab5f6f23e11e7b7c30600c196a04ea087c5af/packages/nimble-components/src/dialog/index.ts#L81) of the nimble dialog ([spec reference](https://github.com/ni/nimble/blob/1d5ab5f6f23e11e7b7c30600c196a04ea087c5af/packages/nimble-components/src/dialog/specs/README.md?plain=1#L63)) which returns true if the dialog is open. The Angular implementation of the `isEmpty` will be same as the [`open` attribute of the `NimbleDialogDirective`](https://github.com/ni/nimble/blob/1d5ab5f6f23e11e7b7c30600c196a04ea087c5af/angular-workspace/projects/ni/nimble-angular/src/directives/dialog/nimble-dialog.directive.ts#L51). The `isEmpty` getter method in the component simply returns the Tiptap's [isEmpty](https://tiptap.dev/api/editor#is-empty) value which will be true if the editor is empty.

With this approach, we can directly bind the `isEmpty` in the HTML of the Angular application. As an example, in the comments feature, we have the requirement to disable the `Comment` button when the editor is empty. So we can do something like the below code in the client to enable or disable the button.
```html
<nimble-rich-text-editor #editor>
    <nimble-button slot="footer-actions">Cancel</nimble-buttont>
    <nimble-button slot="footer-actions" [disabled]="editor.isEmpty">Comment</nimble-buttont>
</nimble-rich-text-editor>
```

So we thought of adding it as a read-only property rather than an event. We will update the spec to mention the Angular implementation for this API.

jattasNI:
Since this is your first PR, I'll just mention: now that I'll the Nimble reviewers have approved you're welcome to complete it yourself once all the status checks are passing. Looks like the build just finished and the last comment thread is almost resolved, so you should be just about good to go! 🚀

packages/nimble-components/src/rich-text-editor/specs/README.md
jattasNI: Resolving this thread since I think it's a sufficient level of detail for the spec. @atmgrifter00 mentioned offline that he still wanted to understand this a bit better, but it doesn't need to block initial submission (we can always update the spec if our plans change).

rajsite:
> Since this is your first PR, I'll just mention: now that I'll the Nimble reviewers have approved you're welcome to complete it yourself once all the status checks are passing. Looks like the build just finished and the last comment thread is almost resolved, so you should be just about good to go! 🚀

@vivinkrishna-ni As Jesse mentioned once the status checks are complete and you have the owners you need you can merge the PR. Is there anything y'all are waiting on?

suseendran-ni:
> > Since this is your first PR, I'll just mention: now that I'll the Nimble reviewers have approved you're welcome to complete it yourself once all the status checks are passing. Looks like the build just finished and the last comment thread is almost resolved, so you should be just about good to go! 🚀
> 
> @vivinkrishna-ni As Jesse mentioned once the status checks are complete and you have the owners you need you can merge the PR. Is there anything y'all are waiting on?

suseendran-ni:
> > Since this is your first PR, I'll just mention: now that I'll the Nimble reviewers have approved you're welcome to complete it yourself once all the status checks are passing. Looks like the build just finished and the last comment thread is almost resolved, so you should be just about good to go! 🚀
> 
> @vivinkrishna-ni As Jesse mentioned once the status checks are complete and you have the owners you need you can merge the PR. Is there anything y'all are waiting on?

@rajsite , as @RickA-NI is working on the visual designs for the comments feature, we have planned to analyze that and updated any content if required. 

Will it be good to submit these changes now and initiate another PR in the future if any change needs to be done in the document? Please advice.

jattasNI:
> Will it be good to submit these changes now and initiate another PR in the future if any change needs to be done in the document? Please advice.

Yes, that would be a good way to do it, especially since #1314 is currently waiting on this.

suseendran-ni:
> 



> > Will it be good to submit these changes now and initiate another PR in the future if any change needs to be done in the document? Please advice.
> 
> Yes, that would be a good way to do it, especially since #1314 is currently waiting on this.

Sure @jattasNI , I will submit this HLD.

rajsite:
@suseendran-ni you stated you will merge the PR but then closed it.
image

If you want to merge it you need to enable auto-merge / click merge, not close.
image

Did you mean to close the PR or merge the PR?

PR #1291, opened 2023-06-08
Fix inconsistencies in combobox filtering

### LINKED ISSUE #1290: combobox clears options filter too aggressively
# 🐛 Bug Report

The change to the combobox made in #1270 to fix #1266 now clears the options filter too aggressively. As a result, there are cases where the options drop-down shows all options instead of a filtered set of options when it used to be filtered.

Two of these cases that I've identified are:
- Clicking an option in the drop-down list and then reopening it
- Blurring the combobox list and then reopening it

## 💻 Repro or Code Sample

Scenario 1:
1. Go to the [storybook page for the combobox](https://nimble.ni.dev/storybook/?path=/story/combobox--underline-combobox)
2. Click on the combobox  to focus it
3. Type "Mary" and press Enter
4. Click on the combobox to re-open the drop-down list
5. Notice the drop-down list is filtered to only include "Mary"
6. Click "Mary" in the options list
7. Click on the combobox to re-open the drop-down list
8. Notice the drop-down list is no longer filtered and includes all options

Scenario 2:
1. Go to the [storybook page for the combobox](https://nimble.ni.dev/storybook/?path=/story/combobox--underline-combobox)
2. Click on the combobox to focus it
3. Type "Mary" and press Enter
4. Click off the combobox to cause it to lose focus
7. Click on the combobox to re-open the drop-down list
8. Notice the drop-down list is no longer filtered and includes all options

## 🤔 Expected Behavior

Prior to #1270, the combobox filter wasn't cleared in the cases described above. The behavior can be seen [in this older storybook page](https://60e89457a987cf003efc0a5b-zlisnnsntl.chromatic.com/?path=/story/combobox--underline-combobox).

## 😯 Current Behavior

See steps to reproduce above

## 🌍 Your Environment

* Version [e.g. 1.8.0]

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1290 

## 👩‍💻 Implementation

Now unconditionally set the filter to the new value in the `value` setter.

## 🧪 Testing

Manually tested in Storybook and wrote two new unit tests (plus changed one existing unit test).

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/combobox/index.ts
+++ b/packages/nimble-components/src/combobox/index.ts
@@ -79,16 +79,14 @@ export class Combobox
     // See: https://github.com/microsoft/fast/issues/6749
     public override set value(next: string) {
         super.value = next;
-        if (!this.valueUpdatedByInput) {
-            // Workaround using index notation to manipulate private member
-            // Can remove when following resolved: https://github.com/microsoft/fast/issues/6749
-            // eslint-disable-next-line @typescript-eslint/dot-notation
-            this['filter'] = '';
-            this.filterOptions();
-            this.selectedIndex = this.options
-                .map(option => option.text)
-                .indexOf(this.value);
-        }
+        // Workaround using index notation to manipulate private member
+        // Can remove when following resolved: https://github.com/microsoft/fast/issues/6749
+        // eslint-disable-next-line @typescript-eslint/dot-notation
+        this['filter'] = next;
+        this.filterOptions();
+        this.selectedIndex = this.options
+            .map(option => option.text)
+            .indexOf(this.value);
     }
 
     private valueUpdatedByInput = false;
--- a/change/@ni-nimble-components-c975dc40-fb4a-45aa-acd8-a57ed2cd9aa3.json
+++ b/change/@ni-nimble-components-c975dc40-fb4a-45aa-acd8-a57ed2cd9aa3.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix inconsistencies in combobox filtering",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/combobox/tests/combobox.spec.ts
+++ b/packages/nimble-components/src/combobox/tests/combobox.spec.ts
@@ -382,7 +382,7 @@ describe('Combobox', () => {
         await disconnect();
     });
 
-    it('clears old filter when value set programmatically', async () => {
+    it('updates filter when value set programmatically', async () => {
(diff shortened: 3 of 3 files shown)


jattasNI (commented):
I played around with Storybook and it felt consistent to me, but I defer to @atmgrifter00 on this one.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -79,16 +79,14 @@ export class Combobox
     // See: https://github.com/microsoft/fast/issues/6749
     public override set value(next: string) {
```
mollykreis, line 80: An oddity that I discovered is that an option in the drop-down list isn't always shown as selected anymore even though the text value matches one of the options.

For example:
- Set the value to _Sue Ann_
- Click off & back on the combobox to open the dropdown again
- Click _Backspace_ a few times until the text value is only _Sue_
- Notice that _Sue_ in the dropdown flickers as green but immediately changes back to white (i.e. not selected)

m-akinc: I can reproduce this in the published Storybook as well. It loses the selection when using a number of different keys (e.g. arrowing through). It's because of this FAST code: https://github.com/microsoft/fast/blob/8c80dfcf9361a5390806c3a55b67c43259a76fb8/packages/web-components/fast-foundation/src/combobox/combobox.ts#L530.

I've overriden this method and removed the line that clears the selected index. Our tests still pass, but I can't be totally sure we haven't introduced some other weird behavior by doing this.

mollykreis: Thanks for looking into it. I'd say that if it's existing behavior, we shouldn't try to address it in this PR. If we think it should be addressed, we can file an issue for it and prioritize it separately.

m-akinc: Ok, reverted that change.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
```diff
@@ -7,7 +7,7 @@ import { processUpdates, waitForUpdatesAsync } from '../../../testing/async-help
 import type { Combobox } from '../nimble-combobox.directive';
```
jattasNI, line 7: Could you add a note to the PR description explaining the CVA part of the change?

m-akinc: Done

rajsite (approved):

mollykreis (approved):

jattasNI (approved):

atmgrifter00 (approved):

PR #1249, opened 2023-05-16
Update guidance on experimental components

jattasNI:
# Pull Request

## 🤨 Rationale

We had a team discussion about encouraging more Nimble contributions for things like app-specific table columns. This update captures policies for those components in CONTRIBUTING.md

## 👩‍💻 Implementation

Update the "Experimental" section of CONTRIBUTING.md

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1292, opened 2023-06-12
Follow new guidance for incubating components, mark table NOT incubating, add sections to Storybook

jattasNI:
# Pull Request

## 🤨 Rationale

Following the guidance agreed upon in #1249, we want to promote the "Incubating" terminology to describe components that aren't ready for general use.

## 👩‍💻 Implementation

1. Add sections to Storybook for **Incubating** and **Components**. The latter was proposed in #1230 so it seems like a good direction for multiple reasons.
2. Add MDX docs for the Incubating section.
3. Add a link to the corresponding component issue from incubating stories. Update tooltip to use standard incubating header.
4. Rename WaferMap stories to "Wafer Map" to match other components which English naming not code naming (and since this change was already going to cause a Chromatic diff)

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/utilities/tests/storybook.ts
+++ b/packages/nimble-components/src/utilities/tests/storybook.ts
@@ -146,14 +146,17 @@ Overrides of properties are not recommended and are not theme-aware by default. 
 ${howToOverride}
 </details>`;
 
-export const usageWarning = (componentName: string, issueLink: string): string => `
+export const incubatingWarning = (
+    componentName: string,
+    issueLink: string
+): string => `
 <style class="code-hide">
-#usage-warning {
+#incubating-warning {
     color: red;
     font: var(${bodyFont.cssCustomProperty});
 }
 </style>
-<div id="usage-warning" class="code-hide">
+<div id="incubating-warning" class="code-hide">
 WARNING - The ${componentName} is still incubating. It is not recommended for application use. 
 For more information on its status, see <a href="${issueLink}">this issue</a>.
 </div>`;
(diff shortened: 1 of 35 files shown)


packages/nimble-components/src/utilities/tests/storybook.ts
```diff
@@ -146,16 +146,19 @@ Overrides of properties are not recommended and are not theme-aware by default.
 ${howToOverride}
 </details>`;
 
-export const usageWarning = (componentName: string): string => `
+export const incubatingWarning = (
+    componentName: string,
+    statusLink: string
+): string => `
```
rajsite, line 152: nit / pref: could you make this a configuration object instead? Will make it easier to understand the params if the list grows

packages/nimble-components/src/utilities/tests/storybook.ts
```diff
@@ -146,16 +146,19 @@ Overrides of properties are not recommended and are not theme-aware by default.
...
     color: red;
     font: var(${bodyFont.cssCustomProperty});
 }
 </style>
-<div id="usage-warning" class="code-hide">
-WARNING - The ${componentName} is still in development and considered
-experimental. It is not recommended for application use.
+<div id="incubating-warning" class="code-hide">
+WARNING - The ${componentName} is still incubating. It is not recommended for application use. 
+For more information on its status, <a href="${statusLink}">see here</a>.
```
rajsite, line 161: This makes the actual link text for screen readers, etc just "see here".
Can you make it so the link is more descriptive?

```suggestion
See the <a href="${statusLink}">incubating component status</a>.
```

rajsite (approved):

PR #1295, opened 2023-06-12
Wafer Map update tracker

munteannatan:
# Pull Request

## 🤨 Rationale

The wafer map component needs to implement the update tracker for its inputs and to have a more fragmented update 

## 👩‍💻 Implementation

Implemented the update tracker for the wafer map inputs.

Split apart each module and created update functions so that it doesn't have to be reinitialized at every update.

## 🧪 Testing

Existing tests were updated where needed, no new tests were added.

## ✅ Checklist


change/@ni-nimble-components-b6bd9b08-a8e3-4e8f-bab1-7da899d6fece.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Implemented the update tracker for the wafermap and split the rendere in the new update functions for each module",
```
jattasNI, line 3:
```suggestion
  "comment": "Implemented the update tracker for the wafermap and split the render in the new update functions for each module",
```

jattasNI (commented):
I did some basic interaction testing with Storybook and left a couple comments, but similar to the previous update tracker PR I'll defer to others on the detailed review.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -153,34 +155,87 @@ export class WaferMap extends FoundationElement {
...
-        if (!this.$fastController.isConnected) {
-            return;
-        }
-        if (!this.renderQueued) {
-            this.renderQueued = true;
-            DOM.queueUpdate(() => this.render());
+    /**
+     * @internal
+     */
+    public update(): void {
```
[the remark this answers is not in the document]
jattasNI: Part of my question was me not having taken the time to understand how the update tracker works yet. I took a bit closer look and saw that each property update only triggers one `requires***Update` accessor to be `true`. If that remains the case then I think this implementation is fine.

But suppose that changing the `dies` property actually required both `DiesRenderInfoUpdate` and `ScalesUpdate` (just a made up example). If `update()` is only called once then it will only process one of those two changes.

Assuming that's accurate and is a use case we want to support, it would be easy to fix by changing this method to use chained `if`s instead of `else if`s. But that's just an idea: I could be misunderstanding the intended use case.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -153,34 +155,87 @@ export class WaferMap extends FoundationElement {
...
+        if (this.waferMapUpdateTracker?.requiresContainerDimensionsUpdate) {
+            this.eventCoordinator?.detachEvents();
```
DStavilaNI, line 175: Feels like repetitive code -  `this.eventCoordinator?.detachEvents()` / `this.eventCoordinator?.updateEvents()` . I think that these could be pulled out and called outside the `else if` chain.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -202,56 +257,64 @@ export class WaferMap extends FoundationElement {
...
-
-    private orientationChanged(): void {
-        this.queueRender();
+        this.waferMapUpdateTracker?.track('quadrant');
+        this.waferMapUpdateTracker?.queueUpdate();
     }
 
     private maxCharactersChanged(): void {
-        this.queueRender();
+        this.waferMapUpdateTracker?.track('maxCharacters');
```
DStavilaNI, line 265: I am not very familiar with the `updateTracker`, but do we have to call the `track` method for each property/attribute change event ? This seems to be more like an initialization that has to be done just once ...

munteannatan: when creating the tracker, we define which properties will be tracked. on every update we set which properties have changed, run the update specific to those changes and after we reset every tracked property. I will add some comments to make everything more clear

packages/nimble-components/src/wafer-map/index.ts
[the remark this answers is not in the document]
munteannatan: the approach you imagined is the correct one, the only difference is that we batch changes together before the wafer map updates and the wafer map tracker is the one responsible for queueing the updates. the flow is as follows:

1. the wafer detects a change and records it in the tracker. when the first change is detected, an update is queued in the tracker. meanwhile more changes are batched until the update runs.
2. the update is finally run. the wafer runs this update but it checks the tracker to see what kind of update to run based on the recorded changes.
3. after the update finishes, all the tracked items are reset and ready for the next changes.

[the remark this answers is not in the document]
munteannatan: sorry for the late reply, the use case here is a bit different because we have a somewhat linear update flow, we do not have independent updates. so the higher order updates, like `DiesRenderInfoUpdate` in your example will trigger a full length update flow which contains the `ScalesUpdate` flow. These are not separate updates, just different starting points in the same update flow. we do not have the option to run only certain segments of the update because it snowballs. we can only chose the altitude

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -153,34 +155,88 @@ export class WaferMap extends FoundationElement {
...
+                this.updateLabelsFontSize();
+            } else if (
+                this.waferMapUpdateTracker.requiresDiesRenderInfoUpdate
+            ) {
+                this.updateDiesRenderInfo();
+            } else if (
+                this.waferMapUpdateTracker.requiresRenderingModuleUpdate
+            ) {
+                this.updateRenderingModule();
+            }
```
rajsite, line 197: so every single one of these code paths ends up calling `renderer.drawWafer`. I don't think the indirection is making it clearer what work this method is doing. Did you have something in mind for needing all the indirection? Would something like the following be clearer?


```suggestion
            if (this.waferMapUpdateTracker.requiresContainerDimensionsUpdate) {
                this.dataManager.updateContainerDimensions();
            } else if (this.waferMapUpdateTracker.requiresScalesUpdate) {
                this.dataManager.updateScales();
            } else if (
                this.waferMapUpdateTracker.requiresLabelsFontSizeUpdate
            ) {
                this.dataManager.updateLabelsFontSize();
            } else if (
                this.waferMapUpdateTracker.requiresDiesRenderInfoUpdate
            ) {
                this.dataManager.updateDiesRenderInfo();
            }
            this.renderer.drawWafer();
```

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -202,56 +258,64 @@ export class WaferMap extends FoundationElement {
     }
 
     private quadrantChanged(): void {
-        this.queueRender();
-    }
-
-    private orientationChanged(): void {
-        this.queueRender();
+        this.waferMapUpdateTracker?.track('quadrant');
```
rajsite, line 261: Can you remove all the conditional checks now? i.e. `this.waferMapUpdateTracker.track('quadrant');`

packages/nimble-components/src/wafer-map/modules/event-coordinator.ts
```diff
@@ -11,13 +11,12 @@ export interface EventCoordinatorCallbacks {
...
 export class EventCoordinator {
-    private readonly zoomHandler: ZoomHandler;
-    private readonly hoverHandler: HoverHandler;
-
-    public constructor(private readonly wafermap: WaferMap) {
-        this.zoomHandler = new ZoomHandler(wafermap);
-        this.hoverHandler = new HoverHandler(wafermap);
+    private readonly zoomHandler = new ZoomHandler(this.wafermap);
+    private readonly hoverHandler = new HoverHandler(this.wafermap);
+    public constructor(private readonly wafermap: WaferMap) {}
```
rajsite, line 16: So this is a funny one. Normal JS supports class members nowadays but the pattern you're doing here (where it is expected the constructor runs before the properties) does not actually happen in JS. See: https://jsbin.com/dekatuc/1/edit?js,console and notice that undefined is printed because the ZoomHandler is constructed before the constructor runs, not after.

The reason this works is because TypeScript actually generates code closer to [what you had originally](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDyAuALApgJwCqIJYDsDm0A3gFDRnTEC+xokMmuBJ50wA9jhPKgK7DxtUACjZI0DfAC44YjNnwBKItWrIAHgAdB8VuCjQAogDdkOeAGE2ggCa4wA1EVLkNqLEfvJoqZGGscQAE9oAC8rAFsACTAcaxA0aABeaBxkAHdoCTwhJCwIADo0sAAzNHCwDQUAbmcyV3dPb19-HCDoRDYTVGjY+Mdk1IysnOwCotLUcsqalg0eACMQLGBWDi5efkEheo94Lx8-AODxsorpBBQ5RiVCFSA)

So in just this case I would recommend keeping the constructor for initializing these two class members (zoom and hover handler) so the source better aligns with what is generated.

packages/nimble-components/src/wafer-map/modules/data-manager.ts
```diff
@@ -54,27 +54,43 @@ export class DataManager {
...
+    }
 
-        this.dataMap = new Map(
-            wafermap.dies.map(die => [`${die.x}_${die.y}`, die])
-        );
+    private readonly computations = new Computations(this.wafermap);
+    private readonly prerendering = new Prerendering(this.wafermap, this);
+    private dataMap!: Map<string, WaferMapDie>;
+
+    public constructor(private readonly wafermap: WaferMap) {}
```
rajsite, line 65: Similar to the discussion in the event-coordinator, because creating Computations and Prerendering rely on the constructor running first we should just make those members in the constructor.

packages/nimble-components/src/wafer-map/index.ts
munteannatan: I left these checks here because there are cases when the waferMapUpdateTracker is still undefined and a property changes. for example, the first attribute gets its default value set which triggers a change event before the waferMapUpdateTracker is initialized. Maybe there is something I do not understand about the initialization and connected callback of the component, but without these the unit tests are failing. Can you also check this behavior? Maybe we can have a call to find  a solution to this when you have the time

packages/nimble-components/src/wafer-map/index.ts
munteannatan: It is something that remained from before the previous change, when the drawWafer was not always the last call. Thaks for noticing!

packages/nimble-components/src/wafer-map/index.ts
rajsite: I think I see what is happening and yea it's a bit tricky:

So this code:
```ts
class Ex{
   @observable
   public myProp = 'a';

  private updateTracker = new UpdateTracker(this);
  
  myPropChanged() {
    updaterTracker.doSomething();
  }
}
```

When compiled desugars to something like:
```ts
class Ex{
   get myProp {
       return this._myProp;
   }
   set myProp(v) {
      this._myProp = v;
      this.myPropChanged();
  }  
  myPropChanged() {
    updaterTracker.doSomething();
  }
  constructor() {
      this.myProp = 'a';
      this.updateTracker = new UpdateTracker(this);
  }
}
```

So the issue here is literally the order of instantiation. The property assignment runs first causing the changed callback to run before the update tracker was constructed.

Instead, you need the updateTracker to be created before any of those observable property assignments run.

```ts
class Ex{
   /* @internal */
   public updateTracker = new UpdateTracker(this);

   @observable
   public myProp = 'a';

  myPropChanged() {
    updaterTracker.doSomething();
  }
}
```

Making sure the objects needed in the changed handlers are created first is also the approach used in TableColumns (which because of our lint rules encourages making it public and marking it internal) See:
https://github.com/ni/nimble/blob/1a5679389fe3439214ff0eb8d0029e7e6e755a61/packages/nimble-components/src/table-column/base/index.ts#L25

packages/nimble-components/src/wafer-map/index.ts
rajsite: Unresolving because the second part of my concern based on the proposed snippet was not addresses.
The question is are all these one line wrapper indirections still needed, notice in my snippet above that they were inlined instead:

https://github.com/ni/nimble/blob/e2a99c375d20a544bd85567f9dae473dafcc8b98/packages/nimble-components/src/wafer-map/index.ts#L201-L219

packages/nimble-components/src/wafer-map/index.ts
rajsite: Hopefully that clears it up and we can initialize the update tracker first but feel free to reach out if you need to discuss / have Qs

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -134,53 +140,69 @@ export class WaferMap extends FoundationElement {
...
+        hoverWidth: number,
+        hoverHeight: number,
+        hoverOpacity: HoverDieOpacity,
+        hoverTransform: string
+    ): void {
+        this.hoverWidth = hoverWidth;
+        this.hoverHeight = hoverHeight;
+        this.hoverOpacity = hoverOpacity;
+        this.hoverTransform = hoverTransform;
     }
```
rajsite, line 175: It's not clear to me what the benefit of this method is. Other spots that mutate the wafermap state manipulate the properties directly like zoom-handler and hover-handler.

Since this method is only used from one spot and does not follow other patterns for manipulating wafermap state I would recommend we don't add it and the rendering module just sets the properties like before

packages/nimble-components/src/wafer-map/modules/event-coordinator.ts
```diff
@@ -11,13 +11,15 @@ export interface EventCoordinatorCallbacks {
...
-    private readonly hoverHandler: HoverHandler;
-
+    private readonly zoomHandler;
+    private readonly hoverHandler;
     public constructor(private readonly wafermap: WaferMap) {
         this.zoomHandler = new ZoomHandler(wafermap);
         this.hoverHandler = new HoverHandler(wafermap);
+    }
 
+    public updateEvents(): void {
```
rajsite, line 21: Can you instead merge attachEvents and updateEvents into a single public function called `attachEvents`?

```ts
    public attachEvents(): void {
        this.zoomHandler.updateZoomBehavior();
        this.wafermap.addEventListener('mousemove', this.onMouseMove);
        this.wafermap.addEventListener('mouseout', this.onMouseOut);
        this.wafermap.canvas.addEventListener('wheel', this.onWheelMove, {
            passive: false
        });
    }
```

That way the public api is `attachEvents` and `detachEvents`.

I recognize other changes are more of an update pattern but this case is more of a attach / create

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
```diff
@@ -20,9 +20,11 @@ export class ZoomHandler {
...
-    private readonly zoomBehavior: ZoomBehavior<Element, unknown>;
+    private zoomBehavior!: ZoomBehavior<Element, unknown>;
 
-    public constructor(private readonly wafermap: WaferMap) {
+    public constructor(private readonly wafermap: WaferMap) {}
+
+    public updateZoomBehavior(): void {
         this.zoomBehavior = this.createZoomBehavior();
         this.zoomBehavior(select(this.wafermap.canvas as Element));
     }
```
rajsite, line 30: Similar to the event cordinator can you just make a public createZoomBehavior after merging together updateZoomBehavior and createZoomBehavior.

I recognize other changes are more of an update pattern but this case is more of a create than an update

rajsite (commented):
Looking good, just a few naming things left!

change/@ni-nimble-components-b6bd9b08-a8e3-4e8f-bab1-7da899d6fece.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @munteannatan as an FYI from [Nimble's contribution policy changes](https://teams.microsoft.com/l/message/19:wo8vmMKMsHfltKXxc0bczZo-X4JlQSV5VxpaRJdh13k1@thread.tacv2/1690832858628?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=9ee065d7-3898-4245-82f6-76e86084b8b1&parentMessageId=1690832858628&teamName=Design%20System&channelName=General&createdTime=1690832858628) make sure to get local review approval before the final owners reviewer reset.
For this PR you don't need a Nimble Buddy PR from @atmgrifter00  as we are already close to finishing but future PRs should have the local approval and nimble buddy before leaving Draft.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -134,53 +140,69 @@ export class WaferMap extends FoundationElement {
...
 
-    private eventCoordinator?: EventCoordinator;
-    private resizeObserver?: ResizeObserver;
+    private readonly eventCoordinator = new EventCoordinator(this);
+    private readonly resizeObserver = this.createResizeObserver();
 
     public override connectedCallback(): void {
         super.connectedCallback();
+        this.canvas.width = this.clientWidth;
+        this.canvas.height = this.clientHeight;
```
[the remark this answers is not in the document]
munteannatan: they are causing unintended behaviour at component initialization. but cnvas.height does not need to be initiated either so i removed everything

rajsite (approved):

PR #1300, opened 2023-06-14
Default buttons appearance updates

aidendk:
# Pull Request

## 🤨 Rationale

Updates the styling of the `default` Nimble Button, Anchor Button, and Toggle Button to reflect the designs in the [Figma Spec](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-77205&t=u8M17crgkG5nt32y-0).

## 👩‍💻 Implementation

Provides a styling update that is up to date and fixes design tokens to reflect new theme colors (specifically the `PowerGreen` color for the `dark` theme). The font for the button has also been changed, and now has a font of `semi-bold`. 

Other components that previously used the same font token as the button (including the switch. radio, and checkbox) have been changed to use the `bodyFont` token to give them a font of `regular`. The font for the tab component was also changed to `semi-bold` to reflect new UX designs in Figma.

The css variables border, box-shadow, and outline are all used in order to best reflect the new button designs in Figma and keep the button consistent with other component styling.

Changing only the `border` css property through `calc()` was also considered, but using this required extra properties like `padding` and calculations that `box-shadow`  and `outline` don't require.

## 🧪 Testing

Chromatic Testing

## ✅ Checklist


packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -49,7 +50,7 @@ export const styles = css`
         background-color: transparent;
         height: 100%;
         width: 100%;
-        border: ${borderWidth} solid transparent;
+        border: 0 solid transparent;
```
m-akinc, line 53: I think this breaks most states of the outline primary button. Or something else does. Even though you're shortly going to redo the primary states, it would be nice to keep them correct in the meantime. See if just setting `border-width: ${borderWidth}` for those states fixes them.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -148,28 +151,27 @@ export const styles = css`
         ButtonAppearance.outline,
         css`
             .control {
+                box-shadow: 0px 0px 0px ${borderWidth}
+                    rgba(${actionRgbPartialColor}, 0.3) inset;
                 background-color: transparent;
```
m-akinc, line 156: I wonder why we're setting `background-color: transparent` in so many places. It's already set that way in the `.control` block at the top. I'd see if you can clean some/all of these up.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -207,17 +204,13 @@ export const styles = css`
         css`
             .control {
                 background-color: rgba(${borderRgbPartialColor}, 0.1);
-                border-color: transparent;
             }
 
             .control:hover {
```
m-akinc, line 209: Empty block can be removed

packages/nimble-components/src/patterns/button/styles.ts
aidendk: If I understand correctly, you meant that the border in the active state was not showing up, correct? I fixed this issue, but the only way I was able to do so was by adding the box-shadow property to the active state of the outline primary button.

m-akinc: Sorry, should have been more specific. I see three states of the primary outline button that are wrong now:
- default state has thin border
image
- hover state has thin border
image
- keyboard focused state is missing inner ring, and outer border is thin
image

If it's not straightforward to restore these, I think you can probably leave them as-is, since you will immediately be restyling the primary button anyway.

change/@ni-nimble-components-37c84290-c5e8-4c31-a22e-e69547cc96c0.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
aidendk: Sure, I'll show these changes to Brandon at the team meeting so the rest of the team can give input as well. The inner border you mentioned is intentional, Brandon has this in his design for the buttons in the Figma Spec.

The diffs that aren't related to the buttons are showing up because I changed the borderHoverColor attribute color for the dark theme to `PowerGreen`, as when I looked through Brandon's other Figma designs (Inputs, Checkboxes, etc.) this color seems to be the new default border color for the dark theme. I'll also double check with him about this in the team meeting.

Regarding the borders changing- this was also intentional. It had to do with issues getting the border-width css attribute to work and ultimately deciding to use only box-shadow for button borders. I'll go more into depth on this in the team meeting as well.

jattasNI, line 1: I may have misheard, but I thought Brandon had mentioned that he wanted tabs to receive the new semi-bold font. In Chromatic I'm seeing that tab toolbar buttons got that change but the titles of tabs are still regular font. Can you double check what's supposed to happen with tabs?

change/@ni-nimble-components-37c84290-c5e8-4c31-a22e-e69547cc96c0.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "default buttons appearance updates (#663)",
```
jattasNI, line 3: Since this is going to impact every client that uses a button, it'd be good to be more specific. (Feel free to update if I got something wrong)
```suggestion
  "comment": "Default appearance for buttons better matches visual design spec. Text is now semi-bold, border is more prominent when focused, and border hover color is changed for dark theme",
```

change/@ni-nimble-tokens-3974f217-50f5-4e6f-a318-be960d230b62.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Default Buttons Updated Styling",
```
jattasNI, line 3: It's often good to have different comments for each package so that clients who use just that package can see more specifically what changed.
```suggestion
  "comment": "Changed button label font weight token from Regular to Semi-Bold",
```

change/@ni-nimble-components-37c84290-c5e8-4c31-a22e-e69547cc96c0.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@ni/nimble-components",
+  "email": "103057880+aidendk@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
jattasNI, line 7: I'm a bit wary of the Chromatic status check. Do we know why the UI Tests are showing as unchanged? Is that because you already approved the diffs in a previous build, so the latest build is unchanged relative to that? Or is Chromatic somehow confused (I did change how Chromatic runs in #1301 so I want to make sure I didn't break something).

If you have an explanation, that's great. If not then it wouldn't hurt to do a sanity check in another clean branch where you 
1. make a change that triggers a diff in a build
2. accept that diff
3. make another change
4. see whether the diff is reported on that build anymore

image

jattasNI (approved):

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -49,7 +50,7 @@ export const styles = css`
         background-color: transparent;
         height: 100%;
         width: 100%;
-        border: ${borderWidth} solid transparent;
+        border: 0px solid transparent;
```
rajsite, line 53: Just gonna capture some thoughts here:

## Control sizing

This is a change in how we do the sizing of components. For example see the following two screenshots in devtools.

main:
image

PR:
image

So for example in those screenshots the definition of how the 16px green padding is applied changes. It becomes 16px to the outer edge of the control instead of 16px to the start of the border.

- Seems like this should be done consistently across all controls otherwise reasoning about sizing becomes difficult / we might end up with increasing inconsistencies
- The "border" draws inset to the control instead of outset. That seems reasonable I think.
  - Is that what figma does for measuring sizes as well? I found [one article](https://threads.com/blog/34432545238/why-we-use-the-css-property-boxshadow-instead-of-border-in-our-app) that seems to indicate borders are outset and don't contribute to size instead of inset. It may be useful to know how figma vs xd handle that. Maybe that could lean us to preferring an outset?

## Performance

This is adding additional box-shadow drawings to the button and based on the above potentially to every control. At least historically drawing box-shadows was pretty expensive. I'm having difficulty finding good references for performance today but one thing that does seem consistent is that animating box shadows is expensive.

- Can we characterize the performance cost of using box-shadows for all borders? Mobile / low-end devices performance ?
- Can we use an approach [to animate opacity instead](https://css-tricks.com/getting-deep-into-shadows/#aa-animating-shadows) of two overlaid box shadows instead of animating the the box-shadow itself?

aidendk: I talked with Molly about the benefits and drawbacks of using the additional box-shadows yesterday, and I think I'm going to revert it back to the normal border styling that we use in main. Based on what you mentioned regarding styling differences and possible performance costs, this issue with box-shadow where the text overflows beyond the background-color (see screenshot), and my discussion with Molly, I don't think it's worth it implementing box-shadow to fix the small issue of a tiny white between the two border pixels that's only visible when zooming in 4-6 times in storybook.

boxshadowbutton

I'm working on getting this code updated to reflect this now, and I'll show the diffs in the team meeting today.

change/@ni-nimble-components-37c84290-c5e8-4c31-a22e-e69547cc96c0.json
aidendk: Yup, the Figma spec shows the tabs as semi-bold font. it'll show as updated in my next commit.

aidendk:
The commit I just made (Button inner border fix) is the final version of the button based on Brandon's figma spec. I haven't finished the toggle button yet, but let me know what you all think of this one for now.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -63,9 +63,22 @@ export const styles = css`
         outline: none;
         margin: 0;
         padding: 0 ${standardPadding};
+        position: relative;
         transition: box-shadow ${smallDelay};
```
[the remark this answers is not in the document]
aidendk: I added a `smallDelay` transition to the outer border, a `mediumDelay` to the box-shadow, and a 0.5s transition to the transition for the outline used in the focus visible state. These changes should work for every button including the Anchor and Toggle buttons. 

The outline in the focus visible state is somewhat flickery, but I initially ran into issues adding a transition because it would create a black outline until the transition animation finished, and the only way to fix this was to make the outline transparent with the `color` css variable, which when used in the same class as the text would cause the text to be transparent as well. I found a fix for this, and as I show in the video it's not very noticeable when you're zoomed out.

https://github.com/ni/nimble/assets/103057880/32a6fe51-211f-4a51-a8b2-978ed02ab685

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +93,44 @@ export const styles = css`
...
+        box-shadow: 0px 0px 0px 1px ${borderHoverColor} inset;
+        transition: box-shadow ${mediumDelay};
     }
 
     .control${focusVisible} {
+        border-color: ${borderHoverColor};
         box-shadow: 0px 0px 0px ${borderWidth} ${borderHoverColor} inset;
+    }
+
+    .control${focusVisible}::before {
```
rajsite, line 106: `::before` pseudoelements are treated as different elements in the DOM tree and should be organized in the CSS that way per our styling guidelines.
See: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +81,56 @@ export const styles = css`
...
+    }
+
+    .control::before:active {
+        outline: none;
     }
 
     :host([disabled]) .control {
         box-shadow: none;
         outline: none;
     }
```
rajsite, line 127: The selection target of this element is `.control` and should be grouped with the other `.control` targetted selectors.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -148,57 +181,57 @@ export const styles = css`
...
+            .control::before:active {
                 background-color: ${fillSelectedColor};
-                border-color: ${fillSelectedColor};
+                outline: none;
+                padding: 1px;
             }
 
             :host([disabled]) .control {
+                border-color: rgba(${borderRgbPartialColor}, 0.3);
+            }
```
rajsite, line 207: The selection target of this element is `.control` and should be grouped with the other `.control` targetted selectors.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -207,32 +240,40 @@ export const styles = css`
...
+            .control::before:active {
                 background-color: ${fillSelectedColor};
-                border-color: ${fillSelectedColor};
+                padding: 1px;
             }
 
             :host([disabled]) .control {
                 background-color: rgba(${borderRgbPartialColor}, 0.1);
                 border-color: transparent;
             }
```
rajsite, line 271: The selection target of this element is `.control` and should be grouped with the other `.control` targetted selectors.

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -10,23 +16,66 @@ export const styles = css`
...
+    :host([disabled]) .control[aria-pressed='true'] {
         border-color: ${fillSelectedColor};
+        background-color: ${fillSelectedColor};
     }
 
     :host([disabled]) .control[aria-pressed='true']:hover {
-        background-color: ${fillSelectedColor};
         border-color: ${fillSelectedColor};
+        background-color: ${fillSelectedColor};
     }
```
rajsite, line 80: The selection target of this element is `.control` and should be grouped with the other `.control` targetted selectors.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -49,7 +50,7 @@ export const styles = css`
         background-color: transparent;
         height: 100%;
         width: 100%;
-        border: ${borderWidth} solid transparent;
+        border: 1px solid transparent;
```
rajsite, line 53: Why is the borderWidth design token no-longer being used?

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +81,56 @@ export const styles = css`
...
+    .control::before {
+        content: '';
+        position: absolute;
+        width: 100%;
+        height: 100%;
+        pointer-events: none;
+        box-sizing: border-box;
+        outline: none;
+        background-clip: content-box;
+        z-index: -1;
```
rajsite, line 110: what's the z-index: -1 about?

packages/nimble-components/src/patterns/button/styles.ts
aidendk: Got it. Also, it seems like declaring css classes with `::before` in the most recent commit's format (ex. `.control::before${focusVisible}`) causes the css classes to not work. Looking at other places that use `::before` like the  [nimble tab styles.ts](https://github.com/ni/nimble/blob/631614c7c6fb333e4018058611c9986d61b0002c/packages/nimble-components/src/patterns/tab/styles.ts), it seems that they use the format `.control${focusVisible}::before`, so I'll switch back to that format since when I do it fixes the problem of the css class not working.

packages/nimble-components/src/patterns/button/styles.ts
aidendk: Sorry, I forgot to change that back to the design token, it's fixed now along with a few other places where I was using 1px instead of the token 👍

packages/nimble-components/src/patterns/button/styles.ts
rajsite: Ah yea makes sense. A real element has focus states, the psedoelements `::before` and `::after` don't so yea you want to target the `::before` pseudelement of the `.control` element when the `.control` element has focus. `.control${focusVisible}::before`

packages/nimble-components/src/patterns/button/styles.ts
aidendk: Since the `::before` class has backgrounds in some conditions (like `:active`), if the z-index didn't exist, the background would overlap on top of the button text and cause the text color to look off.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +81,56 @@ export const styles = css`
...
+    }
+
+    .control${focusVisible}::before {
+        outline: ${borderWidth} solid ${borderHoverColor};
```
rajsite, line 119: The focus animations look very inconsistent to me.

For comparison, the hover animation:

hoveranimation

the outer border appears to grow inward to two pixels over time.

there is no inner border

For the focus animation:

focusanimation

The outer border does not seem to grow inward, it seems to pop-in instantly

The inner border seems to grow outward in size and it looks decoupled from the outer border animation. 

I can't find the reference animation so I can definitely be wrong but it seems overly complex. I would imagine it should feel like one single animation is happening where the outer border grows inward and the inner border either fades in or grows inward during the same time.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +81,56 @@ export const styles = css`
...
+        outline-offset: -3px;
+        transition: outline 0.5s;
+        color: transparent;
```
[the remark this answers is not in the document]
aidendk: I had issues with the transition creating a black outline before the transition finished and the outline became green, which I explained in [this discussion](https://github.com/ni/nimble/pull/1300#discussion_r1238767173) with Jesse.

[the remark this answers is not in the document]
aidendk: From [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/color): 

> The color CSS property sets the foreground [color value](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) of an element's text and [text decorations](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration), and sets the [currentcolor](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) value. currentcolor may be used as an indirect value on other properties and is the default for other color properties, such as [border-color](https://developer.mozilla.org/en-US/docs/Web/CSS/border-color).

The black outline appeared because in `.control::before`, the `color` variable wasn't set, so the outline would default to the `currentcolor` property which is the black colored outline we saw. By changing the color in `.control::before`, the currentcolor variable is no longer used and the default outline color is changed.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -60,10 +61,12 @@ export const styles = css`
         gap: 4px;
         cursor: inherit;
         font: inherit;
-        outline: none;
+        outline: transparent;
         margin: 0;
         padding: 0 ${standardPadding};
+        position: relative;
         transition: box-shadow ${smallDelay};
```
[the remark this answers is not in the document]
aidendk: I looked a few other components including the radio button and card button which have a transition on hover, and both use `smallDelay`. I'll update the button to always use `smallDelay`, except for in `focusVisible` which will need a `largeDelay`. I'll add a comment in the code with reasoning for this.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -60,10 +61,12 @@ export const styles = css`
...
+        transition: border-color ${smallDelay};
```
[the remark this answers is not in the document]
aidendk: Changed all transitions to be smallDelay, see review above.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,26 +81,56 @@ export const styles = css`
...
+        box-sizing: border-box;
+        outline: none;
+        background-clip: content-box;
+        z-index: -1;
+    }
+
+    .control${focusVisible}::before {
+        outline: ${borderWidth} solid ${borderHoverColor};
+        outline-offset: -3px;
+        transition: outline 0.5s;
```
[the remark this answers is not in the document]
aidendk: This delay will be on its own and will now use `largeDelay`. If I use smallDelay or mediumDelay for this, the inner, 1px `focus-visible` specific border shows up before outer 2px border (specifically the 1px box-shadow) is done transitioning. I added a comment explaining this in the code.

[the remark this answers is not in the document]
aidendk: I talked with Brandon about this, and he said that the smallDelay would be best for the transition for the inner 1px outline.

packages/nimble-components/src/patterns/button/styles.ts
aidendk: This should be fixed in my newest commit, I slowed this video (below) down and it looks like the inner border is growing inward now.

https://github.com/ni/nimble/assets/103057880/57011312-428c-4bac-97ba-08b31981ed49

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -10,23 +15,64 @@ export const styles = css`
     .control[aria-pressed='true'] {
         background-color: ${fillSelectedColor};
         border-color: ${fillSelectedColor};
+        position: relative;
     }
 
     .control[aria-pressed='true']:hover {
```
jattasNI, line 21: When I hover over a toggle button in the pressed state, the background is changing to appear unpressed. It looks like this is a regression, likely caused by removing background-color here.

Same issue with focusVisible below.


https://github.com/ni/nimble/assets/10500124/9ef8df2e-b682-4beb-9402-01e5ae071472

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -98,6 +99,33 @@ export const styles = css`
...
+        height: 100%;
+        pointer-events: none;
+        box-sizing: border-box;
+        outline: 0px solid transparent;
+        color: transparent;
+        background-clip: content-box;
+        z-index: -1;
+    }
+
+    .control${focusVisible}::before {
```
jattasNI, line 115: In the latest storybook build I'm not seeing the double border for focus visible any more. Was that an intentional change?

packages/nimble-components/src/toggle-button/styles.ts
aidendk: I looked into it, it's because control::before sets the z-index to -1, so the background that should show up on :active::before is behind the background of the entire matrix story. I'm looking into ways that I can fix this without having to set the z-index to 0, since doing so affects the text color.

packages/nimble-components/src/patterns/button/styles.ts
aidendk: Nope, this issue is also caused by the z-index being -1. I'm looking into fixes now.

aidendk: My newest commit fixes this issue but requires modifying the storybook.ts file which is used for all matrix stories. I double checked the matrix stories that use z-index as well as the chromatic tests for all matrix stories, and everything seems to be fine. This is the simplest way I could find to fix this issue, but we might need to discuss it with the team to make sure that everyone's okay with it.

jattasNI: Yeah, I'm nervous about that approach. Anything that requires global storybook changes to render correctly might also require client apps to make similar changes, and we don't want them to have to do anything special to use our components.

I don't think I'll have a chance to brainstorm better options this week (I'm only working a few hours) but you could try with Mert or put this aside and work on your other projects until people are back next week.

aidendk: My newest commit limits the change I previously made to the storybook.ts file to just the button, toggle button, and anchor button matrix story file. 👍

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -16,3 +16,32 @@ import * as buttonStories from './button.stories';
...
 <Controls of={buttonStories.outlineButton} />
 <Stories of={buttonStories} />
+
+# Usage Docs
+
+<br />
+
+## Sizing
+
+    The label button has two sizes: 32px and 24px, while the icon only button has two smaller sizes: 18px and 16px.
```
m-akinc, line 26: I don't think we use the term "label button" anywhere else. Doesn't immediately make me think of the non-icon button. I would just say "standard button".

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -16,3 +16,32 @@ import * as buttonStories from './button.stories';
...
+
+## Styling / Theme
+
+    Use the Color UI version for backgrounds with color (e.g. purple, blue).
+
+    See the usage details for more information on button styling / usage.
+
+## Accessibility
+
+    Please work with your designer and insure you have a 4.5:1
```
m-akinc, line 46: insure => ensure

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -12,7 +37,29 @@ export const appearanceVariantDescription = `This attribute has no effect on but
...
+    <ul>
+        <li>Icon and text buttons: 
+            <br/>Use a label plus icon to help reinforce a button's functionality, and increase the visual impact of the action.  
+            It should promote additional clarity, and should not be used solely for decoration. 
+        </li>
+        <br/>
+        <li>Icon only buttons: 
+            <br/>Use primarily for the purpose of limited space, and when the button's action is common and expected.
+            <br/>Replicate the usage to create patterns, but do not assume the user will understand the icon's meaning.
+            <br/>When a label isn't present, use a tool-tip to give extra information.
```
m-akinc, line 54: tool-tip => tooltip
(for consistency)

packages/nimble-components/src/patterns/button/styles.ts
m-akinc: I'm seeing that for the buttons in the Docs page, (1) they don't get a double outline when focused, and (2) they don't get a green fill when pressed/active. I'm guessing that's also related to the z-index issue.

m-akinc (changes requested):

aidendk:
My most recent commit fixes the z-index problem with the ::before pseudo element. I used a linear gradient for the background but kept the ::before pseudo element for the innermost `focusVisible` 1px border, as this allows the 1px border next to it to be transparent.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -207,26 +220,38 @@ export const styles = css`
...
             .control:hover {
                 background-color: transparent;
-                border-color: ${borderHoverColor};
+                background-image: linear-gradient(
+                    rgba(${borderRgbPartialColor}, 0.1),
+                    rgba(${borderRgbPartialColor}, 0.1)
+                );
+                background-size: calc(100% - 4px) calc(100% - 4px);
+                background-repeat: no-repeat;
+                background-position: center;
```
rajsite, line 233: These only apply when a background-image is set, so we can just move them up to the `.control ()` selector and avoid duplication. Whenever an image is set these are always the same

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,24 +80,57 @@ export const styles = css`
...
     .control:active {
         box-shadow: none;
         outline: none;
+        background-image: linear-gradient(
+            ${fillSelectedColor},
+            ${fillSelectedColor}
+        );
+        background-size: calc(100% - 2px) calc(100% - 2px);
+        background-repeat: no-repeat;
+        background-position: center;
```
rajsite, line 101: These only apply when a background-image is set, so we can just move them up to the `.control {}` selector and avoid duplication (remove all the other instances of them). Whenever an image is set these are always the same

packages/nimble-components/src/patterns/button/styles.ts
rajsite, line 233: I find the animation to be really jarring. When just using the control at a normal size it feels like it's glitching. I think it's because the space disappears immediately around the block instead of shrinking and then the border grows:
hoveranimation

If the background-size animates in it feels a lot smoother and less jumpy:

hoveranimation2

Could be done with something like this:

image

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -1,3 +1,28 @@
...
+    <ul>
+        <li>For appearance \`ghost\`:
+            <br/>This is the default and standard option for buttons. They create an open / airy feel, and fit comfortably in tight spaces.
+            <br/>Be careful when using that the surrounding context does not confuse this button for emphasized body text, tabs or a standalone links. 
+            <br/>Use this in combination with a primary outline and primary block buttons to create a hierarchy of importance.
+        </li>
+        <br/>
+        <li>For appearance \`outline\`:
+            <br/>This is an alternative standard button when a ghost button is not suitable, and a more visually direct button is needed. 
+            <br/>This can also create a light and airy feel. Use in combination with ghost buttons, not block buttons to create hierarchy.
```
jattasNI, line 15:
```suggestion
            <br/>This can also create a light and airy feel. Use in combination with ghost buttons to create hierarchy. Do not use in combination with block buttons. 
```

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -1,3 +1,28 @@
...
+            <br/>Use this in combination with a primary outline and primary block buttons to create a hierarchy of importance.
+        </li>
+        <br/>
+        <li>For appearance \`outline\`:
+            <br/>This is an alternative standard button when a ghost button is not suitable, and a more visually direct button is needed. 
+            <br/>This can also create a light and airy feel. Use in combination with ghost buttons, not block buttons to create hierarchy.  
+        </li>                
+        <li>For appearance \`block\`:
+            <br/>This standard button should be used when the most visible solution is required. 
+            <br/>It can be an alternative to overly subtle button solutions when it is important to emphasis an action and the functionality of the control.
```
jattasNI, line 19:
```suggestion
            <br/>It can be an alternative to overly subtle button solutions when it is important to emphasize an action and the functionality of the control.
```

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -1,3 +1,28 @@
...
+            <br/>Use in combination with ghost buttons, not outline buttons to create hierarchy.
```
jattasNI, line 20:
```suggestion
            <br/>Use in combination with ghost buttons to create hierarchy. Do not use in combination with outline buttons.
```

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -78,24 +84,55 @@ export const styles = css`
...
+        box-sizing: border-box;
+        outline: 0px solid transparent;
+        color: transparent;
+        background-clip: content-box;
+    }
+
+    .control${focusVisible}::before {
+        outline: ${borderWidth} solid ${borderHoverColor};
+        outline-offset: -3px;
+        transition: outline ${smallDelay} ease-in-out;
```
rajsite, line 127: Can you move the transition definition up to the `.control::before {}` default selector? (assuming there is not a specific reason it needs to be here instead).

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -10,23 +15,79 @@ export const styles = css`
...
+        background-image: linear-gradient(
+            ${fillSelectedColor},
+            ${fillSelectedColor}
+        );
+        background-size: calc(100% - 4px) calc(100% - 4px);
+    }
+
+    .control[aria-pressed='true']:active {
+        box-shadow: none;
+        outline: none;
```
rajsite, line 50: think getting rid of the `outline` on `.control` applies here too.

rajsite (approved):

m-akinc (approved):

PR #1304, opened 2023-06-15
Add tooltip to ellipsized table header

### LINKED ISSUE #857: Tooltip support for truncated cells
## 😯 Problem to Solve
If a table cell (or table header cell) is too small to display the contents, it should be truncated. They user should be able to access the full content via a tooltip.

## 💁 Proposed Solution
 - [ ] #889

## 🤔 Open Questions
Are there any things to make sure we address for accessibility...alt text for truncated content?
Depends on: 
- [ ] #309

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #857
Specifically, finishes the remaining work of adding a tooltip to header when text is ellipsized.

## 👩‍💻 Implementation

Updated the column template and base class. Follows established pattern.

## 🧪 Testing

Added three new automated tests.

## ✅ Checklist


change/@ni-nimble-components-d849f476-f620-4021-bf50-d5f1258be733.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The tooltip has some aggressive padding in Firefox on macOS (I didn't test other browsers). This looks vaguely familiar...we may have solved this in a previous tooltip PR but I can't recall how.

For reference, I used [the fractional width column story](https://60e89457a987cf003efc0a5b-hjirkftres.chromatic.com/?path=/story/incubating-table-column-configuration--fractional-width-column) to test it (struggled to find a case where the columns would ellipsize before the table hit its min width).

image

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -187,6 +187,41 @@ describe('Table', () => {
             expect(headerContent?.textContent).toEqual('foo');
         });
 
+        it('sets title when header text is ellipsized', async () => {
```
jattasNI, line 190: Are there cases we need to think about when the header content includes icons? In that case we recommend that clients set their own title; do we need to ensure we don't accidentally remove it?

m-akinc: I don't think so. They would be setting the title on the icon--we are setting the title on a span around the icon (and whatever else they have as content in their column element). I believe that if there is a title set on both the icon and the span, hovering over the icon should give the icon's tooltip, since it is the inner element.

packages/nimble-components/src/table-column/base/template.ts
```diff
@@ -1,10 +1,22 @@
...
+// prettier-ignore
 export const template = html<TableColumn>`
     <template slot="${x => x.columnInternals.uniqueId}">
-        <span class="header-content">
-            <slot></slot>
+        <span
+            ${ref('headerSpan')}
+            class="header-content"
+            @mouseover="${x => {
+        x.isValidContentAndHasOverflow = !!x.slottedHeaderContent.map(element => element.textContent?.trim()).join(' ')
```
rajsite, line 11: Can you move the duplicated logic for formatting the text to a method

packages/nimble-components/src/table-column/base/template.ts
```diff
@@ -1,10 +1,22 @@
...
+                    && x.headerSpan.offsetWidth < x.headerSpan.scrollWidth;
+    }}"
+            @mouseout="${x => {
+        x.isValidContentAndHasOverflow = false;
+    }}"
+            title=${x => (x.isValidContentAndHasOverflow ? x.slottedHeaderContent.map(element => element.textContent?.trim()).join(' ') : null)}
+        >
+            <slot ${slotted('slottedHeaderContent')}></slot>
```
rajsite, line 19: The overhead of persisting and tracking reference to the slotted content isn't needed.
Can you make a ref to the slot and use it instead in the method above being added
```ts
contentSlot.assignedNodes().map(...)
```

rajsite (approved):

packages/nimble-components/src/table-column/base/template.ts
m-akinc: Done

packages/nimble-components/src/table-column/base/template.ts
m-akinc: Done

packages/nimble-components/src/table-column/base/index.ts
```diff
@@ -41,6 +45,17 @@ export abstract class TableColumn<
...
+    /** @internal */
+    @observable
+    public isValidContentAndHasOverflow = false;
+
+    /** @internal */
+    public headerSpan!: HTMLSpanElement;
+
+    /** @internal */
+    @observable
+    public contentSlot?: HTMLSlotElement;
```
rajsite, line 57: How are you deciding when to use the non-null assertion and when to say it's nullable?
Both of these are created by the same template, not sure why their nullability is treated different

rajsite: Make sure and clean-up side-effects of the change. i.e. can remove the optional chaining from contentSlot reference.

jattasNI (approved):

Issue #846, opened 2022-12-02
Interactively resize column width


nate-ni:
## 😯 Problem to Solve
The end user needs to able to adjust the width of a column. 

## 💁 Proposed Solution
Design will be provided by
- [x] #885 

## 🤔 Open Questions
Now that the table will be interactive, this is a good point to decide about "role" for the Nimble table. We should resolve this before closing this or other "interactive" work items.

tatwood2:
The Licensing NIC team would like to use this enhancement in our desktop Electron with Blazor application for LiMo when it is ready.  Please keep us in the loop as we'd like to uptake it as soon as it is ready.

PR #1305, opened 2023-06-15
Column sizing hld update

atmgrifter00:
# Pull Request

## 🤨 Rationale

These are changes to the column sizing HLD in response to what I believe we need to properly support interactive column sizing (#846).

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A
## ✅ Checklist


packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,15 +15,18 @@ We need to provide users the means for changing the widths of individual columns
...
+    -   A sizing action to the right that would ultimately result in the final right column reaching its minimum size (all columns still within current table width) would _conditionally_ (see behavior descriptions below) begin to push columns out of the table viewport width resulting in a horizontal scrollbar.
 -   Columns can be configured to not allow a user to interactively size them
     -   The implicit behavior present based on the behaviors described above, is that in a sizing action that cascades to a column configured to not be resized is that the column won't be resized towards a minimum size, and the cascade will effectively "skip" this column.
+-   A user will be able to configure the table such that dragging a column will never result in a change in the table's viewport width. This does _not_ mean that a table configured in such a will _not_ have a scrollbar (e.g. a table could still have enough columns in its view to demand a horizontal scrollbar). This API will simply prevent the user from expanding the existing viewport width.
```
jattasNI, line 27: Configuring this behavior will be done via an API, not a user-visible setting, correct? If so I'd suggest rewording to say "A client will be able to configure..." or "The table will expose an API to configure it such that..."

jattasNI, line 27: Could you clarify what it will do instead of expanding the viewport width? Stop expanding the column?

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -35,7 +38,23 @@ There are some column sizing behaviors that we will ultimately expect to support
...
+    /**
+     * Allows a user to disable interactive column sizing
+     */
+    @attr({ attribute: 'disable-column-sizing', mode: 'boolean' })
+    public disableColumnSizing = false;
+
+    /**
+     * Allows a user to limit interactive column sizing to the available viewport width.
+     */
+    @attr({ attribute: 'restict-sizing-to-viewport', mode: 'boolean'})
```
jattasNI, line 53: typo (though I'd still vote for other names)
```suggestion
    @attr({ attribute: 'restrict-sizing-to-viewport', mode: 'boolean'})
```

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -35,7 +38,23 @@ There are some column sizing behaviors that we will ultimately expect to support
 
 ### API
 
-To accomodate the various sizing modes of a column, in addition to the other necessary behaviors, we can add the following properties to `ColumnInternals`:
+`Table`
```
jattasNI, line 41: Naming suggestions:
1. stick with `disable` for both properties rather than `restrict`
2. use `resize` instead of `sizing` to match browser APIs like [resize](https://developer.mozilla.org/en-US/docs/Web/CSS/resize). So `disable-column-resize` and `disableColumnResize`
3. Consider alternate names for the table resizing attribute
    - `disable-resize-grow`
    - `disable-column-resize-autogrow`
4. Consider changing them both to enums to allow for future growth and make the API more expressive (also inspired by `resize='none' | 'horizontal' | 'vertical' | 'both'`)
   - `column-resize = 'none' | 'push-pull'`
   - `column-resize-table-width = 'none' | 'grow'
5. Consider combining the two attributes to a single enum.
   - `column-resize= 'none' | 'grow-table' | 'limit-width'`?

Just a starting point for brainstorming. I'm not attached to any of these if they don't cover the behavior accurately.

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -183,3 +202,10 @@ To help facilitate this, it may be helpful to approach the initial implementatio
 TanStack offers the ability to maintain column sizing state as well as APIs to manage interactive sizing ([`getResizeHandler`](https://tanstack.com/table/v8/docs/api/features/column-sizing#getresizehandler)). By plugging into `getResizeHandler` TanStack can provide all necessary column size state either as the column is being sized (allowing immediate sizing while dragging), or at the end of an interactive operation (columns update size on mouse up for example).
 
 TanStack expects size values to be provided as [pixel values](https://tanstack.com/table/v8/docs/api/features/column-sizing#size). As such, we wouldn't be able to leverage TanStack's APIs in a way to achieve our initial desired behavior. However, it's possible that it would be beneficial to attempt to upstream changes to TanStack that would allow us to leverage it, but I would suggest that we revisit this possibility at a later time.
+
+## Open Issues
+
+1. How do we allow a user to remove any added viewport width (i.e. a column has been grown such that a horizontal scrollbar is visible)? Options considered include:
```
jattasNI, line 208: FYI for other reviewers, we have a meeting with interaction designers on Wednesday to discuss this.

packages/nimble-components/src/table/specs/table-column-width-hld.md
atmgrifter00: At the moment I'm leaning towards the last suggestion. What about:
-    ```column-resize: 'none' | 'cascade' | 'cascade-limit-width';```

packages/nimble-components/src/table/specs/table-column-width-hld.md
atmgrifter00: Changing to the single `column-resize` attribute for now.

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
+-   If a user drags a divider between two columns to the right, then the column on the left will grow larger, and the column on the right will grow smaller by the same pixel amount. Sub-behaviors to this are: - If a shrinking column has reached its minimum pixel size or is not resizable, then the next column in the direction of the sizing action will be affected up to the final column in a given direction. - A sizing action to the left will ultimately stop having an effect when the left-most column reaches its minimum size. - A sizing action to the right that would ultimately result in the final right column reaching its minimum size (all columns still within current table width) would _conditionally_ (see behavior descriptions below) begin to push columns out of the table viewport width resulting in a horizontal scrollbar on the table. - Within a single mouse interaction (i.e. drag-sizing without releasing mouse), if a cascade results in a column not adjacent to the divider being sized, then moving the mouse back in the opposite direction will "revert" the size made to the non-adjacent column.
```
mollykreis, line 21: The formatting here looks like it got messed up. The sub bullet points should be on their own lines.

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
+    ![Column resizing](spec-images/tableColumnResize.gif)
+
 -   Columns can be configured to not allow a user to interactively size them
     -   The implicit behavior present based on the behaviors described above, is that in a sizing action that cascades to a column configured to not be resized is that the column won't be resized towards a minimum size, and the cascade will effectively "skip" this column.
+-   A client will be able to configure the table such that dragging a column will never result in a change in the table's viewport width. Instead, once all columns have been reduced to their minimum width to the right of the divider, further mouse moves to the right will result in no action. This does _not_ mean that a table configured in such a way will _never_ have a scrollbar (e.g. a table could still have enough columns in its view to demand a horizontal scrollbar due to their combined minimum sizes).
```
mollykreis, line 26: Is there a recommendation or idea of why a client would use one mode rather than another? I'm all for having the API be written in a way where we can add new modes in the future without breaking the existing API, but I'm not convinced we should add modes now that we don't have (even theoretical) use cases for.

atmgrifter00: Agreed. I've removed this.

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
+-   If a user drags a divider between two columns to the right, then the column on the left will grow larger, and the column on the right will grow smaller by the same pixel amount. Sub-behaviors to this are: - If a shrinking column has reached its minimum pixel size or is not resizable, then the next column in the direction of the sizing action will be affected up to the final column in a given direction.
+-   A sizing action to the left will ultimately stop having an effect when the left-most column reaches its minimum size. - A sizing action to the right that would ultimately result in the final right column reaching its minimum size (all columns still within current table width) would _conditionally_ (see behavior descriptions below) begin to push columns out of the table viewport width resulting in a horizontal scrollbar on the table. - Within a single mouse interaction (i.e. drag-sizing without releasing mouse), if a cascade results in a column not adjacent to the divider being sized, then moving the mouse back in the opposite direction will "revert" the size made to the non-adjacent column.
```
rajsite, line 22: nit: looks like these are list items that should be on separate lines

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
-    -   If a shrinking column has reached its minimum pixel size or is not resizable, then the next column in the direction of the sizing action will be affected up to the final column in a given direction.
-    -   A sizing action to the left will ultimately stop having an effect when the left-most column reaches its minimum size.
-    -   A sizing action to the right that would ultimately result in the final right column reaching its minimum size (all columns still within current table width) would then begin to push columns out of the table width resulting in a horizontal scrollbar.
+-   If a user drags a divider between two columns to the right, then the column on the left will grow larger, and the column on the right will grow smaller by the same pixel amount. Sub-behaviors to this are: - If a shrinking column has reached its minimum pixel size or is not resizable, then the next column in the direction of the sizing action will be affected up to the final column in a given direction.
```
rajsite, line 21: nit: same as below, would be easier to read as a sublist instead of a large paragraph

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
     -   The implicit behavior present based on the behaviors described above, is that in a sizing action that cascades to a column configured to not be resized is that the column won't be resized towards a minimum size, and the cascade will effectively "skip" this column.
 
+#### Table Sizing
+
+-   Table resizing can result in _asymmetric_ column size state for mirrored actions such as a window maximize operation followed by a window restore down action. The following is a summary of expected behavior:
+    -   If the table currently has a horizontal scrollbar and the table is then sized to be wider, then as the table grows it will first remove excess space for the viewport prior to growing any columns. At this point, after the initial wider size has been completed, further reductions in window size will _not_ have a restorative state to the scrollbar prior to growing the window.
+    -   If the table currently has a horizontal scrollbar and the table is then sized to be smaller, the current proportional width of each column will be maintained.
+
+[Prototype branch](https://60e89457a987cf003efc0a5b-xtgmcptvna.chromatic.com/?path=/story/table-column-configuration--grouping).
```
rajsite, line 34: nit/fyi: stories from builds are deleted after 30 days, so if possible include a gif, but not a deal-breaker (wouldn't expect prototype branches to stick around forever anyway)

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -15,27 +15,45 @@ We need to provide users the means for changing the widths of individual columns
...
 
-To accomodate the various sizing modes of a column, in addition to the other necessary behaviors, we can add the following properties to `ColumnInternals`:
+`Table`
+
+```ts
+    /**
+     * Allows a user to set how interactive column sizing should behave
+     */
+    @attr({ attribute: 'column-resize' })
+    public columnResize: 'none' | 'cascade' = 'cascade';
```
rajsite, line 53: nit: / implementation detail. The default value should be `undefined` (i.e. is not mandatory to be set) so if cascade is the default the enum should be:

```ts
export const TableColumnResize = {
    default: undefined,
    none: 'none'
}

```

Or potentially less proffered, naming the default option `cascade`, but that could cause potential confusion as the string `'cascade'` should not be an option:
```ts
export const TableColumnResize = {
    cascade: undefined,
    none: 'none'
}
```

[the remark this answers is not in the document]
atmgrifter00: Correct, this only applies to column sizing via the dividers (which would only appear when this attribute is not `none`). A reason to support this, I think, are scenarios that involve tables with well-known content with respect to their sizes, and sizing a column differently would not do anything other than possibly create unneeded horizontal scrollable space.

I could be convinced that even this is a feature that we should wait for a request before adding.

packages/nimble-components/src/table/specs/table-column-width-hld.md
atmgrifter00: Yeah, not sure how that got goofed up.

packages/nimble-components/src/table/specs/table-column-width-hld.md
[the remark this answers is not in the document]
atmgrifter00: This API has been removed.

packages/nimble-components/src/table/specs/table-column-width-hld.md
atmgrifter00: I've now just changed the link to the working branch (not a prototype branch anyway). I've also included another gif to demonstrate proposed behavior.

mollykreis (approved):

rajsite (approved):

msmithNI (approved):

jattasNI (approved):

PR #1313, opened 2023-06-16
Fix intermittent hang in tokens build

### LINKED ISSUE #1297: nimble-tokens svg-to-ico task hangs on Windows after a fresh sync (until a retry)
## 🧹 Tech Debt
Almost any time a developer syncs down the Nimble repo on a Windows machine at a new commit point / on a new branch, and does `npm run build`, the build hangs on the nimble-tokens svg-to-ico task:

```
> @ni/nimble-tokens@6.0.0 build:svg-to-ico
> node build/svg-to-ico.js

Icon asset directory :c:\dev\nimble\packages\nimble-tokens\dist\icons\svg
Output ico directory :c:\dev\nimble\packages\nimble-tokens\dist\icons\ico
Number of icons found to convert: 158
.
```

It would be good to investigate & fix this issue if possible.

## Workaround
Cancel the build and start `npm run build` again

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1297

## 👩‍💻 Implementation

The hang was happening when building two icon buffers (i.e. 16x16 and 32x32) at the same time. The solution is to inline the svg-to-ico function (from the svg-to-ico npm package) and modify it to run serially.

I did not seriously consider upstreaming this fix, because I'm removing parallelism, and I don't understand the cause/breadth of this issue enough to push that on all users of the package. The repo did not have a single issue opened in it (ever). It has a total of 15 commits, and is rarely touched. So the likelihood of having a change merged also seems slim.

## 🧪 Testing

Was able to force the hang reliably by starting to delete my node_modules folder, then cancelling when it popped up a dialog about files being used by other processes. Doing this would always cause the next `npm run build` of `nimble-tokens` to hang. After my changes, it no longer hangs.

## ✅ Checklist


packages/nimble-tokens/build/svg-to-ico.js
m-akinc, line 1: I'm not 100% sure people are going to like this approach to solving the problem. I'm about to go on vacation, but I'm putting this PR up as a suggestion. If there are fundamental objections, feel free to close the PR. If you are okay with it, feel free to usher the PR through, or just wait for me to return from vacation.

jattasNI: @msmithNI could you do a bit of testing with this proposed fix locally since you'd been seeing it pretty reliably?

msmithNI: This fix worked for me locally.

It's a bit tricky to try it multiple times since generally I only get the problem after syncing new changes. I think if we update this branch with the latest changes with main, then I could pull it down again and that scenario would trigger the hang without this fix too (I think).

msmithNI: Still works after pulling down the merged changes from main, so yeah I think this fix works.

The svg-to-ico stage of the build is still super fast too, so removing the concurrency of that step seems OK to me if it fixes this issue.

jattasNI (approved):

rajsite (approved):

PR #1315, opened 2023-06-22
Fix status table Storybook links

jattasNI:
# Pull Request

## 🤨 Rationale

I noticed that the links from the [component status table](https://nimble.ni.dev/storybook/?path=/docs/component-status--docs) to component stories were broken on `https://nimble.ni.dev/storybook`. They worked fine in local builds and Chromatic PR builds.

They should look like this:
`https://nimble.ni.dev/storybook/?path=/docs/components-anchor--docs`

But instead they look like this:
`https://nimble.ni.dev/?path=/docs/components-anchor--docs`

## 👩‍💻 Implementation

Changed status table relative links from this form, which is relative to the root of the site:
`/?path=/docs/components-anchor--docs`

to this form, which is relative to the current document: 
`./?path=/docs/components-anchor--docs`

and added `target="_top"` to the anchor columns to escape the storybook iframe context.

Also updated the docs where we describe how to link to stories from anchor tags and scanned for other stories with relative links (there weren't any).

## 🧪 Testing

I tested links in a local storybook build, a local site build, and in the Chromatic PR build. But we won't know for sure whether this fixes the issue in production until we've completed the PR and I can test on nimble.ni.dev/storybook.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/docs/component-status.stories.ts
+++ b/packages/nimble-components/docs/component-status.stories.ts
@@ -23,7 +23,7 @@ const components = [
     },
     {
         componentName: 'Anchor',
-        componentHref: '../?path=/docs/components-anchor--docs',
+        componentHref: './?path=/docs/components-anchor--docs',
         designHref: 'https://xd.adobe.com/view/33ffad4a-eb2c-4241-b8c5-ebfff1faf6f6-66ac/screen/bfadf499-caf5-4ca0-9814-e777fbae0d46/',
         designLabel: 'XD',
         issueHref: 'https://github.com/ni/nimble/issues/324',
@@ -34,7 +34,7 @@ const components = [
     },
     {
         componentName: 'Anchor Button',
-        componentHref: '../?path=/docs/components-anchor-button--docs',
+        componentHref: './?path=/docs/components-anchor-button--docs',
         issueHref: 'https://github.com/ni/nimble/issues/324',
         issueLabel: 'Issue',
         componentStatus: '✅',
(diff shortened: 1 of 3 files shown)


packages/nimble-components/docs/component-status.stories.ts
```diff
@@ -23,7 +23,7 @@ const components = [
     },
     {
         componentName: 'Anchor',
-        componentHref: '/?path=/docs/components-anchor--docs',
+        componentHref: '../?path=/docs/components-anchor--docs',
```
rajsite, line 26: We just want to change the search query parameter on the url right? Can we just do that?:

```suggestion
        componentHref: '?path=/docs/components-anchor--docs',
```

Hopefully that works because what is being shown doesn't make sense:

image

Notice how `storybook` gets removed from the path.

Compared to:

image

You can also test this stuff out by doing a full build (or build of storybook and packages/site) and then `npm run start:eleventy -w packages/site`

rajsite: If we end up needing to hold onto this weird path thing can you add a comment on the comments var linking to the issue. In particular because the relative path doesn't make sense from what I can see so there must be something else happening...

rajsite: Ohhh I see, the component is inside an iframe so the page url actually looks like `https://nimble.ni.dev/storybook/iframe.html?viewMode=docs&id=component-status--docs`

So we want the url to be resolved from the current directory like:
```suggestion
        componentHref: './?path=/docs/components-anchor--docs',
```

Since the iframe.html is "in the same directory" as the root index page. BUT we don't want to navigate the iframe context we want to navigate the top level browser. So the anchor column should be configured `<nimble-table-column-anchor target="_top">`

jattasNI: Thanks for the guidance! I confirmed that your suggested approach works for both the local site build and a local storybook build. And that the previously suggested approach within markdown works for those too (I think Storybook has some special processing for that case). 

So I updated the status table to use the new approach and updated our docs with guidance for markdown links and anchor links.

rajsite (approved):

PR #1321, opened 2023-06-23
Interactive column sizing

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #846

## 👩‍💻 Implementation

Bulk of implementation is the creation of the new `TableLayoutManager` class that is responsible for updating the various aspects of sizing state the table needs.

Beyond that it was mainly just updating the `Table` template to host the new interactive dividers, and a "glass pane" `div` to handle the disabling of hover states in template elements during a column resize operation.

## 🧪 Testing

Added unit tests.

## ✅ Checklist


packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,358 @@
...
+                    - columnInitialWidths.minPixelWidth;
+                currentIndex -= 1;
+            }
+        }
+        return delta > 0
+            ? Math.min(delta, availableSpace)
+            : Math.max(delta, -availableSpace);
+    }
+
+    private performCascadeSizeLeft(
```
atmgrifter00, line 240: You will notice some obvious similarities between this method and `performCascadeSizeRight`. I made some attempts to refactor the methods to share some of the similar code, but didn't come up with something I liked. It's difficult because a lot of it is "reversed" logic. I still believe there is room to at least make some of the logic more consistent between the two, and am very interested in suggestions here.

packages/nimble-components/src/table/styles.ts
```diff
@@ -89,16 +86,72 @@ export const styles = css`
...
+        position: absolute;
+    }
+
+    .column-divider[active] {
+        display: block;
+        z-index: 1;
+    }
+
+    .header-container:hover .column-divider:not([not-active]).left,
+    .header-container:hover .column-divider:not([not-active]).right {
```
atmgrifter00, line 118: These CSS selectors were my admittedly goofy way of dealing with not marking dividers on a column other than the one you started the drag from as `active` (and thus preventing them from appearing). I've already discussed with @mollykreis that whatever solution we come up with from preventing row hover state as you drag a divider should be applicable to other column headers, so my hope is that these CSS selectors (and ultimately the `not-active` attribute) will ultimately go away.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+            .join(' ');
+    }
+
+    /**
+     * Sets up state related to interactively sizing a column.
+     * @param columnIndex The column index currently being hovered over
+     * @param activeColumnDivider The divider that was clicked on (columnIndex - 1 for left divider, columnIndex for right)
+     */
+    public beginColumnInteractiveSize(
+        columnIndex: number,
```
mollykreis, line 99: Why do you need the `columnIndex` here? I would think knowing which divider was clicked would be enough. For example, what's the difference between clicking on the right divider of column 0 vs the left divider of column 1 (i.e. the difference of passing `0, 0` and `1, 0`)?

packages/nimble-components/src/table/index.ts
```diff
@@ -514,6 +559,7 @@ export class Table<
         }
 
         this.observeColumns();
+        this.tableLayoutManager.updateTableViewportMinWidth();
```
mollykreis, line 562: Why does the table viewport min width need to be updated when the columns change? Looking at the implementation of `updateTableViewportMinWidth`, I'm guessing it has to do with the calculation for the total minimum column width. Is that correct? If so, wouldn't that code need to be run again in the `update` code path due to a column configuration changing?

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+    private readonly onDividerMouseUp = (): void => {
+        document.removeEventListener('mousemove', this.onDividerMouseMove);
+        document.removeEventListener('mouseup', this.onDividerMouseUp);
+        this.unflagActiveColumnDividers();
+        this.updateTableViewportMinWidth();
+        this.resetGridSizedColumns();
+        this.table.isColumnBeingSized = false;
+    };
+
+    private getAllColumnsMinimumWidth(): number {
```
mollykreis, line 174: Doesn't look like this handles hidden columns.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+        return totalColumnMiniumWidth;
+    }
+
+    private getTotalColumnFixedWidth(): number {
+        let totalColumnFixedWidth = 0;
+        for (const column of this.table.columns) {
+            totalColumnFixedWidth += (column.columnInternals.currentPixelWidth
+                !== undefined)
+                ? column.columnInternals.currentPixelWidth
+                : 0;
```
mollykreis, line 188: I think this can be simplified
```suggestion
            totalColumnFixedWidth += column.columnInternals.currentPixelWidth ?? 0;
```

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    private setColumnsToFixedSize(): void {
+        this.cacheGridSizedColumns();
+        const headers = this.table.columnHeadersContainer.querySelectorAll(
+            '.header-container'
+        );
+        for (let i = 0; i < headers.length; i++) {
+            this.table.columns[i]!.columnInternals.currentPixelWidth = headers[i]!.getBoundingClientRect().width;
+        }
```
mollykreis, line 207: I don't think this code handles hidden columns correctly. We don't create the `.header-container` element for hidden columns which means that your `headers` variable isn't guaranteed to be a parallel array to `this.table.columns`.

packages/nimble-components/src/table/template.ts
```diff
@@ -38,95 +38,112 @@ export const template = html<Table>`
...
-                                    :nestingLevel="${(x, c) => c.parent.tableData[x.index]?.nestingLevel}"
-                                    :leafItemCount="${(x, c) => c.parent.tableData[x.index]?.leafItemCount}"
-                                    :groupColumn="${(x, c) => c.parent.tableData[x.index]?.groupColumn}"
-                                    ?selectable="${(_, c) => c.parent.selectionMode === TableRowSelectionMode.multiple}"
-                                    selection-state="${(x, c) => c.parent.tableData[x.index]?.selectionState}"
-                                    @group-selection-toggle="${(x, c) => c.parent.onRowSelectionToggle(x.index, c.event as CustomEvent<TableRowSelectionToggleEventDetail>)}"
-                                    @group-expand-toggle="${(x, c) => c.parent.handleGroupRowExpanded(x.index, c.event)}"
-                                >
-                                </${tableGroupRowTag}>
+            <div class="glass-pane" style="
```
atmgrifter00, line 45: This `div` is responsible for preventing hover states on elements during a column sizing operation. I'm not sure of a cleaner way to provide this behavior. Additionally, this proved difficult to write a unit test for, as I believe we ultimately need an API that actually moves a mouse to a position on screen.

Open to a class name other than `glass-pane` as well.

packages/nimble-components/src/table/template.ts
```diff
@@ -38,95 +38,112 @@ export const template = html<Table>`
...
+                --ni-private-glass-pane-pointer-events: ${x => (x.isColumnBeingSized ? 'none' : 'default')};
+            ">
+                <div role="rowgroup" class="header-row-container">
+                    <div class="header-row" role="row">
+                        <span class="header-row-action-container" ${ref('headerRowActionContainer')}>
```
atmgrifter00, line 50: I initially added this wrapper element (and provided it a `ref`) because at one point my implementation needed to more easily track when the size of the combination of the contained elements changed. While this is no longer the case, its presence does clean up a pageObject implementation, so I'm not sure if that is enough to keep this around.

packages/nimble-components/src/table/styles.ts
```diff
@@ -89,16 +83,77 @@ export const styles = css`
...
-        display: flex;
+    .column-divider {
+        border-left: 2px solid ${popupBorderColor};
+        display: none;
+        height: ${controlSlimHeight};
+        cursor: col-resize;
+        position: absolute;
+    }
+
+    .column-divider::before {
```
atmgrifter00, line 109: This CSS allows us to provide a larger hit-test region for the divider without adjusting its visual position. I've started with a width of 8px, which seems pretty forgiving, but am open to suggestion here.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+        return this.table.columns
+            ?.filter(column => !column.columnHidden)
+            .map(column => {
+                const {
+                    minPixelWidth,
+                    currentPixelWidth,
+                    currentFractionalWidth
+                } = column.columnInternals;
+                if (currentPixelWidth) {
+                    const coercedPixelWidth = Math.max(
```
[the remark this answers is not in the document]
atmgrifter00: This idea actually doesn't work for a grid contained within a flex box. The usage of `minmax` will allow for the grid to be sized smaller than what is specified by the max parameter. We need the element with the `grid` display to size to the full width specified by the max parameter.

packages/nimble-components/src/table/index.ts
```diff
@@ -173,11 +193,17 @@ export class Table<
      * @internal
      */
     @observable
+    public tableWidthFactor = 1;
```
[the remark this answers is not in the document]
atmgrifter00: This has been reworked such that we are no longer using a width factor.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,353 @@
...
+            entries => {
+                if (entries[0]) {
+                    this.updateTableViewportMinWidth();
+                }
+            }
+        );
+    }
+
+    public connectedCallback(): void {
+        this.headerRowActionContainerResizeObserver.observe(
```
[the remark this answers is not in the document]
atmgrifter00: There is no longer the need for a `connectedCallback` implementation.

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: I've reworked the logic to remove this.

packages/nimble-components/src/table/index.ts
atmgrifter00: No longer doing this.

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: We now are handling hidden columns.

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: Handling hidden columns now.

packages/nimble-components/src/table/styles.ts
atmgrifter00: Removed.

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: One interesting behavior I noticed in initial Storybook testing: if you start a column resize and then release your mouse outside the browser, the drag operation isn't canceled. The result is that if you then bring your mouse back into the browser, the drag operation is still happening even though the mouse is no longer down. I'm seeing this on Safari and Edge on macOS but interestingly not Firefox (unless I'm doing something weird, I was testing in a rush).

jattasNI: A similar weird behavior: start the drag with a right click instead of a left click. The browser context menu appears but if you dismiss it there is still a drag operation in progress even though the mouse isn't pressed.

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
jattasNI, line 2: I think it'd be worthwhile to add review and update our Storybook docs on table column sizing. I don't see much to change but it might be helpful to modify the parameters for "Fractional width configuration" and `min-pixel-width` to mention that:
1. fractional width is just about initial size and size when the table is resized. It's not updated when you resize a column
2. min-pixel-width is applied when the table resizes or when a column is resized

packages/nimble-components/src/table/styles.ts
```diff
@@ -89,16 +83,77 @@ export const styles = css`
...
+        content: '';
+        position: absolute;
+        width: 8px;
+        height: 100%;
+        left: -5px;
```
jattasNI, line 114: Where does `-5px` come from? Is it possible to express it as a calculation of other named values?

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,263 @@
+import type { Table } from '..';
+import type { TableColumn } from '../../table-column/base';
+import type { TableRecord } from '../types';
+
+/**
+ * This class manages the layout of columns within a Table.
+ * @interal
+ */
+export class TableLayoutManager<TData extends TableRecord> {
```
jattasNI, line 9: I didn't look at this file in any amount of detail. I think someone besides you should understand it and take a closer look but it needn't be every reviewer. I think I heard you mention that person was likely to be @mollykreis but I'm happy to step in and get an overview if that isn't true.

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
atmgrifter00: Regarding your first issue, I think this is a Storybook quirk. You don't see the same behavior in the Angular application. Additionally, you don't see the noted behavior in Chrome with vanilla JS apps in StackBlitz that have a similar setup. Not sure how we want to proceed given that.

For the second issue, I think I can just be more guarded in my event handling where I only process left clicks (or the "primary" button).

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -447,6 +448,58 @@ export class TablePageObject<T extends TableRecord> {
...
+        const mouseUpEvent = new MouseEvent('mouseup');
+        document.dispatchEvent(mouseUpEvent);
+    }
+
+    public getColumnDivider(index: number): HTMLElement {
+        const headerContainers = this.tableElement.shadowRoot!.querySelectorAll('.header-container');
+
+        if (index >= headerContainers.length - 1) {
+            throw new Error(
+                'Attempting to index past the number of available dividers. Max index is two less than the number of columns.'
```
mollykreis, line 482: Would it be better to write this to return `HTMLElement | null` so that you only throw if you are trying to index past the number of columns, not the number of dividers. Then you can write tests that the right-most column doesn't have a divider on its right -- which currently is a bug in the code based on where the hidden columns fall within the DOM.

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -164,7 +190,7 @@ describe('Table Column Sizing', () => {
                 column2ExpectedRenderedWidth: 100
             }
         ];
-        const focused: string[] = [];
+        const focused: string[] = ['both columns use pixelWidth'];
```
mollykreis, line 193: I'm guessing you aren't planning on checking this code in with only one of these tests running?

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Adding interactive column sizing",
+  "packageName": "@ni/nimble-components",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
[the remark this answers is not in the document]
atmgrifter00: To fix this I actually had to remove a call to `Math.floor`.

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
jattasNI: For the first issue, I can confirm that it's only reproducible in Storybook and not in the Angular example app. That may mean it's reproducible in cases where clients use iframes which is not ideal but also likely not a priority for us to fix. I found a couple threads that seem related but didn't look in detail whether there are fixes we could leverage:
- [SO post suggesting it's related to `preventDefault`](https://stackoverflow.com/questions/66703382/how-to-keep-watching-mouse-move-outside-iframe)
- [Firefox bug suggesting the behavior in other browsers is correct](https://bugzilla.mozilla.org/show_bug.cgi?id=1681647)

I'm fine to leave as-is if those threads don't yield a simple fix.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,263 @@
...
+        const leftDividerIndex = this.activeColumnDivider! > 0
+            ? this.activeColumnDivider!
+                  - ((this.activeColumnDivider! + 1) % 2)
+            : 0;
+        const rightDividerIndex = leftDividerIndex === 0
+            || leftDividerIndex === visibleColumnLength + 1
+            ? leftDividerIndex
+            : leftDividerIndex + 1;
+        const dividers = this.table.shadowRoot!.querySelectorAll('.column-divider');
+        Array.from(dividers).forEach((divider, i) => {
```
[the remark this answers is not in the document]
atmgrifter00: 🤦‍♂️

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,263 @@
...
+            });
+        }
+    }
+
+    private resetGridSizedColumns(): void {
+        if (!this.gridSizedColumns) {
+            return;
+        }
+
+        const largestColumnFixedSize = Math.max(
```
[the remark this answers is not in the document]
atmgrifter00: The algorithm has changed to simply be:
currentFractionalWidth = (currentPixelWidth / initialPixelWidth) * initialFractionalWidth

So, now, if we have a series of fractional width columns: [1, 1, 1, 1], with an initial table total column width of 400 (and no minimum column sizes), and then size the first column by 150 pixels (it started off as 100), then the resulting fractional widths will be [2.5, .5, .5, .5].

This means that if I add a new column after the above sizing operation, it will have its expected default render width when considering its default fractional width of 1 (be twice as big as the smallest columns).

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -245,35 +245,18 @@ export class TablePageObject<T extends TableRecord> {
...
 
-        const collapseButton = this.getCollapseAllButton();
-        const buttonWidth = collapseButton!.getBoundingClientRect().width;
-        const buttonStyle = window.getComputedStyle(collapseButton!);
         table.style.width = `${
             rowWidth
-            + buttonWidth
-            + parseFloat(buttonStyle.marginLeft)
-            + parseFloat(buttonStyle.marginRight)
+            + (table.headerRowActionContainer.getBoundingClientRect().width
```
mollykreis, line 250: Nothing on this line immediately jumps out to me as being possibly undefined. Can you double check that the `?? 0` is needed?

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -459,6 +442,114 @@ export class TablePageObject<T extends TableRecord> {
...
+     * @param columnIndex The index of the column to the left of a divider being dragged. Thus, this
+     * can not be given a value representing the last visible column index.
+     * @param deltas The series of mouse movements in the x-direction while sizing a column.
+     */
+    public dragSizeColumnByRightDivider(
+        columnIndex: number,
+        deltas: number[]
+    ): void {
+        const divider = this.getColumnRightDivider(columnIndex);
+        if (divider === null) {
```
mollykreis, line 455: Any reason not to simplify this condition?
```suggestion
        if (!divider) {
```

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -459,6 +442,114 @@ export class TablePageObject<T extends TableRecord> {
...
+     * @param columnIndex The index of the column to the right of a divider being dragged. Thus, this
+     * value must be greater than 0 and less than the total number of visible columns.
+     * @param deltas The series of mouse movements in the x-direction while sizing a column.
+     */
+    public dragSizeColumnByLeftDivider(
+        columnIndex: number,
+        deltas: number[]
+    ): void {
+        const divider = this.getColumnLeftDivider(columnIndex);
+        if (divider === null) {
```
mollykreis, line 491: Any reason not to simplify this condition?
```suggestion
        if (!divider) {
```

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -459,6 +442,114 @@ export class TablePageObject<T extends TableRecord> {
...
+        columnIndex: number,
+        deltas: number[]
+    ): void {
+        const divider = this.getColumnRightDivider(columnIndex);
+        if (divider === null) {
+            throw new Error(
+                'The provided column index has no right divider associated with it.'
+            );
+        }
+        divider.setAttribute('active', 'true');
```
mollykreis, line 460: `active` is a CSS class now rather than an attribute, right? Is this line still needed? If so, should there be an equivalent line setting `active` back to `false` at the end of the function?

Same question for `dragSizeColumnByLeftDivider` below.

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -321,3 +347,487 @@ describe('Table Column Sizing', () => {
...
+    });
+
+    afterEach(async () => {
+        await disconnect();
+    });
+
+    describe('No hidden columns ', () => {
+        const columnSizeTests = [
+            {
+                name: 'sizing left only affects adjacent right column with delta less than min width',
```
mollykreis, line 373: The test names and drag deltas here don't seem to match. For example, this test claims to be "sizing left" but provides a positive drag delta.

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -321,3 +347,487 @@ describe('Table Column Sizing', () => {
...
+        await waitForUpdatesAsync();
+        await pageObject.sizeTableToGivenRowWidth(400, element);
+    });
+
+    afterEach(async () => {
+        await disconnect();
+    });
+
+    describe('No hidden columns ', () => {
+        const columnSizeTests = [
```
mollykreis, line 371: Is there a test that covers moving the mouse farther to the left than can columns can shrink, then moving the mouse to the right a little bit and ensuring column widths don't change (i.e. that the divider gets re-synced to the cursor's position)? I didn't see one, but it's quite possible I missed it. If there isn't one, could you add one?

packages/nimble-components/src/table/index.ts
```diff
@@ -515,6 +586,9 @@ export class Table<
 
         this.observeColumns();
         this.tableUpdateTracker.trackColumnInstancesChanged();
+        this.visibleColumns = this.columns.filter(
```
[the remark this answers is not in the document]
atmgrifter00: Nope, we don't.

packages/nimble-components/src/table/index.ts
```diff
@@ -483,6 +538,22 @@ export class Table<
...
+    public getHeaderContainerElements(): NodeListOf<Element> {
+        return this.columnHeadersContainer.querySelectorAll(
+            '.header-container'
+        );
+    }
+
+    /**
+     * @internal
+     */
+    public getColumnDividerElements(): NodeListOf<Element> {
```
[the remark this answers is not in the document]
atmgrifter00: Nope. Removed.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,255 @@
...
+        const headers = this.table.getHeaderContainerElements();
+        for (let i = 0; i < headers.length; i++) {
+            this.visibleColumns[i]!.columnInternals.currentPixelWidth = headers[i]!.getBoundingClientRect().width;
+        }
+        this.cacheColumnInitialPixelWidths();
+    }
+
+    private getAllowedSizeDelta(
+        activeDividerIndex: number,
+        delta: number
```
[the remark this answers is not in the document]
atmgrifter00: I've changed it to `requestedResizeAmount`, but I'm wondering if @jattasNI or @rajsite have other suggestions that might better indicate that this value can be negative. For whatever reason, I think a negative "amount" sounds odd, whereas a negative "delta" does not.

packages/nimble-components/src/table/testing/table.pageobject.ts
atmgrifter00: Removed.

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
atmgrifter00: Added test.

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
jattasNI, line 1: Reminder to update your PR description (it says tests are still in work)

change/@ni-nimble-components-752d1b02-6122-4d2f-b737-809aff6729d2.json
jattasNI, line 2: I updated from main again to see if it fixes the spurious Chromatic diffs.

jattasNI (approved):
Given that Molly's out and has given this a thorough review, I think it's ok to submit without her official approval (once Milan approves)

packages/nimble-components/src/table/template.ts
```diff
@@ -39,97 +39,112 @@ export const template = html<Table>`
             --ni-private-table-scroll-height: ${x => x.virtualizer.allRowsHeight}px;
             --ni-private-table-row-container-top: ${x => x.virtualizer.rowContainerYOffset}px;
             --ni-private-table-row-grid-columns: ${x => x.rowGridColumns ?? ''};
+            --ni-private-table-cursor-override: ${x => (x.layoutManager.isColumnBeingSized ? 'col-resize' : 'default')};
+            --ni-private-table-row-min-width: ${x => `max(100%, ${x.tableScrollableMinWidth}px)`};
+            --ni-private-table-header-row-min-width: ${x => `max(100%, calc(${x.tableScrollableMinWidth}px + ${x.virtualizer.headerContainerMarginRight}px))`};
```
rajsite, line 44: Can we move more of the css logic into the stylesheet and keep the property definition to a minimum here?
For example:
```
--ni-private-table-scrollable-min-width: ${x => `${x.tableScrollableMinWidth}px)`}
```

And then we already have headerContianerMarginRight as a property `--ni-private-table-header-scrollbar-spacer-width`

style:
```
min-width: max(100%, calc(var(--ni-private-table-scrollable-min-width) + var(--ni-private-table-header-scrollbar-spacer-width));
```

I think that also illustrates as issue with the mental model here. These css custom property definitions are trying to define bindings to property values that can be used in styles, not styles themselves. They are just making observable properties usable in CSS.

So we may want to rename `--ni-private-table-header-scrollbar-spacer-width` to something that better aligns to the property it corresponds to, ie `--ni-private-table-header-container-margin-right`.
Then it'll be easier to say in the CSS that the `.header-row` `min-wdith` is a function of the right margin and scrollable min width, etc.

Wouldn't mind a pass in this PR or a separate one that changed the names of most of these to better align to the property they are exposing. I realize it doesn't work perfectly, for example for booleans and the row grid columns, but I think we can improve the correlation a bit.

packages/nimble-components/src/table/template.ts
```diff
@@ -39,97 +39,112 @@ export const template = html<Table>`
...
-                                            name="${x => x}"
-                                            slot="${x => `row-action-menu-${x}`}">
-                                        </slot>
+                                    <${iconTriangleTwoLinesHorizontalTag} slot="start"></${iconTriangleTwoLinesHorizontalTag}>
+                                    ${x => tableGroupsCollapseAllLabel.getValueFor(x)}
+                                </${buttonTag}>
+                            </span>
+                        </span>
+                        <span class="column-headers-container" ${ref('columnHeadersContainer')}>
+                            ${repeat(x => x.visibleColumns, html<TableColumn>`
```
rajsite, line 76: think you can specify the parent type here and avoid some `as` casts
```suggestion
                            ${repeat(x => x.visibleColumns, html<TableColumn, Table>`
```

packages/nimble-components/src/table/template.ts
```diff
@@ -39,97 +39,112 @@ export const template = html<Table>`
...
+                                <div class="header-container">
+                                    ${when((_, c) => c.index > 0, html<TableColumn, Table>`
+                                        <div class="column-divider left ${(_, c) => `${c.parent.layoutManager.activeColumnIndex === c.index ? 'active' : ''}`}" @mousedown="${(_, c) => c.parent.onLeftDividerMouseDown(c.event as MouseEvent, c.index)}"></div>
+                                    `)}
+                                        <${tableHeaderTag}
+                                            class="header"
+                                            sort-direction="${x => (typeof x.columnInternals.currentSortIndex === 'number' ? x.columnInternals.currentSortDirection : TableColumnSortDirection.none)}"
+                                            ?first-sorted-column="${(x, c) => x === (c.parent as Table).firstSortedColumn}"
+                                            @click="${(x, c) => (c.parent as Table).toggleColumnSort(x, (c.event as MouseEvent).shiftKey)}"
```
rajsite, line 85: with the above can likely skip the `as` Table cast

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,255 @@
...
+        this.dragStart = dragStart;
+        this.currentTotalDelta = 0;
+        this.visibleColumns = this.getVisibleColumns();
+        this.setColumnsToFixedSize();
+        this.initialTableScrollableWidth = this.table.viewport.scrollWidth;
+        this.initialTableScrollableMinWidth = this.table.tableScrollableMinWidth;
+        this.initialColumnTotalWidth = this.getTotalColumnFixedWidth();
+        this.isColumnBeingSized = true;
+        document.addEventListener('mousemove', this.onDividerMouseMove);
+        document.addEventListener('mouseup', this.onDividerMouseUp);
```
rajsite, line 80: Extraneous context:

In the [WebVI days](https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FVI%2FHtmlControls.Design%2FWeb%2FFramework%2FniUIActivityService.js&_a=contents&version=GBmain) found there are a bunch of extra cases to deal with. mouseup is not guaranteed so should also listen for window blur (and probably other events) and then there is the slew separate touch events which were deprecated. Which was funny because touch events were deprecated and the replacement api pointer events wasn't standard yet 🤣 BUT TODAY...

Recommendation:

We should switch to [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) which are mostly analogous to mouse events but provide a unified api across mouse, touch, arbitrary "pointer" concepts.

In particular this use-case seems well suited for the [setPointerCapture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) workflow which will intercept all other events for us, guarantee the life cycle completes, and actually might obviate the need for the pane of glass (if true, sorry it didn't occur to me sooner 😵)

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
```
rajsite, line 81: Have we thought of an intent to cancel the column resize? I found myself wanting to push the `esc` to abort an ongoing resize operation.

rajsite: Might be pretty straightforward to represent under pointer events. i.e. the esc key would just restore the sizes and releasePointerCapture (which fires pointercancel I believe) and leaves you ready for the next pointerdown.

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: Agreed. I've filed [this issue](https://github.com/ni/nimble/issues/1362) for investigating the usage of pointer events.

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: I think we can tackle this along with the pointer events I've created [this issue](https://github.com/ni/nimble/issues/1362) for.

rajsite (approved):

PR #1134, opened 2023-03-24
Storybook 7 upgrade

jattasNI:
# Pull Request

## 🤨 Rationale

We want to be on the latest storybook version because it fixes issues we've had with dependencies and it enables more powerful docs pages via mdx and Component Story Format enhancements.

Notable changes:
- landing page uses correct fonts and errors in console are gone (fixing #825 and #943 properly)
- [no more Docs tab](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#70-docs-changes). Each story group has a docs page under it instead.
- XD previews are gone (feature was deprecated)
- Actions not working yet and some code previews have stopped working; see "Follow ups" below

## 👩‍💻 Implementation

Following the [upgrade guide](https://storybook.js.org/docs/react/configure/upgrading#prereleases) didn't work very well because it is buggy with a monorepo. Fred's general strategy was to run the upgrade command from within `nimble-components` then delete the `node_modules` and `package-lock.json` that it creates there, then re-install everything from the root. The migration tool is responsible for the changes to:
1. storybook configuration files
2. package.json
3. renaming types like `Story` -> `StoryFn` and `storyName` to `name`
4. deleting the `withXD` decorator which is no longer supported

Beyond that there were a few other manual changes:
1. establish static dependencies on components needed by each story. Storybook seems to have gotten more aggressive about compiling away unneeded components so if something isn't imported, it won't be found and components won't render correctly. This meant:
    1. replacing a few literals for element tag names in stories with the `elementTag` export
    2. updating the `nimble-menu` to explicitly depend on `nimble-anchored-region` (see comment on `menu/index.ts`)
 2. Enable 'autodocs' for component and token stories but disable it for test stories
 3. Add a patch file to work around https://github.com/storybookjs/storybook/issues/21820
 4. simplified return types of story generation functions in `storybook.ts` in response to storybook type changes
5. regenerated landing page screenshot and annotation now that docs tab is gone

## 🔜 Follow ups in future PRs

Fixing these issues is tracked by #1146 . I tried upgrading to ts 4.7 but it didn't fix any of them.

- [ ] https://github.com/storybookjs/storybook/issues/21820
- [ ] Figure out why actions aren't firing when you click a button. Likely [need this decorator](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#removed-auto-injection-of-storybookaddon-actions-decorator) but it's not present in current rc: https://github.com/storybookjs/storybook/issues/21308
[...]

PR #1322, opened 2023-06-23
Remove @storybook/html peer dependency from root package.json and re-enable audit for prod

### LINKED ISSUE #1317: Revert loosening of npm audit strictness
## 🧹 Tech Debt
 The `npm audit` step of our pipeline [was failing](https://github.com/ni/nimble/actions/runs/5349242650/jobs/9700181965) due to [a vulnerability with semver](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw). We made it less strict in #1318 to get the build passing, but we should fix the vulnerability and make it strict again instead.

jattasNI:
# Pull Request

## 🤨 Rationale

Fixes #1317.

The `npm audit` of production dependencies was failing because of deps that Storybook brought in. But Storybook should really be a dev dependency.

Storybook is listed as a production dependency because it's in `peerDependencies` in the root `package.json`. We added it there in [this commit](https://github.com/ni/nimble/pull/1134/commits/321cdd399adbc0e755e2f0acddeb0558c6e5bffb) of the PR that migrated us to Storybook 7. I believe the rationale was that it was necessary to apply [a patch](https://github.com/ni/nimble/blob/main/patches/%40storybook%2Bhtml%2B7.0.0.patch) to the package. (I'd like to remove that patch but I think it's not possible until https://github.com/storybookjs/storybook/issues/22384 is available. Currently it's only released in an alpha branch)

## 👩‍💻 Implementation

1. Remove `@storybook/html` from `peerDependencies` in the root `package.json`
2. `git clean -fdx`
3. `npm install` to regenerate `package-lock.json`
4. Re-enable audit for all severity levels for prod dependencies

## 🧪 Testing

I locally verified that the patch was still applied to the file inside node_modules. If it isn't applied, I believe we'd see a build error.

I locally verified that `npm audit --only=prod` succeeds now.

Otherwise relying on the PR build.

## ✅ Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,5 @@
     "beachball": "^2.31.0",
     "cross-env": "^7.0.3",
     "patch-package": "^6.4.7"
-  },
-  "peerDependencies": {
-    "@storybook/html": "*"
   }
 }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
       - run: npm outdated || exit 0
 
       # Audit
-      - run: npm audit --only=prod --audit-level=high
+      - run: npm audit --only=prod
       # https://github.com/ni/nimble/issues/801
       # - run: npm audit --audit-level=critical
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,9 +23,6 @@
         "beachball": "^2.31.0",
         "cross-env": "^7.0.3",
         "patch-package": "^6.4.7"
-      },
-      "peerDependencies": {
-        "@storybook/html": "*"
       }
     },
     "angular-workspace": {
@@ -192,6 +189,7 @@
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
       "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
+      "dev": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.1.0",
         "@jridgewell/trace-mapping": "^0.3.9"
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -45,8 +45,5 @@
     "beachball": "^2.31.0",
     "cross-env": "^7.0.3",
     "patch-package": "^6.4.7"
-  },
```
jattasNI, line 48: I'm not sure why we though it had to be a peerDependency in order to apply the `node_modules` patch. Perhaps that was for a different patch that has since been removed? @rajsite do you remember? Any testing you can think to do besides local build and PR build?

rajsite: The reason is that patch-package is not monorepo aware: https://github.com/ds300/patch-package/issues/356#issuecomment-1334750995

So we are making peer instead of dep to force a single copy* and making it peer at the root so it's pulled to the top.

\* technically this approach still isn't perfect. If the peer at root and a child devDep don't align a second copy will just be made in the child node_modules directory. We wouldn't want it to be a peerDependency in the child because then it becomes a production dependency. We would need a `peerDevDependencies` type concept to avoid that.

rajsite: Some other options:
- Hope and pray - The npm resolution algorithm will generally try to flatten things to the root when possible. Remove the peerDeps and it should "generally" work.
- Try and validate.
  - Maybe scan the lock file and assert there is only one version of the library ever resolved and it's in the root.
  - Maybe make a script that can be run in each package that runs `require.resolve('package-name')` and verifies that the package it resolves to is in the root node_modules folder and not somewhere else.

rajsite: Hope and pray may be fine. At least for these storybook packages I don't think it's likely we introduce multiple copies (ie add stroybook to the angular app or add a package where we develop a storybook plugin)

rajsite: The current lockfile looks good, so hope and pray would be technically fine so I'll mark approved if that seems like the right direction to go

rajsite (approved):

package.json
jattasNI: I'm going to go ahead with hoping it doesn't break on subsequent installs / signing ourselves up to validate when we do change update storybook. This storybook patches fail quite obviously when they aren't applied (compile error) and one of them already requires revisiting on every upgrade (it applies to a file with a uniquified name that changes in every version). So we're already stuck revalidating it the next time we update. And hopefully the next update allows us to delete them anyway!

PR #1327, opened 2023-06-26
Extend table page object

mollykreis:
# Pull Request

## 🤨 Rationale

In writing tests for client applications, I discovered a few additional pieces of information that would be beneficial to query from the table page object, so I've added them. They are:
- The array of currently sorted columns & their sort direction
- The array of currently grouped columns
- The number of rows that are in a group

## 👩‍💻 Implementation

- Added new capability to the table page object
- Added/updated tests to use the new functions

## 🧪 Testing

Ran unit tests

## ✅ Checklist


packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -447,6 +459,47 @@ export class TablePageObject<T extends TableRecord> {
...
+                    && typeof x.columnInternals.groupIndex === 'number'
+            )
+            .sort(
+                (a, b) => a.columnInternals.groupIndex!
+                    - b.columnInternals.groupIndex!
+            )
+            .map(x => x.columnId ?? '');
+    }
+
+    public getChildDisplayCountForGroup(groupRowIndex: number): string {
```
mollykreis, line 495: I went back and forth here on whether this should return a string or a number. I landed on a string because it was more straightforward to implement, and it provided what was required for the tests I'd like to write. I'm open to updating if reviewers think it would be more appropriate to return a number here.

rajsite: Think I'd lean toward a number since it's a "count".

rajsite (approved):

packages/nimble-components/src/table/testing/table.pageobject.ts
jattasNI: I'd also vote number.

jattasNI (approved):

Issue #1090, opened 2023-03-01
Unified strategy for providing localized labels within components


jattasNI:
### Labels for Icons
There are several places where components display icons. Each of these might benefit from: 
1. setting `title` so that sighted users can hover over them to see the icon explained in a tooltip
2. setting `aria-label` or other attributes so that screen reader users can understand their purpose
Those strings should be localized. Since we've decided that localization is an application concern and not a Nimble concern, we need to provide (hopefully consistent) APIs for clients to supply these strings.

We may also decide not to set `title` and to exclude the icon from the accessibility tree if its meaning is provided in other ways. We should consult with Leslie and with web best practice docs to come up with guidance for this.

The icons we've identified so far:
1. increment/decrement buttons on numeric text input #617 
2. action menu button in table cells #859 
3. sort icon in table header (and soon other icons from #885 like grouping, collapse all, column menu)
4. banner and tooltip severity

This will also inform related work like the following, though the solution might be different:
1. #1013 
2. icons as the primary content of a table column header

### Other Labels in Nimble Components
We also have other labels in Nimble components that will need a way for clients to localize them, such as the banner dismiss button, buttons (expand/collapse group, collapse all groups, etc) in the table, and menu items in the table.

## Goals

1. Document policy about when providing this content is necessary
2. Decide on an approach that's consistent across components
4. Document that approach
5. Ensure the work to adopt the approach for existing and known future use cases is tracked

## Possible Implementations

See HLD: #1272

## Tasks
* [x] Write HLD
* [x] Create nimble-label-provider-core and nimble-label-provider-table, and uptake them in nimble-components
* [x] Add label provider integration for Angular and Blazor
* [x] Remove banner dismissButtonLabel (obsolete due to dismiss label on LabelProviderCore)
* [ ] Revisit [per-component accessibility notes doc](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/accessibility.md#component-specific-accessiblity-notes) and try to document them in Storybook instead

mollykreis:
We need to make sure that the solution is scalable because some components, such as the table will have many strings that need to be localized. For example:
- label & tooltip for _Collapse All_ button
- label & tooltip for _Collapse Group/Row_ button
- label & tooltip for _Toggle selection of all rows_ checkbox for whole table & group rows
- label & tooltip for _Toggle selection_ checkbox
- Menu item text for items in the column menus -- _Sort ascending_, _Sort descending_, _Group by column_, _Size to fit_, etc

[...]

PR #1328, opened 2023-06-26
Add/uptake label providers in nimble-components

msmithNI:
# Pull Request

## 🤨 Rationale

Partially addresses #1090 (Angular/Blazor changes will be in subsequent PRs)

## 👩‍💻 Implementation

See the [HLD](https://github.com/ni/nimble/tree/main/specs/labels-and-localization) for the implementation plan for the label providers.

This PR creates the LabelProvider classes in nimble-components, and creates/uptakes label DesignTokens in `nimble-banner`, `nimble-number-field`, and `nimble-table`

## 🧪 Testing

- Verified that each label appears in the expected spots in the DOM for the affected components. Note that the current labels are mostly not user-visible (except that grouped table columns will now have a 'Grouped' tooltip on that icon in the header). Future table changes (i.e. the column header menu) will introduce new visible label tokens.
- Added autotests
    - For label providers: verify token naming, and that label provider attribute updates result in token updates
    - For controls using labels: verify that the label token values are used in the controls as expected
- Updated [Storybook](https://60e89457a987cf003efc0a5b-tbdnyzeaxa.chromatic.com/)
   - Each component using labels has a new "Localizable labels" row in their Storybook controls/API section
   - Under "Tokens" is a new ["Label Providers" section](https://60e89457a987cf003efc0a5b-tbdnyzeaxa.chromatic.com/?path=/docs/tokens-label-providers--docs). Each label provider has a section with a table that has the token name, JS property name, HTML attribute name, and English default value of the label.

## ✅ Checklist

  - Updated nimble-components Readme and Contributing docs with Localization sections.


packages/nimble-components/src/banner/tests/banner.stories.ts
```diff
@@ -121,11 +128,13 @@ export const _banner: StoryObj<BannerArgs> = {
         dismissButtonLabel: {
             name: 'dismiss-button-label',
             description:
-                'Set to a localized label (e.g. `"Close"`) for the dismiss button. This provides an accessible name for assistive technologies.'
+                'Set to a localized label (e.g. `"Close"`) for the dismiss button. This provides an accessible name for assistive technologies.',
+            control: { type: 'none' }
```
msmithNI, line 132: Note: These metadata/args changes just result in the 'Localizable labels' section being the last row in Storybook. With dismissButtonLabel/toggle being unspecified in the args object below, Localizable Labels appeared in the middle (before dismissButton/toggle). Once I gave them values in args, that resulted in a 'Set object' button appearing in their Storybook row without the control.type=none changes.

packages/nimble-components/src/label-provider/core/label-tokens.ts
```diff
@@ -0,0 +1,17 @@
+import { DesignToken } from '@microsoft/fast-foundation';
+import { coreLabelDefaults } from './label-token-defaults';
+
+export const alertDismissLabel = DesignToken.create<string>({
```
msmithNI, line 4: This label/token might apply to more components in the future - the [HLD had a comment](https://github.com/ni/nimble/pull/1272#discussion_r1228408606) with `banner-dismiss` and `toast-dismiss` as the example (but we don't have toast yet).

So I tried to pick a more general name and ended up with `alert-dismiss`. If people thinks it makes more sense to go with `banner-dismiss` (until/unless we have a 2nd component being added that would use it) let me know.

packages/nimble-components/src/label-provider/table/index.ts
```diff
@@ -0,0 +1,96 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-label-provider-table': LabelProviderTable;
+    }
+}
+
+/**
+ * Label provider for the Nimble table (and its sub-components and columns)
+ */
+export class LabelProviderTable extends LabelProviderBase {
```
msmithNI, line 22: Note: The [HLD](https://github.com/ni/nimble/tree/main/specs/labels-and-localization#plan-for-nimble-provided-labels) had a few labels whose tokens are not being added here yet - namely those that will be used for the column header menu & menu items, since it's not implemented yet. I figured we can add the tokens once we work on that (but if anyone wants to argue for having the tokens now, in the name of fewer label changes for clients, let me know)

packages/nimble-components/src/label-provider/core/tests/label-provider-core.stories.ts
```diff
@@ -0,0 +1,22 @@
...
+    LabelProviderArgs,
+    labelProviderMetadata
+} from '../../base/label-provider-stories-utils';
+import { labelProviderCoreTag } from '..';
+import * as labelTokensNamespace from '../label-tokens';
+
+const metadata = {
+    ...labelProviderMetadata,
+    title: 'Tokens/Label Providers',
+    tags: ['autodocs']
```
jattasNI, line 12: Could you add an MDX file for the label providers and disable autodocs? That would let us customize a few things:
- include stories for both providers on the docs page instead of just the core one
- consider hiding the "show code" button on the story. One way to do this is by importing it as `<Canvas>` rather than `<Story>` in the MDX
- add more markdown text providing context for each story

You can look at the button MDX for inspiration. I also have a local branch I can push if that would be helpful.

msmithNI: Took a stab at that, let me know what you think.
- Hid "Show Code" - along with `Canvas`, that also required [sourceState="none"](https://storybook.js.org/docs/react/api/doc-block-canvas#sourcestate)
- Added section headings for each provider/story. Wasn't too sure what additional markdown/text to put in each though so the blurb is pretty short right now.

packages/nimble-components/src/label-provider/base/label-provider-stories-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+    labelTokens: [string, DesignToken<string>][];
+    removeNamePrefix(tokenName: string): string;
+}
+
+const createTemplate = (
+    labelProviderTag: string
+): ViewTemplate<LabelProviderArgs> => html<LabelProviderArgs>`
+<${labelProviderTag}></${labelProviderTag}>
+<p>Element name: <code>${x => x.labelProviderTag}</code></p>
+<table>
```
jattasNI, line 24: Can you comment why you ended up using a native table and adding nimble-table styling to it? It looks a bit janky to me. My main complaints are the lack of row delineation and the inconsistent text alignment between headers and cells. Those would be easy enough to fix, but could we instead consider just using the nimble-table like we did for the component status story?

Edit: I see the theme aware token story is also using native tables. I think that's needed because of the varied contents of the token value column, but we don't have the same requirement for these tables.

change/@ni-nimble-components-6eb6eef1-eec2-4f6d-a98f-f7182ecdad50.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: Any idea why the Chromatic status check isn't reporting that the new stories add baselines that require approval?

rajsite (changes requested):

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -428,6 +428,31 @@ Nimble includes three NI-brand aligned themes (i.e. `light`, `dark`, & `color`).
 
 When creating a new component, create a `*-matrix.stories.ts` Storybook file to confirm that the component reflects the design intent across all themes and states.
 
+## Localization
+
+Nimble components may need to use some strings/labels (such as the label for a close button in a component's template, or menu items in a component-provided menu). Nimble exposes these localizable labels as design tokens, to support both localization and the ability for clients to override the strings.
```
jattasNI, line 433: It would be good to update this introductory paragraph (here and in README and in storybook) to provide some more context about the normal case being that clients provide localized strings and that this is for the rarer case where Nimble is providing the string and apps need to override it. Basically make this an answer to the question "how does Nimble handle localization?"

A rough draft:
Most user-visible strings displayed by Nimble components are provided by the client application and are expected to be localized by the application if necessary. However, some strings are built in to Nimble components and are provided only in English. An application can provide localized versions of these strings by using design tokens set on label provider elements.

change/@ni-nimble-components-6eb6eef1-eec2-4f6d-a98f-f7182ecdad50.json
msmithNI: Do you think we want/need Chromatic diffs for these stories?

Currently I opted them out of it (see [here](https://github.com/ni/nimble/blob/e070de537d61dfd326176d73e7f1070b39ca0c70/packages/nimble-components/src/label-provider/base/label-provider-stories-utils.ts#L53)), like some other documentation-related stories do. I should have commented on it to point it out and get feedback though.

packages/nimble-components/src/label-provider/base/label-provider-stories-utils.ts
msmithNI: Yep I started by adapting the theme-aware token story. I can try out using Nimble table instead.

msmithNI: Switched to nimble-table

packages/nimble-components/src/label-provider/base/label-provider-stories-utils.ts
```diff
@@ -0,0 +1,95 @@
+import { ViewTemplate, html, repeat } from '@microsoft/fast-element';
```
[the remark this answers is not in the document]
msmithNI: Moved

packages/nimble-components/src/label-provider/base/label-user-stories-utils.ts
```diff
@@ -0,0 +1,33 @@
+import type { Meta } from '@storybook/html';
```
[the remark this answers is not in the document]
msmithNI: Moved

packages/nimble-components/src/label-provider/base/label-name-utils.ts
```diff
@@ -0,0 +1,9 @@
+import { spinalCase } from '@microsoft/fast-web-utilities';
```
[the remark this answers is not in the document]
msmithNI: Moved

packages/nimble-components/CONTRIBUTING.md
msmithNI: Updated

change/@ni-nimble-components-6eb6eef1-eec2-4f6d-a98f-f7182ecdad50.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
msmithNI: I don't think we'll close out any of the issues until the Angular/Blazor integration are done too.

change/@ni-nimble-components-6eb6eef1-eec2-4f6d-a98f-f7182ecdad50.json
jattasNI: I don't feel too strongly but I think I'd vote for consistency with the other token stories (they don't opt out of Chromatic). Even though a change to these tokens isn't a visual diff from an app's point of view, as a reviewer I think I'd get some benefit from noticing that Chromatic has detected a change. Seeing those diffs might remind me to think about the change a bit more carefully.

msmithNI: Re-opted into Chromatic snapshots

packages/nimble-components/src/label-provider/base/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import { themeProviderTag } from '../../theme-provider';
+
+/**
+ * Base class for label providers
+ */
+export abstract class LabelProviderBase extends FoundationElement {
+    private readonly deferredTokenUpdates: Map<
+    DesignToken<string>,
+    string | undefined
+    > = new Map();
```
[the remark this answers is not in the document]
msmithNI: I think your suggestions make sense. I had tried something similar in earlier revisions but struggled some with the typing / how to get the property values semi-safely. Got it working so you can take a look now (might need new label-provider-base level tests but haven't done that yet).

A few things I wasn't sure about:
- Is `Observable.getNotifier`+`subscribe` safe to do in the constructor (vs `connectedCallback`)? It seems to work, and it looks like the FAST observable code doesn't necessarily rely on `$fastController`. I think we do need to subscribe before `connectedCallback` (again, for Blazor, the property sets will be before the DOM attach, from what I saw when testing my prototype).
- How to handle `propertyValue === undefined` in handleChange - I ended up deciding the value should be computed from `token.getValueFor(this)` in that case to match what would happen under the initialization codepath.

Also worth noting that the current type for `supportedLabels` ensures the keys are keys/properties on the class, but you'd be allowed to have a key of `$fastController` or something else we inherit from FAST. But the leaf-level tests for the tokens in `label-provider-core.spec.ts` and `label-provider-table.spec.ts` would catch that.

packages/nimble-components/src/number-field/index.ts
```diff
@@ -64,28 +68,28 @@ const nimbleNumberField = NumberField.compose<NumberFieldOptions>({
...
-    stepDownGlyph: html`
+    stepDownGlyph: html<NumberField>`
         <${buttonTag}
             class="step-up-down-button"
             appearance="ghost"
             content-hidden
             tabindex="-1"
         >
-            "Decrement"
+            ${x => numberFieldDecrementLabel.getValueFor(x)}
```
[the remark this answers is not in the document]
msmithNI: Added per-control label tests for the banner, number field, and table.

My first thought was that they might be overkill, but then again, they did catch that I didn't have `tableCellActionMenuLabel` hooked up in the table previously. So they're probably useful.

packages/nimble-components/src/label-provider/base/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        if (this.supportedLabels[property]) {
+            const token = this.supportedLabels[property]!;
+            let value = this[property] as string | undefined;
+            if (value === undefined) {
+                value = token.getValueFor(this);
+            }
+            token.setValueFor(this, value);
+            if (this.themeProvider) {
+                token.setValueFor(this.themeProvider, value);
+            }
```
rajsite, line 44: If the user is deleting the value off the provider, then I think we should just delete the token value off the provider element instead of fetching / calculating and caching it by calculating the value and setting it on the provider. i.e. lean towards the cascade being the source of truth.

```suggestion
            if (value === undefined) {
                value = token.deleteValueFor(this);
            } else {
                token.setValueFor(this, value);
            }
            if (this.themeProvider) {
                token.setValueFor(this.themeProvider, token.getValueFor(this));
            }
```
Edit: Actually after a bit more thought it probably looks different, continue below.

The design token cascade being the source of truth is important because the design token cascade is observable. I think that is missing in the provider base right now. We should be observing the value of the design tokens on the provider.

I think the high level algorithm is:

- Label provider creates observers for all token values it cares about on itself
- If user sets label provider property / attribute, reflect that on the token value for the provider itself (ie set or delete the token value on self but don't reflect to theme provider yet)
- On changes to the token value on provider itself, reflect to theme provider (also do that in connected as needed)

rajsite: I think the test for the above is a cascade of theme providers:

```html
ie start:
<nimble-theme-provider>
    <label-provider-core my-label="test-parent"></label-provider-core>

     <nimble-theme-provider>
         <label-provider-core my-label="test-child"></label-provider-core>
     </nimble-theme-provider>
         <element-in-test>
</nimble-theme-provider>

then clear child:
<nimble-theme-provider>
    <label-provider-core my-label="test-parent"></label-provider-core>

     <nimble-theme-provider>
         <label-provider-core></label-provider-core>
         <element-in-test>
     </nimble-theme-provider>
</nimble-theme-provider>


set parent:
<nimble-theme-provider>
    <label-provider-core my-label="test-parent-new-value"></label-provider-core>

     <nimble-theme-provider>
         <label-provider-core></label-provider-core>
         <element-in-test><!-- Currently element would be stuck with cached value on theme provider instead of observing cascade changes -->
     </nimble-theme-provider>
</nimble-theme-provider>
```

packages/nimble-components/src/label-provider/core/label-tokens.ts
rajsite: Maybe popup-dismiss? We have popupBorderColor in our existing tokens so popup as a generic element is there, don't see one for alert (which feels a bit too specific? like a snackbar or even a persistent tooltip etc feel like they could have a dismiss button)

packages/nimble-components/src/label-provider/core/label-tokens.ts
```diff
@@ -0,0 +1,17 @@
...
+    name: 'alert-dismiss-label',
+    cssCustomPropertyName: null
+}).withDefault(coreLabelDefaults.alertDismissLabel);
+
+export const numberFieldDecrementLabel = DesignToken.create<string>({
+    name: 'number-field-decrement-label',
```
rajsite, line 10: Can we do numeric-increment and numeric-decrement?
Feel like anything numerical that may have those operations could use it, doesn't have to be specifically number-field (though is today)

packages/nimble-components/src/label-provider/table/index.ts
rajsite: I'm alright adding them when we need them

packages/nimble-components/src/table/components/header/template.ts
```diff
@@ -20,8 +21,8 @@ export const template = html<TableHeader>`
         ${when(x => x.sortDirection === TableColumnSortDirection.descending, html`
             <${iconArrowDownTag} class="sort-indicator" aria-hidden="true"></${iconArrowDownTag}>
```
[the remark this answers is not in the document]
msmithNI: Currently I wasn't planning to add sort labels. From the [HLD](https://github.com/ni/nimble/blob/main/specs/labels-and-localization/README.md?plain=1#L27):

> No accessible label specifically for these icons is currently planned. The primary sorted column is already indicated with `aria-sort="ascending"` or `aria-sort="descending"`. The [ARIA APG's sortable table example](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) follows similar logic, and says sort labels are not added to each column header button "to prevent repetitious verbosity that could interfere with understanding of the column titles".

change/@ni-nimble-components-6eb6eef1-eec2-4f6d-a98f-f7182ecdad50.json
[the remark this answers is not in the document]
msmithNI: Done

(banner dismissButtonLabel is the only one I'm aware of so phrased it in terms of that)

packages/nimble-components/src/table/template.ts
```diff
@@ -59,6 +60,7 @@ export const template = html<Table>`
                             @click="${x => x.handleCollapseAllGroupRows()}"
```
jattasNI, line 60: Should we also add `title` for this button so that we show a tooltip? I couldn't find documentation of a decision (the button storybook doesn't have any guidance about tooltips for icon buttons and #1090 just says we might want to do it). But my opinion is that this button especially isn't a standard icon so it could benefit from a tooltip to explain what it does.

I could go either way on the group expand/collapse button and the action menu button needing titles. I don't have justification for my sense that they don't need them other than "I can tell what they do from their icons", which is not very good justification. So maybe they need them too.

packages/nimble-components/src/table/components/header/template.ts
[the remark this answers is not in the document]
rajsite: excellent, maybe add a comment to that effect saying that avoiding a title on the sort icons is intentional?

jattasNI (approved):

packages/nimble-components/src/table/template.ts
msmithNI: Agreed, since this button/icon is pretty nonstandard it could probably use a title/tooltip, I'll add that.
Also agreed that the other buttons don't really need one.

packages/nimble-components/src/label-provider/base/index.ts
msmithNI: Pushed [a commit](https://github.com/ni/nimble/pull/1328/commits/79b8cc9c490a224fbd6646fbbef26d0f676d6548) to try to adopt this proposal.

I ran into multiple issues getting it to work. I think the fact that the label providers are children of the theme provider ends up complicating things quite a bit.

Issues I hit:
- clearing a label property (the `token.deleteValueFor(labelProvider)` code path): FAST has [some code](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/design-token/design-token.ts#L650) that recomputes the local token value on delete. If we delete the value from the label provider first, and it's still set on the parent theme provider at this point, it'll compute the wrong value (which would be evident the next time `labelToken.getValueFor` was called).
   - If we add an explicit `token.deleteValueFor(this.themeProvider)` call at the same time / right afterwards, that ends up triggering our change function (from subscribing to the tokens). That would end up setting an explicit value for the label token on the theme provider (which seems wrong)
   - My current fix for this was to clear the token value (`deleteValueFor`) on the label provider, and do the same for (`deleteValueFor`) the theme provider too, and add code to suppress the change handler in that case (kinda ugly)
- what to do in the connectedCallback: I started by saying 'for all supported label tokens, `token.setValueFor(this.themeProvider, token.getValueFor(thisLabelProvider))`'. I don't think that's right because if the label wasn't set locally, that would explicitly set the default label value on the theme provider (so it wouldn't get cascaded values from its parent anymore). I changed it to 'for label properties on this provider that have a set value, ...' and that seem to work. (I added a separate test for this case - nested theme providers and label providers, with no initial values set, and only set a label value on the parent, and verify that it cascades).

If you're around next week you can take a look and see if what I have makes sense / could be simplified, otherwise we can discusss post-quiet-week.

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -428,6 +428,30 @@ Nimble includes three NI-brand aligned themes (i.e. `light`, `dark`, & `color`).
...
+
+-   element/type(s) to which the token applies, e.g. `number-field` or `table`
+    -   This may not be an exact element name, if this label applies to multiple elements or will be used in multiple contexts
+-   component part/category (optional), e.g. `column-header`
+-   specific functionality or sub-part, e.g. `decrement`
+-   the suffix `label` (will be omitted from the label-provider properties/attributes)
+
+Components using localized labels should document them in Storybook. To add a "Localizable Labels" section:
+
+-   Their story `Args` should extend `LabelUserArgs`
```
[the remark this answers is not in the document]
msmithNI: Since the story `Args` are all interfaces too, they have to be `extends` (`implements` is a TypeScript error for `interface implements interface`)

packages/nimble-components/src/table/components/header/template.ts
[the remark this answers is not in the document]
msmithNI: Added

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -428,6 +428,30 @@ Nimble includes three NI-brand aligned themes (i.e. `light`, `dark`, & `color`).
...
+## Localization
+
+Most user-visible strings displayed by Nimble components are provided by the client application and are expected to be localized by the application if necessary. However, some strings are built into Nimble components and are provided only in English. An application can provide localized versions of these strings by using design tokens set on label provider elements.
+
+There are currently 2 label providers:
+
+-   `nimble-label-provider-core`: Used for labels for all components besides the table
+-   `nimble-label-provider-table`: Used for labels for the table (and table sub-components / column types)
+
+To add new labels to a label provider, follow the existing patterns (create a `DesignToken<string>`, an `@attr`-backed property, and a `propertyChanged()` function that updates the token value).
```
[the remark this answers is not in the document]
msmithNI: Removed

packages/nimble-components/src/label-provider/core/label-tokens.ts
msmithNI: Updated

packages/nimble-components/src/table/template.ts
msmithNI: Added `title`

packages/nimble-components/src/label-provider/base/index.ts
rajsite: Ahhhhhhh okay. So using the token definition on the label provider doesn't work well because it is influencing it's parent and messing with the cascade. I still think the idea that the label provider should behave declaratively makes sense. But what if it's state is just the attribute values instead of the token value.

So same as before:
- Each label-provider has a readonly map of it's propertyName -> label tokens
- The attributes stay simple, they can have a default value of undefined

But now we just care about our attribute state:
- Base class we track for attribute change and just set that on the theme provider if available:
  ```ts
  handleChanged((_element, property) =>{
      if (this.supportedLabels[property]) {
          const token = this.supportedLabels[property];
          const value = this[property];
          if (this.themeProvider) {
              if (value === null || value === undefined) {
                  token.deleteValueFor(this.themeProvider);
             } else {
                  token.setValueFor(this.themeProvider, value);                
            }
          }
      }
  })
  ```
- And in the connectedCallback we reflect the current attribute state
  ```ts
  connectedCallback() {
      super.connectedCallback()
      this.themeProvider = // find it
      if (this.themeProvider) {
        for (const [property, token] of Object.entries(this.supportedLabels)) {
            const value = this[property];
            if (value === null || value === undefined) {
                  token.deleteValueFor(this.themeProvider);
            } else {
                  token.setValueFor(this.themeProvider, value);                
            }
        }
      }
  ```

I think that simplifies things:
- Can just focus on the label provider attribute values and don't need to deal with calculating defaults, just care about what values override the defaults / undefined
- No issue with token values on the cascade. Label Provider doesn't need to be aware of it at all.

I think the only question may be what should happen in disconnected callback of label provider. If the mental model is that the label provider represents just the attribute values then I guess removing the label provider from a theme provider is effectively like all those attributes are removed. So I guess on disconnect it should do the equivalent of those attrs being set to undefined, i.e. deleteValueFor all the attrs on the theme provider in the label provider disconnected callback?

packages/nimble-components/src/label-provider/core/label-tokens.ts
msmithNI: Updated

packages/nimble-components/src/label-provider/base/index.ts
msmithNI: Yeah I think that makes sense, and it definitely simplifies the code.

Updated with that approach, which still works with the previous tests too.

packages/nimble-components/src/label-provider/base/index.ts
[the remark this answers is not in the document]
msmithNI: Previous questions are N/A with the latest updates. Since we're only tracking the attribute state now I also switched to the more standard pattern of Observable.getNotifier+subscribe in the connectedCallback, and unsubscribe in the disconnectedCallback.

packages/nimble-components/src/label-provider/base/index.ts
```diff
@@ -0,0 +1,70 @@
...
+import { Notifier, Observable, type Subscriber } from '@microsoft/fast-element';
+import { themeProviderTag } from '../../theme-provider';
+
+/**
+ * Base class for label providers
+ */
+export abstract class LabelProviderBase
+    extends FoundationElement
+    implements Subscriber {
+    protected abstract readonly supportedLabels: {
```
rajsite, line 11: [Played with it a bit](https://www.typescriptlang.org/play?#code/C4TwDgpgBAIhDOBLA5gOwCoHsDWFXwDFMAnAHgHkAjAKwgGNh0A+KAXigG8AoKXqAbVwgoiVFCGYAZlCq0G6ALoAuKPGDFRyANxcAvly4BDSmuKGGUOgBtD8eFAAyxiFYAKxTADdEAEwjEAIVsIUgBlAFcwMBJgCB8nShd7CAAPWNQfew5BCBAVU01lKFRwgFtE4igAegAqSwALQ1RkaGBMKD8kNDbcVBqq3RZuPihjU3NgKFKQBKSVCKiYuNmreB19LjpMfEnplft2Yb4bRKsARhUAVgAaHmPnKwAmFQBmPVH7LZ2dTZs7KAAsiAAMJ-ZJpPCZRwPdxeXz+ILwEKgSBSKYzB7wFiIUpgKwQUp4YD2OBdDA4PCEEikFEQNF7TFDO68YgQQw+bZWYQM04HdH7HQjE4uM5sKAAcnqLismHFgvup0eYvFAHcSFYfHKDPogA) and I think that can make the types a bit stronger by passing the supported labels as a generic but if not it's fine since it's internal stuff.

rajsite (approved):

packages/nimble-components/src/label-provider/base/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        [P in keyof LabelProviderBase]?: DesignToken<string>;
+    };
+
+    private propertyNotifier?: Notifier;
```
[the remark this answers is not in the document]
msmithNI: Updated

packages/nimble-components/src/label-provider/base/index.ts
msmithNI: Updated

PR #1329, opened 2023-06-27
Update docs on when to use the data$ Observable on the Angular nimble-table

### LINKED ISSUE #1320: Update guidance for `data$` input on the `NimbleTableDirective`
## 🧹 Tech Debt
When trying to use the `nimble-table` in an Angular application, we discovered that in practice, the `data$` input to specify an `Observable` of data should only be used in certain cases. When using it, a client application has no way of knowing when their data has actually been applied to the table, so in cases where that knowledge is important, `setData()` should be used rather than `data$`.

We should have clear guidance documenting the recommendations of using `setData()` vs `data$`.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1320

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-angular-1fa32c6c-a6b6-424d-a969-eccb45a9b0f1.json
+++ b/change/@ni-nimble-angular-1fa32c6c-a6b6-424d-a969-eccb45a9b0f1.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Update docs on when to use the data$ Observable on the Angular nimble-table",
+  "packageName": "@ni/nimble-angular",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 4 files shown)


change/@ni-nimble-angular-1fa32c6c-a6b6-424d-a969-eccb45a9b0f1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
rajsite, line 2: Changes product code so I think should still be a patch. Even a comment could have unexpected side-effects from funky build tooling, etc.

```suggestion
  "type": "patch",
```

rajsite (approved):

jattasNI (approved):

PR #1330, opened 2023-06-28
Concrete properties in text-column view base classes

rajsite:
# Pull Request

## 🤨 Rationale

Changes the text-column cell view and group header view base classes to have concrete observable properties:

  - Is a more "normal" element pattern. Just standard observable properties.
  - Encourages caching results in response to changes instead of calculating new values on accessor request.
    - Cell View elements are configured by the columnConfig and cellRecord/groupHeaderValue. So we can avoid extra work by calculating results only when those change.

## 👩‍💻 Implementation

Changed base class to use observable properties.

## 🧪 Testing

Rely on existing texts.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. We do keep finding that abstract properties in base class lead to more confusion, not less and end up refactoring away (see columnInternals changes). Not sure if there is general guidance yet.

### DIFF
--- a/change/@ni-nimble-components-a91e2528-b7f1-448a-a97a-409d872e2988.json
+++ b/change/@ni-nimble-components-a91e2528-b7f1-448a-a97a-409d872e2988.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Change text-column base class property behavior",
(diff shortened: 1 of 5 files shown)


jattasNI (approved):
Nice simplification!

Issue #1298, opened 2023-06-13
Branded / green color appearance for bit spinner


jattasNI:
## 😯 Problem to Solve

At least one client (Licensing Blazor app)  and possibly others is requesting a mode of the spinner which renders the bits in a branded green color when using the light theme. Their interaction designer (Nakano) spoke with Brandon and he agreed that this makes sense for certain use cases like branded splash screens. 

We don't yet have a clear deadline of when this is needed, but it seems like a simple enhancement.

![image](https://github.com/ni/nimble/assets/10500124/84db3e2d-3c21-4de3-a95d-cc6e5dbfa0d4)

## 💁 Proposed Solution

I don't think we want to allow clients to configure arbitrary colors. Instead we'd likely add a configuration using an existing Nimble concept like `appearance='branded'`.

We'll need to support all themes and provide documentation about when to use default vs branded appearance. Nakano will ask Brandon to add this to the Figma library.

DStavilaNI:
Hello Guys - we are also interested in this, for a separate project (Ozone). Would it be possible to give us an estimate of when this might be completed ?

rajsite:
@DStavilaNI do you know when you would need that ability? The issue describes that the first app that we know needs it is Q2 2024. Helps prioritize if we know when you need it.

jattasNI:
We had considered making this be the next project for our summer intern @aidendk. Hearing about more clients for it makes me think we should do this! (No guarantees how far he'll get, but I'm optimistic).

jattasNI:
Mini HLD:
 - configure using `appearance='accent'` like the button (no appearance set means use the existing grey with value `undefined`)

PR #1338, opened 2023-07-06
Spinner Accent Appearance

aidendk:
# Pull Request

## 🤨 Rationale

See #1298

## 👩‍💻 Implementation

Implements the `accent` appearance variation of the nimble-spinner in Storybook, Blazor, and Angular, and includes spinner docs that more closely reflect Brandon's docs in Figma.

## 🧪 Testing

Tested that appearance would not compile with an arbitrary string value in types.spec.ts of the spinner/tests directory.
Appearance testing was also added to the Angular and Blazor spinner components,

Spinner matrix tests were updated to include the `accent` appearance.

## ✅ Checklist


aidendk:
Let me know what you guys think of the new docs. I decided to try splitting up the major sections like Brandon does in Figma, and included attribute specific docs within the `Usage details` of each variable. I reduced the length of a few descriptions, including descriptions for the sizing.

packages/nimble-components/src/spinner/tests/spinner.stories.ts
```diff
@@ -22,11 +23,28 @@ const spinnerSize = {
 
 interface SpinnerArgs {
     size: keyof typeof spinnerSize;
+    appearance: keyof typeof SpinnerAppearance;
 }
 
 const overviewText = '<p>The `nimble-spinner` is an animating indicator that can be placed in a particular region of a page to represent '
```
jattasNI, line 29: Instead of putting this docs content in the `.stories.ts` file, let's keep the story limited to just the component demo and API docs. Instead you can add a `.mdx` file for the docs. That's a markdown format so it'll be easier to maintain text content than in HTML. And it allows embedding the interactive demo and API docs anywhere within the file. (This is a new pattern we've started adopting recently but eventually I hope all of our components will use this approach).

You can see a great example of this in [Malcolm's label-provider PR](https://github.com/ni/nimble/pull/1328/files#diff-b240e51b1ed36cc5b8c6f4bcd4fd6b1d22dd5006bc1c86618a132215cc2b2c3c) and some simpler examples in other .mdx files already in the repo.

Once you do this I'd suggest a document structure something like:
1. intro paragraph (maybe lines 29-31 of your current docs?)
2. interactive demo and API table
3. Sections for sizing / appearance / accessibility

packages/nimble-components/src/spinner/tests/spinner.stories.ts
```diff
@@ -22,11 +23,28 @@ const spinnerSize = {
...
+    + '<p>It has 3 sizes (64px, 32px, and 16px) and 2 appearance types (default and accent).</p>'
+    + '<h4>Sizing</h4>'
+    + '<p>The 64px spinner is a less commonly used size reserved for large negative empty space. Only one 64px spinner should be used on a single screen at one time.</p>'
+    + '<p>The 32px spinner is a more commonly used size for negative empty space. No more than three 32px spinners should be used on a single screen at one time.</p>'
+    + '<p>The 16px spinner is the most compact option for tight spaces such as grid cells and dense UI interfaces without much negative empty space. '
+    + 'Multiple 16px spinners can be used on a single screen at one time.</p>'
+    + '<h4>Styling / Theme</h4>'
+    + '<p>Light UI controls are made of  dark attributes and designed to work against bright, light backgrounds. </p>'
+    + '<p>Dark UI controls are made of light attributes and designed to work against bold, dark backgrounds. </p>'
+    + '<p>Color UI controls are made of monochromatic attributes and designed to work against bold, medium-dark and dark colored backgrounds.</p>'
```
jattasNI, line 40: I'd vote to remove these lines from this page since they are general info about theming rather than being specific to the spinner. Ideally we'd have another docs page that covers them (possibly a page for the `nimble-theme-provider` component or a general section about the concept of theming) but that's probably out of scope for this PR.

packages/nimble-components/src/spinner/tests/spinner.stories.ts
```diff
@@ -22,11 +23,28 @@ const spinnerSize = {
...
+    + '<h4>-</h4>'
+    + '<p>Use the `default` appearance Color UI version for backgrounds with color (e.g. purple, blue). </p>'
+    + '<p>Do not use the `accent` appearance on any colored backgrounds, instead use the `default` Color UI version.</p>'
+    + '<p>See the `size` and `appearance` usage details for information on customizing the spinner size and guidance for which appearance to use.</p>'
+    + '<h4>Accessibility</h4>'
+    + '<p>A low-motion/low-frame rate version of the spinner is also available.</p>'
```
jattasNI, line 46: You could add a note that this is used when [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) is enabled.

packages/nimble-components/src/spinner/tests/spinner.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+Do not use the `accent` appearance on any colored backgrounds, instead use the `default` Color UI version.
+
+See the `size` and `appearance` usage details for information on customizing the spinner size and guidance for which appearance to use.
+
+## Accessibility
+
+A low-motion/low-frame rate version of the spinner is also available. It can be activated by enabling `prefers-reduced-motion`.
+
+When using colored backgrounds, confirm with your designer that a 3:1 contrast ration is met.
```
jattasNI, line 45:
```suggestion
When using colored backgrounds, confirm with your designer that a 3:1 contrast ratio is met.
```

packages/nimble-components/src/spinner/tests/spinner.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+The 16px spinner is the most compact option for tight spaces such as grid cells and dense UI interfaces without much negative empty space.
+
+Multiple 16px spinners can be used on a single screen at one time.
+
+## Styling / Theme
+
+Use the `default` appearance Color UI version for backgrounds with color (e.g. purple, blue).
+
+Do not use the `accent` appearance on any colored backgrounds, instead use the `default` Color UI version.
```
jattasNI, line 37: I think these two lines are essentially saying the same thing so could be collapsed to one paragraph.:
```suggestion
Use the `default` appearance in the Color UI theme for backgrounds with color (e.g. purple, blue). Do not use the `accent` appearance on any colored backgrounds.
```

packages/nimble-components/src/spinner/index.ts
```diff
@@ -12,7 +14,15 @@ declare global {
  * A Nimble-styled spinner component.
  * A spinner is an animating indicator that can be placed in a particular region of a page to represent loading progress, or an ongoing operation, of an indeterminate / unknown duration.
  */
-export class Spinner extends FoundationElement {}
+export class Spinner extends FoundationElement {
+    /**
+     * @public
+     * @description
+     * Type of spinner
```
jattasNI, line 21: nit: not much better, but it matches the docs from other components' appearance property
```suggestion
     * The appearance the spinner area should have.
```

change/@ni-nimble-components-c20d1978-2348-4144-866b-63c75566bf94.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: It doesn't have to be in this PR, but we'll need to add Angular and Blazor integration to expose the new `appearance` attribute.

jattasNI (approved):
Just wanna say: I love the new docs!!

aidendk:
My newest commit integrates the new spinner appearance with Blazor and Angular. Take a look at the component demos for each site and let me know what you think of how I displayed the `Default` and `Accent` spinner appearances.

change/@ni-nimble-components-c20d1978-2348-4144-866b-63c75566bf94.json
aidendk: Just added integration for both in my most recent commit 😄

packages/nimble-components/src/spinner/styles.ts
```diff
@@ -144,27 +148,71 @@ export const styles = css`
...
-        `
-    ),
-    themeBehavior(
-        Theme.dark,
-        css`
-            :host {
-                --ni-private-spinner-bits-background-color: ${Black15};
+            div.container {
+                opacity: 0.6;
             }
```
rajsite, line 156: Two things:
1. The default styles should be the default styles set on the control when no attributes are set. You don't need to use an appearanceBehavior for the default. In the top `div.container {}` you can set the default opacity style.
2. Check out the [responding to attribute appearance](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#responding-to-attribute-values) guidelines. The key thing about behaviors is that thye should be used when "a large block of styles is overridden based on the attribute configuration". In this case we only are changing a single style so this approach seems a bit much.

I think you could organize a bit more like this:
```ts
export const styles = css`
    div.container { opacity: 0.6; }
    :host([appearance="accent"]) div.container { opacity: 1; }
`.withBehaviors(
    themeBehavior(Theme.light, css`
        :host {
            --ni-private-spinner-bits-background-color: ${Black91};
        }
        :host([appearance="accent"]) {
            --ni-private-spinner-bits-background-color: ${DigitalGreenLight};
        }
    `),
    ...
);
        
```

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleSpinnerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    public void NimbleSpinner_SupportsAdditionalAttributes()
+    {
+        var context = new TestContext();
+        context.JSInterop.Mode = JSRuntimeMode.Loose;
+        var exception = Record.Exception(() => context.RenderComponent<NimbleSpinner>(ComponentParameter.CreateParameter("class", "foo")));
+        Assert.Null(exception);
+    }
+
+    [Theory]
+    [InlineData(SpinnerAppearance.Default, "<nimble-spinner((?!appearance).)*>")]
```
atmgrifter00, line 35: While I appreciate the rigor here, I think it's sufficient to follow our other patterns for the default case, where we just check against the literal `<nimble-spinner>`. A default component will never have other attributes set in it, so this is safe to do. It also allows a test reader to not have to be familiar with regular expressions to know what we're testing.

jattasNI (approved):

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleSpinnerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        var context = new TestContext();
+        context.JSInterop.Mode = JSRuntimeMode.Loose;
+        var exception = Record.Exception(() => context.RenderComponent<NimbleSpinner>(ComponentParameter.CreateParameter("class", "foo")));
+        Assert.Null(exception);
+    }
+
+    [Theory]
+    [InlineData(SpinnerAppearance.Default, "<nimble-spinner>")]
+    [InlineData(SpinnerAppearance.Accent, "appearance=\"accent\"")]
+    public void SpinnerAppearance_AttributeIsSet(SpinnerAppearance value, string expectedMarkupRegEx)
```
atmgrifter00, line 37: You can now change the parameter name to `expectedMarkup`.

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleSpinnerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        var spinner = RenderWithPropertySet(x => x.Appearance, value);
+
+        Assert.Matches(expectedMarkupRegEx, spinner.Markup);
```
atmgrifter00, line 41: I think you would now use `Assert.Contains`.

rajsite (approved):

msmithNI (approved):

atmgrifter00 (approved):

PR #1336, opened 2023-07-05
Initial Angular code bring up for nimble-rich-text-viewer component

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

This PR includes the angular integration for the `nimble-rich-text-viewer` component, and it is intended to make the component available to the Angular application which allows us to get started with the other features of comments dependent on this component.

## 👩‍💻 Implementation

* Added directive and module for `nimble-rich-text-viewer`. 
* Added `nimble-rich-text-viewer` to the example client app.

**Note**: We will add the input properties and attributes once the support for the markdown parser and other functionality in the component ( https://github.com/ni/nimble/pull/1335) is completed.

## 🧪 Testing

Added basic unit tests for the directive. 

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1252, opened 2023-05-17
Export table page object from nimble-angular

mollykreis:
# Pull Request

## 🤨 Rationale

The primary driver of this change is being able to access the `TablePageObject` from tests within an Angular client. However, we decided that now is a good time to add secondary entry points for table related code. Creating secondary entry points for everything within nimble-angular is covered by #172.

## 👩‍💻 Implementation

nimble-angular now has the following entry points:
- `@ni/nimble-angular` - this is the existing entry point and contains all the non-table components
- `@ni/nimble-angular/internal-utilities` - contains template-helper code; creating an additional entry point was the best way I could find to share code between other entry points
- `@ni/nimble-angular/table` - contains nimble-table
- `@ni/nimble-angular/table/testing` - exports the table's page object
- `@ni/nimble-angular/table-column` - contains the base class for column directives
- `@ni/nimble-angular/table-column/anchor` - contains nimble-table-column-anchor
- `@ni/nimble-angular/table-column/text` - contains nimble-table-column-text

## 🧪 Testing

- Built systemlink-lib-angular with these changes to verify that it can still import everything it needs, including the nimble-table and nimble-table-column-text
- Used a locally build npm package to write some tests using the page object in systemlink-lib-angular.

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

Issue #172, opened 2021-10-14
Organize repo to consistently have an import per component


rajsite:
Currently nimble-components and nimble-angular organize each component into a separate directory with separate import strategies. 

Edit August 2023:

The current recommendation would be as follows (these are not coupled, we could do the smaller nimble-components work first):
- nimble-components:
  - Switch to using [npm package entry points](https://nodejs.org/api/packages.html#packages_package_entry_points)
  - Leveraging package entrypoints, elide the `dist/esm` from import paths so for instance a button import changes from `import '@ni/nimble-components/dist/esm/button';` to `import '@ni/nimble-components/button';`
  - Should decide how we migrate to es module resolution within the package. i.e. showing the full extension in the import path.
  - Should decide what is private and what is public apis in nimble-components when using package exports.
     - Could potentially stage the private/public enforcement before the rename to remove `dist/esm`. This could help avoid bugs like we fixed in #2290 
- nimble-angular:
  - Switch to secondary entrypoints per directive. **Note:** we should profile build performance before going down this road. We might also want to be on a nearly-new version of Angular to get modern performance numbers.
  - Switch to Angular single file / standalone component pattern for our directives, pipes, etc.
  - Leveraging secondary entrypoints and single file components an angular app can avoid needing to import a Module per component and would import a button directive from `@ni/nimble-angular/button`
- Go through the comments in the thread and make sure they are addressed

All libraries:
  - When we get to make the big breaking path changes we should update the icon paths from the plural `icons/add` to i.e. just `icon/add`. This reflects our other path conventions to make the path align with the icon name. (Realizing years later I missed that 🤦)

Known issues this change will resolve:
- In nimble-angular today the vast majority of components are exported together from the top-level primary entrypoint. Any import at that level results in all components from that entry point being registered and defined in a page: https://github.com/ni/nimble/issues/1476

rajsite:
> As part of this work we should control our exports. This helps two issues, one is the Angular warning about deep entry points:
> 
> The other is we can prevent implementation details from being exposed externally, like stuff in src/patterns styles.

In addition, we should evaluate our exported .d.ts files to make sure internal libraries are not exposed and move `@types` to devDependencies where possible. See https://github.com/ni/nimble/pull/831/files#r1029565900

rajsite:
Related issue suggesting we should use package exports as described above. That is technically independent of re-organizing the repo so not closing as a duplicate. It's essentially a subtask of this issue https://github.com/ni/nimble/issues/694

[...]

PR #1340, opened 2023-07-07
Create Angular barrel files for all components

### LINKED ISSUE #433: Refactor icon Angular code generation to create barrel file instead of rewriting public-api.ts
## 🧹 Tech Debt

In #377, I created a build script which generates Angular integration for Nimble icons. Since we were unable to use barrel files in Nimble Angular at that time, I had to generate a new `public-api.ts` file which included all of the generated icon files. 

Now that we have switched to Ivy, we should get rid of the `public-api.ts` generation and generate a barrel file for the icons instead. But we should probably wait a bit until we're confident in the stability of all of the Angular projects on Ivy.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #433 

## 👩‍💻 Implementation

There is already a barrel file for the generated icons. This change adds barrel files (`index.ts`) for each of our other components (except for table and table-column). `public-api.ts` is updated to reference those barrel files instead of each individual file for the component.

## 🧪 Testing

Built and ran tests.

## ✅ Checklist


change/@ni-nimble-angular-2e46c39b-28a8-4aef-9b90-ecc3ef77f46e.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Some questions for my own understanding:
1. How does this impact clients? Does it change how we'd recommend they do imports? Or change what can be tree shaken?
2. If it does change anything:
   1. are there any docs or examples that need updating?
   2. are there any clients we should update ourselves or broadcasting we should do to get clients to change their code?
3. How does this relate to our vision of having separate secondary entry points for each component? (Issue #172 with some initial implementation in #1252 and #1336). Does this make that easier? Harder? Unrelated?

rajsite: 1. No impact, what is exported from public-api.ts doesn't change. Doesn't help or hurt tree shaking. Tree-shaking is broken until we have secondary entry points.
2. N/A
3. Doesn't really help or hurt the future goal of secondary entry points. It's only for dev purposes.
   
   I'm on the fence of whether this change actually improves dev experience as-is. We have to deal with some complexities to avoid exporting symbols twice which is captured in the top-level public-api.ts. I think at worst this change could obfuscate those considerations and make them harder to understand / debug and at best is neutral for dev efficiency.

I'd probably advocate for not making this change as I don't think it gives a significant benefit for clients or nimble devs. Instead we should prefer secondary entry points as that will have tangible benefits for both in the future.

change/@ni-nimble-angular-2e46c39b-28a8-4aef-9b90-ecc3ef77f46e.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: Patch change as there should be no observable change in behavior

rajsite (approved):

change/@ni-nimble-angular-2e46c39b-28a8-4aef-9b90-ecc3ef77f46e.json
m-akinc: I am not personally invested in this change, and I think it's marginally better to avoid any risk, so I'm going to abandon.

PR #1342, opened 2023-07-11
Update markdown value accessor name in rich-text-editor spec

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Discussion on changing the name from `markdownValue` to `markdown`: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/517374#1688590052
 
## 👩‍💻 Implementation

Updated the `nimble-rich-text-editor` spec to change the markdown value accessor name from `markdownValue` to `markdown`.

## 🧪 Testing

NA

## ✅ Checklist


vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -102,7 +102,7 @@ Example usage of the `nimble-rich-text-editor` in the application layer is as fo
 
 _Props/Attrs_
 
--   `markdownValue` - is an accessor used to get and set the markdown value.
+-   `markdown` - is an accessor used to get and set the markdown value.
```
rajsite, line 105: This makes sense. 👍

The native input has the concept of [`value`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#value) (the raw string value), [`valueAsNumber`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#valueasnumber), and [`valueAsDate`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#valueasdate).

If we had a concept of a "raw" `value` then I can imagine us doing `valueAsMarkdown`, etc. But I don't think we will ever expose a `value` property for the "raw" value so the `valueAs*` pattern isn't really worth following. So this change seems reasonable to me 👍.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -102,7 +102,7 @@ Example usage of the `nimble-rich-text-editor` in the application layer is as fo
...
     -   `getter` - this will serialize the content by extracting the Node from the editor and convert it into a markdown string using
         [prosemirror-markdown serializer](https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/to_markdown.ts#L30).
     -   `setter` - this will parse the markdown string into a Node and load it back into the editor using
```
rajsite, line 108: Can you capture a discussion in an `Alternatives` section about the choice between getter and setter accessors vs `getMarkdown` and `setMarkdown` methods?

I can't remember how we decided to use accessors instead of methods so if you can capture the reasoning in the spec that would be helpful.

rajsite (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: @rajsite I have added the `Alternatives` section on how we choose `markdown` as an accessor instead of methods.

rajsite: Huh. That justification does not seem very convincing.. Just to be clear, setting the property on the viewer is NOT a one time operation right? The value can be set multiple times whenever the app needs to update the content. I think that statement is misleading and should be clarified.

The best trade-off I can think of is that the property syntax allows for one-way data binding which is easier for clients. However because calculating the markdown representation can be expensive an accessor was used instead of a property that is written to on changes for the editor component. The documentation will need to include discussion on how clients can use the component performantly and avoid reading the accessor unnecessarily.

vivinkrishna-ni: Apologies for the confusion caused. The one-time operation I meant was that the `viewer` component does not get the processed `markdown` value for every change like the `editor` component, instead, it returns the same markdown value that it receives from the client. However, it is misleading and not a convincing justification for this case. So I have removed the statement and updated it as you suggested.

rajsite: looks great! ty!

jattasNI (approved):

jattasNI:
@vivinkrishna-ni No need to wait on the remaining Nimble reviewers for this relatively minor change. I approve "bypassing" and completing the PR.

Issue #1325, opened 2023-06-26
Nimble combobox doesn't scroll down to the selected option


ipopa144:
# 🐛 Bug Report

When using the nimble combobox component with `autocomplete="inline"` option, if the dropdown doesn't fit on the screen, when typing a value, the option is selected, but it doesn't scroll down such that the user can see that the option is selected.

## 💻 Repro or Code Sample

```
<nimble-combobox
    formControlName="..."
    placeholder="Select or enter workspace"
    autocomplete="inline"
    error-text="text"
    [error-visible]="!valid">
...
</nimble-combobox>
```

## 🤔 Expected Behavior

The component should scroll down to the selected value if it is not visible on the screen.

## 😯 Current Behavior

![workspace-selector-bug](https://github.com/ni/nimble/assets/44372517/853b312c-783c-4a0d-b6cd-2d05e6b830ad)

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

atmgrifter00:
This is, indeed, an issue in FAST. While I think we can go ahead with a solution in Nimble, I've still created an [issue in FAST](https://github.com/microsoft/fast/issues/6776).

PR #1343, opened 2023-07-11
Fix Combobox scroll issue.

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1325

## 👩‍💻 Implementation

Simply calling the provided FAST `focusAndScrollOptionIntoView` base function when the current value in the input is different than the current value while typing.

## 🧪 Testing

Added a unit test.

## ✅ Checklist


packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -531,6 +552,30 @@ describe('Combobox', () => {
...
+        await connect();
+        await waitForUpdatesAsync();
+
+        const lastOption = element.options[element.options.length - 1]!;
+        await clickAndWaitForOpen(element);
+        updateComboboxWithText(element, '1');
+        await waitForUpdatesAsync();
+        // This second call seems necessary to allow the requestAnimationFrame in the FAST Combobox implementation
+        // time to run for the first update.
+        await waitForUpdatesAsync();
```
atmgrifter00, line 570: I wasn't able to come up with a testing solution that didn't involve calling `waitForUpdatesAsync` twice in a row after the first update.

rajsite, line 570: At this point (before updating the text to `1000`) should we asset that the option is not visible?

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -26,6 +26,27 @@ async function setup(
     return fixture<Combobox>(viewTemplate);
 }
 
+async function setupWithManyOptions(
+    position?: string,
+    open?: boolean,
+    autocomplete?: ComboboxAutocomplete
```
rajsite, line 32: If these options aren't needed I'd remove them to simplify the fixture for this test. Makes the test setup easier to understand.

rajsite (approved):

packages/nimble-components/src/combobox/tests/combobox.spec.ts
atmgrifter00: Hmmm...not sure how I feel about that...that's basically just an assert to make sure the setup code is proper, as opposed to something related to the issue itself. If you would still like to have the assert there I'm not that against it, I just don't feel strongly about adding it.

rajsite: I'd like the assertion just because I'm less confident about tests initial conditions because we do not control the viewport. It's possible for some reason that the browser window is very long and the line is already in the viewport. We don't control that initial condition / test set-up isn't controlling for that. So a clear assertion before and after about the viewport state makes me more confident this is testing a specific change in behavior.

rajsite: > that's basically just an assert to make sure the setup code is proper,

yep! exactly lol

atmgrifter00: Done.

jattasNI (approved):

PR #1348, opened 2023-07-14
Add rowRecordId to TableCell, TableCellView, and delegated event details

m-akinc:
# Pull Request

## 🤨 Rationale

This is a precursor to adding Angular router support to the anchor column. We want to be able to pass a clicked link's row id as part of the event details of the delegated click event.

## 👩‍💻 Implementation

The delegated event is fired from the TableCellView, so it needs access to the row id. To get it there, we also need to add it to the TableCell. The row id gets propagated through the templates.

## 🧪 Testing

Tested in prototype of Angular router support for anchor column.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-30319fbf-4709-43b2-9ab5-378704cc1cff.json
+++ b/change/@ni-nimble-components-30319fbf-4709-43b2-9ab5-378704cc1cff.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add rowRecordId to TableCell, TableCellView, and delegated event details",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table-column/base/cell-view/index.ts
+++ b/packages/nimble-components/src/table-column/base/cell-view/index.ts
@@ -26,6 +26,9 @@ export abstract class TableCellView<
     @observable
     public column?: TableColumn<TColumnConfig>;
 
+    @observable
+    public rowRecordId?: string;
+
     private delegatedEvents: readonly string[] = [];
 
     /**
@@ -49,7 +52,10 @@ export abstract class TableCellView<
         this.delegatedEventHandler = (event: Event) => {
             this.column?.dispatchEvent(
                 new CustomEvent<DelegatedEventEventDetails>('delegated-event', {
-                    detail: { originalEvent: event }
+                    detail: {
+                        originalEvent: event,
+                        rowRecordId: this.rowRecordId
+                    }
                 })
             );
         };
--- a/packages/nimble-components/src/table-column/base/types.ts
+++ b/packages/nimble-components/src/table-column/base/types.ts
@@ -24,6 +24,7 @@ export interface TableColumnWithPlaceholderColumnConfig {
  */
 export interface DelegatedEventEventDetails {
     originalEvent: Event;
+    rowRecordId?: string;
 }
 
 /**
--- a/packages/nimble-components/src/table/components/row/template.ts
+++ b/packages/nimble-components/src/table/components/row/template.ts
@@ -30,6 +30,7 @@ export const template = html<TableRow>`
                         :cellState="${x => x.cellState}"
                         :cellViewTemplate="${x => x.column.columnInternals.cellViewTemplate}"
                         :column="${x => x.column}"
+                        :rowRecordId="${(_x, c) => c.parent.recordId}"
                         ?has-action-menu="${x => !!x.column.actionMenuSlot}"
                         action-menu-label="${x => x.column.actionMenuLabel}"
                         @cell-action-menu-beforetoggle="${(x, c) => c.parent.onCellActionMenuBeforeToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>, x.column)}"
--- a/packages/nimble-components/src/table-column/base/cell-view/template.ts
+++ b/packages/nimble-components/src/table-column/base/cell-view/template.ts
@@ -25,6 +25,7 @@ export const createCellViewTemplate = (
             :cellRecord="${y => y.cellState?.cellRecord}"
             :columnConfig="${y => y.cellState?.columnConfig}"
             :column="${y => y.column}"
+            :rowRecordId="${y => y.rowRecordId}"
             class="cell-view"
         >
         </${cellViewTag}>
--- a/packages/nimble-components/src/table/components/cell/index.ts
+++ b/packages/nimble-components/src/table/components/cell/index.ts
@@ -29,6 +29,9 @@ export class TableCell<
     @observable
     public column?: TableColumn;
 
+    @observable
+    public rowRecordId?: string;
+
     @attr({ attribute: 'has-action-menu', mode: 'boolean' })
     public hasActionMenu = false;
 


change/@ni-nimble-components-30319fbf-4709-43b2-9ab5-378704cc1cff.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I love small PRs but a couple things become more important when slicing up work like this:
1. providing more big-picture context about how this will be used. Do we have a comment / HLD / draft PR explaining what the current proposal is to fix the router bug? How can I gain confidence that these APIs are needed?
2. providing more detail about the testing plan. There aren't any auto tests of these new APIs in this PR. You mentioned that you tested them in another branch but are those auto tests or one-time-prove-it-works manual tests? If auto tests, why are we relying on integration tests in nimble-angular rather than unit tests in nimble-components? There could be good reasons, but the context helps understand those reasons.

m-akinc: 1. There's no HLD or such document, but I do have a draft PR for the follow-on bug fix: #1349. This approach was discussed over a couple of conversations between Molly, Milan, and myself. Before starting down this path, I got consensus from them that this was how we thought it should be addressed.
2. I did one-time manual testing in my prototype build of TestInsights, but I also have a couple automated tests in my next PR (in `nimble-angular`). One of those tests asserts that the row id gets passed to the callback function, thereby validating the changes in this PR. I didn't think it necessary to test that the `rowRecordId` properties I added get populated correctly, because those are essentially implementation details. As far as validating that the row id ends up in the delegated event details, we do have a test in `table-cell-view.spec` that verifies the behavior of delegated events, but it doesn't do it in the context of a table with row data (so there are no row ids to test). I figured it was already going to be tested by my `nimble-angular` test, so I didn't bother to write a new test in `nimble-components`.

rajsite: >  I figured it was already going to be tested by my nimble-angular test, so I didn't bother to write a new test in nimble-components.

This is a feature of the delegated events implementation so it should have it's own test coverage.

jattasNI (approved):

packages/nimble-components/src/table-column/base/types.ts
```diff
@@ -24,6 +24,7 @@ export interface TableColumnWithPlaceholderColumnConfig {
  */
 export interface DelegatedEventEventDetails {
     originalEvent: Event;
+    rowRecordId?: string;
```
rajsite, line 27: What is the scenario this is undefined for users?
A rendered row always has row ids right? Either an index created by the table or the row ids defined by the user? @mollykreis 

Seems like this should not be nullable. Maybe we should not fire an event in the scenario that somehow rowId is undefined?

packages/nimble-components/src/table-column/base/tests/table-cell-view.spec.ts
```diff
@@ -108,4 +108,24 @@ describe('TableCellView', () => {
...
+
+        let eventDetails: DelegatedEventEventDetails | undefined;
+
+        const column = document.createElement(
+            tableColumnDelegatesClickAndKeydownTag
+        ) as TableColumnDelegatesClickAndKeydown;
+        column.addEventListener('delegated-event', event => {
+            eventDetails = (event as CustomEvent)
+                .detail as DelegatedEventEventDetails;
+        });
```
rajsite, line 123: Prefer to use jasmine spies for tracking events

packages/nimble-components/src/table-column/base/types.ts
m-akinc: If `rowRecordId` should always be defined, then maybe we should throw an error if we find an undefined value. That seems better than silently skipping the event.

rajsite: Brainstormed with Jesse a bit and I think we have a direction to go:

- Looking at the cell-view as a standalone component, it is completely valid for the component to have a recordId that is undefined. From a unit testing perspective we should have tests for the recordId defined and undefined showing it's behavior (fires and doesn't fire an event respectively).
-  We should have integration level jasmine tests for this feature testing the different scenarios for recordId, i.e. the data including recordId and the data not including recordId (and falling back to table generated record ids).
- With the above then we don't throw an exception here, each component in isolation and in integration is behaving as defined and tested.

> we should throw an error if we find an undefined value

This isn't hard and fast rule we have defined. For example all the validity infrastructure on tables avoids throwing exceptions as there can be a settling phase for property setting. Also on the next line if column is undefined the code is silently ignoring firing the error.

rajsite (approved):

packages/nimble-components/src/table-column/base/cell-view/index.ts
```diff
@@ -47,11 +50,19 @@ export abstract class TableCellView<
...
-                })
-            );
+            if (this.rowRecordId) {
+                this.column?.dispatchEvent(
+                    new CustomEvent<DelegatedEventEventDetails>(
+                        'delegated-event',
+                        {
+                            detail: {
+                                originalEvent: event,
+                                rowRecordId: this.rowRecordId
```
mollykreis, line 60: What was the rationale in calling this `rowRecordId` rather than just `recordId`? I think we are pretty consistent throughout our code in calling this a `recordId`, so I think we should have a good justification for differing from that naming here.

rajsite: good catch!

m-akinc: Mutated from my original name `rowId`. Agree it is better as just `recordId`.

packages/nimble-components/src/table-column/base/tests/table-cell-view.spec.ts
```diff
@@ -108,4 +109,39 @@ describe('TableCellView', () => {
...
+        const spy = jasmine.createSpy();
+        column.addEventListener('delegated-event', spy);
+
+        element.column = column;
+        element.dispatchEvent(new PointerEvent('click'));
+
+        expect(spy).not.toHaveBeenCalled();
+    });
+
+    it('passes row id in delegated event details', async () => {
```
mollykreis, line 128: nit:
```suggestion
    it('passes record id in delegated event details', async () => {
```

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -30,6 +30,7 @@ export const template = html<TableRow>`
                         :cellState="${x => x.cellState}"
                         :cellViewTemplate="${x => x.column.columnInternals.cellViewTemplate}"
                         :column="${x => x.column}"
+                        :recordId="${(_x, c) => c.parent.recordId}"
```
mollykreis, line 33: I believe we generally use `_` rather than `_variableName` when something is unused -- at least that's what we use in this file.
```suggestion
                        :recordId="${(_, c) => c.parent.recordId}"
```

packages/nimble-components/src/table/tests/table-delegated-events.spec.ts
```diff
@@ -98,30 +98,36 @@ describe('Table delegated events', () => {
         expect(bubbledEvent!).toBe(clickEvent);
     });
 
-    it('delegates all specified event types', async () => {
+    it('delegates all specified event types (w/ default row ids)', async () => {
```
mollykreis, line 101: nit:
```suggestion
    it('delegates all specified event types with default record ids', async () => {
```

packages/nimble-components/src/table/tests/table-delegated-events.spec.ts
```diff
@@ -141,4 +147,27 @@ describe('Table delegated events', () => {
         pageObject.dispatchEventToCell(0, 0, mouseoverEvent);
         expect(gotDelegatedEvent).toBeFalse();
     });
+
+    it('delegates events with specified row ids', async () => {
```
mollykreis, line 151: nit:
```suggestion
    it('delegates events with specified record ids', async () => {
```

mollykreis (approved):

rajsite:
FYI mert, when using the merge suggestion feature consider using the batch option. Current approach kicked off 10 builds which isn't necessarily a problem but a bit unnecessary. I cancelled 8 of the builds:
image

Issue #1238, opened 2023-05-10
Anchor column doesn't work as expected with the Angular router


mollykreis:
# 🐛 Bug Report

The `nimble-table-column-anchor` element doesn't work correctly with the Angular router. Specifically, if you set an href in your data that works with the browser's navigation (e.g. CTRL+CLICK) then it doesn't work correctly with the Angular router and vice versa. This problem is definitely apparent when your Angular application is configured to have a base href, but it might manifest itself with other configurations as well.

My understanding of the problem is this:
- In normal circumstances, configuring a `routerLink` (or `nimbleRouterLink`) on an element in Angular uses the router link to calculate the `href` attribute to set on the element. This calculation is done by using the router.
- In the case of the anchor column, we are setting the `href` on the underlying anchor element and then the Angular directive assumes the `href` is also the valid `routerLink`. This is not a valid assumption.

## 🤔 Expected Behavior

I should be able to configure my anchor column with an href that will work with the Angular router and with CTRL+CLICK on the link (even if my application has a base href).

## 😯 Current Behavior

If my application has a base href, I have to choose if I want CTRL+CLICK to work or regular clicking (which uses the Angular router).

## 💁 Possible Solution

We could stop assuming that the href specified is a valid Angular router link and instead do something like what we do in the navigation tree -- try navigating the with router and falling back to updating the document's location.

Note: Whatever the final solution, we should ensure that we appropriately document whether the href specified in an Angular app is the router link or the href needed for navigation outside of the Angular router.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

rajsite:
> We could stop assuming that the href specified is a valid Angular router link and instead do something like what we do in the navigation tree -- try navigating the with router and falling back to updating the document's location.

[...]

PR #1349, opened 2023-07-14
Switch to NavigationGuard directive for Anchor Columns

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes: #1238 

## 👩‍💻 Implementation

We no longer directly do any router navigation. Instead, we provide the client with a hook to do their own router navigation on click.
- Removed anchor column directives that derived from `RouterLink` and `RouterLinkWithHref`, since we don't support any form of RouterLink anymore.
- Added new `NimbleTableColumnAnchorNavigationGuardDirective` which calls a client's configured click handler and conditionally supresses the link's default click handler.
- Updated example app
- Updated tests
- Updated Storybook documentation

## 🧪 Testing

Tested in prototype and unit tests pass.

## ✅ Checklist


change/@ni-nimble-angular-cae1f20e-b472-42ea-9f72-309d57cb3171.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: I was about to say this should be `major` since we're removing a public API, but the table is still in "incubating" so I think this can be `patch`

packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.stories.ts
```diff
@@ -87,13 +87,13 @@ interface AnchorColumnTableArgs extends SharedTableArgs {
...
+In an Angular app, you can configure a callback to intercept clicks so that you may invoke the router to perform the navigation instead of the default handler:
 \`\`\`
-<nimble-table-column-anchor nimbleRouterLink [queryParams]="{ id: 'foo' }" replaceUrl>
+<nimble-table-column-anchor [clickDelegate]="doRouterNavigation">
     Link
 </nimble-table-column-anchor>
 \`\`\`
-Note that these property values apply to every link in the column and cannot be configured on a per-link basis.
+The function bound to \`clickDelegate\` should be of type \`ClickDelegate\`. It receives the clicked row id and should return a boolean indicating whether the default click handler should be run. I.e. return \`false\` if you have handled the navigation. The click delegate is only called for left clicks with no modifier keys pressed.
```
jattasNI, line 96: >The click delegate is only called for left clicks with no modifier keys pressed.

Do we have a recommendation if an app wants to support modifier keys or other buttons to enable opening a link in a new tab? If so is it baked enough to document? If not, what are the barriers to coming up with a solution?

angular-workspace/projects/example-client-app/src/app/customapp/customapp.component.html
```diff
@@ -202,9 +202,6 @@
                     <nimble-icon-check title="String 1"></nimble-icon-check>
                 </nimble-table-column-text>
                 <nimble-table-column-anchor
-                    nimbleRouterLink
-                    [queryParams]="{ word: 'foo' }"
```
jattasNI, line 206: There's a text field below whose purpose was to display the query params so I think it can be deleted now.

m-akinc: Oops, good catch.

jattasNI (approved):

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { TableColumnAnchorCellView } from '@ni/nimble-components/dist/esm/table-column/anchor/cell-view';
+
+export type ClickDelegate = (rowRecordId: string | undefined) => boolean;
+
+/**
+ * Directive to allow client to intercept anchor clicks and do router navigation instead of
+ * letting the default action handle the navigation.
+ */
+@Directive({
+    selector: 'nimble-table-column-anchor[clickDelegate]'
```
rajsite, line 11: the directive should match the directive name and clean-up language around "click delegate". From the angular context they don't care about click / event delegates. They just care about being able to service routing themselves or letting the link do it's own thing.

```
    selector: 'nimble-table-column-anchor[navigationGuard]'
```

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
```diff
@@ -0,0 +1,39 @@
+import { Directive, HostListener, Input } from '@angular/core';
+import type { TableColumnAnchorCellView } from '@ni/nimble-components/dist/esm/table-column/anchor/cell-view';
+
+export type ClickDelegate = (rowRecordId: string | undefined) => boolean;
+
+/**
+ * Directive to allow client to intercept anchor clicks and do router navigation instead of
+ * letting the default action handle the navigation.
```
rajsite, line 8: Can you make it dead clear what the app should do, i.e. the behavior of returning true vs false and a short description of how this delegate is expected to be used

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+        if (delegatedEvent.type !== 'click') {
+            return;
+        }
+
+        const clickEvent = delegatedEvent as MouseEvent;
+        // Only call the clickDelegate for plain left clicks.
+        // Those are the only ones that should potentially use the router.
+        if (clickEvent.button !== 0 || clickEvent.ctrlKey || clickEvent.shiftKey || clickEvent.altKey || clickEvent.metaKey) {
+            return;
+        }
```
rajsite, line 28: guessing this logic is derived from here? https://github.com/angular/angular/blob/35a3fde5b71ef3b50282fe6f7b37ca1c92b8d2a0/packages/router/src/directives/router_link.ts#L453

If so can you put a pemallink referencing that

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+
```
rajsite, line 29: We capture the link to target in columnconfig right? Seems like this logic should be copied as well, i.e. for target being defined (and if we do, referenced with a permalink and tested)

https://github.com/angular/angular/blob/35a3fde5b71ef3b50282fe6f7b37ca1c92b8d2a0/packages/router/src/directives/router_link.ts#L457

PR #1353, opened 2023-07-14
Style unused slots to be hidden

### LINKED ISSUE #451: Visually hide unstyled or undocumented slots
## 🧹 Tech Debt

We don't style all the slots that are available in our controls from the fast-foundation templates. For example many controls use the start slots for icon placement but not the end slots.

Ideally we would not expose slots we do not use but that currently requires forking the templates and is not a maintenance burden we currently want to take on for all controls (only as a last resort). FAST is also [not currently planning](https://discord.com/channels/449251231706251264/482296787021856768/951561369034039299) on making that kind of configuration directly available.

For now we should make sure to style our unused slots with display:none and make sure our supported slots and use-cases are documented. This issue should specifically address that unused slots are marked dispaly:none.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #451 

Components with functional differences:
- anchor-tabs - Updated to hide the `start` slot
- anchor-tree-item - Updated to hide `end` slot
- breadcrumb-item - Updated to hide `start` and `end` slots
- combobox - Updated to hide `start` slot (in shared dropdown styles)
- list-option - Updated to hide `start` and `end` slots
- number-field - Updated to hide `start` slot
- select - Updated to hide `start` slot (in shared dropdown styles)
- tabs - Updated to hide `start` slot
- tree-item - Updated to hide `end` slot

## 👩‍💻 Implementation

- Updating the styling of some slots
- Standardize on selectors of `[part='start']` and `[part='end']`

## 🧪 Testing

- For each new hidden slot, manually tested that an element put in that slot isn't rendered in the component

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
+++ b/change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "hide unused slots",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor-menu-item/styles.ts
+++ b/packages/nimble-components/src/anchor-menu-item/styles.ts
@@ -76,7 +76,7 @@ export const styles = css`
         height: ${iconSize};
     }
 
-    :host(.indent-1) .start {
+    :host(.indent-1) [part='start'] {
         grid-column: 1;
     }
 
--- a/packages/nimble-components/src/breadcrumb-item/styles.ts
+++ b/packages/nimble-components/src/breadcrumb-item/styles.ts
@@ -59,12 +59,7 @@ export const styles = css`
 
     .start,
     .end {
-        display: flex;
-        align-items: center;
-    }
-
-    .start {
-        margin-inline-end: 4px;
+        display: none;
     }
 
     slot[name='separator'] {
--- a/packages/nimble-components/src/breadcrumb-item/styles.ts
+++ b/packages/nimble-components/src/breadcrumb-item/styles.ts
@@ -57,8 +57,11 @@ export const styles = css`
         text-decoration: underline;
     }
 
-    .start,
-    .end {
+    [part='end'] {
+        display: none;
+    }
+
+    [part='end'] {
         display: none;
     }
 
--- a/packages/nimble-components/src/anchor-tabs/styles.ts
+++ b/packages/nimble-components/src/anchor-tabs/styles.ts
@@ -10,6 +10,10 @@ export const styles = css`
         grid-template-rows: auto 1fr;
     }
 
+    [part='start'] {
+        display: none;
+    }
+
     .tablist {
         display: grid;
         grid-template-rows: auto auto;
--- a/packages/nimble-components/src/anchor-tree-item/styles.ts
+++ b/packages/nimble-components/src/anchor-tree-item/styles.ts
@@ -113,7 +113,7 @@ export const styles = css`
         height: ${iconSize};
     }
 
-    [part="end"] {
+    [part='end'] {
         display: none;
     }
 `;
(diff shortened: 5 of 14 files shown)


change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Could you add a note to [CSS Guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md) that describes our guidance to hide unused slots and how to target CSS to a slot?

packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -57,14 +57,12 @@ export const styles = css`
         text-decoration: underline;
     }
 
-    .start,
-    .end {
-        display: flex;
-        align-items: center;
+    [part='end'] {
```
jattasNI, line 60: Is this right? Looks like `end` is styled twice but there are no styles for `start`.

change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
jattasNI, line 2: Opinions on calling this a `major` change in case people were relying on these slots being visible? I think I'm ok with `patch` since the APIs were undocumented (incubating?) but could also see a client being annoyed if they were using it and we didn't provide any notification that it was going away.

My vote is to keep it `patch` but post in the UI WG and Web Chapter Blazor channels to give clients a heads up when this goes in.

jattasNI (approved):

change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
rajsite: +1 for patch, it's an internal api to use those slots and not a public api so not breaking 🙃.
Sure not well documented or anything. Should read our minds better 🔮.

packages/nimble-components/src/anchor-menu-item/styles.ts
```diff
@@ -76,7 +76,7 @@ export const styles = css`
         height: ${iconSize};
     }
 
-    :host(.indent-1) .start {
+    :host(.indent-1) [part='start'] {
```
rajsite, line 79: All else equal I think I prefer the class syntax after seeing the diff 🤐

We haven't really leveraged Shadow Parts all that much and not really sure how much we will since our goal is to lean more towards opinionated instead of flexible and exposing parts of our controls to restyle externally. Pretty much all the [existing usages](https://dev.azure.com/ni/_search?action=contents&type=code&text=%22%3A%3Apart%22%20ext%3Ascss) seem questionable too. So I don't think I want to over encourage the use, just keep it as a semi-quiet escape hatch.

At the same time it's internal so if it doesn't feel worth the effort to switch the selectors over that's fine. It does look like every selector targetting `[part=` except one (action part of input) has a corresponding class of the same name. If we go the class route we could clean that up and just remove all the `[part=` selectors from our style selectors.

packages/nimble-components/src/breadcrumb-item/styles.ts
mollykreis: No, it's not right. I messed that up when I was cleaning up the code before posting my PR 🤦

change/@ni-nimble-components-f58de85c-8a74-4a0b-b1c4-fdab36bfa94f.json
mollykreis: I went back and forth on this myself, and decided on patch since it should really only help applications better align with nimble styling (not actually break an application or build). Whether or not that is actually true, I'm not sure. I'll leave this as patch since that's the way we are all leaning.

mollykreis: I'll leave this conversation open as a reminder to post in the UI WG and Web Chapter Blazor channels when this change goes in.

packages/nimble-components/src/anchor-menu-item/styles.ts
mollykreis: I don't mind switching to use the class syntax. The thing that made me a bit hesitant to do that was that the classes only get applied when something is in the slot. I hadn't realized that until I started working on this PR, but I wasn't sure if that would break any styling assumptions.

PR #1355, opened 2023-07-17
Button Appearance Variations

### LINKED ISSUE #663: Adopt nimble-button visual designs for states primary, quiet, and loud
## 😯 Problem to Solve

The Visual Design spec for buttons now includes primary, quiet, and loud states that aren't implemented in the component yet. 

image
This image is from the [XD spec](https://xd.adobe.com/view/33ffad4a-eb2c-4241-b8c5-ebfff1faf6f6-66ac/screen/ef4a806c-a87d-4360-b952-4068d25957a9/) but the Figma spec is more up to date so we should implement it from there.

## 💁 Proposed Solution

Add states to the nimble-component. 

Document in storybook (including guidance about when to use the states). 
 - [Usage gudance is available in Figma](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1604%3A74603&t=ZRkd1pVkbBNgvlIz-1)

Expose in framework wrappers if needed.

## 📋 Tasks
- [ ] This is a subtask of the feature. (It can be converted to an issue.)
-->

aidendk:
# Pull Request

## 🤨 Rationale

Fixes #663 

## 👩‍💻 Implementation

The css classes of the button component and toggle-button component are implemented with cascade layers- using these helps decrease css variable overriding that was occurring due to appearance variants. There are five cascade layer values, `default`, `hover`, `focusVisble`, `active`, and `disabled`. The order listed is the hierarchical order these values are in, with `default` being the lowest and `disabled` being the highest. The classes are organized in this same hierarchical order within the `styles.ts` file, and classes that correspond to each cascade layer are grouped inside of that layer (ex. `:hover` classes will go in the `hover` cascade layer wrapper) in the class order described in our [css guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order).

Currently, the appearance variants change styling through conditional css classes that respond to specific appearance-variants (ex. `:host([appearance-variant='primary'])`)

Two new color tokens were added for this implementation, DigitalGreenDark105 and DigitalGreenDark110.

## 🧪 Testing

Testing that had already been created for a previous implementation of button appearance variants has been updated to work with the new appearance variants.

## ✅ Checklist


packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -21,16 +21,80 @@ import * as buttonStories from './button.stories';
...
+        Use in combination with ghost buttons to create hierarchy.
+
+        When using an outline button, do not use it in combination of a block button to create hierarchy.
+
+        When using a block button, do not use it in combination of a outline button to create hierarchy.
+
+    ### Accent Buttons:
+
+    Use only 1 accent button in a section. It should be used when trying to achieve the most prominent eye catching approach. Consider the contextual implications
+    of using a green colored button in relation to it’s name or metaphor. Remember that color has accessibility constraints and the color alone should not be relied on.
```
m-akinc, line 71: it's -> its

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -21,16 +21,80 @@ import * as buttonStories from './button.stories';
...
+        Use in combination with a primary outline and primary block buttons to create a hierarchy of importance. There is no primary ghost button.
+
+    ### Outline Button:
+
+        Outline is the secondary style and should be considered for use when ghost button is not sufficient.
+
+        Use as an alternative standard button when a ghost button is not suitable. Use similarly to a ghost button to create a clean, light and airy feel.
+        The outline button is more visually direct about the control’s functionality than a ghost button.
+
+        Use in combination with ghost buttons to create hierarchy. Do not use in combination of a block button to create hierarchy.
```
m-akinc, line 44: There are a few places that say "in combination of", which probably should be "in combination with". But that change results in very repetitive text. Maybe these should be like "Use in combination with ghost buttons (but not block buttons) to create a hierarchy of importance."

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -21,16 +21,80 @@ import * as buttonStories from './button.stories';
...
+
+        Use in combination with ghost buttons to create hierarchy.
+
+        When using an outline button, do not use it in combination of a block button to create hierarchy.
+
+        When using a block button, do not use it in combination of a outline button to create hierarchy.
+
+    ### Accent Buttons:
+
+    Use only 1 accent button in a section. It should be used when trying to achieve the most prominent eye catching approach. Consider the contextual implications
```
m-akinc, line 70: prominent eye catching approach -> prominent, eye-catching appearance

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -23,7 +23,7 @@ export const appearanceDescription = `This attribute affects the appearance of t
 </details>
 `;
 
-export const appearanceVariantDescription = `This attribute has no effect on buttons with a \`ghost\` appearance.
+export const appearanceVariantDescription = `This attribute has no effect on buttons with a \`ghost\` appearance. <br/> The appearance variant \`primary-accent\` has no effect on buttons using the "color" theme.
```
m-akinc, line 26: This should just say "accent" rather than "primary-accent", right?

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -253,89 +264,139 @@ export const styles = css`
...
-            :host([appearance-variant='primary']) .control:active {
-                box-shadow: 0px 0px 0px ${borderWidth} ${fillSelectedColor}
-                    inset;
-            }
-
-            :host([appearance-variant='primary'][disabled]) .control {
-                box-shadow: none;
-            }
-        `
+    appearanceVariantBehavior(
```
m-akinc, line 267: Rather than nesting behaviors here, you might be able to simplify things by defining a CSS private variable for the color to use, and then conditionally changing the value of that variable for `[appearance-variant='primary'] or `[appearance-variant='accent']. See how the tooltip styling changes color variables for severities: https://github.com/ni/nimble/blob/bfd00b11c6f63fd57a7902a104627abbdc97195b/packages/nimble-components/src/tooltip/styles.ts#L70

packages/nimble-components/src/utilities/style/appearance.ts
```diff
@@ -20,3 +20,14 @@ export function appearanceBehavior<AppearanceType>(
         styles
     );
 }
+
+export function appearanceVariantBehavior<AppearanceVariantType>(
```
m-akinc, line 24: This function may be unneeded if you can use a private CSS variable instead (see my comment on styles.ts)

packages/nimble-components/src/anchor-button/index.ts
```diff
@@ -36,7 +36,7 @@ export class AnchorButton
      * HTML Attribute: appearance-variant
      */
     @attr({ attribute: 'appearance-variant' })
-    public appearanceVariant: ButtonAppearanceVariant;
+    public appearanceVariant: ButtonAppearanceVariant = ButtonAppearanceVariant.default;
```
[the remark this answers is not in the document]
aidendk: Nope 👍 I'll fix in my newest commit.

packages/nimble-components/src/button/index.ts
```diff
@@ -38,7 +38,7 @@ export class Button
      * HTML Attribute: appearance-variant
      */
     @attr({ attribute: 'appearance-variant' })
-    public appearanceVariant: ButtonAppearanceVariant;
+    public appearanceVariant: ButtonAppearanceVariant = ButtonAppearanceVariant.default;
```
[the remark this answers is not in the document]
aidendk: Nope 👍

packages/nimble-components/src/patterns/button/styles.ts
aidendk: Yup, this works great! I added this in my newest commit, let me know what you think.

m-akinc: This looks good, but there's a lot of duplication. What I was hoping you might be able to do (I haven't looked closely enough to be sure) is something like this:

```css
export const styles = css`
    ...
    .control {
         --ni-private-button-background-color: (some value for the default appearance variant)
         --ni-private-button-text-color: (some value for the default appearance variant)
         background-color: var(--ni-private-button-background-color);
         color: var(--ni-private-button-text-color);
    }
    ...
`;
export const buttonAppearanceVariantStyles = css`
     :host([appearance-variant='primary']) .control {
         --ni-private-button-background-color: ${buttonPrimaryBackgroundColor};
         --ni-private-button-text-color: ${buttonPrimaryFontColor};
     }
     :host([appearance-variant='accent']) .control {
         --ni-private-button-background-color: ${buttonAccentBackgroundColor};
         --ni-private-button-text-color: ${buttonAccentFontColor};
     }
     ...
`;
```
Basically, the idea would be to see if you can set most/all of the attributes using a few private variables that you declare, then just change the values of those variables based on the `appearance-variant` value. If I'm not explaining clearly, I can come by and chat about it.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -254,87 +262,163 @@ export const styles = css`
 
 export const buttonAppearanceVariantStyles = css``.withBehaviors(
     appearanceBehavior(
-        ButtonAppearance.outline,
+        ButtonAppearance.block,
```
m-akinc, line 265: Is there a reason you swapped the order of the outline and block appearanceBehaviors? It makes the diff harder to make sense of.

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -9,6 +9,22 @@ export const comments: { readonly [key in TokenName]: string | null } = {
     dividerBackgroundColor: 'Divider background color',
     headerBackgroundColor: 'Background color for application headers',
     sectionBackgroundColor: 'Background color for sections',
+    buttonPrimaryBackgroundColor:
+        'Background Color for "primary" appearance-variant buttons',
```
m-akinc, line 13: nit: For consistency, do not capitalize "Color" (here and below).

m-akinc (approved):

aidendk:
Regarding Mert's comment about duplication:

This is also something I want to leave open to everyone for their feedback:

Currently there are two ways that I have implemented the styling for the appearance-variants.

**1. Conditionally changing the value of the variable (ex. `:host([appearance-variant='primary'])`)** 

(**Used in most recent commit**)

This implementation gets rid of the need for a MultivaluePropertyStyleSheetBehavior, but requires css variables to be declared multiple times. For example, when the button is in the hover state the outer border should have a color of `borderHoverColor`, as declared in .control:hover towards the top of the css class:

```
    .control:hover {
        border-color: ${borderHoverColor};
        box-shadow: 0px 0px 0px ${borderWidth} ${borderHoverColor} inset;
    }
```
Even though this is declared initially, because we are using :host to conditionally change the value of the variable in the `buttonAppearanceVariantStyles`, the `border-color: ${actionRgbPartialColor}` takes over the `${borderHoverColor}` that was initially declared, as it does not have :host.
Becuase of this, the border color has to be changed every time, as shown in the example below.
```
export const buttonAppearanceVariantStyles = css``.withBehaviors(
    appearanceBehavior(
        ButtonAppearance.outline,
        css`
            :host([appearance-variant='primary']) .control {
                border-color: ${actionRgbPartialColor};
            }

            :host([appearance-variant='primary']) .control:hover {
                border-color: ${borderHoverColor};
            }

           ...
```
<br/>
**2. Using another MultivaluePropertyStyleSheetBehavior for the appearance-variant attribute**

(**Used in [this commit](https://github.com/ni/nimble/blob/6ccd0e42cb63a1f69de83eafb86a2ecb474d323a/packages/nimble-components/src/patterns/button/styles.ts#L266)**)

Using the MultivaluePropertyStyleSheetBehavior does require adding another function to the `appearance.ts` file, but does not require re-applying variables like border-color for previously defined css classes because :host is not used.

```
export const buttonAppearanceVariantStyles = css``.withBehaviors(
    appearanceVariantBehavior(
        ButtonAppearanceVariant.primary,
        css``.withBehaviors(

           ...

            appearanceBehavior(
                ButtonAppearance.outline,
                css`
                    .control {
                        border-color: ${actionRgbPartialColor};
                    }

                    .control:hover {
                    }
```
Here, the border-color is still changed for .control because the border color is a different color when the appearance-variant is `primary`. Although, it is not required for the control:hover css class.

Either one works, but each has their own advantages and disadvantages. Let me know what you guys think! 😄

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -9,6 +9,22 @@ export const comments: { readonly [key in TokenName]: string | null } = {
     dividerBackgroundColor: 'Divider background color',
     headerBackgroundColor: 'Background color for application headers',
     sectionBackgroundColor: 'Background color for sections',
```
jattasNI, line 11: I'm trying to decide if I like these token names and whether they conform to our [token naming guidelines](https://github.com/ni/nimble/blob/123c631d25df675963ee5010eed7ebd203bd2898/packages/nimble-components/CONTRIBUTING.md#token-naming). My hesitation comes from not knowing where the variant parts like "Primary" or "BlockAccent" should go in the name. 

1. Is it a "state" in the guidelines, in which case it should go later:
`buttonBackgroundPrimaryColor`

2. Or is it a new concept, in which case we should describe it in the naming guidelines? I'd be ok with the guidelines saying that variants, if needed, go between "element" and "part".

Could you scan the existing token names to see if there's any precedence? If not we can pick whatever people like better (my vote is 2).

jattasNI: Also, what does "Action" mean in these names?

aidendk: Action referred to an action like clicking the button, as the tokens with "Action" in them were used in the .control:active css class. Looking back at it, this might be a bit confusing, so maybe a name like buttonPrimaryClickColor / buttonPrimaryFillClickColor would work better?

m-akinc: And in a couple cases there are multiple variant words in one token, e.g. `buttonOutlineAccentBorderColor`. It's a little confusing because "outline" sounds like it might be a part, but it's not. "Border" is the part.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -257,84 +265,160 @@ export const buttonAppearanceVariantStyles = css``.withBehaviors(
...
+                ${iconColor.cssCustomProperty}: white;
             }
 
-            :host([appearance-variant='primary']) .control:active::before {
-                border-color: ${fillSelectedColor};
-                border-width: ${borderWidth};
+            :host([appearance-variant='accent'][disabled])
+                slot[name='end']::slotted(*) {
+                ${iconColor.cssCustomProperty}: ${buttonLabelFontColor}
             }
```
rajsite, line 422: make sure your styles are in dom order make sure your styles are in dom order: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

The grouping by appearance variant is fine (technically not aligned with the guidelines) but within each of the appearance varaint groups make sure they are in dom order

packages/nimble-components/src/theme-provider/design-token-comments.ts
rajsite: The links provide strong precedence and (pretty good) alignment to the conventions:
 
element: link
part: (n/a)
states: Active, Disabled, Prominent
types: Font*

The only one that messed up is linkProminentDisabledFontColor should be linkDisabledProminentFontColor as the states should be in alphabetical order

https://github.com/ni/nimble/blob/17bef3c5becd15e8c25c735b42ba255f134a4e32/packages/nimble-components/src/theme-provider/design-token-comments.ts#L152-L185

I think it helps to write it down, your proposal is combinations of:

element: button
parts: fill, border, ~~background~~ (I think fill and backgrounds are synonyms for the part of the control, we are unfortunately inconsistent but I think we should settle on fill),
states: Primary, Accent, ~~"Action"~~ Active (active is the state the button is in where these are applied), Block, Outline

My questions:
- are the fills completely unique to the button only? Could they be used in other controls as `fillPrimaryColor`, `fillActivePrimaryColor`, `fillAccentColor`, `fillAccentActiveColor` instead?
- is the border color button specific? Could it be `borderAccentColor` instead?

change/@ni-nimble-angular-148cc61e-756b-4f60-8409-3c67783094a7.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: I don't see any angular changes, did you forget to push them?

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -257,84 +265,160 @@ export const buttonAppearanceVariantStyles = css``.withBehaviors(
...
-                .control:active {
-                padding: 0px;
+            :host([appearance-variant='primary']) [part='start'] {
+                ${iconColor.cssCustomProperty}: white;
             }
 
-            :host([appearance-variant='primary'][disabled]) .control {
-                background-clip: border-box;
+            :host([appearance-variant='primary'][disabled])
+                slot[name='start']::slotted(*) {
```
rajsite, line 354: I don't think all of these start and end slot selectors for targeting the icon color need `::slotted(*)`. Can you remove those to make it consistent?

change/@ni-nimble-angular-148cc61e-756b-4f60-8409-3c67783094a7.json
aidendk: [Mert recommended](https://github.com/ni/nimble/pull/1355#discussion_r1265879791) in a previous review that changes to the angular testing wasn't necessary since the implementation of the new states didn't change the point of the tests. I forgot to delete that change file after making the fix Mert recommended, I'll do that in my next commit.

packages/nimble-components/src/theme-provider/design-token-comments.ts
aidendk: I asked Brandon about this, once he responds I'll report what he says here 👍

aidendk: @rajsite I talked to Brandon about this, and he believes that the appearance variant color design tokens will only be used for the button. The one exception is the design token for the accent outline border color, which is used for the link rollover. Although, Brandon believes that using one token for both could be problematic if he eventually needs to make a change to the link colors, and the naming would be unusual since in the button case it's used for a border, but in the link use case it's used for text color.

For the appearance variant design tokens in general, he said that the colors he chose were very specific- he said that he doesn't plan on using combinations like Black91 with 75% or 85% opacity or Black15 with 20% opacity anywhere else and usually sticks to 10%, 30%, and 60% opacities for most nimble components. He said that the button variations had specific use cases that prompted him to use other opacities.

rajsite: Thanks for checking! Then I think rewriting the names to follow the conventions with the breakdown shown below makes sense (assuming I didn't miss anything :P):

element: button
parts: fill, border,
states: Primary, Accent, Active, Block, Outline

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -257,84 +265,156 @@ export const buttonAppearanceVariantStyles = css``.withBehaviors(
...
     ),
     appearanceBehavior(
         ButtonAppearance.block,
         css`
-            :host([appearance-variant='primary']) .control {
-                border-color: rgba(${actionRgbPartialColor}, 0.3);
-                box-shadow: 0px 0px 0px ${borderWidth}
-                    rgba(${actionRgbPartialColor}, 0.3) inset;
+            :host([appearance-variant='primary']) [part='start'] {
+                ${iconColor.cssCustomProperty}: white;
```
jattasNI, line 311: Right now we duplicate `${iconColor.cssCustomProperty}: white;` across many rules. Should we consider ways to reduce that duplication or is it needed because the value changes so much depending on the configuration of the button?

Ideas to consider:
- some theme aware tokens for icon color and disabled icon color
- local CSS variables for the same
- finding ways to apply the color on a less specific selector. If it only changes based on theme and disabled state then maybe we don't need to set it in rules that target specific appearance variants

jattasNI (approved):

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -74,7 +74,6 @@ export const styles = css`
         pointer-events: none;
         box-sizing: border-box;
         outline: 0px solid transparent;
-        outline: none;
```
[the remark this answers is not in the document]
aidendk: Sure, this line was removed from the .control::before of the normal button styling in the default button styling PR and I forgot to update it for the toggle button as well, so I removed it here.

packages/nimble-components/src/patterns/button/styles.ts
aidendk: My most recent commit has a solution that Mert came up with to fix this, which was cascade layers for the styles of the button. By doing this, a hierarchy could be created that allowed the `${iconColor.cssCustomProperty}: white;` along with other repeated styles to be removed since they don't have to be reiterated with cascade layers.

These layers have to be wrapped around every css class which could be unwanted, but at the same time it helps get rid of repeating css that could be confusing. The hierarchy of the layers follows the [css guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes).

msmithNI (approved):

aidendk:
@rajsite @atmgrifter00 Here's the links to the branches I was working on: button-variations (this branch) [button-variations2](https://github.com/ni/nimble/tree/button-variations2) and [button-variations3](https://github.com/ni/nimble/blob/button-variations3/packages/nimble-components/src/patterns/button/styles.ts).

Button-variations currently declares a layer for every css class, which is the functionality we're trying to move away from, but it's worth mentioning as it has correct styling for the buttons.

For button-variations2, I only used the `default` and `disabled` cascade layers. I reorganized the layers and grouped together the css classes that went with default (control, hover, focusVisible, active, etc.), and the ones that went with disabled (just disabled). If you run this in storybook you'll notice a few css styling errors, I intentionally didn't fix these- I'm not adding any extra styling to indicate that using only two cascade layers would require more reiterated css than using multiple.

For button-variations3, I used all 5 of the cascade layers, and grouped each cascade layer with its corresponding css classes (ex. all of the :hover classes went together in the hover layer). I kept all of the classes in the same order within each layer, so they still follow the grouping selectors guidelines. The cascade layers are organized from lowest to highest priority. A few places like `.withBehaviors` and the `buttonAppearanceVariantStyles` required their own layers but there's still significantly less repetition of layer names than there is in this branch (button-variations). The styling in this branch also correctly represents the Figma styling of the button appearance variants.

In terms of styling guidelines, I think that the cascade layers should be organized in their hierarchy, preferably with names that we already use (ex. default / control, hover, focusVisible, active, disabled, ..., then lastly css classes that don't have a layer, as these are the highest priority by default). Within those cascade layers, css classes would follow the [group selectors by target and document order](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order) guidelines.

Let me know what you think! 😄

atmgrifter00:
Between the 3 options (taking into consideration the issue you mentioned for button-variations2), I think my preference is for `button-variations3'. I personally find it helpful to find all of the CSS related to a particular layer contained within a single section.

change/@ni-nimble-blazor-803b27ab-c364-4038-afe2-da2ac9153dbf.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @rajsite @atmgrifter00 Here's the links to the branches I was working on: button-variations (this branch) [button-variations2](https://github.com/ni/nimble/tree/button-variations2) and [button-variations3](https://github.com/ni/nimble/blob/button-variations3/packages/nimble-components/src/patterns/button/styles.ts).

Button-variations currently declares a layer for every css class, which is the functionality we're trying to move away from, but it's worth mentioning as it has correct styling for the buttons.

For button-variations2, I only used the `default` and `disabled` cascade layers. I reorganized the layers and grouped together the css classes that went with default (control, hover, focusVisible, active, etc.), and the ones that went with disabled (just disabled). If you run this in storybook you'll notice a few css styling errors, I intentionally didn't fix these- I'm not adding any extra styling to indicate that using only two cascade layers would require more reiterated css than using multiple.

For button-variations3, I used all 5 of the cascade layers, and grouped each cascade layer with its corresponding css classes (ex. all of the :hover classes went together in the hover layer). I kept all of the classes in the same order within each layer, so they still follow the grouping selectors guidelines. The cascade layers are organized from lowest to highest priority. A few places like `.withBehaviors` and the `buttonAppearanceVariantStyles` required their own layers but there's still significantly less repetition of layer names than there is in this branch (button-variations). The styling in this branch also correctly represents the Figma styling of the button appearance variants.

In terms of styling guidelines, I think that the cascade layers should be organized in their hierarchy, preferably with names that we already use (ex. default / control, hover, focusVisible, active, disabled, ..., then lastly css classes that don't have a layer, as these are the highest priority by default). Within those cascade layers, css classes would follow the [group selectors by target and document order](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order) guidelines.

Let me know what you think! 😄

rajsite: Between the 3 options (taking into consideration the issue you mentioned for button-variations2), I think my preference is for `button-variations3'. I personally find it helpful to find all of the CSS related to a particular layer contained within a single section.

rajsite: FYI @aidendk @atmgrifter00  prefer using the comment thread on the file tab so conversations can be threaded.

Haven't had a chance to look yet, just moving the thread over

change/@ni-nimble-blazor-803b27ab-c364-4038-afe2-da2ac9153dbf.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
+  "comment": "Blazor integration for primary and accent appearance-variant buttons (#663)",
+  "packageName": "@ni/nimble-blazor",
+  "email": "103057880+aidendk@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
rajsite, line 7: original from @aidendk :

Regarding Mert's comment about duplication:

This is also something I want to leave open to everyone for their feedback:

Currently there are two ways that I have implemented the styling for the appearance-variants.

**1. Conditionally changing the value of the variable (ex. `:host([appearance-variant='primary'])`)** 

(**Used in most recent commit**)

This implementation gets rid of the need for a MultivaluePropertyStyleSheetBehavior, but requires css variables to be declared multiple times. For example, when the button is in the hover state the outer border should have a color of `borderHoverColor`, as declared in .control:hover towards the top of the css class:

```
    .control:hover {
        border-color: ${borderHoverColor};
        box-shadow: 0px 0px 0px ${borderWidth} ${borderHoverColor} inset;
    }
```
Even though this is declared initially, because we are using :host to conditionally change the value of the variable in the `buttonAppearanceVariantStyles`, the `border-color: ${actionRgbPartialColor}` takes over the `${borderHoverColor}` that was initially declared, as it does not have :host.
Becuase of this, the border color has to be changed every time, as shown in the example below.
```
export const buttonAppearanceVariantStyles = css``.withBehaviors(
    appearanceBehavior(
        ButtonAppearance.outline,
        css`
            :host([appearance-variant='primary']) .control {
                border-color: ${actionRgbPartialColor};
            }

            :host([appearance-variant='primary']) .control:hover {
                border-color: ${borderHoverColor};
            }

           ...
```
<br/>
**2. Using another MultivaluePropertyStyleSheetBehavior for the appearance-variant attribute**

(**Used in [this commit](https://github.com/ni/nimble/blob/6ccd0e42cb63a1f69de83eafb86a2ecb474d323a/packages/nimble-components/src/patterns/button/styles.ts#L266)**)

Using the MultivaluePropertyStyleSheetBehavior does require adding another function to the `appearance.ts` file, but does not require re-applying variables like border-color for previously defined css classes because :host is not used.

```
export const buttonAppearanceVariantStyles = css``.withBehaviors(
    appearanceVariantBehavior(
        ButtonAppearanceVariant.primary,
        css``.withBehaviors(

           ...

            appearanceBehavior(
                ButtonAppearance.outline,
                css`
                    .control {
                        border-color: ${actionRgbPartialColor};
                    }

                    .control:hover {
                    }
```
Here, the border-color is still changed for .control because the border color is a different color when the appearance-variant is `primary`. Although, it is not required for the control:hover css class.

Either one works, but each has their own advantages and disadvantages. Let me know what you guys think! 😄

change/@ni-nimble-blazor-803b27ab-c364-4038-afe2-da2ac9153dbf.json
jattasNI: @aidendk could you update the PR description to capture where you landed?

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -5,87 +5,86 @@ import {
     borderHoverColor,
     borderWidth,
     fillSelectedColor,
+    fillSelectedRgbPartialColor,
     smallDelay
 } from '../theme-provider/design-tokens';
 import { styles as buttonStyles } from '../patterns/button/styles';
 
 export const styles = css`
```
jattasNI, line 13: The Chromatic diffs for the toggle button matrix story look wrong to me: all of the buttons that were checked and disabled are now rendering as unchecked and disabled. Is that intentional?

image

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -22,16 +22,93 @@ import * as buttonStories from './button.stories';
...
+    These appearances have the default styling of the nimble-button. Each should be considered for use before using appearance variant buttons.
+
+    <br/>
+
+    #### Ghost Button:
+
+        Ghost is the default style and should be the first consideration for use.
+
+        Use as the default and standard option to create a clean airy and open UI feel. Ghost buttons fit comfortably in tight spaces and help control the visual density of the UI.
+        Be careful when using that the surrounding context does not confuse this button for emphasized body text, tabs or a standalone links.
```
jattasNI, line 40:
```suggestion
        Be careful when using that the surrounding context does not cause this button to be confused for emphasized body text, tabs or standalone links.
```

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -22,16 +22,93 @@ import * as buttonStories from './button.stories';
 
 <br />
 
+## Styling
+
+<br />
+
+    ### Appearances
```
jattasNI, line 29: Could you make the indentation in your changes more consistent? I think we don't usually bother to indent various levels of headers and their content. It can all just be left aligned.

packages/nimble-components/src/toggle-button/styles.ts
aidendk: It's intentional, but I need to double check with Brandon that it's correct- Brandon's [toggle button spec] shows the disabled button as being the gray colored background, and it's unclear if this design is for both checked and unchecked toggle buttons. I believe he's back this week, so I'll ask him.

aidendk: I double-checked with Brandon and the prior design was correct- it should have a different appearance when disabled in the checked state. I'm fixing this now 👍

jattasNI: Looking better. I noticed that one more state is incorrect: disabled + checked + hovered (unfortunately we don't have automated testing for hover states so had to check in manually). The disabled state should take precedence but I'm still seeing hover feedback for all themes in the "outline" and "ghost" appearance ("block" seems fine). Hover feedback is correctly disabled for regular buttons and anchor buttons.

I was testing in Firefox by hovering over all the "Checked .... Disabled" buttons in the [toggle button theme matrix](https://60e89457a987cf003efc0a5b-gbmhgmvpwd.chromatic.com/?path=/story/tests-toggle-button--toggle-button-theme-matrix)

packages/nimble-components/src/button/tests/button.mdx
```diff
@@ -22,19 +22,97 @@ import * as buttonStories from './button.stories';
 
 <br />
 
+## Styling
+
+<br />
```
rajsite, line 27: We shouldn't be using `<br>` tags for effectively styling purposes like this. Can you create a tech debt task to create a stylesheet for our mdx files that can styles our headers as we need and remove all these `<br>` tags?

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -13,166 +13,189 @@ import {
...
+    buttonBorderAccentOutlineColor,
+    buttonAccentOutlineFontColor
 } from '../../theme-provider/design-tokens';
 import { appearanceBehavior } from '../../utilities/style/appearance';
 import { ButtonAppearance } from './types';
 
 export const styles = css`
+    @layer default, hover, focusVisible, active, disabled;
+
     ${display('inline-flex')}
```
rajsite, line 32: move this into the default layer (it just emits a full style on it's own)

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -13,166 +13,189 @@ import {
...
 
-    :host([disabled]) slot[name='end']::slotted(*) {
-        opacity: 0.3;
+    @media (prefers-reduced-motion) {
+        .control {
+            transition-duration: 0s;
+        }
     }
 
     :host([content-hidden]) [part='end'] {
```
rajsite, line 201: I feel like instead of relying on styles outside of a layer having a higher importance we should instead make a layer named `top` and put it at the top of the layer order. An easier story to say every style needs to be in one of the layers.

Also if we have `top` maybe we should rename `default` to `base`.
`@layer base, hover, focusVisible, active, disabled, top;`

Base feels a little more appropriate than `default` because it has more styles than just default state styles, i.e. it has styles for the content-hidden state as well.

Also open to other naming ideas.

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -5,87 +5,106 @@ import {
...
+                ${fillSelectedColor}
+            );
+            border-color: rgba(${fillSelectedRgbPartialColor}, 0.3);
+            position: relative;
+            transition: box-shadow ${smallDelay} ease-in-out,
+                border-color ${smallDelay} ease-in-out,
+                background-size ${smallDelay} ease-in-out;
+            background-size: 100% 100%;
+            background-repeat: no-repeat;
+            background-position: center;
```
rajsite, line 30: none of these are necessary to redefine right? These were all set in the default / base of `.control` in the buttonStyles already this is just overwriting everything with the same value?

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -5,87 +5,106 @@ import {
...
+            content: '';
+            position: absolute;
+            width: 100%;
+            height: 100%;
+            pointer-events: none;
+            box-sizing: border-box;
+            outline: 0px solid transparent;
+            background-clip: content-box;
+            transition: outline ${smallDelay} ease-in-out;
+        }
```
rajsite, line 43: These styles are completely defined in buttonStyles already it looks like. We can get rid of this altogether right?

packages/nimble-components/src/button/tests/button.mdx
aidendk: Sure, here's the tech debt issue #1413 👍

packages/nimble-components/src/patterns/button/styles.ts
rajsite: re-opening because there was a mixup. The display style needs to move into the default layer, we can keep the layer definition outside

```css
@layer base, hover, focusVisible, active, disabled, top;

@layer base {
    ${display('inline-flex')}

    :host {
   ...
   }
}
```

rajsite (approved):

PR #1357, opened 2023-07-18
Fix icon color for disabled menu items and disabled text fields

### LINKED ISSUE #1356: Nimble icon not disabled when used with nimble menu
# 🐛 Bug Report

We wanted to create a dropdown list containing some menu items. So we started using the `nimble-menu` component. Each menu item has a nimble icon associated with it. Under certain conditions, we need to disable some menu items. When we tried setting up the `disabled` property of the menu item, it is getting unclickable, the text in the button is getting greyed out. But, the nimble icon associated with the menu item is not greyed out.

When we checked this in the [storybook](https://nimble.ni.dev/storybook/?path=/story/components-menu--menu-item&args=disabled:!true), here again the nimble icon is not disabled when we try to disable the menu item.

## 💻 Repro or Code Sample

```
<nimble-menu-item [disabled]="true">
    <nimble-icon-eye slot="start"></nimble-icon-eye>
    Menu Item
</nimble-menu-item>
```

## 🤔 Expected Behavior

The icon in the menu item should be greyed out, along with the text.

## 😯 Current Behavior

The icon is not getting greyed out, only the text in the menu item is greyed out.

image

## 🌍 Your Environment

* OS & Device: Windows 10 PC
* Browser: Chrome
* Version 114.0.5735.199

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1356 
Also addresses the color of the `start` icon in a disabled text field.

## 👩‍💻 Implementation

The icons in the `start` slot of the `nimble-menu-item`, `nimble-anchor-menu-item`, and `nimble-text-field` did not have the correct color when disabled. I've updated the styling to set `${iconColor.cssCustomProperty}`.

## 🧪 Testing

- Verified locally in storybook that disabled `nimble-menu-item` elements, disabled `nimble-anchor-menu-item` elements, and disabled `nimble-text-field` elements with icons have the correct icon color.
- Updated the matrix test for the menu to have a disabled menu item with an icon. Previously it only had a disabled anchor menu item with an icon.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
+++ b/change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix icon color for disabled menu items",
+  "comment": "Fix icon color for disabled menu items and disabled text fields",
   "packageName": "@ni/nimble-components",
   "email": "20542556+mollykreis@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
+++ b/change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix icon color for disabled menu items",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor-menu-item/styles.ts
+++ b/packages/nimble-components/src/anchor-menu-item/styles.ts
@@ -9,6 +9,7 @@ import {
     controlHeight,
     fillHoverColor,
     fillSelectedColor,
+    iconColor,
     iconSize
 } from '../theme-provider/design-tokens';
 
@@ -71,7 +72,7 @@ export const styles = css`
     }
 
     slot[name='start']::slotted(*) {
-        fill: currentcolor;
+        ${iconColor.cssCustomProperty}: currentcolor;
         width: ${iconSize};
         height: ${iconSize};
     }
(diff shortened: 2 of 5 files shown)


change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Should we make similar changes for `tree-item` and `anchor-tree-item`?

jattasNI (approved):

change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
mollykreis: We should revisit the disabled styling of the tree item components because I don't think they are correct (simply applying 50% opacity to the whole item, icon and all, when disabled). However, I didn't see any visual designs that showed what the disabled state of a tree item should be. I'm guessing they should be styled the same as a disabled menu item, but I didn't want to clutter this PR with other changes that I wasn't sure about.

We probably need to reach out to designers and double check that the same pattern from the menu items applies to the tree items.

change/@ni-nimble-components-39e27099-91fe-44c4-8501-23b6e9f4c012.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: Do we need a similar change for text field?
image

I think we intentionally avoid it for slotted action buttons requiring the app to style but maybe we should style slotted start icons?

mollykreis: Based on [the XD spec](https://xd.adobe.com/view/33ffad4a-eb2c-4241-b8c5-ebfff1faf6f6-66ac/screen/842889a5-67ba-4350-91c1-55eee48f4fa2/specs/), I think you are correct about the 'start' icon in text fields. I've added the styling fix for that in this PR since it was a very small change.

Issue #801, opened 2022-11-02
Critical vulnerability in socket.io-parser


jattasNI:
## 🧹 Tech Debt
`npm audit` started failing due to [a new critical vulnerability in socket.io-parser](https://github.com/advisories/GHSA-qm95-pgcg-qqfq). This blocks the main PR workflow.

The package is pulled in by `karma` and `@11y/eleventy`:
- eleventy has fixed [a similar issue](https://github.com/11ty/eleventy/issues/2213) in the past but [I haven't seen a discussion of this vulnerability yet](https://github.com/11ty/eleventy/issues?q=is%3Aissue+socket.io)
- karma has [an issue to track the vulnerability](https://github.com/karma-runner/karma/issues/3823) but no fix yet

```
npm ls socket.io-parser

├─┬ @ni/nimble-components@15.1.1 -> ./packages/nimble-components
│ └─┬ karma@6.3.4
│   └─┬ socket.io@3.1.2
│     └── socket.io-parser@4.0.4
└─┬ @ni/site@1.0.0-beta.0 -> ./packages/site
  └─┬ @11ty/eleventy@1.0.2
    └─┬ browser-sync@2.27.10
      ├─┬ browser-sync-ui@2.27.10
      │ └─┬ socket.io-client@4.5.2
      │   └── socket.io-parser@4.2.1
      └─┬ socket.io@4.5.2
        └── socket.io-parser@4.2.1
```

rajsite:
@fredvisser does Snyk replace the functionality that `npm audit` provided? If so then the action for this issue is to remove the audit steps from `main.yml`. Marking as "discussion" to see if we can make this issue go away.

fredvisser:
Snyk doesn't support monorepos by default, so isn't picking up any of the package specific package.json files. So the current setup isn't good enough.
While they have a CLI flag that will find the other package.json files, we're using a web hook which doesn't appear to expose the option. 

We could switch to a their Github Action or we could add the [Github Dependency Review Action](https://github.com/actions/dependency-review-action) which would keep the workflow entirely with Github.

rajsite:
Fixed in #1930

rajsite:
@fredvisser looking at the dependency review check instructions it says to check the following page: https://github.com/ni/nimble/security/dependabot

where there I see non-dismissed critical vulns, but it doesn't look like dependency review check is complaining about those, see build:
https://github.com/ni/nimble/actions/runs/8439038563

and log:
[...]

PR #1335, opened 2023-06-30
Rich text viewer | Implementation of markdown parser

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

This PR includes the logic for parsing markdown input and rendering the rich text content into the viewer component.
Issue: https://github.com/ni/nimble/issues/1288
AzDo User Story: [Show the rich text content by parsing the markdown input](https://dev.azure.com/ni/DevCentral/_backlogs/backlog/ASW%20SystemLink%20Platform/Initiatives/?workitem=2426496)

## 👩‍💻 Implementation

* Added [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown/tree/master) package for parsing markdown strings, and [prosemirror-model ](https://github.com/ProseMirror/prosemirror-model) package for DOM serializer to serialize the content and render the rich text content in the UI.
* Added `markdownValue` accessors to allow setting and retrieving the markdown value. When setting the markdown input string, the component performs operations to convert the markdown string into corresponding document fragments for loading it into the container of the viewer component.
* Enabled `emphasis`, `list`, and `autolink` from the [markdown-it](https://github.com/markdown-it/markdown-it/tree/master), which is internally used by prosemirror-markdown. This allows users to add bold, italics, numbered, and bulleted lists, as well as direct links in a CommonMark flavor. All other basic markdown formats are disabled.
* Added `json()` from the "@rollup/plugin-json" in the rollup.config.js because prosemirror-markdown and markdown-it internally use JSON files during the build process. Referred to this solution from the discussion [here](https://discuss.prosemirror.net/t/unexpected-token-on-importing-of-markdown-stuff/2360/4). 
* Also, added `nodePolyfills()` from [rollup-plugin-polyfill-node](https://www.npmjs.com/package/rollup-plugin-polyfill-node) as the `markdown-it` requires `punycode`([code reference](https://github.com/markdown-it/markdown-it/blob/master/lib/index.js#L14)) but as mentioned in the [punycode](https://www.npmjs.com/package/punycode), the right way to use it is `require('punycode/')`. So use `nodePolyfills()` to rollup the bundle with the `punycode` properly imported in the `all-components.bundle.js`.

Spec update:
1. Latest `comments` desktop and mobile mockup links and updated a few GitHub permalinks. 
2. Removed the `fit-to-content` attribute for the viewer component. 
3. `markdown` accessor is changed to a property for the viewer component.

## 🧪 Testing

Added unit tests and manually tested and verified the changes.

[...]

PR #1361, opened 2023-07-18
Update dependencies to latest supported

### LINKED ISSUE #1146: Update storybook when SB7 regressions are fixed
## 🧹 Tech Debt

As part of #1134 we filed some issues to Storybook about workarounds we had to adopt and features we had to disable. This issue tracks the work to adopt the fixes once they're available.

- [x] https://github.com/storybookjs/storybook/issues/21820. When this is fixed we can remove `patches/@storybook+html+7.0.0-rc.8.patch` and delete the peer dependency on `@storybook/html: *` from the root `package.json`. This is fixed in 7.0.8 but we can't uptake until this is fixed: https://github.com/storybookjs/storybook/issues/22384
- [x] Remove workaround in `storybook-declarations.d.ts` added in #1154 once this issue is fixed: https://github.com/storybookjs/storybook/issues/21887
- [x] Some code previews (table, tooltip, dialog) are not available in the Docs page. https://github.com/storybookjs/storybook/issues/21893
- [x] The `configureActions` parameter `depth` in `preview.js` changed meaning. We adapted in #1154 but if they change it back we should revert that. https://github.com/storybookjs/storybook/issues/21933
- [x] When we try to uptake the official release instead of an RC, nimble-table rows don't render. Root cause not yet identified, but this Nimble PR shows the issue. #1147

rajsite:
# Pull Request

## 🤨 Rationale

Updates dependencies to the latest resolved by our package.json files without modifying those specifiers. Hoping it pre-emptively helps the discussion of [issues around dependency changes](https://github.com/ni/nimble/pull/1335#discussion_r1260409646) but useful regardless.

As a bonus, fixes #1146 

## 👩‍💻 Implementation

- Regenerate the lock file (delete lock file, node_modules, and npm install)
- Update CI node and npm version to reflect Skyline recent update and better reflect local dev
- Pin Blazor playwright version to correspond to latest released on nuget.org (as of writing [1.36.0](https://www.nuget.org/packages/Microsoft.Playwright)) vs latest released on npm which the previous specifier would resolve to (as of writing [1.36.1](https://www.npmjs.com/package/playwright)).
- Update tests and stories based on lint rule behavior changes

## 🧪 Testing

Manual test storybook build but rely on CI for most.

## ✅ Checklist


change/@ni-nimble-components-e81dea16-5dd0-4d2d-8814-c5f263d310f1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
jattasNI, line 2: I likely need a refresher on why this change doesn't impact the built bits. Wouldn't changing our build tool versions potentially modify the build output?

And the facts that this is 
1. causing some tests to fail
2. causing chromatic diffs
3. changing the version of Angular packages that we test against 
...also give me pause. 

I guess our `package.json`s already said that clients could satisfy Nimble's dependencies with these newer versions. We just never tested against them and now that we are, we're discovering that they behave differently? Does that imply that we should have been pinning our package.json dependencies?

(possible we should have a call to talk through these points)

change/@ni-nimble-components-e81dea16-5dd0-4d2d-8814-c5f263d310f1.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Was hoping this would fix #801 and allow us to re-enabling `npm audit` for dev dependencies, but I still see 3 critical vulnerabilities in `minimist`. Oh well.

change/@ni-nimble-tokens-53d62bd1-be2a-4fe1-9e78-13c91edb04b2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Lint disable without behavior change",
+  "packageName": "@ni/nimble-tokens",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
```
jattasNI, line 7: GitHub's rich diff for the package-lock.json (click the 📄 icon) shows changes to nimble dependencies. Is that normal? 

image

change/@ni-nimble-components-e81dea16-5dd0-4d2d-8814-c5f263d310f1.json
msmithNI: In this case one of the changes was an update to `@tanstack/table-core` which had a behavior change around sorting undefined values. That's now resolved (and our tests did catch it) but since we're not only updating devDependencies, I'd say build output is definitely affected so these should probably be `patch` at least.

change/@ni-nimble-tokens-53d62bd1-be2a-4fe1-9e78-13c91edb04b2.json
jattasNI: Discussed offline and these are normal diffs for a PR that modifies package versions.

change/@ni-nimble-components-e81dea16-5dd0-4d2d-8814-c5f263d310f1.json
jattasNI: Decision after offline discussion was to bump the tanstack version in package.json to reflect this change, which means this definitely merits a `patch`

rajsite:
@msmithNI @atmgrifter00 waiting on blazor reviewer

rajsite:
> @msmithNI @atmgrifter00 waiting on blazor reviewer

Gonna override y'all. Can address concerns in a follow-up

Issue #1014, opened 2023-01-30
Table allows configuration of datetime column type


nate-ni:
## 😯 Problem to Solve
There is a need to provide a date (and/or time) column type. Initially, only date may be needed.
Having a column for this type provides support for sorting, grouping and filtering. It will also be able to integrate locale specific formatting to support i18n and l10n.

### In
Displaying the date to the user in a locale correct format for numeric displays
Providing display format options (long year, short year, etc.)
Sorting of column works as expected

## Open Questions
- Do we need to support time display in the initial version?
  - Probably
- Do we provide numeric (01/01/2023) AND text (January 01, 2023) versions of formatting?
  - How does that work along with locale specific formatting?
- Will we need a 3rd party library for formatting? (date-fns, moment, luxon, etc.??). If so which one?
- What type(s) of input do we accept for this column type (Date object, timecode seconds, timecode ms.?)
- Which width modes should the column support? Both fixed pixel width and fractional width?

## Prior Art
- [x] #887

### Example of date being used
![image](https://user-images.githubusercontent.com/109235103/218489678-62dd6988-cfd3-4bb3-a28e-8555c8c2fe60.png)

![Image](https://user-images.githubusercontent.com/109235103/218791672-d00441d0-f99c-47d0-96ad-45b27033802a.png)

```[tasklist]
## 📋 Tasks
- [x] Create custom elements for date column with basic API and default date formatting
- [x] Unit tests for behavior with invalid dates
- [x] Unit tests for formatting number edge cases (Inf, NaN, etc)
- [x] Add API to configure custom date formatting
- [x] Unit tests for invalid state with bad formatting configuration
- [x] Respect Nimble locale token for date formatting
- [x] Angular integration
- [x] Blazor integration
```

PR #1363, opened 2023-07-19
Initial version of date column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1014 

## 👩‍💻 Implementation

This is the first, basic version of the column. It does not support any configuration. It always formats the datetime in the way that will become the default when we later add format configuration. The column derives from `TableColumnTextBase`.
- component defined
- initial Storybook stories/documentation created
- unit tests written

## 🧪 Testing

Tested in Storybook and created unit tests.

## ✅ Checklist


packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * A cell view for displaying date/time fields as text
+ */
+export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
+TableColumnDateTextCellRecord,
+TableColumnDateTextColumnConfig
+> {
+    private columnConfigChanged(): void {
+        this.placeholder = this.columnConfig.placeholder;
```
rajsite, line 24: @jattasNI There was strong discussion about removing the concept of placeholder values from the table: https://github.com/ni/nimble/issues/1364

As this and number text column start dev should we confirm that decision? How should that impact this PR?

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    private cellRecordChanged(): void {
+        if (typeof this.cellRecord.value === 'number') {
+            const formatter = new Intl.DateTimeFormat(undefined, {
+                dateStyle: 'medium',
+                timeStyle: 'medium'
+            });
+            try {
+                this.text = formatter
+                    .format(this.cellRecord.value)
+                    .replace('\u202f', ' '); // on Chrome, the space before AM/PM is a narrow non-breaking space. For testing consistency across browsers, replace it with a regular space.
```
rajsite, line 36: I'm not sure this is a buggy behavior worth overriding. Did you check from that perspective?

If the goal is just making tests easier it should be a part of the page object. I'm not sure there is a clear benefit in trying to override the details of the native date formatter to get the exact same rendered text across browsers. Is there a specific argument for that?

One argument against could be that many apps are likely to use the same formatter. So our date text will render similarly to most other web apps in Chrome.

m-akinc: I don't think it's a bug--just a mostly trivial difference between how each browser formats the date. It seems more targeted to doing the replacement here than in the page object. If I do it in the page object, I either have to do this substitution for *all* text rendered in cells, or add an esoteric new function to fetch a rendered date from a cell. I don't see any practical downsides to replacing one narrow non-breaking space with a regular space. That was my thought process, at least.

rajsite: Seems like the date table column could have it's own page object

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
m-akinc: Has design/UX had any input regarding placeholders in the table? If not (and I suspect not), I can pivot to removing that from our table columns, as that would be a precursor to this PR.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
rajsite: > don't see any practical downsides to replacing one narrow non-breaking space with a regular space.

- It's arbitrary and not clear why it's actually needed
- Only thing it cites is improving testing, why is that leaking in product code
- The logic is duplicated poorly across multiple files
- Doesn't link to anything explaining the behavior, why it's allowed or not allowed, etc.
- No regression testing behavior for the browser-specific workaround
- No guidance or analysis of other browser-specific behaviors we need to be aware of with this api

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
rajsite: That is the open question cited in the linked issue that should be resolved

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text-matrix.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+        date: new Date('Dec 21, 2020, 3:45:03 PM').valueOf()
+    },
+    {
+        id: '2'
+    }
+] as const;
+
+// prettier-ignore
+const component = (): ViewTemplate => html`
+    <label style="color: var(--ni-nimble-control-label-font-color); font: var(--ni-nimble-control-label-font)">Date Text Table Column</label>
```
rajsite, line 37: Don't use the bare tokens: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#design-tokens

If that pattern was copied from somewhere clean it up.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
m-akinc: I've created PR #1365 to remove placeholder support from table columns.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text-matrix.stories.ts
m-akinc: It was copied from the anchor column matrix stories, but I'm cleaning that up in my PR removing placeholder support.

jattasNI (approved):

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+    it('reports column configuration valid', async () => {
+        await connect();
+        await waitForUpdatesAsync();
+
+        const firstColumn = element.columns[0] as TableColumnDateText;
+
+        expect(firstColumn.checkValidity()).toBeTrue();
+    });
+
+    const badValueData = [
```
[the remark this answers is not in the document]
m-akinc: I went ahead and added these tests for the three column types. We will have to remember to do the same for the icon and mapping columns.

packages/nimble-components/src/table-column/date-text/group-header-view/index.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * The group header view for displaying date/time fields as text.
+ */
+export class TableColumnDateTextGroupHeaderView extends TableColumnTextGroupHeaderViewBase<
+TableNumberFieldValue,
+TableColumnDateTextColumnConfig
+> {
+    private groupHeaderValueChanged(): void {
+        if (typeof this.groupHeaderValue === 'number') {
+            const formatter = new Intl.DateTimeFormat(undefined, {
```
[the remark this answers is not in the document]
m-akinc: I would prefer to kick this can down the road a little bit, to when I add configuration. I'll have a better sense then for how much complication it adds vs benefit.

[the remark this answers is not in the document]
rajsite: Seems like it would be easy to make it private static (pretty much as good as it gets) and then handle it being dependent on state in the future.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -0,0 +1,48 @@
...
+export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
+TableColumnDateTextCellRecord,
+TableColumnDateTextColumnConfig
+> {
+    private cellRecordChanged(): void {
+        if (typeof this.cellRecord.value === 'number') {
+            const formatter = new Intl.DateTimeFormat(undefined, {
+                dateStyle: 'medium',
+                timeStyle: 'medium'
+            });
```
rajsite, line 28: Same as below lets go ahead and make this private static and then see how it changes based on state in the future

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
```diff
@@ -0,0 +1,120 @@
...
+import {
+    SharedTableArgs,
+    sharedTableActions,
+    sharedTableArgTypes,
+    sharedTableArgs
+} from '../../base/tests/table-column-stories-utils';
+import { tableColumnTextTag } from '../../text';
+
+const simpleData = [
+    {
```
rajsite, line 19: nit / optional / friday afternoon:
Some simpsons birthdays ([floating timeline 2023](https://simpsons.fandom.com/wiki/Floating_Timeline#With_Floating_Timeline_(2023))):
Patty and Selma Bouvier (twins) - June 11, 1979
Maggie Simpson - January 12, 2022
Homer Simpson - May 12, 1984
Marge Simpson - March 19, 1984
Bart Simpson - April 1, 2013
Lisa Simpson - May 9, 2015

rajsite (approved):

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
m-akinc: I had considered that, but realized these characters have perpetually changing birthdays. Anyway, I've changed them to birthdays (but left the random times).

Issue #1364, opened 2023-07-19
Remove placeholder concept from table columns


rajsite:
# 🎙 Discussion

In [discussions about the enum columns](https://github.com/ni/nimble/pull/1274#discussion_r1248310568) it was determined that a concept of placeholder / default mappings was inappropriate. In that case a user should pre-format the data and add a state to their mappings to represent the placeholder / default value if they so choose. We aren't trying to compensate for poorly formatted data and trying to distinguish between mapped values, unmapped values, and undefined. If a value is unmapped for enum columns, it renders blank.

The statement was also made that in general placeholders don't make sense for the table. We should prefer to not render invalid data over showing a placeholder as that pollutes the contents of the table. Showing lots of rows of placeholder text in a column just clutters the table. 

There was strong agreement from the team to remove the concept of placeholder / default-mapping from enum column.

A concrete decision for all column types might need buy-in from UX?

jattasNI:
Some additional rationale captured for posterity:
 - in number columns we hope to right align valid values to improve scanning but left align things that don't have a consistent decimal location. Placeholders could make this look ugly with right-aligned valid values and left-aligned placeholder text
 - we aren't aware of specific requests to support placeholder content. If there are any we can add the feature back then, along with guidance about when it should be used rather than showing an empty cell.

PR #1365, opened 2023-07-19
Remove placeholder support from table columns

m-akinc:
# Pull Request

## 🤨 Rationale

Discussed in #1364 

## 👩‍💻 Implementation

Removed `placeholder` support/references from:
- all table (column) specs/hlds
- text base table column (and tests)
- anchor table column (and tests)
- text table column (and tests)
- Angular directives/tests
- Blazor files

## 🧪 Testing

Removed some tests related to placeholders and updated others to test for blank instead of placeholder value. All automated tests pass.

## ✅ Checklist


rajsite (approved):

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor.directive.ts
```diff
@@ -35,14 +35,6 @@ export class NimbleTableColumnAnchorDirective extends NimbleTableColumnBaseDirec
         this.renderer.setProperty(this.elementRef.nativeElement, 'hrefFieldName', value);
```
rajsite, line 35: Looking at the chromatic tests I think this case looks extra weird because all values on the row are undefined. But my question is when you group on an undefined / absent value, how should the group header render?

image

This probably applied before too if placeholder was not set, guess that just wasn't covered in tests.

rajsite: We also may have multiple potential "blank" groupings, i.e. number columns would represent different NaN values as blank

m-akinc: Yeah, I would not consider this in scope for this PR, but it's a fair question. My implementation for the mapping columns printed the actual record value in the group header. This was to address the scenario with multiple blank group headers. We could do the same for number/date columns if we want. When the record value is actually missing/null, we could make the header show something like "*Blank* (1)". That's probably better than what we currently have.

rajsite: Agree that it's not in scope for this PR and was a pre-existing concern just unmasked with these changes. Also think the cases for a string column are something like: null, undefined, empty string, and string with various white space. Think all of those would look weird as group headers.

The advice today is to pre-format your data to avoid those cases. Can you create a discussion issue for us to gather client feedback around the behavior for the future? Then we can link to that and resolve the comment for this PR.

m-akinc: Created #1367

change/@ni-nimble-angular-4fac20cb-60b2-4732-b697-bb71f0a26238.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: All these can be `patch` since it's an incubating component.

But you will need to fix the breaking change in systemlink-lib-angular since the sl-table columns expose these.

jattasNI (approved):

jattasNI:
Given that these changes are fairly mechanical, I'm ok if you don't wait for the remaining owners to approve.

PR #1231, opened 2023-05-08
Ensure Blazor components can take additional attributes

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1191

## 👩‍💻 Implementation

- Added missing `[Parameter(CaptureUnmatchedValues = true)]` to two components (breadcrumb item and tabs)
- Added missing `AdditionalAttributes` parameter to table columns (base class)
- Added test to each component

## 🧪 Testing

Ensured new test fails w/o `[Parameter(CaptureUnmatchedValues = true)]` and passes with it.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1366, opened 2023-07-19
Blazor integration for label providers

msmithNI:
# Pull Request

## 🤨 Rationale

Partially addresses #1090 (a few small followup items remain in the work item once Angular/Blazor PRs go in)

## 👩‍💻 Implementation

See the [HLD](https://github.com/ni/nimble/tree/main/specs/labels-and-localization) for the implementation plan for the label providers.

The Blazor integration is fairly rudimentary (just the label provider APIs). To fully use them in a Blazor app, the client apps need to handle localizing the strings themselves, as outlined in the Readme updates.

## 🧪 Testing

Added basic autotests. Added label providers to example app MainLayout (matches guidance from the Readme).

## ✅ Checklist

    - Updated Readme to add Localization guidance. Also added brief guidance on NimbleThemeProvider since it wasn't mentioned in this Readme.


packages/nimble-blazor/NimbleBlazor/Components/NimbleLabelProviderCore.razor.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string? NumericIncrement { get; set; }
+
+    [Parameter]
+    public string? NumericDecrement { get; set; }
+
+    /// <summary>
+    /// Gets or sets a collection of additional attributes that will be applied to the created element.
+    /// </summary>
+    [Parameter(CaptureUnmatchedValues = true)]
+    public IReadOnlyDictionary<string, object>? AdditionalAttributes { get; set; }
```
atmgrifter00, line 23: Not sure if we should just skip adding the `AdditionalAttributes` junk. This component certainly doesn't seem like it will benefit from the use of arbitrary attributes (plus the fact that I believe we want to ultimately move away from this API for all components).

atmgrifter00 (approved):

packages/nimble-blazor/README.md
```diff
@@ -94,6 +98,22 @@ There are currently extra manual steps required to use the Nimble design tokens
 The SCSS compilation happens before the rest of Blazor's compilation, so this approach works fine with Blazor CSS isolation.  
 Note: This approach requires periodically updating the Nimble tokens SCSS files manually (whenever the Nimble Blazor NuGet version is updated).
 
+### Localization
```
jattasNI, line 101: For parity's sake I guess we should link to these docs from Storybook in a `## Blazor` section.

packages/nimble-blazor/NimbleBlazor/Components/NimbleLabelProviderCore.razor.cs
msmithNI: Removed

packages/nimble-blazor/README.md
msmithNI: Done

rajsite (approved):

packages/nimble-blazor/NimbleBlazor/Components/NimbleLabelProviderCore.razor.cs
rajsite: @atmgrifter00 for consistency based from the last discussion I was advocating for putting AdditionalAttributes on everything: https://github.com/ni/nimble/pull/1231#discussion_r1189089284
My understanding is that is the current direction. The reasoning being these are web components, not blazor components, and have additional properties that are valid.

This does keep coming up, so if we want to reverse that decision and opt-in on only some components adding AdditionalAttributes configuration then we can. In that case I would like a way to see easily in the component that it was decided that it was opted out, not that it was accidentally forgotten.

rajsite: Our current conventions also state to always include it: https://github.com/ni/nimble/blob/main/packages/nimble-blazor/CONTRIBUTING.md#conventions

So if a direction change is being made that should be updated as well

msmithNI: Fair points. Added it back for consistency.

PR #1368, opened 2023-07-20
New API for Grid Dimensions in Wafer Map Component HLD

munteannatan:
# Pull Request

## 🤨 Rationale

The wafer needs new API for internal grid dimensions

## 👩‍💻 Implementation

The Azure PR for including these changes is [here](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/528792).

## ✅ Checklist


packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
```diff
@@ -0,0 +1,26 @@
...
+
+The inputs will be validated before use. If they are undefined or the data is missing the previous behavior of calculating the values based on the input data will be implemented.
+
+## Alternative Implementations / Designs
+
+The API can be changed to use four numeric inputs for each of the values, or a new Type can be created to structure all the dimensions together.
+
+## Open Issues
+
+-   Should the detection of partial data on the new inputs trigger a visual warning?
```
jattasNI, line 26: The strategy we've taken with the table is:
1. if the client developer configured the component incorrectly, the component renders blank to make it very obvious that something is wrong
2. the component [exposes "validity" APIs](https://nimble.ni.dev/storybook/?path=/docs/incubating-table--docs) to report whether the configuration is valid and explain what's wrong with it. These can be called in tests to assert there are no problems or when debugging to see why the table is rendering blank (they wouldn't typically be called in production)

We've been pretty happy with that approach so you could consider it for this component. Item 2 may be "nice to have" but I think defining consistent behavior (and autotesting that behavior) when configuration is invalid is important.

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Implementation / Design
+
+The new attributes and properties will be described in the component definition. They will be created with change listeners which will trigger the appropriate updates in the wafer.
+
+The inputs will be validated before use. If they are undefined or the data is missing the previous behavior of calculating the values based on the input data will be implemented.
+
+## Alternative Implementations / Designs
+
+The API can be changed to use four numeric inputs for each of the values, or a new Type can be created to structure all the dimensions together.
```
jattasNI, line 22: Could you also describe why these alternatives were rejected in favor of the proposed approach?

munteannatan: They are not rejected, sorry if that is the way it appears. I chose this approach because the PointCoordinates is already a public type, merging all 4 of them in 1 property will need a new public type, and having them all as attributes will increase the chance of missing information, but they are all valid in my opinion

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
munteannatan: I think we can try and implement this approach covering only this case for now and expand it in a later feature or tech debt

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -95,6 +95,11 @@ _The key elements of the component's public API surface:_
...
     -   bottomRight - ![Bottom Right Quadrant](./Resources/bottom_right.png)
     -   `orientation` - represent the orientation of the notch on the wafer map outline (only visual). As only four static orientations are possible, it can be represented by an Enum with the following values: top, bottom, left, right.
+    -   `origin` - represents the origin coordinates of the dies grid for rendering the wafer map. Leaving the value `undefined` will set the value to the minimum corner of the bounding box of the dies coordinates. It will be an object of the type `PointCoordinates` with the following attributes:
+        -   x: number
+        -   y: number
+    -   `numberOfRows` - represents the number of values on the Y axis. Leaving the value `undefined` will set the value to the difference between the maximum and the minimum Y values of the bounding box of the dies coordinates plus one.
+    -   `numberOfColumns` - represents the number of values on the X axis. Leaving the value `undefined` will set the value to the difference between the maximum and the minimum X values of the bounding box of the dies coordinates plus one.
```
rajsite, line 102: Can these just be `rows` and `columns`

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
```diff
@@ -0,0 +1,28 @@
+# Nimble Wafer Map Grid Dimensions
+
+## Problem Statement
+
+The Wafer Map displays dies on a canvas inside a circular border based on an internal grid which is calculated based on the data input.
+This approach is limiting the performance of the wafer and the grid dimensions can be erroneous if the displayed data is just a subset of a larger data set.
+To mitigate these problems, we will reveal three new API inputs which will allow the users to set the grid dimensions themselves, thus saving processing time and correcting inconsistencies generated by the lack of data context.
+
```
rajsite, line 8: Can you give a more concrete / very specific examples of  how the user of the wafer map would leverage these apis?

Like snippets of API usage and what it accomplishes.

In particular I'm trying to understand the component workflow and how the API is expected to be used by applications.

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -95,6 +95,11 @@ _The key elements of the component's public API surface:_
     -   topRight - ![Top Right Quadrant](./Resources/top_right.png)
     -   bottomRight - ![Bottom Right Quadrant](./Resources/bottom_right.png)
     -   `orientation` - represent the orientation of the notch on the wafer map outline (only visual). As only four static orientations are possible, it can be represented by an Enum with the following values: top, bottom, left, right.
+    -   `origin` - represents the origin coordinates of the dies grid for rendering the wafer map. Leaving the value `undefined` will set the value to the minimum corner of the bounding box of the dies coordinates. It will be an object of the type `PointCoordinates` with the following attributes:
+        -   x: number
+        -   y: number
```
rajsite, line 100: Could we have originX and originY properties?
Is there significant risk  / concern of the properties being set one at a time? I think with the batching implementation there really isn't a concern.

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
```diff
@@ -0,0 +1,28 @@
+# Nimble Wafer Map Grid Dimensions
+
+## Problem Statement
+
+The Wafer Map displays dies on a canvas inside a circular border based on an internal grid which is calculated based on the data input.
+This approach is limiting the performance of the wafer and the grid dimensions can be erroneous if the displayed data is just a subset of a larger data set.
+To mitigate these problems, we will reveal three new API inputs which will allow the users to set the grid dimensions themselves, thus saving processing time and correcting inconsistencies generated by the lack of data context.
```
rajsite, line 7: Is this implying that the circular border will be removed when these properties are set? Can you clarify what is the visual impact change as a result of this?

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
rajsite, line 8: Is this API effectively a way to programmatically zoom into a certain area of the wafer? Can you clarify what the interactive impact this has?

Can the user zoom out interactively and see more of the data? Can the zoom in interactively to a region of this subset of data?

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
```diff
@@ -0,0 +1,28 @@
...
+
+The new attributes and properties will be described in the component definition. They will be created with change listeners which will trigger the appropriate updates in the wafer.
+
+The inputs will be validated before use. If they are undefined the previous behavior of calculating the values based on the input data will be implemented.
+
+If part of the input data is missing, the wafer will not render, and a new `validity` API will be populated with the reason for the error, `invalidGridDimensions`.
+
+## Alternative Implementations / Designs
+
+Yes.
```
rajsite, line 24: The goal of this section is to list the alternatives considered / researched for posterity and to give some discussion about why the proposed solution of the spec was selected over the alternatives.

Can you list the alternatives?

One alternative I'm curious about is that based on the description this change sounds like the goal is to enable a form of optimization. I'm assuming that means some form of benchmark or some example data sets were used that signified the need to add new APIs specific to optimization. Can you capture that information here?

Given that information, I'm curious if alternatives were considered to improve the performance of the wafer map generally? What are the performance goals we need to achieve. It is quite possible that we can leverage features of the browser to make the wafer map itself faster for large calculations. I'd like to know what alternatives were considered.

packages/nimble-components/src/wafer-map/specs/README.md
jattasNI: Yeah, a nice benefit of that is that they could be attributes since they'd have primitive types, whereas an object can't be expressed as an attribute easily. Though that's not a huge deal since some of the existing configuration is already objects/arrays. But still better to try to limit that when possible.

change/@ni-nimble-components-4c6ed2fc-5a8c-4af1-bc8d-d145422fca42.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
mollykreis, line 2: It doesn't look like any code is actually changing in this PR, so the type should be none.
```suggestion
  "type": "none",
```

packages/nimble-components/src/wafer-map/specs/README.md
munteannatan: I am worried they can be mistaken for the actual data rows and columns and not just their number

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
munteannatan: no, the border will remain visible. this change will solve an existing issue regarding the way the data points are stretched to cover the whole canvas when they do not have a circular shape. This issue is popping up when the existing wafer values are filtered by value or by the wafer zone, thus warping the display

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
munteannatan: yes, essentially, it's a way to set the predefined zoom level of the data if you wish to see it that way. Right now, we are calculating the bounding box of the input data with the assumption that it is circular, and the missing data points are accidental. When the user is voluntary trying to display an incomplete subset for various purposes the data display gets warped. this change is a way to fix that.

the way you describe using it to zoom in and out programmatically is mostly a side effect. Do you recommend to discourage that usage by limiting the user input to being greater than the existing bounding box? or should we encourage it by filtering out the values outside the ones specified? Is there a different solution to our problem?

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
munteannatan: Sorry if the document made the performance gains seem the target of this change, I tried to clarify the goals of fixing display issues with incomplete sets of data more proeminent. The improved performance is just a side effect of removing a data parsing heuristic from the update loop.

We are still researching ways to improve the performance of the wafer. A recent project saw the implementation of WebGL to render the canvas with 60 fps for a data quantity of around 3 M points as an alternative for advanced graphics capable devices. But those projects are still in research phase as we are focusing on aligning different parts of our product to create a better synergy for the end users

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
munteannatan: I provided a visual representation of what is expected. if there should be more examples please let me know

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
rajsite: > A recent project saw the implementation of WebGL to render the canvas with 60 fps for a data quantity of around 3 M points as an alternative for advanced graphics capable devices.

Is that an example of a user requirement? It sounds like a demonstration of what a browser is capable of. But just to be clear are we expecting users to need to supply 3M data points to the nimble wafermap?

What are the functional performance requirements?

packages/nimble-components/src/wafer-map/specs/features/grid-dimension.md
rajsite: I appreciate you including the screenshot but I think I'm even more confused about what this feature is trying to accomplish. 

Like I don't know what I'm supposed to takeaway from this screenshot:
image

Is that a highlighted row? where is the actual wafer data? should the outer part of the wafer map still be shown?

You may need to reach out to help us understand what's going on clearer. If you can catch @jattasNI or I to help understand the goals and expected visual / user interaction impact of the changes that would help to make sure the details are captured in the spec.

rajsite: @munteannatan I would encourage you to try and reach out and resolve these questions sooner than later depending on your timelines.

As an FYI Jesse will be OOO for the next two weeks and we both have NI-Tech commitments. So be aware that response times will be longer as we approach NI-Tech and bandwidth is reduced.

jattasNI (commented):
FYI I'm going on vacation tomorrow. No need to wait on me to complete the PR.

packages/nimble-components/src/wafer-map/specs/README.md
rajsite: Another option. This represents additional metadata of the `dies` property configuration. It describes aspects of the shape of that property.

- `dies-origin-x`, `dies-origin-y`, `dies-rows`, `dies-cols`

The validation property `invalidGridDimensions` introduces the concept of die grid we could use:

- `die-grid-x`, `die-grid-y`, `die-grid-rows`, `die-grid-cols`
- `die-grid-origin-x`, `die-grid-origin-y`, `die-grid-rows`, `die-grid-cols`

I think I lean towards the (`die-grid-x`, `die-grid-y`, `die-grid-rows`, `die-grid-cols`) properties as the name grid lends itself towards the concept of rows and columns and clarifies the layout. Also the shared name prefix emphasizes the relationship between these properties.

rajsite (commented):
Looks good with the new images and discussions! Just need to figure out the naming

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -95,6 +95,10 @@ _The key elements of the component's public API surface:_
     -   topRight - ![Top Right Quadrant](./Resources/top_right.png)
     -   bottomRight - ![Bottom Right Quadrant](./Resources/bottom_right.png)
     -   `orientation` - represent the orientation of the notch on the wafer map outline (only visual). As only four static orientations are possible, it can be represented by an Enum with the following values: top, bottom, left, right.
```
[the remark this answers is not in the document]
munteannatan: is it clearer now or should i detail more?

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -226,8 +232,8 @@ N/A
 
 ### Performance
 
-As mentioned in the "Risks" section, the biggest challenge from the performance standpoint is the representation of large number of datapoints and interacting with these datapoints in the browser.
-Currently the only thing we would like to achieve is to measure (benchmark) the time it takes a `nimble-wafer-map` component to be loaded with various number of datapoints. Initially this could be a manual benchmark on a demo site hosted within the repository. Longer term we may automate and trend this by including benchmarking on the PR build.
+As mentioned in the "Risks" section, the biggest challenge from the performance standpoint is the representation of large number of data points and interacting with these data points in the browser.
```
DStavilaNI, line 235: This feature, as I understood it, (and also from the title of this PR) does not address performance issues. I am not sure it makes sense to refer here to performance if no performance improvements/changes are actually expected from this feature ... 

If the idea is to add bench-marking, as part of the storybook rendering, then the title should reflect that. However, again, if the intent is to do a "manual" benchmark using browser tools, I am not convinced it should be treated or otherwise included in the HLD ...

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -239,10 +245,11 @@ During these manual benchmark runs we would like to obtain more details on:
 
 Based on this information we can plan how we would like to approach performance improvements in the future and probably detect major bottlenecks.
 
-For the moment we don't have a clear understanding how long it will take to load a `nimble-wafer-map` component but we desire the following behavior:
+After doing various benchmarks, we are able to attest that the initial rendering performance of the `nimble-wafer-map` component is lower than 100 ms even for for wafer maps between 10k and 100k data points. These values may be different based on the system configuration.
+
+Another information of note is that although the initial render is is fast even for large data sets, the actual responsive behavior of the component, especially the zooming, is degrading visibly for a higher load.
```
DStavilaNI, line 250: "the initial render `is` is fast" - typo

packages/nimble-components/src/wafer-map/specs/Resources/bottom_left.png
DStavilaNI, line 1: I think that it would help to add an arrow to display the actual direction ... it is a bit hard to read from the current images.

image

change/@ni-nimble-components-4c6ed2fc-5a8c-4af1-bc8d-d145422fca42.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: moving to draft as it has local reviewer feedback to address, can move out of draft when ready for owner review

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -89,12 +89,16 @@ _The key elements of the component's public API surface:_
         -   x: number
         -   y: number
         -   value: string
-    -   `quadrant` - represents the orientation of the dies on the wafer map (the layout of the values on the dies). It can be represented by a const(as suggested [here](https://github.com/ni/nimble/blob/12a84ea7ad9103ab848aa2cd9f724e8853751a10/packages/nimble-components/docs/coding-conventions.md#use-const-objects-instead-of-typescript-enums)) with the following values:
+    -   `quadrant` - represents the direction of the two axes, X and Y, which are used for displaying the die grid on the wafer map. The four possible combinations are represented with the position of the quadrant respective to the specific axes directions . It can be represented by a const(as suggested [here](https://github.com/ni/nimble/blob/12a84ea7ad9103ab848aa2cd9f724e8853751a10/packages/nimble-components/docs/coding-conventions.md#use-const-objects-instead-of-typescript-enums)) with the following values:
```
atmgrifter00, line 92: I realize I may be late to the game for this particular API, but `quadrant` as a term, doesn't map intuitively to what this feature does. It seems to me all this is really doing is axis inversion, which I think could be represented much more understandably as two separate boolean attributes: `invertX` and `invertY` (or something even more specific, but the `invert` term is clear in its purpose).

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -89,12 +89,16 @@ _The key elements of the component's public API surface:_
...
     -   topLeft - ![Top Left Quadrant](./Resources/top_left.png)
     -   bottomLeft - ![Bottom Left Quadrant](./Resources/bottom_left.png)
     -   topRight - ![Top Right Quadrant](./Resources/top_right.png)
     -   bottomRight - ![Bottom Right Quadrant](./Resources/bottom_right.png)
-    -   `orientation` - represent the orientation of the notch on the wafer map outline (only visual). As only four static orientations are possible, it can be represented by an Enum with the following values: top, bottom, left, right.
+    -   `orientation` - represents the orientation of the notch on the wafer map outline. As only four static orientations are possible, it can be represented by an Enum with the following values: `top`, `bottom`, `left`, `right`. This value does not influence the die grid display in any measure, it affects only the circle outline of the wafer.
+    -   `dieGridOriginX` - represents the origin coordinates for the X axis of the dies grid for rendering the wafer map. Leaving the value `undefined` will set the value to the minimum X corner value of the bounding box of the dies coordinates.
```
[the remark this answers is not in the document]
atmgrifter00: Yeah, I think there is missing context here. I would presume these four properties are meant to allow a user to provide a particular window into the data provided in the `dies` property. Is that accurate? If so, I would further guess the value in these APIs it that it allows the view to have access to all of the data, but allow a front-end app to adjust the window into that data, avoiding the need for the view to retrieve the data from a possible expensive network request again.

Even if I'm way off-base, it would be valuable to describe the reasons these properties are being provided.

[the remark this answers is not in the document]
atmgrifter00: By my reading of the provided grid-dimensions document, `dieGridOriginX` and `dieGridOiginY` don't seem like they could be derived from `orientation`. I believe the way it's described, if all dies (provided to the `dies` property) had, for example, y-values between 3-5, you could provide a `dieGridOriginY` of 1, and a `dieGridRows` of 7, to get the dies to render in a constrained, centered vertical area in the circle of the wafer.

Is there a reason we can't make these attributes more semantically aligned? For instance, isn't another way to view `dieGridOriginY`, just `dieGridMinY`, and then `dieGridCols` just `dieGridMaxY`?

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -104,14 +108,16 @@ _The key elements of the component's public API surface:_
...
     -   `colorScaleMode` - represent an Enum value that determent if the colorScale is represent a continues gradient values (linear), or is set categorically (ordinal).
     -   `highlightedValues` - represent a list of strings of dies values that will be highlighted in the wafer map view
-    -   disabled - it's represented by a boolean value and refers to the state of the `nimble-wafer-map` component. If true, the component should be rendered dimmed out and no user interaction should be allowed.
+    -   `disabled` - it's represented by a boolean value and refers to the state of the `nimble-wafer-map` component. If true, the component should be rendered dimmed out and no user interaction should be allowed.
+    -   `validity` - readonly object of boolean values that represents the validity states that the wafer map's configuration can be in. The object's type is WaferMapValidity, and it contains the following boolean properties:
+        -   `invalidDieGridDimensions` : true when only one or two of the `dieGridOriginX`, `dieGridOriginY`, `dieGridRows` or `dieGridCols` is `undefined`.
```
atmgrifter00, line 113: I'm not sure this description is accurate for when this validity state would be `true`. Why would this be `true` based on any subset of these attributes being undefined, when there seems to be a well-defined meaning to what undefined results in for each property (which doesn't seem like it would be in conflict with its associated property that has a non-undefined value)?

Wouldn't we want this to be `true` only if `dieGridOriginX` was more than `dieGridCols` (and similar for the other properties)? I'm not sure we would say it's `true` even in the scenario that the provided values provided a coordinate space where nothing in the `dies` mapped to (but perhaps we could/should?).

If I'm not understanding something please let me know.

packages/nimble-components/src/wafer-map/specs/README.md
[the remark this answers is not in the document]
rajsite: Is your concrete proposal:
`dieGridMinX`, `dieGridMinY`, `dieGridMaxX`, `dieGridMaxY`
That sounds great to me!

I think we could even simplify at that point to:
`diesMinX`, `diesMinY`, `diesMaxX`, `diesMaxY`
I think that is nicely aligned with the dies X and Y properties they are constraining.

packages/nimble-components/src/wafer-map/specs/README.md
rajsite: For me thinking in invert-x or invert-y terms isn't mentally clearer. I think I'd have to toggle the properties until it looks right lol.

I think what would be clearer for me is defining it in terms of the origin location on a "screen". From that perspective I think the names top-left, top-right, bottom-left, bottom-right make sense. The term quadrant doesn't really though because we aren't like focusing on one quadrant of the coordinate system, we are re-mapping the coordinate system.

My proposal would be the name `dies-origin` (or `dies-grid-origin` to match the outcome of the other discussion) and then keep the enum values top-left, top-right, bottom-left, bottom-right, thoughts @atmgrifter00 ?

side note: It's kinda like dealing with game engine coordinates lol

image

packages/nimble-components/src/wafer-map/specs/README.md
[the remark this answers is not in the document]
munteannatan: Yes, the new API is meant to describe just the bounding box limits of the data input. Those are used when calculating the amount of vertical and horizontal space a single die will take based on the component dimensions.

They do not influence the orientation or the quadrant, the data will conform to those properties respectively when displayed.

The use-case we are going to apply the new API for is when we have partial data, but we want the display not to get stretched, just like having the full data. Another use case can be as mentioned reducing the bounding box if you want to 'zoom' in, but I do not know a situation when that would be needed.

I think it's a matter of choice in how we want to abstract a delimited area in API. In the database the actual dimensions are the Y-diameter, X-diameter, and the coordinates of the center. Here we chose to use the starting point and the width and height. Having the minimum and maximum coordinates is also a valid choice.Are there any

packages/nimble-components/src/wafer-map/specs/README.md
munteannatan: The slight performance improvement this change will create is related to the internal parsing of the data for finding the bounding box, which I think can be more apparent on larger data sets.

I added this section with some actual benchmark data because of a concern raised by @rajsite that the wafer documentation was not clear which data volume it was comfortable handling, where the current limit is and if we are planning on improving in the near to medium future.

packages/nimble-components/src/wafer-map/specs/README.md
munteannatan: I agree that the naming can create confusion and tbh it took me a while as well to get comfortable with the concept.

I think we may have a misalignment in our goals because we are not actual domain experts in the semiconductor area, so a lot of the jargon may sound strange to us. At the same time this component is targeted towards other developers who may or may not have the domain knowledge, but they may be tasked to implement features using our component.

packages/nimble-components/src/wafer-map/specs/README.md
munteannatan: I mentioned that when any subset of these attributes is undefined the validity flag will be raised because in that case the provided bounding box is obviously incomplete, so we are not going to use it to display the wafer. @rajsite mentioned that in this case is customary to not display anything and to raise the error flag for debugging

The `dieGridOriginX` value being greater than `dieGridCols` is not a concern because the columns are counted starting from the originX position, so if that position is 7 and we have 3 columns, the last column will be in the 9 position.

There still is the case you mentioned in which the provided bounding box does not cover any actual data although it will appear valid. Do you think this is something we should check internally, or should showing just the empty display align better with the nimble policy? If we are going to check if the input data aligns with the bounding box it may cost us some performance.

packages/nimble-components/src/wafer-map/specs/README.md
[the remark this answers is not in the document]
rajsite: @munteannatan We gave a concrete proposal but sounds like there is still open discussion. I'd recommend creating a meeting to resolve it quicker as the timezone difference will slow down the question-response cycle.

DStavilaNI (approved):

atmgrifter00 (approved):

rajsite (approved):

PR #1369, opened 2023-07-20
Spec update to include placeholder support for rich text editor

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Moving the placeholder text to `nimble-rich-text-editor`. Initially planned to have the placeholder textbox for adding a new comment in the `sl-new-comment` component. 

For more details, see the [discussion on comments HLR PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/523493#1689119376).

[Link to visual design](https://www.figma.com/file/7tTImJ8UueTcmgqcCWmD5q/Rich-Text-Editor-Visuals?type=design&node-id=7565-31314&mode=design&t=dCWzSlPK4OC9wdow-4) (Not finalized yet)

## 👩‍💻 Implementation

- Added a `placeholder` attribute in the spec for showing the placeholder text in the editor.
- Added `footer-hidden` attribute to hide the footer section.
- Updated `markdown` API to methods instead of an accessor and the decision on that change.
- Updated package info in the spec for the additional placeholder support in Tiptap and replaced `@tiptap/pm` with the actual required `Prosemirror` libraries.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/rich-text-editor/specs/README.md
+++ b/packages/nimble-components/src/rich-text-editor/specs/README.md
@@ -120,9 +120,9 @@ _Props/Attrs_
 
 _Methods_
 
--   `getMarkdown()` - this will serialize the content by extracting the Node from the editor and convert it into a markdown string using
+-   `getMarkdown()` - this will serialize the content by extracting the Node from the editor and return the converted markdown string
     [prosemirror-markdown serializer](https://github.com/ProseMirror/prosemirror-markdown/blob/9049cd1ec20540d70352f8a3e8736fb0d1f9ce1b/src/to_markdown.ts#L30).
--   `setMarkdown()` - this will parse the markdown string into a Node and load it back into the editor using
+-   `setMarkdown(value)` - this will parse the input markdown string into a Node and load it back into the editor using
     [prosemirror-markdown parser](https://github.com/ProseMirror/prosemirror-markdown/blob/9049cd1ec20540d70352f8a3e8736fb0d1f9ce1b/src/from_markdown.ts#L199).
 
 _Alternatives_


packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
...
 _Methods_
 
--   none
+-   `hideFooter()` - hides the footer section that contains all formatting options and the `footer-actions` slot.
+-   `clearContent()` - clears the content in the editor through [clearContent()](https://tiptap.dev/api/commands/clear-content) command in Tiptap.
+
+_Open Discussion_
+
+-   We are proposing a design that involves displaying only a placeholder text while initially hiding the footer section in the default view. Upon focusing on the editor, we will use
+    Tiptap's [focus event](https://tiptap.dev/api/events#focus) to show the footer section with all the formatting options. As per the
```
vikisekarNI, line 140:
```suggestion
-   We are proposing a design that involves displaying only placeholder text initially and footer section will be hidden. Upon focusing the editor, we will capture
    Tiptap's [focus event](https://tiptap.dev/api/events#focus) to show the footer section. As per the
```

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
...
+    UX for comments feature, we will revert back to the original state (i.e,. without the footer section) only if the user clicks the `cancel` or `ok` button. Therefore,
+    we came to this decision of exposing the above methods.
```
vikisekarNI, line 142:
```suggestion
    we came to this decision of exposing the above `hideFooter()` and `clearContent` methods.
```

vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
...
+-   `hideFooter()` - hides the footer section that contains all formatting options and the `footer-actions` slot.
+-   `clearContent()` - clears the content in the editor through [clearContent()](https://tiptap.dev/api/commands/clear-content) command in Tiptap.
+
+_Open Discussion_
+
+-   We are proposing a design that involves displaying only placeholder text initially and footer section will be hidden. Upon focusing the editor, we will capture
+    Tiptap's [focus event](https://tiptap.dev/api/events#focus) to show the footer section. As per the
+    UX for comments feature, we will revert back to the original state (i.e,. without the footer section) only if the user clicks the `cancel` or `ok` button. Therefore,
+    we came to this decision of exposing the above `hideFooter()` and `clearContent()` methods.
+-   There isn't a specific reason to expose two separate methods when considering the comments feature; they could be combined into a single method. However, we are uncertain about the
```
rbell517, line 143: Reset is a comment use-case specific operation. If we were using this as a description text editor it would still be reasonable to cancel the edit to reset to the previous contents and hide the footer until focus, but it would only be a clear if the previous contents were empty. These sound like decisions for the application, not the rich text editor. So I agree with the decision to keep these methods separate.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -110,7 +110,10 @@ _Props/Attrs_
 -   `isEmpty` - is a read-only property that indicates whether the editor is empty or not. This will be achieved through Tiptap's
     [isEmpty](https://tiptap.dev/api/editor#is-empty) API. The component and the Angular directive will have a getter method
     that can be used to bind it in the Angular application.
--   `fitToContent` - is a boolean attribute allows the text area to expand vertically to fit the content.
+-   `fit-to-content` - is a boolean attribute allows the text area to expand vertically to fit the content.
```
rbell517, line 113: If you're changing `fitToContent` to snake case would you want to do the same for `isEmpty`?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
...
+    we came to this decision of exposing the above `hideFooter()` and `clearContent()` methods.
+-   There isn't a specific reason to expose two separate methods when considering the comments feature; they could be combined into a single method. However, we are uncertain about the
+    appropriate name for this method. One idea we have is to name it `resetEditor()`, combining the functionality of hiding the footer and clearing the existing content in the editor. This method
+    would not modify the markdown value and so we are not sure `resetEditor()` is the appropriate name for it. So, the question is, can we consolidate both functionalities into a single method
+    and find a suitable name for it?
+-   The rationale behind not suggesting it as a property/attribute is that we don't see a compelling reason to expose both showing and hiding functionalities of the footer section to the client.
+    Instead, our proposal is to only expose a method that allows the client to hide the footer when necessary. The rich text editor can show the footer section exclusively when focused, while
```
rbell517, line 148: I guess the part I'm missing is what is the operation the application would do to show the footer in the first place? The API is unbalanced if you have a 'hide' but not a 'show'. A 'displayFooter' boolean attribute (or 'hideFooter' if we want the default value to be 'false') would work for both cases

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: Yea should consistently show as either the property (camelCase) or attribute (snake-case) form in the spec. Most specs use the attribute case. Make sure to do a quick search and align when in spec comments when changed.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -110,7 +110,10 @@ _Props/Attrs_
 -   `isEmpty` - is a read-only property that indicates whether the editor is empty or not. This will be achieved through Tiptap's
     [isEmpty](https://tiptap.dev/api/editor#is-empty) API. The component and the Angular directive will have a getter method
     that can be used to bind it in the Angular application.
--   `fitToContent` - is a boolean attribute allows the text area to expand vertically to fit the content.
+-   `fit-to-content` - is a boolean attribute allows the text area to expand vertically to fit the content.
+-   `placeholder` - is an attribute to include a placeholder text for the editor when it is empty. This can be achieved through
```
rajsite, line 114: Can you state explicitly is the placeholder treated as markdown or a plain string?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
 
 _Methods_
 
--   none
+-   `hideFooter()` - hides the footer section that contains all formatting options and the `footer-actions` slot.
+-   `clearContent()` - clears the content in the editor through [clearContent()](https://tiptap.dev/api/commands/clear-content) command in Tiptap.
```
rajsite, line 135: How is this different than assigning `el.markdown = ''` or `el.markdown = undefined;`?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -128,7 +131,22 @@ application's performance is enhanced as the operation is performed only once, t
 
 _Methods_
 
--   none
+-   `hideFooter()` - hides the footer section that contains all formatting options and the `footer-actions` slot.
```
rajsite, line 134: To hide parts of other controls we use a boolean attribute. See the conventions here: https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#attribute-common-name-patterns

Because showing the footer is probably a reasonable default for a rich text editor generally, then based on the conventions I would expect the footer to be visible by default and then setting the `footer-hidden` attribute to hide the footer.

One thing I'm not completely sure about, wonder if @jattasNI has any thoughts. Should the whole hide footer behavior be the built-in behavior of the rich text editor? Something something progressive disclosure you only see the footer toolbar when it has focus?
Maybe we need footer-visibility: 'auto' (default / undefined), 'visible', 'hidden'
- Auto: hides the footer when not focused, shows the footer when focused
- Visible: forces the footer to always be visible (app can set it to visible if it knows there is unsaved / dirty content)
- Hidden: forces the footer to not be visible (completeness? not sure of the use case)

packages/nimble-components/src/rich-text-editor/specs/README.md
jattasNI: I think writing down the following would help us come to a good API:
1. a list of the states that the component can be in
2. a list of the operations that can trigger a transition between those states
3. for each operation, a proposal of whether it can be detected by the rich-text-editor and if so whether the editor should automatically transition. Or or whether the client needs an API to tell the editor to transition.

Once we understand those states and transitions we can propose either attributes that represent the states or methods that represent the transitions (or both).

For example, I believe the states are something like this: (focused and not focused might be additional states)
1. no text, not showing footer
2. no text, showing footer
3. has text, showing footer

1 is the initial state

1->2 is triggered when the user focuses the editor. The editor can detect this so the rich text component could make this transition automatically. Should it?

2->1 requires clicking the cancel button which is provided by the client so the rich text editor needs to provide an API to make that transition

vikisekarNI: > 2->1 requires clicking the cancel button which is provided by the client so the rich text editor needs to provide an API to make that transition

For slotted button cancel operation, we definitely need an attribute `footer-hidden` (API) to allow clients to set the footer state for those button. Making an attribute API also makes it symmetric to show or hide the corresponding UI part. 

> 1->2 is triggered when the user focuses the editor. The editor can detect this so the rich text component could make this transition automatically. Should it?

For comments workflow, this transition applies only when adding a new comment through `sl-new-comment` component which always has the rich text editor available at top of the page. For editing/deleting a comment we quickly replace the editor with viewer on cancel / OK. 

For other use cases like description boxes as well this might likely apply based on the requirements. 

If it's too early for us to decide on a default footer state (hide initially and show in focus) for different use cases of rich text editor, can we defer this decision and enable clients to decide the visibility for their use cases by prioritizing implementation of `footer-hidden` API? Because this will allow clients to set this state based on their requirements.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Updated

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Initially, our plan was to clear the content without modifying the markdown value in the component. We considered this for scenarios like clicking a cancel button while adding a comment, where we preferred not to change the markdown value since it had not been parsed yet. Additionally, the reason for updating the markdown value was not entirely evident, and considering that Tiptap offers a method for clearing content, we decided to expose this API.

However, if we set the markdown value to an empty string, we can clear the content from the client. As this functionality can be achieved directly using the markdown and to avoid duplicating the same functionality in two different places, I have currently removed this method.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: As per the conventions, I have updated it as a `footer-hidden` boolean attribute instead of a method as you mentioned in the first part of the comment.

jattasNI (approved):

jattasNI:
Reminder that you should "re-request review" from reviewers once you've made updates or responded to their questions (circular arrows button next to each reviewer name). It looked like you'd resolved everything so I re-requested review from Milan and Richard.

mollykreis (approved):

rbell517 (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -396,10 +403,13 @@ This component is dependent on the [`tiptap`](https://tiptap.dev/) third party l
 library. For the currently supported features, we will include the following libraries that will be added to the package.json
 
 -   [@tiptap/core](https://www.npmjs.com/package/@tiptap/core)
--   [@tiptap/pm](https://www.npmjs.com/package/@tiptap/pm)
 -   [@tiptap/starter-kit](https://www.npmjs.com/package/@tiptap/starter-kit)
+-   [@tiptap/extension-placeholder](https://www.npmjs.com/package/@tiptap/extension-placeholder)
+-   [@tiptap/extension-link](https://www.npmjs.com/package/@tiptap/extension-link)
```
rajsite, line 408: Is the link package required as part of placeholder footer? Why is this being included here?

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: Good change! 👍

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: Just so the decision is spelled out, is the idea that the rich-text-editor itself won't define the UX of only showing the toolbar when it has focus and that is an application specific / usage specific workflow? Does that sound like the decision we are aligning on? @jattasNI @rbell517 

We can imagine that there will be enough variations of usage of rich text controls that we don't want to constrain / make consistent that part of the UX?

I can see that being a reasonable starting decision and the proliferation of usage of the rich-text-control is expected to be minimal, so changing the behavior may be reasonable in the future (for instance blazor apps desire the same UX for non-comment usages, maybe general description editing or something).

If that's the current decision then it should be clarified in the spec that the behavior described around footer hiding interactions is how an example application will use the `footer-hidden` property and not that the behavior of the rich text component itself:

https://github.com/ni/nimble/blob/91cb5fff3baf50cc76fb82248d47d1985b3b82ae/packages/nimble-components/src/rich-text-editor/specs/README.md?plain=1#L159-L161

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: @rajsite To reconfirm that this approach works in all cases, we did a second pass check of the actual use case of the `clearContent()` method API and found that it might be required in a few instances. 

For example, in comments, if the user adds a new comment and clicks the `Cancel` button, the editor will not be cleared since Angular's change detection is not triggered. This happens because the editor's `el.markdown` value remains unchanged (i.e., the markdown value does not update when the user doesn't click the `Ok` button. Only upon clicking `Ok`, the markdown value changes, not with every input). 

With the `clearContent()` method exposed, on click of the `Cancel` button, we can clear the content from the editor without changing the markdown value. So we would like to bring back this method as it gives a valid use case that may not be achieved by setting the `el.markdown = ''`. Please let us know your thoughts.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: I have updated the behavior of the footer hiding interactions and how the example application can utilize this API.

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: We ran into the same behavior with the nimble-table and the data property. Angular's change detection would say the value was not updated if we re-assigned the same data reference even if the data was changed.

I'm betting if in Angular you did not rely on data binding but instead used ViewChild and got the component reference and assigned directly to the directive it would work. But I can see how relying on data binding would cause issues with change detection.

I think this is a good indication that for the rich text editor we should use explicit getMarkdown and setMarkdown methods instead of relying on accessors and data binding.

We should not change the rich text viewer API, it makes sense as a property with the convenience of data binding. But it seems like for reasons similar to the table we should use methods for the rich text editor instead. I'd suuggest we change to getMarkdown and setMarkdown instead of adding a clearContent method.

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: If this PR is adding links support then I have several questions:

- SystemLink has some requirements around external linking behaviors used for custom properties in the table. Being able to restrict what URLs can do, etc. Does that apply here? What configuration is needed? Do we have the right APIs to configure that
- Are there visuals needed on URLs that link externally?
- Are the urls configured securely? (link rel noopener, no referrer etc)
- Do links open in a new window or the current window? Do links need Angular Router integration?
- Does this include the visuals of the link editor? If so can you include a screenshot of what that dialog looks like? Also I'm assuming that dialog will be defined in nimble itself as part of the rich text editor template, is that the implementation plan?
- If the link dialog is defined in nimble, what is the localization impact for the strings in the dialog? Do we need to add additional label provider? https://nimble.ni.dev/storybook/?path=/docs/tokens-label-providers--docs

I don't think the addition of link support as part of the spec was clear to other reviewers. The spec itself still says in the `_Additional features out of scope of this spec_` that links are not supported: "Support for adding hyperlinks to the existing words in the editor.", what exactly is being covered here?

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Closing this comment as the `clearContent()` method is removed as we can achieve it by setting the `markdown` value to an empty string.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: I have captured a few pieces of information related to these questions in the spec. I have noted some of them down here, but for a few, we don't have immediate answers. We still need to align with the necessary people to come to a decision.
1. As mentioned in the spec, we will currently support only the absolute URL format, which is one of the checks in the hyperlink support for the custom property table. The other validation involves checking if it is an external link, but we do not have an answer for that yet, and we are unsure about the configuration and API requirements to support it. Presently, the server-side does not perform any validation for links that are inserted into comments.
2. For the initial pass, we do not plan to display any visuals for external links.
3. Yes, Tiptap by default adds the rel attribute with noopener, noreferrer values also provide ways to add/override them.
4. Yes, links will open in a new window by default as they have the attribute `target="_blank"`.
5. Currently, we are not planning to support links that are hidden behind text. We will only support absolute URL links, and in markdown, the `autolink` format will be the corresponding representation of it. Therefore, there will be no dialogue to add links to the selected texts.
6. Since the dialog for the link is not currently supported, we will not have the localization.

> "Support for adding hyperlinks to the existing words in the editor.", what exactly is being covered here?

We will indeed support only absolute URLs where the href and the content are the same, and they will be considered as a link in the editor, as shown in the image below.
image

Please note that the information provided regarding the link support represents our current plans and is not a final decision. However, we would like to hear your thoughts on the same.

rajsite: I think the link stuff should be split into a separate spec update. I don't think the reviewers who have currently approved are clearly aware of the approval of link support.

My concerns / thoughts are:
- Any restrictions / limitations placed on the links that can be inserted beyond what CommonMark defines seem application specific. I am not convinced application specific restrictions should be enabled by default on the rich text control.
  - "Presently, the server-side does not perform any validation for links that are inserted into comments" if it will only be enforced by the client then it does not exist. I'd want to hear more justification about enforcing it client side only.
- I disagree with opening links in a new window by default. If that is a SystemLink comments requirement (which I would also push back on) then it should be a configurable option.
- The point still stands that the Globalization section of the spec needs to be updated. The linked PRs have merged so we need to localize the buttons in the toolbar.

Again to avoid blocking the current changes, I'd suggest a separate spec update for the link support.

vivinkrishna-ni: Thanks for sharing your thoughts! As you said, I have moved all the link related changes to a separate PR and it is still in progress as we need to align the link specific requirements with the necessary people. I will also be covering the Globalization part in the draft PR I have created for link support.

rajsite (approved):

PR #1370, opened 2023-07-20
Updates to contributing docs based on recent pain points

jattasNI:
# Pull Request

## 🤨 Rationale

Wanted to capture formal guidance for some common pieces of PR feedback so that we have something to point to when giving that feedback.

1. process for approving chromatic diffs
2. handling invalid component state
3. reducing owner burden

## 👩‍💻 Implementation

Changes in CONTRIBUTING.md. See PR comments for additional rationale.

## 🧪 Testing

N/A

## ✅ Checklist


CONTRIBUTING.md
```diff
@@ -97,7 +97,7 @@ If a fix for the vulnerability isn't available or if it isn't practical to uptak
 
 This repository uses [Chromatic](https://www.chromatic.com) to facilitate visual component review, and adds GitHub status checks to the build pipeline. The workflow is as follows:
 
-1. The `UI Tests` status check is designed to highlight any visual changes included in the changeset. The developer (that's you!) should review the `UI Tests` status check in Chromatic, and if all changes are intentional or expected, mark the components as **approved**.
+1. The `UI Tests` status check is designed to highlight any visual changes included in the changeset. The developer (that's you!) should review the `UI Tests` status check in Chromatic, and if all changes are intentional or expected, mark the components as **approved**. Please wait until you're nearly ready to submit to do this; if you approve a snapshot and then the appearance changes as a result of PR feedback, the diffs can become confusing.
```
jattasNI, line 100: I've noticed in the button appearance PRs that have gone through several rounds of feedback that it is hard to see a diff between main and the latest proposed changes if the diffs were approved from an intermediate state. I hope this will fix the problem.

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -263,6 +263,16 @@ With an attribute defined there are several ways to react to updates. To minimiz
 
     Some valid use cases are reflecting correct aria values based on the updated attribute or forwarding updates to child components.
 
+#### Only throw exceptions in production for Nimble logic errors
```
jattasNI, line 266: I welcome feedback on this. I tried to be assertive but may have overstepped.

I found one case that doesn't obey the guidelines as written: dialog and drawer both throw when the client calls `close` on something that's already closed (or open on something that's open). I'm open to feedback about whether we should change them to noop or update the guidelines to allow that class of exceptions. Maybe it's ok to throw from method calls but not from state updates?

CONTRIBUTING.md
jattasNI: I updated this section as discussed in team meeting. We can make further changes to the UI Review policy if needed now that we understand its purpose better.

CONTRIBUTING.md
```diff
@@ -128,9 +128,21 @@ You can also configure this task to execute via a keyboard shortcut by [configur
...
+
+1. When your code is ready to submit, create a [Draft pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests).
+2. Perform a self review of the changes. Ensure they follow conventions and that the status checks are passing.
+3. Add a local peer reviewer. If you are new to Nimble, also add the contact from the Nimble team who has been helping you get familiar with the repo.
+4. Once those reviewers have approved, mark the PR as "Ready for review". This will add owners to the review.
+5. Each file in a pull request requires the approval of at least one of its code owners (though in general for interesting changes we wait for everyone to review). Owners for different files are listed in [`CODEOWNERS`](/.github/CODEOWNERS).
+
+If a PR requires significant refactoring at any point in this process, please move it back to Draft and re-do the steps before adding owners again.
```
rajsite, line 141: Once owners (as defined by CODEOWNERS) are added I don't think can be removed even if moved back to draft. It's a slightly annoying minor nit as folks often accidentally create the normal PR and switch to draft so get the notifications anyway.

```suggestion
If a PR requires significant refactoring at any point in this process, please move it back to Draft and re-do the steps before exiting draft.
```

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -263,6 +263,16 @@ With an attribute defined there are several ways to react to updates. To minimiz
...
 
+#### Don't throw exceptions when a component is misconfigured
+
+Components should be robust to being configured in invalid ways and should typically not throw exceptions. This matches native element behavior and helps avoid situations where client code must be set component state in a specific order.
+
+Instead of throwing an exceptions, components should ignore invalid state and render in a predictable way. This could mean reverting to a default or empty state. This behavior should be covered by auto tests.
+
+Components can also consider exposing an API that checks the validity of the component configuration. Clients can use this to assert about the validity in their tests and to discover why a component is invalid when debugging. See the `nimble-table` for an example of this.
+
+It is acceptable to throw exceptions in production code and tests when a case gets hit that should be impossible: for example an invalid enum value.
```
rajsite, line 274: Components should be robust to being configured **via properties and attributes** in invalid ways

It is acceptable to report errors as exceptions when invoking methods as the user can synchronously respond to that behavior and be aware in call stacks. Compared to properties which may be set in varying orders and resolve asynchronously in the component (i.e. only observed after connect or in a batch at a later time).

(full comment may be too much detail but that's the message trying to throw around)

CONTRIBUTING.md
```diff
@@ -97,8 +97,8 @@ If a fix for the vulnerability isn't available or if it isn't practical to uptak
...
+1. The `UI Tests` status check is designed to highlight any visual changes included in the changeset. The developer (that's you!) should review the `UI Tests` status check in Chromatic, and if all changes are intentional or expected, mark the components as **approved**. If you approve an initial build and then make further changes to a snapshot, this check will show the difference between those revisions, not between main and the latest revision.
+2. The `UI Review` status check is designed to collect feedback from UX and visual designers. It shows the difference between the latest revision and main. Rather than blocking PR completion on this feedback, you can also approve this check by validating the story changes yourself. However, you should still demo your changes to relevant designers either in a team meeting or one-on-one. This can happen either before or after the PR completes, as long as designer feedback is addressed promptly. If you don't have access to a designer, please reach out to Nimble team members for help.
```
rajsite, line 101: May be worth just resummarizing this next week as a few folks were out. In part because I may have been confused and came away with the idea that we were going to want code owners to have to do the UI Review but what's written here sounds more preferable to me (in-terms of reducing the amount code owners need to double check).

rajsite (approved):

CONTRIBUTING.md
jattasNI: Yeah, my understanding of our plan was "no policy changes for now, but let's see if we want to change anything later now that we understand the two checks a little better". I think that was Fred's proposal too but we can re-iterate next week.

Issue #533, opened 2022-05-03
nimble-accordion Component


jattasNI:
## 😯 Problem to Solve

We want an accordion control which allows stacking several regions and allowing the user to collapse them to save space.

Here is an example in SystemLink Server's [package-viewer](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemsManagement/projects/systems/src/app/software/packages-viewer/packages-viewer.component.html), which uses mat-accordion. There is [a proposal](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/287607?_a=files&path=/Specs/Enterprise/Manage%20Notebook%20Execution%20Results/Specification.md) to bring this component to SLE for an app to manage notebook execution.
<img width="1331" alt="image" src="https://user-images.githubusercontent.com/10500124/166562762-0585baae-b5d2-4053-aa6a-4242549c61a0.png">

## 💁 Proposed Solution

- FAST component: https://explore.fast.design/components/fast-accordion
- Visual Design: #778 

## 📋 Tasks

- [ ] nimble-accordion interaction design issue
- [ ] nimble-accordion interaction design spec
- [ ] nimble-accordion visual design issue
- [ ] nimble-accordion visual design spec
- [ ] #784
- [ ] nimble-accordion custom element
- [ ] nimble-accordion unit tests
- [ ] nimble-accordion Storybook tests
- [ ] nimble-accordion Storybook docs
- [ ] nimble-accordion Angular wrapper
- [ ] nimble-accordion Blazor wrapper
- [ ] Add to Angular example app
- [ ] Update Component Status table in README and incubating status

jattasNI:
Another client [needs this](https://teams.microsoft.com/l/message/19:0c59ac4a4cc1438894cb42f4b21703f9@thread.skype/1664802527322?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=41626d4a-3f1f-49e2-abdc-f590be4a329d&parentMessageId=1664802527322&teamName=ASW%20SystemLink&channelName=UI%20Working%20Group&createdTime=1664802527322&allowXTenantAccess=false) for the Steps and PDV workflow:
<img width="575" alt="image" src="https://user-images.githubusercontent.com/10500124/193619709-63102f3d-8de1-4952-a5d8-72bb6e991607.png">

jattasNI:
Linking to #1150 since it tracks generating usage documentation for when to use the accordion.

[...]

PR #1371, opened 2023-07-20
Nimble accordion spec

### LINKED ISSUE #784: nimble-{name} spec

aidendk:
# Pull Request

## 🤨 Rationale

Fixes #784

## 👩‍💻 Implementation

Provides a spec for the nimble-accordion component

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,153 @@
...
+An Angular directive will be created for this component. The component will not have form association, so a `ControlValueAccessor` will not be created.
+
+### Blazor integration
+
+A blazor wrapper will be created for the component.
+
+### Additional requirements
+
+-   _User interaction: Do the FAST component's behaviors match the visual design spec? When they differ, which approach is preferable and why?_
+    -   Partially, extra functionality for Nimble components in inner accordion items will be required. Adding this functionality through the `content` section of the accordion will be preferrable because it doesn't require deviating from FAST's API.
```
m-akinc, line 129: What "extra functionality" is required? I'm not sure I understand.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,153 @@
...
+    <fast-accordion-item>
+        <fast-checkbox slot="start"></fast-checkbox>
+            <fast-number-field placeholder="number" slot="end"></fast-number-field>
+                Accordion four content
+            <div slot="heading">Accordion four</div>
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+### API
```
m-akinc, line 77: I would organize this section a bit differently. Just start with this list:
-   _Component Name:_
-   _Properties/Attributes:_ 
-   _Methods:_ 
-   _Events:_ 
-   _CSS Classes and Custom Properties that affect the component:_ 
-   _Slots:_

And have a copy of that list for each of the two different components (accoridion and accordion item). Then only go into detail about what deviates from the FAST API (otherwise say "Unchanged"). Much of what you have below is just re-describing the FAST API.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+[This screenshot](https://github.com/ni/nimble/issues/533#issue-1224634916) shows a possible integration of the accordion. It shows the usage of a dropdown select in the inner accordion items. The main conflict with this is that the FAST accordion does not have `start` or `end` slots for the inner accordion items, only for the headers. If this functionality is being used in SystemLink currently, it would most likely need to be included with the nimble-accordion to make switching to Nimble easier for these users.
```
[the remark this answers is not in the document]
aidendk: Yeah, I might be looking to far into [this example](https://github.com/ni/nimble/issues/533#issue-1224634916) and thinking that we have to implement the inner component functionality as well.

packages/nimble-components/src/accordion/specs/README.md
aidendk: That was mentioning the implementation of divs for component functionality in the accordion children but based on your [comment above](https://github.com/ni/nimble/pull/1371#discussion_r1269942724) it's most likely unnecessary, so I'll change this to No Additional Requirements.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,135 @@
...
+                Accordion four content
+            <div slot="heading">Accordion four</div>
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+### API
+
+[FAST accordion API](https://github.com/microsoft/fast/blob/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion/accordion.spec.md)
+
```
m-akinc, line 80: Can you give these component APIs some visual separation by adding subheaders for "Accordion" and "Accodion Item"?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,135 @@
...
+-   _Component Name:_ `nimble-accordion`
+-   _Properties/Attributes:_ appearance (`block`, `outline`, `ghost`), `error-visible` (boolean)
```
m-akinc: Could you format these into a list for readability? I.e.

-    _Properties/Attributes:_ 
     -    `appearance` - `block`, `outline`, `ghost`
     -    `error-visible` - boolean

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,135 @@
...
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_ Unchanged
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ `error-visible` and accordion appearances (`block`, `outline`, `ghost`)
```
m-akinc, line 92: I don't think these actually affect the accordion item. I would move the following paragraphs to after the API list of the accordion, i.e. right before the heading I suggested you add for Accordion Item.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+    The Figma design also includes a state that is most likely `error`, as it has a red border and red exclamation mark icon. For this state, the styling of the accordion header button could be altered to have a red border. This could be controlled with a boolean attribute such as `error-visible`, which would have a default of "" in the accordion attributes- when necessary, "error-visible" would be added to the accordion attributes and this would make the accordion switch to the red styling. Adding the error icon could be done by putting the icon in the `end` slot of the accordion heading.
+
+    A css class for the nimble-accordion-item can be used to get a green border (or red for `error-visible`) around the entire accordion (parent and children of accordion).
```
m-akinc, line 98: I don't think we'd put a class like this on the accordion item (if anything, it would be on the accordion), but regardless, this gets into a level of implementation detail that probably doesn't need to be in this spec.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+A blazor wrapper will be created for the component.
+
+### Additional requirements
+
+-   _User interaction: Do the FAST component's behaviors match the visual design spec? When they differ, which approach is preferable and why?_
+    -   No Additional Requirements
+-   _Styling: Does FAST provide APIs to achieve the styling in the visual design spec?_
+    -   Partially, extra styling will be needed for the `error-visible` boolean and `Block` / `Outline` / `Ghost` appearances.
+-   _Testing: Is FAST's coverage sufficient? Should we write any tests beyond Chromatic visual tests?_
```
m-akinc, line 116: I think `error-visible` and `appearance` values should be covered by Chromatic visual tests.

m-akinc (approved):

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Design
+
+Simple accordion group
+
+```
+<fast-accordion>
+    <fast-accordion-item>
+        <span slot="heading">Panel one</span>
+        <span slot="icon">^</span>
```
rajsite, line 23: I don't see an `icon` slot in the existing fast templates or specs. So this is a new api. What is it for?
Remember slots are for nimble users to add customization, what would this slots be customizing that we would want to allow customization for?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+        Panel three content
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+Complex accordion group with components in slots and single expand-mode
+
+```
+<fast-accordion expand-mode="single">
```
rajsite, line 42: like above, make sure to frame this in terms of the nimble component names you are proposing

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+            <fast-number-field placeholder="number" slot="end"></fast-number-field>
+                Accordion four content
+            <div slot="heading">Accordion four</div>
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+### API
+
+#### FAST Accordion
```
rajsite, line 79: The proposal is for a Nimble Accordion, make sure to update the other location below

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
+    -   `error-visible` - boolean
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_
+    -   `appearance` - `block`, `outline`, `ghost`
+    -   `error-visible` - boolean
```
rajsite, line 102: Do you actually expect to respond to css classes being set? Can you explain these?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+-   _Slots:_ Unchanged
+
+The Figma design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` buttons. The nimble-button pattern with these styles seems to fit the styling of the accordion, so these styles can be used. The attributes of `block`, `outline`, and `ghost` will most likely be implemented with the name `appearance`. A few styling elements like the background color change on click that is used in the button will have to be removed.
```
rajsite, line 105: Something to keep in mind for this implementation direction. If you end up in a position where you include the shared button styles and then have to override a significant number of them that would be a big drawback for maintainability. In order to understand the styles you now only have to look at the button styles but then keep in your head how they need to be overridden.

What I want to avoid is if you open devtools and inspect the styles and see tons of constantly overridden styles then it is probably not a good fit for the styles to be shared.

An extreme (somewhat misleading) example of styles being overridden excessively:
image

It's not that overriding styles is bad. Overriding styles is expected for things like state changes from the default, i.e. override the border on hover. So I'm not discouraging you from finding good patterns to share styles, it's been very valuable for buttons, etc, just making sure you are aware of what a concern may be in finding the right styles to share.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+The Figma design also includes a state that is most likely `error`, as it has a red border and red exclamation mark icon. For this state, the styling of the accordion header button could be altered to have a red border. This could be controlled with a boolean attribute such as `error-visible`, which would have a default of "" in the accordion attributes- when necessary, "error-visible" would be added to the accordion attributes and this would make the accordion switch to the red styling.
```
rajsite, line 107: Why the passive language "most likely error" or "could be controlled by an attribute"? The spec is a proposal of what is intended to be implemented. Go ahead and phrase it so there is less wiggle room and it's clear what the outcome of the spec being implemented should be.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+A css class for the nimble-accordion-item can be used to get a green border (or red for `error-visible`) around the entire accordion (parent and children of accordion).
```
rajsite, line 109: Can you expand on exactly what the api of this would look like?
Sound like you are saying `<nimble-accordion-item class="make-it-green">`. If that is the proposal make sure to spell out what the proposed class name is etc.

In this case, that seems like the wrong direction. We have a pattern (you may be familiar with 😉) of having a appearance variant that uses the brand accent color. 

But but in this case, are you sure having a green / accent mode is what the spec is proposing? I looked briefly at the spec and it seems like the different green borders are associated with hover states / keyboard focus states. They do not seem to be different modes of the control. Is that what you are seeing as well?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+An Angular directive will be created for this component. The component will not have form association, so a `ControlValueAccessor` will not be created.
+
+### Blazor integration
+
+A blazor wrapper will be created for the component.
+
+### Additional requirements
+
+-   _User interaction: Do the FAST component's behaviors match the visual design spec? When they differ, which approach is preferable and why?_
```
rajsite, line 121: How does FAST handle nesting accordions which the figma designs seem to show? I don't see examples.

And then to double check because I don't see examples in the FAST repo, are there ARIA guidelines around keyboard interactions for accordions? How do the ARIA guidelines handle nested accordians? Is that how FAST handles them?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+</fast-accordion>
+```
+
+Complex accordion group with components in slots and single expand-mode
+
+```
+<fast-accordion expand-mode="single">
+    <fast-accordion-item>
+        <fast-checkbox slot="start"></fast-checkbox>
+            <fast-select slot="end">
```
atmgrifter00, line 45: I think a lot of the indentation in this markup is misleading/wrong. As an example, the above `<fast-select...>` (sic) component should be a sibling of the `<fast-checkbox...>` component, not a child of it (which is what the indentation implies).

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,144 @@
...
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+#### FAST Accordion Item
+
+[FAST accordion-item API](https://github.com/microsoft/fast/tree/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item)
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
```
atmgrifter00, line 96: Is there a reason to explicitly provide `appearance` on the items themselves? Why not provide it on the `nimble-accordion` component? I wouldn't expect that we want items to have different appearances within a single accordion component.

packages/nimble-components/src/accordion/specs/README.md
aidendk: Sorry, I pulled the `slot="icon"` part from this section in [fast's accordion spec](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/accordion/accordion.spec.md#implementation). Not sure why they had `icon` as a slot since they don't list it [here](https://github.com/microsoft/fast/tree/archives/fast-element-1/packages/web-components/fast-foundation/src/accordion#css-parts-1), I should have caught that!

I'll use one of the examples from the fast component explorer instead and rename the tag names to `nimble-accordion` / `nimble-accordion-item` 👍

change/@ni-nimble-components-d6d8517b-312c-467f-8b69-bdaa581e67fa.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: nit: PR description refers to anchor, not accordion.

packages/nimble-components/src/accordion/specs/README.md
aidendk: In their [spec](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/accordion/accordion.spec.md), they said that "Optional support for nested accordions. Ideally this behavior will "just work" and no special behaviors will need to be added to enable/support this. The expectation here should be that an accordion takes content and whatever content is inside the accordion panel respects its own interaction model." FAST didn't create any examples of this in their component explorer, but I just copied and pasted an accordion inside of the parent component and it works fine.

I attached a clip where I showed the functionality of FAST's accordion and the accordion that [W3 uses as an example](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/), the styling isn't great since I tweaked both to have nested accordions, but from what I can tell, both handle them the same way.

https://github.com/ni/nimble/assets/103057880/29e48613-c5b8-4a98-b0e6-99d4105dfad2

These are [aria's guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) for keyboard interactions: enter or space for actions, tab and shift + tab for moving up and down, arrow keys for moving up and down between headers, and home / end (optional) for moving to the first and last accordion header, respectively. I tested all of these in the FAST component explorer, and they're functional.

packages/nimble-components/src/accordion/specs/README.md
aidendk: Yeah, that's a good point. I'll move those attributes to the nimble-accordion component's `Properties/Attributes` section 👍

packages/nimble-components/src/accordion/specs/README.md
aidendk: No, I think I misinterpreted this section. Similar to [this comment](https://github.com/ni/nimble/pull/1371#discussion_r1272326316), I intended to specify that the styling of these attributes would be controlled by the attribute values, not the css classes. I'll delete this portion from the `CSS Classes and Custom Properties...` section. 👍

packages/nimble-components/src/accordion/specs/README.md
aidendk: I talked with Mert about this and took another look at the Figma spec, and we decided to not use the button pattern styles. This should solve this issue of overriding too many styles, I'll update the spec to reflect this decision.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,140 @@
...
+<nimble-accordion expand-mode="single">
+    <nimble-accordion-item>
+        <nimble-checkbox slot="start"></nimble-checkbox>
+        <div slot="heading">Accordion one</div>
+        <nimble-select slot="end">
+            <nimble-option value="1">Option 1</nimble-option>
+            <nimble-option value="2">Option 2</nimble-option>
+            <nimble-option value="3">Option 3</nimble-option>
+        </nimble-select>
+            Accordion one content
```
atmgrifter00, line 48: I might just remove the various "Accordion <x> content", as I don't think it really helps illustrate the API/functionality of the accordion. It's just string literal content mixed with other component content, but its presence is more confusing than anything (I get that this is how the FAST example illustrates the API, but they don't provide other content like we are here).

m-akinc (approved):

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+[Nimble issue #533: nimble-accordion Component](https://github.com/ni/nimble/issues/533)
+
+[Visual Design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-85131&mode=design&t=DxDRlqT7MeCPLOxi-0)
+
+---
+
+## Design
+
+Simple accordion group
```
jattasNI, line 17: nit: some suggestions to improve these examples:
1. I don't think a client would have a use case for an accordion item with no content in the default slot of an item: they'd probably just remove the item in that case. So I'd prefer if the examples didn't emphasize that case so much: right now only 2 of the items have any content in the default slot.
2. When the default slot has content, it's likely to be more complex than just a single direct child. For example it could be a `div` that contains lots of controls, like in a settings pane. 

Could you update the examples to better reflect these use cases?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+-   _Security: Any requirements for security?_
+    -   No Additional Requirements
+
+---
+
+## Open Issues
+
+Should we only have arrow icons for open / closed indicators on the accordion, or should other visuals like +/- be implemented as well?
+
+Brandon's Figma spec has designs for two accordion item heights, 24px and 32px. In previous cases like the nimble button, we've only implemented one height (in that case only the 32px button was implemented). Do we want to continue doing this, and if so, which height should be implemented?
```
jattasNI, line 135: My vote is to stick with a single height. I assume we'd match what we did with the button and implement the larger one, but we could confirm with Brandon when he gets back.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+-   _Performance: does the FAST component meet Nimble's performance requirements?_
+    -   No Additional Requirements
+-   _Security: Any requirements for security?_
+    -   No Additional Requirements
+
+---
+
+## Open Issues
+
+Should we only have arrow icons for open / closed indicators on the accordion, or should other visuals like +/- be implemented as well?
```
jattasNI, line 133: I'm curious where the idea came from to add other visuals? I don't see them in the design spec. If it's just something we think someone might want later, I'd vote to remove it from the spec for now and add it on-demand if someone ever asks for it.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+Brandon's Figma spec has designs for two accordion item heights, 24px and 32px. In previous cases like the nimble button, we've only implemented one height (in that case only the 32px button was implemented). Do we want to continue doing this, and if so, which height should be implemented?
+
+Brandon's Figma spec also only shows designs for one accordion item at a time. How should accordion items in each appearance type be separated from each other (ex. a line between accordion items, transparent space between accordion items, etc.)?
```
jattasNI, line 137: I agree this needs to be resolved before implementation, and thanks for mentioning it in the spec. But I don't think it needs to be resolved before submitting the spec (unless you think a different appearance would drastically affect the structure of the component and whether we could leverage the FAST template). So I'd suggest removing this from Open Issues and instead making it a note where you link to the visual design.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+The Figma design also includes an `error` state. This will be controlled with the boolean attribute `error-visible`, which has a default attribute value of "". When needed, "error-visible" would be added to the accordion attributes, changing the color of the accordion item border color to red. This will be implemented in the accordion styling through conditional css styles based on `error-visible`'s value (ex. :host([error-visible]) will have styling for when `error-visible` is true).
```
jattasNI, line 98: It's interesting that the design shows a way to indicate that there's an error, but no way to indicate what that error is. Assuming that's not an oversight, I wonder if the idea is that an accordion item would show an error if-and-only-if something in its content is in an error state, and that the content would be responsible for explaining what the error is.

If that's the case, we should plan to add usage guidance explaining that in the storybook documentation for `error-visible`. An alternative could be that the accordion detects that by scanning its children for error states and enters its own error state automatically, but I'd vote that's overkill. 

When he's back, can you double check with Brandon about this?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+-   _Component Name:_ `nimble-accordion`
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+The Figma design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. Because of the difference in border functionality and possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value (ex. :host([appearance='outline']) will have styling for the outline button).
```
jattasNI, line 82: >shared styles will not be used

Could you clarify this? Are you saying that the accordion won't share styles with other components or that the accordion won't share styles across the three appearances?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+Documentation will be added to advise against using multiple Nimble Accordions with different appearances next to each other.
```
jattasNI, line 84: It'd be good to check with Brandon if he has guidance about when to use each `appearance`. Is it unique guidance for the accordion or can we re-use some guidance from the button? (This needn't block this PR but you could add a note mentioning it)

packages/nimble-components/src/accordion/specs/README.md
jattasNI: That accessibility summary would be good to capture in the corresponding section below.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+[FAST accordion-item API](https://github.com/microsoft/fast/tree/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item)
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
```
jattasNI, line 96: Do we have use cases in mind for the start and end slots? Anything in the visual designs? If not, we should follow our new [Hide unused slots](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#hide-unused-slots) guideline that was introduced in #1353.

If we do that then you could simplify the examples not to make use of the slots.

packages/nimble-components/src/accordion/specs/README.md
aidendk: That's what Mert and I assumed the functionality of the error state would be, and I'll add a note in the spec to make sure we create that guidance in the docs. Sure, I'll let you know what he says!

packages/nimble-components/src/accordion/specs/README.md
aidendk: That's what I put in the spec, but I still can't decide if using the button styling patterns is worth it or not. The main conflict is that it's hard to know how much overriding we'll have to do if we use the button styling patterns, but at the same time it could save us from a lot of extra css since it mostly follows the same appearance states as the accordion. Maybe we can leave this as an open issue in the spec and I can first try using the button patterns to see if it works, and if not, switch to creating new styles for the accordion?

packages/nimble-components/src/accordion/specs/README.md
aidendk: Sure, I'll add that along with the other comments in the visual design section and I'll ask him when he gets back 👍

packages/nimble-components/src/accordion/specs/README.md
aidendk: The visual designs don't show any use case for the start slot, but the end slot will be used for displaying the error icon when in the `error-visible` state.

packages/nimble-components/src/accordion/specs/README.md
jattasNI: I'm ok with leaving it unspecified in the spec.

My guess is that we shouldn't try too hard to share the styles with the buttons. We have several other components that support block and outline appearance (text field, text area, select) and I don't think any of them use the button styles (they do use the appearance behavior infrastructure). I'd suggest starting with the same pattern for the accordion.

aidendk: Sounds good, I included it in the open issues section and advised that we leave it open until we spend some time trying to implement it.

aidendk:
Just remembered that the accordion doesn't use .control for styling, it uses .region and .heading, so it's probably not worth going through the trouble of getting those to work with the button shared styles (if even possible). I'll update the docs to reflect this.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Design
+
+Accordion group, appearance ghost with an error-visible accordion item
+
+```
+<nimble-accordion appearance="ghost">
+    <nimble-accordion-item>
+        <span slot="heading">Accordion one</span>
+            <div slot="default">
```
mollykreis, line 31: I'm assuming this `div` is going in the default slot, not a slot with the name `"default"`? If so, you don't need to specify `slot="default"`. Simply having the content be a child of the `nimble-accordion-item` with not slot specified will do what you want.

Also, since this `div` isn't a child of the `span` on the line above, it should be out-dented once.

Same for the other instances below.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+    <nimble-accordion-item expanded>
+        <span slot="heading">Accordion two</span>
+            <div slot="default">
+                <nimble-checkbox></nimble-checkbox>
+                <span>Accordion two content</span>
+            </div>
+    </nimble-accordion-item>
+    <nimble-accordion-item error-visible>
+        <span slot="heading">Accordion three</span>
+        <span slot="end"><nimble-icon-exclamation-mark></nimble-icon-exclamation-mark></span>
```
mollykreis, line 45: I don't think we should have the client slot the exclamation icon here. Instead, I think the component should automatically add it when `error-visible` is set.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+[FAST accordion-item API](https://github.com/microsoft/fast/tree/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item)
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ The `start` slot will not be used based on the visual designs, but the `end` slot will be used to display the error icon in the `error-visible` accordion-item state.
```
mollykreis, line 130: It would probably be good to include `collapsed-icon` and `expanded-icon` here. I'm guessing the `nimble-accordion-item` will set the content of those slots and it would be unsupported to override those icons?

packages/nimble-components/src/accordion/specs/README.md
rajsite, line 31: This is a little confusing in the way the FAST spec is written but there is not a slot named default (`<slot name="default"></slot>`) there is a *default* slot ([`<slot></slot>`](https://github.com/microsoft/fast/blob/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts#L52))

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+The Figma design also includes an `error` state. This will be controlled with the boolean attribute `error-visible`, which has a default attribute value of "". When needed, "error-visible" would be added to the accordion attributes, changing the color of the accordion item border color to red. This will be implemented in the accordion styling through conditional css styles based on `error-visible`'s value (ex. :host([error-visible]) will have styling for when `error-visible` is true).
+
+Usage guidance for the `error-visible` state will be created, as its usage may be unclear.
+
+### Angular integration
+
+An Angular directive will be created for this component. The component will not have form association, so a `ControlValueAccessor` will not be created.
```
mollykreis, line 138: nit: I assume directives will be created for both components.
```suggestion
Angular directives will be created for these components. Neither will have form association, so `ControlValueAccessor`s will not be created.
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Blazor integration
+
+A blazor wrapper will be created for the component.
```
mollykreis, line 142: nit: Blazor wrappers for both components
```suggestion
Blazor wrappers will be created for both components.
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+                <nimble-checkbox></nimble-checkbox>
+                <span>Accordion one content</span>
+            </div>
+    </nimble-accordion-item>
+    <nimble-accordion-item expanded>
+        <span slot="heading">Accordion two</span>
+            <div slot="default">
+                <nimble-checkbox></nimble-checkbox>
+                <span>Accordion two content</span>
+            </div>
```
rajsite, line 41: Can you double check the indentation of the snippets? This makes it look like the stuff in the default slot is a child of the header but they are siblings.

You can use something like [prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeKBLAtgIwDZwC0AhmJAE4AmG0ABMQA4NzHnFRhwC8AOiPhDABrPgD4eUWlNrps+IqQrVohDPCzjJ07amoA3WgGc8EGLxAALFtSgBzMQEEyEKjUnQ4qAPT7N2nZi4BIRgVsI4EAAeot6B8iFhQhHREv7STkputB60kLAIMKnSsXLBii7KUKrqfsVxZc6uKmpwGkVpuhgGxqbmVsQ29iCiGRVZMADuEN6+7f6yQUQ4AK4wMNCiACpwkTC0AEKr61Ali4Qraxtz2iONldlQcLnQ8LBzp-HlTVUtbVpSC0+dzc1VatQ6+iMJjMfH6g0cwLoMAs5Dgnh8XXB83qRFCcHCURiXhxCXxSUJ1zqpQUiKgWLSAJJX0qoL+DPZMkhPRhlmsGDsYgAcnBDPBKLRRt9aAB5R4zTGUjm0YWiuDiyX3WVPADCLwKivmxOpJFprPpDMBDUyzRqBvZnW60L6fIFwxVYoltNomym8r05qV7rVnutkh9EFouvyb3+HI+VrGNrBBvjNNDWNTJtDZveTNNvwDnK6UN6sJdQ1uodoADMIMtyH7C5alkcNttdgdWycjWcLsdCxqsjk8q9CrHM8yQQWJBPaZoQAAaEAQBgwNyGZCgVjkCATAAKrAQG5QxDwE2IAE8N0ucGxhHAYABlYhYOAAGX5cGQ1dPhjgN7vIQH0fBhSH5WxkBgchln-EA-ywDBIOg2DDHAggAEVllML8kB-PA-yXAArQxIkfNC4Ew7Dv1-WCAEcsPgPcdwYY8QGIQxCEeNU1UXEAoOIDA8HA3UsCwYhkDYvA8F41C7AIBw1nIDALjgPc4HID9Hmo-DYIsGAsDwAB1CwWkMUDOEfI81C6NQLwksBDGvEA9BggBJKBKAKR8wCU1cHA8x8YAvAhtIIkAGB3P9DLYBgJIikV1L0L8l35P9yBgJjiFsMTQtg0DyDSiS9IM3iIv5GBDIwShkWQAAOAAGJdUXojBUUy7LxNwmilxgYgcEq6qLGQAAWJdlj-TY+uPPCwtaHA1U8yg33YWxliyuAADEXDEtZwIk4hVggEAAF9jqAA) to help write the snippets. It'll help find other problems like mismatched tags, etc.

packages/nimble-components/src/accordion/specs/README.md
rajsite, line 31: This is a little confusing in the way the FAST spec is written but there is not a slot named default (`<slot name="default"></slot>`) there is a *default* slot ([`<slot></slot>`](https://github.com/microsoft/fast/blob/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts#L52)).

Think you can reduce the verbosity in these example quuite a bit by relying on the default slot behavior:

```html
    <nimble-accordion-item>
        <div slot="heading">Accordion one</div>
        <nimble-checkbox></nimble-checkbox>
        Accordion one content
    </nimble-accordion-item>
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+The Figma design does not have a design for the "error" focusVisible state. This might be confusing, as keyboard users who tab to the error accordion will suddenly see the red border and exclamation mark icon disappear in the focusVisible state.
+
+When should each `appearance` type be used? Will there be specific guidance on this, and if so, will it be similar to the guidance for other Nimble components that use the same `appearance` types (ex. the Nimble-Button)?
+
+## Design
+
+Accordion group, appearance ghost with an error-visible accordion item
+
+```
```
rajsite, line 27: use the html code fence so there is syntax highlighting
```suggestion
```html
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+                <nimble-checkbox></nimble-checkbox>
+                <span>Accordion three content</span>
+            </div>
+    </nimble-accordion-item>
+</nimble-accordion>
+```
+
+Accordion group, appearance block with a nested accordion
+
+```
```
rajsite, line 56: use the html code fence so there is syntax highlighting
```suggestion
```html
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Background
+
+[Nimble issue #533: nimble-accordion Component](https://github.com/ni/nimble/issues/533)
+
+[Visual Design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-85131&mode=design&t=DxDRlqT7MeCPLOxi-0)
+
+Visual Design Comments:
+
+Brandon's Figma spec has designs for two accordion item heights, 24px and 32px. In previous cases like the nimble button, we've only implemented one height (in that case only the 32px button was implemented). Do we want to continue doing this, and if so, which height should be implemented?
```
rajsite, line 15: I expect it is uncontroversial to start with the 32px implementation. As mentioned we haven't focused on 24px implementations and the examples in the linked issue in the drawers both look like 32px use cases.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+-   _Component Name:_ `nimble-accordion`
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+The Figma design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. These designs are similar to those of the nimble-button and its `appearance` types, but the accordion does not use the `.control` css class, it uses `.region` and `.heading`. Because of this, differences in border functionality, and the possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion and accordion-item will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value (ex. :host([appearance='outline']) will have styling for the outline button).
```
rajsite, line 116: use the code escape for code snippets
```suggestion
The Figma design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. These designs are similar to those of the nimble-button and its `appearance` types, but the accordion does not use the `.control` css class, it uses `.region` and `.heading`. Because of this, differences in border functionality, and the possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion and accordion-item will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value `(ex. :host([appearance='outline'])` will have styling for the outline button).
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Documentation will be added to advise against using multiple Nimble Accordions with different appearances next to each other.
+
+#### Nimble Accordion Item
+
+[FAST accordion-item API](https://github.com/microsoft/fast/tree/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item)
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
```
rajsite, line 126: individual accordion items will be able to be configured in an error state?
Maybe I'm missing it but does the figma document have any examples of what an accordion looks like with multiple accordion items? Like it's not clear to me if the different items would share a border or not (presumably not) and what is the spacing between them.

Also curious how the nested accordions work for all the borders. Do borders only render on the inner most accordions?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,170 @@
...
+-   _Performance: does the FAST component meet Nimble's performance requirements?_
+    -   No Additional Requirements
+-   _Security: Any requirements for security?_
+    -   No Additional Requirements
+
+---
+
+## Open Issues
+
+Using expand-mode="single" when nesting accordions causes issues- when trying to open nested accordions, the parent accordion closes, which makes sense as the functionality of the attribute is to only allow one accordion open at a time. Unless a reasonable fix can be found, we should advise against using this attribute with nested accordions in the docs.
```
rajsite, line 170: Does this make sense or is this a bug we need to handle?

rajsite: Seems weird for the child accordion's accordion-item to impact the parent accordion's accordion-items

packages/nimble-components/src/accordion/specs/README.md
aidendk: That's how I interpreted Brandon's spec, but he doesn't have any examples of multiple accordion items so it's hard to be certain. I'll ask Brandon about this and the border sharing, I agree it's unclear how spacing would work for different appearances. 

Here's an example from FAST's component explorer where I added nested accordions- It seems like the borders render on every accordion.
fastNestedAccordionBorder

packages/nimble-components/src/accordion/specs/README.md
aidendk: https://github.com/ni/nimble/assets/103057880/0a04c02d-1e2f-4148-9418-50c8902bb95c

Here's a video of the interactions with nested accordions in the single expand mode ^
In the first half of the video, I'm trying to close the opened accordion item through click and focusVisible, but when single expand mode is enabled, you can't close the accordion item you've opened by clicking on it, you have to click on a different accordion item. At least one accordion has to be open at all times, which could be confusing to users as the expected functionality would be that accordion items are closed until you open them and can be closed by clicking on that specific accordion item.

I look into the nested accordions in the second half of the video, where I show that once you open a child accordion, the parent accordion closes- when you open the parent accordion again, the previously selected child accordion is open, but when clicking on it, it doesn't close. Also, when opening the second child accordion, the first child accordion stays open, which goes against the functionality of single expand mode.

Regardless of the issues with the nested accordion, I don't think that single expand mode is very beneficial- having one accordion item open before users click on it, and then not allowing them to close it unless they click on a different accordion item isn't what most users would expect from the functionality of an accordion. Discussing this in the team meeting to get everyone's opinions might be helpful.

packages/nimble-components/src/accordion/specs/README.md
aidendk: Sure, I'll include them. If I understand your question correctly, it is supported to override the `collapsed-icon` and `expended-icon`- I've been looking into the accordion item implementation in Nimble, and I was able to set these icons to nimble icons in the export of the accordion item component in index.ts [See FAST's example regarding this](https://github.com/microsoft/fast/tree/archives/fast-element-1/packages/web-components/fast-foundation/src/accordion#accordionitem).

packages/nimble-components/src/accordion/specs/README.md
aidendk: I updated the open issue to reflect what I discussed in the comment above 👍

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Visual Design Comments:
+
+Brandon's Figma spec has designs for two accordion item heights, 24px and 32px. The 32px will be implemented as this is what we have previously focused on when choosing between the two sizes.
+
+Brandon's Figma spec also only shows designs for one accordion item at a time. How should accordion items in each appearance type be separated from each other (ex. a line between accordion items, transparent space between accordion items, etc.)? Because there's only one accordion item design at a time, it is also unclear if error-visible is a nimble-accordion-item specific state, or if it should affect all accordion items and be used in the nimble-accordion component.
+
+The Figma design does not have a design for the "error" focusVisible state. This might be confusing, as keyboard users who tab to the error accordion will suddenly see the red border and exclamation mark icon disappear in the focusVisible state.
+
+When should each `appearance` type be used? Will there be specific guidance on this, and if so, will it be similar to the guidance for other Nimble components that use the same `appearance` types (ex. the Nimble-Button)?
```
jattasNI, line 21: A collection of edits and formatting changes to help position this section more clearly:
```suggestion
There are several open questions with the current visual design. They're tracked here for background but won't block initial approval of this component spec.

1. the Figma spec has designs for two accordion item heights, 24px and 32px. The 32px will be implemented first as this matches other components. We can implement the 24px variation when we receive client requests for it.

2. the Figma spec only shows designs for one accordion item at a time. How should accordion items in each appearance type be separated from each other (ex. a line between accordion items, transparent space between accordion items, etc.)? Because there's only one accordion item design at a time, it is also unclear if error-visible is a nimble-accordion-item specific state, or if it should affect all accordion items and be used in the nimble-accordion component.

3. The Figma spec does not have a design for the "error" focusVisible state. This might be confusing, as keyboard users who tab to the error accordion will suddenly see the red border and exclamation mark icon disappear in the focusVisible state.

4. When should each `appearance` type be used? Will there be specific guidance on this, and if so, will it be similar to the guidance for other Nimble components that use the same `appearance` types (ex. the Nimble-Button)?
```

jattasNI: Now that I see it in context, I would also be happy with tracking these in a separate visual design issue. That way we don't have to update the spec when we resolve them.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,152 @@
...
+        Accordion one content
+    </nimble-accordion-item>
+    <nimble-accordion-item expanded>
+        <span slot="heading">Accordion two</span>
+        <nimble-checkbox></nimble-checkbox>
+        Accordion two content
+    </nimble-accordion-item>
+    <nimble-accordion-item error-visible>
+        <span slot="heading">Accordion three</span>
+        <nimble-icon-exclamation-mark></nimble-icon-exclamation-mark>
```
jattasNI, line 41: Including the exclamation mark here is a bit ambiguous.

Are we saying that clients have to provide their own exclamation mark to indicate the error-visible state? I don't think we'd want that; if it's part of the design then the accordion should provide it.

Or are we just showing that you can include an icon as content? If that's the goal then let's pick a different icon for the example.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,152 @@
...
+    -   No Additional Requirements
+-   _Styling: Does FAST provide APIs to achieve the styling in the visual design spec?_
+    -   No Additional Requirements
+-   _Testing: Is FAST's coverage sufficient? Should we write any tests beyond Chromatic visual tests?_
+    -   No Additional Requirements
+-   _Documentation: Any requirements besides standard Storybook docs and updating the Example Client App demo?_
+    -   No Additional Requirements
+-   _Tooling: Any new tools, updates to tools, code generation, etc?_
+    -   No Additional Requirements
+-   _Accessibility: keyboard navigation/focus, form input, use with assistive technology, etc._
```
jattasNI, line 139: Could you also evaluate how well the FAST accordion follows the ARIA guidance in that document under "WAI-ARIA Roles, States, and Properties"?

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,152 @@
...
+    -   No Additional Requirements
+-   _Performance: does the FAST component meet Nimble's performance requirements?_
+    -   No Additional Requirements
+-   _Security: Any requirements for security?_
+    -   No Additional Requirements
+
+---
+
+## Open Issues
+
```
jattasNI, line 151: Before approval we should come up with a plan to resolve these open issues. It could be some combination of:
1. file issues to FAST to track them as bugs
2. fix the issues in FAST or find workarounds
3. disable (or just don't document) features like single expand mode or nesting in our first implementation until these behaviors are fixed
4. decide we like some of these behaviors 

Opening this thread for opinions but as I mentioned in standup I'd vote to keep this moving by cutting scope if possible. (We'd still want to file issues to FAST if we did that)

packages/nimble-components/src/accordion/specs/README.md
aidendk: Sure, I'll create an issue and link to it in the accordion spec,

packages/nimble-components/src/accordion/specs/README.md
aidendk: Yeah, that's true, I'd assume that we wouldn't want clients to change the error icon, and they shouldn't have to include it themselves. I'll update this in the next commit.

change/@ni-nimble-components-d6d8517b-312c-467f-8b69-bdaa581e67fa.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
```
rajsite, line 2: The shoelace design system just switched to using the native `summary` and `detail` elements in their accordion style component because by leveraging them it allowed the [native browser search feature](https://github.com/shoelace-style/shoelace/pull/1470) to search the contents.

If the FAST templates do not allow browser search into the collapsed accordions and are not leveraging the native `summary` + `details` components for the accordion that would be a strong reason to fork their template (maybe just for accordion-item) to have that capability.

packages/nimble-components/src/accordion/specs/README.md
aidendk: We discussed a plan to resolve these issues in today's meeting, and I documented that plan in the spec. 👍

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+The `nimble-accordion` is a vertical stack of interactive headings. The headings are controls that can be used to reveal or hide content in specific sections. The Nimble accordion is based upon [FAST's accordion component](https://github.com/microsoft/fast/tree/archives/fast-element-1/packages/web-components/fast-foundation/src/accordion).
+
+### Background
+
+[Nimble issue #533: nimble-accordion Component](https://github.com/ni/nimble/issues/533)
+
+[Visual Design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-85131&mode=design&t=DxDRlqT7MeCPLOxi-0)
+
+There are several open questions with the current visual design. They're tracked in this GitHub issue [#1399](https://github.com/ni/nimble/issues/1399).
```
rajsite, line 13: The issue was closed before this PR went in, so this line can be removed and anything that needs to be captured in the spec should be updated

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+### API
+
+#### Nimble Accordion
+
+[FAST accordion API](https://github.com/microsoft/fast/blob/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion/accordion.spec.md)
+
+-   _Component Name:_ `nimble-accordion`
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
+    -   `expand-mode` - `multi`
```
rajsite, line 84: If there is only a single expand-mode option then we can avoid having the property at all for now.

Instead there should be a note here stating that expand-mode will be omitted / unconfigurable from the FAST base class.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+When single expand mode is enabled, the first accordion item is initially open, and can't be closed by clicking on it- it is only closed by clicking a different accordion item. At least one accordion has to be open at all times, which could be confusing to users as the expected functionality would be that all accordion items are initially closed until you open them and can be closed by clicking on that specific accordion item.
+
+This issue will be discussed with UX designers to determine if having this `single` expand mode functionality is necessary.
+
+Additionally, when using nested accordions with the single expand mode, once you open a child accordion, the parent accordion closes- when you open the parent accordion again, the previously selected child accordion is open, but when clicking on it, it doesn't close. Also, when opening the second child accordion, the first child accordion stays open, which goes against the functionality of single expand mode.
+
+This issue with nested accordions will be filed to FAST, and the `expand-mode` attribute will be updated to include the `single` expand mode once FAST resolves this.
```
rajsite, line 102: A few lines up it states we will discuss the single mode behavior with UX. Seems like the line should say "once FAST resolves the issue and UX confirms the single mode behavior"

packages/nimble-components/src/accordion/specs/README.md
rajsite: For purposes of the spec are we still waiting on feedback here?

- is error states for individual accordion items or the full accordion?
- how do borders behave for nested accordions?
-

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+-   _Styling: Does FAST provide APIs to achieve the styling in the visual design spec?_
+    -   No Additional Requirements
+-   _Testing: Is FAST's coverage sufficient? Should we write any tests beyond Chromatic visual tests?_
+    -   No Additional Requirements
+-   _Documentation: Any requirements besides standard Storybook docs and updating the Example Client App demo?_
+    -   No Additional Requirements
+-   _Tooling: Any new tools, updates to tools, code generation, etc?_
+    -   No Additional Requirements
+-   _Accessibility: keyboard navigation/focus, form input, use with assistive technology, etc._
+    -   These are [aria's guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) for keyboard interactions: enter or space for actions, tab and shift + tab for moving up and down, arrow keys for moving up and down between headers, and home / end (optional) for moving to the first and last accordion header, respectively. I tested all of these in the FAST component explorer, and they're functional. It is expected that these interactions will still be functional when implemented in Nimble.
```
rajsite, line 144: Just to confirm, is FAST following all guidelines for accordions and for nested accordions?

aidendk: Yes, they follow all of w3's guidelines. The only exception is the `single` expand mode nested accordion, which won't be included until FAST fixes this issue and UX designers decide that using the `single` expand mode is useful.

aidendk:
@rajsite I answered [these questions](https://github.com/ni/nimble/pull/1371#discussion_r1282213834) in the newest commit 👍. Error-visible will be set on individual accordion items, and all nested accordions will have the `ghost` appearance.

packages/nimble-components/src/accordion/specs/README.md
aidendk: I added answers to [these questions](https://github.com/ni/nimble/pull/1371#discussion_r1282213834) in the newest spec commit 👍. Error-visible will be set on individual accordion items, and all nested accordions will have the `ghost` appearance.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+-   _Properties/Attributes:_
+    -   `appearance` - `block`, `outline`, `ghost`
+    -   FAST's attribute `expand-mode` will be omitted from the FAST base class due to issues discussed below.
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ Unchanged
+
+The Figma design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. These designs are similar to those of the nimble-button and its `appearance` types, but the accordion does not use the `.control` css class, it uses `.region` and `.heading`. Because of this, differences in border functionality, and the possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion and accordion-item will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value (ex. `:host([appearance='outline'])` will have styling for the outline button).
```
atmgrifter00, line 88:
```suggestion
The design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. These designs are similar to those of the nimble-button and its `appearance` types, but the accordion does not use the `.control` css class, it uses `.region` and `.heading`. Because of this, differences in border functionality, and the possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion and accordion-item will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value (ex. `:host([appearance='outline'])` will have styling for the outline button).
```

No reason to call out Figma by name in these specs.

atmgrifter00 (approved):

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+-   _Slots:_ Unchanged
+
+The design includes appearances of the accordion header that reflect those of the `block`, `outline`, and `ghost` appearances used in other components. These designs are similar to those of the nimble-button and its `appearance` types, but the accordion does not use the `.control` css class, it uses `.region` and `.heading`. Because of this, differences in border functionality, and the possibility of excess overridden css classes if patterns are used, shared styles will not be used, and new styles for the accordion and accordion-item will be created. The appearances of `block`, `outline`, and `ghost` will be implemented under the HTML attribute `appearance`, which will use conditional css styles based on `appearance`'s value (ex. `:host([appearance='outline'])` will have styling for the outline button).
+
+Documentation will be added to advise against using multiple Nimble Accordions with different appearances next to each other.
+
+When using nested accordions, the parent accordion will have the selected `appearance` type, and all children accordions will be forced to have the `ghost` appearance type.
```
mollykreis, line 92: I haven't been following along very closely to this HLD, so I might have missed the discussion on this -- why do all child accordions need to be forced to have a `ghost` appearance type? It would be good to have a sentence justifying/explaining why this is the case.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
+-   _Methods:_ Unchanged
+-   _Events:_ Unchanged
+-   _CSS Classes and Custom Properties that affect the component:_ Unchanged
+-   _Slots:_ The `start` and `end` slots will not be used. The `collapsed-icon` and `expanded-icon` slots are set by the nimble-accordion-item in its index.ts, so overriding them will not be supported.
+
+The design also includes an `error` state. This will be specific to each accordion item and controlled with the boolean attribute `error-visible`, which has a default attribute value of "". When needed, "error-visible" would be added to the accordion-item attributes, changing the color of the accordion item border color to red. This will be implemented in the accordion styling through conditional css styles based on `error-visible`'s value (ex. `:host([error-visible])` will have styling for when `error-visible` is true).
```
mollykreis, line 116: I think it would make more sense to say the default value of the boolean attribute is `false` rather than an empty string.

```suggestion
The design also includes an `error` state. This will be specific to each accordion item and controlled with the boolean attribute `error-visible`, which has a default attribute value of `false`. When needed, "error-visible" would be added to the accordion-item attributes, changing the color of the accordion item border color to red. This will be implemented in the accordion styling through conditional css styles based on `error-visible`'s value (ex. `:host([error-visible])` will have styling for when `error-visible` is true).
```

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+    -   No Additional Requirements
+-   _Styling: Does FAST provide APIs to achieve the styling in the visual design spec?_
+    -   No Additional Requirements
+-   _Testing: Is FAST's coverage sufficient? Should we write any tests beyond Chromatic visual tests?_
+    -   No Additional Requirements
```
mollykreis, line 138: Since we are forking FAST's template, there should be additional tests brought in to make sure our component and its template work as expected. Depending on how significant the change is to our fork of the template, we should either write a bunch of tests from scratch or we should copy the tests from FAST (as described [here](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#extend-fast-foundation)).

packages/nimble-components/src/accordion/specs/README.md
rajsite, line 92: I'm concerned a little about how this is described. It may be updated based on implementation but generally we want to avoid a situation where we are manipulating attributes that are owned by the user.

So we wouldn't for example want a parent accordion setting the actual appearance attribute on a child accordion. You can end up with a conflict in state where the user expects the attribute to be set one way and we are fighting the value they set and overwriting it with another.

Instead we would want something out of band. Not communicating with public attributes. Alternatively we say that complexity isn't worth it and we just document that child accordions should be configured as ghost by the user.

Spec-wise I think you should say if it is enforced via docs or by the components is an open question.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+The design also includes an `error` state. This will be specific to each accordion item and controlled with the boolean attribute `error-visible`, which has a default attribute value of "". When needed, "error-visible" would be added to the accordion-item attributes, changing the color of the accordion item border color to red. This will be implemented in the accordion styling through conditional css styles based on `error-visible`'s value (ex. `:host([error-visible])` will have styling for when `error-visible` is true).
+
+Usage guidance for the `error-visible` state will be created, as its usage may be unclear.
+
+The shoelace design system switched to using `summary` and `details` elements in their accordion styled component, which allowed the [native browser search feature](https://github.com/shoelace-style/shoelace/pull/1470) to search inside of accordions that are closed. FAST does not allow this, so their template for the accordion item will be forked and changed to allow the native browser search feature.
```
mollykreis, line 120: Can you provide some more details here on what the new template will look like and how significant the changes are from what FAST has? If you have prototyped this working, it would be helpful to put the template code from it here to give everyone an idea of what it'll look like.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+When single expand mode is enabled, the first accordion item is initially open, and can't be closed by clicking on it- it is only closed by clicking a different accordion item. At least one accordion has to be open at all times, which could be confusing to users as the expected functionality would be that all accordion items are initially closed until you open them and can be closed by clicking on that specific accordion item.
+
+This issue will be discussed with UX designers to determine if having this `single` expand mode functionality is necessary.
+
+Additionally, when using nested accordions with the single expand mode, once you open a child accordion, the parent accordion closes- when you open the parent accordion again, the previously selected child accordion is open, but when clicking on it, it doesn't close. Also, when opening the second child accordion, the first child accordion stays open, which goes against the functionality of single expand mode.
+
+This issue with nested accordions will be filed to FAST. Once FAST resolves the issue and UX confirms the `single` expand mode behavior, the `expand-mode` attribute will be updated to include the `single` expand mode.
```
rajsite, line 102: This section was not updated to align with the other discussion, i.e. not "temporarily omitted to have one default value" but the `expand-mode` property will be omitted altogether.

Also can you move this `expand-mode` discussion to a Future Work section? What should be captured in the spec body is that `expand-mode` is being restricted / omitted and the rest of the discussion should be in Future Work.

packages/nimble-components/src/accordion/specs/README.md
rajsite, line 116: > which has a default attribute value of ""

A boolean attribute value of empty string means true, i.e. `<accordion-item error-visible>` is the same as `<accordion-item error-visible="">`
A false boolean attribute has an attribute value null and is absent, i.e. `<accordion-item>` is error-visible false.

All boolean attributes have a default value of false which I think is what this was trying to express.
I think youu can just remove the `, which has a default attribute value of ""` part.

rajsite (approved):

rajsite (commented):
Just a bit more clean-up needed, getting close!

packages/nimble-components/src/accordion/specs/README.md
aidendk: Yeah, with the addition of the `details` and `summary` html elements we should add some extra tests to make sure that they're functioning correctly.

packages/nimble-components/src/accordion/specs/README.md
rajsite: There were several questions brought up at stand-up that need to be addressed with Brandon because they can have a large impact on implementation direction:

- Do we need the block appearance mode behavior where nested accordion headers extend all the way to the edge while nested accordion items are indented. (That requirement may require creation of a nested-accordion element to accomplish, may require more research)

Design questions:
- Should the hover borders appear around the full accordion items or just the header. Concerns that it is not clearly indicating what part of the control you can interact with. Concerns that it creates visual noise particularly with accordions with many / long accordion-items.
- Do we have use cases for the different appearance modes? Is there a discussion of just having block (for the outermost accordion) and ghost (for nested accordions)? What's the use case for outline accordions?
- Should outline accordions only outline the header (as block accordions only show the block background for the header)?

Moving back to draft per new contribution policy as clarifications about nested sub accordion requirements could result in more significant changes (creation of additional elements and maybe some prototyping).

aidendk: These questions were discussed with Brandon and overall, it seems that the Accordion will be put to the side for now and revisited at a later date due to other priorities, the need for an IxD, and the need for more specific accordion UX designs based on the questions presented.

Some possible paths to take were discussed, including:
- It's currently unclear how appearances other than the ghost border will function inside of nested accordions, as UX designs for these cases aren't fully created yet.
- The hover border should only appear around the accordion header (if we decide to keep a hover animation at all, as many other examples of accordions do not have a hover animation).
- Thoughts on keeping different appearance modes are still unclear, but it's been discussed that designers would benefit from having multiple appearance options available when using accordions.
- I don't believe the question regarding outlines around only the header was discussed, but if we're changing the hover animation to only show up around the header, then I would presume that the outline border would follow suit as a green hover border taking up only the header portion of the outline in the expanded mode (instead of the entire outline border) would look odd.

packages/nimble-components/src/accordion/specs/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+`expand-mode` is a FAST attribute, but is being temporarily restricted / removed from then nimble-accordion due to issues with the `single` expand-mode. See "Future Work" for more info.
+
+#### Nimble Accordion Item
+
+[FAST accordion-item API](https://github.com/microsoft/fast/tree/57f3c22c6341d8a21d48b1ffb7fcbfab1ffd02d8/packages/web-components/fast-foundation/src/accordion-item)
+
+-   _Component Name:_ `nimble-accordion-item`
+-   _Properties/Attributes:_
+    -   `error-visible` - boolean
```
atmgrifter00, line 102: After reading the [Accordion IxD](https://github.com/ni/nimble/blob/28e170974e54cb8a2f97d0804119c191f9f60369/packages/nimble-components/src/accordion/specs/IxD.md) (yet to be approved), one interesting description was that an `accordion-item` should represent an error state if one of its children was in an error state. This is interesting in that it could cause us to re-evaluate this API, if, say, we wanted this behavior to occur with no user configuration.

Often, accordion-items are meant to be layout components, and not have any technical association with their content. With the current proposed API along with the guidance given in the IxD, to achieve the desired behavior a client would have to (in code-behind):
- explicitly detect when the error state of a component with an `accordion-item` changed
- find the `accordion-item` parent(s) it resides in (could be more than one?)
- explicitly change the `error-visible` state in each parent

Instead, is it better to have the `accordion-item` do automatic detection of these DOM changes, and then update its visual state as needed, thus requiring no user configuration? There are some questions that we would need to answer before being comfortable with this:
- How would we handle non-Nimble components that offered an error state, which we would also want to be reflected by the `accordion-item`?
- Would we be comfortable with using a `MutationObserver` to manage the automatic notification?
- Is it even okay to take this control away from the user (forces the `accordion-item` to always have this visual state when a child Nimble component is an error state)? Does it make sense to offer an API to turn this auto-detection on or off?

rajsite:
Closing the PR temporarily until accordion IxD is approved and this PR can be re-openend.

packages/nimble-components/src/accordion/specs/README.md
rajsite: I think consistency with our other component behaviors would be for the applications doing the validation and specifying of error states.

Should be a common thing frameworks deal with (propagating error states up a tree like FormGroups in Angular, etc).

Similar to how we aren't providing hooks in the HTML 5 validation apis directly. You have to listen for those events and set error visible on the component yourself. Or in more apps today you listen for Angular's validation events. Also similar to how disabling a text field doesn't automatically disable slotted action buttons in the text field, that is application responsibility.

Certainly doesn't represent all state. I.e. selection state across tree-items in managed by the element. But I think we are pretty consistent that validation state is managed by the app and not the components. Trying to think of contrary examples. Also could be convinced this is an exception.

PR #1373, opened 2023-07-21
Table column resize fires event

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #846

## 👩‍💻 Implementation

Simply have the `Table` listen to changes of the already present `isColumnBeingSized` observable of the `TableLayoutManager`. When it changes to `false`, it indicates an interactive column sizing operation has been completed.

## 🧪 Testing

Added unit test.

## ✅ Checklist


packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -24,7 +24,7 @@ export class TableLayoutManager<TData extends TableRecord> {
     private dragStart = 0;
     private leftColumnIndex?: number;
     private rightColumnIndex?: number;
-    private initialColumnPixelWidths: {
+    private initialColumnWidths: {
```
atmgrifter00, line 27: All changes to this file are simply a name change.

packages/nimble-components/src/table/index.ts
```diff
@@ -243,6 +244,8 @@ export class Table<
         this.table = tanStackCreateTable(this.options);
         this.virtualizer = new Virtualizer(this, this.table);
         this.layoutManager = new TableLayoutManager(this);
+        this.layoutManagerNotifier = Observable.getNotifier(this.layoutManager);
+        this.layoutManagerNotifier.subscribe(this);
```
rajsite, line 248: I think one of these, probably subscribe, can take the property name to observe. Could maybe filter out some unnecessary notifications since we just care about 1 prop.

packages/nimble-components/src/table/index.ts
```diff
@@ -346,6 +349,15 @@ export class Table<
                 this.tableUpdateTracker.trackColumnPropertyChanged(args);
             }
         }
+        if (
```
rajsite, line 352: can do an `else if`

rajsite (approved):

PR #1379, opened 2023-07-23
Updated parameterized test pattern

rajsite:
# Pull Request

## 🤨 Rationale

Updates the parametrized test pattern to address a few shortcomings in the previous pattern:
- Avoids a lot of duplicated boilerplate
- Doesn't encourage an approach that results in eslint disables being used often (false positive of await used in a loop)
- Prevents accidentally focusing a test and checking it in
- Gives decent typing, may take run-time to see if it needs improvements

## 👩‍💻 Implementation

Exposed new parameterization helpers named `parameterize` and `parameterizeNamedList` and marked the existing method deprecated.
Created a tech debt task to track migrating to the new pattern: https://github.com/ni/nimble/issues/1551

## 🧪 Testing

Added some tests for the new helpers and migrated a couple of existing tests to the new pattern.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Added detailed comments to the helper like the deprecated helper did.


packages/nimble-components/src/utilities/tests/parameterized.ts
```diff
@@ -1,23 +1,7 @@
 // eslint-disable-next-line no-restricted-globals
```
jattasNI, line 1: This is in `src/utilities/tests`, whose README says is for nimble-components internal test utilites. We also have `src/testing`, whose README says is for things that could also be used by clients. Do you think this belongs in the latter place (either now or eventually once it's had some runtime)?

packages/nimble-components/src/utilities/tests/parameterized.ts
```diff
@@ -64,3 +30,117 @@ export const getSpecTypeByNamedList = <T extends { name: string }>(
     (x: T) => focusList.includes(x.name),
     (x: T) => disabledList.includes(x.name)
 );
+
+// eslint-disable-next-line no-restricted-globals
```
jattasNI, line 34: Could be worth a comment both here and in the `.spec.ts` saying that clients should not disable this lint rule.

packages/nimble-components/src/utilities/tests/parameterized.ts
```diff
@@ -64,3 +30,117 @@ export const getSpecTypeByNamedList = <T extends { name: string }>(
...
+
+/**
+ * Used to create a parameterized test using an object of test names and arbitrary test values.
+ * In the following example:
+ *  - the test named `catsAndDogs` is focused for debugging
+ *  - the test named `frogs` is configured to always be disabled
+ *  - the test named `men` will run normally as it has no override
+ * @example
+ * const rainTests = {
+ *     catsAndDogs: 'idiom',
```
jattasNI, line 55: nit: is it possible to use a syntax that allows for spaces in the names of the tests? That would make the specs read a bit easier.

Not critical if it requires a major refactor but if it's easy to just change these examples to quote the property names then I'd vote to encourage readability in the printed test name.

change/@ni-nimble-components-5528f3eb-a2ce-4e80-9df4-6ddb04f8b015.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I think the follow up work you mentioned in the PR description (moving away from the now-deprecated APIs) is indeed worth tracking in tech debt.

jattasNI (approved):

packages/nimble-components/src/utilities/tests/parameterized.ts
rajsite: Yes... it is possible, but wondering what our guidance should be.

Example 1:

```ts
const rainTests = {
    // eslint-disable-next-line @typescript-eslint/naming-convention
    'really really really really really long test name cats and dogs': {
        testData: 0
    },
} as const;
describe('Different rains', () => {
    parameterize(rainTests, (spec, name, value) => {
        spec(`of type ${name} exist`, () => {
            expect(value).toBeDefined();
        });
    }, {
        // eslint-disable-next-line @typescript-eslint/naming-convention
        'really really really really really long test name cats and dogs': fit,
    });
});
```
Pros:
- Don't need to think of an identifier for each test case. These names get [really long](https://github.com/ni/nimble/blob/931138bb9d3601051afc3e394885a8a90e9dc83b/packages/nimble-components/src/table/tests/table-column-sizing.spec.ts#L185) and nuanced so I guess it's nice to not need to think of a short identifier name that represents this case.

Cons:
- We would need to update our eslint rules to allow for funky key names. Maybe we just do it for tests? Or do we have eslints on each line? Maybe there is a rule config we can loosen vs completely disable in tests
- Makes the overrides look clunky but that's not a big deal really

Example 2:
```ts
const rainTests = {
    catsAndDogs: {
        specName: 'really really really really really long test name cats and dogs',
        testData: 0
    },
} as const;
describe('Different rains', () => {
    parameterize(rainTests, (spec, _name, value) => {
        spec(`of type ${value.specName} exist`, () => {
            expect(value).toBeDefined();
        });
    }, {
        catsAndDogs: fit
    });
});
```

Pros:
- More normal configuration object shape
- More normal keys for disabling in override
- Semi forced / encouraged to come up with a shorter name, could maybe help with distinguishing a test case

Cons:
- A fair bit more mental energy to think of test case names
- End up with a frequently unused `name` parameter

jattasNI: Consensus with coding practices WGwas to start with Example 1 and consider changing our eslint defaults so that naming-convention doesn't fire for this case

packages/nimble-components/src/utilities/tests/parameterized.ts
rajsite: I think once it has some runtime we should move it to a new package in the styleguide repo to share across applications
I think that's captured sufficiently here: https://github.com/ni/javascript-styleguide/issues/125#issuecomment-1631682110

packages/nimble-components/src/utilities/tests/parameterized.ts
rajsite: Made a pattern that allows configuring via the named list pattern (which also aligns more closely with that already exists) and I think it avoids all the trade-offs discussed.

change/@ni-nimble-components-5528f3eb-a2ce-4e80-9df4-6ddb04f8b015.json
rajsite: Created https://github.com/ni/nimble/issues/1551

rajsite:
@jattasNI I meant to reset you since I changed the pattern 😅
If you have any concerns I can address them in a follow-up

PR #1380, opened 2023-07-24
Add essential APIs into the rich-text-editor spec

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

As part of the comments feature, it is necessary to expose the following APIs from the `nimble-rich-text-editor` to configure and meet the specific requirements of comments,
1. `disabled`
2. `error-visible`
3. `error-text`

## 👩‍💻 Implementation

Added the above APIs to the design section of the `nimble-rich-text-editor` spec.

## ✅ Checklist


packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -111,6 +111,9 @@ _Props/Attrs_
     [isEmpty](https://tiptap.dev/api/editor#is-empty) API. The component and the Angular directive will have a getter method
     that can be used to bind it in the Angular application.
 -   `fitToContent` - is a boolean attribute allows the text area to expand vertically to fit the content.
+-   `maxlength` - is a number attribute that restricts the number of input characters in the editor. This is achieved through
```
vikisekarNI, line 114: Are we going to update the visual design for characters limit similar to the snip below? If not, it will be good to explicitly mention it here and get Nimble code owners advise on the same. 

image

vivinkrishna-ni: As per the current visual design, we are not showing this information, so I have mentioned explicitly in the spec that will not show this information in the UI and use it only to limit the characters.

vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
jattasNI: One personal pet peeve is when I paste a large amount of text into a field that limits the max length of text. If there's no feedback that I've hit the limit, I may not notice that text has been truncated. 

I'd defer to your team and the designers about whether that's worth improving and how we might do so. Many of our other input components have a red underline and glyph that we use to show error state. Are we not doing that here for scope reasons or UX reasons?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -111,6 +111,11 @@ _Props/Attrs_
...
```
jattasNI, line 114: It'd be good to mention that the name `maxlength` matches attributes on native elements (could even link to MDN docs for reference). I was initially confused why we didn't choose something more inline with our other APIs like `max-length` or `maximum-length`. But given there is precendence with native elements, I like the idea of following that precedence as you're currently suggesting.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -111,6 +111,11 @@ _Props/Attrs_
...
+    Tiptap's [CharacterCount extension](https://tiptap.dev/api/extensions/character-count#limit). However, we will not display
+    the number of characters currently entered in the editor along with the maximum length (e.g., 10/100 characters entered).
+    We use this extension only to limit the number of characters that can be entered into the editor.
+-   `disabled` - is a boolean attribute to disable the editor by restricting all the user interactions within the component.
```
jattasNI, line 118: Will there also be visual styling to reflect the disabled state?

jattasNI, line 118: Does the disabled state apply to content that the client slotted in too?

atmgrifter00, line 118:
```suggestion
-   `disabled` - is a boolean attribute to disable the editor by preventing all user interactions within the component.
```

rajsite: For slotted buttons that has not been the pattern used by other controls. i.e. action buttons in a text field need to be disabled by the host application. I'd expect stuff placed in the footer slot to have state managed by the app and nimble will control layout / placement.

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: I'm pretty nervous about this API. It feels like "max length" is a very application specific definition.

Is it the max length of visible characters? Then how many is: 👨‍👩‍👦 which is treated as a "single character" for text selection and rendering in a browser with your mouse but is a string of length 8 in javascript.

How long is the following string?:
```
[LOL](http://www.google.com)
```
Visually it is 3 characters, but in markdown it is a javascript string of length 28. I think the plugin would treat this as 3 characters: https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-character-count/src/character-count.ts#L51

Right now I lean towards this being a concept that is application specific. Instead the control should have an error state that the application can configure like how text field, etc. behave.

Based on understanding the exact requirements we can make sure that the API exposes enough information for the application to decide how to put the control in an error state. My guess of the requirement is that we want to be able to limit the size of the generated markdown that is sent to the server, not the visible character count. In which case we would not need an additional API. Instead when the user triggers a submit and the markdown is calculated the application can decide if it's too large and configure the error state of the control.

But that is based on a guess of the application requirements. Please let us know what they actually are.

vikisekarNI: As mentioned rightly, we were thinking `maxlength` for only visible characters and it seems to be tricky with markdown length coming into picture. But yes, our application requirements is to restrict the **markdown content to 10k characters**. It will be a simple effort to error validate the same in application side and set the error flag through `error-visible` and `error-text`

However, our primary thought process to add `maxlength` was to limit the number of visible characters (say default is 10k) that an user can enter in UI to avoid any performance issues like freezing. 
1. Freezing seems to be quite uncommon when user types and enters input 
2. Freezing seems to be common while user tries to copy paste large characters and then realizes that editor is frozen. [Tiptap claims that it supports more than 200k characters](https://tiptap.dev/examples/book), but in real application environment with other background work, this number may get even lower based on application design and load. 

We would like to confirm if this is too much of a performance corner case to defer for later?

**Note:**
If we choose to have `maxlength` it's good to be provided with a visual cue like in below snip for a better UX. This informs user that any text input greater than this limit will be simply restricted/truncated.

image

jattasNI: I tried pasting a large amount of text into comment fields in AzDO and GitHub.
1. AzDO work item comment accepted the text but performed very poorly (still hung after 20 seconds)
2. AzDO PR comment truncated the text with no feedback (besides that it stopped accepting further input)
3. GitHub accepted the text but performed very poorly (10+ seconds to type a character)

This suggests to me that fully preventing the freezes could be a corner case not worth prioritizing right now.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -396,6 +405,7 @@ library. For the currently supported features, we will include the following lib
 -   [@tiptap/core](https://www.npmjs.com/package/@tiptap/core)
 -   [@tiptap/pm](https://www.npmjs.com/package/@tiptap/pm)
 -   [@tiptap/starter-kit](https://www.npmjs.com/package/@tiptap/starter-kit)
+-   [@tiptap/extension-character-count](https://www.npmjs.com/package/@tiptap/extension-character-count)
```
rajsite, line 408: If the direction of not handling maxlength is agreed then reminder to remove this dep

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: I have removed `markdown` and updated the spec as suggested.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Removed

rajsite (approved):

PR #1381, opened 2023-07-24
Add formatting API to date column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1014
Adding ability to customize formatting of dates/times.

## 👩‍💻 Implementation

- add `format` attribute and nineteen other custom formatting configuration attributes
- add unit tests
- update Storybook story

## 🧪 Testing

Unit tests pass, and tested manually in Storybook.

## ✅ Checklist


packages/nimble-components/src/table-column/date-text/group-header-view/index.ts
```diff
@@ -17,23 +18,19 @@ export class TableColumnDateTextGroupHeaderView extends TableColumnTextGroupHead
...
-                this.text = '';
-            }
-        } else {
-            this.text = '';
-        }
+        this.updateText();
+    }
+
+    private updateText(): void {
+        this.text = TableColumnDateTextCellView.formatNumericDate(
```
mollykreis, line 31: I expect this to be reworked at least a little bit based on my comment regarding the static formatter of the cell view causing problems, but I'll make a note here as well -- it doesn't seem quite right to me for the group header to be calling into the cell to do formatting for it. I think it would be better to have shared logic in an independent location that they both could call to avoid code duplication.

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -10,7 +11,34 @@ import type { ColumnInternalsOptions } from '../base/models/column-internals';
 
 export type TableColumnDateTextCellRecord = TableNumberField<'value'>;
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface TableColumnDateTextColumnConfig {}
+export interface TableColumnDateTextColumnConfig {
+    format?: 'custom';
```
mollykreis, line 15: Was there a specific rational for hard-coding the possible strings rather than following the existing pattern in nimble of creating enums? Part of the reason I ask is because it looks like these values are duplicated in a few different places.

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -22,6 +50,76 @@ declare global {
...
+    public customDateStyle?: 'full' | 'long' | 'medium' | 'short';
+
+    @attr({ attribute: 'custom-time-style' })
+    public customTimeStyle?: 'full' | 'long' | 'medium' | 'short';
+
+    @attr({ attribute: 'custom-hour-cycle' })
+    public customHourCycle?: 'h11' | 'h12' | 'h23' | 'h24';
+
+    public override connectedCallback(): void {
+        this.updateColumnConfig();
```
mollykreis, line 120: I could be wrong, but I was thinking that you didn't need to call your change handler within the `connectedCallback` if none of the `xyzChanged()` functions returned early when the component isn't connected. Is that not the case? Did you find that overriding `connectedCallback` was required? If it is required, you need to make sure to call `super.connectedCallback()`.

mollykreis (changes requested):

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -20,24 +21,80 @@ export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
 TableColumnDateTextCellRecord,
 TableColumnDateTextColumnConfig
 > {
-    private static readonly formatter = new Intl.DateTimeFormat(undefined, {
-        dateStyle: 'medium',
-        timeStyle: 'medium'
-    });
+    private static formatter: Intl.DateTimeFormat;
```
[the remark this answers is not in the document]
m-akinc: Of course. Duh. I'm not sure it's worth maintaining a map that stores one formatter per column, so I'm just creating one for every cell view and group header view.

packages/nimble-components/src/table-column/date-text/index.ts
m-akinc: It looked like it was going to be a bunch of enums that wouldn't get reused much, but on second look, I can see benefit to creating them.

packages/nimble-components/src/table-column/date-text/index.ts
m-akinc: Without it, the group headers don't render a date. Somehow the TableColumnDateTextGroupHeaderView doesn't get a columnConfig.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -20,24 +21,80 @@ export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
...
-                this.text = '';
+                return '';
             }
         } else {
-            this.text = '';
+            return '';
         }
     }
+
+    public static updateFormatter(
```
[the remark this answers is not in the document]
m-akinc: I've added a validator, and I'm setting a validity flag based on whether construction of the formatter object throws an exception.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -20,22 +21,23 @@ export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
...
 TableColumnDateTextColumnConfig
 > {
-    private static readonly formatter = new Intl.DateTimeFormat(undefined, {
-        dateStyle: 'medium',
-        timeStyle: 'medium'
-    });
+    private formatter?: Intl.DateTimeFormat;
+
+    private columnConfigChanged(): void {
+        this.formatter = createFormatter(this.columnConfig);
```
mollykreis, line 27: Now that I'm seeing what it looks like for everything to have its own formatter (cell view, column, and group header view), I'm wondering if it would be possible to avoid recreating formatters so often (particularly since the cell view will exist for every single row in the table).

Would it be possible to put the formatter in the columnConfig so that each cell view and each header view are all referencing the instance owned by the column itself?

packages/nimble-components/src/table-column/date-text/types.ts
```diff
@@ -0,0 +1,107 @@
+/**
```
mollykreis, line 1: Can you write some simple tests for these types following the patterns we use elsewhere in nimble?

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -26,187 +26,536 @@ async function setup(): Promise<Fixture<Table<SimpleTableRecord>>> {
...
+            ]);
+            await waitForUpdatesAsync();
+            column.format = 'custom';
+            column.customYear = 'numeric';
+            column.customNumberingSystem = 'fullwide';
+            await waitForUpdatesAsync();
+            expect(pageObject.getRenderedCellContent(0, 0)).toBe('２０１２');
+        });
+
+        it('sets invalid flag on column when custom options are incompatible', async () => {
```
mollykreis, line 454: Would it be valuable to add a few more tests around the validity of this columns? A few cases I can think of are:
- The column becomes valid again when custom options become compatible
- The column is initially valid and/or valid when not using custom formatting

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
```diff
@@ -20,22 +21,23 @@ export class TableColumnDateTextCellView extends TableColumnTextCellViewBase<
...
-            } catch (e) {
-                this.text = '';
-            }
-        } else {
-            this.text = '';
+        this.updateText();
+    }
+
+    private updateText(): void {
+        if (this.formatter) {
```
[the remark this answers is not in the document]
m-akinc: 1. If `cellRecordChanged()` runs before `columnConfigChanged()`, then `this.formatter` will be `undefined`. `cellRecord` and `columnConfig` are both bound to equivalent properties on `cellState` in the `TableCell` template. I'm not sure which fires the "changed" handler first.
2. If there is an inconsistency in the configuration of the formatting options, then `this.formatter` will be `undefined`.

In case 2, there will be a validation error, so no cells will render, and `updateText` won't be called. In case 1, I expect that the short-circuiting call to `updateText()` would be followed immediately by one where `this.formatter` was defined. So it should be moot whether we actually set a value for `text` on that first call, because it will be overwritten.

The same applies for the group header view.

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
m-akinc: I talk about this in a response to an earlier (unresolved) comment. Let's continue the discussion in that thread: https://github.com/ni/nimble/pull/1381#discussion_r1272773193

packages/nimble-components/src/table-column/date-text/cell-view/index.ts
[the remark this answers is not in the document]
mollykreis: I'm curious to hear your rationale for it not being appropriate to put the formatter in the columnConfig.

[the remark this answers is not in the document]
m-akinc: I feel like @rajsite made a point of it being just for "dumb" configuration data, and a formatter object is instead something functional created using that configuration. To be honest, since we're not serializing the columnConfig, I'm not sure what practical reasons there would be for enforcing that restriction. It certainly would be convenient to share the formatter via columnConfig.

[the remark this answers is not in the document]
m-akinc: Talked to Milan, and he clarified his definition of "dumb" config data. Basically he meant state that was independent and not mutable by/exposed to other parts of the system. So the formatter fits that definition and can reasonably be stored in the columnConfig. 😊

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
m-akinc: Added one test for each bullet

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -1,16 +1,53 @@
...
+    HourCycle,
+    MonthFormat,
+    WeekdayFormat
+} from './types';
+import { TableColumnDateTextValidator } from './models/table-column-date-text-validator';
 
 export type TableColumnDateTextCellRecord = TableNumberField<'value'>;
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface TableColumnDateTextColumnConfig {}
+export interface TableColumnDateTextColumnConfig {
```
mollykreis, line 28: Do all of these properties need to be part of the column config? I would have expected just the formatter to be there, and then the other properties influence how the settings with which the formatter is created.

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -31,6 +140,162 @@ export class TableColumnDateText extends TableColumnTextBase {
...
+    protected customTimeStyleChanged(): void {
+        this.updateColumnConfig();
+    }
+
+    protected customHourCycleChanged(): void {
+        this.updateColumnConfig();
+    }
+
+    private updateColumnConfig(): void {
+        const columnConfig: TableColumnDateTextColumnConfig = {
```
mollykreis, line 225: I was playing around with the column, and I realized that when you remove the `custom-*` attributes, the property values get set to `null` rather than `undefined` (there's a bug to FAST somewhere that I could probably dig up about this if we need to), but the `null` value causes an exception to be thrown when creating the formatter. I think we need to handle that within this code so that someone can add & then remove one of the `custom-*` attributes without the table becoming invalid.

packages/nimble-components/src/table-column/date-text/index.ts
m-akinc: That's a good point. We don't need them in there anymore.

packages/nimble-components/src/table-column/date-text/index.ts
m-akinc: Added a workaround

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -31,6 +120,143 @@ export class TableColumnDateText extends TableColumnTextBase {
...
+            formatter: this.createFormatter(this.format)
```
mollykreis, line 206: nit: I don't see a reason to pass `this.format` to `this.createFormatter` since it can read `this.format` from its own implementation.

mollykreis (approved):

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
```diff
@@ -113,9 +167,181 @@ export const dateTextColumn: StoryObj<TextColumnTableArgs> = {
             description:
                 'Set this attribute to identify which field in the data record should be displayed in each column. The field values must be of type `number` and represent the number of milliseconds since January 1, 1970 UTC. This is the representation used by the `Date` type.',
             control: { type: 'none' }
+        },
+        format: {
```
jattasNI, line 171: Could you add docs for the column's validity object field? It might be helpful to mention that a good way to debug invalid configuration is to use [MDN's Try It widget](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#try_it) or a browser console. The browser actually gives very helpful exception messages.

change/@ni-nimble-components-8fdb7efd-10a8-4e79-9dd7-079527db6644.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: Can be patch since table is still incubating
```suggestion
  "type": "patch",
```

jattasNI (commented):
Thank you @m-akinc and @mollykreis for using the new "peer review" process. It felt a lot quicker to review this PR and the discussions you already had looked really productive.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
```diff
@@ -113,9 +167,181 @@ export const dateTextColumn: StoryObj<TextColumnTableArgs> = {
...
+            description:
+                'Refer to the option [documentation for the `Intl.DateTimeFormat()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat). The values presented here are not a comprehensive list.',
+            options: [undefined, 'arab', 'fullwide', 'latn', 'tamldec'],
+            control: { type: 'radio' }
+        },
+        customFormatMatcher: {
+            name: 'custom-format-matcher',
+            description:
+                'Refer to the option [documentation for the `Intl.DateTimeFormat()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat).',
+            defaultValue: { summary: 'best fit' },
```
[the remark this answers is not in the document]
m-akinc: These were the only ones with straight-forward defaults. But I'm fine removing them.

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
m-akinc: I had seen that definition, but I'm not sure where that gets me. Are you imagining something like `public customDay: typeof DateTimeFormatOptions.day;`? For some reason, I can't get that interface to resolve, but assuming I could, I'm not sure that is what we'd want.

packages/nimble-components/src/table-column/date-text/models/format-helper.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function formatNumericDate(
+    formatter: Intl.DateTimeFormat,
+    date: TableNumberFieldValue
+): string {
+    if (typeof date === 'number') {
+        try {
+            return formatter.format(date);
+        } catch (e) {
+            return '';
```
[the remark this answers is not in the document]
m-akinc: It will throw if you pass it something it can't format, like `'foo'` or a value which is out of range for a `Date`. I have tests for a whole range of bad input values that ensure we render a blank cell (near the top of the spec file).

change/@ni-nimble-components-8fdb7efd-10a8-4e79-9dd7-079527db6644.json
m-akinc: Man, I keep doing this.

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
jattasNI: Yeah, that's more or less what I was envisioning. If TypeScript had exposed its underlying types then I was hoping we could either use them directly or use them to construct our types. That way our types would update automatically if new options were added. But I also don't see a way to do that if we can't access those interfaces or their member types.

It could be reasonable to file an issue to TypeScript asking for those types to be exported. I think it's a valid use case to want to provide an API that wraps the native API and not have to re-implement the types. But not critical to do immediately.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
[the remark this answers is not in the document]
jattasNI: Yeah, after scanning the MDN docs more closely I think removing them is the right call. They have their own documentation of default values so it doesn't add much to expose our take on it here.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -26,187 +26,558 @@ async function setup(): Promise<Fixture<Table<SimpleTableRecord>>> {
...
+        });
 
-        const firstColumn = element.columns[0] as TableColumnDateText;
+        it('reports column configuration valid', () => {
+            expect(column.checkValidity()).toBeTrue();
+        });
 
-        expect(firstColumn.checkValidity()).toBeTrue();
-    });
+        const badValueData = [
```
jattasNI, line 90: @rajsite what's the outlook on #1379 being completed so we could use the new templatized pattern here? If it still needs to bake longer do you have an opinion on whether it'd be better to at least use the soon-to-be-deprecated `getSpecTypeByNamedList`?

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -26,187 +26,558 @@ async function setup(): Promise<Fixture<Table<SimpleTableRecord>>> {
...
-        await waitForUpdatesAsync();
-        tablePageObject.dispatchEventToCell(0, 0, new MouseEvent('mouseover'));
-        await waitForUpdatesAsync();
-        expect(pageObject.getCellTitle(0, 0)).toEqual(
-            'Dec 10, 2012, 10:35:05 PM'
-        );
-    });
+        it('honors customDateStyle property', async () => {
+            await element.setData([
+                { field: new Date('Dec 10, 2012, 10:35:05 PM').valueOf() }
```
jattasNI, line 263: In the past I've seen date tests be flaky when run on machines in different time zones or locales. Is there anything we need to do to ensure these pass reliably?

I didn't think too much about it but as long as the way we construct dates and the way we format dates both use the same time zone (either local or UTC), then time zones shouldn't be an issue. Could you check that's the case?

And until we implement a way to set locale I think the DateTimeFormatter will also use the machine's locale. If that's the case then some of these tests may fail when the OS setting isn't en-us. Could you either try that or think through whether it's possible?

If any of them don't pass in other locations then we can talk about options. We may be willing to live with it until we implement the design token for setting locale. Or we may want to see if we can add a workaround for now.

jattasNI (approved):

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
m-akinc: Did a little StackOverflow searching to find the right syntax, but I got it working. Now we define types like this:
`export type WeekdayFormat = Intl.DateTimeFormatOptions['weekday'];`

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
rajsite: Yep parameterized tests should use whatever the current pattern is, which is currently getSpecTypeByNamedList. Don't think it's worth blocking PRs on the new pattern. After the new pattern is merged tests should prefer it.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
m-akinc: There's only one test that sets a specific output time zone, and I've already updated the input time string to include that same time zone. The rest of the tests will use the same (local) time zone for the input and output dates, so there shouldn't be a problem when running in different time zones.

On the other hand, the tests expect specific, locale-dependent, formatted strings. This breaks if the locale is something other than 'en-us'. I think we will want to explicitly set the locale to 'en-us' in the tests once we support that. I don't think we need to do anything special until then, though.

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
jattasNI: Awesome! You just made my day 😁

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
jattasNI: I have my locale set to `en-ca` (Canada) which uses "p.m." instead of "PM". When I cloned your branch and ran the tests locally, they passed! I haven't debugged how that's possible: playing with DateTimeFormat on mdn definitely shows it using my local machine locale. Anyway, I agree that it's ok to keep as-is until we add the locale feature.

m-akinc: In Windows, I found that they all pass if I just change my "Country or region" to Canada (with "Regional format" set to Recommended), but if I explicitly set my "Regional format" to English (Canada), the tests failed. How are you setting your locale? Is it just for the OS, or for all programs?

jattasNI: I have it set in macOS System Preferences. I'm not sure how consistent apps are at following that.

Further confusing matters

- the sample dates in the settings dialog use AM 
- the DateTimeFormatter with a locale of `en-ca` or `undefined` in Safari uses AM
- the DateTimeFormatter with a locale of `en-ca` or `undefined` in Firefox uses a.m.
 
I think we should stop looking at this before we get sucked in any deeper :-D

image

packages/nimble-components/src/table-column/date-text/tests/types.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    it('TimeStyle fails compile if assigning arbitrary string values', () => {
+        // @ts-expect-error This expect will fail if the enum-like type is missing "as const"
+        const format: TimeStyle = 'hello';
+        expect(format).toEqual('hello');
+    });
+
+    it('HourCycle fails compile if assigning arbitrary string values', () => {
+        // @ts-expect-error This expect will fail if the enum-like type is missing "as const"
+        const format: HourCycle = 'hello';
```
rajsite, line 96: make sure the references in this file get updated, the expect error may mask that

rajsite (approved):

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -31,6 +124,141 @@ export class TableColumnDateText extends TableColumnTextBase {
...
+        if (!this.format) {
+            options = {
+                dateStyle: 'medium',
+                timeStyle: 'medium'
+            };
+        } else {
+            options = this.getCustomFormattingOptions();
+        }
+        try {
+            return new Intl.DateTimeFormat(undefined, options);
```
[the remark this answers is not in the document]
m-akinc: I've got a follow-on PR tee'd up that add support for setting the locale via `nimble-theme-provider`. In those changes, I update the tests to run under a specific locale ('en'). Since that should be along shortly, can we just wait for that?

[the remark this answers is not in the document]
rajsite: Are you saying we should not update the karma config or GitHub actions settings? I think we should anyway and should have tests that show it can be unconfigured and pickup the system locale.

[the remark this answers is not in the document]
m-akinc: Added the locale flag to our karma config. I'll do the GitHub actions changes separately, as suggested.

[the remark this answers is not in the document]
m-akinc: Created PR https://github.com/ni/nimble/pull/1409 for the GitHub actions changes.

PR #1384, opened 2023-07-24
Fix tabs and anchor-tabs styling to allow tabs-toolbar content to be right aligned

mollykreis:
# Pull Request

## 🤨 Rationale

When fixing #451 (PR #1353), the `start` slot of the `tabs` and `anchor-tabs` were hidden, but the `grid-template-columns` styling of each component didn't get updated to account for not having `start` content. As a result, any usage of the `tabs-toolbar` for right-aligned buttons broke because the right-aligned buttons became left-aligned.

## 👩‍💻 Implementation

Updated `grid-template-columns` styling of the `tabs` and `anchor-tabs` to account for those components no longer having a supported `start` slot.

## 🧪 Testing

- Added test case to the `tabs` and `anchor-tabs` matrix tests to show right-aligned content.
- Manually verified that the new buttons added to the matrix tests renders in the correct location (previously it did not)

## ✅ Checklist


packages/nimble-components/src/anchor-tabs/tests/anchor-tabs-matrix.stories.ts
```diff
@@ -37,6 +37,7 @@ const component = (
         ${when(() => toolbar, html`
             <${tabsToolbarTag}>
                 <${buttonTag} appearance="ghost">Toolbar Button</${buttonTag}>
+                <${buttonTag} appearance="ghost" style="margin-left: auto;">Right-aligned Button</${buttonTag}>
```
rajsite, line 40: hmmm should our tabs toolbar align better with the toolbar and support the end slot idea natively?

Was wondering if we should get rid of tabs-toolbar altogether and have an appearance mode on toolbar but I guess they behave pretty different and it's not clear that the normal toolbar interactions are preferred:
Toolbar:
- start, default, and end slots all available
- uses arrow keys to move between items
- has own aria role defined

Tabs toolbar:
- only default slot
- just does simple layout of elements, no custom keyboard interactions
- no aria role defined (maybe fine since it's kinda transparent, the buttons slotted in in are just next in dom order anyway)

Scanning other usages in systemlink:
- Would have been nice to specify standard gap between items slotted in toolbar
- See use cases for specifying a start and end but maybe a default / middle is useful? I can't tell what's going on in [this UI](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/LabManagement/src/app/work-orders/components/work-order-details/work-order-details.component.scss&version=GBmaster&line=38&lineEnd=39&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) or where to find it deployed.

If it seems like it's useful for the tabs-toolbar to have more slots styled, I'm assuming we wouldn't want to do it in this PR and could make a separate tech debt? (unless it does make sense to tackle now).

rajsite (approved):

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs-matrix.stories.ts
jattasNI: You can see the UI you linked at `/labmanagement/workorders?ff-workorders=&ff-testplans=` and then click on any of the work orders in the grid to get to its details page. I didn't look closely at the CSS you linked but here's what the toolbar looks like.

image

jattasNI: If we decide to change this, I agree it'd likely make sense as a follow up.

jattasNI (approved):

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs-matrix.stories.ts
mollykreis: @rajsite brought up valid questions, and it isn't clear to me either whether or not the usages of the `tabs-toolbar` today should behave like a toolbar from an accessibility point of view. I'd really like to get this fix in sooner rather than later since our apps have incorrect styling without it. I've created an issue to figure out the answer to these questions and implement it: #1385

PR #1388, opened 2023-07-25
Per-component accessibility doc updates

### LINKED ISSUE #1090: Unified strategy for providing localized labels within components

### LINKED ISSUE #1393: Fix validateDOMNesting warnings that appear in some Storybook stories (dev/debug builds only)
## 🧹 Tech Debt / Issue
Currently if you run `npm run storybook` locally, some of our stories print out warnings in DevTools:
- `Warning: validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.`
- `Warning: validateDOMNesting(...): <h4> cannot appear as a descendant of <p>.`
- `Warning: validateDOMNesting(...): <summary> cannot appear as a descendant of <p>.`
- `Warning: validateDOMNesting(...): <details> cannot appear as a descendant of <p>.`

These don't appear in the release builds - however they add extra noise in debug builds and are easy to address.

These errors currently happen in for these stories: **Anchor Button, Button, Checkbox, Spinner, Toggle Button, Wafer Map**

### Problem
The Storybook rendering code is being strict and pointing out that having elements like `<ul>` or `<details>` inside `<p>` elements is [technically invalid HTML](https://html.spec.whatwg.org/multipage/grouping-content.html#:~:text=List%20elements%20(in%20particular%2C%20ol,it%20should%20be%20marked%20up. ).

The rendering code also automatically adds `<p>` tags in some cases, so it's easiest to find the problematic DOM structure via DevTools (Elements view).

### Problematic Examples
```
Some text without an explicit 'p' tag
<ul>
<li>List Item</li>
</ul>
```

```
Some text without an explicit 'p' tag
<details>
<summary>My summary</summary>
</details>
```

```
<details>
<summary>My summary</summary>
Some text without an explicit 'p' tag
<ul>
<li>List Item</li>
</ul>
</details>
```

### Solutions
```
<p>Some text with an explicit 'p' tag</p>
<ul>
<li>List Item
</ul>
```

```
Some text with an automatically created 'p' tag (extra newline before content that follows)

<details>
<summary>My summary</summary>
</details>
```

```
<details>
<summary>My summary</summary>
Some text with an automatically created 'p' tag (extra newline before content that follows)

<ul>
<li>List Item</li>
</ul>
</details>
```

msmithNI:
# Pull Request

## 🤨 Rationale

Resolves #1090

In [PR discussion on #1272](https://github.com/ni/nimble/pull/1272#discussion_r1215890160), we decided that the per-component accessibility notes would be better documented in Storybook (by the other component docs), instead of in accessibility.md.

Resolves #1393 

## 👩‍💻 Implementation

For components with accessibility notes not already documented in Storybook:
- Create an mdx docs page for them, if it wasn't already present. Based on PR comments on the recent PR by Jesse, I think this is the general direction we're going for component docs.
- Add a Usage Docs / Accessibility section at the bottom, and add the notes from accessibility.md. The button/spinner mdx have these sections, so again following the existing patterns.

## 🧪 Testing

Manual verification in Storybook.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/docs/accessibility.md
+++ b/packages/nimble-components/docs/accessibility.md
@@ -23,21 +23,3 @@ export const template = html<CoolNewButton>`
 
 ## Setting ARIA attributes
 When setting [ARIA attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes) on nimble components, avoid using id-based attributes, such as `aria-labelledby`. The id-based attributes may not behave as expected when used with Shadow DOM components, such as `nimble`.
-
-# Component-specific accessiblity notes
-
-## Buttons
-`nimble-button`, `nimble-anchor-button`, `nimble-menu-button`, `nimble-toggle-button`
-
-When using the Nimble buttons with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally to ensure the button is adequately accessible with a screen reader.
-
-## nimble-banner
-When using `nimble-banner`, the title content should always be provided for accessibility. To hide the title visually, `titleHidden` can be set to `true`.
-
-## nimble-dialog
-When using `nimble-dialog`, the title content should always be provided for accessibility. To hide the title visually, `headerHidden` can be set to `true`.
-
-## Icons
-When using an icon standalone, the `title` attribute should usually be set, to provide accessible text (which will also show as a tooltip).
-
-When using an icon as the content of a Nimble button, place the icon in the `start` slot, and follow the button guidance above.
--- a/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
+++ b/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
@@ -13,7 +13,7 @@ import * as anchorButtonStories from './anchor-button.stories';
 <Title of={anchorButtonStories} />
 <Description of={anchorButtonStories} />
 
-<DocsStory of={anchorButtonStories.outlineAnchorButton} />
+<DocsStory of={anchorButtonStories.outlineAnchorButton} expanded={false} />
 <Controls of={anchorButtonStories.outlineAnchorButton} />
 <Stories of={anchorButtonStories} />
 
--- a/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
+++ b/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
@@ -6,6 +6,7 @@ import {
     Title,
     Description
 } from '@storybook/blocks';
+import ContentHiddenDocs from '../../patterns/button/tests/content-hidden-docs.mdx';
 import * as anchorButtonStories from './anchor-button.stories';
 
 <Meta of={anchorButtonStories} />
@@ -15,12 +16,9 @@ import * as anchorButtonStories from './anchor-button.stories';
 
 <DocsStory of={anchorButtonStories.outlineAnchorButton} expanded={false} />
 <Controls of={anchorButtonStories.outlineAnchorButton} />
-<Stories of={anchorButtonStories} />
 
 # Usage Docs
 
 ## Accessibility
 
-When using the button with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button
-even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally
-to ensure the button is adequately accessible with a screen reader.
+<ContentHiddenDocs buttonElement="nimble-anchor-button" />
--- a/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
+++ b/packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
@@ -1,0 +1,26 @@
+import {
+    Controls,
+    DocsStory,
+    Meta,
+    Stories,
+    Title,
+    Description
+} from '@storybook/blocks';
+import * as anchorButtonStories from './anchor-button.stories';
+
+<Meta of={anchorButtonStories} />
+
+<Title of={anchorButtonStories} />
+<Description of={anchorButtonStories} />
+
+<DocsStory of={anchorButtonStories.outlineAnchorButton} />
+<Controls of={anchorButtonStories.outlineAnchorButton} />
+<Stories of={anchorButtonStories} />
+
+# Usage Docs
+
+## Accessibility
+
+When using the button with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button
+even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally
+to ensure the button is adequately accessible with a screen reader.
--- a/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
+++ b/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
@@ -2,7 +2,7 @@ export const appearanceDescription = `This attribute affects the appearance of t
 
 <details>
     <summary>Appearance Usage</summary>
-    There is a hierarchy to which button appearance should be used.
+    <p>There is a hierarchy to which button appearance should be used.</p>
     <ul>
         <li>For appearance \`ghost\`:
             <br/>This is the default and standard option for buttons. They create an open / airy feel, and fit comfortably in tight spaces.
@@ -27,7 +27,7 @@ export const appearanceVariantDescription = `This attribute has no effect on but
 
 <details>
     <summary>Primary Button Usage</summary>
-    Make a button primary to distinguish it visibly for one of the following reasons:
+    <p>Make a button primary to distinguish it visibly for one of the following reasons:</p>
     <ul>
         <li>to indicate the action that allows the user to accomplish their most common or important goal</li>
         <li>to indicate the action that allows the user to complete their task</li>
@@ -41,7 +41,7 @@ export const iconDescription = `When including an icon, set \`slot="start"\` on 
 
 <details>
     <summary>Icon Usage</summary>
-    Icons should be consistent and only used when necessary.
+    <p>Icons should be consistent and only used when necessary.</p>
     <ul>
         <li>Icon and text buttons: 
             <br/>Use a label plus icon to help reinforce a button's functionality, and increase the visual impact of the action.  
--- a/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
+++ b/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
@@ -14,7 +14,8 @@ export const appearanceDescription = `This attribute affects the appearance of t
         <li>For appearance \`outline\`:
             <br/>This is an alternative standard button when a ghost button is not suitable, and a more visually direct button is needed. 
             <br/>This can also create a light and airy feel. Use in combination with ghost buttons to create hierarchy. Do not use in combination with block buttons.
-        </li>                
+        </li>
+        <br/>
         <li>For appearance \`block\`:
             <br/>This standard button should be used when the most visible solution is required. 
             <br/>It can be an alternative to overly subtle button solutions when it is important to emphasize an action and the functionality of the control.
--- a/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
+++ b/packages/nimble-components/src/patterns/button/tests/doc-strings.ts
@@ -2,7 +2,8 @@ export const appearanceDescription = `This attribute affects the appearance of t
 
 <details>
     <summary>Appearance Usage</summary>
-    <p>There is a hierarchy to which button appearance should be used.</p>
+    There is a hierarchy to which button appearance should be used.
+
     <ul>
         <li>For appearance \`ghost\`:
             <br/>This is the default and standard option for buttons. They create an open / airy feel, and fit comfortably in tight spaces.
@@ -27,7 +28,8 @@ export const appearanceVariantDescription = `This attribute has no effect on but
 
 <details>
     <summary>Primary Button Usage</summary>
-    <p>Make a button primary to distinguish it visibly for one of the following reasons:</p>
+    Make a button primary to distinguish it visibly for one of the following reasons:
+
     <ul>
         <li>to indicate the action that allows the user to accomplish their most common or important goal</li>
         <li>to indicate the action that allows the user to complete their task</li>
@@ -41,7 +43,8 @@ export const iconDescription = `When including an icon, set \`slot="start"\` on 
 
 <details>
     <summary>Icon Usage</summary>
-    <p>Icons should be consistent and only used when necessary.</p>
+    Icons should be consistent and only used when necessary.
+
     <ul>
         <li>Icon and text buttons: 
             <br/>Use a label plus icon to help reinforce a button's functionality, and increase the visual impact of the action.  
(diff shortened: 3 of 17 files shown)


packages/nimble-components/docs/accessibility.md
```diff
@@ -23,21 +23,3 @@ export const template = html<CoolNewButton>`
...
-
-# Component-specific accessiblity notes
-
-## Buttons
-`nimble-button`, `nimble-anchor-button`, `nimble-menu-button`, `nimble-toggle-button`
-
-When using the Nimble buttons with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally to ensure the button is adequately accessible with a screen reader.
-
-## nimble-banner
-When using `nimble-banner`, the title content should always be provided for accessibility. To hide the title visually, `titleHidden` can be set to `true`.
```
msmithNI, line 35: The banner and dialog already have this general guidance captured in Storybook, in the API doc rows for titleHidden / headerHidden, so currently I didn't create mdx pages for them with the Usage Docs / Accessibility section.

If you think it's still better to add those docs/sections for them, in addition to what's in the API section (or maybe replacing some of the titleHidden/headerHidden API doc text), let me know.

packages/nimble-components/docs/accessibility.md
```diff
@@ -23,21 +23,3 @@ export const template = html<CoolNewButton>`
...
-
-## nimble-dialog
-When using `nimble-dialog`, the title content should always be provided for accessibility. To hide the title visually, `headerHidden` can be set to `true`.
-
-## Icons
-When using an icon standalone, the `title` attribute should usually be set, to provide accessible text (which will also show as a tooltip).
-
-When using an icon as the content of a Nimble button, place the icon in the `start` slot, and follow the button guidance above.
```
msmithNI, line 43: Similar to above, I thought that this line was already documented sufficiently by the `icon` API doc row for buttons in Storybook.

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
```diff
@@ -0,0 +1,26 @@
...
+    Description
+} from '@storybook/blocks';
+import * as anchorButtonStories from './anchor-button.stories';
+
+<Meta of={anchorButtonStories} />
+
+<Title of={anchorButtonStories} />
+<Description of={anchorButtonStories} />
+
+<DocsStory of={anchorButtonStories.outlineAnchorButton} />
```
msmithNI, line 16: I noticed that for the components we've converted to mdx docs instead of autodocs:
- their main story background is white instead of light gray; and
- the theme switcher at the top of the page doesn't update their main story background, so the dark/color themes don't render well:

ComponentMdxThemeColorIssues

After going through some Storybook code and experimenting, I think the easiest way to fix that issue is to use `DocsStory` instead of `Canvas` or `Story` for the main story in mdx docs.

Background info:
- The Storybook withBackground decorator, for docs view rendering, [creates a selector](https://github.com/storybookjs/storybook/blob/0792ef2db7e81a2719e24c9d587963eb8b4e8e47/code/addons/backgrounds/src/decorators/withBackground.ts#L38) that assumes that there will be an `Anchor` as the parent of the main story's `Canvas`. Our mdx docs aren't set up that way (but the autodocs are).
- However, we get that for free if we use [DocsStory](https://github.com/storybookjs/storybook/blob/265f21491c7c01cffaa5db569bd967185ff8d3c3/code/ui/blocks/src/blocks/DocsStory.tsx#L23), which has the `Anchor` and a child `Canvas` for the main story.

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -2,7 +2,7 @@ export const appearanceDescription = `This attribute affects the appearance of t
 
 <details>
     <summary>Appearance Usage</summary>
-    There is a hierarchy to which button appearance should be used.
+    <p>There is a hierarchy to which button appearance should be used.</p>
```
msmithNI, line 5: When I was running `npm run storybook` locally, I saw warnings/errors in dev tools about this DOM structure:
```
Warning: validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.
    at ul
    at p
    at details
    at div
    at __WEBPACK_DEFAULT_EXPORT__ (http://localhost:6006/vendors-node_modules_storybook_blocks_dist_index_mjs.iframe.bundle.js:3340:12874)
    at div
    at http://localhost:6006/vendors-node_modules_storybook_blocks_dist_index_mjs.iframe.bundle.js:1608:2994
    at td
    at tr
    at ArgRow (http://localhost:6006/vendors-node_modules_storybook_blocks_dist_index_mjs.iframe.bundle.js:472:8207)
    at tbody
    at table
```
It looks like once it's rendered, a `<p>` gets automatically added around these lines, and the `<ul>` ends up as child of `<p>`, which is [technically invalid](https://html.spec.whatwg.org/multipage/grouping-content.html#:~:text=List%20elements%20(in%20particular%2C%20ol,it%20should%20be%20marked%20up. ):
ButtonTextSS

So I added the `<p>` around just the text and the errors went away (renders the same, the `p` and `ul` become siblings)

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
```diff
@@ -0,0 +1,26 @@
...
+<Title of={anchorButtonStories} />
+<Description of={anchorButtonStories} />
+
+<DocsStory of={anchorButtonStories.outlineAnchorButton} expanded={false} />
+<Controls of={anchorButtonStories.outlineAnchorButton} />
+<Stories of={anchorButtonStories} />
+
+# Usage Docs
+
+## Accessibility
```
jattasNI, line 22: Interestingly, I just recently had to explain this behavior in [an SLE PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/528023#1690318766). Being able to point at these docs would have been great (I'd forgotten about accessibility.md).

What do you think about calling out the `title` guidance here too? I realize it's included in the API docs above and it's not necessarily part of accessibility, but I bet it would help people if we added something like "The text content should typically match the value provided for the `title` attribute."

Same for the other button components if you agree here.

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+When using the button with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button
```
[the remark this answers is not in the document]
jattasNI: Equivalent feedback applies to the other button files.

jattasNI (approved):

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
```diff
@@ -0,0 +1,26 @@
...
+import * as anchorButtonStories from './anchor-button.stories';
+
+<Meta of={anchorButtonStories} />
+
+<Title of={anchorButtonStories} />
+<Description of={anchorButtonStories} />
+
+<DocsStory of={anchorButtonStories.outlineAnchorButton} expanded={false} />
+<Controls of={anchorButtonStories.outlineAnchorButton} />
+<Stories of={anchorButtonStories} />
```
rajsite, line 18: Everytime I change the theme I'm noticing the list of stories grow longer on the doc page. All the stories seem to get re-added to the page. Seems like a bug to report to storybook.

I'm also skeptical of the value of placing this list of stories on the docs page. Seems like the DocsStory and the Controls give the right value. I'd opt for reporting the issue to storybook to be good citizens and removing the Stories from our doc pages. @jattasNI @msmithNI 

image

packages/nimble-components/src/toggle-button/tests/toggle-button.mdx
```diff
@@ -0,0 +1,25 @@
...
+<Controls of={toggleButtonStories.outlineButton} />
+<Stories of={toggleButtonStories} />
+
+# Usage Docs
+
+## Accessibility
+
+When using the button with only an icon, `contentHidden` should be set to `true` and text content should be provided within the button
+even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally
+to ensure the button is adequately accessible with a screen reader.
```
rajsite, line 25: mdx files are components, so we can create re-usable mdx components for these shared docs
See "MDX files are components": https://mdxjs.com/docs/using-mdx/#components

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
```diff
@@ -41,7 +41,7 @@ export const iconDescription = `When including an icon, set \`slot="start"\` on
 
 <details>
     <summary>Icon Usage</summary>
-    Icons should be consistent and only used when necessary.
+    <p>Icons should be consistent and only used when necessary.</p>
```
rajsite, line 44: Can you give the same treatment to our remaining usages of summary / detail in other stories. Makes it more likely folks will copy the right pattern in the future.

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
jattasNI: Good catch! I'm ok with removing the stories; they also take up a lot of vertical space so make it less likely that readers will make it to the bottom of the page to see the usage guidance, which is likely more important.

FYI @fredvisser since you're also thinking about reorganizing docs pages.

msmithNI: Removed `<Stories />` from the mdx files

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
msmithNI: Updated the remaining problematic ones.

Basically it looks like if there's a line of text then `<details>` or `<summary>` as the immediate next line, that triggers the error (as they all get combined into a single `<p>`). If there's a blank line inbetween it becomes a separate paragraph from the `details`/`ul` so it's fine.

I went with the newline approach for text already written as template literal strings since I thought that it's easier to read than the explicit `<p></p>` tags.

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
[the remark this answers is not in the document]
msmithNI: Updated the first button reference in the text to use the tag name. I think once is sufficient (so kept later sentences to still just say 'button').

packages/nimble-components/src/toggle-button/tests/toggle-button.mdx
msmithNI: Split this text out into `patterns/button/tests/content-hidden-docs.mdx` and referenced it as a component in the button mdx files.

packages/nimble-components/src/patterns/button/tests/doc-strings.ts
rajsite: Can I instead advocate for always having the explicit `<p></p>` tags? It's easier in review to always check the child text is wrapped in a tag then to try and figure out when it is or is not necessary.

> Makes it more likely folks will copy the right pattern in the future.
> > I went with the newline approach for text already written as template literal strings since I thought that it's easier to read than the explicit `<p></p>` tags.

My goal was consistency, I still think the newline approach will be easy for people to copy and paste tweak and mess-up because there are inconsistent patterns used to copy and paste.

rajsite: Unmarking resolve

msmithNI: Discussed this some offline. The issue is a bit more complicated than just 'ensure text inside `<details>` is in a `<p>` tag'. The warning can be printed when not using `<details>` (e.g. text then `<ul>`), and plain text in `<details>` is fine (unless again it's followed by something like `<ul>`).

Settled on filing a new issue ( #1393 ) that documents the warnings with some examples + fixes.

rajsite (approved):

rajsite:
The wafer map changes are straightforward doc formatting changes so I don't think we need to block on reviews from @DStavilaNI @zszilagy (but y'all feel free to pipe in if we need to address something in a follow-up)

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
msmithNI: Discussed and this was referring to the content about title/tooltip for icon-only buttons. Currently that's captured in the "Icon Usage" section of the 'icon' API for buttons. Agreed that it makes sense to include that in this Accessibility section too.

Now the button accessibility text looks like the following:
<hr/>

When using the `nimble-button` with only an icon:
- Add the `content-hidden` attribute and provide text content within the button even though it will not be visible on the screen. The button will use the text content to configure the appropriate ARIA attributes internally to ensure the button is adequately accessible with a screen reader.
- Consider providing a tooltip to provide extra information (with the `title` attribute). Typically, the `title` attribute should match the text content in the button.

rajsite: > typically, the title attribute should match the text content in the button

Wait is that accurate advice? My uunderstanding is the title / tooltip should represent additional context while the aria label should describe the button.

icon: 📈
aria: buy stock
title / tooltip: buying stock increases the amount of stock you own

msmithNI: I took out that line. I don't think it's always true. In some cases title will match (that's what we did for the table Collapse All button because we didn't think the icon would sufficiently convey that). But in other cases (well-known icon) it makes sense that the tooltip would be additional information. So it's probably case-by-case.

Issue #1103, opened 2023-03-07
Table allows configuration of boolean column type


atmgrifter00:
## 😯 Problem to Solve
For some columns in the table, there is a need to show an item as a boolean value. This could be visualized as either text or as an icon.

## 💁 Proposed Solution
None at the moment. It could be that we solve this through a combination of other defined work:
- #1011 (see [HLD](https://github.com/ni/nimble/pull/1054): possibly allowing custom formatting of various types...could solve how to display boolean text in a localized way)
- #1013 (allow clients to specify the icons to use for each boolean value?)

## Open Questions
| Question | Initial Recommendation | Resolved | Decision |
|:---|:---|:---|:---|
| What icons should we use to represent True/False? Should this be user-configurable?| Icons and boolean text should be supplied by clients | Yes |  |

## Design
- TODO

```[tasklist]
## 📋 Tasks
- [ ] TBD
```

PR #1392, opened 2023-07-26
Enum-text table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1103

## 👩‍💻 Implementation

Created new `nimble-table-column-enum-text` component plus the supporting elements `nimble-mapping-text`.

These changes include some refactors (by Milan and myself) from the original design. Base classes for the column and mapping element are in anticipation of the icon column and its associated mapping elements.

Angular and Blazor support will be implemented in follow-on PRs.

## 🧪 Testing

Unit tests written for all new components. Tested in Storybook.

## ✅ Checklist


packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.stories.ts
```diff
@@ -16,20 +16,11 @@ import {
 import { tableColumnTextTag } from '../../text';
 import { AnchorAppearance } from '../../../anchor/types';
 
-const columnTypeOverviewText = `This page contains information about the types of columns that can be displayed in a \`nimble-table\`.
-See the **Table** page for information about configuring the table itself and the **Table Column Configuration** page for
-information about common column configuration.`;
-
 const metadata: Meta<SharedTableArgs> = {
-    title: 'Incubating/Table Column Types',
+    title: 'Incubating/Table Column - Anchor',
```
m-akinc, line 20: We rename all the table column stories so they are directly under Incubating.

packages/nimble-components/src/table-column/base/index.ts
m-akinc, line 1: @rajsite can you comment here why this change was necessary?

packages/nimble-components/src/mapping/base/index.ts
```diff
@@ -0,0 +1,14 @@
+import { attr } from '@microsoft/fast-element';
+import { FoundationElement } from '@microsoft/fast-foundation';
+import type { MappingKey } from './types';
+
+/**
+ * Base class for mapping configuration elements
+ */
+export abstract class Mapping extends FoundationElement {
+    @attr({ mode: 'fromView' })
```
atmgrifter00, line 9: As this is a departure from how we normally declare attributes (i.e. using the `fromView` mode), do you think it would make sense to provide a comment as to the reasoning behind using that?

packages/nimble-components/src/mapping/base/tests/story-helpers.ts
```diff
@@ -0,0 +1,3 @@
+export function keyDescription(renderedThing: string): string {
```
atmgrifter00, line 1: I presume this is being exported here with the expectation that there will be a separate stories file that will leverage the same description?

packages/nimble-components/src/table-column/enum-base/index.ts
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Base class for table columns that map values to content (e.g. nimble-table-column-enum-text and nimble-table-column-icon)
+ */
+export abstract class TableColumnEnumBase<
+    TColumnConfig extends TableColumnEnumColumnConfig,
+    TEnumValidator
+>
+    extends TableColumn<TColumnConfig>
+    implements Subscriber {
+    /** @internal */
```
atmgrifter00, line 38: nit: Can we move `internal` members below non-internal public members?

packages/nimble-components/src/mapping/text/index.ts
```diff
@@ -0,0 +1,21 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-mapping-text': MappingText;
+    }
+}
+
+/**
+ * Maps data values to text.
+ */
+export class MappingText extends Mapping {}
```
atmgrifter00, line 14: On the surface (i.e. from a constrained view of just what is present in this PR), this comes across as odd, in that the `MappingText` offers no API, and is just a concrete version of the API presented by `Mapping`, which has no additional derivatives. My presumption is that ultimately there will be other necessary `Mapping` derivatives that have their own API needs, but what isn't clear is whether or not the `Mapping` APIs (notably `label`) are necessary for those derivatives. I see the `label` being used as the display content, and so I'm wondering if we would still being using that for a separate `Mapping` type.

packages/nimble-components/src/mapping/base/index.ts
```diff
@@ -0,0 +1,14 @@
...
+    public key?: MappingKey;
+
+    @attr()
+    public label?: string;
```
atmgrifter00, line 13: Has there already been discussion on the proper name for this attribute, as `label` just seems like an overloaded term that seems a little odd here. Other options I can think of are `display`, `value`, or `text`. I do _not_ feel strongly about this, and `label` can certainly suffice. It just struck me as a slightly odd term.

packages/nimble-components/src/mapping/base/index.ts
```diff
@@ -0,0 +1,14 @@
+import { attr } from '@microsoft/fast-element';
+import { FoundationElement } from '@microsoft/fast-foundation';
+import type { MappingKey } from './types';
+
+/**
+ * Base class for mapping configuration elements
+ */
+export abstract class Mapping extends FoundationElement {
+    @attr({ mode: 'fromView' })
+    public key?: MappingKey;
```
atmgrifter00, line 10: In my ideal world, `MappingKey` would be provided via a generic on the type, so we could declare `MappingText` as `MappingText<string>`, so that it wouldn't be possible in TS to assign `key` on `MappingType` a value of `type number`. However, I don't think there is a means to enforce that the generic type only be a type as defined by `MappingKey` (doing `Mapping<T extends MappingKey>` opens up the type to non-supported types, which we don't want).

All of this is to simply say, I guess this is the best we can do?

packages/nimble-components/src/mapping/text/index.ts
atmgrifter00, line 14: Given that this class is just a thin, concrete wrapper around the `Mapping` class, it's unclear how this class participates in actually mapping a data value to a text value. Can this be explained in the documentation for the class?

packages/nimble-components/src/mapping/base/tests/story-helpers.ts
m-akinc: Yep, since I already have both columns implemented, but I'm splitting up the changes, you may see stuff like this or seemingly unnecessary base classes, etc.

packages/nimble-components/src/table-column/enum-base/index.ts
m-akinc: Actually, we need this member to come first in the order. The members get initialized in order, and in the `*Changed()` handlers, we want the validator to be initialized already. I'll add a comment to that effect.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -0,0 +1,405 @@
...
+                'number'
+            ));
+            await connect();
+            await waitForUpdatesAsync();
+            const column = model.col1;
+            expect(column.checkValidity()).toBeFalse();
+            expect(column.validity.invalidMappingKeyValueForType).toBeTrue();
+        });
+
+        describe('is invalid with invalid boolean key values:', () => {
```
atmgrifter00, line 325: Shouldn't there be a test for valid boolean key values?

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The `nimble-table-column-mapping` is a component that supports rendering specific number, boolean, or string values as mapped text. `nimble-table-column-icon` is a specialized version of `nimble-table-column-mapping` that instead maps values to icons and/or spinners and has a minimal, fixed width. The actual mappings are defined by child elements `nimble-mapping-icon`, `nimble-mapping-spinner`, and `nimble-mapping-text`.
+The `nimble-table-column-enum-text` is a component that supports rendering specific number, boolean, or string values as mapped text. `nimble-table-column-icon` is a specialized version of `nimble-table-column-enum-text` that instead maps values to icons and/or spinners and has a minimal, fixed width. The actual mappings are defined by child elements `nimble-mapping-icon`, `nimble-mapping-spinner`, and `nimble-mapping-text`.
```
atmgrifter00, line 5: Is "`nimble-table-column-icon` is a specialized version of `nimble-table-column-enum-text`..." accurate, or should it be `nimble-table-column-icon` is a specialized version of `TableColumnEnumBase`?

packages/nimble-components/src/table-column/enum-base/models/table-column-enum-base-validator.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Validator for TableColumnEnumText. Implementations MUST include enumBaseValidityFlagNames in validity flag names set.
+ */
+export abstract class TableColumnEnumBaseValidator<
+    ValidityFlagNames extends readonly string[]
+> extends ColumnValidator<
+    typeof enumBaseValidityFlagNames | ValidityFlagNames
+    > {
+    public validate(
+        supportedMappingElements: readonly (typeof Mapping)[],
```
atmgrifter00, line 24: Do we even expect this set to be dynamic for a particular concrete column type? If not, this seems like a strange API, where we could instead provide this as a constructor parameter, and just have this state available on the class instance, instead of having to always pass some const as a parameter to a method.

packages/nimble-components/src/mapping/base/index.ts
m-akinc: Now I'm not sure what this really does for us, so I'm removing it.

packages/nimble-components/src/mapping/base/index.ts
m-akinc: I don't think anyone brought this up before. I see your point, and I lean slightly towards changing it to `text`. I'll leave this comment open and urge other reviewers (please) to weigh in.

packages/nimble-components/src/mapping/base/index.ts
atmgrifter00: All it's going to do is have it so that a programmatic change to this attribute isn't reflected back to the DOM. It's unclear to me if that would provide us any benefit.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
m-akinc: Removed that bit about it being a specialized version of something else.

packages/nimble-components/src/table-column/base/template.ts
m-akinc, line 1: This was done because we reuse the header template in the enum-base template. That template just augments this base one by adding a slot for the mapping elements (see `table-column/enum-base/template.ts`). Without this change, only part of the template would be wrapped in a `<template>` element, which is wrong.

packages/nimble-components/src/table-column/enum-text/index.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+export class TableColumnEnumText extends mixinGroupableColumnAPI(
+    mixinFractionalWidthColumnAPI(
+        TableColumnEnumBase<
+        TableColumnEnumColumnConfig,
+        TableColumnEnumTextValidator
+        >
+    )
+) {
+    protected get supportedMappingElements(): readonly (typeof Mapping)[] {
```
atmgrifter00, line 34: I'm wondering if it's really important for this member to be defined as an abstract member. Its only usage is to be passed as a constructor parameter, so it doesn't really have utility as a `protected` member (useful to derivations). The `createValidator()` implementation seems like it could easily just be:
```ts
    public override createValidator(): TableColumnEnumTextValidator {
        return new TableColumnEnumTextValidator(
            this.columnInternals,
            [MappingText]
        );
    }
```

atmgrifter00 (approved):

packages/nimble-components/src/table-column/enum-text/index.ts
```diff
@@ -31,15 +31,10 @@ export class TableColumnEnumText extends mixinGroupableColumnAPI(
...
-        return [MappingText] as const;
-    }
-
     public override createValidator(): TableColumnEnumTextValidator {
-        return new TableColumnEnumTextValidator(
-            this.columnInternals,
-            this.supportedMappingElements
-        );
+        return new TableColumnEnumTextValidator(this.columnInternals, [
+            MappingText
```
atmgrifter00, line 36: If prettier allows it, I think it would look best as:
```ts
        return new TableColumnEnumTextValidator(
            this.columnInternals,
            [MappingText]
         );
```

m-akinc: It does not 👎

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -0,0 +1,420 @@
...
+
+                    expect(
+                        pageObject.getRenderedGroupHeaderContent(0)
+                    ).toContain(value.name);
+                }
+            );
+        }
+    });
+
+    it('sets group header text to key value when unmatched (instead of blank)', async () => {
```
mollykreis, line 212: Sorry if I'm starting a discussion that has already been had and settled on, but it seems really odd to me that the group header would show the key value while the column itself would show blank. To me that sees very inconsistent. What's the rationale for the two different behaviors?

packages/nimble-components/src/table-column/enum-base/models/mapping-config.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Common state shared across Mapping Config
+ */
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+export abstract class MappingConfig {}
```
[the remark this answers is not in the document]
m-akinc: I had moved `label` into the shared mapping element, but forgot to move it in the mapping config.

packages/nimble-components/src/table-column/enum-base/index.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public abstract createValidator(): TEnumValidator;
+
+    /**
+     * Called when any Mapping related state has changed
+     */
+    protected abstract updateColumnConfig(): void;
+
+    // Assumes the mapping element state is validated
```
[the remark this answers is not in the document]
m-akinc: I don't see what that would improve. There would still be an assumption that validation had run. Checking if the validation state _isn't_ valid would not catch the case where validation hadn't been run. In a sense, `createColumnConfig` is double-checking that the state is valid (by checking `key !== undefined`) and doing something appropriate if it is not.  In `TableColumnEnumText.createMappingConfig()` we do also handle the case where our configuration is invalid by throwing an error. That's because we assume it would be a programming error, but maybe if we are loosening the assumption that validity has been checked, we can just return a `MappingTextConfig` with an empty string. Thoughts?

packages/nimble-components/src/table-column/enum-base/models/table-column-enum-base-validator.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.validateNoMissingKeys(mappings);
+        this.validateNoMissingLabels(mappings);
+    }
+
+    private validateKeyValuesForType(
+        keys: (MappingKey | undefined)[],
+        keyType: MappingKeyType
+    ): void {
+        const invalid = keys.some(
+            key => key !== undefined
```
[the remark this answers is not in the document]
m-akinc: Yes, I'll add the comment.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
mollykreis: Did we discuss at some point the sorting inconsistency when the cell shows blank but not all blanks are ordered together when sorting by that column? I think a similar argument to grouping needing to show the raw value could be made for needing to show the raw value when sorting. I'm not sure if that makes me lean towards showing the raw value in both cases or neither case, but I'm not convinced that grouping is a unique problem that should be solved in a special way.

jattasNI (approved):

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
m-akinc: That's a good point. I still feel there is a little more of an argument to do something special in the group header, since there is another column (icon, which hasn't been reviewed yet...) where the group header shows something different than the cell. Leslie wanted the icon _and_ its label to be displayed in the group header whereas the cells just show the icon. On the other hand, it would be exceedingly weird if sorting columns made displayed values change. If the general consensus is that we should change the group headers to be blank, like the cells, I'm okay with that.

packages/nimble-components/src/mapping/base/index.ts
m-akinc: I'm changing it to `text`.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+        lastName: '',
+        status: 'foo',
+        isChild: true
+    },
+    {
+        firstName: 'Frank',
+        lastName: '',
+        status: 'bar',
+        isChild: false
+    }
```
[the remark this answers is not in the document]
m-akinc: 1. These two entries were to display the behavior for rows that don't match a mapping. This is also part of why grouping is turned on. There's a behavior (being questioned by Molly in another thread) where the unmapped key value is shown in the group header instead of a blank. If we consider unmapped values more of a corner case, then I don't need to demonstrate it in the story.

2. I used to have a second column that used `isChild`, but I had removed it. Forgot to remove the record properties.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+    render: createUserSelectedThemeStory(html<EnumTextColumnTableArgs>`
+        ${incubatingWarning({ componentName: 'table', statusLink: 'https://github.com/orgs/ni/projects/7/views/21' })}
+        <${tableTag}
+            ${ref('tableRef')}
+            data-unused="${x => x.updateData(x)}"
+        >
+            <${tableColumnTextTag} field-name="firstName" >
+                Name
+            </${tableColumnTextTag}>
+            <${tableColumnEnumTextTag}
```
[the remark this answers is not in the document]
jattasNI: My (admittedly weak) justification is that you can the keys/mappings/key-type from the Code view of the Docs page. And (weaker still) if someone's looking at the story's source in the repo.

packages/nimble-components/src/table-column/enum-base/index.ts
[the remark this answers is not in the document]
mollykreis: I think there are a few things that are a bit confusing to me, and perhaps they can be addressed with minimal code changes.
- Is the assumption that _validation has run_? Or, is the assumption that _the column & its mappings are valid_? Can the language in the comments be more precise?
- The comment above `createMappingConfig` is confusing to me because it says that the function "assumes the mapping element state is validated", but there isn't an implementation associated with the function. Should the comment instead say something to the effect of "An implementation of this function may assume that the `mapping` passed in is valid"? If so, why are we doing a check for an allowed mapping type in our implementation? Should performing a check really be the guidance -- "throw an Error from this function if an invalid `Mapping` is passed"?
- It seems like the comments above the functions are intended to be documentation for the functions rather than comments associated with the code. If that's the case, can you switch them to using `/** */` style comments that document the function and also make note of any assumptions associated with the function?
- If every mapping column implementation is expected to check the validity of the configuration before calling `createColumnConfig()`, why not put the validity check there? I agree it doesn't guarantee that the caller has updated their validity, but I think that would still be better than having identical logic in every column implementation for `this.validator.isValid() ? this.createColumnConfig() : undefined`

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.stories.ts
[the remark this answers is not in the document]
jattasNI: Ah, of course, thanks,. Personally I'd vote that the unmapped case is uncommon enough that it isn't worth documenting here.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.stories.ts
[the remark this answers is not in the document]
m-akinc: I think it will look redundant for almost all users, so I'd vote for not adding them (and this goes for the icon column story that will be coming, too).

packages/nimble-components/src/table-column/enum-base/index.ts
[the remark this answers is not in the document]
m-akinc: 1. I pushed the `this.validator.isValid` check down into `createColumnConfig`, like you suggested.
2. I attempted to run the validation (using a new abstract method) in the base class (i.e. in `updateColumnConfig`), but a typing incompatibility makes that impractical. Instead, I've just updated the comments to be more directly instructive.

packages/nimble-components/src/table-column/enum-base/models/mapping-key-resolver.ts
```diff
@@ -0,0 +1,30 @@
...
+ * given keyType. The converted value can then be used to compare against
+ * values in the table data.
+ */
+export const resolveKeyWithType = (
+    key: MappingKey | undefined,
+    keyType: MappingKeyType
+): MappingKey | undefined => {
+    if (keyType === 'number') {
+        const converted = nullableNumberConverter.fromView(key) as
+            | number
```
atmgrifter00, line 16: Is that first `|` (in front of `number`) necessary?

m-akinc: Yes, the formatter adds it back if I remove it.

packages/nimble-components/src/table-column/enum-base/index.ts
```diff
@@ -0,0 +1,132 @@
...
+     * Validation state should be updated before calling this function.
+     */
+    protected createColumnConfig(): TableColumnEnumColumnConfig | undefined {
+        if (!this.validator.isValid) {
+            return undefined;
+        }
+        const mappingConfigs = new Map<MappingKey, MappingConfig>();
+        this.mappings.forEach(mapping => {
+            const key = resolveKeyWithType(mapping.key, this.keyType);
+            if (key !== undefined) {
```
mollykreis, line 92: Given all the other updates, would it make sense now to remove this `key !== undefined` check? Or, do you think it is still beneficial to keep it?

packages/nimble-components/src/table-column/enum-base/models/mapping-config.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Common state shared across Mapping Config
+ */
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
mollykreis, line 4: Can you stop disabling this lint rule now that the class isn't empty?

packages/nimble-components/src/table-column/enum-base/models/table-column-enum-base-validator.ts
```diff
@@ -0,0 +1,81 @@
...
+        keys: (MappingKey | undefined)[],
+        keyType: MappingKeyType
+    ): void {
+        const typedKeys = keys.map(x => resolveKeyWithType(x, keyType));
+        const invalid = new Set(typedKeys).size !== typedKeys.length;
+        this.setConditionValue('duplicateMappingKey', invalid);
+    }
+
+    private validateNoMissingKeys(mappings: Mapping[]): void {
+        const invalid = mappings.filter(mapping => mapping.key === undefined).length > 0;
```
mollykreis, line 73: While there probably aren't going to be tons of mappings, it would be more efficient to use `some` rather than `filter` so that you can stop looking once one invalid mapping key is found and to avoid making a copy of the array.

```suggestion
        const invalid = mappings.some(mapping => mapping.key === undefined);
```

packages/nimble-components/src/table-column/enum-base/models/table-column-enum-base-validator.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.setConditionValue('missingKeyValue', invalid);
+    }
+
+    private validateNoMissingText(mappings: Mapping[]): void {
+        const invalid = mappings.filter(mapping => mapping.text === undefined).length > 0;
```
mollykreis, line 78: Same here.

```suggestion
        const invalid = mappings.some(mapping => mapping.text === undefined);
```

packages/nimble-components/src/table-column/enum-text/group-header-view/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    private updateText(): void {
+        const value = this.groupHeaderValue;
+        if (value === undefined || value === null) {
+            this.text = '';
+            return;
+        }
+
+        const config = this.columnConfig?.mappingConfigs.get(value);
+        this.text = config ? (config as MappingTextConfig).text : '';
```
mollykreis, line 38: From what I can see, this logic appears to be the same as in the cell view but written differently. Can you pick one style and make them consistent so that it is more apparent that they are doing the same thing?

packages/nimble-components/src/table-column/enum-base/index.ts
m-akinc: I still think it's okay to have, especially because I would otherwise need to add a non-null assertion anyway.

rajsite: I think this can at least behave consistently with createMappingConfig. The idea being this is an impossible state since we have guarded the running of this function with validation so we should throw instead.

packages/nimble-components/src/table-column/enum-base/models/mapping-text-config.ts
```diff
@@ -0,0 +1,10 @@
+import { MappingConfig } from './mapping-config';
+
+/**
+ * Mapping configuration corresponding to a text mapping
+ */
+export class MappingTextConfig extends MappingConfig {
+    public constructor(text: string) {
+        super(text);
+    }
```
rajsite, line 9: If just forwarding everything through can just leave off the constructor

packages/nimble-components/src/table-column/enum-text/index.ts
```diff
@@ -0,0 +1,78 @@
...
+            groupHeaderViewTag: tableColumnEnumTextGroupHeaderViewTag,
+            delegatedEvents: [],
+            sortOperation: TableColumnSortOperation.basic
+        };
+    }
+
+    protected override updateColumnConfig(): void {
+        this.validator.validate(this.mappings, this.keyType);
+        this.columnInternals.columnConfig = this.createColumnConfig();
+    }
```
rajsite, line 57: I think at this point this is completely generic (and the other branch shows identical implementations for each column type). Can we move this to the base class?

Even more "correctly" it is validation that is column specific. So the validator execution could be overridable but the `updateColumnConfig` algorithm is always the same, i.e. run validation then create the columnConfig.

packages/nimble-components/src/table-column/enum-base/tests/shared.ts
```diff
@@ -0,0 +1,9 @@
+export const sharedMappingValidityDescription = `Readonly object of boolean values that represents the validity states that the column's configuration can be in.
```
rajsite, line 1: Can you name this file something more descriptive, like `shared-storybook-docs.ts`

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    private updateText(): void {
+        const value = this.cellRecord.value;
+        if (value === undefined || value === null) {
+            this.text = '';
+            return;
+        }
+
+        const config = this.columnConfig?.mappingConfigs.get(value);
+        this.text = config ? (config as MappingTextConfig).text : '';
```
rajsite, line 40: Treat the element more standalone and behave robustly. So an instanceof check instead of a cast / assertion

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
```diff
@@ -0,0 +1,52 @@
+import { DesignSystem } from '@microsoft/fast-foundation';
```
rajsite, line 1: We haven't been consistent on it for the cell view and group header view elements but every element should have those trivial registration check tests:

https://github.com/ni/nimble/blob/158b1d29e9cb8d6591662404ee4ed0921b3e1361/packages/nimble-components/src/breadcrumb/tests/breadcrumb.spec.ts#L3-L13

For the elements outside this PR could you capture those in a tech debt task and make sure we have them in a follow-up PR?

packages/nimble-components/src/table-column/enum-text/group-header-view/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
```
rajsite, line 39: same as cell view avoid the as cast and do an actual instanceof check

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -0,0 +1,418 @@
...
+
+        expect(pageObject.getRenderedCellContent(0, 0)).toBe('alpha');
+    });
+
+    describe('various string values render as expected', () => {
+        const focused: string[] = [];
+        const disabled: string[] = [];
+        for (const value of wackyStrings) {
+            const specType = getSpecTypeByNamedList(value, focused, disabled);
+            // eslint-disable-next-line @typescript-eslint/no-loop-func
```
rajsite, line 159: I think this eslint disable isnt needed

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -0,0 +1,418 @@
...
+            );
+        }
+    });
+
+    describe('various string values render in group header as expected', () => {
+        const focused: string[] = [];
+        const disabled: string[] = [];
+        for (const value of wackyStrings) {
+            const specType = getSpecTypeByNamedList(value, focused, disabled);
+            // eslint-disable-next-line @typescript-eslint/no-loop-func
```
rajsite, line 187: I think this disable also isn't needed?

packages/nimble-components/src/table-column/enum-text/index.ts
m-akinc: I had to make a few other changes to make this work properly. Now `createColumnConfig` is abstract and returns a generic type instead of `TableColumnEnumColumnConfig`. This way, classes derrived from `TableColumnEnumBase` can _actually_ have different column config types. Otherwise, we would get assignment errors when trying to do `this.columnInternals.columnConfig = this.createColumnConfig();`.

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
m-akinc: Could you explain what you mean by "treat the element more standalone?" I understand what change you're suggesting, but not the motivation.

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
m-akinc: I went ahead and added them in this PR.

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
rajsite: The cast for `as MappingTextConfig` is knowledge outside the component. The property accepts any type of MappingConfig based on it's type. Based solely on the component on it's own and it's types it can't assume that only MappingTextConfig are assigned.

packages/nimble-components/src/table-column/enum-text/cell-view/index.ts
rajsite: I guess that's fine for this PR since it's already done but really based on the history of this PR and the discussions we have been having for smaller PRs I would not recommend that in the future.

packages/nimble-components/src/mapping/base/index.ts
```diff
@@ -0,0 +1,22 @@
...
+export abstract class Mapping extends FoundationElement {
+    /**
+     * The data value that is mapped to another representation
+     */
+    @attr()
+    public key?: MappingKey;
+
+    /**
+     * In the case of a mapping that maps to text, this is that mapped text.
+     * For mappings to non-textual representations, this is a display string
```
rajsite, line 17: It's not displayed right? That's the whole distinction?

```suggestion
     * For mappings to non-textual representations, this is a string
```

rajsite (approved):

packages/nimble-components/src/mapping/base/index.ts
m-akinc: Well, no, it could be. For the icon/spinner mappings, the `text` value appears in the tooltip, and also next to the icon/spinner in the group header. So it should be formatted appropriately for display.

rajsite: "For mappings to non-textual representations, this is a string that can be displayed in contexts such as group headers or tooltips and used as an accessible label"

Issue #1331, opened 2023-06-29
Allow anchor components to indicate that they represent external links.


atmgrifter00:
## 😯 Problem to Solve

It can be helpful to provide visual context to a hyperlink that it is representing an external link. We should consider adding this ability to our anchors. Here is an example of such a visualization that Wikipedia uses:

![image](https://github.com/ni/nimble/assets/26874831/77d48942-eac9-4227-bf1a-a3e5ae87574e)

## 💁 Proposed Solution

Perhaps this is simply an extra attribute we provide for our anchor components (e.g. `external: boolean`). We would then just style the anchor to include the extra visual based on that attribute's presence.

leslieab:
From a Figma video I was watching.

![Image](https://github.com/ni/nimble/assets/28602266/ef5a6e14-73f4-4ef0-beb1-67d916f034ae)

fredvisser:
We _will_ support an icon to indicate that the link will open in a new tab/context (#1412), but won't support an icon for external links.

- If external links are opened in new tabs, they can use the new tab icon. 
- Using icons to indicate external links is not a common pattern.
- Detecting whether user generated links are external within frontend code is brittle.

PR #1394, opened 2023-07-27
Spec update to include absolute URL support for rich text editor

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

As a requirement for the comments feature, we need to support absolute URL links in both the `nimble-rich-text-editor` and `nimble-rich-text-viewer` components. Therefore, I am initiating this PR to add the necessary support for absolute URL links in the editor and viewer.

Related discussion on link support: https://github.com/ni/nimble/pull/1369#discussion_r1272857724

## 👩‍💻 Implementation

Updated spec to add information on the design for supporting absolute URL links in a rich text editor and viewer components.

## ✅ Checklist


vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,10 +352,67 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
+            [
+                iconUpRightFromSquareTag, { slot: 'end' }
+            ]
+        ];
+    }
+}
+```
+
+1.  As in the above schema, we will modify the `target` value according to the boolean attribute `externalLink` that is configured by the client component.
+2.  We also set the `rel` attribute value to `noopener noreferrer nofollow` to enhance security and ensure responsible linking practices.
```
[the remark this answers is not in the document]
vivinkrishna-ni: Sure, I have removed the `nofollow` from the default value.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,27 @@ _Props/Attrs_
 -   `markdown` - for retrieving and modifying the markdown value. If the client modifies the markdown value, it will be parsed into a Node using the
     [prosemirror-markdown parser](https://github.com/ProseMirror/prosemirror-markdown/blob/9049cd1ec20540d70352f8a3e8736fb0d1f9ce1b/src/from_markdown.ts#L199).
     The parsed node will then be rendered in the viewer component as rich text.
+-   `external-link` - is a boolean attribute determining whether all absolute links within the viewer component should open in a new tab. By default, all the links will open in the same tab
```
mollykreis, line 245: The name `external-link` here doesn't seem appropriate. At least in my experience, an "external" link is a link that will take you to a new site. But here, you are using that term to mean a link that will open in a new tab, which is significantly different.

An alternative could be to make this a string attribute called something like `link-target` where the value is one of the valid `target` values on anchor elements (i.e. `"_self" | "_blank" | "_parent" | "_top"`).

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,27 @@ _Props/Attrs_
...
+    as per the accessibility guidelines on WCAG (links are below). If opening in a new tab, `nimble-icon-up-right-from-square` will be placed right next to all the links
```
mollykreis, line 246: Do we have any designer guidance or approval on the use of `nimble-icon-up-right-from-square` to mean "this link will open in a new tab"? We should make sure visual designers agree that this is the correct icon to use, and we should also make sure they are aware that we'll ultimately need a way to indicate that a link will go to an external site (#1331 and #1412).

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,27 @@ _Props/Attrs_
...
+    in the viewer instance to indicate the link will be opened in a new tab.
+    -   Accessibility guidelines to open link only in new tab when required: <https://www.w3.org/TR/WCAG20-TECHS/G200.html>
+    -   Accessibility guidelines on opening a link in a new tab: <https://www.w3.org/TR/WCAG20-TECHS/G201.html>
+
+_Alternative implementations:_
+
+For comments feature in SLE, there is a requirement to have all links open in a new tab. However, we are also considering an alternative approach where links will open in a new tab only if
```
mollykreis, line 253: Where is the requirement for having all links open in a new tab coming from? Based on my understanding, that would go against the guidelines you linked right above this paragraph _Accessibility guidelines to open link only in new tab when required: <https://www.w3.org/TR/WCAG20-TECHS/G200.html>_.

Could you also clarify why we'd want some links to have different default behavior than other links? My initial thought is that will be confusing and seemingly inconsistent to the user, but perhaps I'm missing some precedent/guidelines we are following with that.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: I have updated the spec as per the suggestion.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: Thanks for bringing this up, Molly. 

Initially, Jesse helped us by getting some bandwidth of Rick Ashby for the visual design of rich text components since we have not worked with any visual designers so far. Looks like Rick is on a vacation from Jul 27 - Aug 15.

Are there any other visual designers helping out in #1412 who can quickly help on this, as this will be supporting both issues? 

It will be much helpful if we can align on a fallback strategy for this item, if we couldn't get any quick help from a visual designer based on their priorities.

packages/nimble-components/src/rich-text-editor/specs/README.md
suseendran-ni: Hi @mollykreis ,

As a part of the Comments requirement, the expectation is to open the links in new tab. Please find the below comment in the Feature - https://dev.azure.com/ni/DevCentral/_backlogs/backlog/ASW%20SystemLink%20Platform/Initiatives/?workitem=2383256

image


We observed, in Test Insights, in the grid. the SLE Links are opened in a different tab.
image

But when we explored how the Azdo PR comments work, we noticed that the links that belong to the same domain are opened in the same tab and the links that belong to the different domain are opened in a different tab
image

Hence we decided to add this in the `alternative implementation` section to get your suggestions

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,36 @@ _Props/Attrs_
...
     [prosemirror-markdown parser](https://github.com/ProseMirror/prosemirror-markdown/blob/9049cd1ec20540d70352f8a3e8736fb0d1f9ce1b/src/from_markdown.ts#L199).
     The parsed node will then be rendered in the viewer component as rich text.
+-   `link-target` - is an attribute that accepts valid `target` values for the `anchor` element, applicable to all links within the viewer component.
+    The following values are accepted by this attribute:
+    1.  `_self`: the current browsing context. (Default)
+    2.  `_blank`: usually a new tab, but users can configure browsers to open a new window instead.
+    3.  `_parent`: the parent browsing context of the current one. If no parent, behaves as `_self`.
+    4.  `_top`: the topmost browsing context (the "highest" context that's an ancestor of the current one). If no ancestors, behaves as `_self`.
+
+Above definitions are from the [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#:~:text=separated%20link%20types.-,target,-Where%20to%20display).
```
rajsite, line 252: The extra context is not needed, can just say it aligns with the anchor target attribute exactly and link to the mdn reference. Also to be clear the target attribute is not an enum, it takes arbitrary strings. We should behave exactly like the built-in target attribute.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,36 @@ _Props/Attrs_
 -   `markdown` - for retrieving and modifying the markdown value. If the client modifies the markdown value, it will be parsed into a Node using the
     [prosemirror-markdown parser](https://github.com/ProseMirror/prosemirror-markdown/blob/9049cd1ec20540d70352f8a3e8736fb0d1f9ce1b/src/from_markdown.ts#L199).
     The parsed node will then be rendered in the viewer component as rich text.
+-   `link-target` - is an attribute that accepts valid `target` values for the `anchor` element, applicable to all links within the viewer component.
```
rajsite, line 245: We should call it `anchor` target, that's the term used elsewhere in nimble for anchors.

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: As of fairly recently we have been developing clearer guidelines around when to open links in new tabs based on the WCAG guidelines referenced.

There are two conditions we need to consider:

- From an accessibility perspective: If it is appropriate to open a link in a new tab, how do we show the user the link will open in a new tab. The visual to use has not been defined yet.
- From a security perspective: If the link is "not trusted", how do we show the user the link is not trusted. The visual to use has not been defined yet.

So for the comments feature we have two questions:

1. Is it appropriate for the links to always open in a new tab?
2. Are there concerns about the link being untrusted?

## Is it appropriate for the links to always open in a new tab

The Rich Text Viewer just represents text with links on a page. There is nothing unique about the links in the Rich Text Viewer. Based on the WCAG discussions, the links are on rendered content, there is no implicit behavior of data loss, no explicit switch between security contexts, etc. These are plain links.

So **no**, the comment viewer should not make links always open in a new tab. They should behave like normal links.

The `anchor-target` feature can be useful for cases inside of iframes etc. So I think it is a useful addition generally. **However** it is not a feature that should be used in the comment viewer in SystemLink to always force all links to open in a new window.

## Are there concerns about the link being untrusted

Right now I don't think there are because the comment editor only supports `absolute URLs`. The user can always see very clearly what a link is and make the assessment if it is untrusted. The concern is currently mitigated because we do not support applying URLs to arbitrary text.

**Proposal:** We do not add visuals for opening the url in a new tab or indicating the URL is untrusted as based on the above discussion they are not currently needed. We can optionally add the `anchor-target` attribute as that is a generally useful feature for the rich text viewer but the SystemLink Comment feature will not use that to force all comment links to open in new tabs.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -236,6 +242,36 @@ _Props/Attrs_
...
+they are external (i.e., not from the same domain/origin); for internal links, they will open in the same tab. Here, we will show the same warning mentioned above for the external
+links. Below is proposed high level design for the same,
+
+We will introduce an attribute named `link-configuration` that can accept the following enum values:
+
+1.  `openInSameTab` - This value is the default setting. If chosen, all links within the component will open in the same tab.
+2.  `openInNewTab` - When selected, all links in the component will open in new tab. A warning icon will be placed next to each link, conveying the user that
+    clicking the link will result in a new tab opening.
+3.  `externalLinkBasedOnOrigin` - With this setting, links will be validated based on their origins. If a link's origin differs from the current origin, it is categorized as
+    external link and will open in a new tab. Conversely, links originating from the same domain will be treated as internal link and open in the same tab.
```
rajsite, line 272: I don't think this alternate design is sufficient to be a core features in nimble.

I believe the concern generally is that if a URL is trusted or not-trusted and to warn the user about untrusted URLs.

Just because a URL is a different origin does not mean it is untrusted and should have a warning.
For example, a URL on ni.com would not show a warning for a URL to systemlink.ni.com or for a url to systemlinkcloud.com. Those URLs are different origins, but they are not untrusted URLs.

We may need additional discussion but as far as I can tell the definition of an "untrusted URL" is application specific. Just having a different origin is not a sufficient indication that a URL is not trustworthy.

> If a link's origin differs from the current origin, it is categorized as external link and will open in a new tab

Based on the WCAG guidelines this is not a sufficient reason to always open a URL in a new tab. If the URL is untrusted we need a way for the user to visualize that the URL is untrusted but we do not have that visual yet and that visual is not a requirement yet because we only currently support absolute URLs.

packages/nimble-components/src/rich-text-editor/specs/README.md
vivinkrishna-ni: Updated the spec.

packages/nimble-components/src/rich-text-editor/specs/README.md
suseendran-ni: @rajsite thanks for your input. 

We have explained this to Josh and rechecked the importance of this requirement. But what we understood is, opening the link in a new tab is an important requirement (as most of the users are not software developers).

Attaching the comment from Josh for your reference - https://dev.azure.com/ni/DevCentral/_workitems/edit/2383256

image

Considering the importance of this requirement, can you please guide us on how to proceed further? 

As mentioned in the spec, exposing the `anchor-target` API could be one option. Please let us know if there are any other better ways to achieve this.

I can schedule a meeting with Josh to discuss this if required.

rajsite: Thanks for the follow-up. I added more details to the linked azdo thread to continue the discussion with POs.

change/@ni-nimble-components-27f5d5a2-511b-45fd-991a-bee83da87ed5.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: I think continued review of the API in this spec contingent based on discussions in: https://dev.azure.com/ni/DevCentral/_workitems/edit/2383256

So I'm moving the PR to draft per our current contribution guidelines until that discussion is resolved. Let me know if there are any concerns

jattasNI: @vikisekarNI I see you moved this PR out of draft and back to "ready for review" but I still see several open comment threads. Could you please close those discussions with a short summary of what was decided or what input you still need from reviewers?

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,10 +330,56 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
+2.  Set [openOnClick](https://tiptap.dev/api/marks/link#open-on-click) to `false` for the editor, to restrict the user from opening a link from the editor by clicking. Users can open the link only by
+    `Right-click >> Open link in new tab` from the editor.
+3.  Set [autoLink](https://tiptap.dev/api/marks/link#autolink) to `true`, to add the valid link automatically when typing.
+4.  Set [linkOnPaste](https://tiptap.dev/api/marks/link#link-on-paste) to `false` which will attach the URL to current selected text in the editor, converting it into a hyperlink. If it is `true`,
+    pasting a link to the selection will add the link behind the word which is not supported for the initial pass.
 
 The `nimble-rich-text-viewer` will be responsible for converting the input markdown string to HTML Fragments with the help of
 `prosemirror-markdown` parser, which is then converted to HTML string and rendered into the component to view all rich text content.
 
+_Implementation details for supporting absolute link:_
```
rajsite, line 349: We should have symmetry in validation both ways.
So if tiptap validates links are http and http for the rich-text-editor,
how should the rich-text-viewer behave for links that fail validation?
First reaction is not treat them as links, they just become text. I assume that is how the editor side will behave for unsupported schemes, just treat them as text

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,10 +330,56 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
 -   Italics - `*Italics*`
 -   Numbered list - `1. Numbered list`
 -   Bulleted list - `* Bulleted list`
+-   Absolute URL links - `<Absolute URI link>` (For more details on the markdown syntax for absolute URL links, see [Autolinks in CommonMark](https://spec.commonmark.org/0.30/#autolink))
+
+_Configurations on Tiptap to support only absolute links_:
+
+Install the [link extension](https://tiptap.dev/api/marks/link) mark from Tiptap and initialize the `Links` with the following configurations:
+
+1.  Set regular expression in [validate](https://tiptap.dev/api/marks/link#validate) field to support only `HTTP` and `HTTPS` absolute links in the editor.
```
rajsite, line 339: Still think that any validation being done client side needs to be replicated server side. For example that the server does the same validation (markdown only with certain features is accepted, must validate with the same schema, must have links with certain schemas. etc). How / where is that being tracked? As more features are added etc how would they stay aligned?

change/@ni-nimble-components-27f5d5a2-511b-45fd-991a-bee83da87ed5.json
vikisekarNI: Thanks for the heads up @jattasNI. I will respond with decisions we made in our WCAG call for link support on open comments. I will also re-enable draft mode, since we planned to do prototyping for below comment from Milan. We'll mark it ready for review, once we have the feasibility/outcome of prototype.

image

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: This API is removed for the scope of this PR and will be updated later.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: Since `external-link` is removed for the scope of this PR, visual design icon support is also removed.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: Summarizing the key decisions made and resolving this particular comment for scope of this PR: 
1. All links will open in same tab (or default link behavior) until further notice. Respective updates has been done on the HLD.
2. An ADR will be published for WCAG accessibility guidelines for links, along with other platform level recommendations on which the scope of link support for the `nimble-rich-text-*` components will be revisited.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: This is a valid concern on getting to a right definition for the trusted or untrusted links specific to application. 

This alternative approach is not applicable now since the current working decision is to go with the default behavior of same tab. Contents has been removed from this HLD

packages/nimble-components/src/rich-text-editor/specs/README.md
suseendran-ni: @rajsite, yes, currently the `rich-text-viewer` does not have any validation to consider `http` / `https`. 

But as you mentioned, we will update the `rich-text-viewer` to render the links as plain text, if the links do not have a `http` / `https` scheme. And that is how the `rich-text-editor` will also behave.

packages/nimble-components/src/rich-text-editor/specs/README.md
suseendran-ni: @rajsite, currently in the backend, we do not plan to add any validation on the markdown content. Because even if try to create a comment (that includes links that do not belong to `http` or `https` schema) directly via API, when it gets rendered in the UI, it will be rendered as plain text (because of the rules we configured during the markdown parser).

Currently, SLE email notification does not support HTML content. In the future, when it supports HTML content, we can implement the same parsing logic (we are doing in the client to convert markdown to HTML) to generate the HTML content for mailing.

Please let me know your thoughts.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: Prototyped the feasibility of validation for HTTP/HTTPS scheme in `rich-text-viewer`'s markdown parsing logic (Prose Mirror/markdown -it) and updated the spec document to mention about the possibility of symmetrical validation in both viewer and editor.

vikisekarNI (approved):

jattasNI:
Friendly reminder that you should click "re-request review" when you've addressed reviewers feedback and want them to weigh in again. This is helpful because GitHub offers ways for reviewers to filter PRs by the ones that are awaiting our input, but PRs won't show up in that filter if you don't re-request.

I already did this for the reviewers that needed it on this PR, just a tip for the future!

image

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -48,7 +53,8 @@ The `nimble-rich-text-viewer` provides support for converting the input markdown
 
 -   Allowing the user to tag or mention by entering `@` in the editor and selecting the user name from the drop-down list.
 -   Support for adding images to the editor either by uploading or by pasting it.
--   Support for adding hyperlinks to the existing words in the editor.
+-   Support for adding hyperlinks to the existing text in the editor. This allows users to add links to existing text in the editor. When the
+    link button in the formatting options is clicked, a dialog opens, providing a space to enter the hyperlink for the selected text.
```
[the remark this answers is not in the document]
vikisekarNI: We are yet to research and explore on the hyperlink dialog implementation strategy. Currently it's added under out of scope section. This should have been captured earlier by us, under out of scope. 

Since this point is added in out of scope now, the diff shows only this point without showing the actual heading of this section.

Hope it should be fine to resolve this comment. Let us know otherwise.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -34,6 +34,11 @@ Both components provide support for the following basic text formatting options:
 2. Italics
 3. Numbered List
 4. Bulleted List
+5. Absolute URL links - Supports only [absolute URI](https://spec.commonmark.org/0.30/#absolute-uri) with a valid [scheme](https://spec.commonmark.org/0.30/#scheme).
```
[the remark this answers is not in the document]
vikisekarNI: Updated as requested

packages/nimble-components/src/rich-text-editor/specs/README.md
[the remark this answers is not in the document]
jattasNI: Sorry about that! Github had collapsed the header for that section so I thought it was part of the Features plan.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,20 +335,69 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
+                href: node.attrs.href as Attr,
+                rel: node.attrs.rel as Attr
+            }
+        ];
+    }
+}
+```
+
+1.  We will set the `rel` attribute value to `noopener noreferrer` to enhance security and ensure responsible linking practices.
+2.  In the `toDOM` function, we have incorporated the `anchorTag` to render all links within the viewer component as `nimble-anchor` elements.
```
jattasNI, line 383: One thing to double check (if you haven't already): does the nimble-anchor render as expected if the user applies bold or italic styling to it? I noticed the anchor matrix story has a test case for italic but I didn't see one for bold.

jattasNI (approved):

jattasNI:
I will pester the remaining Nimble reviewers tomorrow to submit their feedback.

packages/nimble-components/src/rich-text-editor/specs/README.md
vikisekarNI: On a quick check, looks like yes. nimble-anchor is able to support this rendering both in viewer and editor.

**Viewer**
MicrosoftTeams-image

**Editor**
MicrosoftTeams-image (1)

vikisekarNI: Partial formatting also looks to work fine. `nimble-anchor` is awesome! 😃 

Will resolve this thread after getting your acknowledgement or further pre-cautionary inputs to be considered.

MicrosoftTeams-image (2)

jattasNI: Great, thanks for checking. It'd be good to add these as test cases when you get to implementation.

atmgrifter00:
I will decline to review.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,20 +335,69 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
 -   Bulleted list - `* Bulleted list`
+-   Absolute URL links - `<Absolute URI link>` (For more details on the markdown syntax for absolute URL links, see [Autolinks in CommonMark](https://spec.commonmark.org/0.30/#autolink))
+
+_Configurations on Tiptap to support only absolute links_:
+
+Install the [link extension](https://tiptap.dev/api/marks/link) mark from Tiptap and initialize the `Links` with the following configurations:
+
+1.  Set regular expression in [validate](https://tiptap.dev/api/marks/link#validate) field to support only `HTTP` and `HTTPS` absolute links in the editor.
+2.  Set [openOnClick](https://tiptap.dev/api/marks/link#open-on-click) to `false` for the editor, to restrict the user from opening a link from the editor by clicking. Users can open the link only by
+    `Right-click >> Open link in new tab` from the editor.
```
m-akinc, line 346: Does this mean that link markdown in the editor is somehow rendered as an operable link? I would have expected it to just look like plain markdown text, e.g. "<http://ni.com>". Only in the viewer would I expect it to look and behave like a link.

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,20 +335,69 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
+            {
+                href: node.attrs.href as Attr,
+                rel: node.attrs.rel as Attr
+            }
+        ];
+    }
+}
+```
+
+1.  We will set the `rel` attribute value to `noopener noreferrer` to enhance security and ensure responsible linking practices.
```
m-akinc, line 382: Does setting `noopener` do anything if we're not also setting `target="_blank"`? We are opening links in the same tab by default, right?

m-akinc (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: As long as it's crystal clear on the backend owner's part that no validation of the data going into the backend is happening (i.e. only client side validation which is a known bad practice) and it passes muster for their security audits. I.e. those decisions have sign-off from the right folks understanding and documenting those concerns, ie. @rbell517

I think it's weird that SystemLink's security requirements are being encoded in the nimble rich text editor component this way and that it's the only line of validation that happens. 

Ultimately I won't block this spec on it. The behavior being implemented for the rich-text components as described here seems fine (only autolinking http and https urls).

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: The rich text editor is a wysiwyg editor. The user does not type markdown into it so they would not see the commonmark markdown representation, they would see it render as a link.

It will render as a link during edit (in WYSIWYG fashion) but not be operable as to avoid accidental navigation (except by new window interactions). I agree that it's not a perfect tradeoff and they are not clear how to operate. But I think it's okay enough for now and the UX could be iterated on as needed.

FWIW powerpoint as a WYSIWYG editor also prevents link clicks from opening by default and they include in the alt text instructions for how to follow the link:

image

That's just one example though, finding the right UX may take additional investigation.

packages/nimble-components/src/rich-text-editor/specs/README.md
rajsite: It doesn't hurt or help for the context of this spec and should probably be used for future revisions. I don't feel strongly that it should be avoided.

rajsite (approved):

packages/nimble-components/src/rich-text-editor/specs/README.md
```diff
@@ -320,20 +335,69 @@ markdown based on [CommonMark](http://commonmark.org/) flavor:
...
+    pasting a link to the selection will add the link behind the word which is not supported for the initial pass.
 
 The `nimble-rich-text-viewer` will be responsible for converting the input markdown string to HTML Fragments with the help of
 `prosemirror-markdown` parser, which is then converted to HTML string and rendered into the component to view all rich text content.
 
+_Implementation details for supporting absolute link:_
+
+For the `nimble-rich-text-viewer` component, we will set up the `link` mark in the Prosemirror schema as below, allowing links in the component to open with default behavior (same tab).
+
+Only links with schema HTTP and HTTPS will be treated as links within `nimble-rich-text-editor` and `nimble-rich-text-viewer` and rest of the links with any other schemas will be rendered as plain text.
```
rajsite, line 358: I assume for the nimble-rich-text-editor this validation applies to both user entered links and to the setMarkdown() method right? The setMarkdown codepath seems fairly unique so I think we should state it explicitly

packages/nimble-components/src/rich-text-editor/specs/README.md
rbell517: Yes this security question has come up a few times, and essentially where I have landed is we should always consider the contents of a comment as untrusted, in both the frontend and any usage of it in the backend. Just as the backend shouldn't assume its getting contents from a specific frontend, the frontend shouldn't assume its getting content from a specific backend; these should be trust boundaries for both. 

That said, even if the backend treats the comment contents as untrusted, that does not mean it will treat it the same as untrusted content in the frontend. Its not rendering the content, but it is inserting it into a database, so the security context changes to validating for SQL injections instead of XSS, for example. This will change again for other usages like email notifications, etc. Hopefully the posture is clear enough.

rajsite: > This will change again for other usages like email notifications

So if I understand as of now that means we would accept arbitrary strings even if they don't conform to markdown encoded in our supported subset of CommonMark features and custom extensions (i.e. mentions syntax and however that will look in the future) in addition to XSS concerns being detected in properly structured formats.

That seems like something that would be challenging to clean-up without data loss in the future when we could make sure to start enforcing that now. I'd really hope we don't introduce that technical debt into the system if we don't need to but yea ultimately as stated that's not a nimble concern and sounds like the concerns are understood.

I guess the remaining question is if that is tracked anywhere @rbell517 @suseendran-ni  (security concerns for string insertion and technical debt of markdown validation so we know what's actually being dumped in our databases), but that is orthogonal to acceptance of this PR (but wouldn't mind seeing how that is captured).

suseendran-ni: Created a tech debt to track this conversation
https://dev.azure.com/ni/DevCentral/_workitems/edit/2511073

@vivinkrishna-ni @vikisekarNI @rbell517 @rajsite

Issue #1302, opened 2023-06-14
Combobox does not update value when programmatically updating with duplicate entries in Angular


rajsite:
# 🐛 Bug Report

Create a combobox that uses ngModel and has multiple options with the same label but different ngValues associated with them.
If you try and assign one of the duplicate ngValues to the combobox, the combobox does not recognize them as an accepted value.

## 💻 Repro or Code Sample

See [example](https://stackblitz.com/edit/nimble-combobox-duplicate-entries-label?file=src%2Fmain.ts).

## 🤔 Expected Behavior

When you press the button in the example the combobox should show the label of the corresponding model value.

## 😯 Current Behavior

The combobox lookup map is loosing the duplicate options model value.

## 💁 Possible Solution

The combobox should be aware of all the model values so it can do the lookup correctly and show the label (even if there are duplicate labels).

## 🔦 Context

See [Skyline PR 504385](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/504385)

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

Issue #1273, opened 2023-05-30
Filterable `NimbleSelect`.


atmgrifter00:
## 😯 Problem to Solve

We are seeing a lot of requests from clients to have a filterable dropdown component that supports model values associated with each option (in Angular), with no need for arbitrary value support. Ideally, this filtering would allow for _any_ string matching within the text of each displayed option.

## 💁 Proposed Solution

Add new typing-based filtering to the `NimbleSelect`. Behaviorally, it could operate in a very similar fashion to [this dropdown component](https://select2.org/dropdown). The key addition here is a new input area in the dropdown itself allowing a user to type. This likely should be a _conditional_ ability of the `Select` as well, allowing a user to not always have a dedicated area for a user to type.

[Prototype branch](https://github.com/ni/nimble/tree/filterable-select) (and [Storybook](https://60e89457a987cf003efc0a5b-ldjwdkyulp.chromatic.com/?path=/story/components-select--underline-select))

[Initial visual designs (WIP)](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1529-41777&mode=design) -- see references to "filter" in WIP Experiments section

jattasNI:
One use case where this came up was the SLE user picker. It would be good to be able to type a user's first or last name to narrow the list of visible users. [Discussion on Teams here](https://teams.microsoft.com/l/message/19:a045e5de02a8458484bd3705ac3df6ed@thread.skype/1679332622128?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=41626d4a-3f1f-49e2-abdc-f590be4a329d&parentMessageId=1679332622128&teamName=ASW%20SystemLink&channelName=LIMS%20UI-UX&createdTime=1679332622128)

atmgrifter00:
Some observations made while demoing the prototype:
- We should place the filter input at the bottom of the popup if the popup is shown above the component. Otherwise the input would bounce around as the user types.
- We should perhaps stop updating the display text of the select as the user navigates the dropdown via the arrow keys.
    - A corollary to this would be that we need to visually distinguish between the current selected item and items that are currently focused in the drop-down via the key navigation.

jattasNI:
## Interaction design questions

1. What gestures can the user make to focus the filter input text box? Click in it? Tab to it? It gets focus when the select is opened?

We plan to have the dev team make a proposal and run it by IxD.

jattasNI:
[...]

PR #1395, opened 2023-07-27
Change Combobox CVA to handle duplicate values

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1302

While it's understood that the `Combobox` will ultimately change so that clients will not be able to associate arbitrary models with a value, until we provide a [filterable Select](https://github.com/ni/nimble/issues/1273) component, we can temporarily provide better behavior for the `Combobox` when it comes to duplicate values.

## 👩‍💻 Implementation

Essentially we will now allow the CVA to track multiple model values per identical display text. Further, we can even support multiple options sharing the same model value instance, by using the element reference of each option as a unique identifier.

Currently, if a user selects a duplicate value in the Combobox, it will return the first model value found amongst the duplicate options.

## 🧪 Testing

Added unit tests.

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
```diff
@@ -43,20 +43,20 @@ export class NimbleComboboxListOptionDirective implements AfterViewInit, OnDestr
...
         }
     }
 
     private updateComboboxValue(value: unknown): void {
-        this.combobox!.removeOption(this._currentTextContent);
+        this.combobox!.removeOption(this._currentTextContent, this.elementRef);
         this.changeDetector.detectChanges();
         this._currentTextContent = this.elementRef.nativeElement.text;
-        this.combobox!.addOption(this._currentTextContent, value);
+        this.combobox!.addOption(this._currentTextContent, value, this.elementRef);
```
rajsite, line 60: The remove + add on value update will change the order in the optionmap, it won't align with the dom order anymore. Is that observable?
Should have an updateOptionValue or something that preserves the map order or does it not matter?

I'm now also a bit worried about reading from the nativeElement.text again here... If the text was modified it's only observed if ngValue also happens to be updated? Is there a temporal issue? It seems like the text must be assigned before ngValue for this to work since we only read the text value on ngValue update.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -124,9 +133,11 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
     }
 
     private getValueStringFromValue(value: unknown): string | undefined {
-        for (const [optionKey, optionValue] of this._optionMap.entries()) {
-            if (this._compareWith(optionValue, value)) {
-                return optionKey;
+        for (const [displayValue, optionValues] of this._optionMap.entries()) {
+            for (const modelValue of optionValues.values()) {
+                if (this._compareWith(modelValue, value)) {
+                    return displayValue;
```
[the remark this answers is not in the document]
atmgrifter00: Changed so that we always get the first option found.

angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
atmgrifter00: I'm no longer doing add + remove, and instead just doing a direct update of the option when its model value changes. Additionally, there is a now a mutation observer that will handle if the text is changed in the DOM for keeping the internal state up-to-date for the options. These operations _do not_ affect the current display text or model value associated with the combobox.

angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
```diff
@@ -26,7 +26,8 @@ export class NimbleComboboxListOptionDirective implements AfterViewInit, OnDestr
     public set ngValue(value: unknown) {
         if (this.combobox) {
             this._modelValue = value;
-            this.updateComboboxValue(value);
+            this.changeDetector.detectChanges();
```
rajsite, line 29: Guessing this will change a bit after https://github.com/ni/nimble/pull/1470
So can take a look again after that pone merges

atmgrifter00: Updated.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -64,14 +66,16 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
     @HostListener('blur')
     private onTouched: () => void;
 
-    public constructor(
-        private readonly _renderer: Renderer2,
-        private readonly _elementRef: ElementRef
-    ) {}
+    private readonly observer: MutationObserver;
+
+    public constructor(private readonly _renderer: Renderer2, private readonly _elementRef: ElementRef) {
```
rajsite, line 71: Can you type the `ElementRef<Combobox>` think we can then remove some `as HTMLElement` and `as Combobox` casting

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -158,4 +203,14 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
     private setProperty(key: string, value: unknown): void {
         this._renderer.setProperty(this._elementRef.nativeElement, key, value);
     }
+
+    private readonly mutationCallback = (mutations: MutationRecord[]): void => {
+        const optionTextUpdate = mutations.find(mutation => {
+            return ((mutation.type === 'characterData' && mutation.target.parentElement instanceof ListOption)
```
rajsite, line 209: I'm really sorry but I can't remember. What cases does the mutation observer handle that the add, remove, and queue update in combobox list option don't handle? Shouldn't that handle all option text content updates and list item children changes?

Maybe we just need the map updates but not the mutationobserver stuff?

atmgrifter00: Yeah this was just being done in case someone messed with the `textContent` of the native ListOption direction. We agreed this is overkill so I removed this code. This required a slight change to the `addOption` logic to only add the option to the set being tracked if it didn't already exist.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -119,20 +120,45 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
...
+        }
+        this._optionToModelValueMap.delete(option);
+    }
+
+    public updateOption(modelValue: unknown, option: ListOption): void {
+        this._optionToModelValueMap.set(option, modelValue);
+        this.updateDisplayValue();
     }
 
     public queueOptionUpdate(listOption: ListOption, modelValue: unknown): void {
```
rajsite, line 160: Can we align the signature with addOption and give it a more descriptive name, i.e.:

```ts
updateOptionModelValue(displayValue: string, modelValue: unknown, option: ListOption)
```

Concern is some of these methods access the `option.text` directly while others expect it to be passed in via `displayValue`, just trying to get it to be consistent.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -119,20 +120,45 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
...
+                    options.splice(removeIndex, 1);
+                }
+            } else {
+                this._displayTextToOptionsMap.delete(displayValue);
+            }
+        }
+        this._optionToModelValueMap.delete(option);
+    }
+
+    public updateOption(modelValue: unknown, option: ListOption): void {
```
rajsite, line 155: Is this method used anymore? I can't see what is calling it.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
rajsite, line 160: This method should be marked internal

rajsite (approved):

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
atmgrifter00: Yeah, I just got rid of that extra argument everywhere...it was unneeded.

angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
```diff
@@ -42,17 +42,17 @@ export class NimbleComboboxListOptionDirective implements AfterViewInit, OnDestr
     public ngAfterViewInit(): void {
         if (this.combobox) {
             this._currentTextContent = this.elementRef.nativeElement.text;
```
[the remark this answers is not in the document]
atmgrifter00: Crud...no.

Issue #1387, opened 2023-07-25
Resizing table columns to the left sometimes adds horizontal scrollbar


mollykreis:
# 🐛 Bug Report

I was playout around with the table columns resize behavior, and I noticed that resizing columns to the left sometimes adds a horizontal scrollbar with a very small scroll amount. Looking in dev tools, it looks like the minimum scroll width and the actual size of the table only differs by 1px. Therefore, it seems like there is some rounding error at play.

![table-scroll](https://github.com/ni/nimble/assets/20542556/02d89073-e046-459f-9d3f-d580b23caf07)

## 🤔 Expected Behavior

Resizing columns to the left should never create a horizontal scrollbar.

## 😯 Current Behavior

Resizing columns to the left can sometimes create a horizontal scrollbar.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #1396, opened 2023-07-27
Fix table horizontal scroll issue

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1387

## 👩‍💻 Implementation

Just remove unnecessary calls to `Math.round` on the delta value in the cascade implementations. The current implementation works well with individual columns being given an adjusted pixel size on a sub-pixel boundary.

## 🧪 Testing

Added a unit test.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
+++ b/packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
@@ -869,4 +869,12 @@ describe('Table Interactive Column Sizing', () => {
         expect(actualFractionalWidths).toEqual(expectedFractionalWidths);
         expect(actualPixelWidths).toEqual(expectedPixelWidths);
     });
+
+    it('sizing to left when columns are all on sub-pixel boundary and just above minimum size, does not cause total column width to change', async () => {
+        await pageObject.sizeTableToGivenRowWidth(202, element); // columns now on half-pixel boundary
+        pageObject.dragSizeColumnByLeftDivider(3, [-1]);
+        await waitForUpdatesAsync();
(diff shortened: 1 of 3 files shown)


packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -869,4 +869,12 @@ describe('Table Interactive Column Sizing', () => {
         expect(actualFractionalWidths).toEqual(expectedFractionalWidths);
         expect(actualPixelWidths).toEqual(expectedPixelWidths);
     });
+
+    it('sizing to left when columns are all on sub-pixel boundary and just above minimum size, does not cause total column width to change', async () => {
+        await pageObject.sizeTableToGivenRowWidth(202, element); // columns now on half-pixel boundary
```
rajsite, line 874: Is sizing the table to a 202px row width a guarantee that columns are now on a half-pixel boundary?
Isn't that a function of a bunch of things related to the styling, etc?

How do you feel about coming up with an assertion that can validation the assumption that the table is actually at a half-pixel boundary before the test runs. Alternatively is there a test set-up that would give a clear guarantee the setup is what we expect?

atmgrifter00: It's a guarantee with the current test setup, where there are four columns that should each be sized to 50.5 pixels since they are equally sized. I can provide an assertion to this effect prior to sizing a column.

jattasNI (approved):

rajsite (approved):

Issue #1383, opened 2023-07-24
nimble-tabs allows switching to a hidden tab using keyboard


mollykreis:
# 🐛 Bug Report

The `nimble-tabs` component allows you to switch to a hidden tab using the left/right arrows. This should not be allowed.

## 💻 Repro or Code Sample

1. Go to the storybook page for the Tabs Matrix: https://nimble.ni.dev/storybook/?path=/story/tests-tabs--tabs-theme-matrix
2. Focus the first tab element on the page
3. Notice that you can use the Right Arrow button to move from Tab One to Tab Two
4. Notice that if you press the Right Arrow button again, the content of the tab panel will be that of the hidden Tab Three that is part of the story

## 🤔 Expected Behavior

A hidden tab should behave as though it is not in the DOM. A user should not be able to navigate to that tab's content using the keyboard (or the mouse).

## 😯 Current Behavior

You can navigate to the hidden tab using the arrow keys.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

m-akinc:
Check anchor tabs as well

m-akinc:
PR to FAST: https://github.com/microsoft/fast/pull/6789

PR #1397, opened 2023-07-27
Fix keyboard navigation with hidden anchor tabs

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1383
The tabs control needs to be fixed in FAST, but we have similar flawed logic being used for the anchor tabs.

## 👩‍💻 Implementation

The code already handles disabled tabs in the keyboard navigation logic. We just need to expand that to handle hidden tabs the same way. I updated `isFocusableElement()` to also look for the presence of `hidden`.

Also, to make Home/End work properly I had to replace a call to `isDisabledElement()` with `isFocusableElement()` (which used to be equivalent, but now are not).

## 🧪 Testing

Wrote two new test cases and tested manually in Storybook.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/anchor-tabs/index.ts
+++ b/packages/nimble-components/src/anchor-tabs/index.ts
@@ -104,8 +104,12 @@ export class AnchorTabs extends FoundationElement {
         return el.getAttribute('aria-disabled') === 'true';
     };
 
+    private readonly isHiddenElement = (el: Element): el is HTMLElement => {
+        return el.getAttribute('hidden') !== null;
+    };
+
     private readonly isFocusableElement = (el: Element): el is HTMLElement => {
-        return !this.isDisabledElement(el);
+        return !this.isDisabledElement(el) && !this.isHiddenElement(el);
     };
 
     private readonly setTabs = (): void => {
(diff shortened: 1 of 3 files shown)


packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
```diff
@@ -203,6 +204,20 @@ describe('AnchorTabs', () => {
             initialFocusIndex: 0,
             keyName: keyEnd,
             expectedFinalFocusIndex: 1
+        },
+        {
+            name: 'should skip hidden tab when arrowing right',
```
mollykreis, line 209: The test cases for hidden tabs don't seem quite as complete as the test cases for disabled tabs. Is there a reason not to have a test for arrowing left or for pressing the Home key?

packages/nimble-components/src/anchor-tabs/index.ts
```diff
@@ -104,8 +104,12 @@ export class AnchorTabs extends FoundationElement {
         return el.getAttribute('aria-disabled') === 'true';
     };
 
+    private readonly isHiddenElement = (el: Element): el is HTMLElement => {
+        return el.getAttribute('hidden') !== null;
```
mollykreis, line 108: Is there a reason not to use `hasAttribute('hidden')` here instead?

mollykreis (approved):

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
m-akinc: I know they exercise the same relevant code paths, so I didn't write them. But then the same logic would apply to disabled tabs... so I guess I should add more test cases for consistency.

jattasNI (approved):
I think you mentioned you'd filed an issue/created a PR in FAST. If so can you make sure it's linked from the Nimble issue?

rajsite (approved):

PR #1405, opened 2023-07-31
Facilitate getting cells by row/column ID

### LINKED ISSUE #1308: Enable finding a table cell by row ID & column ID
## 😯 Problem to Solve

There has been a request to write tests that verify that a cell has the correct content given a row ID and a column ID. Right now, there isn't a way to accomplish this -- either using the existing page object or by querying the DOM directly.

It would be nice to have this capability from our existing page object and from a playwright page object (#1306).

We'll need to consider what to do if the requested item isn't on screen. Scroll to it? Throw an error?

See discussion here: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/503741#1686845187

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1308

## 👩‍💻 Implementation

- Added `columnId` property/attribute to `TableCell`, including logic to update the value when the `column` property changes, or the column's `columnId` changes.
- Added two tests for `TableCell` that validate the `columnId` is updated properly
- Added public `getRenderedCellViewById()` function to the table page object, which takes a `recordId` and a `columnId`.
- Added a test file for the table page object with one function testing `getRenderedCellViewById()`

## 🧪 Testing

Added unit tests for the new `TableCell.columnId` logic and the new `getRenderedCellViewById` function.

## ✅ Checklist


packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -61,6 +77,37 @@ export class TableCell<
...
     }
+
+    /**
+     * @internal
+     *
+     * The event handler that is called when a notifier detects a change. Notifiers are added
+     * to the column, so `source` is expected to be an instance of `TableColumn`, and `args`
+     * is the string name of the property that changed on that column.
+     */
+    public handleChange(source: unknown, args: unknown): void {
```
msmithNI, line 88: Adding observers to every cell just to monitor the columnId feels kind of heavyweight to me.

My vote would be to set column-id for cells in the `TableRow`'s template, similar to how the [table sets record-id on rows in it's template](https://github.com/ni/nimble/blob/bde26d30926b908fcac8f473948b3b360cd6b445/packages/nimble-components/src/table/template.ts#L125).

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -604,6 +618,19 @@ export class TablePageObject<T extends TableRecord> {
         return rows.item(rowIndex);
     }
 
+    private getRowById(recordId: string): TableRow {
+        const row: TableRow | null = this.tableElement.shadowRoot!.querySelector(
+            `nimble-table-row[record-id="${recordId}"]`
```
msmithNI, line 623: Since we don't have many restrictions on record ID / column ID formats, they can contain double-quotation marks which would result in an invalid selector here (and in `getCellById`).
We probably need to add escaping - it looks like [CSS.escape()](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static) should work.

packages/nimble-components/src/table/components/cell/index.ts
m-akinc: Doing it that way doesn't seem to trigger an update when the column's `columnId` changes. Would you expect it to?

msmithNI: Tried it locally and I see it update:
CellColumnIdChange
That's with `@attr({ attribute: 'column-id' }) public columnId?: string;` in `TableCell`, and `column-id="${x => x.column.columnId}"` on the cell in the `row\template.ts`.

m-akinc: Oops, I had a failing unit test, but it was because there was no row involved. Was just creating a table cell directly and assigning it a column.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -616,6 +643,19 @@ export class TablePageObject<T extends TableRecord> {
         return cells.item(columnIndex);
     }
 
+    private getCellById(recordId: string, columnId: string): TableCell {
+        const row = this.getRowById(recordId);
+        const cell: TableCell | null = row.shadowRoot!.querySelector(
+            `nimble-table-cell[column-id="${CSS.escape(columnId)}"]`
```
rajsite, line 649: ah, excellent catch, yea the ids can be arbitrary strings. And using CSS.escape is the right escape for this case since it's going in a CSS selector. 👍

rajsite (changes requested):

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: Credit goes to Malcolm for that one!

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -87,6 +87,51 @@ describe('TableRow', () => {
...
+    it('updates cell.columnId when column changes', async () => {
+        await connect();
+
+        const textColumn1 = new TableColumnText();
+        textColumn1.columnId = 'foo';
+        textColumn1.fieldName = 'stringData';
+        const textColumn2 = new TableColumnText();
+        textColumn2.columnId = 'bar';
+        textColumn2.fieldName = 'moreStringData';
+        element.columns = [textColumn1];
```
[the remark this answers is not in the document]
m-akinc: I believe I've addressed all your concerns.

packages/nimble-components/src/table/components/cell/tests/table-cell.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-classes-per-file */
```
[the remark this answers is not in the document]
m-akinc: I initially had tests in here that I moved to the TableRow tests. I overlooked removing the old supporting code. I'll try to give my PRs another pass before leaving draft in the future.

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -24,208 +28,254 @@ async function setup(): Promise<Fixture<TableRow<SimpleTableRecord>>> {
...
-        expect(listener.spy).toHaveBeenCalledTimes(1);
-        const expectedDetails: TableRowSelectionToggleEventDetail = {
-            newState: true,
-            oldState: false
-        };
-        const event = listener.spy.calls.first().args[0] as CustomEvent;
-        expect(event.detail).toEqual(expectedDetails);
-    });
+// prettier-ignore
+async function setupTable(): Promise<Fixture<Table<SimpleTableRecord>>> {
```
rajsite, line 32: nit: can you move this down to the `describe('in table',` block it's used in? Makes it easier to see with the tests using it

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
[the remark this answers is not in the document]
rajsite: Looking great, thank you!

rajsite (approved):

PR #1407, opened 2023-08-01
Show url of focused anchor tab at bottom of page

### LINKED ISSUE #1401: Anchor tabs do not show URL in toolbar and accessibility issues
# 🐛 Bug Report

When tabbing to an anchor tab in Chrome I do not see the URL status bar pop-up at the bottom of the screen and when switching tabs using the arrow keys I also do not see the URL status bar showing the updated active tab url.

When tabbing / arrowing to nimble-anchor, nimble-anchor-tree-item, nimble-menu-item I do see the URL status bar show and change:
image

I can also see that when switching between anchor tabs with the arrow keys that the aria-selected property does not change. Also the tab-index of the child anchor is always `-1` (which I'm guessing is the reason the url bar is not updating as the anchor itself is not active):

image

We should make sure the aria and link activation are happening correctly.

## 💻 Repro or Code Sample

Visit the example angular app and tab to the anchor tab items, notice the URL bar indicating where the link navigates does not appear.

## 🤔 Expected Behavior

See above.

## 😯 Current Behavior

See above.

## 💁 Possible Solution

See above.

## 🔦 Context

Testing anchor tabs changes from a PR.

## 🌍 Your Environment

Chrome Version 115.0.5790.110 (Official Build) (64-bit)

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1401

## 👩‍💻 Implementation

- Set `delegatesFocus=true` for anchor tab so that inner anchor gets focus
- Removed `tabindex=-1` from native anchor element, so that it can receive keyboard focus
- No longer sharing styling between the anchor tab and the standard tab. Moved the shared styling back under the `tab` directory, then added a modified copy of the contents to the anchor tab stylesheet. We have to change the `::before` and `::after` pseudoelements to be on the `a` rather than the `:host`, because it is now the `a`'s focus state that they depend on.

## 🧪 Testing

Tested manually in Storybook

## ✅ Checklist


packages/nimble-components/src/anchor-tab/styles.ts
```diff
@@ -1,14 +1,124 @@
...
+        width: 0px;
+        height: 0px;
+        border-bottom: ${borderHoverColor}
+            var(--ni-private-focus-indicator-width) solid;
+        transition: width ${smallDelay} ease-in;
+    }
+
+    @media (prefers-reduced-motion) {
+        a::before {
+            transition-duration: 0.01s;
```
atmgrifter00, line 75: Any reason to not just set this to 0 seconds? While I do see some other cases where we use the '.01s' value, most `trasition-duration` values are set to '0s'. I'm not sure if there's even a noticeable difference between the two.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
```diff
@@ -255,6 +255,11 @@ describe('AnchorTabs', () => {
             const specType = getSpecTypeByNamedList(test, focused, disabled);
             // eslint-disable-next-line @typescript-eslint/no-loop-func
             specType(test.name, async () => {
+                // Can't focus a link without an href, so set them all here.
+                tab(0).href = 'foo';
```
atmgrifter00, line 259: Any reason to not set the `href` attribute in the fixture template? I don't see any tests that rely on that attribute not being set. Should there be?

packages/nimble-components/src/anchor-tab/styles.ts
atmgrifter00, line 1: Is the main motivation to not share the styles the requirement that the `::before` and `::after` pseudo-selectors now needed to be applied to an anchor element (instead of the `host`)? If so, would there still be value in sharing the styles, but just moving the `::before` and `::after` CSS out of the shared style, and into a styles file specific to each component?

packages/nimble-components/src/anchor-tab/styles.ts
m-akinc: It's copied from the tab styling. I'll set them both to 0.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
m-akinc: There are tests that will click on a tab (e.g. to change the active tab), and if an href is set it will attempt a navigation, which aborts the test run. I don't really need to (and can't practically) test navigating, so I generally want no hrefs set.

packages/nimble-components/src/anchor-tab/styles.ts
m-akinc: The `::before` and `::after` stuff accounts for about 70% of the styling. I'm not sure there's much benefit in separating the rest into a shared file.

atmgrifter00 (approved):

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
rajsite: Let's go ahead and make a separate fixture relevant for these tests then

rajsite (approved):
approved assuming straightforward test clean-up

PR #1409, opened 2023-08-01
Set build locale to en_US.UTF-8

m-akinc:
# Pull Request

## 🤨 Rationale

In another PR for the date-text column, I included some tests that are sensitive to the machine's locale and expect it to be some form of 'en'. While it seems all GitHub's hosted runners are in the US and would therefore have a compatible default locale, it is safer to explicitly set the locale.

## 👩‍💻 Implementation

Setting the env variables LC_ALL, LANG, and LANGUAGE is supposed to be sufficient.

## 🧪 Testing

Build succeeds.

## ✅ Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,9 @@ on:
 env:
   GITHUB_SERVICE_USER: "rajsite"
   GITHUB_SERVICE_EMAIL: "rajsite@users.noreply.github.com"
+  LC_ALL: "en_US.UTF-8"
+  LANG: "en_US.UTF-8"
+  LANGUAGE: "en_US.UTF-8"
 
 jobs:
   build:
--- a/packages/nimble-components/karma.conf.js
+++ b/packages/nimble-components/karma.conf.js
@@ -29,7 +29,7 @@ const commonChromeFlags = [
     '--disable-infobars',
     '--disable-translate',
     '--force-prefers-reduced-motion',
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -9,6 +9,9 @@ on:
 env:
```
rajsite, line 9: The goal isn't just en but en-US, there are lots of formats of en locales..

The other PR was merged without that discussion, can you make that update in this PR for the karma config:
https://github.com/ni/nimble/pull/1381/files#diff-bc258864f2b77d1453fe6b926921f8db150e8e3ed98a5447dfb7a178dd0d2b7dR32

packages/nimble-components/karma.conf.js
```diff
@@ -29,7 +29,7 @@ const commonChromeFlags = [
     '--disable-infobars',
     '--disable-translate',
     '--force-prefers-reduced-motion',
-    '--lang=en'
+    '--lang=en-US'
```
rajsite, line 32: Just to sanity check, does this run correctly locally on a windows build? (not confirmed via CI which is linux)

rajsite (approved):

packages/nimble-components/karma.conf.js
m-akinc: Yes, tests still pass when run locally on Windows.

PR #1257, opened 2023-05-18
Automatic theme / Runtime Configuration HLD

rajsite:
# Pull Request

## 🤨 Rationale

HLD for proposed changes to make the components reactive to the system theme by default. Also adds an element to control the page default theme and refactors token layout.

## 👩‍💻 Implementation

See HLD.

## 🧪 Testing

N/A

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1410, opened 2023-08-01
Add locale support to theme provider and use in date-text column

### LINKED ISSUE #1293: Provide locale setting on nimble-theme-provider
## 😯 Problem to Solve

The `nimble-theme-provider` element should have a way to specify a locale that can be used by various components for sorting and formatting. We should also consider having the default value of `locale` be determined by the browser.

Some places where this could be used:
- locale-aware string sorting on columns in the `nimble-table`
- number formatting in the `nimble-number-field`
- number/date formatting on columns in the `nimble-table`
- date formatting in a future `nimble-date-picker`

This has been discussed in a few different places:
- [Table column sort HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-column-sort-hld.md#configurable-locale-for-sorting)
- [Number column HLD](https://github.com/ni/nimble/pull/1268)

## 💁 Proposed Solution

Add a new property/attribute to the `nimble-theme-provider` that allows components to access the current locale of an application. This attribute can have a default value based on the application, but it should be configurable by an application using nimble.

More investigation is still needed to solidify a recommendation on string sorting, such as for string column sorting in the table. We should consider how the [`Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) compares to the current implementation of [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1293
Part of #1014 
Adding design token ("lang") for locale and honoring that token's value when formatting dates in the date-text column.

## 👩‍💻 Implementation

- introduced "lang" design token with dynamic default that uses the page's `lang` property, if valid, otherwise uses "en-US"
- new `DocumentElementLang` singleton watches page element's `lang` property
- added `lang` property to theme provider which sets design token value
- using design token value when creating formatter used by date-text column
- added new validation flag to theme provider that is set when the locale is malformed
- setting 'en-US' locale for all date-text unit tests (which they were already assuming)
- added unit tests

## 🧪 Testing

Unit tests pass (even when system locale is something other than 'en').

## ✅ Checklist


packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,6 +38,11 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
+    @attr({
```
mollykreis, line 41: Could this be changed to `@attr` without the extra configuration? I'm guessing that you did this to be consistent with the other attributes in this file. Maybe we should update all of them unless there is a reason for the way they are written that I'm not aware of.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -43,6 +53,17 @@ export class ThemeProvider extends FoundationElement {
     })
     public theme: Theme = Theme.light;
 
+    public langChanged(
+        _prev: string | undefined,
+        next: string | undefined
+    ): void {
+        if (next !== undefined && next !== null) {
```
mollykreis, line 60: It's interesting that we're checking for `null` here when that isn't called out as an allowed value. Any idea why? Looks like that's also the case for `directionChanged` and `themeChanged`.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,6 +38,11 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
+    @attr({
+        attribute: 'lang'
+    })
+    public override lang: string = document.documentElement.lang;
```
mollykreis, line 44: I think it would be good to have some unit tests for the new attribute.

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add locale support to date-text column",
```
mollykreis, line 3: This PR is not only adding locale support to the date-text column, it's adding locale support to the nimble-theme-provider as a whole. I think that's important to capture in the change file.

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: I think you could link #1293 to this PR. Or, possibly just close it if it is just a duplicate work item.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: I'm not sure if the property could reasonably end up with the value `null`. I assume there is a way, since the other property change handlers are written the same way.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: Added a couple tests to validate connection between attribute and design token.

packages/nimble-components/src/theme-provider/index.ts
msmithNI: Calling `element.removeAttribute(attributeName)` will cause the `Changed()` methods to be called with `next` as `null` (similarly, going from no attribute set to another value yields `_prev` as `null`)

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -259,6 +272,26 @@ export class TableColumnDateText extends TableColumnTextBase {
         };
         return options;
     }
+
+    private validateConfig(formatter: Intl.DateTimeFormat | undefined): void {
```
mollykreis, line 276: It doesn't look like you're actually using `formatter` in this function other than to determine if the configuration is valid (`formatter !== undefined`) or invalid (`formatter === undefined`). I think the intent of this function would be clearer if you pass a boolean rather than the formatter.

mollykreis (approved):

packages/nimble-components/src/theme-provider/index.ts
m-akinc: I updated the signatures as suggested.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +38,29 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
-    @attr({
-        attribute: 'direction'
-    })
+    @attr()
+    public override lang: string = document.documentElement.lang;
```
msmithNI, line 42: I think we might need to explicitly fallback to 'en-US' as the default if this isn't a non-empty string:
- On a page where the `lang` attribute is not present on the `html` element, `document.documentElement.lang` is `''`. It looks like writing `lang=""` is also [allowed (it means a page with undetermined languge)](https://www.w3.org/International/questions/qa-no-language#undetermined)
- If we pass `''` as the locale to `Intl.DateTimeFormat` that yields an error:
    ```
    > new Intl.DateTimeFormat('', { dateStyle: 'medium', timeStyle: 'medium' });
    Uncaught RangeError: Incorrect locale information provided
    ```

I don't think we want our controls to error out if locale is unset, only if someone explicitly set `themeProvider.lang` to something invalid themselves.

(You might want to wait for @rajsite to weigh in before making changes though)

msmithNI (changes requested):

packages/nimble-components/src/theme-provider/index.ts
rajsite: Pulling the document.documentElement.lang at whatever random time this property gets evaluated is not sufficient. To respond to whatever the document configuration is we should actually observe it for changes. The patterns for having the theme provider defaults defined by the document are part of https://github.com/ni/nimble/pull/1257

So for now we should do what the other properties do and initialize to a know value that can be overwritten, ie. `en-US`.

I'm also not sure about making every element that wants to use a locale having to do all the validation. Seems like something that should be localized to where the local is configured. This should probably be a stand-up discussion topic or something.

m-akinc: We talked and came up with a plan:
- the `lang` token's default value will be whatever the page's `lang` attribute is (using a `MutationObserver` to track the value)
    - if the page's `lang` value is invalid, the token instead defaults to the system's default locale (`(new Intl.DateTimeFormat(undefined)).resolvedOptions().locale`)
- if the theme provider has a `lang` value set on it, that overrides the `lang` token's default.
    - if the theme provider's `lang` value is invalid, a validity flag is set on the theme provider, and the default `lang` token value is used instead.

Note that an "invalid" `lang` value is one that causes the `Intl.DateTimeFormat` constructor to throw a range error. Other seemingly "wrong' values can still resolve to a valid locale based on the resolution algorithm. For example, on my machine, the value "foo" resolves to "en-US". However "" and "123" do not resolve and are thus invalid.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +54,44 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
-    @attr({
-        attribute: 'direction'
-    })
+    @attr()
+    public override lang = '';
```
m-akinc, line 58: I want `lang` to initially be `undefined`, i.e. no override of the `lang` design token's default. However I can't type it as `string | undefined` because `HTMLElement`'s `lang` property is typed as just `string`. The best I could think to do was to initialize to `''` and just treat that value the same as `null`/`undefined` in my `langChanged` handler. This seems reasonable, because `''` is otherwise an invalid value -- if we pass that to the `Intl.DateTimeFormat` constructor, it throws an error.

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: Large refactor so please move to draft based on current policy https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#code-review

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -106,11 +110,22 @@ export class TableColumnDateText extends TableColumnTextBase {
...
+    };
+
     public override connectedCallback(): void {
         super.connectedCallback();
+        lang.subscribe(this.langSubscriber, this);
         this.updateColumnConfig();
     }
 
+    public override disconnectedCallback(): void {
+        lang.unsubscribe(this.langSubscriber, this);
```
mollykreis, line 126: You need to call `super.disconnectedCallback();` here.

```suggestion
    public override disconnectedCallback(): void {
        super.disconnectedCallback();
        lang.unsubscribe(this.langSubscriber, this);
```

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +54,44 @@ export const theme = DesignToken.create<Theme>({
...
+
+    public langChanged(
+        _prev: string | undefined | null,
+        next: string | undefined | null
+    ): void {
+        this.langIsInvalid = false;
+        if (
+            next !== undefined
+            && next !== null
+            && next !== ''
```
mollykreis, line 82: It looks like you're checking for all falsey values here, so why not just write this as
```suggestion
            next
```

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +54,44 @@ export const theme = DesignToken.create<Theme>({
...
+            // eslint-disable-next-line no-cond-assign
+            && !(this.langIsInvalid = isInvalidLang(next))
```
mollykreis, line 84: I think we should attempt not to have to disable our lint rules unless we have a good reason to. Is there a reason you decided that this would be best written in a way that requires disabling a lint rule?

packages/nimble-components/src/theme-provider/tests/theme-provider.spec.ts
```diff
@@ -16,6 +19,79 @@ describe('Theme Provider', () => {
...
+        let element: ThemeProvider;
+        let connect: () => Promise<void>;
+        let disconnect: () => Promise<void>;
+
+        afterEach(async () => {
+            await disconnect();
+        });
+
+        it('value is set to "foo" when theme provider lang attribute is assigned value "foo"', async () => {
+            ({ element, connect, disconnect } = await setup());
```
mollykreis, line 41: This test and the next few confused me until I saw your PR comment about `"foo"` and `"bar"` being acceptable locale values. I don't think it's immediately obvious that those are valid, so these tests seem wrong just looking at them. Could you make them more realistic by using a more correct locale? Perhaps even some that we typically see/use within NI products (e.g. `de-DE`, `fr-FR`, etc).

packages/nimble-components/src/theme-provider/tests/theme-provider.spec.ts
```diff
@@ -16,6 +19,79 @@ describe('Theme Provider', () => {
...
+        it('value defaults to page lang when theme provider lang attribute is malformed', async () => {
+            document.documentElement.lang = 'bar';
+            ({ element, connect, disconnect } = await setup('123'));
+            await connect();
+            await waitForUpdatesAsync();
+            expect(lang.getValueFor(element)).toBe('bar');
+            expect(element.validity.invalidLang).toBeTrue();
+        });
+
+        it('value updates when theme provider lang attribute goes from bad to good', async () => {
```
mollykreis, line 74: I think one more test here would be good that does this opposite of this (verifies that the correct thing happens when the theme provider's lang attribute is valid and then gets set to something invalid).

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +54,44 @@ export const theme = DesignToken.create<Theme>({
...
+    @attr()
     public direction: Direction = Direction.ltr;
 
-    @attr({
-        attribute: 'theme'
-    })
+    @attr()
     public theme: Theme = Theme.light;
 
+    public get validity(): ValidityObject {
```
mollykreis, line 66: Is there a reason not to have a `checkValidity()` function here like we do for our other components with validation?

mollykreis, line 66: This will ultimately need to be exposed through Angular & Blazor, right? Is the plan to do that in a different PR, or should they be added to this one?

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -8,13 +8,34 @@ import { Direction } from '@microsoft/fast-web-utilities';
...
 declare global {
     interface HTMLElementTagNameMap {
         'nimble-theme-provider': ThemeProvider;
     }
 }
 
+function isInvalidLang(value: string): boolean {
+    try {
+        // eslint-disable-next-line no-new
+        new Intl.DateTimeFormat(value);
```
mollykreis, line 23: Based on [these MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat), it doesn't look like you need to use the `new` keyword here. Could you avoid disabling the eslint rule by writing this like:

```suggestion
        Intl.DateTimeFormat(value);
```

packages/nimble-components/src/theme-provider/index.ts
m-akinc: I want to avoid calling `isInvalidLang` unless we need to, i.e. when `next` is truthy, and if I do the assignment inside the `if` block, I have to duplicate the `else` with `lang.deleteValueFor`. The documentation for the rule itself says, "There are valid reasons to use assignment operators in conditional statements." For these reasons, I just disabled the rule in this case.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: I'll do it in a follow-up PR.

packages/nimble-components/src/theme-provider/index.ts
rajsite: > The documentation for the rule itself says, "There are valid reasons to use assignment operators in conditional statements." For these reasons, I just disabled the rule in this case.

Eslint's description of the rule does not necessarily imply anything about the NI JavaScript Styleguide configuration of the rule. We follow airbnb's convention without modification which always disallows without description of exceptions: 

https://github.com/airbnb/javascript/blob/cda44dad403979d313a4b02ce55f0bb6cc3a24b2/packages/eslint-config-airbnb-base/rules/errors.js#L23C1-L24C43

So I'm going to push back and ask to come up with a pattern that avoids the mutation in the conditional

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +53,46 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
-    @attr({
-        attribute: 'direction'
-    })
+    @attr()
+    public override lang = '';
```
rajsite, line 57: I'll comment because I'm still thinking about it. A concern I have about overriding the lang attribute is that if we override the lang attribute we should match the attributes behavior identically. We shouldn't override an attribute and not align with the browser behavior.

My concern is that this does not behave identically. Setting to empty string will reflect that value to the dom. I don't think `<div lang>` and `<div>` are identical in terms of how they impact lang settings. Not sure but also not desirable attribute manipulation behavior, we should minimize changing the behavior of attributes that we don't own and setting the attr to empty string is a change in behavior to normal elements, i.e. any newly created `div` element won't magically get a `lang` attribute added to it.

m-akinc: We could leave `lang` without an initializer and just suppress the typescript error about the missing initializer by using a non-null assertion. I realize that's a lie, but if we have a comment that explains why we're lying, maybe that's a reasonable solution.

mollykreis: Is there a reason not to give this attribute a different name (e.g. `language` or `locale`) rather than trying to force it to align with the browser behavior?

m-akinc: Might be confusing to have two separate attributes that both effectively specify the locale to use for the contents. I could see someone using `lang` on accident when they should have used `language`.

rajsite: My concern is the DOM state getting out of sync with token state:

```html
<html lang="en">
<style>
div:lang(en) {color: red;}
div:lang(fr) {color: blue;}
</style>
<nimble-theme-provider language="fr">
    <div>Hello!</div>
</nimble-theme-provider>
</html>
```

https://jsbin.com/luvomuqosu/edit?html,js,output

To align the DOM state you have to do something like:

```html
<html lang="en">
<style>
div:lang(en) {color: red;}
div:lang(fr) {color: blue;}
</style>
<nimble-theme-provider lang="fr" language="fr">
    <div>Hello!</div>
</nimble-theme-provider>
</html>
```

https://jsbin.com/yupayavuza/edit?html,js,output

But if we could use the native lang then the DOM state is aligned:

```html
<html lang="en">
<style>
div:lang(en) {color: red;}
div:lang(fr) {color: blue;}
</style>
<nimble-theme-provider lang="fr">
    <div>Hello!</div>
</nimble-theme-provider>
</html>
```

https://jsbin.com/vaveyameju/edit?html,js,output


Our usage guidelines then become something like whenever you need to update the locale for a region of the page you have to use the theme provider. Don't set it on a random div, wrap the content in a theme provider and set `lang` and then the token state and the actual dom state align

rajsite: Discussed in post-stand-up:
1. No pushback on using the same attribute so that the DOM configuration and token configuration are controlled via a single property
2. The semantics of the lang attribute + token need to align with the DOM
   - If unset resolves to parent value?
   - If set to invalid value, resolve to parent?
   - If set to empty string (mdn describes as the "unknown" state) what is the behavior? Resolved to parent?
   - If the root (html element lang attr) is unset, resolve to system value? Is that the same as DateFormat(undefined) locale?
   - If the root is set to invalid value, resolve to system?
   - If the root is set to empty string, resolve to system?

Some research needed there, so moving to draft as per new contributing policy. Can `@mention` for discussions as researched and peer review, etc.

m-akinc: >If unset resolves to parent value?

Yes. But in the case of the token, we can't watch every ancestor element, so we only look at the root. If lang is unset on the theme provider, we fall back to the root's lang. If the root's lang is unset, the lang is _unknown_, equivalent to a value of empty string.

>If set to invalid value, resolve to parent?

No. The value remains that specific invalid value. It is up to consumers of the locale to determine how to handle an invalid value.

>If set to empty string (mdn describes as the "unknown" state) what is the behavior? Resolved to parent?

No, remains empty string.

>If the root (html element lang attr) is unset, resolve to system value? Is that the same as DateFormat(undefined) locale?

No, root unset just means _unknown_ locale. The `DateTimeFormat()` constructor resolves `undefined` to the browser's language, but empty string (unknown) and invalid values result in a thrown error.

>If the root is set to invalid value, resolve to system?

No, remains invalid value.

>If the root is set to empty string, resolve to system?

No, means the same as unset: _unknown_.

So...
1. our token may have an invalid/empty value. 
2. we may fail to create a date-time formatter with the token value
3. should we have a fall-back locale to create our formatter with when the token value is bad? Use the browser locale?

rajsite: > Yes. But in the case of the token, we can't watch every ancestor element, so we only look at the root. If lang is unset on the theme provider, we fall back to the root's lang. If the root's lang is unset, the lang is unknown, equivalent to a value of empty string.

Nope. We just discussed this. It's not the root but theme-providers in the cascade. Theme providers can be nested. Should look to the cascade (which is just how DesignTokens work, this is nothing new).

```html
<html lang="fr">
  <theme-provider lang="en">
    <theme-provider>
        <!-- Token lang value and DOM lang value is en, not fr -->
    </theme-prover>
  </theme-provier>
</html>
```

> our token may have an invalid/empty value.

The theme-provider lang attribute can be configured to an invalid value. Whether we use that same value on the token is the question. Need to understand the trade-offs of letting the token value be invalid to align with DOM and making every component responsible for invalid values or choosing to differ from the DOM and having the token value always resolve to a valid value.

rajsite: What research was done to determine the default locale? 

Why was the DateFormatter(undefined) option chosen as the default over navigator.language? 

https://stackoverflow.com/questions/57666095/browser-default-locale-intl-datetimeformat-vs-navigator-language

m-akinc: > Nope. We just discussed this. It's not the root but theme-providers in the cascade and then the root.

I was just describing the single theme provider case. My understanding is that we don't have to do anything special to handle the nested theme-provider case. We're just concerned with what the token defaults to, because that is what applies when the theme provider cascade provides no `lang` value.

> Need to understand the trade-offs...

If we allow our token to have an invalid/unknown value, there are a couple primary drawbacks:
- any code using the token value must be aware of and add logic to handle an invalid/unknown value
- different consumers of the token may handle invalid/unknown values inconsistently, i.e. not falling back to another locale, or falling back to a different one

The only advantage of allowing invalid/unknown values is:
- resulting behavior/errors may make it easier to detect and fix a bad value that was specified by accident

Based on these points, I would advocate for preventing invalid/unknown locale values in our `lang` token.

> What research was done to determine what we should use as the default locale?
> Why was the DateFormatter(undefined) option chosen as the default over navigator.language? 

I thought following the precedent of `Intl.DateTimeFormat`'s handling of an unspecified locale was a good approach. I didn't know about `navigator.language`. It appears that `Intl.DateTimeFormat` falls back to the system locale, whereas `navigator.language` uses the browser's language setting. Both seem reasonable to me, but maybe we should prefer the latter, since it is more tightly scoped.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +53,50 @@ export const theme = DesignToken.create<Theme>({
...
+            if (!this.langIsInvalid) {
+                shouldSetNewTokenValue = true;
+            }
+        }
+
+        if (shouldSetNewTokenValue) {
+            lang.setValueFor(this, next!);
+        } else {
+            lang.deleteValueFor(this);
+        }
```
rajsite, line 94: This boolean flag tracking is pretty confusing. Think it can be simplified just making sure each code path updates the validity and token value as needed.

```ts
if (next === null || next === undefined) {
    lang.deleteValueFor(this);
    this.langIsInvalid = false;
    return;
}


if (isInvlaidLang(next)) {
    lang.deleteValueFor(this);
    this.langIsInvalid = true;
} else {
    lang.setValueFor(this, next);
    this.langIsInvalid = false;
}
```

rajsite: Actually based on the other thread I'm not sure if in the isInvalidLang(next) should result in clearing the token (i.e. falling back to the cascade).

The other thread:

>> If set to invalid value, resolve to parent?
>
> No. The value remains that specific invalid value. It is up to consumers of the locale to determine how to handle an invalid value.

Says the DOM value does not fallback to the cascade at that point. Trying to understand if there is a con for falling back to the cascade.

m-akinc: Based on the conversation in the other thread I thought we had decided that our token value would not match the DOM behavior, i.e. our token would not take on invalid/unknown values. Are we on the same page? If yes, then the logic will stay the same (except I can refactor a bit for clarity).

packages/nimble-components/src/utilities/models/page-locale.ts
```diff
@@ -0,0 +1,27 @@
+import { observable } from '@microsoft/fast-element';
+
+/**
+ * Observable class to subscribe to changes in the page's lang attribute
+ */
+class PageLocale {
```
rajsite, line 6: lets make a more straight forward name aligning to the html element lang attr

packages/nimble-components/src/theme-provider/index.ts
m-akinc: Okay, talked on Teams -- we're going to fall back to the cascade, then if there are no valid values in the cascade, the token will try to use the documentElement's `lang` value, then if that is invalid/missing, we'll fall back to "en-US".

packages/nimble-components/src/theme-provider/index.ts
m-akinc: We changed our decision to fall back to "en-US" instead of navigator.language.

Where should we document this? We don't have a Storybook story for the theme provider, but maybe we should make one? Don't know what it would render, though.

rajsite: Can follow the label-provider doc patterns, see @msmithNI

m-akinc: I don't know how that would translate to this `lang` token documentation. For label-provider, we created a "Label Providers" Docs page under the "Tokens" section. It seems like the analog would be a "Theme Provider" Docs page, but I question the discoverability of a "lang" token under "Theme Provider" docs. Since the section is "Tokens", should we just give the `lang` token its own docs page as a sibling of "Base Colors", "Theme-aware Tokens", and "Label Providers"? If it helps, we could discuss this over Teams.

rajsite: Think the discoverability of label provider under tokens is questionable for discoverability as well so I don't think your alternative proposals provide significant advantage.

I'd suggest making theme-provider next to label-providers for now and it can document its theme, direction, and lang tokens. We can revisit how we document them in storybook in the future.

If that is not palatable now, I don't have the bandwidth for this topic this week. Can brainstorm with another team member.

m-akinc: Added mdx for theme provider with documentation for theme, lang, and direction.

msmithNI (approved):

packages/nimble-components/src/theme-provider/tests/theme-provider.stories.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export default metadata;
+
+const langDescription = `Defines the language of the element. See [external documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) for details.
+
+Users should set \`lang\` on the root \`html\` element of the page to reflect the language of the content. If necessary, users may override the language for a subtree by inserting a \`nimble-theme-provider\` element and setting its \`lang\` attribute. Nimble elements will not honor a \`lang\` value set on any other type of ancestor element.`;
+
+export const themeProvider: StoryObj = {
+    render: createUserSelectedThemeStory(hiddenWrapper(html`<style></style>`)),
```
jattasNI, line 24: This is probably feature creep, so feel free to reject. But it'd be cool to populate the HTML with a theme provider and some elements that respect the tokens set on it (like a table with a date column for the `lang` attribute and maybe also some content that uses background tokens that respect the `theme` attribute). Then the Storybook controls could affect the theme provider and users could see the impact interactively.

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
jattasNI, line 1: It'd be good to mention in the date column docs that it respects the lang token to do locale-specific formatting (and maybe link to the new theme provider docs). I think adding it to `dateTextColumnDescription` is fine, even though I believe that's not yet visible to Storybook users since we don't have a Docs page for the columns yet. We will someday!

packages/nimble-components/src/theme-provider/tests/theme-provider.spec.ts
```diff
@@ -16,6 +19,90 @@ describe('Theme Provider', () => {
         );
     });
 
+    describe('lang token', () => {
```
jattasNI, line 22: I may be missing something, but are we actually testing `langChanged()` and the mutation observer? It looks like we only modify the document `lang` before the `theme-provider` is attached.

Assuming that is a gap and that it's possible to write tests for it, I'd hope for tests that cover all the combinations of `if (next)` and `isInvalidLang(next)` in `langChanged()`.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +49,50 @@ export const theme = DesignToken.create<Theme>({
...
  */
 export class ThemeProvider extends FoundationElement {
-    @attr({
-        attribute: 'direction'
-    })
+    @attr()
+    public override lang = '';
+
+    @attr()
     public direction: Direction = Direction.ltr;
```
rajsite, line 56: same as above, we should not be writing to the attribute because then it breaks normal inheritance

packages/nimble-components/src/theme-provider/index.ts
rajsite: Unresolving, please update to a pattern similar to the first comment in the thread instead of the `houldSetNewTokenValue` pattern

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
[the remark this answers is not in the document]
m-akinc: I have a personal note to expose the `lang` property in the Angular directive for the theme provider, and that will include the usual tests. Why do we need to additionally test changing the `lang` in SLE? We do have a test for the date-text column that changing the `lang` token affects the rendered string.

packages/nimble-components/src/theme-provider/tests/theme-provider.stories.ts
```diff
@@ -0,0 +1,50 @@
...
+    argTypes: {
+        theme: {
+            description:
+                'The display theme to use. One of `Theme.light`, `Theme.dark`, or `Theme.color`.',
```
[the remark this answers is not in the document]
rajsite: We should do the attribute values. That's what we do for other enum style attributes and what users would use in Angular templates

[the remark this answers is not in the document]
m-akinc: I know that's our convention for the values listed by radio buttons, but I thought it would be better to document the `Theme` members, as that's what we want clients to use. But I don't feel strongly, so I changed it for `lang` and for `direction`. I also realized that the `Direction` enum is a FAST type, and we weren't re-exporting it. So I did that, too.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -8,13 +8,29 @@ import { Direction } from '@microsoft/fast-web-utilities';
...
 }
 
+function isInvalidLang(value: string): boolean {
+    try {
+        Intl.DateTimeFormat(value);
+        return false;
+    } catch (e) {
+        return true;
+    }
+}
```
rajsite, line 27: nit: can you avoid the unnecessary negation. Let's just call this `isValidLang`

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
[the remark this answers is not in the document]
mollykreis: My understanding of @jattasNI's statement is that we need to verify (at least manually) that changing the locale in SLE sets the appropriate language on the page such that the `nimble-theme-provider` picks up that value.

Looking at what happens today, SLE does set the `lang` attribute on `document.documentElement`, but it is just the two letter language code (e.g. `de` rather than `de-DE`). Will that work as expected, or do we need to make an SLE change to have it set the longer language code?

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -8,13 +8,29 @@ import { Direction } from '@microsoft/fast-web-utilities';
...
 
 declare global {
     interface HTMLElementTagNameMap {
         'nimble-theme-provider': ThemeProvider;
     }
 }
 
+function isInvalidLang(value: string): boolean {
+    try {
+        Intl.DateTimeFormat(value);
```
mollykreis, line 22: The use of `Intl.DateTimeFormat` here seems a bit arbitrary. Could you pass the value into the constructor of `Intl.Locale` instead, or does `DateTimeFormat` give extra validation? Part of my question stems from the fact that we'll need this in the number column soon as well, so I would hope that any locale that's valid for the date formatter would also be valid for the number formatter. Therefore, it would be nice to write this in a way that wasn't tied to a specific usage of the locale.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +49,50 @@ export const theme = DesignToken.create<Theme>({
...
 
+    public get validity(): ValidityObject {
+        return {
+            invalidLang: this.langIsInvalid
+        };
+    }
+
+    private langIsInvalid = false;
+
+    public checkValidity(): boolean {
```
mollykreis, line 69: Should we use the `Validator` base class here so that we aren't manually tracking the validity keys in this class?

packages/nimble-components/src/theme-provider/tests/theme-provider.spec.ts
m-akinc: We are testing `langChanged()`. It gets called both when we set the attribute in html and when we set the attribute programmatically. We have test cases that cover all code paths in `langChanged()`.

You're right that I missed testing the mutation observer. I've now added a test to cover that, where we change the document element's `lang` after the theme provider is attached.

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +49,50 @@ export const theme = DesignToken.create<Theme>({
  * @internal
  */
 export class ThemeProvider extends FoundationElement {
-    @attr({
-        attribute: 'direction'
-    })
+    @attr()
+    public override lang = '';
```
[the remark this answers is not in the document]
m-akinc: Had to add a TS suppression to do that. Leaving it uninitialized requires adding `undefined` to the type, but I can't do that because its type has to match the `lang` attribute that it's overriding. If I don't use the `override` keyword, that also produces an error.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: Sorry, in all the discussion about our fallback behavior, I lost sight of the original ask.

packages/nimble-components/src/theme-provider/index.ts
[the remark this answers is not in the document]
rajsite: You could possibly use a non-null assertion

```ts
    @attr
    public override lang!;
```

change/@ni-nimble-components-df093e6a-7cba-46ed-8d5d-07670af90719.json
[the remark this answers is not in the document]
m-akinc: It may be dependent on what is using it. For example, the `Intl.DateTimeFormat` constructor supports a `localeMatcher` option that affects how the runtime finds a suitable locale to use based on what was requested. I'm willing to bet that most locale-aware code would work similarly though, such that the region subtag isn't necessary to resolve to a suitable locale.

packages/nimble-components/src/theme-provider/tests/theme-provider.stories.ts
m-akinc: I just added a table, as that reflects the `theme` as well as the `lang`. I don't know how to demo the `direction` setting, but I left it selectable anyway.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: That works, but I did have to add an eslint suppression (for calling `new` without using the object). For some reason I was able to construct an `Intl.DateTimeFormat` without using `new`, but can't for `Intl.Locale`.

packages/nimble-components/src/theme-provider/index.ts
m-akinc: It's so trivial that I think it's actually simpler this way. If we end up having to do more validation, I would probably start using a `Validator`.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
```diff
@@ -106,7 +106,7 @@ interface TextColumnTableArgs extends SharedTableArgs {
     validity: () => void;
 }
 
-const dateTextColumnDescription = 'The `nimble-table-column-date-text` column is used to display date-time fields as text in the `nimble-table`. The date-time values must be of type `number` and represent the number of milliseconds since January 1, 1970 UTC. This is the representation used by the [JavaScript `Date` type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).';
+const dateTextColumnDescription = 'The `nimble-table-column-date-text` column is used to display date-time fields as text in the `nimble-table`. The date-time values must be of type `number` and represent the number of milliseconds since January 1, 1970 UTC. This is the representation used by the [JavaScript `Date` type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date). Dates are formatted in a locale-specific way based on the value of the `lang` token, which can be set via the [`nimble-theme-provider`](/docs/tokens-theme-provider--docs).';
```
jattasNI, line 109: Storybook links are finicky to get working in local builds, PR builds, and the deployed site. We have guidance about how they're supposed to look (should work right, but I didn't test it for this particular link): https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/creating-storybook-component-documentation.md#markdown

```suggestion
const dateTextColumnDescription = 'The `nimble-table-column-date-text` column is used to display date-time fields as text in the `nimble-table`. The date-time values must be of type `number` and represent the number of milliseconds since January 1, 1970 UTC. This is the representation used by the [JavaScript `Date` type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date). Dates are formatted in a locale-specific way based on the value of the `lang` token, which can be set via the [`nimble-theme-provider`](?path=/docs/tokens-theme-provider--docs).';
```

packages/nimble-components/src/theme-provider/tests/theme-provider.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+            defaultValue: {
+                summary:
+                    '`lang` of the document element if set, otherwise "en-US".'
+            },
+            options: ['en-US', 'fr-FR', 'de-DE'],
+            control: { type: 'radio' }
+        },
+        direction: {
+            description:
+                'The text direction of the element. Either `"ltr"` or `"rtl"`. The `Direction` type exposes these values.',
```
jattasNI, line 104: nit: don't want to give clients the wrong idea about how well this will work
```suggestion
                'The text direction of the element. Either `"ltr"` or `"rtl"`. The `Direction` type exposes these values. Note: Right-to-left support in Nimble is untested. If you need this capability, please file an issue.',
```

packages/nimble-components/src/theme-provider/tests/theme-provider.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+    args: {
+        ...sharedTableArgs(simpleData)
+    }
+};
+
+export default metadata;
+
+const langDescription = `Defines the language of the element. See [external documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) for details.
+
+Users should set \`lang\` on the root \`html\` element of the page to reflect the language of the content. If necessary, users may override the language for a subtree by inserting a \`nimble-theme-provider\` element and setting its \`lang\` attribute. Nimble elements will not honor a \`lang\` value set on any other type of ancestor element.`;
```
jattasNI, line 55: nit: distinguishing between end-users and developers
```suggestion
Applications should set \`lang\` on the root \`html\` element of the page to reflect the language of the content. If necessary, users may override the language for a subtree by inserting a \`nimble-theme-provider\` element and setting its \`lang\` attribute. Nimble elements will not honor a \`lang\` value set on any other type of ancestor element.`;
```

jattasNI (approved):

packages/nimble-components/src/theme-provider/tests/theme-provider.spec.ts
```diff
@@ -16,6 +19,122 @@ describe('Theme Provider', () => {
...
+            );
+        }
+
+        let element: ThemeProvider;
+        let connect: () => Promise<void>;
+        let disconnect: () => Promise<void>;
+        let pageLangToRestore: string | undefined;
+
+        beforeEach(() => {
+            pageLangToRestore = document.documentElement.lang;
```
rajsite, line 40: When accessing the property value you can't distinguish between an absent lang and a lang of empty string. The current code is using a truthiness check to assume that a lang of empty string means remove the lang but it can't actually accurately reflect the state of what's configured.

I'd avoid all the mental gymnastics and switching between manipulating the property and the attribute and just use the attribute consistently in this case.

```
pageLangToRestore = document.documentElement.getAttribute('lang');
if (pageLangToRestore === null) {
    document.documentElement.removeAttribute('lang');
} else {
    document.documentElement.setAttribute('lang', pageLangToRestore);
}

```

rajsite (approved):

packages/nimble-components/src/theme-provider/index.ts
```diff
@@ -33,19 +53,49 @@ export const theme = DesignToken.create<Theme>({
...
-    public direction: Direction = Direction.ltr;
-
-    @attr({
-        attribute: 'theme'
-    })
+    @attr()
+    public override lang!: string;
+
+    @attr()
+    public direction?: Direction;
```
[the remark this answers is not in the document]
m-akinc: Created #1478

PR #1411, opened 2023-08-01
Add docs for 'target' attribute on anchor elements

mollykreis:
# Pull Request

## 🤨 Rationale

We should document guidelines of when it is appropriate to use the `target` attribute on anchor elements.

Ultimately, we should also have documented guidelines for how to convey that an anchor will open in a new tab/window, but we don't have those guidelines yet (see #1412).

## 👩‍💻 Implementation

- Created mdx usage docs for setting the `target` attribute on anchor attributes
- Moved more components to use mdx storybook format:
    - anchor tabs
    - anchor
    - menu / menu item / anchor menu item
    - tree / tree item / anchor tree item
    - anchor table column
- Included the new usage information on all anchor related components:
    - anchor button
    - anchor tabs
    - anchor
    - anchor menu item
    - anchor tree item
    - anchor table column

## 🧪 Testing

Manually verified that the updated storybook pages look correct.

## ✅ Checklist


packages/nimble-components/src/utilities/tests/storybook.ts
m-akinc, line 1: Our precedent for doc strings that are specific to a certain group of controls is to put them under `patterns`. E.g. see `patterns/button/tests/doc-strings.ts`.

m-akinc (approved):
I think I agree that we do not to add anything to the anchor tabs' documentation.

packages/nimble-components/src/utilities/tests/storybook.ts
```diff
@@ -178,3 +178,11 @@ export const disableStorybookZoomTransform = `
     }
 </style>
 `;
+
+export const anchorTargetGuidelines = `
```
rajsite, line 182: Forgot to clean-up after a refactor? Don't forget to self review the changes 👍

packages/nimble-components/src/anchor-button/tests/anchor-button.stories.ts
```diff
@@ -35,8 +36,8 @@ const metadata: Meta<AnchorButtonArgs> = {
     parameters: {
         docs: {
             description: {
-                component:
-                    'An anchor button is a component with the visual appearance of a button, but it navigates like an anchor/link when pressed.'
+                component: `An anchor button is a component with the visual appearance of a button, but it navigates like an anchor/link when pressed.
+                    ${targetGuidelines}`
```
rajsite, line 40: This is very much usage docs so we should go ahead and add mdx files for these stories and use that pattern.

See the button mdx docs: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/button/tests/button.mdx

Specifically things to note:
- See the pattern for sharing doc fragments from mdx files, i.e. the `ContentHiddenDocs`
- You'll need to remove the autodocs tags: https://github.com/ni/nimble/blob/fde3626559e2469f8540c45d9ff774838b49bb75/packages/nimble-components/src/anchor/tests/anchor.stories.ts#L23

packages/nimble-components/src/anchor-button/tests/anchor-button.stories.ts
```diff
@@ -35,8 +36,8 @@ const metadata: Meta<AnchorButtonArgs> = {
...
             }
```
rajsite, line 41: Looking at the button mdx docs I'm also realizing we are accidentally pulling in all mdx files as stories:

image

Because of:

https://github.com/ni/nimble/blob/fde3626559e2469f8540c45d9ff774838b49bb75/packages/nimble-components/.storybook/main.js#L4-L9

Would be nice in this PR or a tech debt issue + follow-up to rename our mdx files that are actually stories to be `*.stories.mdx` and then update that selector to reflect that:

```js
export const stories = [
    '../docs/**/*.stories.mdx',
    '../docs/**/*.stories.ts',
    '../src/**/*.stories.mdx',
    '../src/**/*.stories.ts'
];
```

change/@ni-nimble-components-63ba584c-0659-44df-8b5c-65ebf12532b9.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
mollykreis: If anchor-tabs are trying to behave like tabs, just with each tab having its own URL, it doesn't seem like it would ever be within the usage of the anchor-tab to open in a new tab/window by default. However, I think the same argument could probably be made for the anchor-tree-item as well.

Is your position that we should include the docs on all our anchor elements? I don't mind doing that. It would certainly be less presumptive about how the components are being used.

packages/nimble-components/src/anchor-button/tests/anchor-button.stories.ts
mollykreis: I made a quick effort to try to rename the relevant files, but I ran into problems. It looks like storybook does not treate `.mdx` files and `.stories.mdx` files the same, so we can't simply change the extension because we want to. That behavior is discussed in this storybook issue: https://github.com/storybookjs/storybook/issues/20496

I can create a tech debt item for fixing the fact that all of our mdx files are showing up in storybook since it looks like we'll need to have a bit more discussion about how we want to distinguish internal/external mdx files.

rajsite: Lovely, thanks for investigating. Yea we can make a tech debt to figure out how to handle those helper mdx files in the storybook nav tree. But I think we should go ahead and use the pattern for the PR even if it's adding a bit more to that tech debt.

change/@ni-nimble-components-63ba584c-0659-44df-8b5c-65ebf12532b9.json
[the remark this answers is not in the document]
rajsite: Yea I think we can start by having the same guidance on all anchors. I don't think the advice is wrong in those cases, and if we refine it more we can specialize (i.e. anchor-tabs should likely never need a blank target, etc). Think it's better than not providing guidance in this case.

packages/nimble-components/src/anchor-button/tests/anchor-button.stories.ts
mollykreis: Created #1419

packages/nimble-components/src/menu/tests/menu.stories.ts
```diff
@@ -262,19 +246,24 @@ export const nestedMenu: StoryObj<MenuArgs> = {
...
-    parameters: {
-        docs: {
-            description: {
-                story: 'The menu can be configured to display a custom header with arbitrary content.  This could include other nimble components, or even custom elements.'
-            }
-        }
-    },
     // prettier-ignore
     render: createUserSelectedThemeStory(html`
+        <style>
```
mollykreis, line 251: This change wasn't necessary as part of this PR, but I noticed that we didn't have our custom menu header content styled using nimble tokens. Since we specifically have docs saying that you can do this, I thought it was important to set a good example by using tokens to style it.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+<Title of={anchorTabsStories} />
+<Description of={anchorTabsStories} />
+
+<DocsStory of={anchorTabsStories.anchorTabs} expanded={false} />
+<Controls of={anchorTabsStories.anchorTabs} />
+
+# Usage Docs
+
+## Accessibility
```
m-akinc, line 22: Why is this Accessibility header here? I don't think the target docs are really an accessibility topic (more general principles of web design), and the following paragraph about Angular is definitely not accessibility. Not sure if we should separate that second paragraph under an Angular subheading, or just leave everything directly under Usage Docs.

packages/nimble-components/src/anchor-button/tests/anchor-button.mdx
```diff
@@ -22,3 +23,5 @@ import * as anchorButtonStories from './anchor-button.stories';
 ## Accessibility
 
 <ContentHiddenDocs buttonElement="nimble-anchor-button" />
+
+<TargetDocs />
```
m-akinc, line 27: Is this really an accessibility concern, or more general?

packages/nimble-components/src/anchor/tests/anchor.mdx
```diff
@@ -0,0 +1,24 @@
...
+<Description of={anchorStories} />
+
+<DocsStory of={anchorStories.anchor} expanded={false} />
+<Controls of={anchorStories.anchor} />
+
+# Usage Docs
+
+## Accessibility
+
+<TargetDocs />
```
m-akinc, line 24: Again, is this really accessibility?

packages/nimble-components/src/anchor/tests/anchor.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+<Title of={anchorStories} />
+<Description of={anchorStories} />
+
+<DocsStory of={anchorStories.anchor} expanded={false} />
+<Controls of={anchorStories.anchor} />
+
+# Usage Docs
+
+## Accessibility
```
m-akinc, line 22: Not sure why, but this is rendering weird:
image

m-akinc (approved):

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.mdx
mollykreis: My thoughts were that routing correctness was part of making an app accessible, but I can see your point of view that it's really just overall usability rather than strictly an accessibility consideration. I don't have a strong preference. I'll move them out from under the _Accessibility_ header.

packages/nimble-components/src/anchor/tests/anchor.mdx
mollykreis: Ah, I see the problem for this component specifically. There are styles in the anchor story that are affecting these docs. I'll fix that.

packages/nimble-components/src/anchor/tests/anchor.stories.ts
```diff
@@ -33,15 +32,15 @@ const metadata: Meta<AnchorArgs> = {
     // prettier-ignore
     render: createUserSelectedThemeStory(html`
         <style class='code-hide'>
-            * {
+            span {
```
mollykreis, line 35: I made this change so that the `*` styling wouldn't mess with the styling of the mdx docs.

packages/nimble-components/src/menu/tests/menu.stories.ts
```diff
@@ -262,19 +246,24 @@ export const nestedMenu: StoryObj<MenuArgs> = {
...
+                font-weight: var(${bodyEmphasizedFontWeight.cssCustomProperty});
+            }
+
+            .header-secondary-text {
+                font: var(${bodyFont.cssCustomProperty});
+                color: var(${bodyFontColor.cssCustomProperty});
+            }
+        </style>
         <${menuTag}>
             <div style="display: grid; font-family: Source Sans Pro; font-size: 11px;">
```
rajsite, line 264: Can you move this to the styles with a `header-wrapper` class or seomthing?

packages/nimble-components/src/menu/tests/menu.mdx
```diff
@@ -0,0 +1,33 @@
...
+-   `nimble-menu-item` - Use this child element to execute a function when the item is activated or to include a sub-menu. To include a sub-menu, place a `nimble-menu` as a child of a `nimble-menu-item`.
+-   `nimble-anchor-menu-item` - Use this child element to navigate to a URL when the item is activated.
+
+### `nimble-anchor-menu-item` Target Configuration
+
+<TargetDocs />
+
+## Custom Header
+
+The menu can be configured to display a custom header with arbitrary content. This could include other nimble components, or even custom elements.
```
rajsite, line 33: I think we may want to render the customMenu story below this description so someone can see the show code example

packages/nimble-components/src/menu/tests/menu.mdx
```diff
@@ -0,0 +1,33 @@
...
+<Controls of={menuStories.menu} />
+
+# Usage Docs
+
+## Child Elements
+
+-   `nimble-menu-item` - Use this child element to execute a function when the item is activated or to include a sub-menu. To include a sub-menu, place a `nimble-menu` as a child of a `nimble-menu-item`.
+-   `nimble-anchor-menu-item` - Use this child element to navigate to a URL when the item is activated.
+
+### `nimble-anchor-menu-item` Target Configuration
```
rajsite, line 27: For markdown headers prefer less punctuation, makes links and ToCs a little easier generally

```suggestion
### Anchor Menu Item Target Configuration
```

packages/nimble-components/src/tree-view/tests/tree-view.mdx
```diff
@@ -0,0 +1,29 @@
...
+<Controls of={treeViewStories.treeItem} />
+
+# Usage Docs
+
+## Child Elements
+
+-   `nimble-tree-item` - Use a this child element if you want a tree item that calls a callback, has a value, and/or has child items.
+-   `nimble-anchor-tree-item` - Use this child element to navigate to a URL when the item is activated.
+
+### `nimble-anchor-tree-item` Target Configuration
```
rajsite, line 27:
```suggestion
### Anchor Tree Item Target Configuration
```

packages/nimble-components/src/patterns/anchor/tests/target-docs.mdx
```diff
@@ -0,0 +1,5 @@
+In general, when adding an anchor element to a web application, the anchor should not have a `target` configured on it to control how the link is opened.
+Instead, the user should be able to decide how they want to navigate using browser features, such as `CTRL + Click` or `Right-click >> Open New Tab`. For more information
```
rajsite, line 2: Can you split this into two paragraphs starting at "For more information"

packages/nimble-components/src/patterns/anchor/tests/target-docs.mdx
```diff
@@ -0,0 +1,5 @@
+In general, when adding an anchor element to a web application, the anchor should not have a `target` configured on it to control how the link is opened.
```
rajsite, line 1: I think specifically we are concerned about a target configured for `_blank` with the intention of opening in a new window or tab.

The other target configurations are use-case specific and not really a big deal / worth calling out (i.e. `_top` / `_parent` are for iframe things and other names are for targeting specific named windows which is real uncommon nowadays).

packages/nimble-components/src/anchor/tests/anchor.stories.ts
```diff
@@ -33,15 +32,15 @@ const metadata: Meta<AnchorArgs> = {
...
                 font: var(${bodyFont.cssCustomProperty});
             }
         </style>
-        Click on the <${anchorTag}
+        <span>Click on the <${anchorTag}
```
rajsite, line 39: Can you give the span a class name to target (just in-case span's get emitted in mdx as well to avoid conflicts) and also the `code-hide-top-container` so it's not shown in the showcode view. Actually the `code-hide-top-container` class may be enough to target.

rajsite: I don't feel strong enough for any particular option. Showing the span is probably fine.

rajsite (approved):

Issue #1011, opened 2023-01-30
Table allows configuration of number column type


nate-ni:
## 😯 Problem to Solve
A user needs to display numeric values to the user. These values may be integer or decimal based numbers.

The end user would like to be able to sort, filter, etc. based on these values. 
Note: This feature would not provide that direct functionality but is a consideration for this feature.

## 💁 Proposed Solution
The user should be able to provide simple, supported, configurations easily. 

## Open Questions
| Question                                                                      | Initial Recommendation     | Resolved | Decision |
|-------------------------------------------------------------------------------|----------------------------|----------|----------|
| Which formatting options do we need to provide?                               | Integer and Decimal        | No       |          |
| Should we provide config for prefix? (e.g. currency symbols)                  | No                         | No       |          |
| Should we provide config for prefix or suffix? (e.g. static string for units) | Yes                        | No       |          |
| How should we provide alignment?                                              | Always right-align content | No       |          |
| Which width modes should we support?                                | Both fixed pixel width and fractional width | No       |          |

## Other Considerations
Supporting the users local for number formatting
More info: [Intl.NumberFormat - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#examples)

## References/Prior Art
[jQuery Grid Cells Formatting](https://www.jqwidgets.com/jquery-widgets-documentation/documentation/jqxgrid/jquery-grid-cellsformatting.htm)

```[tasklist]
## 📋 Tasks
- [ ] #887
- [ ] #1054
- [ ] #1224
- [x] Create base classes for text columns
- [x] Define API for numeric formatting
- [x] Define API for text alignment
- [x] Create custom elements for number column with basic API and default numeric formatting
- [x] Unit tests for formatting number edge cases (Inf, NaN, etc)
- [x] Add API to configure integer numeric formatting
- [x] Add API to configure decimal numeric formatting
- [ ] https://github.com/ni/nimble/issues/1293
- [x] Respect Nimble locale token for numeric formatting
- [x] Unit tests for invalid state with bad formatting configuration
- [x] Add API to configure text alignment
- [x] Angular integration
- [x] Blazor integration
```

jattasNI:
I was curious about what kind of numeric formatting options we need. Here are the use cases I found in SLE that use either a number column type or a custom format function:
[...]

Issue #887, opened 2022-12-06
Visual Design for basic table content types


nate-ni:
## 😯 Problem to Solve
Provide designs for basic content types outside of what was already completed from the initial design 
- #883

### In table cells
- Numeric values (Related dev task: #1011)
- Hyperlinks (Related dev task: #1012)
- Icons (Related dev task: #1013 )
- DateTime (Related dev task: #1014)

### In table header cells
- Icon
- Text and Icon

## 🤔 Open Questions
1. What size should the icons be?
2. Do we need default alignments for different data types?

## 🖼️ References/Prior Art
Design for basic table
- #883

jattasNI:
@RickA-NI here are a couple prototypes of right-aligned numeric columns. They reveal some interesting questions for you to provide guidance on:
1. For numeric columns, should we right align the values, both the values and the header, or neither?
2. If we right align the values, should we also set [`font-variant-numeric: tabular-nums`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric) to request that the browser render digits with equal widths? Does our font support this setting? (I couldn't see a difference when I played with it)

![image](https://user-images.githubusercontent.com/10500124/216631469-ebc5513e-ce2f-42a3-86b3-d38feedbd469.png)

![image](https://user-images.githubusercontent.com/10500124/216631867-a26c4615-949b-4328-a36f-524fd486f358.png)

RickA-NI:
1) My initial reaction is that headers should match their data. I'm not sure what the ramifications are for all the stuff (sorting direction, menu icons, grab handles, etc.) we've talked about stick up there but visually it looks better and keeps the header from becoming dissociated from the data as in your example.
2) Source Sans Pro renders digits with equal widths by default. You shouldn't need to turn that on unless we're trying to force it into a monospaced like mode where the decimal places and commas take up equal space to a didgit.

nate-ni:
I've always thought....strings to the left, numbers to the right...but it is challenging to visually connect data when you don't have borders in the headers to group columns, like we've got.

![Image](https://user-images.githubusercontent.com/109235103/216718117-aeca7292-7b43-4703-941a-2b3412293376.png)

RickA-NI:
I found some blog on table design and it had three (and a half) rules for alignment that seem to jive pretty well with what we're all aligning on here:
![image](https://user-images.githubusercontent.com/104936907/216789239-73cb9987-2398-4426-883c-b5abad2ddfe0.png)

[...]

PR #1414, opened 2023-08-02
Initial implementation of nimble-table-column-number-text

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1011 

This PR adds the initial implementation of the `nimble-table-column-number-text` with configuration to render the number using a default formatter or as an integer.

There are a number of additional configurations that still need to be added to the column, such as:
- Text alignment (right vs left)
- Additional formatting options (decimal and custom)
- Prefix & suffix
- Locale-aware formatting

## 👩‍💻 Implementation

The new column follows the same patterns as other table columns in nimble and uses the existing `TableColumnTextCellViewBase` and `TableColumnTextGroupHeaderViewBase`.

More interesting/complex implementation will come in future PRs.

## 🧪 Testing

- New unit tests
- New storybook page (manually tested)
- New matrix tests

## ✅ Checklist


packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -0,0 +1,406 @@
...
+        }
+    });
+
+    describe('with integer formatting', () => {
+        beforeEach(async () => {
+            columnInstances.column1.format = NumberTextFormat.integer;
+            await waitForUpdatesAsync();
+        });
+
+        const testCases: readonly NumericTestCase[] = [
```
msmithNI, line 365: We should probably have the similar (round-down and round-up) test cases for negative numbers too. Intl.NumberFormat` has configuration for that (`roundingMode`), though the default behavior looks fine.

msmithNI (approved):

packages/nimble-components/src/table-column/number-text/models/format-helper.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (absoluteValue > 0 && (absoluteValue < 1e-6 || absoluteValue >= 1e16)) {
+        // Intl.NumberFormat formats numbers like '1.23E4' and '1.23E-4' for scientific. We want the output to be closer
+        // to Number.toString(), i.e. lowercase 'e' and including the exponent + sign ('1.23e+4' and '1.23e-4').
+        const exponentSeparator = number > 1 ? 'e+' : 'e';
+        return scientificFormatter
+            .formatToParts(number)
+            .map(s => (s.type === 'exponentSeparator' ? exponentSeparator : s.value))
+            .join('');
+    }
+    return formatter.format(number);
```
rajsite, line 41: Where is this requirement coming from?

If it's to aid in testing or something then we should capture the logic in a page object instead. Similar to the [date formatter discussion](https://github.com/ni/nimble/pull/1363#discussion_r1268471402).

Otherwise I'd like to understand more why we aren't taking the formatting output as-is.

My concern is that this is delegating to an external library based on locale. I don't know that `e` is a world-wide globally accepted representation of exponentiation. That's what we are delegating to the browser for in the first place. I don't think we should presume to override it. I also don't think there is real significant value in overriding it.

msmithNI: The [spec says 'formatting similarly to Number.toString()'](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md?plain=1#L174) which formats numbers for scientific notation differently than `Intl.NumberFormat` does by default. I also thought that the form with `e+` was easier to read (`1.22317846123e+21` vs `1.22317846123E21`, with the 2nd form the capital `E` almost blends into the numbers so it's harder to read to me / I prefer the 1st form) so I went with that for implementation.

It does complicate the code here though so it's worth discussion. I looked back at #887 and there wasn't really any discussion/examples for numbers with scientific notation there. We could always start with the default `Intl.NumberFormat` form and revisit if needed.

mollykreis:
Moving this PR back to draft to address some formatting changes discussed in person:
- Use 6 digits of precision
- Remove custom E -> e+ formatting
- Don't set `useGrouping: false`

I will also update the HLD to reflect these changes.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -0,0 +1,155 @@
...
+} from '../../base/tests/table-column-stories-utils';
+import { tableColumnTextTag } from '../../text';
+import { NumberTextFormat } from '../types';
+
+const simpleData = [
+    {
+        firstName: 'Homer',
+        lastName: 'Simpson',
+        age: 45.2358734623,
+        favoriteNumber: 0.0000002358967325
```
atmgrifter00, line 24: Pretty sure Homer's favorite number is π, because "Mmmmmm....piiiiii".
```suggestion
        favoriteNumber: π
```

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            sortOperation: TableColumnSortOperation.basic
+        };
+    }
+
+    private formatChanged(): void {
+        this.updateColumnConfig();
+    }
+
+    private updateColumnConfig(): void {
+        this.columnInternals.columnConfig = this.createFormatterConfig();
```
rajsite, line 52: How would you feel about encapsulating the Formatter behaviors?

Something like:
```ts

class NumberFormatter {
    abstract format (number: number);
}

class DefaultFormatter extends NumberFormatter {
    /* encapsulates all the logic around doing the auto formatting*/
}

class IntegerFormatter extends NumberFormatter {
   /* same, encapsulates the integer formatting stuff */
}

class CustomFormatter extends NumberFormatter {
    /* just pass the formatter reference */
    constructor(private formatter: Intl.NumberFormat) { }
}

And then we just pass the specific instance via columnConfig. Don't need to spread out the state for specific types of formatters on the columnConfig
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+            {
+                name: 'MAX_SAFE_INTEGER + 9999 renders as expected',
+                value: Number.MAX_SAFE_INTEGER + 9999,
+                expectedRenderedString: '9.0072E15'
+            },
+            {
+                name: 'MIN_SAFE_INTEGER - 9999 renders as expected',
+                value: Number.MIN_SAFE_INTEGER - 9999,
+                expectedRenderedString: '-9.0072E15'
+            }
```
rajsite, line 346: instead of "as expected" maybe "as an exponential with most significant digits of MAX_SAFE_INTEGER" or something

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+            {
+                name: 'MAX_SAFE_INTEGER + 9999 renders as expected',
+                value: Number.MAX_SAFE_INTEGER + 9999,
+                expectedRenderedString: '9,007,199,254,750,990'
+            },
+            {
+                name: 'MIN_SAFE_INTEGER - 9999 renders as expected',
+                value: Number.MIN_SAFE_INTEGER - 9999,
+                expectedRenderedString: '-9,007,199,254,750,990'
+            }
```
rajsite, line 449: Maybe split into four tests?

i.e. for the positive cases split to the two tests

```ts
            {
                name: 'MAX_SAFE_INTEGER +1 renders as the exact value',
                value: Number.MAX_SAFE_INTEGER + 1,
                expectedRenderedString: '9,007,199,254,740,992'
            },
            {
                name: 'MAX_SAFE_INTEGER + 1.5  renders with rounding precision errors',
                value: Number.MAX_SAFE_INTEGER + 1.5,
                expectedRenderedString: '9,007,199,254,750,992'
            },
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -0,0 +1,490 @@
...
+    beforeEach(async () => {
+        columnInstances = new ColumnInstances();
+        ({ element, connect, disconnect } = await setup(columnInstances));
+        pageObject = new TablePageObject<SimpleTableRecord>(element);
+    });
+
+    afterEach(async () => {
+        await disconnect();
+    });
+
```
rajsite, line 58: should have the tag name and construuction tests for this component as well

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const numberTextColumnDescription = 'The `nimble-table-column-number-text` column is used to display number fields as text in the `nimble-table`.';
+
+const formatDescription = `Configures the way that the numeric value is formatted to render within the column.
+
+<details>
+    <summary>Format Options</summary>
+
+    <ul>
+        <li>\`default\`: Integers are shown with no trailing zeros, the value is limited to 6 significant digits, and exponential notation is used for numbers that are large (\`>= 1e6\`) or small (\`< 1e-6\`) in magnitude.
```
rajsite, line 99:
```suggestion
        <li>\`default\`: Integers are shown with no trailing zeros, the value is limited to 6 significant digits, and exponential notation is used for numbers that are large (\`>= 1e6\`) or small (\`<= 1e-6\`) in magnitude.
```

rajsite (approved):

packages/nimble-components/src/table-column/number-text/models/integer-formatter.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The formatter for a number-text column whose format is configured to be 'integer'.
+ */
+export class IntegerFormatter {
+    private static readonly formatter = new Intl.NumberFormat(undefined, {
+        maximumFractionDigits: 0,
+        useGrouping: true
+    });
+
```
rajsite, line 9: Chatted with @atmgrifter00 a bit about the following examples:
image


The concerns were the following:

- Us ever showing negative zero in the table for our pre-made configs (default and integer) is not a great built-in behavior. Negative zero vs zero distinctions are really only relevant for specific engineering scenarios and shouldn't be the default representation.
- The name "integer" does not reflect what is happening well from a programmer configuration expectation. "Normal" / programmer expected float to integer conversion should not have negative zeros (because that can't be represented in the expected two's complement integer format) and should truncate instead of round.

I think the following changes would address those concerns:

- We should rename the `integer` format to `round`. Seems to better represent what is described by the docs and makes it clearer what will happen to your numbers.
- We should use the `signDisplay: "negative"` configuration for all these modes. i.e. the `round` mode and in both the formatters used in the `default` mode.

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/models/default-formatter.ts
```diff
@@ -0,0 +1,53 @@
...
+        undefined,
+        {
+            maximumSignificantDigits: DefaultFormatter.maximumDigits,
+            maximumFractionDigits: DefaultFormatter.maximumDigits - 1,
+            // Use 'lessPrecision' rounding priority to help ensure that no more than 6 digits are rendered
+            // in the formatted value, even if 'maximumSignificantDigits' or 'maximumFractionDigits' alone
+            // would result in more than 6 rendered digits.
+            roundingPriority: 'lessPrecision',
+            useGrouping: true,
+            // @ts-expect-error - The version of TypeScript currently being used does not include 'negative' as a valid signDisplay value
```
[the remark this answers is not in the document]
mollykreis: Update: It looks like the value of `"negative"` isn't supported in the version of FF our tests run against (115), despite the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility) claiming that it is. The value is, however, supported in the version of FF that is installed on my machine (116). I'm working on finding a good alternative.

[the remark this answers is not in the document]
mollykreis: I ended up writing code to coerce `-0` to `0` before calling format since I couldn't use `signDisplay: 'negative'`.

I also realized that `roundingPriority` isn't supported in all browsers either, so I had to add an additional condition when using the `default` format for when to use 6 significant digits, 5 decimal digits, or exponential notation with 6 significant digits.

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -0,0 +1,553 @@
...
+                expectedRenderedString: '1E6'
+            },
+            {
+                name: 'does not use exponential notation for 0.001',
+                value: 0.001,
+                expectedRenderedString: '0.001'
+            },
+            {
+                name: 'does not use exponential notation for 0.000995',
+                value: 0.000999999,
```
rajsite, line 362: value doesn't match test name

rajsite (approved):

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text-matrix.stories.ts
```diff
@@ -0,0 +1,70 @@
+import type { StoryFn, Meta } from '@storybook/html';
```
rajsite, line 1: I think we need a usage note for round-to-integer stating that our numbers are represented as treated as 64-bit floats so notably cannot represent large 64-bit integers losslessly. And maybe linking to the mdn MAX_SAFE_INTEGER docs.

We'll need to add an mdx page so don't feel strongly if in this PR or follow-up.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -171,12 +171,12 @@ Nimble will introduce `nimble-table-column-number-text` which formats a numeric
...
     -   This could be extended to other pre-configured formats in future. Their configuration attributes would similarly be prefixed with the name of the format mode.
-    -   **Note:** all of the above will be implemented using a `Intl.NumberFormat` formatter. For all modes besides `custom` Nimble will configure the formatter with defaults to match the [visual design spec](https://github.com/ni/nimble/issues/887) (e.g. `useGrouping: false` to achieve `1000` rather than `1,000` and `signDisplay: auto` to achieve `1` rather than `+1`).
+    -   **Note:** all of the above will be implemented using a `Intl.NumberFormat` formatter. For all modes besides `custom` Nimble will configure the formatter with defaults to match the [visual design spec](https://github.com/ni/nimble/issues/887). The exception is that we will set `useGrouping: true` to achieve `1,000` rather than `1000` because this styles the values in a way that is more human readable.
```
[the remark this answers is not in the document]
mollykreis: There was a bit of discussion around this, but it was just with the dev team. Something things we talked about were:
- commas will likely make the value more readable
- SLE doesn't have commas today, but SLS had commas. We don't think the change in behavior was intentional.

I agree that it would be good to extend the discussion to include the designers.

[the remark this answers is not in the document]
jattasNI: Those are great points in favor of commas. I'm fine either way, just want to make sure we have some authority/rationale to point to when someone inevitably complains.

If designers aren't able to give guidance immediately then I'm ok proceeding with commas for now and revisiting this when we get firmer guidance. One idea though: why not use `useGrouping: auto` instead of `true`? The result is the same for `en-us` but the MDN docs imply it might make different decisions in other locales where grouping is less expected.

jattasNI (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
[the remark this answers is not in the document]
mollykreis: Based on an offline discussion with the dev team and designers, we decided to render values with commas.

Regarding using `useGrouping: auto`, that [isn't supported in FF yet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility). In FF, values other than `boolean | undefined` are in preview mode.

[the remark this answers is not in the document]
rajsite: Seems like once widely available in ffx (release + 6 months) we should switch to that as a better behavior and clean-up some of the workarounds for other missing options, could that be captured in a tech debt?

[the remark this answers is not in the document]
mollykreis: Created #1442

PR #1415, opened 2023-08-02
Hierarchical Cascade Layers with @layer Guidelines

aidendk:
# Pull Request

## 🤨 Rationale

Provides guidelines for css styling used in #1355 

## 👩‍💻 Implementation

Added docs and examples to css guidelines for using `@layer`

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/docs/css-guidelines.md
```diff
@@ -126,6 +126,92 @@ For example:
...
+    }
+}
+```
+In this example, using @layer gives the `content:hover` non-host class precedence to apply to all `:host` classes in the `hover` state.
+
+States should flow from plain base -> hover -> focus -> active -> error -> disabled -> top (which overrides all others)
+
+Note: Custom layer names should be avoided.
+
+Corresponding states will be placed in layers with each other, in the order specified by [group selectors by target and document order](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order).
```
m-akinc, line 183:
```suggestion
Selectors should be organized by layer, then within a layer they should be ordered by [target and document order](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order).
```

packages/nimble-components/docs/css-guidelines.md
```diff
@@ -126,6 +126,92 @@ For example:
...
+
+@layer hover {
+    .content:hover {
+        border-color: ${borderHoverColor};
+    }
+}
+```
+In this example, using @layer gives the `content:hover` non-host class precedence to apply to all `:host` classes in the `hover` state.
+
+States should flow from plain base -> hover -> focus -> active -> error -> disabled -> top (which overrides all others)
```
m-akinc, line 179: I think it's worth mentioning (after this line?) that selectors that are not in a layer take precedence over all layers, so if you use layers, you should put all selectors in a layer.

m-akinc (approved):

packages/nimble-components/docs/css-guidelines.md
```diff
@@ -126,6 +126,95 @@ For example:
 
 Useful reference: [When do the :hover, :focus, and :active pseudo-classes apply?](https://bitsofco.de/when-do-the-hover-focus-and-active-pseudo-classes-apply/)
 
+## Hierarchical Cascade Layers with @layer
```
rajsite, line 129: Some high-level thoughts:

- Since these are guidelines, the title should be a clear action
- I think the advice is a bit too hard to evaluate. I think we can simplify considerably and just say new control styles should be organized this way.
- This feels directly related to the previous topic and is describing implementation details for that topic, I think we can merge it there as a subheader.
- The example should be simplified and made standalone. I'd avoid referencing source code directly if it's not needed. Think folks can will try and search for examples in source if they want it, we don't need to maintain that link.

So at a high-level I think the docs should be organized like this (the following is outlining and should be word smithed):

```
### Organize styles with CSS Cascade Layers

New controls and existing controls undergoing major refactoring should migrate to using [CSS cascade layers](link to mdn) for organizing styles that override states.
Cascade layers help enforce that property overrides have higher specificity over the base styles without modifying selectors to increase specificity.

For consistency, control styles that adopt CSS Cascade Layers should follow these practices:
- Specify the following layers in: `@layer base, hover, focusVisible, active, disabled, top`
   - Avoid changing the layer order of adding additional layers.
- Ensure all styles are placed in a layer and no styles for the control are outside a layer.
- Ensure the layers in the stylesheet are ordered based on the layer order.
- Ensure the styles within a layer follow existing best practices such as grouping using document order.

Example:
` ` `css
@layer base, hover, focusVisible, active, disabled, top

@layer base {
    :host {
        border: green;
   }
}
@layer hover {}
@layer focusVisible {}
@layer active {}
@layer disabled {
    :host([disabled]) {
        border: gray;
   }
}
@layer top {}
` ` `
```

packages/nimble-components/docs/css-guidelines.md
```diff
@@ -126,6 +126,37 @@ For example:
 
 Useful reference: [When do the :hover, :focus, and :active pseudo-classes apply?](https://bitsofco.de/when-do-the-hover-focus-and-active-pseudo-classes-apply/)
 
+### Organize styles with CSS Cascade Layers 
+
+New controls and existing controls undergoing major refactoring should migrate to [CSS Cascade Layers with @layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) for organizing styles that override states. Cascade Layers with @layer define the order of precedence when multiple cascade layers are present, and help enforce that property overrides have higher specificity over base styles without modifying selectors to increase specificity.
```
rajsite, line 131:
```suggestion
New controls and existing controls undergoing major refactoring should migrate to [CSS Cascade Layers with `@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) for organizing styles that override states. Cascade Layers with `@layer` define the order of precedence when multiple cascade layers are present. This helps enforce that property overrides have higher specificity over base styles without needing to modify selectors to increase specificity.
```

packages/nimble-components/docs/css-guidelines.md
```diff
@@ -126,6 +126,37 @@ For example:
...
+
+For consistency, control styles using CSS Cascade Layers should follow these practices:
+-   Define the cascade layers: `@layer base, hover, focusVisible, active, disabled, top`
+    -   Avoid changing the order of layers, changing layer names, or creating additional layers.
+-   Layers in the stylesheet should follow the order by which they are defined above.
+-   Ensure that all styles are in a layer- no styles for the control should be outside of a layer.
```
rajsite, line 137: I think that's how semicolons work 😅
```suggestion
-   Ensure that all styles are in a layer; no styles for the control should be outside of a layer.
```

rajsite (approved):

rajsite:
We can merge these changes. @jattasNI worth reviewing if we should update recommendations / being aware of for future code reviews

Issue #1211, opened 2023-04-28
Create mixins for Angular table column directives 


m-akinc:
## 🧹 Tech Debt
Multiple column types are likely to mix-in the FractionalWidthColumnAPI and/or the GroupableColumnAPI. The Angular directives for those column types will also expose those APIs. We should have mixins for those directives. Once we have them, we should update the Text and Anchor column types to use them.

rajsite:
Re-opening as #1587 was reverted in #1591
An [Angular issue](https://github.com/angular/angular/issues/42594#issuecomment-868091258) describes why using TypeScript Mixins does not work for Angular Directives and recommends checking out how Angular Material handles mixins. I didn't look but we can check what Angular Material does and see if it's useful for us.

mollykreis:
tldr - After looking into the approach that Angular Material uses, it doesn't see particularly useful for us. Closing this as not planned because there isn't a clear path forward to is significantly better than our current approach of copy/pasting between directives.

The approach taken in Angular Material is to mixin a base class that isn't a `@Directive` and doesn't contain any properties marked as `@Input`. Instead, that mixed in base class provides only the TS implementation of the getter(s) and setter(s) for properties. Then, the actual `@Directive` has to configure an `inputs` array as part of its decorator, which in our case also has to rename each input to follow our naming conventions. Therefore, our implementation would need to look something like the following:

```ts
@Directive({
    selector: 'nimble-table-column-text',
    // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property, @angular-eslint/no-input-rename
    inputs: ['fractionalWidth: fractional-width', 'minPixelWidth: min-pixel-width', 'groupIndex: group-index', 'groupingDisabled: grouping-disabled']
})
export class NimbleTableColumnTextDirective extends mixinFractionalWidthColumnAPI(mixinGroupableColumnAPI(NimbleTableColumnBaseDirective<TableColumnText>)) {
    /* implementation specific to the text column */
}
```

It's not ideal to have to call the mixin function and still have to know all the properties & their mapped inputs. This also doesn't allow a mixin to add a new input without having to update each usage of that mixin. I attempted to single-source the input string within the mixin file, but that didn't work because the mixin is in a different entry point (`table-column` rather than `table-column/text`); therefore, the string array isn't statically known at compile time like is required.

[...]

Issue #572, opened 2022-05-18
Research FastElement 2.0


mollykreis:
## 🧹 Tech Debt
We should updated to FastElement 2.0 so that we can use some new features from FAST.

Features we should leverage when we move to FastElement 2.0:
- Advanced event binding to specify things such as _capture: true_ as is done in the nimble-menu-button.
    - FAST comment: https://github.com/microsoft/fast/issues/4531#issuecomment-963323586
    - nimble change that could leverage this feature: https://github.com/ni/nimble/pull/569
- Better typing of `c.parent` in bindings
    - See discussion from nimble PR: https://github.com/ni/nimble/pull/1056#discussion_r1107806254
    - See FAST change that improved typing: https://github.com/microsoft/fast/pull/6397

rajsite:
When switching to FE2 we should update our exported button type:
https://github.com/ni/nimble/blob/f3f0e75635bae19ca3b751b69d0479387cba5bab/packages/nimble-components/src/button/types.ts#L9

To export the newly added actual [foundation ButtonType](https://github.com/microsoft/fast/blob/a42ff75a9130a5ce13006c272c2e2bbbaffdbb41/packages/web-components/fast-foundation/src/button/button.options.ts#L6).

jattasNI:
In #1134 we added an import in `nimble-components/src/menu/index.ts` of the anchored region component so that FAST's implementation could find Nimble's anchored region when using `tagFor`. Both `tagFor` and anchored region have seen big changes in FAST vNext, so this workaround should be revisited and possibly removed.

m-akinc:
We should look at replacing the boolean converter we use in the date-text table column with the FAST `nullableBooleanConverter`.

jattasNI:
Given the slow progress on FastElement 2.0 it doesn't seem worthwhile to proactively track potential issues. Closing as low ROI for now, we can re-open if it comes closer to release.

PR #1417, opened 2023-08-03
Angular integration for date-text table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1014

## 👩‍💻 Implementation

nimble-angular
- directive for date-text column
- tests
- added column to example
- added `checkValidity()` and `validity` to `NimbleTableColumnBaseDirective`. These must have been previously overlooked.
- fixed two text column tests that were accidentally testing the table instead

nimble-components
- The date-text column's `customHour12` property needs to have three possible states: `true`, `false`, or `undefined`. It was being declared as a boolen-mode attribute, though, which was preventing it from being set to `undefined` (it would be translated to `false`). I now instead configure a custom `optionalBooleanConverter` to handle conversion of values to/from the view.

## 🧪 Testing

Wrote tests and they pass.

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/table-column/nimble-table-column-base.directive.ts
```diff
@@ -72,4 +72,12 @@ export class NimbleTableColumnBaseDirective<T extends TableColumn> {
     }
 
     public constructor(protected readonly renderer: Renderer2, protected readonly elementRef: ElementRef<T>) {}
+
+    public checkValidity(): boolean {
+        return this.elementRef.nativeElement.checkValidity();
+    }
+
+    public get validity(): TableColumnValidity {
```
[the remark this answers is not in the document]
m-akinc: I looked briefly at it, trying to see what it would take, but I didn't get anywhere. I have a feeling that would be at odds with the implementation we're currently using (based on `src/utilities/models/tracker.ts`). I'm also not sure how likely it is that client code would be accessing individual flags in that object, vs dumping the object to the console for debugging, for example. What do you think?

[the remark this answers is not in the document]
mollykreis: I was wondering the same thing about how likely it is that someone would actually write code that uses the output of that function vs using it in debugging. I mostly asked because depending on how we went about it, it might be easier to implement now rather than after we have a bunch of column types. A few ideas that immediately come to mind are:
- Export the Validity type for each column. This isn't a great experience, but it would at least allow someone to do something like `(myDateColumn.validity as DateTextColumnValidity).invalidCustomFormatters`
- Make `get validity()` abstract on the base directive so that each column has to implement it to return their specific validity type. I'm not 100% sure if TypeScript would make that typing easy or if we'd have to jump through any hoops to get that to work.

I don't have a problem waiting until we have a better use-case for needing improved typing before we come up with a plan. I just wanted to make sure we were explicitly making the decision not to do anything now.

angular-workspace/projects/ni/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
```diff
@@ -0,0 +1,301 @@
...
+        return this.elementRef.nativeElement.customHourCycle;
+    }
+
+    // Renaming because property should have camel casing, but attribute should not
+    // eslint-disable-next-line @angular-eslint/no-input-rename
+    @Input('custom-hour-cycle') public set customHourCycle(value: HourCycleFormat) {
+        this.renderer.setProperty(this.elementRef.nativeElement, 'customHourCycle', value);
+    }
+
+    public get fractionalWidth(): number | null | undefined {
```
atmgrifter00, line 258: Man...we need to get those [mixins](https://github.com/ni/nimble/issues/1211) created.

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -82,7 +83,7 @@ export class TableColumnDateText extends TableColumnTextBase {
     @attr({ attribute: 'custom-format-matcher' })
     public customFormatMatcher: FormatMatcherAlgorithm;
 
-    @attr({ attribute: 'custom-hour12', mode: 'boolean' })
+    @attr({ attribute: 'custom-hour12', converter: optionalBooleanConverter })
```
atmgrifter00, line 86: @rajsite, this is being done to support three different functionalities for this property, where true = 12 hour, false = 24 hour, and undefined = locale dependent. Would it be more conventional to represent this API as an enum for Nimble, and then just translate it appropriately under the hood, or do we feel we should align our API directly to that of `Intl.NumberFormat`?

mollykreis: I like the suggestion of creating an enum in nimble for this. I think a nullable boolean will likely cause problems (or at the least confusion) in our API since it does diverge from what is standardly done.

angular-workspace/projects/ni/nimble-angular/table-column/nimble-table-column-base.directive.ts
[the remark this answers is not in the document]
m-akinc: I would vote for giving it some run time to see if there's a need for improved typing. I have a feeling we won't need it.

angular-workspace/projects/ni/nimble-angular/table-column/date-text/tests/nimble-table-column-date-text.directive.spec.ts
```diff
@@ -0,0 +1,1671 @@
...
+            public customEra: EraFormat = 'long';
+            public customYear: YearFormat = 'numeric';
+            public customMonth: MonthFormat = 'short';
+            public customDay: DayFormat = '2-digit';
+            public customHour: HourFormat = 'numeric';
+            public customMinute: MinuteFormat = '2-digit';
+            public customSecond: SecondFormat = 'numeric';
+            public customTimeZoneName: TimeZoneNameFormat = 'shortOffset';
+            public customFormatMatcher: FormatMatcherAlgorithm = 'basic';
+            public customHour12?: boolean = false;
```
mollykreis, line 365: Should this be of type `Hour12Format` rather than optional boolean?

angular-workspace/projects/ni/nimble-angular/table-column/date-text/tests/nimble-table-column-date-text.directive.spec.ts
```diff
@@ -0,0 +1,1671 @@
...
```
mollykreis, line 1052: Should this be of type `Hour12Format` rather than optional boolean?

packages/nimble-components/src/table-column/date-text/index.ts
mollykreis: I'm not sure about the new enum. Having the values actually be `true`, `false`, and `undefined` within the enum and still requiring a custom converter doesn't seem correct to me. I think if we're going to make an enum for this, it should be an enum following our existing conventions where the enum values map to a string (or `undefined`), and then there isn't a custom converter that is required.

atmgrifter00: +1.

m-akinc: Changed enum to have string values and made resulting updates.

packages/nimble-components/src/table-column/date-text/types.ts
```diff
@@ -25,3 +25,10 @@ export type DayPeriodFormat = Intl.DateTimeFormatOptions['dayPeriod'];
 export type DateStyle = Intl.DateTimeFormatOptions['dateStyle'];
 export type TimeStyle = Intl.DateTimeFormatOptions['timeStyle'];
 export type HourCycleFormat = Intl.DateTimeFormatOptions['hourCycle'];
+
+export const Hour12Format = {
+    default: undefined,
+    twelveHour: 'twelve-hour',
```
atmgrifter00, line 31: Since these are just string values we don't need the kebab-casing. Additionally, we might consider making the values even easier for a user to parse (e.g. '12hour', and '24hour').

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
```diff
@@ -265,7 +266,9 @@ export const dateTextColumn: StoryObj<TextColumnTableArgs> = {
         customHour12: {
             name: 'custom-hour12',
             description:
-                'Refer to the option [documentation for the `Intl.DateTimeFormat()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat).'
+                'Refer to the option [documentation for the `Intl.DateTimeFormat()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat).',
```
atmgrifter00, line 269: Do we still want to just re-direct users to this documentation, as our API is now misaligned?

packages/nimble-components/src/table-column/date-text/types.ts
mollykreis: I agree that kebab-casing is not what we typically do, so I don't think that should be our choice here. I'm on the fence for _12_ vs _twelve_ though. We don't generally use numbers, but our API already is using `cutom-hour12`.

atmgrifter00: I actually made the suggestion more to avoid the more verbose 'twenty-four`, where I think '24' is just _way_ easier to parse for a user.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.stories.ts
m-akinc: I've added a note about how our values map to the ones in the linked documentation.

packages/nimble-components/src/table-column/date-text/types.ts
m-akinc: I hadn't seen Molly's response when I update the values to '12Hour` and '24Hour`. Now I'm noticing that that is not the capitalization you suggested. I was just using camel casing. Any objections to this?

FYI, the reason I used kebab-cased values in the first place was because I noticed our use of `AttributeHelpers.GetEnumNamesAsKebabCaseValues` in nimble-blazor. It's weird that we have (and use) that helper if our convention for values is not to use kebab casing.

atmgrifter00: I believe I was incorrect in my statement about [kebab-casing](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#attribute-naming-convention). I think I thought what I did because I came across some internal APIs that didn't use the kebab-casing, and moreover, there are very few cases where we have enum values that would require it. Given that, I think it would be appropriate for the values to be `12-hour` and `24-hour` (convention is for it to be lower-case as well).

m-akinc: After discussion, we decided that we should be aligning as closely as possible with the JS API, meaning having a tri-state boolean rather than a new enum.

atmgrifter00 (approved):

angular-workspace/projects/ni/nimble-angular/table-column/date-text/tests/nimble-table-column-date-text.directive.spec.ts
```diff
@@ -0,0 +1,1671 @@
...
+                        custom-era="long"
+                        custom-year="numeric"
+                        custom-month="short"
+                        custom-day="2-digit"
+                        custom-hour="numeric"
+                        custom-minute="2-digit"
+                        custom-second="numeric"
+                        custom-time-zone-name="shortOffset"
+                        custom-format-matcher="basic"
+                        custom-hour12="24hour"
```
mollykreis, line 110: Why is `custom-hour12` being set to `24hour`? That isn't a valid value, right? And the test is asserting this gets coerced to `false`. I'm not against having a few extra tests for the conversion of this property -- in fact that would probably be a good thing. But, I think this test should use valid values.

Right now the property is defined as
```
@Input('custom-hour12') public set customHour12(value: boolean | string | undefined)
```
Should that be stricter on which strings are allowed?
```
@Input('custom-hour12') public set customHour12(value: boolean | 'true' | 'false' | undefined)
```

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -498,6 +477,38 @@ describe('TableColumnDateText', () => {
...
+                            custom-format-matcher="basic"
+                            custom-hour12="false"
+                            custom-time-zone="America/Chicago"
+                            custom-calendar="hebrew"
+                            custom-day-period="short"
+                            custom-numbering-system="fullwide"
+                            custom-date-style="long"
+                            custom-time-style="long"
+                            custom-hour-cycle="h23"
+                            >
```
mollykreis, line 505: I think this should be out-dented one level.
```suggestion
                        >
```

packages/nimble-components/src/utilities/models/converter.ts
```diff
@@ -0,0 +1,17 @@
+import { booleanConverter, ValueConverter } from '@microsoft/fast-element';
+
+export const optionalBooleanConverter: ValueConverter = {
```
mollykreis, line 3: FAST actually has a nullableBooleanConverter, and the implementation looks very similar to this. Can you double check that the existing one in FAST isn't sufficient for this case?

https://github.com/microsoft/fast/blob/218cfc014f5c76464493725f83bd883e8b07defa/packages/web-components/fast-element/src/components/attributes.ts#L96

angular-workspace/projects/ni/nimble-angular/table-column/date-text/tests/nimble-table-column-date-text.directive.spec.ts
m-akinc: > Why is `custom-hour12` being set to `24hour`?

That was a mistake.

> Should that be stricter on which strings are allowed?

Yeah, that makes sense to me.

packages/nimble-components/src/utilities/models/converter.ts
m-akinc: You've linked to the `nullableNumberConverter`. I do see that there is a `nullableBooleanConverter` in the main branch, but I don't see it in the branch that we use. Anyway, theirs wouldn't work for us because we need setting to `undefined` not to convert to `null`.

mollykreis: Why is the distinction between `null` and `undefined` important here? The value given to the formatter will ultimately be `true`, `false`, or `undefined`, right?

m-akinc: Actually, I guess you're right. We pass `this.customHour12 ?? undefined`, so even if the property gets a `null` value, we'll pass `undefined` to the formatter. Still, since there is no `nullableBooleanConverter` in the version of FAST that we use, we have to define our own, and it seems reasonable (to me) to have it pass `null` and `undefined` through unchanged. Is there an argument for doing it a different way?

mollykreis: Will the `nullableBooleanConverter` be available when we complete #1427? If so, we should have a note somewhere that this converter should be replaced with that one once it is available.

m-akinc: It still isn't available. I've added a comment reminding us to adopt the FAST converter when it's available.

mollykreis (approved):

packages/nimble-components/src/table-column/date-text/index.ts
```diff
@@ -82,7 +83,9 @@ export class TableColumnDateText extends TableColumnTextBase {
     @attr({ attribute: 'custom-format-matcher' })
     public customFormatMatcher: FormatMatcherAlgorithm;
 
-    @attr({ attribute: 'custom-hour12', mode: 'boolean' })
+    // Later versions of FAST (than the legacy branch we're on) have a nullableBooleanConverter.
```
jattasNI, line 86: Could you link this code comment to #572 and add a note in that issue mentioning this code? That might help increase the odds of us fixing this when we upgrade.

packages/nimble-components/src/utilities/models/converter.ts
jattasNI, line 3: Could you add a comment explaining when this should be used?

jattasNI (approved):

packages/nimble-components/src/utilities/models/converter.ts
```diff
@@ -0,0 +1,17 @@
...
+    toView(value: unknown): string {
+        return typeof value === 'boolean' ? value.toString() : '';
```
rajsite, line 5: empty string isn't a valid option right? won't this result in `<my-el custom-hour12>` which is none of absent, "true", or "false"

```suggestion
        return typeof value === 'boolean' ? value.toString() : null;
```
Is there a testing gap?

packages/nimble-components/src/utilities/models/converter.ts
```diff
@@ -0,0 +1,17 @@
...
+    },
+
+    fromView(value: string): boolean | null | undefined {
+        if (value === undefined) {
+            return undefined;
+        }
+        if (value === null) {
+            return null;
+        }
```
rajsite, line 14: We are trying to advocate that it is unexpected behavior that FAST lets null values through, lets not propagate that behavior

```suggestion
        if (value === undefined || value === null) {
            return undefined;
        }
```

packages/nimble-components/src/utilities/models/converter.ts
```diff
@@ -0,0 +1,17 @@
...
+        return booleanConverter.fromView(value) as boolean;
```
rajsite, line 15: Seems weird to not be symmetric if we are leveraging fast converters.

And the [fast converter](https://github.com/microsoft/fast/blob/c2b4076a8d1d527551b3c6680c17ba6fd030c147/packages/web-components/fast-element/src/components/attributes.ts#L77) is doing weird cases we don't care about like the value number `0` and falling back to true for misconfigurations. true seems like the wrong fallback behavior over absent.

FAST's reasoning is that any value for an attribute should be true `<my-el val>`, `<my-el val="">`, `<my-el val="boo!">`, `<my-el val="false">`. They are putting in exceptions to be friendly I guess where they are going out of spec to treat `<my-el val="false">` as a special case for false. Irregardless we have already stated normal boolean behavior is not what we expect so relying on this is wrong.

Lets handle our expected cases with a more appropriate fallback. `'true'`, `'false'`, `true`, `false` and fallback to `undefined` instead of `true`.

packages/nimble-components/src/utilities/models/converter.ts
m-akinc: I've updated our test to fail with the current code and pass with the change.

packages/nimble-components/src/utilities/models/converter.ts
rajsite: Unresolving as the fromView does not handle the cases give, see:

> Lets handle our expected cases with a more appropriate fallback. 'true', 'false', true, false and fallback to undefined instead of true.

missing the boolean `true` and boolean `false` cases.

rajsite: The requirement that fromView handles the possible property values (the actual booleans) comes from my understanding of this topic: https://www.fast.design/docs/fast-element/defining-elements#customizing-attributes

Specifically:

> When the property value changes, the converter's fromView method will also be called, ensuring that the type is correct. After this, the mode will be determined. If the mode is set to reflect then the converter's toView method will be called to allow the type to be formatted before writing to the attribute using setAttribute.

Which seems to indicate fromView needs to handle property values and attribute string values.

rajsite (approved):

angular-workspace/projects/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -107,6 +108,7 @@ export class CustomAppComponent {
             stringValue2: `bar ${tableData.length}`,
             href: '/customapp',
             linkLabel: 'Link',
+            date: new Date().valueOf(),
```
rajsite, line 111: Use deterministic values in demos (like we do in stroybook)

Issue #1013, opened 2023-01-30
Table allows configuration of icon column type


nate-ni:
## 💁 Proposed Solution
- [x] #887

Open questions:

1. which column width modes should it support? Just fixed pixel width?

PR #1418, opened 2023-08-03
Icon table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1013

## 👩‍💻 Implementation

Created new `nimble-table-column-icon` component plus the supporting elements `nimble-mapping-icon` and `nimble-mapping-spinner`.

Angular and Blazor support will be implemented in follow-on PRs.

## 🧪 Testing

Unit tests written for all new components. Tested in Storybook.

## ✅ Checklist


packages/nimble-components/src/mapping/icon/index.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    // Allow icons to be defined asynchronously from when the property is configured
+    private async resolveIconAsync(icon: string): Promise<void> {
+        try {
+            await customElements.whenDefined(icon);
+        } catch (ex) {
+            // If any error (i.e. invalid custom element name) don't continue
+            // Don't update the resolvedIcon as it was already set to undefined before async resolution
+            // (in-case other async resolutions were started)
+            return;
```
atmgrifter00, line 44: Is there validity state we would want to update here?

packages/nimble-components/src/mapping/icon/index.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        this.resolvedIcon = isIconClass(elementClass) ? icon : undefined;
+    }
+
+    private iconChanged(): void {
+        const icon = this.icon;
+        if (typeof icon !== 'string') {
```
atmgrifter00, line 60: nit: Is this just another way of saying:
```ts
if (icon === undefined) { ... }
```
If so, I think the `undefined` check is clearer in intent, since icon is always either a string or `undefined`. The `typeof` check, to me, is somewhat suggestive that there could be other possible values for `icon` outside of `undefined` and string values (like `null` even), which isn't the case.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -363,6 +366,49 @@ describe('TableColumnEnumText', () => {
             }
         });
 
+        describe('invalid mappings', () => {
```
atmgrifter00, line 369: I don't think this `describe` block is helpful from an organizational perspective. The test would read "validation invalid mappings is invalid with icon or spinner mappings", as opposed to "validation is invalid with icon or spinner mappings". Seems like we can just discard this inner `describe` block, and just move all of the helper functions/types as you see fit (which may be nowhere).

packages/nimble-components/src/table-column/icon/cell-view/template.ts
```diff
@@ -0,0 +1,19 @@
+import { html, when } from '@microsoft/fast-element';
+import type { TableColumnIconCellView } from '.';
+import { spinnerTag } from '../../../spinner';
+
+export const template = html<TableColumnIconCellView>`
+    ${when(
```
atmgrifter00, line 6: Could you do:
```
   ${when(x => x.iconTemplate, html`
        {x => x.iconTemplate}
    `)}
    ${when(x => !x.iconTemplate,
        html<TableColumnIconCellView>`
        <${spinnerTag}
            title="${x => x.text}"
            aria-label="${x => x.text}">
        </${spinnerTag}>
    `
    )}
```

Basically just trying to get rid of the redundant `visual` observable (this would also require making `iconTemplate` an `observable`).

packages/nimble-components/src/table-column/icon/group-header-view/styles.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const styles = css`
+    ${display('inline-flex')}
+
+    :host {
+        gap: ${smallPadding};
+        align-items: center;
+    }
+
+    .no-shrink {
```
atmgrifter00, line 17: I don't see this being used. Remove?

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    for (const test of [
+        { type: 'string', key: 'a' },
+        { type: 'number', key: 10 },
+        { type: 'boolean', key: true }
+    ]) {
+        // eslint-disable-next-line @typescript-eslint/no-loop-func
+        it(`displays icon mapped from ${test.type}`, async () => {
+            ({ element, connect, disconnect, model } = await setup(
+                [{ key: test.key, text: 'alpha', icon: 'nimble-icon-xmark' }],
```
atmgrifter00, line 75: I think it might be good to just go ahead and use the exported tag.

atmgrifter00, line 75:
```suggestion
                [{ key: test.key, text: 'alpha', icon: `${iconXmarkTag}` }],
```

I think it would be good to go ahead and use the exported tag for the icon.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -186,6 +188,30 @@ export class TablePageObject<T extends TableRecord> {
...
     }
 
+    public getRenderedCellIcon(
+        rowIndex: number,
+        columnIndex: number
+    ): Icon | null {
+        const icon = this.getRenderedCellView(rowIndex, columnIndex).shadowRoot!
+            .firstElementChild;
+        if (!icon || !(icon instanceof Icon)) {
+            return null;
```
atmgrifter00, line 198: Shouldn't we throw an error here? We do for other `getRenderedCell<>` variants (like `getRenderedCellAnchor`)?

atmgrifter00:
So, is it expected that the spinner doesn't animate in the column and group header views? I would expect that it would animate.

packages/nimble-components/src/mapping/icon/index.ts
m-akinc: No, that is handled by the `TableColumnIconValidator`, which sets a validity flag when the resolved icon is `undefined`.

packages/nimble-components/src/table-column/icon/group-header-view/styles.ts
m-akinc: Somewhere along the way, some code got accidentally removed. I updated the templates for the icons and spinners to include the class `no-shrink`. I also updated the templates to put label text inside a `span` that conditionally sets the `title`. Thanks for catching this.

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: Sure, I just need to fix up the test to expect an error instead of `null`.

m-akinc:
> So, is it expected that the spinner doesn't animate in the column and group header views? I would expect that it would animate.

I was unconditionally preventing the animation to avoid flaky Chromatic snapshots, but I'm not sure why I didn't do it conditionally like the spinner stories. Made it conditional. Presumably it's going to work fine.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
m-akinc: Did this for the whole file.

packages/nimble-components/src/mapping/icon/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    public resolvedIcon?: string;
+
+    // Allow icons to be defined asynchronously from when the property is configured
+    private async resolveIconAsync(icon: string): Promise<void> {
+        try {
+            await customElements.whenDefined(icon);
+        } catch (ex) {
+            // If any error (i.e. invalid custom element name) don't continue
+            // Don't update the resolvedIcon as it was already set to undefined before async resolution
+            // (in-case other async resolutions were started)
```
atmgrifter00, line 43:
```suggestion
            // (in case other async resolutions were started)
```

packages/nimble-components/src/mapping/icon/index.ts
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+
+        const elementClass = customElements.get(icon)!;
+        if (icon !== this.icon) {
+            // Possible the icon has changed while waiting for async resolution
+            // Don't update the resolvedIcon as it was already set to undefined before async resolution
+            // (in-case other async resolutions were started)
```
atmgrifter00, line 51:
```suggestion
            // (in case other async resolutions were started)
```

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -0,0 +1,398 @@
...
+            ));
+            await connect();
+            await waitForUpdatesAsync();
+            expect(model.col1.checkValidity()).toBeFalse();
+            expect(
+                model.col1.validity.invalidMappingKeyValueForType
+            ).toBeTrue();
+        });
+
+        describe('invalid mappings', () => {
```
atmgrifter00, line 310: Again another case where the describe block offers strange chain semantics of "TableColumnIcon validation invalid mappings is invalid with text mapping" instead of "TableColumnIcon validation invalid with text mapping".

atmgrifter00 (approved):

packages/nimble-components/src/table-column/icon/group-header-view/styles.ts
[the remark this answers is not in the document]
m-akinc: We apply it to both icons and spinners in our templates. I kind of like that the name documents the purpose for having the class, rather than putting a seemingly redundant `icon-content` class on our icons and spinners. Since you don't feel strongly, I'd prefer to leave it.

packages/nimble-components/src/table-column/icon/cell-view/template.ts
m-akinc: Realized I can't do this. There is a third state where the value doesn't match any mappings, and we don't show either template. Switched back to using the `visual` property.

change/@ni-nimble-components-6bfc2c11-64d5-45f8-85b7-85d6afbf3d5f.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @mollykreis The PR from a very brief scroll through is looking pretty solid. I'm likely not going to get to a full review for a few days due to NI-Tech stuff, would you be okay doing a secondary buddy review to get any feedback you could find sooner?

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+        super.validate(mappings, keyType);
+        this.validateIconNames(mappings);
+    }
+
+    private validateIconNames(mappings: Mapping[]): void {
+        const isMappingIcon = (mapping: Mapping): mapping is MappingIcon => mapping instanceof MappingIcon;
+        const invalid = mappings
+            .filter(isMappingIcon)
+            .map(mappingIcon => mappingIcon.resolvedIcon)
+            .some(resolvedIcon => resolvedIcon === undefined);
```
mollykreis, line 41: nit: I don't expect there to be too many mapping icons, but it seems unnecessary to convert the array of MappingIcon instances to a string array before looking at the value of those strings. Why not do this:

```suggestion
        const invalid = mappings
            .filter(isMappingIcon)
            .some(mappingIcon => mappingIcon.resolvedIcon === undefined);
```

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+}
+
+class Model {
+    public col1!: TableColumnIcon;
+}
+interface ModelFixture<T> extends Fixture<T> {
+    model: Model;
+}
+
+describe('TableColumnIcon', () => {
```
mollykreis, line 34: This should also include the _should export its tag_ and _can construct an element instance_ tests.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+
+        model.col1.fieldName = 'field2';
+        await waitForUpdatesAsync();
+
+        expect(
+            pageObject.getRenderedCellIcon(0, 0) instanceof IconCheck
+        ).toBeTrue();
+    });
+
+    it('changing mapping icon updates display', async () => {
```
mollykreis, line 119: It would be good to have another similar test here that verifies that changing the severity updates the display.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -186,6 +188,34 @@ export class TablePageObject<T extends TableRecord> {
         return anchor as Anchor;
     }
 
+    public getRenderedCellIcon(
+        rowIndex: number,
+        columnIndex: number
+    ): Icon | null {
```
mollykreis, line 194: This can't actually return `null`, can it? You throw an error if the icon can't be found.

```suggestion
    ): Icon {
```

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -186,6 +188,34 @@ export class TablePageObject<T extends TableRecord> {
...
+                `Icon not found at cell ${rowIndex},${columnIndex}`
+            );
+        }
+        return icon;
+    }
+
+    public getRenderedCellSpinner(
+        rowIndex: number,
+        columnIndex: number
+    ): Spinner | null {
```
mollykreis, line 208: Same here. It doesn't look like this function can actually return `null` since it throws in that case.

```suggestion
    ): Spinner {
```

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -186,6 +188,34 @@ export class TablePageObject<T extends TableRecord> {
...
+            .firstElementChild;
+        if (!icon || !(icon instanceof Icon)) {
+            throw new Error(
+                `Icon not found at cell ${rowIndex},${columnIndex}`
+            );
+        }
+        return icon;
+    }
+
+    public getRenderedCellSpinner(
```
mollykreis, line 205: I was wondering the benefit of actually returning the `Spinner` instance here since I don't think there is much relevant state on it, so I did a code search to see how it's used. I don't actually see it used in any of the tests, so can it just be removed?

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -186,6 +188,34 @@ export class TablePageObject<T extends TableRecord> {
         return anchor as Anchor;
     }
 
+    public getRenderedCellIcon(
```
mollykreis, line 191: I have a bit of a reservation about both this function and the equivalent function for the spinner below. They seem a bit too coupled to the mapping-icon column with their assumption that the icon/spinner will be the first child element. For example, `getRenderedCellAnchor` directly above doesn't do that.

I think the spinner version could look exactly like the anchor implementation since you know the element's tag. The icon version seems a bit trickier. I don't have a great alternative to the implementation you have off the top of my head, but perhaps if we can't come up with a more flexible selector, the function name should somehow indicate that it is specific to the `mapping-icon` column and isn't designed to be flexible for use with any column that could possibly contain an icon...just a thought, don't know if it's really the best solution here.

packages/nimble-components/src/mapping/icon/tests/mapping-icon.stories.ts
```diff
@@ -0,0 +1,43 @@
...
+            control: { type: 'none' }
+        },
+        icon: {
+            control: { type: 'none' },
+            description:
+                'The tag name of the Nimble icon to render, e.g. `nimble-icon-check`.'
+        },
+        severity: {
+            description:
+                'One of `error`, `warning`, `success`, `information`, or unspecified. Controls the color of the icon.'
```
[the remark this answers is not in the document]
m-akinc: Sure, I guess it's better for them to use that type instead of string literals. I just referenced it by name. I don't think there's anything helpful to actually link to.

packages/nimble-components/src/mapping/icon/tests/mapping-icon.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { html } from '@microsoft/fast-element';
+import { DesignSystem } from '@microsoft/fast-foundation';
+import { MappingIcon, mappingIconTag } from '..';
+import { fixture, type Fixture } from '../../../utilities/tests/fixture';
+import { waitForUpdatesAsync } from '../../../testing/async-helpers';
+import { Icon } from '../../../icon-base';
+
+describe('Icon Mapping', () => {
```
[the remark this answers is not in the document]
m-akinc: Had to add these for the spinner mapping and text mapping, too.

packages/nimble-components/src/table-column/enum-base/models/mapping-icon-config.ts
```diff
@@ -0,0 +1,31 @@
...
+    </${icon}>`;
+
+/**
+ * Mapping configuration corresponding to a icon mapping
+ */
+export class MappingIconConfig extends MappingConfig {
+    public readonly iconTemplate: ViewTemplate<IconView>;
+    public constructor(
+        text: string,
+        public readonly severity: IconSeverity,
```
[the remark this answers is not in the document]
m-akinc: It's kind of hidden. The `MappingIconConfig` provides a template, and the template references the `severity` of the cell-view and group-header-view.

packages/nimble-components/src/table-column/icon/cell-view/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * The cell view for the icon column
+ */
+export class TableColumnIconCellView
+    extends TableCellView<
+    TableColumnEnumCellRecord,
+    TableColumnEnumColumnConfig
+    >
+    implements IconView {
+    public severity: IconSeverity;
+    public text!: string;
```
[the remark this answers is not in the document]
rajsite: It's thee same feedback as here: https://github.com/ni/nimble/pull/1392#discussion_r1283717336
The element should be treated standalone from the context of the element itself. It can't assume anything about a validator and should behave well based on it's types.

Non-null assertions and `as` casts are overriding the type. It can be done for internal properties we control, like we know a ref will be bound to a template at a specific time that the compiler can't know, but that doesn't apply to the elements public api.

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    public override validate(
+        mappings: Mapping[],
+        keyType: MappingKeyType
+    ): void {
+        super.validate(mappings, keyType);
+        this.validateIconNames(mappings);
+    }
+
+    private validateIconNames(mappings: Mapping[]): void {
```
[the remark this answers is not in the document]
m-akinc: I think we're okay. `MappingIcon.resolvedIcon` is observable, and we create `Notifier`s for our mappings (in `TableColumnEnumBase.mappingsChanged`) that cause a re-validation when a change occurs. So before the icon is resolved, we may fail validation, but when resolution happens, we'll revalidate and pass.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
m-akinc: None of our column types had these tests, so I added them.

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: Moot, since I removed it.

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: We don't have to worry about the spinner function, because I removed it. I can't think of a more general way to get our icons (except to set a certain class on every Nimble icon), so I'm renaming the function.

packages/nimble-components/src/mapping/spinner/tests/mapping-spinner.stories.ts
```diff
@@ -0,0 +1,34 @@
...
+export const spinnerMapping: StoryObj = {
+    render: createUserSelectedThemeStory(hiddenWrapper(html`<style></style>`)),
+    argTypes: {
+        key: {
+            description: mappingKeyDescription('a spinner'),
+            control: { type: 'none' }
+        },
+        text: {
+            description:
+                'A textual description of the value which will be used as the tooltip and accessible name of the spinner. The text is also displayed next to the spinner in a group header.'
```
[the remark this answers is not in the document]
m-akinc: > It worked more reliably in Chrome and Safari.

Was it not completely reliable? On Windows, in Chrome, it is consistent.

I can reproduce this in Firefox on Windows, too. If I pause the animation, then the tooltip comes up consistently. I think it's a Firefox bug. I tried to create a simple Stackblitz that reproduced it, but I couldn't. I don't think it warrants spending more time on this.

packages/nimble-components/src/table-column/icon/cell-view/index.ts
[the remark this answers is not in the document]
m-akinc: I changed the types of `text` (and `severity`) to accept `undefined`, even though we validate that they are not. If we didn't do the validation, the correct behavior would be to propagate that `undefined` value to the template so that the attributes did not get set.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
```diff
@@ -0,0 +1,136 @@
...
+    },
+    {
+        firstName: 'Bart',
+        lastName: 'Simpson',
+        status: 'success',
+        isChild: true
+    }
+];
+
+const iconColumnDescription = 'The `nimble-table-column-icon` column renders string, number, or boolean values as a Nimble icon or `nimble-spinner` in the `nimble-table`.';
```
mollykreis, line 49: @jattasNI had made this comment on my number-text PR, and I think it's relevant here too. We should make a statement of what the sort/group behavior is. Specifically in this case, it would be good to note that the raw values are sorted & grouped, not the mapped values. It would probably be good to go back and add that same note to the enum-text column, possibly in a separate PR.

packages/nimble-components/src/mapping/spinner/tests/mapping-spinner.stories.ts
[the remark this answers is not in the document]
jattasNI: Sorry, poor choice of language: it was 100% reliable in Chrome and Safari.

I think it's pretty unlikely that apps will be storing any critical info in a tooltip for a spinner in a table cell, so I'm ok to not spend any more time investigating this.

jattasNI (approved):

packages/nimble-components/src/table-column/icon/cell-view/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    private updateState(): void {
+        this.visual = undefined;
+        if (!this.columnConfig?.mappingConfigs || !this.cellRecord) {
+            return;
+        }
+        const value = this.cellRecord.value;
+        if (value === undefined || value === null) {
+            return;
+        }
```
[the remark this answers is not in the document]
m-akinc: Actually, setting the value of `this.visual` to `undefined` does cause us to update and render an empty cell.

I did not have a test for this previously, so I added one.

packages/nimble-components/src/table-column/icon/group-header-view/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    private updateState(): void {
+        this.visual = undefined;
+        if (!this.columnConfig) {
+            return;
+        }
+        const value = this.groupHeaderValue;
+        if (value === undefined || value === null) {
+            return;
+        }
```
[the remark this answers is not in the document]
m-akinc: Same explanation as for the cell view. It turned out the test I wrote for this didn't pass! That led me to find (and file) a FAST bug related to nested templates. I was able to work around it just by adding a newline to the template.

packages/nimble-components/src/table-column/icon/cell-view/index.ts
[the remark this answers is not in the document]
mollykreis: Ah, yes, you are absolutely right. I overlooked the line setting `this.visual` to `undefined`.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
m-akinc: I went ahead and updated the description for both elements.

packages/nimble-components/src/table-column/icon/cell-view/index.ts
```diff
@@ -0,0 +1,80 @@
...
+        this.updateState();
+    }
+
+    private cellRecordChanged(): void {
+        this.updateState();
+    }
+
+    private updateState(): void {
+        this.visual = undefined;
+        if (!this.columnConfig?.mappingConfigs || !this.cellRecord) {
```
rajsite, line 53: mappingConfigs isn't nullable right?

```suggestion
        if (!this.columnConfig || !this.cellRecord) {
```

packages/nimble-components/src/table-column/icon/cell-view/index.ts
```diff
@@ -0,0 +1,80 @@
...
+            this.iconTemplate = mappingConfig.iconTemplate;
+        } else if (mappingConfig instanceof MappingSpinnerConfig) {
+            this.visual = 'spinner';
+            this.text = mappingConfig.text;
+        }
+    }
+}
+
+const iconCellView = TableColumnIconCellView.compose({
+    baseName: 'table-column-icon-cell-view',
```
rajsite, line 74: add the basic definition tests

packages/nimble-components/src/table-column/icon/group-header-view/index.ts
```diff
@@ -0,0 +1,77 @@
...
+            this.iconTemplate = mappingConfig.iconTemplate;
+        } else if (mappingConfig instanceof MappingSpinnerConfig) {
+            this.visual = 'spinner';
+            this.text = mappingConfig.text ?? '';
+        }
+    }
+}
+
+const iconGroupHeaderView = TableColumnIconGroupHeaderView.compose({
+    baseName: 'table-column-icon-group-header-view',
```
rajsite, line 70: add the basic definition tests

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Validator for TableColumnIcon
+ */
+export class TableColumnIconValidator extends TableColumnEnumBaseValidator<
+    typeof iconValidityFlagNames
+> {
+    public constructor(
+        columnInternals: ColumnInternals<unknown>,
+        supportedMappingElements: readonly (typeof Mapping)[]
+    ) {
+        super(columnInternals, iconValidityFlagNames, supportedMappingElements);
```
rajsite, line 25: Do we need to take supportedMappingElements as a constructor parameter? It doesn't vary per instance of TableColumnIconValidator right?

Seems like the icon validator and the text validator can pass the supported mappings array directly to the base, don't need it parameterized by the constructor.

packages/nimble-components/src/table-column/icon/group-header-view/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.updateState();
+    }
+
+    private groupHeaderValueChanged(): void {
+        this.updateState();
+    }
+
+    private updateState(): void {
+        this.visual = undefined;
+        if (!this.columnConfig?.mappingConfigs) {
```
rajsite, line 49: mappingConfigs isn;t nullable right?


```suggestion
        if (!this.columnConfig) {
```

packages/nimble-components/src/mapping/icon/tests/mapping-icon.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        await waitForUpdatesAsync();
+
+        expect(element.resolvedIcon).toBeUndefined();
+
+        DesignSystem.getOrCreate()
+            .withPrefix('nimble')
+            .register(
+                TestIcon.compose({
+                    baseName: 'test-icon',
+                    template: html`<nimble-xmark-icon></nimble-xmark-icon>`
```
rajsite, line 47: it'll be a mute point but fyi make sure to use the tag name constants and not hard code

rajsite (approved):

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -0,0 +1,467 @@
...
+        expect(tableColumnIconTag).toBe('nimble-table-column-icon');
+    });
+
+    it('can construct an element instance', () => {
+        expect(
+            document.createElement('nimble-table-column-icon')
+        ).toBeInstanceOf(TableColumnIcon);
+    });
+
+    for (const test of [
```
mollykreis, line 80: Is there a reason not to use the existing `getSpecTypeByNamedList` pattern here?

mollykreis (approved):

packages/nimble-components/src/mapping/icon/tests/mapping-icon.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                })()
+            );
```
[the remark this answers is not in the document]
m-akinc: Looks like I don't even need to define a `TestIcon` class.

packages/nimble-components/src/table-column/icon/cell-view/index.ts
m-akinc: I had run into a situation where we were accessing `mappingConfigs` when it wasn't defined, but I can't seem to reproduce now (in Storybook or in unit tests). I guess I'll just make that change.

packages/nimble-components/src/mapping/icon/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.resolvedIcon = isIconClass(elementClass) ? icon : undefined;
+    }
+
+    private iconChanged(): void {
+        const icon = this.icon;
+        if (icon === undefined) {
+            this.resolvedIcon = undefined;
+            return;
+        }
+        const elementClass = customElements.get(icon);
```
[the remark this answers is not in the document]
m-akinc: If I do that, I will have to add a couple of non-null assertions which are lies or do some other null handling elsewhere. I think the intent is clearer if I just change the test to `!icon` (empty string can be handled the same as `null` and `undefined`).

packages/nimble-components/src/mapping/icon/tests/mapping-icon.spec.ts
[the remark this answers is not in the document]
rajsite: I'm not really sure what you mean, but you can't register the same class twice with the different names like the source seems to be doing with the XMark class:
image

please create a new class or clearly document what's going on and why that's not an issue here because I don't understand it.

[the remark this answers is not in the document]
rajsite: Maybe compose is creating a new antonymous class or something. I dunno just create a new text class that extends `IconXmark`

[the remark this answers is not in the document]
m-akinc: I'm not sure if `IconXmark` wasn't being defined before I registered it with a new name, or what. Anyway, I'm just using a new class now.

PR #1420, opened 2023-08-08
Rich text viewer | Updated the default minimum height and an example string

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

The default minimum height of the component should be one line height and currently it creates some space in the view comments in SLE (as in the below image).
image

## 👩‍💻 Implementation

* Removed having the minimum height of the component and if there is no content, the component will behave like an empty `div` element.
* Updated example string for the link as `Absolute link` which makes more sense for the feature it actually supports.

## 🧪 Testing

Manually test and verified the change.

## ✅ Checklist


suseendran-ni (approved):

packages/nimble-components/src/rich-text-viewer/styles.ts
```diff
@@ -17,7 +18,7 @@ export const styles = css`
         inline-size: auto;
         overflow: auto;
         block-size: 100%;
-        min-block-size: 36px;
+        min-block-size: ${bodyFontLineHeight};
```
vikisekarNI, line 21: Would you mind adding a default sizing test case with one line string to notice this new layout behavior in storybook?

vikisekarNI (changes requested):
minor: Looking forward to a default sizing test case with one line to witness the layout behavior in storybook

packages/nimble-components/src/rich-text-viewer/styles.ts
vivinkrishna-ni: I have added a default height sizing test case for single line usage in storybook.
image

vikisekarNI (approved):

m-akinc (approved):

packages/nimble-components/src/rich-text-viewer/styles.ts
rajsite, line 21: Do we really need to specify a minimum height in nimble? If the component is completely empty why should it use any space? It would be like an empty `div` or a `span`. Can we just remove the min size altogether?

vivinkrishna-ni: Yes, it makes sense not to have a minimum height for the viewer component and so I removed it.

rajsite (approved):

packages/nimble-components/src/rich-text-viewer/tests/rich-text-viewer-matrix.stories.ts
```diff
@@ -109,6 +109,10 @@ export const differentContentsInMobileWidth: StoryFn = createStory(html`
             [
                 'Long word',
                 'ThisIsALongWordWithoutSpaceToTestLongWordInSmallWidth'
+            ],
```
vikisekarNI, line 112: nit: Out of scope Typo. Change `sup` to `sub` for all instances in this or other files

image

vikisekarNI (approved):

PR #1421, opened 2023-08-09
Angular support for enum-text table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1103

## 👩‍💻 Implementation

- added directives for `nimble-table-column-enum-text` and supporting element `nimble-mapping-text`. Both of these are in a single module
- added tests for new directives
- added enum-text column to example program

## 🧪 Testing

Automated tests pass

## ✅ Checklist


angular-workspace/projects/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -105,7 +106,8 @@ export class CustomAppComponent {
             stringValue1: `new string ${tableData.length}`,
             stringValue2: `bar ${tableData.length}`,
             href: '/customapp',
-            linkLabel: 'Link'
+            linkLabel: 'Link',
+            statusCode: (Math.floor(Math.random() * 10) % 2) === 0 ? 100 : 101
```
mollykreis, line 110: I don't have a strong opinion either way, but I'm curious if there's a reason to make this random rather than deterministic (i.e. `statusCode: (tableData.length % 2 === 0) ? 100 : 101`).

mollykreis (approved):

atmgrifter00 (approved):

angular-workspace/projects/ni/nimble-angular/table-column/enum-text/nimble-mapping-text.directive.ts
```diff
@@ -0,0 +1,32 @@
+import { Directive, ElementRef, Input, Renderer2 } from '@angular/core';
```
rajsite, line 1: Keep consistency with nimble components' organization.

```
/mapping/text
/table-column/enum
```

So add the extra entry points and modules.

If the change in consistency was made to save a couple module import lines that's not worth it. Eventually we will move to single file components so the extra module import won't happen.

rajsite (changes requested):

angular-workspace/projects/ni/nimble-angular/table-column/enum-text/nimble-mapping-text.directive.ts
```diff
@@ -0,0 +1,32 @@
...
+import { type MappingText, mappingTextTag } from '@ni/nimble-components/dist/esm/mapping/text';
```
rajsite, line 2: > Both of these are in a single moduleBoth of these are in a single module

Pretty confident we have not done that for any components in nimble. They all get there own module. If you found a location that was exporting multiple components from the same module we should probably fix that to be consistent with the rest.

rajsite: May be useful to capture something in CONTRIBUTING to enforce consistency

angular-workspace/projects/ni/nimble-angular/table-column/enum-text/nimble-mapping-text.directive.ts
rajsite: Re-opening. Can you move the mapping elements to secondary entrypoints to align with the other table related elements. That was the description the paths above were illustrating along with the description "add the extra entry points and modules" (was trying to think ahead :P)

rajsite (commented):
Should align with other table elements. Separate entry points and modules

angular-workspace/projects/ni/nimble-angular/table-column/enum-text/nimble-mapping-text.directive.ts
m-akinc: Whoops, read too fast and missed that aspect.

rajsite (approved):

PR #1424, opened 2023-08-10
Rich Text Editor | Add setMarkdown() and getMarkdown() methods to set and get content from Tip Tap editor 

aagash-ni:
# Pull Request

## 🤨 Rationale
This PR contains logic for parsing the assigned markdown content, setting it to the editor and also serializing the editor's data to markdown content.

AzDo Feature: https://dev.azure.com/ni/DevCentral/_backlogs/backlog/ASW%20SystemLink%20LIMS/Features/?workitem=2350963
Issue: https://github.com/ni/nimble/issues/1288

## 👩‍💻 Implementation

* Exposed getMarkdown() and setmarkdown() methods to facilitate the exchange of data with the editor.
* Used MarkdownParser from [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown/tree/master) package for parsing markdown strings, DOM serializer from [prosemirror-model ](https://github.com/ProseMirror/prosemirror-model) package to serialize the content as DOM structure, XML Serializer to serialize it to HTML string and sets it to the editor.
* To serialize the tip-tap [document ](https://prosemirror.net/docs/ref/#model.Document_Structure)in getMarkdown() used MarkdownSerializer from [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown/tree/master) package to serialize the node based on schema.
* Enabled `emphasis` and  `list` rules in MarkdownParser, this allows users to set bold, italics, numbered, and bulleted lists in a CommonMark flavor to the editor. All other basic markdown formats are disabled.
* Added only respective nodes and marks for bold, italics, numbered and bulleted lists in `MarkdownSerializer`

## 🧪 Testing

* Added unit tests and visual tests for the component.
* Manually tested and verified the functionality of the supported features.

## ✅ Checklist


packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -419,4 +419,618 @@ describe('RichTextEditor', () => {
             );
         });
     });
+
+    it('set a empty string should clear a value in the editor', () => {
```
vikisekarNI, line 423:
```suggestion
    it('setting an empty string should clear a value in the editor', () => {
```

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -145,6 +170,131 @@ export class RichTextEditor extends FoundationElement {
...
+            node: Node
+        ): void {
+            const start = (node.attrs.start as number) || 1;
+            const maxW = String(start + node.childCount - 1).length;
+            const space = state.repeat(' ', maxW + 2);
+            state.renderList(node, space, i => {
+                const nStr = String(start + i);
+                return `${state.repeat(' ', maxW - nStr.length) + nStr}. `;
+            });
+        };
```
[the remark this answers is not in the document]
aagash-ni: Yes `ol`  has only  `start` attribute, For parsing and serializing we are using `defaultMarkdownParser`  and `defaultMarkdownSerializer` from the `prosemirror-markdown` package, both are using same [schema ](https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/schema.ts)internally, Especially In that schema it is having `order` [attribute ](https://github.com/ProseMirror/prosemirror-markdown/blob/b7c1fd2fb74c7564bfe5428c7c8141ded7ebdd9f/src/from_markdown.ts#L251C3-L251C40)for `orderedList` node, But once we set the parsed content to tip-tap editor, internally editor will build its own schema, marks and nodes based on extensions, which don't have order attribute in `orderedList` node and only having start attribute, So taking the `orderedList` [node ](https://github.com/ProseMirror/prosemirror-markdown/blob/b7c1fd2fb74c7564bfe5428c7c8141ded7ebdd9f/src/to_markdown.ts#L94C2-L101C7) from `prosemirror-markdown` package and changed it to `start` instead of `order` is working as expected.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -145,6 +170,131 @@ export class RichTextEditor extends FoundationElement {
...
+
+    /**
+     * This function takes the Fragment from parseMarkdownToDOM function and return the serialized string using XMLSerializer
+     */
+    private getHtmlContent(value: string): string {
+        const markdownDocumentFragment = this.parseMarkdownToDOM(value);
+        return this.xmlSerializer.serializeToString(markdownDocumentFragment);
+    }
+
+    private initializeMarkdownParser(): MarkdownParser {
```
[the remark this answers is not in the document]
aagash-ni: Moving the common parsing functionalities to a separate folder is in the pipeline, Will create subsequent PR for this implementation.

packages/nimble-components/src/rich-text-editor/index.ts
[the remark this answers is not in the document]
aagash-ni: Added reference for tip-tap document which is specific to `prose-mirror` describes the tree of nodes

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -419,4 +419,618 @@ describe('RichTextEditor', () => {
...
+            element.setMarkdown('*Italics*');
+
+            expect(pageObject.getEditorTagNames()).toEqual(['P', 'EM']);
+            expect(pageObject.getEditorLeafContents()).toEqual(['Italics']);
+        });
+
+        it('italics markdown string("_") to "em" HTML tag', async () => {
+            element.setMarkdown('_Italics_');
+
+            await connect();
```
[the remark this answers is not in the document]
aagash-ni: Removed Connect() as it is already added in beforeEach block

packages/nimble-components/src/rich-text-editor/index.ts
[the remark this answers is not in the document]
aagash-ni: To summarize, We are using a combination of `prose-mirror` and `Tiptap` where `TipTap` leverages extension support which will be useful in `@mention` in the future, and also it's internal handling in many cases whereas `prose-mirror` leverages support for markdown serializing and parsing, as Tip Tap doesn't [support markdown](https://tiptap.dev/guide/output#not-an-option-markdown) as output.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -145,6 +170,131 @@ export class RichTextEditor extends FoundationElement {
...
 
+    /**
+     * This function load tip tap editor with provided markdown content by parsing into html
+     * @public
+     */
+    public setMarkdown(value: string): void {
+        if (this.$fastController.isConnected) {
+            const htmlConvertedContent = this.getHtmlContent(value);
+            this.tiptapEditor.commands.setContent(htmlConvertedContent);
+            this.tiptapEditor.commands.focus();
```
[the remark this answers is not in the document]
aagash-ni: Removed focus() when setting markdown

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -163,6 +213,105 @@ export class RichTextEditor extends FoundationElement {
...
+         * Setting the escape argument as false to skip esc method under MarkdownSerializerState as regExpression
+         * adding unnecessary slashes while serializing and not to affect text nodes.
+         * https://github.com/ProseMirror/prosemirror-markdown/blob/b7c1fd2fb74c7564bfe5428c7c8141ded7ebdd9f/src/to_markdown.ts#L409C3-L417C4
+         */
+        const textNode = function text(
+            state: MarkdownSerializerState,
+            node: Node
+        ): void {
+            state.text(node.text!, false);
+        };
```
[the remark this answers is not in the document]
vikisekarNI: Escaping should be turned on for special markdown syntax characters like `*`,`\`,`~`,`[`,`]`,`_` to be interpreted as text. Am I missing something here?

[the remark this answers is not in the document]
aagash-ni: Removed skipping esc func while serializing text node,

- Assigned defaultMarkdownSerializer text node as MarkdownSerializer text node which will add escape character while [serializing](https://github.com/ProseMirror/prosemirror-markdown/blob/b7c1fd2fb74c7564bfe5428c7c8141ded7ebdd9f/src/to_markdown.ts#L249C1-L259C4) based on special markdown character.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -447,4 +447,662 @@ describe('RichTextEditor', () => {
...
+                }
+            );
+        }
+    });
+
+    describe('Should return markdown without any changes when various wacky string values are assigned', () => {
+        const focused: string[] = [];
+        const disabled: string[] = [];
+
+        const wackyStringsWithoutSpecialMarkdownCharacter = [
```
aagash-ni, line 1036: Removed `-Infinity` and `-2147483648/-1` from the wacky strings and asserted it in the previous block as it contains special markdown character in it, Remaining wacky strings were added here in `wackyStringsWithoutSpecialMarkdownCharacter` and asserted.

vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+export const multipleSubPointsContentInMobileWidth: StoryFn = createFixedThemeStory(
+    createMatrix(mobileWidthComponent),
+    lightThemeWhiteBackground
+);
+
+multipleSubPointsContentInMobileWidth.play = (): void => {
+    document
+        .querySelector('nimble-rich-text-editor')!
+        .setMarkdown(
+            '1. Sub point 1\n   1. Sub point 2\n       1. Sub point 3\n          1. Sub point 4\n             1. Sub point 5\n                 1. Sub point 6\n                    1. Sub point 7\n'
```
m-akinc, line 146: Can you format this string as a template literal so we can see how the indentations line up?
```
`1. Sub point 1
    1. Sub point 2
        1. Sub point 3
        ...
`
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+);
+
+editorWhenMarkdownValueSetInColorThemeDarkGreenBackground.play = playFunction;
+
+export const editorWhenMarkdownValueSetInDarkThemeBlackBackground: StoryFn = createFixedThemeStory(
+    createMatrix(styledComponent),
+    darkThemeBlackBackground
+);
+
+editorWhenMarkdownValueSetInDarkThemeBlackBackground.play = playFunction;
```
m-akinc, line 126: You should be able to use `createMatrixThemeStory` to render all three of these themes in the same story. You'll have to modify your `playFunction` to get all editors instead of just one.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+multipleSubPointsContentInMobileWidth.play = (): void => {
+    document
+        .querySelector('nimble-rich-text-editor')!
+        .setMarkdown(
+            '1. Sub point 1\n   1. Sub point 2\n       1. Sub point 3\n          1. Sub point 4\n             1. Sub point 5\n                 1. Sub point 6\n                    1. Sub point 7\n'
+        );
+};
+
+export const longWordContentInMobileWidth: StoryFn = createFixedThemeStory(
+    createMatrix(mobileWidthComponent),
```
m-akinc, line 151: `createMatrix` isn't doing anything here, right?

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -1,18 +1,35 @@
...
     footerActionButtons: boolean;
+    getMarkdown: undefined;
+    setMarkdown: undefined;
+    editorRef: RichTextEditor;
+    setMarkdownData: (args: RichTextEditorArgs) => void;
 }
 
+const plainString = 'Plain text' as const;
+
+const markdownString = 'Supported rich text formatting options:\n1. **Bold**\n2. *Italics*\n3. Numbered lists\n   1. Option 1\n   2. Option 2\n4. Bulleted lists\n   * Option 1\n   * Option 2' as const;
```
m-akinc, line 23: Please use template string so that indentations/newlines are easier to understand:
```
`Supported rich text formatting options:
1. **Bold**
2. *Italics*
...`
```

packages/nimble-components/src/rich-text-editor/tests/types.ts
m-akinc, line 1: This doesn't need to be exported from its own file. It can just be defined in the `rich-text-editor.stories.ts` file.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+
+export const editorWhenMarkdownValueSetInDarkThemeBlackBackground: StoryFn = createFixedThemeStory(
+    createMatrix(styledComponent),
+    darkThemeBlackBackground
+);
+
+editorWhenMarkdownValueSetInDarkThemeBlackBackground.play = playFunction;
+
+export const plainTextContentInMobileWidth: StoryFn = createFixedThemeStory(
+    createMatrix(mobileWidthComponent),
```
m-akinc, line 129: `createMatrix` isn't doing anything, right?

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+editorWhenMarkdownValueSetInColorThemeDarkGreenBackground.play = playFunction;
+
+export const editorWhenMarkdownValueSetInDarkThemeBlackBackground: StoryFn = createFixedThemeStory(
+    createMatrix(styledComponent),
+    darkThemeBlackBackground
+);
+
+editorWhenMarkdownValueSetInDarkThemeBlackBackground.play = playFunction;
+
+export const plainTextContentInMobileWidth: StoryFn = createFixedThemeStory(
```
m-akinc, line 128: Instead of `createFixedThemeStory` here and below, you can just use `createStory`. You're not trying to test the specific themes here.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -23,13 +26,32 @@ const metadata: Meta = {
...
+const [
+    lightThemeWhiteBackground,
+    colorThemeDarkGreenBackground,
+    darkThemeBlackBackground,
+    ...remaining
+] = backgroundStates;
+
+if (remaining.length > 0) {
+    throw new Error('New backgrounds need to be supported');
+}
```
m-akinc, line 47: If you replace uses of `createFixedThemeStory` with `createMatrixThemeStory` and `createStory` as I suggest in a couple other comments, you can remove this.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -64,6 +86,77 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
+            '1. Sub point 1\n   1. Sub point 2\n       1. Sub point 3\n          1. Sub point 4\n             1. Sub point 5\n                 1. Sub point 6\n                    1. Sub point 7\n'
+        );
+};
+
+export const longWordContentInMobileWidth: StoryFn = createFixedThemeStory(
+    createMatrix(mobileWidthComponent),
+    lightThemeWhiteBackground
+);
+
+longWordContentInMobileWidth.play = (): void => {
```
m-akinc, line 155: This test renders like this:
image
Is this meant to go beyond the bounds of the editor? I'm not sure what behavior it's testing.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -56,6 +65,22 @@ export class RichTextEditor extends FoundationElement {
...
+
+    /**
+     * @internal
+     */
+    public constructor() {
+        super();
+        this.domSerializer = DOMSerializer.fromSchema(schema);
+        this.xmlSerializer = new XMLSerializer();
+        this.markdownParser = this.initializeMarkdownParser();
+        this.markdownSerializer = this.initializeMarkdownSerializer();
```
m-akinc, line 81: I would do these initializations in the declarations above, so you don't need a constructor.

m-akinc (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Updated the width as 300px in component styling and while setting long word it will wrap on to the next line
image

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Initialized in the declarations and removed the constructor.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Updated as `createStory`

packages/nimble-components/src/rich-text-editor/tests/types.ts
aagash-ni: Moved to the current file instead of exporting it from a separate file

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
aagash-ni: Updated to template string.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Removed `createMatrix`.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -36,19 +67,56 @@ const metadata: Meta<RichTextEditorArgs> = {
...
+        data-unused="${x => x.setMarkdownData(x)}"
+    >
         ${when(x => x.footerActionButtons, html`
             <${buttonTag} appearance="ghost" slot="footer-actions">Cancel</${buttonTag}>
             <${buttonTag} slot="footer-actions">OK</${buttonTag}>`)}
     </${richTextEditorTag}>
     `),
     argTypes: {
+        data: {
+            name: 'setMarkdown(value)',
```
jattasNI, line 81: I noticed that when I toggle from "Plain string" to "...all supported...", the editor ends up in a state where some of the toggle buttons are active but the cursor is not present inside the editor. This was unexpected to me: I thought the buttons would only be active to reflect the state of the text at the cursor. If the editor doesn't have focus, I'd expect all the buttons to be toggled off. Instead it's behaving as if the cursor is at the end of the text even thought the cursor isn't visible.

This is only a minor quirk so if it's a challenge to fix then I'm ok leaving it as-is, but I think it's worth spending some time investigating it.

image

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -447,4 +447,662 @@ describe('RichTextEditor', () => {
...
+        expect(pageObject.getEditorLeafContents()).toEqual(['']);
+
+        element.setMarkdown('new markdown string');
+        expect(pageObject.getEditorTagNames()).toEqual(['P']);
+        expect(pageObject.getEditorLeafContents()).toEqual([
+            'new markdown string'
+        ]);
+    });
+
+    describe('supported rich text formatting options from markdown string to its respective HTML elements', () => {
```
jattasNI, line 467: Are these test cases identical to the ones that already verify the markdown to DOM conversion for the rich text viewer? I mostly want to make sure that our coverage is good for both components.

We could also think about ways to share the tests. For example, share a data structure with markdown values and expected tags / leaf contents then have both components test suites iterate over that data. But I'm also OK with duplicating the tests as long as we're diligent about keeping them in sync.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -153,6 +167,31 @@ export class RichTextEditor extends FoundationElement {
         return true;
     }
 
+    /**
+     * This function load tip tap editor with provided markdown content by parsing into html
+     * @public
+     */
+    public setMarkdown(markdown: string): void {
+        if (this.$fastController.isConnected) {
```
rajsite, line 175: These guards seem incorrect to me, the users should be able to call setMarkdown and getMarkdown just fine before the element is connected to the DOM.

ie
```ts
const el = document.createElement('nimble-rich-text-editor');
el.setMarkdown('some markdown');
el.getMarkdown();
```

It seems like tiptap supports this concept: https://tiptap.dev/api/editor#element

> You can even initiate your editor before mounting it to an element. This is useful when your DOM is not yet available. 

I didn't think about it too much but conceptually I think we can have in the template:
```ts
<div class="container">
    <section class="editor-container" ${ref('editorContainer')}>
    </section>
</div>
```

```css
.editor-container {
    display: contents;
}
```

```ts
class RichTextEditor {
    constructor(){
        this.editor = document.createElement('div');
        // set classes, role, etc on the element
        this.tiptTapEditor = new Editor({element})
    }

    connectedCallback() {
        if (!this.editor.isConnected) {
            this.editorContainer.append(this.editor);
        }
        this.bindEditorEventThings();
    }
}
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
rajsite: Notice how below a filter was used to exclude some strings from the list. Don't copy and paste and modify the list, prefer to exclude items from the list.

rajsite: And then make a separate test where those excluded items are explicitly tested so someone can clearly see how just that set of strings are tested. (currently they are merged up into the above unrelated special markdown strings test)

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -447,4 +447,662 @@ describe('RichTextEditor', () => {
...
+
+                    await disconnect();
+                }
+            );
+        }
+    });
+
+    describe('Should return markdown with escape character (back slash) when various special markdown syntax are assigned', () => {
+        const specialMarkdownStrings: { name: string, value: string }[] = [
+            { name: '> blockquote', value: '\\> blockquote' },
```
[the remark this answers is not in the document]
rajsite: Use the String.raw tag template literal to make these easier to read: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw

Can alias it as `const r = String.raw` or something

Might not help with the ones that are testing backticks but the others will be clearer

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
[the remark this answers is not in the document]
aagash-ni: Updated using `String.raw` in all backslash places except string having backticks.

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Moved the editor initialization to the constructor and appended it to dom in `connectedCallback()`. Removed the guard in `setMarkdown()` and `getMarkdown()` methods, where the consumer can set and get markdown irrespective of dom.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
aagash-ni: As mentioned [here](https://github.com/ni/nimble/pull/1424#discussion_r1291010057), refactoring PR parsing logic between viewer and editor and moving it to the same files under `models` is in the pipeline.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
aagash-ni: This is the expected behavior in [ TipTap editor ](https://tiptap.dev/examples/default), that it will make the toggle button active based on the end of the content being set.

One way is to add editor focus programmatically when setting markdown, which will make the cursor active at the document

```
   public setMarkdown(markdown: string): void {
        const html = this.getHtmlContent(markdown);
        this.tiptapEditor.commands.setContent(html);
        this.tiptapEditor.commands.focus()
  }
```

jattasNI (commented):
Looking good but I'd like to take one last look once the predecessor PR has completed and this has been retargeted to main.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -163,7 +205,100 @@ export class RichTextEditor extends FoundationElement {
...
+        return this.domSerializer.serializeFragment(
+            parsedMarkdownContent.content
+        );
+    }
+
     private initializeEditor(): void {
+        // Creating div element to land tiptap editor
+        this.editor = document.createElement('div');
+        this.editor.className = 'editor';
+        this.editor.ariaMultiLine = 'true';
```
rajsite, line 299: The aria* properties don't work on firefox, you'll need to do `setAttribute('aria-multiline', 'true')` instead

vivinkrishna-ni: Thanks for your input. We have updated to `setAttribute` for setting the aria attribute and the test case passed.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
jattasNI: Hmm, I'd be wary of having the `setMarkdown()` method change the focus. It's a side effect that would be unexpected to me and the multiple editor case you mentioned would also be problematic.

If this is TipTap's default behavior then let's leave it as-is and see if any users have feedback about it.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -3,6 +3,15 @@ import { DesignSystem, FoundationElement } from '@microsoft/fast-foundation';
 import { keyEnter, keySpace } from '@microsoft/fast-web-utilities';
```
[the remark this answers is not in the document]
jattasNI: I went ahead and added the change file so that we could get a passing build and complete our reviews sooner.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -163,7 +205,100 @@ export class RichTextEditor extends FoundationElement {
...
+        }
+
+        return this.domSerializer.serializeFragment(
+            parsedMarkdownContent.content
+        );
+    }
+
     private initializeEditor(): void {
+        // Creating div element to land tiptap editor
+        this.editor = document.createElement('div');
```
jattasNI, line 297: Any reason this `div` can't be added in the template? That would allow you to delete this code that creates it and the code that appends it in `connectedCallback()` and it would make it easier to see the structure in the declarative HTML.

As a general rule we strive for:
1. DOM structure in templates
2. styling logic in CSS
3. as little as possible in TS (though there will always be things we need to do like validation / event handling / configuring 3rd party libraries that only offer JS APIs)

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -163,7 +205,100 @@ export class RichTextEditor extends FoundationElement {
...
+        const marks = {
+            italic: defaultMarkdownSerializer.marks.em!,
+            bold: defaultMarkdownSerializer.marks.strong!
+        };
+        return new MarkdownSerializer(nodes, marks);
+    }
+
+    /** This function takes a markdown string, parses it using the ProseMirror MarkdownParser, serializes the parsed content into a
+     * DOM structure using a DOMSerializer, and returns the serialized result.
+     * If the markdown parser returns null, it will clear the editor component by creating an empty document fragment.
```
jattasNI, line 282: This comment doesn't seem quite right to me: the function doesn't clear the editor component, it returns an empty document fragment. It's the caller that applies that to the editor component.

As a general rule we only document private methods if they're doing something that's not obvious from their name and implementation. Our rationale from the Clean Code philosophy is that they tend to be wrong / out of date like this. It's up to you and your team, but I'd be fine with deleting this comment and many of the other ones on private methods entirely.

jattasNI (approved):

packages/nimble-components/src/rich-text-editor/index.ts
rajsite: Yea this is an edge case, need a reference to the element for the library before shadow dom instantiation / attach to the dom. Was the best I could think of

jattasNI: Ah, gotcha. Let's update the comment to capture that better. Something like "Create div from the constructor because the TipTap editor requires its host element before the template is instantiated"

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -75,7 +76,7 @@ export class RichTextEditorPageObject {
      * To click a formatting button in the footer section, pass its position value as an index (starting from '0')
      * @param buttonIndex can be imported from an enum for each button using the `ButtonIndex`.
      */
-    public async clickFooterButton(buttonIndex: number): Promise<void> {
+    public async clickFooterButton(buttonIndex: ToolbarButton): Promise<void> {
```
rajsite, line 79: nit: none of these parameters should have the name `index` in them. that's an internal implementation detail we are hidding with the enum. just call them `button: ToolbarButton`, etc.

packages/nimble-components/src/rich-text-editor/index.ts
rajsite: Re-opening again, unless I am mistaken I still do not see any test that asserts that the getMarkdown and setMarkdown methods actually work **before** the element is connected to the DOM.  Every test seems to connect first and then test something.

I'll mark the PR approved assuming we have this test case and it goes well. @mollykreis or @m-akinc can one of y'all verify this test case is added (that the element can be created and calling set and get markdown functions even though the element is not attached to the dom, and after attached we see the expected content)

rajsite (approved):

packages/nimble-components/src/rich-text-editor/index.ts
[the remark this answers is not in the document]
rajsite: I have a slight preference for having a folder structure of:

```
/rich-text
    /editor
        ... unshared models and existing content here
    /viewer
    /models
        ... shared models here
```

to avoid having a top-level empty `rich-text` folder assuming that doesn't have any other complication.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Added separate describe block `RichTextEditor Before DOM Connection` to check the `setMarkdown()` and `getMarkdown()` before the element attached to dom and asserted once after attached to dom.

m-akinc (approved):

PR #1426, opened 2023-08-10
Improved rendering performance

munteannatan:
# Pull Request

## 🤨 Rationale

The rendering part of the pipeline gets triggered each time a transform event update happens, and it takes a while to get the expected response, especially with very large amounts of data.

This can be improved by removing the sorting of the data as a part of the transform event update and the filtering of the drawn values based on their position relative to the transformed canvas.

## 👩‍💻 Implementation

broke apart the rendering into sorting and drawing.

created a filter which checks that dies and text can be drawn on the canvas before drawing them.

## 🧪 Testing

local benchmark tests show up to 10x increase in rendering performance when transform events are triggered with a data load of 1M based on the zoom level.
before
after

## ✅ Checklist


packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -49,61 +64,70 @@ export class RenderingModule {
...
+        ]);
+        const dieWidth = this.wafermap.dataManager.dieDimensions.width;
+        const dieHeight = this.wafermap.dataManager.dieDimensions.height;
+        trCanvasMinPointX -= dieWidth;
+        trCanvasMinPointY -= dieHeight;
 
         for (const die of this.dies) {
-            if (!prev) {
-                this.wafermap.canvasContext.fillStyle = die.fillStyle;
+            if (
```
DStavilaNI, line 81: I think that the content of this statement should be encapsulated as a function -> "isDieVisible" ... especially considering that the same logic is used in 2 places - i.e. for text rendering as well.

DStavilaNI (changes requested):

packages/nimble-components/src/wafer-map/modules/rendering.ts
munteannatan: there is no big overall difference between this function call and no function call
no function
function

DStavilaNI (approved):

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -10,13 +10,28 @@ export class RenderingModule {
...
+            if (b.fillStyle > a.fillStyle) {
+                return -1;
+            }
+
+            return 0;
+        });
+        this.updateDrawnWafer();
+    }
+
+    public updateDrawnWafer(): void {
```
atmgrifter00, line 27: nit: I think the name change for this function seems unnecessary/overly verbose. The idea that the wafer has already been "drawn" seems irrelevant, and generally the term "update" doesn't always mean a complete refresh happens, which is happening in the implementation. So, `drawWafer` still seems entirely appropriate for the method name.

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -10,13 +10,28 @@ export class RenderingModule {
...
         this.wafermap.canvasContext.save();
         this.clearCanvas();
         this.scaleCanvas();
-        this.sortDies();
         this.renderDies();
-        this.renderText();
+        if (!this.wafermap.dieLabelsHidden) {
```
atmgrifter00, line 32: nit: Any reason not to put this logic in the `renderText` method? I don't feel too strongly about this, but sometimes if each step has conditions associated with it (doesn't seem to the case here for now), then avoiding a bunch of `if` statements in a row can help maintain clarity of intent.

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -49,61 +64,76 @@ export class RenderingModule {
...
-            if (prev && die.fillStyle !== prev.fillStyle && die.fillStyle) {
-                this.wafermap.canvasContext.fillStyle = die.fillStyle;
+            if (
+                this.isDieVisible(
+                    die,
+                    transformedCanvasMinPoint,
+                    transformedCanvasMaxPoint
+                )
+            ) {
+                if (fillStyle !== die.fillStyle) {
```
atmgrifter00, line 90: Why bother with the `if` block here? The only possible difference the `if` block provides is if they were the same, but `context.fillStyle` _wasn't_ set to `die.fillStyle`, and you wanted to maintain that misalignment, which I'm guessing isn't the case.

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -49,61 +64,76 @@ export class RenderingModule {
...
+        const dieHeight = this.wafermap.dataManager.dieDimensions.height;
+        const dieSize = dieWidth * dieHeight * (this.wafermap.transform.k || 1);
         if (dieSize >= this.minDieDim) {
+            const fontsize = this.wafermap.dataManager.labelsFontSize;
+            const context = this.wafermap.canvasContext;
+            context.font = `${fontsize.toString()}px sans-serif`;
+            context.fillStyle = '#ffffff';
+            context.textAlign = 'center';
+            context.lineCap = 'butt';
+            const approxTextHeight = context.measureText('M');
```
atmgrifter00, line 110: nit: Just go ahead and spell out `approximate`. The extra 5 characters is fine. :)

packages/nimble-components/src/wafer-map/modules/rendering.ts
munteannatan: Setting the canvas context fillStyle is an intensive operation so we want to limit it as much as we can.
The dies are presorted based on their fill style because of that, so we want to change it only when a new batch of dies needs to be drawn

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -49,61 +64,76 @@ export class RenderingModule {
...
+                    this.isDieVisible(
+                        die,
+                        transformedCanvasMinPoint,
+                        transformedCanvasMaxPoint
+                    )
+                ) {
+                    context.fillText(
+                        die.text,
+                        die.x + dieWidth / 2,
+                        die.y + dieHeight / 2 + approxTextHeight.width / 2,
```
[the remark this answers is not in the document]
munteannatan: yeah, the linter complains, should I add linter exceptions?

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -170,18 +170,23 @@ export class WaferMap extends FoundationElement {
             this.eventCoordinator.detachEvents();
             if (this.waferMapUpdateTracker.requiresContainerDimensionsUpdate) {
                 this.dataManager.updateContainerDimensions();
+                this.renderer.updateSortedDies();
```
rajsite, line 173: If I'm reading this right all the code paths are still resulting in drawWafer being called?

If that's the case I'd recommend moving the drawWafer call out of updateSortedDies and keep it where it was at the end of this function running in all code paths.

An alternative could be renaming the call to `updateSortedDiesAndDrawWafer`.  As just the name `updateSortedDies` it seems to be hiding the important two responsibilities of updating the dies sort order and being responsible for calling drawWafer.

packages/nimble-components/src/wafer-map/modules/rendering.ts
rajsite: Think it's worth adding a comment around any performance cliffs / surprising API perf behavior that you find, just a quick `Updating the canvas  context fillStyle may be slow so avoid if unnecessary`. A link to issues or discussion on StackOverflow, etc. with context could be nice if you have it.

Helpful for refactors etc to know what to double check for. Also those kind of perf cliffs may differ across browsers or platforms or go away over time so the context is useful.

rajsite (approved):

packages/nimble-components/src/wafer-map/modules/rendering.ts
munteannatan: I rechecked the assumptions we had with an even larger 5M dataset and the repeated assignment of the canvas fillStyle only increased the runtime with 20 ms out of a 600 total, which is less than what we were fearing. I think the filtering reduced the impact it had in the past on performance.

Given that we plan to get rid of the merge sorting and replace it with a bucket sorting when creating the render info, i removed the if block altogether

rajsite (approved):

PR #1429, opened 2023-08-11
Blazor support for enum-text table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1103 

## 👩‍💻 Implementation

Added Blazor integration for both the `nimble-table-column-enum-text` and related `nimble-mapping-text` elements. Included tests and added enum text column to example app.

## 🧪 Testing

Tests pass and column works in example app.

## ✅ Checklist


packages/nimble-blazor/NimbleBlazor/MappingKeyType.cs
```diff
@@ -0,0 +1,17 @@
+namespace NimbleBlazor;
+
+#pragma warning disable CA1720
```
m-akinc, line 3: I'm not sure if I'll get pushback on this. The rule disallows type names, including `String` and `Boolean` in identifiers. These names represent JS types, and it would be weird to call them something else.

mollykreis: I'll let someone else, like @atmgrifter00, that is more aware of the problems that could arise from these names weigh in, but a possible alternative (if deemed necessary) could be `StringKey`.

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -218,6 +218,11 @@
                     <NimbleIconKey title="First Name"></NimbleIconKey>
                 </NimbleTableColumnText>
                 <NimbleTableColumnAnchor HrefFieldName="Href" LabelFieldName="LinkLabel" ColumnId="link-column" >Link</NimbleTableColumnAnchor>
+                <NimbleTableColumnEnumText FieldName="StatusCode" KeyType="MappingKeyType.Number" ColumnId="status-column">
```
atmgrifter00, line 221: This isn't using the implementation strategy as [outlined in the spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md#blazor-integration). Did that not work? Ultimately I think it would be good to not have these string representations of types in the C# API.

packages/nimble-blazor/NimbleBlazor/MappingKeyType.cs
atmgrifter00: As I indicated in another comment, if I have my way, this type goes away.

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor
m-akinc: Thanks for reminding me. I am now using generics.

packages/nimble-blazor/NimbleBlazor/MappingKeyType.cs
m-akinc: Yep, this type went away.

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleTableColumnEnumTextTests.cs
```diff
@@ -0,0 +1,188 @@
...
+    [Fact]
+    public void NimbleTableColumnEnumText_WithFieldNameAttribute_HasMarkup()
+    {
+        var column = RenderWithPropertySet<int, string>(x => x.FieldName!, "Status");
+
+        var expectedMarkup = @"field-name=""Status""";
+        Assert.Contains(expectedMarkup, column.Markup);
+    }
+
+    private IRenderedComponent<NimbleTableColumnEnumText<TKey>> Render<TKey>()
```
atmgrifter00, line 154:
```suggestion
    private IRenderedComponent<NimbleTableColumnEnumText<TKey>> RenderTableEnumColumn<TKey>()
```

atmgrifter00:
Is it possible to write an acceptance test that demonstrates that you can _not_ add a mapping with a key value that isn't typed the same as declared by the column? This might be difficult, as the compiler should basically preemptively disallow this.

Regardless, we should show that these new column types work as expected by writing acceptance tests for them.

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnEnumText.razor
```diff
@@ -0,0 +1,22 @@
+﻿@namespace NimbleBlazor
+@inherits NimbleTableColumn
+@attribute [CascadingTypeParameter(nameof(TKey))]
+@typeparam TKey
+
+<nimble-table-column-enum-text
```
mollykreis, line 6: This doesn't have `@attributes="AdditionalAttributes"`, but it doesn't look like our other columns have that in Blazor either. Do you know if that's intentional? @atmgrifter00, perhaps you know?

mollykreis (approved):

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnEnumText.razor
atmgrifter00: Yeah, that's not intentional. For now, we should have that on all of our public component APIs.

packages/nimble-blazor/NimbleBlazor/Components/NimbleMappingText.razor
```diff
@@ -0,0 +1,7 @@
+@namespace NimbleBlazor
+@typeparam TKey
+
+<nimble-mapping-text
+    key="@Key"
+    text="@Text">
```
atmgrifter00, line 6: I think even this component should get an `AddtionalAttributes` parameter.

atmgrifter00 (approved):

packages/nimble-blazor/NimbleBlazor/Components/NimbleMappingText.razor
rajsite: yep, every component is the current direction: https://github.com/ni/nimble/blob/main/packages/nimble-blazor/CONTRIBUTING.md#conventions

rajsite (approved):

msmithNI (approved):

PR #1430, opened 2023-08-14
Blazor support for date-text column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1014 

## 👩‍💻 Implementation

Added Blazor integration for the nimble-table-column-date-text element. Included tests and added date-text column to example app.

## 🧪 Testing

Tests pass and column works in example app.

## ✅ Checklist


packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleTableColumnDateTextTests.cs
```diff
@@ -0,0 +1,238 @@
...
+        var expectedMarkup = @"custom-format-matcher=""basic""";
+        Assert.Contains(expectedMarkup, table.Markup);
+    }
+
+    [Fact]
+    public void NimbleTableColumnDateText_WithCustomHour12Attribute_HasTableMarkup()
+    {
+        var table = RenderWithPropertySet(x => x.CustomHour12, false);
+
+        var expectedMarkup = @"custom-hour12=""false""";
```
m-akinc, line 144: This test is currently failing because the `custom-hour12` attribute is behaving like a standard boolean. There is a change in the Angular support PR that fixes this (in nimble-components). When that PR goes in, I'll merge and the test should pass.

m-akinc:
Requesting buddy review from @mollykreis

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnDateText.razor
```diff
@@ -0,0 +1,39 @@
+﻿@namespace NimbleBlazor
+@inherits NimbleTableColumn
+
+<nimble-table-column-date-text
```
mollykreis, line 4: I believe this should have `@attributes="AdditionalAttributes"`.

packages/nimble-blazor/NimbleBlazor/DateTextColumnTypes.cs
```diff
@@ -0,0 +1,276 @@
+﻿// Suppressing rule "Identifiers should not contain type names" because "long" and "short" come from the underlying API
+#pragma warning disable CA1720
+namespace NimbleBlazor;
+
+public enum DateTextFormat
+{
+    Custom
```
mollykreis, line 7: Shouldn't this enum also include `Default`?

m-akinc: The default is to not specify a value, so I didn't think I needed it as an enum value. But I guess I can just add a `Default` that maps to `null`.

mollykreis: `Default` mapping to `null` is what I'd expect. I believe that's what we do with our other enums in Blazor.

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleTableColumnDateTextTests.cs
```diff
@@ -0,0 +1,238 @@
...
+    [Fact]
+    public void NimbleTableColumnDateText_WithCustomLocaleMatcherAttribute_HasTableMarkup()
+    {
+        var table = RenderWithPropertySet(x => x.CustomLocaleMatcher, LocaleMatcherAlgorithm.Lookup);
+
+        var expectedMarkup = @"custom-locale-matcher=""lookup""";
+        Assert.Contains(expectedMarkup, table.Markup);
+    }
+
+    [Fact]
```
atmgrifter00, line 49: If we wanted to be as [robust as other tests](https://github.com/ni/nimble/blob/808e6d7db959cd324502c1d07be7566e03ad32ae/packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleBannerTests.cs#L34C4-L34C4), we would test all enum values with a `Theory`.

atmgrifter00:
Should we go ahead and write another acceptance test for this column? I'm somewhat on the fence about this, in that I'm not quite convinced that we get a lot of value from acceptance tests for checking that a component does, in fact, handle its `RenderContent` correctly (despite my request for the acceptance test for the enum text column). The obvious value we get from acceptance tests is when we need to test our custom JS logic, but that doesn't apply here.

Thoughts @rajsite, @msmithNI, @m-akinc, @mollykreis?

packages/nimble-blazor/Tests/NimbleBlazor.Tests/Unit/Components/NimbleTableColumnDateTextTests.cs
m-akinc: That was fun

packages/nimble-blazor/NimbleBlazor/DateTextColumnTypes.cs
```diff
@@ -4,12 +4,17 @@ namespace NimbleBlazor;
...
 public enum DateTextFormat
 {
+    Default,
     Custom
 }
 
 internal static class DateTextFormatExtensions
 {
-    private static readonly Dictionary<DateTextFormat, string> _enumValues = AttributeHelpers.GetEnumNamesAsKebabCaseValues<DateTextFormat>();
+    private static readonly Dictionary<DateTextFormat, string?> _enumValues = new Dictionary<DateTextFormat, string?>
```
mollykreis, line 13: This should have a consistent implementation as our other enums in Blazor.

A few examples:
- [SpinnerAppearance](https://github.com/ni/nimble/blob/main/packages/nimble-blazor/NimbleBlazor/SpinnerAppearance.cs)
- [AnchorAppearance](https://github.com/ni/nimble/blob/main/packages/nimble-blazor/NimbleBlazor/AnchorAppearance.cs)
- [BannerSeverity](https://github.com/ni/nimble/blob/main/packages/nimble-blazor/NimbleBlazor/BannerSeverity.cs)

mollykreis (approved):

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
```diff
@@ -67,6 +67,7 @@ public void UpdateTableData(int numberOfRows)
                     Faker.Name.Last(),
                     "https://nimble.ni.dev",
                     "Link",
+                    DateTime.UtcNow,
```
rajsite, line 70: UUse deterministic values in demos (like we do in stroybook)

rajsite (approved):

msmithNI (approved):

m-akinc:
@atmgrifter00 I'm going to assume from the lack of responses to the question about acceptance tests and the approvals that there's not a need to create an acceptance test in this case.

PR #1431, opened 2023-08-14
Improve cell view typing

mollykreis:
# Pull Request

## 🤨 Rationale

There were a few typing inconsistencies related to creating a cell view for the nimble-table. This PR addresses those inconsistencies and updates usages accordingly.

The problems with the cell view that are fixed in this PR are:
1. The cell view used the non-null (`!`) assertion when declaring `columnConfig`, which isn't correct. Within `ColumnInternals`, `columnConfig` is declared as `public columnConfig?: TColumnConfig;`, so it is valid for the `columnConfig` to be `undefined`. Therefore, we should make sure the cell view type reflects that possibility.
2. When a row's record didn't contain the expected fields, the cell state would be set to `{ cellRecord: {}, columnConfig: {} }`. This leads to both the `cellRecord` and the `columnConfig` on the cell view to be empty objects, which shouldn't be allowed given the typing. This also forces cell view implementations to check for scenarios that shouldn't be allowed given the configured TypeScript types.

## 👩‍💻 Implementation

To fix the problems described above, a few type changes were made:
1. Make `cellRecord` and `columnConfig` both optional within the `TableCellState` interface and in the base `TableCellView` class.
2. Update the code that validates a row's record to set an undefined `TableCellState` on the cell if the record doesn't contain the expected fields rather than creating a `TableCellState` object with `cellRecord` and `columnConfig` both set to `{}`.
3. Update usages to ensure they are correctly handling the possibility of an `undefined` `cellRecord` and `columnConfig`.

## 🧪 Testing

Ran existing auto tests

## ✅ Checklist

### DIFF
--- a/angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
+++ b/angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
@@ -33,11 +33,11 @@ export class NimbleTableColumnAnchorNavigationGuardDirective {
 
         const cellView = delegatedEvent.target as TableColumnAnchorCellView;
         // Based on: https://github.com/angular/angular/blob/35a3fde5b71ef3b50282fe6f7b37ca1c92b8d2a0/packages/router/src/directives/router_link.ts#L457
-        if (typeof cellView.columnConfig.target === 'string' && cellView.columnConfig.target !== '_self') {
+        if (typeof cellView.columnConfig?.target === 'string' && cellView.columnConfig.target !== '_self') {
             return;
         }
 
-        const href = cellView.cellRecord.href;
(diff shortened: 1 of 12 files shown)


mollykreis:
@atmgrifter00 - can you buddy this for me?

angular-workspace/projects/ni/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts
```diff
@@ -33,11 +33,11 @@ export class NimbleTableColumnAnchorNavigationGuardDirective {
```
rajsite, line 33: Looks like the groupHeaderValue needs the same treatment? 
https://github.com/ni/nimble/blob/db6fc1fd69e6919797f500197c0647a24c5f9f61/packages/nimble-components/src/table-column/base/group-header-view/index.ts#L25

Issue #682, opened 2022-08-11
NimbleCombobox in Angular invalid text access


atmgrifter00:
## 😯 Problem to Solve

During the process of bringing over the `NimbleCombobox` into SLE, one of the concessions made in the implementation was that we could no longer set a property of an object to be the current text that was present in the input.

The implementation in question looked like the following:
```
    private readonly workspaceValidator: ValidatorFn = (control: AbstractControl): { [key: string]: any } | null => {
        return typeof control.value === 'string'
            ? { forbiddenWorkspace: { value: control.value } }
            : null;
    };
```

To match the current behavior, we had to import the `Combobox` type, use a `ViewChild` of `ElementRef<Combobox>` type, and access the `value` via the `nativeElement`, which we prefer to avoid.

This feature request is meant to raise this concern and offer a couple of options.

## 💁 Proposed Solution

**Option 1**

Abandon the idea of returning a symbol instance in the "not found" case for the Control Value Accessor, and instead return an instance of a new class whose lone property would be the current text of the input. The class could still even be called `OptionNotFound`, but it would offer a property like `currentValue`.

Pros:
- Solves the particular case outlined above
- Provides users access to the current text of the combobox in any value notification mechanism

Cons:
- Doing the value validation check is now more cumbersome. What is the right way for a user to write something akin to this:
```
       if (value === OPTION_NOT_FOUND) {
            ... // bad form logic
        } else {
            ... // valid form logic
        }
```

**Option 2**

Expose the native `value` properties that FAST exposes on their components directly on the Angular directives for all of our components.

Pros:
- Aligns the Angular API with what a framework like Blazor demands (explicit exposure of the `Value` properties).
- Consistent with native HTML element APIs that are used in Angular that still leverage Angular Control Value Accessors (e.g. `input` and `select`).
- Don't have to change the CVA API of the `NimbleCombobox`, in that it can continue to just return the `OPTION_NOT_FOUND` symbol instance.

Cons:
- Possible confusion for user for how to bind to value (like in form scenarios)
- Accessing value text in form binding scenarios requires the extra implementation to create a `ViewChild`

Issue #1303, opened 2023-06-14
Select and Combobox Refresh


rajsite:
## 📌 User Story

The following are topics that should be addressed (included or closed):

Both: 
- [x] Placeholder option (when a user has not selected a value): https://github.com/ni/nimble/issues/350
- [ ] Placeholder for empty options list (when filter on combobox or select results in all options being removed): https://github.com/ni/nimble/issues/1309
- [ ] Labels: https://github.com/ni/nimble/issues/94 https://github.com/ni/nimble/issues/280
- [x] Limit the height of the dropdown: https://github.com/ni/nimble/issues/1870
- [x] Clear selected value: https://github.com/ni/nimble/issues/1951
- [ ] Dynamically fetched options (show a spinner while fetching options): 
    - https://github.com/ni/nimble/issues/1311
    - https://github.com/ni/nimble/issues/506#issuecomment-1611766299
    - https://github.com/ni/nimble/issues/1866
- [x] Groupable options: https://github.com/ni/nimble/issues/791
- [ ] ~~"Load more" / "Show all" / Action button / Checkbox (use case: dynamic loading of additional options)~~
- [ ] ~~More filter options~~
    - ~~Case sensitive~~
- [ ] Slottable content in options (i.e. icons): https://github.com/ni/nimble/issues/1135 https://github.com/ni/nimble/issues/1254
- [ ] ~~Slottable content in selector area: https://github.com/ni/nimble/issues/1310~~
- [ ] ~~Editable options (add new entries): https://github.com/ni/nimble/issues/507~~
- [ ] ~~Context menu key interactions: https://github.com/ni/nimble/issues/798~~
- [ ] ~~Virtualization of options: https://github.com/ni/nimble/issues/710~~
- [ ] ~~Visual when used in table: https://github.com/ni/nimble/issues/1190~~

Select: 
- [x] Filterable: https://github.com/ni/nimble/issues/1273
- [ ] ~~Multiselect: https://github.com/ni/nimble/issues/341~~ will implement as a picker component

Combobox:
- [ ] ~~Autocomplete mode that allows searching any part of the string: https://github.com/ni/nimble/issues/535~~
- [ ] Control Value Accessor behavior: https://github.com/ni/nimble/issues/682 https://github.com/ni/nimble/issues/1168
- [x] Dynamic options + good filtering modes allow us to make the open api private again, discussion for why it was made public: https://github.com/ni/nimble/issues/1324#issuecomment-1631540180
[...]

PR #1432, opened 2023-08-15
Enable Angular combobox to read value & provide more detailed docs

mollykreis:
# Pull Request

## 🤨 Rationale

I was given the feedback that our existing docs for the combobox are confusing and misleading because we only document the behavior of the native element without any additional documentation that explains the behavior in our supported frameworks (or even implies that the API behavior might be different). It is also not ideal that our previous "recommended" workflow for reading the value associated with an `OPTION_NOT_FOUND` Angular combobox was through the `nativeElement`. Therefore, I've added a getter for the `value` in Angular.

I expect that at least part of the documentation and implementation will need to be updated when all the issues captured in #1303 are addressed, but I think this change is still good to enable clients to be more successful with the combobox before a more complete fix is in place.

Fixes https://github.com/ni/nimble/issues/682

## 👩‍💻 Implementation

- Add getter for `value` on the Angular combobox directive
- Created `.mdx` story for the comboxbox so that we can create docs with improved formatting.

## 🧪 Testing

- Manually verified that the doc page looks correct.
- New unit tests for reading the `value` of a combobox from Angular

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/combobox/tests/combobox.mdx
+++ b/packages/nimble-components/src/combobox/tests/combobox.mdx
@@ -1,0 +1,31 @@
+import {
+    Controls,
+    DocsStory,
+    Meta,
+    Stories,
+    Title,
+    Description
+} from '@storybook/blocks';
+import * as comboboxStories from './combobox.stories';
+
+<Meta of={comboboxStories} />
+
+<Title of={comboboxStories} />
+<Description of={comboboxStories} />
+
+<DocsStory of={comboboxStories.underlineCombobox} expanded={false} />
+<Controls of={comboboxStories.underlineCombobox} />
+
+# Usage Docs
+
+## Combobox value
+
+### Native element and Blazor
+
+The value of the combobox comes from the text content of the selected autocomplete `nimble-list-option`, or, if no matching autocomplete is found, the value is the user-entered text.
+
+### Angular
+
+In Angular, an autocomplete `nimble-list-option` can be created with an `ngValue`. In that case the `ngModel` of the combobox will be the `ngValue` of the matched autocomplete option.
+If no matching autocomplete option is found, the `ngModel` of the combobox will be set to `OPTION_NOT_FOUND`, and if the application needs to access the user-entered text, that should be done
+through the `value` on the `nativeElement` associated with the `nimble-combobox`.
--- a/packages/nimble-components/src/combobox/tests/combobox.mdx
+++ b/packages/nimble-components/src/combobox/tests/combobox.mdx
@@ -28,4 +28,4 @@ The value of the combobox comes from the text content of the selected autocomple
 
 In Angular, an autocomplete `nimble-list-option` can be created with an `ngValue`. In that case the `ngModel` of the combobox will be the `ngValue` of the matched autocomplete option.
 If no matching autocomplete option is found, the `ngModel` of the combobox will be set to `OPTION_NOT_FOUND`, and if the application needs to access the user-entered text, that should be done
-through the `value` on the `nativeElement` associated with the `nimble-combobox`.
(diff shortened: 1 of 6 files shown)


atmgrifter00 (approved):

packages/nimble-components/src/combobox/tests/combobox.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+### Native element and Blazor
+
+The value of the combobox comes from the text content of the selected autocomplete `nimble-list-option`, or, if no matching autocomplete is found, the value is the user-entered text.
+
+### Angular
+
+In Angular, an autocomplete `nimble-list-option` can be created with an `ngValue`. In that case the `ngModel` of the combobox will be the `ngValue` of the matched autocomplete option.
+If no matching autocomplete option is found, the `ngModel` of the combobox will be set to `OPTION_NOT_FOUND`, and if the application needs to access the user-entered text, that should be done
+through the `value` on the `nativeElement` associated with the `nimble-combobox`.
```
rajsite, line 31: > that should be done through the `value` on the `nativeElement` associated with the `nimble-combobox`.

I don't think we should ever describe accessing the `nativeElement` as the expected workflow. Is that the case anywhere else? If accessing the value property is an expected workflow it should be exposed on the combobox directive.

Can we expose just a getter for this use case or are there valid use cases to independently set the value as well? I suspect we don't want to independently set the value as then the ngModel wouldn't be aligned. @atmgrifter00

atmgrifter00: Full disclosure @rajsite, I was fine with the current documentation in lieu of getting around the issue raised [here](https://github.com/ni/nimble/issues/682)(particularly regarding the second solution proposal).

I believe your suggestion is ultimately what makes the most sense (and is aligned with the second proposal). Do you think we just want to avoid mentioning `nativeElement` altogether? I believe what prompted this doc change is that we had a client who wanted access to the current text of the `Combobox`, but could find no docs on how to do so. This change is simply meant to offer a temporary relief of that confusion.

rajsite: > Do you think we just want to avoid mentioning nativeElement altogether?

Proper usage of our controls should not necessitate accessing `nativeElement` in Angular.

> and is aligned with the second proposal

The second proposal says to expose the underlying `value` for all controls. I don't think that's appropriate either (too easy to shoot yourself in the foot and screw up ngModel state). It's an exception just for this case. And I think the rework renders this moot as the ngModel value will ultimately just be the text value of the selected option.

So this is just a workaround till the rework. So we could either document it that way and link to bugs or we can expose just the get accessor for value (assuming that is the use case needed today)

mollykreis: I've added a `value` getter to the directive and updated these docs.

rajsite: FYI added that this PR fixes the issue meyer's referenced to the description https://github.com/ni/nimble/issues/682

rajsite (approved):

packages/nimble-components/src/combobox/tests/combobox.mdx
rajsite: Would be nice in a tech debt or something to clean-up SLE to use value

jattasNI (approved):

packages/nimble-components/src/combobox/tests/combobox.mdx
mollykreis: I've created a tech debt item in SLE for this and assigned it to myself. I'll take a look at it in a few days, when the new version of nimble should have made its way into apps.

PR #1435, opened 2023-08-16
Improved the prerendering of dies to bucket them based on colors

munteannatan:
# Pull Request

## 🤨 Rationale

The die info for rendering was first prepared and then sorted based on the color which costs us a lot of performance.

## 👩‍💻 Implementation

created color buckets to store the values of the dies and rendered them after one cluster after another

## 🧪 Testing

fixed failing tests expecting the previous behavior

observed performance improvements are the time to load has halved
before
after

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/wafer-map/modules/rendering.ts
+++ b/packages/nimble-components/src/wafer-map/modules/rendering.ts
@@ -106,8 +106,7 @@ export class RenderingModule {
             transformedCanvasMinPoint[1] -= dieHeight;
 
             Object.entries(this.wafermap.dataManager.renderInfo).forEach(
-                ([fillStyle, dies]) => {
-                    context.fillStyle = fillStyle;
+                ([_fillStyle, dies]) => {
                     for (const die of dies) {
                         if (
                             this.isDieVisible(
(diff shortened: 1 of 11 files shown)


packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -76,18 +61,22 @@ export class RenderingModule {
...
-                    die,
-                    transformedCanvasMinPoint,
-                    transformedCanvasMaxPoint
-                )
-            ) {
-                context.fillStyle = die.fillStyle;
-                context.fillRect(die.x, die.y, dieWidth, dieHeight);
+        Object.entries(this.wafermap.dataManager.renderInfo).forEach(
+            ([fillStyle, dies]) => {
+                context.fillStyle = fillStyle;
```
atmgrifter00, line 66: Just so I have a clear understanding of the root cause of the issue, the `context.fillStyle` set is what I presume is the bottleneck?

munteannatan: After the tests I did we will postpone this improvement and discuss with my team how to move forward.

I think we will revisit in a week or two

PR #1439, opened 2023-08-17
Fully import date-text column types before exporting them

m-akinc:
# Pull Request

## 🤨 Rationale

Previous attempt to fully export date-text column types didn't work because the types were being imported with `import type`.

## 👩‍💻 Implementation

`import` instead of `import type`

## 🧪 Testing

None

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
```diff
@@ -2,7 +2,7 @@ import { Directive, ElementRef, Input, Renderer2 } from '@angular/core';
 import { type TableColumnDateText, tableColumnDateTextTag } from '@ni/nimble-components/dist/esm/table-column/date-text';
 import { BooleanValueOrAttribute, NumberValueOrAttribute, toBooleanProperty, toNullableNumberProperty } from '@ni/nimble-angular/internal-utilities';
 import { NimbleTableColumnBaseDirective } from '@ni/nimble-angular/table-column';
-import type {
+import {
```
rajsite, line 5:
```suggestion
export {
```

just use the re-export syntax https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#re-exporting_aggregating

and delete all the single line exports added below

angular-workspace/projects/ni/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
```diff
@@ -2,7 +2,7 @@ import { Directive, ElementRef, Input, Renderer2 } from '@angular/core';
...
     DateTextFormat,
     DateStyle,
```
rajsite, line 7: consider testing locally with a packed build to avoid this cycle / validate the change

rajsite (approved):

angular-workspace/projects/ni/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
m-akinc: I can't because I'm actually using those types as well as re-exporting them.

PR #1440, opened 2023-08-18
Changed quadrant to originLocation and updated defaults

munteannatan:
# Pull Request

## 🤨 Rationale

As agreed in the [wafer map spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/wafer-map/specs/README.md), the 'quadrant' attribute was confusing to use, so we needed to change it to 'originLocation'.

## 👩‍💻 Implementation

Changed the name of the attribute and some of the logic. 
The new default value is bottomLeft, the default in the [cartesian coordinate system](https://en.wikipedia.org/wiki/Cartesian_coordinate_system).
small sketch with the origins

## 🧪 Testing

Existing tests were updated.

## ✅ Checklist


packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -135,7 +135,7 @@ const metadata: Meta<WaferMapArgs> = {
         highlightedValues: 'set1',
         maxCharacters: 4,
         orientation: WaferMapOrientation.left,
-        quadrant: WaferMapQuadrant.bottomLeft
+        originLocation: WaferMapOriginLocation.bottomLeft
```
atmgrifter00, line 138: Okay, so despite the behavior change in the attribute, it makes sense to keep this the same (as the default), since it matches the true default of the attribute (which it didn't before)?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -36,8 +36,10 @@ export class WaferMap extends FoundationElement {
      */
     public readonly waferMapUpdateTracker = new WaferMapUpdateTracker(this);
 
-    @attr
-    public quadrant: WaferMapQuadrant = WaferMapQuadrant.topLeft;
+    @attr({
```
atmgrifter00, line 39: I would try and put this all on one line. There are only a few places in our code that span multiple lines, but I don't think putting it on one line will break the build. There are instances that are longer than what you have here with no disabling of prettier.

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
munteannatan: yes, I think before we went with a top left origin similar to graphics rendering origin. I do not remember why the storybook was using this default though

change/@ni-nimble-components-35ae0106-c2e5-4744-adc3-d07e2cd543fa.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: Since this component is still incubating, all changes can be marked patch. (Should we move it out of incubating sometime? That can be a conversation for another day if you want!)
```suggestion
  "type": "patch",
```

jattasNI (approved):

change/@ni-nimble-components-35ae0106-c2e5-4744-adc3-d07e2cd543fa.json
munteannatan: I think there are still some items on the checklist before moving it out, and after we do these changes to the API

DStavilaNI (approved):

rajsite (approved):

PR #1444, opened 2023-08-18
Implement 'alignment' attribute on number-text column

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1011. It adds an `alignment` attribute to the `nimble-table-column-number-text` element that can be set to `right`, `left`, or `default` (defined in a new `NumberTextAlignment` enum). When `alignment` is set to `default`, the alignment is determined by the column's `format` property.

## 👩‍💻 Implementation

- Added `alignment` property to the base text cell view that defaults to `left` so that the base text cell view can control the alignment
- Added new `NumberTextAlignment` enum with values of `right`, `left`, and `default`
- Added `alignment` attribute & property to the `nimble-table-column-number-text`

## 🧪 Testing

- New unit tests
- Extended storybook story and manually verified the changes look correct
- Added additional matrix tests to cover each `format` option being configured with each `alignment` option

## ✅ Checklist


mollykreis:
@m-akinc, can you provide a buddy review for this PR?

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -267,4 +275,101 @@ describe('TableColumnNumberText', () => {
             expect(pageObject.getGroupHeaderTitle(0)).toBe('');
         });
     });
+
+    describe('alignment', () => {
```
m-akinc, line 279: nit: Test output may read better without this describe block. I don't think it's functionally necessary.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -86,6 +86,7 @@ export default metadata;
 interface TextColumnTableArgs extends SharedTableArgs {
```
m-akinc, line 86: nit: rename to `NumberTextColumnTableArgs`

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -105,6 +106,21 @@ const formatDescription = `Configures the way that the numeric value is formatte
...
+const alignmentDescription = `Configures the alignment of the value within the column.
+
+<details>
+    <summary>Default Alignment</summary>
+
+    The default alignment of the value depends on the column's format.
+    <ul>
+        <li>\`default\` format: Values are left-aligned by default.
+        </li>
+        <li>\`roundToInteger\` format: Values are right-aligned by default.
```
m-akinc, line 118: I think it's clearer without "by default".
```suggestion
        <li>\`default\` format: Values are left-aligned.
        </li>
        <li>\`roundToInteger\` format: Values are right-aligned.
```

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -8,14 +8,15 @@ import { TableColumnSortOperation } from '../base/types';
...
-import { NumberTextFormat } from './types';
+import { NumberTextAlignment, NumberTextFormat } from './types';
 import type { NumberFormatter } from './models/number-formatter';
 import { RoundToIntegerFormatter } from './models/round-to-integer-formatter';
 import { DefaultFormatter } from './models/default-formatter';
 
 export type TableColumnNumberTextCellRecord = TableNumberField<'value'>;
 export interface TableColumnNumberTextColumnConfig {
     formatter: NumberFormatter;
+    rightAlign: boolean;
```
m-akinc, line 19: Is there any situation that we would center-align? I guess if one comes up, we can change this to `alignment` at that point.

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -65,6 +74,22 @@ export class TableColumnNumberText extends TableColumnTextBase {
...
+        if (this.alignment === NumberTextAlignment.left) {
+            return false;
+        }
+
+        if (this.alignment === NumberTextAlignment.right) {
+            return true;
+        }
+
+        // Look at format to determine the default alignment
+        if (this.format === NumberTextFormat.roundToInteger) {
```
m-akinc, line 88: The thing I don't like about this is that if the data is all integral to begin with, it's unlikely a user is going to set format to `roundToInteger`, even though we probably would want to right-align in that case. I can't think of a better solution, though. Is there a strong argument not to right-align by default? When there are decimal places, it's unlikely we're going to get the decimal points to line up between rows, whether we left-align or right-align. What's the rationale for left-aligning non-integral values?

packages/nimble-components/src/table-column/number-text/types.ts
```diff
@@ -7,3 +7,15 @@ export const NumberTextFormat = {
 } as const;
```
m-akinc, line 7: This is out-of-scope for this PR, but it seems weird to me that `NumberTextFormat` has a `roundToInteger` value. I would think that would be orthogonal to other formatting options (and so would be its own separate boolean attribute). Do we have a sense for what other values `NumberTextFormat` might have in the future?

m-akinc (approved):

packages/nimble-components/src/table-column/number-text/types.ts
mollykreis: The currently approved API for the column can be found [here](https://github.com/ni/nimble/blob/886850629165bf624f46fb8a1f43c9ed0cb65133/packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md#api). However, the current direction we're leaning is to remove the `custom` format option since we've found that it is challenging (and in some cases impossible) to use `Intl.NumberFormat` alone to format a number the way we want.

The only other format in the spec right now is `decimal`, which adds a `decimal-digits` field to specify the number of decimals to show. However, `roundToInteger` is really the `decimal` format with `decimal-digits` set to `0`, so I've been starting to question the usefulness of having `roundToInteger` once `decimal` is implemented. I might end up proposing that we remove it.

I don't know if that answers your question, but I think if it doesn't, we should probably have a larger in-person discussion with the dev team to talk about some of the items in the approved API since we've started realizing that it doesn't make as much sense as we originally thought.

packages/nimble-components/src/table-column/number-text/index.ts
mollykreis: The idea is that we only want to right-align the number if we know that the decimal separators in all values within the column will align. Since that is unlikely with the `default` format (unless as you stated, they just happen to all be integers), the decision was to left-align by default.

A `decimal` format is also coming shortly, and it will `right-align` by default since those number will always display with the same number of decimals on each row.

packages/nimble-components/src/table-column/number-text/index.ts
mollykreis: I'll leave this conversation open so that others can weigh in if they have an opinion. I think that it makes sense for the property in `TableColumnNumberTextColumnConfig` to stay as `rightAlign`, but I think the bigger question is whether or not we want the alignment in `TableColumnTextCellViewBase` to be more flexible (i.e. an enum of `right`, `left`, `center` rather than a `boolean` indicating right vs left). The `TableColumnTextCellViewBase` will have a larger impact if we decide to change it in the future than internal config for this specific column.

jattasNI: I agree that number columns should only support right and left. I bet we'll never support center alignment there.

I could imagine other text columns wanting center alignment someday. So if it's simple to add to the base class now but might be complex to do later, then I'd lean towards adding it now.

jattasNI (approved):

change/@ni-nimble-components-883f2b11-5389-4299-84e4-80c6f0b1d2c1.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The next time Rick shows up to one of our team meetings, let's demo this feature for him. I recall there was a lot of debate about whether to also right align column headings and the decision was "lets start out not doing so and see how it looks" so now we have a chance to show him how it looks.

mollykreis: Sure, I can certainly demo it. I'm not sure we're quite in a place to see the full effect of alignment & the headers though because we don't have menus in the column headers yet.

packages/nimble-components/src/table-column/number-text/index.ts
mollykreis: My biggest concern about adding the enum was that I didn't want to add a mode that we didn't actually need & didn't have a way to have UI tests for. I'm not sure why I didn't think about it until now, but I'll go ahead and change the boolean to an enum but only support "right" and "left" for now. Then, we can wait to add support for "center" until we need it, but it won't be a breaking change that time.

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -50,9 +54,14 @@ export class TableColumnNumberText extends TableColumnTextBase {
...
 
+    private alignmentChanged(): void {
+        this.updateColumnConfig();
+    }
+
     private updateColumnConfig(): void {
         const columnConfig: TableColumnNumberTextColumnConfig = {
-            formatter: this.createFormatter()
+            formatter: this.createFormatter(),
+            rightAlign: this.shouldRightAlign()
```
rajsite, line 64: Can we pass through as `alignment` instead? We are effectively resolving from a numericcalignment to a base alignment, so could just pass through that resolved value since that's what's used  (instead of coming up with a third intermediate)

PR #1119, opened 2023-03-14
Add invalid styling and update focus styling for text area component

m-akinc:
# Pull Request

## 🤨 Rationale

See #486

This PR adds invalid styling for the text area **and** updates the general focus/hover styling to match the updated design doc, i.e. only color/thicken the bottom border rather than the entire border.

## 👩‍💻 Implementation

- Added two new attributes/properties: `error-visible` and `error-text`, following the pattern already set by other components
- Needed to fork the FAST template to add our error exclamation mark icon and the error text that appears underneath the text area. Forking was necessary because the FAST text area doesn't have any slots we could use.
- One challenge is that the exclamation icon needs to appear just inside the top right corner of the text area, and it needs to be be offset differently depending on whether there is a vertical scrollbar. A vertical scrollbar can appear or disappear when the user resizes the control (dragging the resize handle) or types in the control. I used an `input` event handler and a `ResizeObserver` to detect when the scrollbar visibility had changed.
- Updated the Storybook stories
- Updated the Angular and Blazor integration and tests
- Updated the non-error state focus/hover style to match the design doc. This meant only coloring/highlighting the bottom border rather than the full border. I did not implement the double-underline style from the design doc, because it is still unclear if that is a valid, distinct state.

Design doc:
image
(Note that I checked with Brandon to confirm that we **don't** want the weird spacing to the right of the exclamation mark as is shown in the design doc. He agrees with the spacing that I ultimately used instead.)
Old:
image
New:
image

## 🧪 Testing

Tested in Storybook.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1446, opened 2023-08-22
Rich Text Editor | Implementation of additional APIs

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

See the [API section of the spec](https://github.com/ni/nimble/tree/main/packages/nimble-components/src/rich-text-editor/specs#api).

## 👩‍💻 Implementation

* Implemented the following APIs:
    1. `disabled` - boolean attribute when enabled, set the Tiptap's editable state to false through [setEditable()](https://tiptap.dev/api/editor#set-editable) and set the `disabled` attribute of "nimble-toggle-button".
    2. `footer-hidden` - hide the footer through styles.
    4. `placeholder` - installed Tiptap's [Placeholder](https://tiptap.dev/api/extensions/placeholder) extension and configured the styles to show the placeholder text when the editor is empty.
    5. `empty` - read-only property that returns "true" if the editor is empty by getting the current "textContent" of the editor.
    6. `error-visible` - styled the editor with error color and added an exclamation icon to the editor same as in the "nimble-textarea". Used the "ResizeObserver" to update the scrollbar width and position of the error icon in the editor accordingly. 
    7. `error-text` - shows the text in the bottom left of the editor just like other nimble components.
* Bumped all the `@tiptap` extensions to the latest version.
 
## 🧪 Testing

* Added visual matrix tests to test the combinations of all the UI-related APIs.
* Added unit tests for the APIs which cannot be covered in visual tests.
* Manually verified the functionalities in the local storybook build.

## ✅ Checklist


packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -93,6 +167,39 @@ export class RichTextEditor extends FoundationElement {
...
+    /**
+     * @internal
+     */
+    public disabledChanged(): void {
+        if (this.tiptapEditor) {
+            this.tiptapEditor.setEditable(!this.disabled);
+            this.editor.setAttribute(
+                'aria-disabled',
+                this.disabled ? 'true' : 'false'
+            );
```
vikisekarNI, line 183: While it makes sense absolutely to set `aria-disabled` to `true` when `disabled` attribute changes, I am wondering whether custom slot contents will still be left enabled?

Do we also need to disable slot content in template then (or) remove setting this attribute? It will be good to get thoughts from Nimble team. Since this is a minor one, we can keep it open and get directions from Nimble team's expertise.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -93,6 +167,39 @@ export class RichTextEditor extends FoundationElement {
...
+     * Update the placeholder text and view of the editor.
+     * @internal
+     */
+    public placeholderChanged(): void {
+        if (this.tiptapEditor) {
+            const placeholderExtension = this.tiptapEditor.extensionManager.extensions.filter(
+                extension => {
+                    return extension.name === 'placeholder';
+                }
+            )[0] as Extension<PlaceholderOptions>;
```
vikisekarNI, line 197: Consider moving this into a separate function `getTipTapExtension(extensionName: string)`

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -146,6 +178,29 @@ export const styles = css`
...
+        ${
+            /**
+             * Minimum height is the combination of existing minimum height of the tiptap div, footer height and extra padding added to the tiptap div.
+             * With this calculation, the editor will extend to use the footer height when it is hidden.
+             *
+             * Use case: If the footer is initially hidden and is dynamically enabled when the editor is focused, there will be no layout shift,
+             * and the footer will smoothly appear within the editor.
+             */ ''
+        }
+        min-height: calc(var(--ni-private-rich-text-editor-editor-minimum-height) + var(--ni-private-rich-text-editor-footer-section-height) + 4px);
```
vikisekarNI, line 199: What is the additional `4 px` here? Buffer?

packages/nimble-components/src/rich-text-editor/template.ts
```diff
@@ -6,19 +6,36 @@ import { iconBoldBTag } from '../icons/bold-b';
...
+            <section
+                ${ref('editorContainer')}
+                class="editor-container"
+            >
             </section>
-            <section class="footer-section" part="footer-section">
-                <${toolbarTag}>
+            <${iconExclamationMarkTag}
+                severity="error"
+                class="error-icon ${x => (x.scrollbarWidth >= 0 ? 'scrollbar-width-calculated' : '')}"
```
vikisekarNI, line 25: `scrollbar-width-calculated` needs to be an attribute instead of a class which will conflict with user defined class names and for other reasons stated in [architectural philosophies in contributing document](https://github.com/ni/nimble/blob/51b3fa5759b35d653ad3866535c726399c89711b/packages/nimble-components/CONTRIBUTING.md#why-attributes-over-classes)

Consider having this as an attribute and make necessary selector changes in `styles.ts`

vikisekarNI: Suggestion - As per [attribute naming conventions](https://github.com/ni/nimble/blob/51b3fa5759b35d653ad3866535c726399c89711b/packages/nimble-components/CONTRIBUTING.md#attribute-common-name-patterns), consider naming it as `scrollbar-visible`. 

Open for alternative names.

packages/nimble-components/src/rich-text-editor/template.ts
```diff
@@ -6,19 +6,36 @@ import { iconBoldBTag } from '../icons/bold-b';
...
+                style="--ni-private-rich-text-editor-scrollbar-width: ${x => x.scrollbarWidth}px;"
+            ></${iconExclamationMarkTag}>
+            <section class="footer-section">
+                <${toolbarTag}
+                    ?disabled="${x => x.disabled}"
+                    aria-disabled="${x => x.disabled}"
```
vikisekarNI, line 31: Is `nimble-toolbar` handling this `aria-disabled` internally already?

vikisekarNI: Same comment applies for `nimble-toggle-button` as well. If it's already implemented we can stop with just setting `disabled`

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,6 +156,45 @@ export class RichTextEditorPageObject {
             .map(el => el.textContent || '');
     }
 
+    public isFooterVisible(): boolean {
+        const footerSection = this.richTextEditorElement.shadowRoot!.querySelector(
+            '.footer-section'
+        )!;
```
vikisekarNI, line 162: minor: Query selection part can go into a separate page object function `getFooter()` that can be reused in all higher order functions?

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,6 +156,45 @@ export class RichTextEditorPageObject {
...
+    }
+
+    public getFooterHeight(): string {
+        const footerSection = this.richTextEditorElement.shadowRoot!.querySelector(
+            '.footer-section'
+        )!;
+        return window.getComputedStyle(footerSection).height;
+    }
+
+    public async setFooterHiddenAttribute(): Promise<void> {
```
vikisekarNI, line 173:
```suggestion
    public async hideFooter(): Promise<void> {
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -47,7 +48,7 @@ describe('RichTextEditor', () => {
     it('should initialize Tiptap editor', () => {
         expect(pageObject.editorSectionHasChildNodes()).toBeTrue();
         expect(pageObject.getEditorSectionFirstElementChildClassName()).toBe(
```
vikisekarNI, line 50: minor: This should have been caught earlier. Can we remove any html structure details on the function name? As we will anyway update the page object function on updating the html structure or tip tap structure? 

```suggestion
        expect(pageObject.getEditableAreaClassName()).toBe(
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -1302,6 +1309,164 @@ describe('RichTextEditor', () => {
...
+        expect(element.empty).toBeTrue();
+
+        await pageObject.setEditorTextContent('       ');
+        expect(element.empty).toBeTrue();
+
+        element.setMarkdown('  ');
+        expect(element.empty).toBeTrue();
+    });
+
+    it('should reflect the "placeholder" value to its internal attribute', () => {
```
vikisekarNI, line 1465: Some more test cases needed? 
1. `empty` should be `true` even when `placeholder` content is set
2. `placeholder` should be cleared when `placeholder` attribute is set to `''` after setting `placeholder`

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -50,6 +58,16 @@ Note: The content in the \`footer-actions\` slot will not adjust based on the st
 client application to make any necessary adjustments. For example, if the buttons should be disabled when the rich-text-editor is disabled, the
 client application must implement that functionality.
 `;
+const fitToContentDescription = `Setting \`fit-to-content\` allows the editor to grow vertically to fit the content instead of enabling the
+vertical scrollbar when it reaches the certain height.
```
vikisekarNI, line 62:
```suggestion
vertical scrollbar when content reaches default height.
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -103,11 +127,47 @@ const metadata: Meta<RichTextEditorArgs> = {
...
+            description:
+                'Whether the editor should be styled to indicate that it is in an invalid state.'
+        },
+        errorText: {
+            description:
+                'A message to be displayed when the editor is in the invalid state explaining why the value is invalid.'
+        },
+        placeholder: {
+            description:
+                'Placeholder text to show it in the editor when it is empty.'
```
vikisekarNI, line 141:
```suggestion
                'Placeholder text to show when editor is empty.'
```

vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/index.ts
vivinkrishna-ni: Yes, the custom slot elements will not be disabled when we set the `disabled` attribute. It is the responsibility of the clients to disable the slot contents. It is the same in the [nimble-text-field](https://nimble.ni.dev/storybook/?path=/docs/components-text-area--docs) where the slot element will not be disabled based on the `disabled` attribute. Also, we clearly mentioned in the storybook docs the same way as the text field mentioned for slot elements.
image

As per the [MDN reference for aria-disabled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled), the custom `<div>` control used in the editor should set the `aria-disabled` value.
image

packages/nimble-components/src/rich-text-editor/styles.ts
vivinkrishna-ni: This `4px` should not come here. Previously, I added the minimum height of the editor section as `32px` but for a minimum one-line height, it should be `36px`. Font size of paragraph (18px) + Padding (top 8px + bottom 8px) + Border (top 1px + bottom 1px) = `36px`. 

Therefore, I removed the `4px` from the calculation.

packages/nimble-components/src/rich-text-editor/template.ts
vivinkrishna-ni: Thanks for the references, I have updated it to an attribute instead of a class and named the attribute as `scrollbar-visible` as suggested for the error icon.

packages/nimble-components/src/rich-text-editor/template.ts
vivinkrishna-ni: `nimble-toolbar` employs the FAST toolbar. It appears that the [template of the FAST toolbar](https://github.com/microsoft/fast/blob/6091f89a6a0b8bf3baa4cd4ed2aae1078ba9865b/packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts#L21C9-L36C10) does not contain `disabled` or `aria-disabled` attributes. However, I believe we should apply the disabled value to its child elements i.e. the toggle buttons. 

Also, the `nimble-toggle-button` internally designates the `aria-disabled` state as true when the `disabled` attribute is set. Therefore, I have also removed it from here.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
vivinkrishna-ni: Updated

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Written both tests as mentioned.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -50,6 +58,16 @@ Note: The content in the \`footer-actions\` slot will not adjust based on the st
 client application to make any necessary adjustments. For example, if the buttons should be disabled when the rich-text-editor is disabled, the
 client application must implement that functionality.
 `;
+const fitToContentDescription = `Setting \`fit-to-content\` allows the editor to grow vertically to fit the content instead of enabling the
+vertical scrollbar when content reaches the default height.
+
+To observe the changes when toggling, add more than five lines in the editor; this will enable the vertical scrollbar to view the hidden content.
+If the \`fit-to-content\` option is enabled, the editor will grow vertically to accommodate the content, instead of displaying the vertical scrollbar.`;
```
m-akinc, line 65:
```suggestion
const fitToContentDescription = `Setting \`fit-to-content\` allows the editor to grow vertically to fit the content instead of displaying a
vertical scrollbar when content exceeds the visible height.`;
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -50,6 +58,16 @@ Note: The content in the \`footer-actions\` slot will not adjust based on the st
...
+const inputEventDescription = `This event is fired when there is a change in the content of the editor.
+
+The client can utilize this event to execute various functionalities for every input to the editor. For instance, it can be utilized to assess
+whether the editor is empty by utilizing the \`empty\` read-only property with each input, and subsequently disabling a button if the editor is
+indeed empty.`;
```
m-akinc, line 70: I don't think we need to explain why events are useful.

```suggestion
const inputEventDescription = `This event is fired when there is a change in the content of the editor.`;
```

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -103,11 +127,46 @@ const metadata: Meta<RichTextEditorArgs> = {
...
+            description:
+                'Setting `footer-hidden` hides the footer section which consists of all formatting option buttons and the `footer-actions` slot content.'
+        },
+        fitToContent: {
+            description: fitToContentDescription
+        },
+        empty: {
+            name: 'empty',
+            description:
+                'Read-only boolean value. Returns true if editor is empty.',
```
m-akinc, line 152: We should explicitly say this includes when the editor has only whitespace (assuming that's the behavior we want).

packages/nimble-components/src/rich-text-editor/template.ts
```diff
@@ -6,19 +6,30 @@ import { iconBoldBTag } from '../icons/bold-b';
...
 import { iconListTag } from '../icons/list';
 import { iconNumberListTag } from '../icons/number-list';
+import { errorTextTemplate } from '../patterns/error/template';
+import { iconExclamationMarkTag } from '../icons/exclamation-mark';
 
 // prettier-ignore
 export const template = html<RichTextEditor>`
-    <template>
+    <template
+        aria-disabled="${x => x.disabled}"
```
m-akinc, line 15: I don't think we need to set `aria-disabled` on the host element. It is not the element with the accessibility role -- that's the tiptap editor element.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -294,6 +402,8 @@ export class RichTextEditor extends FoundationElement {
         this.editor.className = 'editor';
         this.editor.setAttribute('aria-multiline', 'true');
         this.editor.setAttribute('role', 'textbox');
+        this.editor.setAttribute('aria-disabled', 'false');
+        this.editor.setAttribute('aria-label', 'Rich Text Editor');
```
m-akinc, line 406: We can't just set `aria-label` to an unlocalized constant string. We should [apply the `ARIAGlobalStatesAndProperties` mixin](https://github.com/ni/nimble/blob/12beb26dae4b4151fa3d06a1e6f0ec8fabd52b19/packages/nimble-components/src/drawer/index.ts#L136) and then [reflect the `ariaLabel` property value to the inner editor element's `aria-label` attribute](https://github.com/ni/nimble/blob/12beb26dae4b4151fa3d06a1e6f0ec8fabd52b19/packages/nimble-components/src/drawer/template.ts#L7).

m-akinc (approved):

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -33,7 +36,68 @@ declare global {
...
  */
-export class RichTextEditor extends FoundationElement {
+export class RichTextEditor extends FoundationElement implements ErrorPattern {
+    /**
+     * Whether to disable user from editing and interacting with toolbar buttons
+     *
+     * @public
+     * HTML Attribute: disabled
+     */
+    @attr({ attribute: 'disabled', mode: 'boolean' })
```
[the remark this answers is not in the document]
vivinkrishna-ni: Removed for both disabled and placeholder attributes.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -33,7 +36,68 @@ declare global {
...
+    @attr({ attribute: 'placeholder' })
+    public placeholder = '';
+
+    /**
+     * True if the editor is empty, false otherwise
+     *
+     * @public
+     */
+    public get empty(): boolean {
+        return this.tiptapEditor.state.doc.textContent.trim().length === 0;
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for pointing this out. Tiptap's [isEmpty](https://tiptap.dev/api/editor#is-empty) does not return `true` when the editor contains only whitespace. But as per the comments feature requirement, we need to disable the `OK` button when the editor is empty or contains only whitespace.

I have updated the spec accordingly with the reasoning.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -85,6 +158,9 @@ export class RichTextEditor extends FoundationElement {
             this.editorContainer.append(this.editor);
         }
         this.bindEditorTransactionEvent();
+        this.resizeObserver = new ResizeObserver(() => this.onResize());
+        this.resizeObserver.observe(this);
+        this.bindEditorUpdateEvent();
```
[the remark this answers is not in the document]
vivinkrishna-ni: The width is calculated in the `onResize` method during the initialization. Even in the text area, we didn't use `queueUpdateScrollbarWidth()` in the `connectedCallback()`.

 I verified the component in the storybook manually during the initialization (with and without the scrollbar) and it seems the icon was placed in the right position as expected. Please let me know if am I missing some cases here.

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -64,6 +79,23 @@ export const styles = css`
         width: 100%;
     }
 
+    :host([disabled]) .container,
+    :host([disabled]) .container::after {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the catch. I have fixed this issue.

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -135,17 +171,53 @@ export const styles = css`
...
         margin-block: 0;
     }
 
+    .ProseMirror p.is-editor-empty:first-child::before {
+        color: ${controlLabelFontColor};
+        content: attr(data-placeholder);
+        float: left;
+        height: 0;
+        pointer-events: none;
+        word-break: break-all;
```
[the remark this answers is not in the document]
vivinkrishna-ni: Done

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -164,4 +236,15 @@ export const styles = css`
         gap: ${standardPadding};
         place-items: center;
     }
+
+    :host([error-visible]) .error-icon {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, it makes sense not to have the `scrollbar-visible` so removed this condition from here and also removed setting the display from the styles.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,10 +156,50 @@ export class RichTextEditorPageObject {
...
+        this.richTextEditorElement.setAttribute('footer-hidden', '');
+        await waitForUpdatesAsync();
+    }
+
+    public async setDisabledState(): Promise<void> {
+        this.richTextEditorElement.setAttribute('disabled', '');
+        await waitForUpdatesAsync();
+    }
+
+    public hasButtonDisabled(buttonIndex: ToolbarButton): boolean {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Done

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,10 +156,50 @@ export class RichTextEditorPageObject {
...
+        const button = this.getFormattingButton(buttonIndex)!;
+        return button.hasAttribute('disabled');
+    }
+
+    public getDataPlaceholderValue(): string {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Done

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -99,22 +182,33 @@ const multipleSubPointsContent = `
...
         );
 };
+
+export const fitToContentTest: StoryFn = createStory(html`
+    ${createMatrix(componentFitToContent, [
+        fitToContentStates,
+        footerHiddenStates,
+        [
+            ['No width', ''],
+            ['Width 360px', 'width: 360px']
```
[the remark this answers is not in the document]
vivinkrishna-ni: Removed width case from the test and added a static width instead.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -53,12 +106,21 @@ const editorSizingTestCase = (
     </div>
 `;
 
-export const richTextEditorThemeMatrix: StoryFn = createMatrixThemeStory(
-    createMatrix(component)
-);
+const stateMatrix = createMatrix(component, [
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have made the test cases as suggested. For the empty content story test alone, I just vary both the `disabledValues` and `placeholderValueStates` as to see the placeholder text color in a disabled state.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -1302,6 +1309,170 @@ describe('RichTextEditor', () => {
...
+            for (const value of formattingButtons) {
+                const specType = getSpecTypeByNamedList(
+                    value,
+                    focused,
+                    disabled
+                );
+                // eslint-disable-next-line @typescript-eslint/no-loop-func
+                specType(
+                    `for "${value.name}" button`,
+                    // eslint-disable-next-line @typescript-eslint/no-loop-func
```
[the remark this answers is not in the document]
vivinkrishna-ni: Removed from here and in other test cases.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -33,7 +36,68 @@ declare global {
...
+     */
+    @attr({ attribute: 'placeholder' })
+    public placeholder = '';
+
+    /**
+     * True if the editor is empty, false otherwise
+     *
+     * @public
+     */
+    public get empty(): boolean {
```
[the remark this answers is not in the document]
vivinkrishna-ni: In order to create a comment in the comments feature, we will disable the OK button when the editor is either empty or contains only whitespace. This is necessary because the backend service does not permit the creation of a comment with just the whitespace. Therefore, it is essential to prevent users from creating a comment unless there is valid content within the editor. I have also updated the spec with the reasoning.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
vivinkrishna-ni: Sure, updated it as per the suggestion.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
vivinkrishna-ni: Done

packages/nimble-components/src/rich-text-editor/template.ts
vivinkrishna-ni: Thanks for the explanation. Removed `aria-disabled` on the host element.

packages/nimble-components/src/rich-text-editor/index.ts
vivinkrishna-ni: Thanks for the references, I have updated the code by applying the mixin and reflecting the `aria-label` value to the inner element. As we are not directly binding the value to the template, I used `ariaLabelChanged` to update the attribute value.

packages/nimble-components/src/rich-text-editor/template.ts
vivinkrishna-ni: There is no need to utilize the `scrollbar-visible` condition, as the icon positions itself precisely where we intended. I verified that the UI glitch shown in [this comment for text area](https://github.com/ni/nimble/pull/1119#discussion_r1154889610) is not reproducible for the editor. So I removed this condition from here.

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -135,17 +171,53 @@ export const styles = css`
...
+        ${
+            /**
+             * Minimum height is the addition of existing minimum height of the tiptap editor div and the footer section height.
+             * With this calculation, the editor will extend to use the footer height when it is hidden.
+             *
+             * Use case: If the footer is initially hidden and is dynamically enabled when the editor is focused, there will be no layout shift,
+             * and the footer will smoothly appear within the editor.
+             */ ''
+        }
+        min-height: calc(var(--ni-private-rich-text-editor-tiptap-editor-minimum-height) + var(--ni-private-rich-text-editor-footer-section-height));
```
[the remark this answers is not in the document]
vivinkrishna-ni: Setting the minimum height in the `:host` will not consider having the border. The behavior will be something like the below when enabling and disabling the `footer-hidden` as the border is set to the container and not the `:host`.
footer-hidden-state

In the comments feature, the initial state of the editor should be without the footer and once the editor is focused, the footer will be enabled. Therefore, we did not want to have any layout shift when enabling or disabling the `footer-hidden`.
footer-hidden-expected-state

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -135,17 +171,53 @@ export const styles = css`
...
         border: ${borderWidth} solid transparent;
         border-top-color: rgba(${borderRgbPartialColor}, 0.1);
         height: var(--ni-private-rich-text-editor-footer-section-height);
         overflow: hidden;
+        visibility: visible;
+    }
+
+    :host([footer-hidden]) .footer-section {
+        visibility: hidden;
+        height: 0px;
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, we can set `display: none;` instead of visibility and height. However, while doing so, we have to also add the border width (top + bottom = 2px) of the footer section to the `.ProseMirror` div to hide the footer without any layout shift in the UI as mentioned in the [other comment](https://github.com/ni/nimble/pull/1446#discussion_r1302893177).

packages/nimble-components/src/rich-text-editor/index.ts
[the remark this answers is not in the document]
m-akinc: We should be making the API decision based on what makes the most sense in the general sense, not because that's what the comments feature needs. But in this case, since you could probably argue for either behavior, I think it's fine to go with that.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -33,7 +41,71 @@ declare global {
...
+    @attr
+    public placeholder = '';
+
+    /**
+     * True if the editor is empty or contains only whitespace, false otherwise.
+     *
+     * @public
+     */
+    public get empty(): boolean {
+        // Tiptap [isEmpty](https://tiptap.dev/api/editor#is-empty) returns true even if the editor has white spaces.
```
m-akinc, line 103:
```suggestion
        // Tiptap [isEmpty](https://tiptap.dev/api/editor#is-empty) returns false even if the editor has only whitespace.
```

packages/nimble-components/src/rich-text-editor/styles.ts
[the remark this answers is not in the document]
m-akinc: I think you could avoid that by setting the footer's `box-sizing: border-box` so that the height includes the border, but adding the extra border height seems fine, too.

jattasNI:
Sorry for the delayed response on this: NITech has significantly reduced our owners' bandwidth this week. I will do my thorough review on Friday.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -50,6 +58,8 @@ Note: The content in the \`footer-actions\` slot will not adjust based on the st
 client application to make any necessary adjustments. For example, if the buttons should be disabled when the rich-text-editor is disabled, the
 client application must implement that functionality.
 `;
+const fitToContentDescription = `Setting \`fit-to-content\` allows the editor to grow vertically to fit the content instead of displaying a
```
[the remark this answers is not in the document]
vivinkrishna-ni: If `fit-to-content` is false, the editor will grow to a certain height i.e. `max-height` of the editor text area is set to `132px` and will enable the vertical scrollbar if the content crosses that height. This height we got from the [current visual design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=2488%3A82463&mode=dev) we have for the rich text editor.
fit-to-content-false

The height is subject to change if the visual design for the rich text editor is changed and finalized. Please let us know if the editor can have a default height like the text area component and based on the boolean attribute it can grow without enabling the scrollbar. 

Related discussion in the spec PR for the origin of this property: https://github.com/ni/nimble/pull/1289#discussion_r1235999977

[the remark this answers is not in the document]
jattasNI: Ah, cool, I guess I just needed to add another line to see that behavior!

The behavior I was expecting:
1. When `fit-to-content` is `false`, typing in the editor never impacts its rendered height. The rendered height is always determined by normal CSS rules set by the parent like `height` and `min-height` and `max-height`. It almost behaves this way now, except for when it's smaller than `max-height` and new input currently causes it to grow until it hits max height (what's shown in your video).
2. When `fit-to-content is `true`, typing in the editor can affect the component's height but it's constrained by whatever the client set for `min-height` and `max-height`. The current behavior I'm seeing is that it will grow and shrink the editor successfully but it isn't constrained by the min and max height. 
 
I'm also seeing strange behavior where the `nimble-rich-text-editor` element respects min-height and max-height but the rendered content grows beyond the bounds of the nimble element. In this screenshot the max-height is set to 300px and the element (blue box) respects that but its contents don't.
image

This suggestion is open to debate but I think it is closer to what Milan was suggesting in the spec discussion you linked. We want a way to completely turn off the behavior where the element sets its own size. And we want the element to respect the CSS it's configured with as much as possible.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -1302,6 +1304,160 @@ describe('RichTextEditor', () => {
...
+    it('should fire "input" event when there is an input to the editor', async () => {
+        const inputEventListener = createEventListener(element, 'input');
+
+        await pageObject.setEditorTextContent('input');
+        await inputEventListener.promise;
+
+        expect(inputEventListener.spy).toHaveBeenCalledTimes(1);
+    });
+
+    it('should fire "input" event when setting the content through "setMarkdown"', async () => {
```
jattasNI, line 1384: Is there a concrete need to fire the "input" event when the content is set programmatically? As a general rule we've tried to only fire events on interactive input, not programmatic. Usually applications are aware of the programmatic changes (since they're the ones making them) so they don't need to be notified of them. And often apps don't want to be notified of them, since it can cause unintended side effects.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the detailed clarification. I have updated the styles to the expected behavior you have stated above. I have updated the min and max height set by the client and verified if it reflects it to the internal control div to make the scrollbar visible irrespective of the `fit-to-content` state.

I also updated the `Fit to content` visual test to verify if the editor is behaving as expected for the given heights.

Also, I updated the description in the storybook.

packages/nimble-components/src/rich-text-editor/template.ts
```diff
@@ -6,19 +6,27 @@ import { iconBoldBTag } from '../icons/bold-b';
...
 export const template = html<RichTextEditor>`
     <template>
         <div class="container">
             <section ${ref('editorContainer')} class="editor-container">
             </section>
-            <section class="footer-section" part="footer-section">
+            <${iconExclamationMarkTag}
+                severity="error"
+                class="error-icon"
+                style="--ni-private-rich-text-editor-scrollbar-width: ${x => x.scrollbarWidth}px;"
```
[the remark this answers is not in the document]
vivinkrishna-ni: In rich text editor, the scrollbar width is calculated during the initialization without any delays. We are updating or queuing to update the scrollbar width in `update` event and also during the `connectedCallback`. So when I toggle on and off the `error-visble` it renders the error icon without any issue.
scrollbar-width-test


However, in the text area, the icon rendered before the calculation and once the scrollbar calculated, causes some shift in the position. Rationale for implementing `scrollbar-width-calculated` logic in the text area for your reference: https://github.com/ni/nimble/pull/1119#discussion_r1154889610

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -54,11 +107,34 @@ const editorSizingTestCase = (
...
+    createMatrix(component, [
+        [disabledStates[0]],
+        [footerHiddenStates[0]],
+        errorStates,
+        [placeholderValueStates[0]]
+    ])
+);
+richTextEditorThemeMatrixForLongContent.play = longTextPlayFunction;
+
+export const richTextEditorThemeMatrixForEmptyContent: StoryFn = createMatrixThemeStory(
```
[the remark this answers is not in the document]
vivinkrishna-ni: I was referring to the `nimble-text-area` for the theme matrix test cases. All the states were implemented in the same theme matrix test.

One constraint we see in the story matrix test, we couldn't find a straightaway approach to load different contents in the same story as we set the content using the `setMarkdown` method. So for instance, to write a test for `placeholder` there should be no content whereas, on the other hand, to test if the error icon is placed in the right position, there should be lengthy content to enable the scrollbar and check. 

Therefore, we just came up with the below order to cover all the states in all combinations. I have updated the name of the tests to provide an idea of what behavior is intended to be tested in the particular tests.
1.  Rich Text Editor Theme Matrix - This test covers the basic rich text formats (text and button state) in combinations of `disabled`, `error`, and `footer-hidden` states.
2. Error state theme matrix with lengthy content - This test covers, where the error icon is placed when the vertical scrollbar is enabled.
3. Placeholder state theme matrix - This is to test the placeholder text in an empty and in a `disabled` state.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: It makes sense not to fire the `input` event when they set content programmatically. So I removed the firing input event when setting the markdown content programmatically.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,8 +495,100 @@ export class RichTextEditor extends FoundationElement {
...
+        this.updateEditorHeightQueued = false;
+        this.updateEditorMinMaxHeight();
+    }
+
+    private onResize(): void {
+        this.scrollbarWidth = this.editor.offsetWidth - this.editor.clientWidth;
+        this.updateEditorMinMaxHeight();
+    }
+
+    private updateEditorMinMaxHeight(): void {
```
jattasNI, line 551: We really try hard to avoid encoding style values and doing style calculations in TS code. The CSS layout engine is optimized to do this for us so we'd rather rely on it than on JS/TS code, which is slower and less easy to read. I would like us to try to encode this logic in CSS rules rather than calling `getComputedStyle` and doing pixel calculations.

Another concern is about who controls the min and max height of the editor. I was expecting that to be something provided by clients by setting CSS rules on the text editor but I see some hardcoded values here that indicate it's intrinsic to the editor.

I don't know enough about the constraints that TipTap applies to offer more concrete suggestions or to know if these suggestions are possible. If you'd like to brainstorm more, we could set up a call with me and @m-akinc (I'm free on Tuesday after the UI WG). Or if there's time pressure for this PR we could consider removing the sizing code for now and splitting it into a follow up PR.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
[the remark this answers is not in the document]
jattasNI: Thanks, I hadn't thought about the constraint of calling `setMarkdown()` from matrix stories. I agree that complicates things. I'm good with the current approach.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -93,6 +196,46 @@ export class RichTextEditor extends FoundationElement {
...
+            this.queueUpdateScrollbarWidth();
+            this.queueUpdateEditorHeight();
+        }
+    }
+
+    /**
+     * @internal
+     */
+    public ariaLabelChanged(): void {
+        this.editor.setAttribute('aria-label', this.ariaLabel ?? '');
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated as per the suggestion.

packages/nimble-components/src/rich-text-editor/index.ts
vivinkrishna-ni: I have removed the entire logic from the TS code and modified the CSS styles to achieve what we wanted. Please review the new implementation and let us know your feedback.

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -21,18 +28,27 @@ export const styles = css`
...
+        height: 82px;
         --ni-private-rich-text-editor-footer-section-height: 40px;
         ${
             /** Minimum width is added to accommodate all the possible buttons in the toolbar and to support the mobile width. */ ''
         }
         min-width: 360px;
     }
 
+    :host([fit-to-content]) {
+        max-height: inherit;
```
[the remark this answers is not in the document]
rajsite: I have a feeling based on this feedback and the discussions in the other threads that the style updates needed for fit-to-content may require a few more rounds of feedback.

You may want to consider splitting it into a separate PR but totally up to you.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,10 +156,45 @@ export class RichTextEditorPageObject {
...
+
+    public async hideFooter(): Promise<void> {
+        this.richTextEditorElement.setAttribute('footer-hidden', '');
+        await waitForUpdatesAsync();
+    }
+
+    public async setDisabledState(): Promise<void> {
+        this.richTextEditorElement.setAttribute('disabled', '');
+        await waitForUpdatesAsync();
+    }
```
rajsite, line 176: Try to keep thee api aligned when wrapping public apis in helpers. so `setFooterHidden(footerHidden: boolean)` and `setDisabled(disabled: boolean)`. I'd also recommend on this making them configurable as suggested in the example signature

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,10 +156,45 @@ export class RichTextEditorPageObject {
...
+
+    public isButtonDisabled(buttonIndex: ToolbarButton): boolean {
```
rajsite, line 178: Can you clean-up all the references to "index"? We aren't passing an index, we are passing a ToolbarButton.
So name the parameter `index` and can you clean-up the docs for all the other button helpers like `getButtonCheckedState`, etc to remove discussion of indexes?

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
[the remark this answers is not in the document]
jattasNI: The default you picked seems fine to me. I'm good with pretty much any default height for the component as long as clients can set their preferred height. (Just a note: it's no longer a one-line default height if you set `footer-hidden`. That's totally fine, just pointing out the futility of trying to pick a "perfect" default 😀)

I noticed one more unexpected behavior in my testing: when `fit-to-content` is `false`, adding a new line that triggers the vertical scrollbar also causes the editor to grow its height by a couple pixels. Deleting the newline to hide the scrollbar causes it to shrink again. I would expect its height never to change in response to text changes.

https://github.com/ni/nimble/assets/10500124/f592f577-d726-47b5-8e94-ac12987cb31e

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -103,11 +119,47 @@ const metadata: Meta<RichTextEditorArgs> = {
...
+        empty: {
+            name: 'empty',
+            description:
+                'Read-only boolean value. Returns true if editor is either empty or contains only whitespace.',
+            control: false
+        },
+        input: {
+            name: 'input',
+            description:
+                'This event is fired when there is a change in the content of the editor.',
```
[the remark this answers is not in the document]
jattasNI: Looks great!

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,8 +481,83 @@ export class RichTextEditor extends FoundationElement {
...
+    }
+
+    /**
+     * Stopping the native input event propagation emitted by the contenteditable element in the Tiptap
+     * since there is an issue(linked below) in ProseMirror where selecting the text and removing them
+     * does not trigger the native HTMLElement input event. So using the "update" event emitted by the
+     * Tiptap to capture it as an "input" customEvent in the Rich text editor.
+     *
+     * Prose mirror issue: https://discuss.prosemirror.net/t/how-to-handle-select-backspace-delete-cut-type-kind-of-events-handletextinput-or-handledomevents-input-doesnt-help/4844
+     */
```
jattasNI, line 508: nit: minor copy editing
```suggestion
    /**
     * Stopping the native input event propagation emitted by the contenteditable element in the Tiptap
     * since there is an issue (linked below) in ProseMirror where selecting the text and removing it
     * does not trigger the native HTMLElement input event. So using the "update" event emitted by the
     * Tiptap to capture it as an "input" customEvent in the rich text editor.
     *
     * Prose Mirror issue: https://discuss.prosemirror.net/t/how-to-handle-select-backspace-delete-cut-type-kind-of-events-handletextinput-or-handledomevents-input-doesnt-help/4844
     */
```

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -93,6 +178,50 @@ export class RichTextEditor extends FoundationElement {
...
+
+    /**
+     * @internal
+     */
+    public ariaLabelChanged(): void {
+        if (this.ariaLabel) {
+            this.editor.setAttribute('aria-label', this.ariaLabel);
+        } else {
+            this.editor.removeAttribute('aria-label');
+        }
```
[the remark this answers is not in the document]
jattasNI: It'd be good to make sure we have test coverage to validate the cases mentioned in that linked discussion. See the icon for a list of the relevant tests: 
https://github.com/ni/nimble/blob/5daaa35c1a0f024e1f9cd08bcf9ba96ebd88ca35/packages/nimble-components/src/icon-base/tests/icons.spec.ts#L98C1-L98C1

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
```diff
@@ -50,15 +59,24 @@ Note: The content in the \`footer-actions\` slot will not adjust based on the st
 client application to make any necessary adjustments. For example, if the buttons should be disabled when the rich-text-editor is disabled, the
 client application must implement that functionality.
 `;
+const fitToContentDescription = `Setting \`fit-to-content\` allows the editor to grow vertically to fit the content instead of displaying a
+vertical scrollbar when content exceeds the visible height.
+
+However, \`fit-to-content\` will not take effect if the \`max-height\` of the element is set to the element i.e. vertical scrollbar will be enabled if it
+exceeds the specified maximum height.`;
```
jattasNI, line 66: nit: wording suggestion
```suggestion
When \`fit-to-content\` is set the component will only grow to its \`max-height\`. A vertical scrollbar will be enabled when the content it exceeds the specified maximum height.`;
```

jattasNI (commented):
Getting very close!

packages/nimble-components/src/rich-text-editor/index.ts
jattasNI: Other than the feedback Milan mentioned about `inherit` and the resize jitter I mentioned in the other thread, I think the new approach is looking nice.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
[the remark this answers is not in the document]
vivinkrishna-ni: I have fixed the above issue.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -93,6 +178,50 @@ export class RichTextEditor extends FoundationElement {
...
+            );
+        }
+    }
+
+    /**
+     * Update the placeholder text and view of the editor.
+     * @internal
+     */
+    public placeholderChanged(): void {
+        if (this.tiptapEditor) {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Removed this check

packages/nimble-components/src/rich-text-editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the references. I have updated the condition as linked above and wrote test cases similar to the ones written for icons.

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -139,15 +177,33 @@ export const styles = css`
         margin-block: 0;
     }
 
+    .ProseMirror p.is-editor-empty:first-child::before {
+        color: ${controlLabelFontColor};
+        content: attr(data-placeholder);
```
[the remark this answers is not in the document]
vivinkrishna-ni: Apologies for missing adding the doc link to the above code. I have added the link to the comment above this class.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
vivinkrishna-ni: I have updated these methods as configurable as suggested.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
vivinkrishna-ni: Did you mean to name the parament as `button: ToolbarButton` which is passed as an argument to these methods? As I could see in the other comment name it as a `button`: https://github.com/ni/nimble/pull/1424#discussion_r1300301659

I have removed the comments on discussing about the indexes in other helper methods.

rajsite: Sorry, yep, name the parameter `button` 👍

packages/nimble-components/src/rich-text-editor/styles.ts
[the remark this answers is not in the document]
vivinkrishna-ni: @rajsite @jattasNI Due to a time constraint to integrate this component into SLE, which relies entirely on this PR and the angular support PR, we have decided to branch out the `fit-to-content` related changes into a separate PR as suggested.

We will address all the comments provided for `fit-to-content` in the follow-up PR which will be created right after we come up with a better implementation approach and after this PR goes in. Therefore, I have removed all the changes related to `fit-to-content` and other style changes related to `inherit` from this PR including storybook docs, storybook tests, etc,..

Please note that I will also be removing the `fit-to-content` related changes from the Angular PR (https://github.com/ni/nimble/pull/1450) which is already approved by every reviewer and will be added along with the follow-up PR.

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.stories.ts
vivinkrishna-ni: I have updated as suggested but the `fit-to-content` will be moved to a different PR.

packages/nimble-components/src/rich-text-editor/index.ts
vivinkrishna-ni: Thanks for the feedback. As I mentioned in the other comment, we will be moving this to a follow-up PR. Therefore, I am closing this comment.

mollykreis:
I decline to review.

atmgrifter00:
I decline to review.

jattasNI (commented):
I played around in storybook after the removal of fit-to-content and it's looking good to me. I'm also OK with the plans described in the open threads. Just want to do one more check once those threads have been resolved.

packages/nimble-components/src/rich-text-editor/template.ts
```diff
@@ -6,19 +6,28 @@ import { iconBoldBTag } from '../icons/bold-b';
...
+                ?scrollbar-width-calculated="${x => x.scrollbarWidth >= 0}"
```
rajsite, line 22: Don't write to attributes that don't actually exist (the public api for icons do not define a scrollbarWidthCalculated property nor should they).

Instead follow the pattern of text-area and set a class when scrollbar width is defined:
https://github.com/ni/nimble/blob/db5ebd42c334810bb29be9c86cee763da487b39b/packages/nimble-components/src/text-area/template.ts#L61

packages/nimble-components/src/rich-text-editor/styles.ts
```diff
@@ -164,4 +218,15 @@ export const styles = css`
         gap: ${standardPadding};
         place-items: center;
     }
+
+    :host([error-visible]) .error-icon {
+        display: none;
+    }
+
+    :host([error-visible]) .error-icon[scrollbar-width-calculated] {
```
rajsite, line 226:
```suggestion
    :host([error-visible]) .error-icon.scrollbar-width-calculated {
```

rajsite (approved):

packages/nimble-components/src/rich-text-editor/template.ts
rajsite: Properties and attributes should be defined by a component itself, the css classes should be owned by component users. The rich text editor is a user of those icon components

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -99,22 +168,21 @@ const multipleSubPointsContent = `
                         1. Sub point 9`;
 
 export const multipleSubPointsContentInMobileWidth: StoryFn = createStory(mobileWidthComponent);
```
jattasNI, line 170: I approved many of the Chromatic diffs but I noticed that the ["multiple sub points content in mobile width" diff](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=64eff0a166bacb2e89143cf7) looks suspicious. Since the default height is now smaller, the diff no longer shows all the sub-points so the test isn't covering what it used to. I think it'd be reasonable to explicitly set the height for this story (and perhaps many of these tests) to ensure they continue testing what we want them to even if we change our minds about the default height.

Feel free to approve the diffs yourself to complete the PR once this is fixed.

jattasNI (approved):
Looks good once you fix the Storybook change and approve the Chromatic diffs. No need to get approval from the remaining reviewers. Congrats!

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
vivinkrishna-ni: Thanks! I have updated the height for all the mobile width content test cases and will approve after taking a final look at the Chromatic diffs.

PR #1452, opened 2023-08-24
Blazor support for icon table column

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1013

## 👩‍💻 Implementation

- Blazor components for nimble-table-column-icon, nimble-mapping-icon, and nimble-mapping-spinner
- Tests for above components, including one acceptance test for icon column
- Added icon column to example app
- Created shared base type for both icon column and enum-text column
- Updated Storybook docs for icon column and enum-text column to explain workaround to issue in Blazor where the key value of a mapping element isn't property treated as a string unless you wrap it like this: `@("foo")`.

## 🧪 Testing

Automated tests pass and example app works correctly.

## ✅ Checklist


m-akinc:

packages/nimble-blazor/NimbleBlazor/Components/NimbleMappingIcon.razor.cs
```diff
@@ -0,0 +1,36 @@
...
+    public TKey? Key { get; set; }
+
+    /// <summary>
+    /// Gets or sets the text used for the tooltip and accessible name of the icon.
+    /// </summary>
+    [Parameter]
+    public string? Text { get; set; }
+
+    /// <summary>
+    /// Gets or sets the (element name of the) Nimble icon being mapped to.
```
mollykreis, line 20: We don't have a good way to refer to these element names in Blazor, do we? And the name that is required isn't even the Blazor component name that a Blazor developer would be familiar with.

It seems like we either need some sort of docs telling Blazor developers how to reference the correct icon name, or more likely, we need a tech debt item to export icon names as constants in Blazor.

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnIcon.razor.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace NimbleBlazor;
+
+public partial class NimbleTableColumnIcon<TKey> : NimbleTableColumnMapping<TKey>, IFractionalWidthColumn
+{
+    /// <summary>
+    /// The fractional/proportional width to use for this column
+    /// </summary>
+    [Parameter]
+    public double FractionalWidth { get; set; } = 1;
```
mollykreis, line 11: I'm guessing that the reason why `FractionalWidth` and `MinPixelWidth` are defined here rather than in `NimbleTableColumnMapping` is that they are the most likely to deviate between the different mapping column instances even though they currently exist on both mapping column types?

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.stories.ts
```diff
@@ -40,7 +40,10 @@ const simpleData = [
 
 const enumTextColumnDescription = `The \`nimble-table-column-enum-text\` column renders string, number, or boolean values as mapped text in the \`nimble-table\`.
 
-When sorting or grouping the column, the raw data values are used, not the mapped text being displayed.`;
+When sorting or grouping the column, the raw data values are used, not the mapped text being displayed.
+
+**Blazor**: When setting a child mapping elment's \`Key\` value to a string, you must wrap it in \`@()\` so that the compiler treats it as a string, e.g.
```
mollykreis, line 45: Can we move the docs for the enum-text and icon columns into mdx files so that we can put the Blazor-specific usage notes in a "Usage Docs" section like we do for other components (e.g. the Angular-specific usage notes for the anchor column).

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnIcon.razor.cs
m-akinc: That's correct.

packages/nimble-blazor/NimbleBlazor/Components/NimbleMappingIcon.razor.cs
m-akinc: Good point. Filed item #1458.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.mdx
```diff
@@ -15,10 +14,17 @@ import * as textMappingStories from '../../../mapping/text/tests/mapping-text.st
...
-<ArgTypes of={tableColumnEnumTextStories.enumTextColumn} />
+<DocsStory of={tableColumnEnumTextStories.enumTextColumn} expanded={false} />
+<Controls of={tableColumnEnumTextStories.enumTextColumn} />
 
-### Text Mapping
+# Usage Docs
+
+## Blazor Usage
+
+When setting a child mapping elment's `Key` value to a string, you must wrap it in `@()` so that the compiler treats it as a string, e.g.
```
mollykreis, line 24:
```suggestion
When setting a child mapping element's `Key` value to a string, you must wrap it in `@()` so that the compiler treats it as a string, e.g.
```

packages/nimble-components/src/table-column/icon/tests/table-column-icon.mdx
```diff
@@ -16,15 +15,22 @@ import * as spinnerMappingStories from '../../../mapping/spinner/tests/mapping-s
...
-<ArgTypes of={tableColumnIconStories.iconColumn} />
+<DocsStory of={tableColumnIconStories.iconColumn} expanded={false} />
+<Controls of={tableColumnIconStories.iconColumn} />
 
-### Icon Mapping
+# Usage Docs
+
+## Blazor Usage
+
+When setting a child mapping elment's `Key` value to a string, you must wrap it in `@()` so that the compiler treats it as a string, e.g.
```
mollykreis, line 25:
```suggestion
When setting a child mapping element's `Key` value to a string, you must wrap it in `@()` so that the compiler treats it as a string, e.g.
```

mollykreis (approved):

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnMapping.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Components;
+
+namespace NimbleBlazor;
+
+public class NimbleTableColumnMapping<TKey> : NimbleTableColumn, IGroupableColumn
```
atmgrifter00, line 6: I take it this is the analog for the `TableColumnEnumBase<T>` class in `nimble-components`. Any reason for the name change? If we like this one better, should we change the one in `nimble-components`?

jattasNI (approved):

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnMapping.cs
m-akinc: Renamed this one to `NimbleTableColumnEnumBase`.

atmgrifter00 (approved):

msmithNI (approved):

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnEnumBase.cs
```diff
@@ -0,0 +1,64 @@
...
+        if (typeof(TKey) == typeof(bool))
+        {
+            return "boolean";
+        }
+        if (typeof(TKey) == typeof(string))
+        {
+            return "string";
+        }
+
+        throw new ArgumentException("TKey was not a numeric, boolean, or string.");
```
rajsite, line 62: I may have missed it buut do we have a test covering this error case?

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnEnumBase.cs
```diff
@@ -0,0 +1,64 @@
...
+    public bool? GroupingDisabled { get; set; }
+
+    private readonly Type[] cSharpNumberTypes = new[]
+    {
+        typeof(int),
+        typeof(uint),
+        typeof(short),
+        typeof(ushort),
+        typeof(long),
+        typeof(ulong),
```
rajsite, line 39: I don't think we should support 64-bit integers. They can't be losslessly converted to JS numbers

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnEnumBase.cs
```diff
@@ -0,0 +1,64 @@
...
+        typeof(byte),
+        typeof(sbyte),
+        typeof(float),
+        typeof(double),
+        typeof(decimal)
```
rajsite, line 44: Similarly we probably shouldn't support decimal since that can't hit in a 64-bit double

rajsite (approved):

packages/nimble-blazor/NimbleBlazor/NimbleTableColumnEnumBase.cs
m-akinc: Yep, we have `NimbleTableColumnIcon_WithObjectTypeParameter_ThrowsException`

Issue #1454, opened 2023-08-25
Rich Text Editor | Blinking caret `|` in editor disappears after clicking any button in Firefox browser.


aagash-ni:
## 🧹 Tech Debt

Once the editor gets focused, the blinking caret will be visible until we click other buttons in the Firefox browser (Changing focus). But once clicked, the editor internally has its focus but the blinking caret **disappears**.

Before button click:
![image](https://github.com/ni/nimble/assets/123377167/b1ac0c60-5de7-4459-9ea0-c6aa671bdda2)

After button click: (caret does not blink even when the editor is in focus)
![image](https://github.com/ni/nimble/assets/123377167/fa022a49-9f53-43a4-b84c-723af38fef03)

Note: This issue can be reproducible only in the shadow DOM (Firefox browser). It seems like this [bug ](https://bugzilla.mozilla.org/show_bug.cgi?id=1496769) is a Firefox issue and still it is not fixed in Firefox browser

Reference to the conversation: https://github.com/ni/nimble/pull/1416#discussion_r1295162427

PR #1455, opened 2023-08-25
Rich Text Editor | Fix for Editor's blinking caret disappearing issue in Firefox browser

aagash-ni:
# Pull Request

## 🤨 Rationale

This PR contains workaround for resolves https://github.com/ni/nimble/issues/1454 Firefox issue when the editor gets focused, the blinking caret will be visible until we click format buttons (Bold, Italic ...). But once any of the buttons are clicked, the editor internally has its focus but the blinking caret disappears.

Issue Link : https://github.com/ni/nimble/issues/1454

## 👩‍💻 Implementation

- Triggering `Blur()` and setting `focus()` in the editor will make the caret re-appears when clicking the formatting buttons.

## 🧪 Testing

- Verified the caret visibility in the Firefox browser when clicking on the format button. Since Firefox is reporting the editor as an active element even though Caret is not rendered, manually verified this behavior.
- Added a test case to verify the active element in the browser especially when clicking the format button.

## ✅ Checklist


change/@ni-nimble-components-46b1fdfb-63b3-4a04-ac71-1124855a4e45.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for Firefox caret disappearing issue",
+  "packageName": "@ni/nimble-components",
+  "email": "aagash.maridas@solitontech.com",
```
vikisekarNI, line 5: nit: Would you like to obfuscate your email address based on [this suggestion](https://github.com/ni/nimble/blob/4574201ae49f8e30da7a0d87db78109f4236f340/CONTRIBUTING.md?plain=1#L79)? 

https://github.com/ni/nimble/blob/4574201ae49f8e30da7a0d87db78109f4236f340/CONTRIBUTING.md?plain=1#L79

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,6 +355,18 @@ export class RichTextEditor extends FoundationElement {
                 return false;
         }
     }
+
+    // In Firefox browser, Once the editor gets focused, the blinking caret will be visible until we click format buttons (Bold, Italic ...) in the Firefox browser (Changing focus).
+    // But once the any of the buttons clicked, the editor internally has its focus but the blinking caret disappears.
```
vikisekarNI, line 360:
```suggestion
    // But once any of the toolbar button is clicked, editor internally has its focus but the blinking caret disappears.
```

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,6 +355,18 @@ export class RichTextEditor extends FoundationElement {
...
+    // As a workaround, manually triggering blur and setting focus on editor will hepls making the caret re-appears.
```
vikisekarNI, line 361:
```suggestion
    // As a workaround, manually triggering blur and setting focus on editor makes the blinking caret to re-appear.
```

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -101,6 +101,7 @@ export class RichTextEditor extends FoundationElement {
      */
     public boldButtonClick(): void {
         this.tiptapEditor.chain().focus().toggleBold().run();
+        this.focusEditorInFirefoxEnv();
```
vikisekarNI, line 104:
```suggestion
        this.focusEditorInFirefox();
```

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,6 +355,18 @@ export class RichTextEditor extends FoundationElement {
                 return false;
         }
     }
+
+    // In Firefox browser, Once the editor gets focused, the blinking caret will be visible until we click format buttons (Bold, Italic ...) in the Firefox browser (Changing focus).
+    // But once the any of the buttons clicked, the editor internally has its focus but the blinking caret disappears.
+    // As a workaround, manually triggering blur and setting focus on editor will hepls making the caret re-appears.
+    // It seems like this issue in not fixed in firefox browser yet, Created Issue https://github.com/ni/nimble/issues/1454 tracks removal of this workaround.
```
vikisekarNI, line 362: I don't think the created issue tracks removal of this workaround. Did you meant to say fix? 

However consider removing this redundant information, as it's already in PR description. 

```suggestion
    // It seems like this issue in not fixed in firefox browser yet, Created Issue https://github.com/ni/nimble/issues/1454 tracks removal of this workaround.
```

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-editor/index.ts
aagash-ni: Updated.

vikisekarNI (approved):

m-akinc (approved):

change/@ni-nimble-components-46b1fdfb-63b3-4a04-ac71-1124855a4e45.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Pro tip: in your PR description if you link to the issue using words like `resolves #1454` or `fixes #1454` then GitHub will automatically close it when the PR completes. It's handy to do this for issues like bugs or small features that don't require any validation or discussion post-commit.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,6 +355,17 @@ export class RichTextEditor extends FoundationElement {
                 return false;
         }
     }
+
+    // In Firefox browser, Once the editor gets focused, the blinking caret will be visible until we click format buttons (Bold, Italic ...) in the Firefox browser (Changing focus).
+    // But once any of the toolbar button is clicked, editor internally has its focus but the blinking caret disappears.
+    // As a workaround, manually triggering blur and setting focus on editor makes the blinking caret to re-appear.
+    private focusEditorInFirefox(): void {
```
jattasNI, line 362: Could you update the comment to include a link to the Mozilla issue tracking this bug? That way we can periodically check if it has been fixed and maybe remove this workaround someday.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -101,6 +101,7 @@ export class RichTextEditor extends FoundationElement {
      */
     public boldButtonClick(): void {
         this.tiptapEditor.chain().focus().toggleBold().run();
+        this.focusEditorInFirefox();
```
jattasNI, line 104: Anything stopping us from writing a unit test to cover this behavior? We already run the tests for nimble-components in both chrome and firefox so if you can come up with a test that failed in FF before but passes now then our infrastructure will validate it.

You might be able to achieve this with a combination of `document.activeElement` (which should be the rich text editor) and `document.activeElement.shadowRoot.activeElement` (which should be the element containing tiptap).

It's possible that Firefox was always reporting the rich text editor as the activeElement, even though it wasn't rendering the cursor. That'd make it hard to write a test that failed in Firefox before this change. However I still think it'd be worthwhile to have testing that verifies the browser is *reporting* the correct activeElement.

And some general feedback: if the PR is only being validated with manual tests then it's helpful to explain the rationale for that in the Testing section of the description.

packages/nimble-components/src/rich-text-editor/index.ts
```diff
@@ -347,6 +355,17 @@ export class RichTextEditor extends FoundationElement {
                 return false;
         }
     }
+
+    // In Firefox browser, Once the editor gets focused, the blinking caret will be visible until we click format buttons (Bold, Italic ...) in the Firefox browser (Changing focus).
+    // But once any of the toolbar button is clicked, editor internally has its focus but the blinking caret disappears.
+    // As a workaround, manually triggering blur and setting focus on editor makes the blinking caret to re-appear.
+    private focusEditorInFirefox(): void {
+        const browserInfo = navigator.userAgent;
+        if (browserInfo.includes('Firefox')) {
```
jattasNI, line 364: User agent detection is [notoriously fickle](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#browser_name_and_version). Browsers often lie about their user agent and you have to be careful about detecting the browser name (Firefox) vs the rendering agent (Mozilla). Is this code safe to run on all browsers? If so then I'd vote to remove the user agent check and run it unconditionally. (We could rename the function to something like `forceFocusEditor()`if we do this).

packages/nimble-components/src/rich-text-editor/index.ts
vikisekarNI: Thanks @jattasNI. I thought of adding a comment for testing this behavior and couldn't find any good way to test the blinking caret as you mentioned below. At least `activeElement` validation will be good.

> It's possible that Firefox was always reporting the rich text editor as the activeElement, even though it wasn't rendering the cursor. That'd make it hard to write a test that failed in Firefox before this change. However I still think it'd be worthwhile to have testing that verifies the browser is reporting the correct activeElement.

Makes sense on the PR description elaboration. Internally aligned with @aagash-ni on the same to pursue it in follow-up PRs.

packages/nimble-components/src/rich-text-editor/index.ts
vikisekarNI: Thanks for relieving the condition check here! I was also skeptical on this implementation and was exploring whether any other recommended ways are possible. 

Since this was a browser specific bug and not a feature we were also not able to use [feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection). Just curious to hear if there are any other good ways of identifying a browser, however these situations are rare.

jattasNI: I'm not aware of strategies other than the ones discussed in that MDN article.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,22 +156,30 @@ export class RichTextEditorPageObject {
...
+    public getFormattingButton(
+        index: ToolbarButton
+    ): ToggleButton | null | undefined {
+        const buttons: NodeListOf<ToggleButton> = this.richTextEditorElement.shadowRoot!.querySelectorAll(
+            'nimble-toggle-button'
+        );
+        return buttons[index];
+    }
+
+    public getActiveElementFromDocument(): Element | null {
```
jattasNI, line 168: Let's try to avoid exposing `Element` types on the public API of the page object since that exposes info about the DOM structure of the component, which should be an internal implementation detail. Instead let's expose methods like 
```ts
isRichTextEditorActiveElement(): boolean {
    return 
        document.activeElement === this.richTextEditorElement &&
        document.activeElement?.shadowRoot?.activeElement === this.getTiptapEditor();
}
```

(or those two checks could be split into separate public methods if needed)

jattasNI (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -103,6 +103,11 @@ describe('RichTextEditor', () => {
         expect(okButtonSpy).toHaveBeenCalledTimes(1);
     });
 
+    it('Should return editor as active element when clicking formatting button', async () => {
```
jattasNI, line 106:
```suggestion
    it('Should return editor as active element after clicking formatting button', async () => {
```

jattasNI (approved):
If @rajsite doesn't get to this by the time you resolve my feedback, I think it's ok to submit.

packages/nimble-components/src/rich-text-editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -156,8 +156,13 @@ export class RichTextEditorPageObject {
             .map(el => el.textContent || '');
     }
 
-    private getEditorSection(): Element | null | undefined {
-        return this.richTextEditorElement.shadowRoot?.querySelector('.editor');
+    public async isRichTextEditorActiveElement(): Promise<boolean> {
+        await waitForUpdatesAsync();
```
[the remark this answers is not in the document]
aagash-ni: Removed `await waitForUpdatesAsync()` from `isRichTextEditorActiveElement()` method

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
aagash-ni: Updated

change/@ni-nimble-components-46b1fdfb-63b3-4a04-ac71-1124855a4e45.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "Fix for Firefox caret disappearing issue",
+  "packageName": "@ni/nimble-components",
+  "email": "123377167+aagash-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
rajsite, line 7: Testing the change a bit I noticed this behavior:

mouse down

When doing a mousedown on the italic button you see the bold button showing focus temporarily until you finish the mouse up.

I can also reproduce it on main so it might not be related to this specific change but could you double check and track in an issue if needed?

rajsite (approved):

PR #1456, opened 2023-08-25
Update component status table to use nimble icons

mollykreis:
# Pull Request

## 🤨 Rationale

Update the component status table to use the `nimble-table-column-icon` rather than emoji icons.

## 👩‍💻 Implementation

- Update component status table to use icon column
- Update markdown table that is the legend for the icons
- Add an enum for framework-specific status to avoid hard-coding strings

## 🧪 Testing

Verified that the storybook docs looked correct

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/docs/component-status.mdx
+++ b/packages/nimble-components/docs/component-status.mdx
@@ -20,9 +20,9 @@ Here are components that are either implemented or on the roadmap for Nimble. So
 
 | Icon | Description                                                                                                                                                              |
 | ---- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| ✅   | The component is ready for general use.                                                                                                                                  |
-| ⚠️   | The component is "[incubating](/docs/incubating-docs--docs)". These are not ready for general use and may make breaking changes without a "major" semantic version bump. |
-| ⭕   | The component doesn't exist yet. Please submit requests by commenting on the linked issue.                                                                               |
+| <${iconCheckTag} severity="success"></${iconCheckTag}>   | The component is ready for general use.                                                                                                                                  |
+| <${iconTriangleTag} severity="warning"></${iconTriangleTag}>   | The component is "[incubating](/docs/incubating-docs--docs)". These are not ready for general use and may make breaking changes without a "major" semantic version bump. |
+| <${iconXmarkTag} severity="error"></${iconXmarkTag}>   | The component doesn't exist yet. Please submit requests by commenting on the linked issue.                                                                               |
 
 ## Future work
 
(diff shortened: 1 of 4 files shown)


packages/nimble-components/docs/component-status.stories.ts
```diff
@@ -280,64 +285,64 @@ const components = [
...
         componentName: 'Rich Text Editor',
         componentHref: './?path=/docs/incubating-rich-text-editor--docs',
         issueHref: 'https://github.com/ni/nimble/issues/1288',
         issueLabel: 'Issue',
-        componentStatus: '⚠️',
-        angularStatus: '⚠️',
-        blazorStatus: '⚠️',
+        componentStatus: ComponentFrameworkStatus.incubating,
+        angularStatus: ComponentFrameworkStatus.incubating,
+        blazorStatus: ComponentFrameworkStatus.doesNotExist,
```
mollykreis, line 299: I changed the `blazorStatus` of the Rich Text Editor and Rich Text Viewer to "does not exist" rather than "incubating" since Blazor wrappers haven't been created for those components (not is there an immediate plan to create them).

All other status for other components were left unchanged.

packages/nimble-components/docs/component-status.mdx
```diff
@@ -18,11 +18,11 @@ Here are components that are either implemented or on the roadmap for Nimble. So
...
-| ⚠️   | The component is "[incubating](/docs/incubating-docs--docs)". These are not ready for general use and may make breaking changes without a "major" semantic version bump. |
-| ⭕   | The component doesn't exist yet. Please submit requests by commenting on the linked issue.                                                                               |
+| Icon                                                             | Description                                                                                                                                                              |
+| ---------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| <nimble-icon-check severity="success"></nimble-icon-check>       | The component is ready for general use.                                                                                                                                  |
```
mollykreis, line 23: Note: I tried to use the exported tag names here, but I couldn't get them to work correctly within an mdx file.

mollykreis:
@m-akinc, can you provide a buddy review for this PR?

m-akinc (approved):

packages/nimble-components/docs/component-status.mdx
jattasNI: You could import the files containing the elements to express the direct dependency and ensure the elements are available to be loaded. I guess it works without that because the story depends on those icon tags and the MDX depends on the story, but making the dependency explicit from here too seems slightly more correct to me.

`import '../src/icons/check';` etc

jattasNI (approved):
I think it's fine to submit this without @rajsite's review

PR #1457, opened 2023-08-25
Number text decimal format

mollykreis:
# Pull Request

## 🤨 Rationale

This PR is part of #1011. It adds support for making a table column have `decimal` formatting, which includes:

- Adding a new `format` value of `decimal`
- Adding `decimal-digits` attribute
- Adding validation that `decimal-digits` is in the allowed range
- Making the default alignment of a `decimal` column `right`

## 👩‍💻 Implementation

The general approach of these new capabilities is aligned with other columns that have been implemented. A few notes that might be of interest:

- This PR adds a `TableColumnNumberTextValidator` to validate the `decimal-digits`. The column is invalid if its `format` is `decimal` and `decimal-digits` is outside of the supported range.
- Added a new `DecimalFormatter` class that extends `NumberFormatter`. It differs slightly from the `RoundToIntegerFormatter` and `DecimalFormatter` in that the `formatter` is not static because the number of decimals being rendered can vary between columns.

## 🧪 Testing

- New unit tests
- Added additional functionality to storybook story & manually verified it was correct
- Added another column to the matrix tests to show the decimal format

## ✅ Checklist


mollykreis:
@atmgrifter00, can you buddy this for me?

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
```diff
@@ -0,0 +1,75 @@
+import { getSpecTypeByNamedList } from '../../../../utilities/tests/parameterized';
+import { DecimalFormatter } from '../decimal-formatter';
+
+describe('DecimalFormatter', () => {
+    const testCases: readonly {
+        name: string,
+        decimalDigits?: number,
```
atmgrifter00, line 7: Perhaps there should be a case that tests the `undefined` case?

mollykreis: There are a few test cases that don't specify a value for `decimalDigits` (-∞, ∞, and NaN), but you're right, there isn't a reason why those can't specify a value so that the test as a whole can be simplified.

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -59,18 +73,36 @@ export class TableColumnNumberText extends TableColumnTextBase {
...
+
+        if (this.validator.isValid()) {
+            const columnConfig: TableColumnNumberTextColumnConfig = {
+                formatter: this.createFormatter(),
+                alignment: this.determineCellContentAlignment()
+            };
+            this.columnInternals.columnConfig = columnConfig;
+        } else {
+            this.columnInternals.columnConfig = undefined;
+        }
```
atmgrifter00, line 95:
```ts
this.columnInternals.columnConfig = this.validator.isValid()
    ? {
            formatter: this.createFormatter(),
            alignment: this.determineCellContentAlignment()
       }
     : undefined;
```
Maybe? I don't feel strongly, just thought I'd offer a "less lines" possibility.

mollykreis: I liked the suggestion until I realized that the typing only knows that `this.columnInternals.columnConfig` is of type `undefined | unknown` (not particularly helpful). Therefore, I think I'm going to leave this as-is so that we can get strict typing of `columnConfig` before assigning it to `this.columnInternals.columnConfig`.

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
mollykreis: I updated the test to make `decimalDigits` required.

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -59,18 +73,36 @@ export class TableColumnNumberText extends TableColumnTextBase {
         this.updateColumnConfig();
     }
 
+    private decimalDigitsChanged(): void {
+        // decimalDigits is only used by the 'decimal' format option. Only update the
+        // column configuration if that is the current format option.
```
jattasNI, line 78: It [doesn't look like the date-text column is this targeted](https://github.com/ni/nimble/blob/4574201ae49f8e30da7a0d87db78109f4236f340/packages/nimble-components/src/table-column/date-text/index.ts#L208) in updating its column config when settings change for a format other than the current one.

If you think this is worth doing, could we do it consistently for other column types? (I'm OK with this happening in a separate PR if that's cleaner). 

I'm also OK changing this to be unconditional like the date column is now.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -100,12 +103,19 @@ const formatDescription = `Configures the way that the numeric value is formatte
...
     </ul>
 </details>
 `;
 
+const validityDescription = `Readonly object of boolean values that represents the validity states that the column's configuration can be in.
+The object's type is \`TableColumnValidity\`, and it contains the following boolean properties:
+-   \`invalidDecimalDigits\`: \`true\` when \`format\` is configured to \`decimal\` and \`decimal-digits\` is set to a number less than 0 or greater than 20.
+`;
+
 const alignmentDescription = `Configures the alignment of the value within the column.
```
jattasNI, line 119: Reminder to update the "Default Alignment" section to mention the behavior of the decimal format.

jattasNI (approved):
This is following established patterns so I think it's ok to submit before @rajsite is back.

packages/nimble-components/src/table-column/number-text/index.ts
mollykreis: There isn't a good reason for that check here, so I'll remove it. I think that'll be better for consistency and to make the code paths more straightforward.

PR #1461, opened 2023-08-28
Make DateText columnConfig `undefined` when the column is invalid

mollykreis:
# Pull Request

## 🤨 Rationale

After a brief team discussion, we decided that it would make more sense to make the `columnConfig` `undefined` when the date-text column was invalid rather than having the `columnConfig` contain invalid options (i.e. an `undefined` formatter`). This is consistent with the number-text column.

## 👩‍💻 Implementation

Set `columnConfig` to `undefined` when creating the `DateTimeFormat` object throws an exception rather than creating a `columnConfig` object with an `undefined` `formatter`.

## 🧪 Testing

- Ran existing unit tests

## ✅ Checklist


mollykreis:
@m-akinc, could I get a buddy review from you for these changes?

m-akinc (approved):

jattasNI (approved):

Issue #1150, opened 2023-04-03
IxD recommendation/documentation/best practices for Tree, Accordion, Disclosure, Table row details


fredvisser:
## 😯 Problem to Solve

We need to document our recommendations for when SLE apps should use the Tree vs. Accordion vs. Disclosures vs. Table row detail expanders. 

This discussion can include the use of the tree in the navigation menu, but it must also include –
1. The right-rail content that is common in many of the SLE apps. 
2. A recommendation for fixing the design of apps that use sparse-grouped tables (where only the child rows have any content for the columns)
3. When to use disclosure areas vs accordions
4. When to use trees

## 📋 Tasks
- [ ] Create a IxD spec

PR #1467, opened 2023-08-29
Accordion IxD Specification

leslieab:
# Pull Request

## 🤨 Rationale
For #[1150](https://github.com/ni/nimble/issues/1150) - IxD recommendation for Tree, Accordion, Disclosure, Table row details

- Interaction design specification for the accordion component
- Clarity on the accordion vs. disclosure, tree, etc. relative to the accordion

## 👩‍💻 Implementation
N/A

## 🧪 Testing
N/A

## ✅ Checklist
- [x] Submit project documentation
- [x] Review 


packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+
+The client user can specify whether one panel or more than one panel can be opened at a time. If the former, clicking to open a closed panel will close the open panel (if applicable).
+
+<p>
+<kbd><img src="spec-images/Accordion-4.png"></kbd>
+<br>
+<sub>Click (one panel open at a time)</sub>
+</p>
+
+If a component in a panel is in an error state, the error state is indicated on the associated header button when closed (if not also when open).
```
jattasNI, line 115: You use the phrase "if not also" a few times in the doc and I want to make sure I understand the intent. Does it mean something like:
1. "and maybe also (but just a suggestion)"?
or
2. "but only if not"?
or
3. something else?

I'd love it if we could come up with language that's more precise (e.g. "must" / "should" / "could" like in [this framework](https://en.wikipedia.org/wiki/MoSCoW_method))

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+Do not use an accordion:
+
+-   For a single panel of content
+-   To organize unrelated information into panels
+-   For selectable or navigable lists or trees
+
+Best practice is to use multiple disclosures in one accordion and avoid one disclosure in multiple accordions.
+
+<p>
+<kbd><img src="spec-images/Accordion-13.png"></kbd>
```
jattasNI, line 34: nitpick: why are there `<kbd>` tags? I'd never heard of them before but learned they're for [formatting text that indicates keyboard shortcuts](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) so it seems odd to put them around images.

I'd vote to use markdown syntax for images if we can:
```
![You can write alt text here or have blank space between the brackets](spec-images/Accordion-13.png)
```

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+
+`TAB` to the accordion focuses the first header button (if not also visually indicating the entire accordion).
+
+<p>
+<kbd><img src="spec-images/Accordion-8.png"></kbd>
+<br>
+<sub>TAB to accordion</sub>
+</p>
+
+`TAB` focuses the next focusable element (if not also visually indicating the header button and/or the entire accordion). All focusable elements in the accordion are included in the `TAB` sequence.
```
jattasNI, line 158: Should the tab sequence include focusable elements in collapsed sections? My inclination would be that it should only include focusable elements in open sections, but want to confirm.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+<br>
+<sub>Nested accordion in "info" panel</sub>
+</p>
+
+### Keyboard Interactions
+
+| Key                | Description                                           |
+| ------------------ | ----------------------------------------------------- |
+| `TAB`              | Moves focus to next focusable element                 |
+| `SPACE` or `ENTER` | Opens/closes accordion section when header is focused |
```
jattasNI, line 148: Just wanna confirm: does ARIA give guidance about space vs enter vs both? If so it'd be good to mention that we're abiding / deviating from its guidance.

jattasNI:
Don't take my feedback as discouraging: I love this format and the fact that we get to have all these discussions up front! And the content you have is mostly exactly what I was hoping to see.

change/@ni-nimble-components-225009b9-77d6-4cbd-a94e-d9ba2751b047.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
NIbokeefe: I prefer questions and mockups to be presented in person if you'd like input.  I want to avoid having to dig through threads as much as possible.

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: I was trying to make a border or box shadow around the images (because they have white backgrounds that bleed into the document background), but GitHub doesn't allow custom CSS styles in Markdown.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+n
+
+## Background
+
+-   [IxD recommendation for Tree, Accordion, Disclosure, Table row details (#1150)](https://github.com/orgs/ni/projects/7/views/13?pane=issue&itemId=24667320)
+-   [ViD Accordion Styles (Figma)](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-85131&mode=design)
+
+## Usage
+
+### Accordions & Disclosures
```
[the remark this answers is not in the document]
leslieab: I have an accordion component and a disclosure component because the ARIA patterns include both. Fast controls do as well. Disclosures can have lots of different permutations than just looking like an expander panel. I guess functionally, the accordion doesn't necessarily have to have literal disclosure components inside of it.

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: Yes, ARIA has both ENTER and SPACE.

change/@ni-nimble-components-225009b9-77d6-4cbd-a94e-d9ba2751b047.json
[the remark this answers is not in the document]
leslieab: I can present in person, if that's better for visual designers. However, it would be useful to have these reviewed by visual designers to ensure the information contains what they need from IxD. The intent is to have things documented to support asynchronous workflows and capture discussions about the design so we can go back and re-examine them as needed.

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: You're right, I felt a bit weird about that phrasing as well. Great suggestion, thanks for the reference.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+
+<p>
+<kbd><img src="spec-images/Accordion-6.png"></kbd>
+<br>
+<sub>"Expand" button</sub>
+</p>
+
+### Nested Accordions
+
+Accordion panels can have one level of nested accordions. Nested accordions are always ghost style. [**QUESTION:** Can accordions be nested anywhere in the panel? Can there be multiple child accordions nested in an accordion?]
```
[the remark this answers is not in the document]
leslieab: I didn't include it because it felt like a visual design thing. One thing I learned from this first spec was that I think I need to reach out to designers (if not also clients?) to get their cases/context for the component, which could help inform what visual styles are needed. But I don't feel like it's an interaction design decision. @NIbokeefe thoughs?

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+<sub>ENTER to close an opened section</sub>
+</p>
+
+## Validation
+
+TBD
+
+## Open Issues
+
+See content marked "**QUESTION**" or "**NOTE**".
```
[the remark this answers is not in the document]
leslieab: I agree, I think having a "Future Considerations" section in the template to aggregate any open questions from discussion is a good idea.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,193 @@
...
+
+<p>
+<kbd><img src="spec-images/Accordion-12.png"></kbd>
+<br>
+<sub>ENTER to close an opened section</sub>
+</p>
+
+## Validation
+
+TBD
```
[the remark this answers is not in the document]
leslieab: The document is being provided on the document in GitHub, so I wasn't sure what to link. These specs are meant to be quick. I think ideally I make a post soliciting feedback from stakeholders on my PRs on Teams (as in person meetings could be scheduled as needed).

Likewise, engaging with user research is necessary, but 1) it seems like it would be a block to posting the spec, 2) if I'm responsible for making this happen for every component I think we need to meet with user research and develop a low-effort protocol for doing this, and 3) it seems like user testing should happen with the visual design as well?

So I'm not really sure.

packages/nimble-components/src/accordion/specs/IxD.md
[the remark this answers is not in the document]
jattasNI: Nice! Feel free to add that change to this PR if you want. Or let me know if you'd rather I do it in a separate PR.

packages/nimble-components/src/accordion/specs/IxD.md
jattasNI: My vote would be to live with GitHub's mediocre styling rather than trying to work around it.

packages/nimble-components/src/accordion/specs/IxD.md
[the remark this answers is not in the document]
leslieab: The latter, please and thank you!

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: How do I add line breaks between text and an image? It appears to be rendering as floats instead of blocks?

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,108 @@
...
+Hovering over a header button indicates interactivity:  
+![ ](spec-images/Accordion-2.png)
+
+Clicking anywhere on a header button opens (if closed) or closes (if opened) the associated panel of content. [**QUESTION:** Do panels only display block content (one component per "line") or do they display flex content?]  
+![ ](spec-images/Accordion-3.png)
+
+The client user can specify whether one panel or more than one panel can be opened at a time. If the former, clicking to open a closed panel will close the open panel (if applicable).  
+![ ](spec-images/Accordion-4.png)
+
+If a component in a panel is in an error state, the error state is visually indicated on the associated header button when closed (and maybe also be visually indicated when open).
```
atmgrifter00, line 73: This point may be worth clarification from a technical perspective. I presume the desired behavior is that if there is any component within an accordion-item/disclosure that has an error-state (the technical meaning of is that a component has an `error-visible` attribute set), that the accordion-item itself would visually represent that error at the very least when it is collapsed.

If this is, in fact, how we want the component to behave (i.e. requires no user configuration), then there are technical implications to that, namely that we will be putting a burden on the component to track DOM changes to _all_ of its children.

The alternative to this is that we instead create a user-facing API for the accordion-item itself that allows a user to specify that it's in an error state (i.e. the `error-visible` attribute).

Personally, I would like the `accordion-item` to behave as described here, however I'm not sure if we should vet the performance implications of this.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,108 @@
...
+
+If a component in a panel is in an error state, the error state is visually indicated on the associated header button when closed (and maybe also be visually indicated when open).  
+![ ](spec-images/Accordion-5.png)
+
+### Additional Options
+Client users may need/want to have additional buttons and/or icons in the header. For example, an expand button:  
+![ ](spec-images/Accordion-6.png)
+
+### Nested Accordions
+Accordion panels can have one level of nested accordions. Nested accordions are always ghost style. [**QUESTION:** Can accordions be nested anywhere in the panel? Can there be multiple child accordions nested in an accordion?]
```
atmgrifter00, line 81: If you're looking for a technical response, the answer is "yes, `accordions` can be placed anywhere respective to other children of an `accordion-item`, and yes, there could be multiple child `accordions` in an accordion".

One question I have about the behavior you have described is the limitation of "one level". If a user has made an `accordion` with two or more levels what is your expected behavior? There is certainly no technical burden, and it's a visually recursive pattern simply limited to the horizontal space.

atmgrifter00:
Since we are mentioning the term "disclosure" frequently, it would be good to get some clarity on what we feel are the complete set of differences between an accordion and a disclosure component. One difference is captured already in this document, in that an accordion can have several collapsible sections (particularly indented at the same visual level), whereas a disclosure only has a single collapsible section. However, it's important to note that an accordion can have a single collapsible section, making that particular use-case fairly analogous to a disclosure.

Another possible difference, which could be more convention than anything else, is that disclosure's are typically described and visually represented as a "button". From ARIA's disclosure pattern description:

_" When the controlled content is hidden, the button is often styled as a typical push button with a right-pointing arrow or triangle to hint that activating the button will display additional content."_.

So, I think my questions are summarized with:
- If we wanted Nimble to have an actual "disclosure" component would we want to adopt this visual representation of using a button?
- Is there a good reason for users to not use an accordion with a single collapsible section?

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Additional Options
+Client users may need/want to have additional buttons and/or icons in the header. For example, an expand button:  
+![ ](spec-images/Accordion-6.png)
+
+### Nested Accordions
+Accordion panels can have one level of nested accordions. Nested accordions are always ghost style. [**QUESTION:** Can accordions be nested anywhere in the panel? Can there be multiple child accordions nested in an accordion?]  
+![ ](spec-images/Accordion-7.png)
+
+### Keyboard Interactions
```
atmgrifter00, line 84: @leslieab, any reason to not include the other optional keyboard interactions ARIA outlines for the accordion (ARROWUP, ARROWDOWN, HOME, and END)?

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,108 @@
+# Nimble Accordion (IxD)
```
rajsite, line 1: @atmgrifter00 moved to comment thread:

Since we are mentioning the term "disclosure" frequently, it would be good to get some clarity on what we feel are the complete set of differences between an accordion and a disclosure component. One difference is captured already in this document, in that an accordion can have several collapsible sections (particularly indented at the same visual level), whereas a disclosure only has a single collapsible section. However, it's important to note that an accordion can have a single collapsible section, making that particular use-case fairly analogous to a disclosure.

Another possible difference, which could be more convention than anything else, is that disclosure's are typically described and visually represented as a "button". From ARIA's disclosure pattern description:

_" When the controlled content is hidden, the button is often styled as a typical push button with a right-pointing arrow or triangle to hint that activating the button will display additional content."_.

So, I think my questions are summarized with:
- If we wanted Nimble to have an actual "disclosure" component would we want to adopt this visual representation of using a button?
- Is there a good reason for users to not use an accordion with a single collapsible section?

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: This is from the discussion we had during a meeting when we decided that we would not allow a client-user to nest accordion(s) inside of an accordion more than one level deep. We did not discuss _how_ we would do that. Is that something interaction design should brainstorm on? I'm fine saying that this is the only thing we are "artificially" limiting, and everything else that is technically feasible remains technically feasible.

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: - Maybe. This looks like what FAST does. The ARIA pattern has examples of disclosures that look like push buttons, <details> tags, menu buttons, and menu bars. It seems to be some kind of a catch all for when you want a component that looks like another component but without all the complex implementation built into it. So maybe it isn't a component.
- No, I guess not.

packages/nimble-components/src/accordion/specs/IxD.md
leslieab: No, I'll add them.

packages/nimble-components/src/accordion/specs/IxD.md
atmgrifter00: No, I think you're right that this isn't an "interaction concern". If anything it might be a visual design concern if we decided that there would be a user-facing consequence of nesting more than one-level deep that was different than "it just working as expected".

rajsite (commented):
Can't think of many IxD questions beyond what is discussed. Think the things I want to clarify are more Visual Design related.

jattasNI (approved):

change/@ni-nimble-components-225009b9-77d6-4cbd-a94e-d9ba2751b047.json
jattasNI, line 1: Note that our build pipeline enforces some basic formatting rules on markdown and it's currently failing. We have [some instructions](https://github.com/ni/nimble/blob/cc709bb825c03fa299e401b152df19ae3c243e1a/CONTRIBUTING.md#linting) about how to configure VSCode to automatically fix these for you. They're written for a developer audience though so feel free to reach out if you need help setting it up.

change/@ni-nimble-components-225009b9-77d6-4cbd-a94e-d9ba2751b047.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
+  "comment": "Initial submission of accordion ixd spec",
+  "packageName": "@ni/nimble-components",
+  "email": "leslie.bloomfield@ni.com",
```
jattasNI, line 5: Most of us have chosen to configure our system to use an obfuscated github email rather than expose our NI emails to the world.

To fix this [go to github settings](https://github.com/settings/emails) and click "Keep my email address private". Then you can copy the obfuscated email from there and manually update it here. There are also instructions there on how to configure git to always use the obfuscated email in the future.

jattasNI (approved):

change/@ni-nimble-components-225009b9-77d6-4cbd-a94e-d9ba2751b047.json
leslieab: silvenon.mdx is deprecated so I can't install that, but I have everything else installed. I ran npm run format and npm run lint after I committed my most recent change. But now when I try to "Sync changes" it says "Can't push refs to remote. Try running "Pull" first to integrate your changes."

mollykreis (approved):

rajsite (approved):

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Behavior
+
+#### Minimum Width
+
+Most accordions should fill 100% of the width of a parent container, and the parent container should have a reasonable minimum width that allows the accordion content to be usable (not pictured).
+
+#### Minimum Height
+
+When the accordion height exceeds that of the viewport, the whole accordion should vertically scroll (not pictured). Content should never scroll vertically or horizontally inside individual panels.
```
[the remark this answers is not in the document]
leslieab: No, I think individual controls that scroll are fine. It's simply the content area of the accordion that should not scroll.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,172 @@
...
+
+> NOTE: NEEDS VISUAL DESIGN & TECH WRITER?
+
+Accordion panels can have one level of nested accordions. If the client-user attempts to nest an accordion more than one level deep, the accordion component should indicate that limitation. Nested acccordions are always ghost style; the client-user should not be able to specify a block or outline appearance.
+
+![ ](spec-images/accordion-9.png)
+
+#### Error State
+
+If a component in a panel is in an error state, the error state is visually indicated on the associated header button when closed (and may also be visually indicated when open).
```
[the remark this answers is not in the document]
leslieab: Sorry, I don't understand your question, can you explain it?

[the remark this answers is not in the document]
leslieab: Yeah, that'd be a visual design question.

packages/nimble-components/src/accordion/specs/IxD.md
```diff
@@ -0,0 +1,172 @@
...
+#### Show & Hide
+
+The client-user should be able to specify whether only one content panel can be opened at a time.
+
+![ ](spec-images/accordion-7.png)
+
+#### Additional Header Options
+
+The client-user may need/want to have additional content in the header. For example, an expand button:  
+![ ](spec-images/accordion-8.png)
```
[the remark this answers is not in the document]
leslieab: Added, and also clarified in the ARIA section.

packages/nimble-components/src/accordion/specs/IxD.md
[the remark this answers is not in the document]
leslieab: Yeah, the accordion itself should never throw an error and therefore doesn't need an error message. However, the child components in an accordion panel might throw at error, in which case the error state and message will be displayed on that component. My concern is that if an accordion is closed, the end-user won't know that there is an error. But you're right, a disappearing error indicator on the header button might be a little bit weird. I will add that it should be visually indicated when both closed and open, if that addresses your concerns?

Issue #1469, opened 2023-08-30
Nimble combobox throws error when having more than 300 options


cristiancristea19:
# 🐛 Bug Report
When having more than 300 items in the options list, the `nimble-combobox` throws this error:
![MicrosoftTeams-image (1)](https://github.com/ni/nimble/assets/79762770/94be0c52-4cf4-4c2d-9e4c-632ef2d5a848)

## 💻 Repro or Code Sample
I reproduced the bug in the SystemLink example app in this [PR](https://ni.visualstudio.com/DefaultCollection/DevCentral/_git/Skyline/pullrequest/550620) and in the Nimble Angular example app.

## 🤔 Expected Behavior
The error should not be thrown even if we have more than 300 options in the combobox.

## 😯 Current Behavior
The error is displayed in the console.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #1470, opened 2023-08-30
Combobox many options bug

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1469

## 👩‍💻 Implementation

The culprit here was calling `changeDetector.detectChanges()` for every set of the `ngValue` of a `list-option`. This caused a recursive stack call that became too large.

To solve this, I've now just had each set of an `ngValue` on a `list-option` queue an update to its `combobox` CVA owner, which will then process those updates in the `ngAfterViewChecked` lifecycle hook, thus, no longer requiring a call to `changeDetector.detectChanges()`, as the views should have been updated as a result of any model updates.

## 🧪 Testing

Added unit tests.

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
```diff
@@ -225,6 +225,32 @@ describe('Nimble combobox control value accessor', () => {
...
+
+        it('update option value with new model, followed by updating combobox value with same model updates display correctly', async () => {
+            const newModelValue: TestModel = { name: 'newName', value: 100 };
+            testHostComponent.dynamicOption = newModelValue;
+            testHostComponent.selectedOption = newModelValue;
+            fixture.detectChanges();
+            await waitForUpdatesAsync();
+
+            expect(combobox.value).toBe('newName');
+        });
```
atmgrifter00, line 237: I'll note that this test would have also passed if I went with an implementation that used `window.requestAnimationFrame()` to process the `list-option` updates (as opposed to the lifecycle hook). I'm not sure there's a way to write the test that would demonstrate that the provided implementation avoids a race condition that we posited would be present with the `window.requestAnimationFrame()` approach.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
```diff
@@ -225,6 +225,32 @@ describe('Nimble combobox control value accessor', () => {
...
+            testHostComponent.selectOptions = newOptions;
+            let exceptionThrown = false;
+            try {
+                fixture.detectChanges();
+            } catch (e) {
+                exceptionThrown = true;
+            }
+
+            expect(exceptionThrown).toBeFalse();
+        });
```
atmgrifter00, line 253: Unsure about this test (at least the way its written...couldn't think of an alternative though). I'll say that in the previous implementation this test does, indeed, fail, but not because we caught an exception. The type of error that is produced doesn't seem available for capture.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
```diff
@@ -225,6 +225,32 @@ describe('Nimble combobox control value accessor', () => {
...
+            }
+            testHostComponent.selectOptions = newOptions;
+            let exceptionThrown = false;
+            try {
+                fixture.detectChanges();
+            } catch (e) {
+                exceptionThrown = true;
+            }
+
+            expect(exceptionThrown).toBeFalse();
```
rajsite, line 252: prefer the jasmine exception helpers

```suggestion
    expect(() => {
        fixture.detectChanges();
    }).not.toThrow();
```

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
```diff
@@ -442,5 +468,31 @@ describe('Nimble combobox control value accessor', () => {
...
```
rajsite, line 495: prefer the jasmine exception helpers

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -61,7 +64,18 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
     @HostListener('blur')
     private onTouched: () => void;
 
-    public constructor(private readonly _renderer: Renderer2, private readonly _elementRef: ElementRef) {}
+    public constructor(
+        private readonly _renderer: Renderer2,
+        private readonly _elementRef: ElementRef,
+        private readonly changeDetector: ChangeDetectorRef
```
rajsite, line 70: Is this used anywhere?

rajsite (approved):

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/tests/nimble-combobox-control-value-accessor.directive.spec.ts
rajsite: Yea it was something about writing to ngModel before the options were evaluated in a rAF. Nothing immediately came to mind how to exercise it further.

angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
atmgrifter00: Nope!

PR #1474, opened 2023-08-31
Update table page object to return spinner from icon column

m-akinc:
# Pull Request

## 🤨 Rationale

Clients like SLE will need ability to get/assert spinner that is rendered in an icon column's cell.

## 👩‍💻 Implementation

Changed two existing functions that previously only returned rendered icons to now return either the rendered icon or spinner.

## 🧪 Testing

Auto-tests pass

## ✅ Checklist


change/@ni-nimble-components-efb8d221-fd75-4cab-ae9d-fb2e7458ffec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
m-akinc, line 2: Technically this is a breaking change to our table page object's API, but it's incubating and there are no clients using it.

m-akinc:
@mollykreis could you give this a quick buddy?

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -187,27 +188,37 @@ export class TablePageObject<T extends TableRecord> {
...
+                `Icon or Spinner not found at cell ${rowIndex},${columnIndex}`
             );
         }
-        return icon;
+        return content;
     }
 
-    public getRenderedIconColumnGroupHeaderIcon(groupRowIndex: number): Icon {
-        const icon = this.getGroupRowHeaderView(groupRowIndex).shadowRoot!
+    public getRenderedIconColumnGroupHeaderContent(
```
mollykreis, line 208: This isn't returning the whole group header content. So I'm wondering if the name should be updated to indicate that it'll just return the icon/spinner, or if the function should be updated to return an object with `text: string` and `notSureWhatToNameThis: Icon | Spinner`. Perhaps the question is whether or not there is another function on the page object that would let you separately get the text, or if you have to return it from here to have it be available to tests?

m-akinc: There is already a function that returns the text content, but it was named `getRenderedGroupHeaderContent`. I decided to rename four functions to make them more specific about their behavior:
`getRenderedGroupHeaderContent` -> `getRenderedGroupHeaderTextContent`
`getRenderedCellContent` -> `getRenderedCellTextContent`
`getRenderedIconColumnGroupHeaderContent` -> `getRenderedIconColumnGroupHeaderContentElement`
`getRenderedIconColumnCellContent` -> `getRenderedIconColumnCellContentElement`

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -187,30 +188,40 @@ export class TablePageObject<T extends TableRecord> {
         return anchor as Anchor;
     }
 
-    public getRenderedIconColumnCellIcon(
+    public getRenderedIconColumnCellContentElement(
```
mollykreis, line 191: Calling the icon/spinner the "Content Element" makes these function names a bit long & awkward to read. What do you think about just calling it an "Icon" even though it could be an icon or a spinner? Technically we called it the "icon" column, so maybe it isn't too confusing to extend that terminology here even though it might be a spinner rather than an icon. That would make these functions `getRenderedIconColumnCellIcon` and `getRenderedIconColumnGroupHeaderIcon`. Thoughts?

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -223,7 +234,7 @@ export class TablePageObject<T extends TableRecord> {
             'nimble-table-group-row'
         );
         return Array.from(groupRows).map((_, i) => {
-            return this.getRenderedGroupHeaderContent(i);
+            return this.getRenderedGroupHeaderTextContent(i);
```
mollykreis, line 237: I can't put a comment on the actual line, but it seems like this function should be `getAllRenderedGroupHeaderTextContent` rather than `getAllRenderedGroupHeaderContent`.

mollykreis (approved):

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -190,40 +191,50 @@ export class TablePageObject<T extends TableRecord> {
     public getRenderedIconColumnCellIcon(
```
rajsite, line 191: For `getRenderedIconColumnCellIcon` and `getRenderedIconColumnGroupHeaderIcon` can we return a tagName string instead of an element instance? We have been trying to avoid page objects exposing element references.

I'm betting it'll improve the expect statements as well instead of `expect(blah instanceof Icon).toBeTrue()` we can expect a tag `expect(tagName).toBe(iconTagName)`

rajsite (approved):

jattasNI (approved):

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: The drawback would be that you wouldn't be able to get `Icon.severity`. Would you prefer we add a separate function for that?

rajsite: Yep!

m-akinc: Okay, added two new functions: one to get the icon severity and one to get the Aria label.

PR #1475, opened 2023-08-31
Update number column HLD

mollykreis:
# Pull Request

## 🤨 Rationale

Update the HLD for the number column to adjust for a change in the API direction.

## 👩‍💻 Implementation

None

## 🧪 Testing

None

## ✅ Checklist


mollykreis:
@atmgrifter00 - can you provide a buddy review?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -336,7 +334,13 @@ Nimble already has a mechanism for clients to provide custom columns by deriving
 
 ### Additional unit APIs
 
-We considered adding attributes `prefix-field-name` and `suffix-field-name` which would allow those values to be populated with a dynamic field value rather than a constant string. This could help use cases where values have disparate types or units. We decided this is not necessary initially but could revisit in the future. See [this thread](https://github.com/ni/nimble/pull/1268#discussion_r1212385898) for more discussion.
+We considered adding attributes for `prefix` and `suffix` that could be used to specify any string to append to the front or back of the formatted number. We decided against this idea because it poses a localization challenge where units that are a suffix in one locale might be a prefix in another. Therefore, we decided to require units to be those supported by ECMAScript so that localization would be built in to the formatter.
```
atmgrifter00, line 337: Do we have an example of units are a suffix in one locale and a prefix in another? If so, it might be good to just cite it here.

atmgrifter00 (approved):

mollykreis:
@atmgrifter00, I ended up pushing an additional significant change to the HLD proposing that we get rid of the `roundToInteger` format. Therefore, I reset your approval so that you can take another pass at the new proposal.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -336,7 +334,13 @@ Nimble already has a mechanism for clients to provide custom columns by deriving
 
 ### Additional unit APIs
 
-We considered adding attributes `prefix-field-name` and `suffix-field-name` which would allow those values to be populated with a dynamic field value rather than a constant string. This could help use cases where values have disparate types or units. We decided this is not necessary initially but could revisit in the future. See [this thread](https://github.com/ni/nimble/pull/1268#discussion_r1212385898) for more discussion.
+We considered adding attributes for `prefix` and `suffix` that could be used to specify any string to append to the front or back of the formatted number. We decided against this idea because it poses some localization challenges. For example, the prefix or suffix might change based on the value of the number (e.g. `1 gram` vs `2 grams`). Additionally, there may be cases where units that are a suffix in one locale are a prefix in another (e.g. `56.8 degrees Celsius` in English will be represented as `摂氏 56.8 度` in Japanese). Therefore, we decided to require units to be those supported by ECMAScript so that localization would be built in to the formatter.
```
jattasNI, line 337: My biggest hesitation is that we don't have an answer for a client that wants to display a non-ECMAScript unit in the cell text (e.g. `5 V`). The closest we have to needing that that I'm aware of are:
1. [specs grid](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/TestInsights/src/app/specs/component/specs-grid/specs-grid.component.ts) which displays a column for units next to the values. 
2. [tags grid](https://test.lifecyclesolutions.ni.com/tags/) which can display a column for units (%, BYTES) next to the values.

I can easily see someone asking us to display those unit values in the same cell. What should we tell them?

1. If units are uniform across all records:
    1. display units in the column header
    2. write a custom column type for those units that knows how to scale them (e.g. bytes / file size, elapsed time)
2. If units are not uniform across all records:
    1. add a separate column to display units
    2. change to a string column
    3. write a more intelligent custom column type that knows how to scale and sort values that weren't originally uniformly scaled

2.1 and 2.2 would not be sorted correctly, but the legacy grid has the same limitation.

I'm not sure if clients would be satisfied by those options or not. We could try to find out (e.g. chat with Adam A and TJ about these options). Or we could brainstorm more ideas to meet their (possibly imaginary) requirements. Or we could call that out as explicitly not supported for now and wait for them to come to us with a concrete request. But I'm hesitant to do the last option since I feel they've expressed a lot of interest in this topic so I think we should try to satisfy their goals.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -165,22 +165,17 @@ Nimble will introduce `nimble-table-column-number-text` which formats a numeric
...
+    -   `'decimal'` - format all values as decimal values (e.g. 123.45), always displaying `decimal-digits` digits after the separator and never displaying exponential notation. If required, values will be rounded to reach the specified number of decimial digits. Configuring `decimal-digits` to `0` will round the value to the nearest integer and display it with no decimal places. Will be displayed right-aligned by default.
+    -   This could be extended to other pre-configured formats in future. Their configuration attributes would be prefixed with the name of the format mode.
+    -   **Note:** all of the above will be implemented using a `Intl.NumberFormat` formatter. Nimble will configure the formatter with defaults to match the [visual design spec](https://github.com/ni/nimble/issues/887). The exception is that we will set `useGrouping: true` to achieve `1,000` rather than `1000` because this styles the values in a way that is more human readable.
 -   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2.
```
rajsite, line 174:
```suggestion
-   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 if unspecified.
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -165,22 +165,17 @@ Nimble will introduce `nimble-table-column-number-text` which formats a numeric
...
-    -   This could be extended to other pre-configured formats in future. Their configuration attributes would similarly be prefixed with the name of the format mode.
-    -   **Note:** all of the above will be implemented using a `Intl.NumberFormat` formatter. For all modes besides `custom` Nimble will configure the formatter with defaults to match the [visual design spec](https://github.com/ni/nimble/issues/887). The exception is that we will set `useGrouping: true` to achieve `1,000` rather than `1000` because this styles the values in a way that is more human readable.
+    -   `'decimal'` - format all values as decimal values (e.g. 123.45), always displaying `decimal-digits` digits after the separator and never displaying exponential notation. If required, values will be rounded to reach the specified number of decimial digits. Configuring `decimal-digits` to `0` will round the value to the nearest integer and display it with no decimal places. Will be displayed right-aligned by default.
```
rajsite, line 171: Need to add `round-to-integer` mode?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -165,22 +165,17 @@ Nimble will introduce `nimble-table-column-number-text` which formats a numeric
...
 -   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2.
--   `custom-*` - when format is `custom`, these attribute-cased values will be passed to the equivalent camelCased fields of the `options` parameter of the `Intl.NumberFormat` constructor. For example, `options.maximumFractionDigits` will be set to the value of `custom-maximum-fraction-digits`. These fields are all string, boolean, or number and their property equivalents will be strictly typed. Initially the column will expose configurations that are [not marked experimental](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility).
+-   `unit` - an optional string that specifies the units to apply to the value when it is displayed. The value must be [one of the units sanctioned for use in ECMAScript](https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers). When a unit is specified, the `style` of the `Intl.NumberFormat` formatter will be set to `unit`.
```
rajsite, line 175: It's not super clear which format modes unit and unitDisplay apply to. Is it all modes?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: Also maybe we should say that decimal-digits is only used by the decimal formatter?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: Oh I see the comment above about getting rid of roundToInteger. I think I forgot the context of that discussion 😅
Is the idea that with decimal-digits and format decimal it becomes redundant to have roundToInteger?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: Maybe we should chat about this again in a post standup discussion. I can't remember where we landed exactly or if we landed somewhere but because tag columns are extra complex (variables types per row) that is separate work and probably something we should design and help implement. I don't remember if specs grid unit requirements are funky or not.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
atmgrifter00: Correct.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
mollykreis: The proposal was for those attributes to apply to all modes, but I've moved those from here down into the _alternatives_ section.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
mollykreis: Since there are still a lot of open questions about units, we've decided to defer the work to define & implement the API associated with them. I've removed them from the API section of this document and added the alternatives we've discussed to the Alternatives section.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
mollykreis: I've applied your suggestion.

> Also maybe we should say that decimal-digits is only used by the decimal formatter?

The docs already say "when format is `decimal`, ....". Do you think we need to be more explicit than that?

change/@ni-nimble-components-a30a4473-2c1e-48e6-9fc4-ae68c2617219.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Does this reduction in scope mean that #1011 will be complete as soon as we remove the roundToInteger format?

jattasNI (approved):

change/@ni-nimble-components-a30a4473-2c1e-48e6-9fc4-ae68c2617219.json
mollykreis: It won't quite be complete, but it'll be really close. I think the open items still are:
- remove roundToInteger format
- localize the column (there's an open PR for this already: #1499)
- add a few things (like alignment) to the Angular directive to ensure it matches the web component API
- implement Blazor support

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: Could be explicit about the behavior in other formats. Maybe "For other formats specifying the decimal digits results in a invalidDecimalDigits column validation error"

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -336,7 +331,56 @@ Nimble already has a mechanism for clients to provide custom columns by deriving
...
+
+We could add attributes for `unit` and `unit-display` that mirror the `unit` and `unitDisplay` configuration options on the `Intl.NumberFormatter`.
+
+**Pros:**
+
+-   Built-in localization support
+
+**Cons:**
+
+-   `Intl.NumberFormatter` supports a fairly limited amount of `unit` values with no way to extend the supported set
```
rajsite, line 348: Can you add a link to the supported units for NumberFormatter?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -370,7 +414,7 @@ We will add standard unit tests, Blazor/Angular tests, and Chromatic tests for n
 
 #### Number column
 
--   number edge cases (-Inf, Inf, -0, +0, NaN, Number.MAX_SAFE_INTEGER + n, Number.MIN_SAFE_INTEGER -n) should render as numbers
+-   number edge cases (-Inf, Inf, -0, +0, NaN, Number.MAX_SAFE_INTEGER + n, Number.MIN_SAFE_INTEGER - n) should render as numbers
```
rajsite, line 417:
```suggestion
-   number edge cases (-Inf, Inf, -0, +0, NaN, Number.MAX_SAFE_INTEGER + n, Number.MIN_SAFE_INTEGER - n) should render as numbers (the exact presentation on the numbers will depend on the formatting options chosen)
```

rajsite (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
mollykreis: Is that the behavior we want? That's not the behavior that's currently implemented. Right now, if the format is anything other than `decimal`, then we don't perform validation on the value of `decimalDigits`. I did that to be consistent with the date-text column, which does validation by trying to create a formatter. As a result, it only validates the values that will be used as part of that formatter.

rajsite: Ah, that was me misinterpreting the validation code when scanning quickly. "For other formats the decimal digits is ignored"

Issue #1436, opened 2023-08-16
Interaction design for tab/toolbar with too much content to fit in screen width


jattasNI:
## 📌 Request Background

There is an [SLE bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/2488661?src=WorkItemMention&src-action=artifact_link) about the expand/collapse toolbar button being placed strangely on small screens. The team that's fixing it requested guidance from us about how to fix it.

![image](https://github.com/ni/nimble/assets/10500124/84579c6c-4777-41a5-b921-c1f408712f5f)

Initial suggestions included adding a way to scroll or growing the height of the toolbar. I've also seen experiences that put items in an overflow menu.

## 🎯 Core Requirements

Ideally we could have a good default behavior without requiring changes to how the content was added to the toolbar. So ideas that use a different presentation like a menu might be difficult because it might require annotating each item with more info like a menu item name.

## 🥅 Acceptance Criteria

- [ ] Interaction Designer comments on this issue with a recommendation

leslieab:
The "..." menu is the overflow menu. If a toolbar doesn't have it, add it. If a toolbar does add it, add the icon buttons to it as items starting with the least important.

jattasNI:
We briefly brainstormed about this in a team meeting today. It'll be a challenge for Nimble to automatically handle it because clients can provide arbitrary toolbar content like icon buttons, select, text, or menus. For some of that there isn't an obvious mapping to how it would render as a menu item.

Instead perhaps Nimble would just give guidance that apps should move content to a menu when it doesn't fit. Then we'd push the responsibility to apps to implement that using CSS container or media queries.

(Just recording brainstorming, not necessarily a decision)

jattasNI:
I captured the guidance above in #1206. There will be lots more questions to answer but that belongs in that bigger mobile support discussion so I'm closing this for now.

Issue #1206, opened 2023-04-24
Audit and improve component behavior on small screens


jattasNI:
## 😯 Problem to Solve

The SLE LIMS app has a requirement that it "work on mobile" but we haven't designed Nimble components with this in mind. This issue is a starting point to track doing the work of auditing components and to track known issues.

**Known issues**
1. Can't access tabs if the screen width is too narrow to see all items. Should we allow horizontal scroll? Show some arrow affordances to indicate that tabs are overflowing?
2. Breadcrumb wraps its content if the screen width is too narrow to see all items. This looked odd to me: I think I expected it to clip, possibly with ... at the start or the middle.
3. Drawer has a fixed width which is often wider than a mobile screen

jattasNI:
The current plan is for @fredvisser to check with SLE POs on the priority of this and then put it on our designers' backlog as appropriate.

m-akinc:
Related: #1436

PR #1480, opened 2023-09-01
Add mobile guidance to toolbar story

jattasNI:
# Pull Request

## 🤨 Rationale

Addresses #1436 

## 👩‍💻 Implementation

The IxD guidance is that content in the toolbar should move to an overflow menu when there is insufficient screen width to display it in toolbar buttons (see comment on linked issue). This is not something Nimble can easily detect, so I'm adding guidance and an example to the toolbar documentation in Storybook.

## 🧪 Testing

Viewed rendered storybook locally.

## ✅ Checklist


jattasNI:
We decided in an offline discussion that this has enough open questions that it's not worth doing now. I'll capture those in #1206.

Issue #1473, opened 2023-08-31
Rich Text Editor | Able to have different type of list element in same level


saikrishnan-ni:
## 🧹 Tech Debt
We're able to have different type of list item on same level for sublist items in `nimble-rich-text-editor` which uses `tiptap` editor. This behavior is from of `tiptap` editor which can be reproduced from the steps followed in this https://github.com/ni/nimble/pull/1462#discussion_r1310933574. We should either find a way to prevent mixing sublist types or have tests for it.

PR #1462, opened 2023-08-29
Rich text editor | Fix nested list highlighting

saikrishnan-ni:
# Pull Request

## 🤨 Rationale
Fixes https://github.com/ni/nimble/issues/1434

- This PR fixes the behavior of list buttons in case of having nested lists. Previously for nested lists, the list buttons used to get highlighted based all all the tags preceding it.
image
- With this change only the list item based on last tag will get highlighted
image

## 👩‍💻 Implementation

- Used `findParentNode` method from tiptap/core to find the current node type
- Highlight the list button only based on current node type

## 🧪 Testing

- Added unit test for testing this use case

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1481, opened 2023-09-04
Rich text editor | Add tests for different list elements in same level behavior

saikrishnan-ni:
# Pull Request

## 🤨 Rationale

- Resolves https://github.com/ni/nimble/issues/1473, where different type of list element is possible to be added in same level. 
It can be reproduced using the steps from https://github.com/ni/nimble/pull/1462#discussion_r1310933574 

## 👩‍💻 Implementation

- Added unit and ui test for the possibility of having different type of list element in same level

## 🧪 Testing

- Added unit and ui testing

## ✅ Checklist


vikisekarNI (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -174,6 +174,18 @@ multipleSubPointsContentInMobileWidth.play = (): void => {
         .setMarkdown(multipleSubPointsContent);
 };
 
+const differentListElementContentInSameLevel = `
+1. Point 1
+    * Sub point 1
+    1. Sub point 2`;
+
+export const differentListElementInSameLevel: StoryFn = createStory(mobileWidthComponent);
```
m-akinc, line 182: I don't know if this visual test is worth having. Does it validate something unique that the unit test doesn't?

For example, if the content was:
```
1. Point 1
    * Sub point 1
    1. Sub point 2
    * Sub point 3
    1. Sub point 4
```
Then I could see that this would show whether sub poin 4 was numbered with a "1" or a "2". But as it is, it seems like this will not verify anything unique.

change/@ni-nimble-components-1b233ea2-a531-4fc0-8b4a-949d74ee0e43.json
m-akinc, line 1: I'm a bit skeptical of this whole PR. Is this really a behavior that we want to assert, or is it just a quirk of the implementation? It seems weird to me that we support mixing list types at the same indentation level. For one, if you interleave bulleted and numbered lists, the numbered list items will keep restarting at "1", which seems a bit odd.

I would lean toward treating mixing list types as undefined behavior that we don't document or test.

@rajsite or @jattasNI  may disagree, though, so I will defer to their judgment.

change/@ni-nimble-components-1b233ea2-a531-4fc0-8b4a-949d74ee0e43.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Added test for know behavior  of having different list types in same level in rich text editor",
```
m-akinc, line 3:
```suggestion
  "comment": "Added test for different list types at same level in rich text editor",
```

m-akinc (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor-matrix.stories.ts
saikrishnan-ni: I'll update the test have the list as suggested and it shows the fourth point to show as `1` considering it a new list after a different list element preceding it. But this case doesn't seem to be manually reproduceable.

change/@ni-nimble-components-1b233ea2-a531-4fc0-8b4a-949d74ee0e43.json
saikrishnan-ni: We had this behavior discussed in https://github.com/ni/nimble/pull/1462#discussion_r1310933574, where it was decided to either resolve or add tests for this behavior. As this behavior is from `Tiptap` and resolving it might need more effort, we've added the tests.

jattasNI (approved):

packages/nimble-components/src/rich-text-editor/tests/rich-text-editor.spec.ts
```diff
@@ -625,6 +625,23 @@ describe('RichTextEditor', () => {
                 'bold, italics and bullet list'
             ]);
         });
+
+        it('should have different type of list at same level possible', () => {
+            element.setMarkdown(`- Bullet List
```
rajsite, line 630: Should we also have a test for the viewer?

change/@ni-nimble-components-1b233ea2-a531-4fc0-8b4a-949d74ee0e43.json
rajsite: Yep, think it's reasonable to add tests for now since it's a state users can get into pretty easily. If we ever change the behavior we'd have to consider how existing comments behave a bit more carefully but hopefully the tests would catch that.

Can we link to a prosemirror or tiptap issue that describes the behavior in a comment by the test? If one doesn't exist we should file one and maybe they'll change the behavior for us. If it's soon enough we could avoid migration concerns.

rajsite (approved):

change/@ni-nimble-components-1b233ea2-a531-4fc0-8b4a-949d74ee0e43.json
rajsite: Looks good! Thanks for filling it 👍

PR #1483, opened 2023-09-05
Fix bug in number-text column where -0 is sometimes displayed

mollykreis:
# Pull Request

## 🤨 Rationale

There are some scenarios in the number-text column when using the `decimal` format where -0 can be displayed even though we attempt to avoid that. Specifically, when the number does not exactly equal -0 but when rounded to show `decimal-digits` decimal places, only zeros will be rendered, then the number will show as `-0.00`.

## 👩‍💻 Implementation

I couldn't find a good way to look at a number and see that it would render only zeros when rounded to `decimal-digits` decimal places. Therefore, the approach I came up with was multiplying the number by 10^decimal-digits and then rounding. If that number is 0, then that means only zeros will be rendered in the number.

## 🧪 Testing

Manually tested
Added new unit test for this case

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table-column/number-text/models/decimal-formatter.ts
+++ b/packages/nimble-components/src/table-column/number-text/models/decimal-formatter.ts
@@ -5,6 +5,7 @@ import { NumberFormatter } from './number-formatter';
  */
 export class DecimalFormatter extends NumberFormatter {
     private readonly formatter: Intl.NumberFormat;
+    private readonly tenPowDecimalDigits: number;
 
     public constructor(decimalsToDisplay: number) {
         super();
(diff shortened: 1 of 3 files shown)


mollykreis:
@atmgrifter00, could I get you to buddy this for me?

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/models/decimal-formatter.ts
```diff
@@ -13,12 +14,19 @@ export class DecimalFormatter extends NumberFormatter {
             minimumFractionDigits: decimalsToDisplay,
             useGrouping: true
         });
+        this.tenPowDecimalDigits = 10 ** decimalsToDisplay;
     }
 
     protected format(number: number): string {
         // The NumberFormat option of `signDisplay: "negative"` is not supported in all browsers nimble supports.
```
jattasNI, line 21: I should have asked for this when the comment was first added, but can you update the comment to explain why we don't want to render -0? Was it because grouping doesn't distinguish between +0 and -0?

jattasNI (approved):

packages/nimble-components/src/table-column/number-text/models/decimal-formatter.ts
mollykreis: We did this with our other two formatters as well (`round-to-integer` and `default`). The thought is that for most cases, people don't think about -0 being different than +0, so we didn't think it made sense for any of our built-in formatters to display -0.

Grouping, however, is an interesting question. Grouping will always be based on the raw value, but it will be rendered based on the formatter. For example, 1.23 and 1.22 with 1 decimal will both render as 1.2, but they won't group together because their values are different. That is a separate concern than -0 though.

rajsite: Yea we might want to make a more descriptive usage note for the grouping docs about this that can be linked to from the number column formatting docs. Think we would need to create a docs page for grouping to do that. Think that would be separate from this PR though.

rajsite (approved):

PR #1490, opened 2023-09-05
Update Storybook dependencies and hide unwanted stories from sidebar

jattasNI:
# Pull Request

## 🤨 Rationale

The latest version of Storybook added a way to hide certain stories from the sidebar. I've been wanting this for "internal" stories that are only used to be composed into higher level MDX docs pages.

https://github.com/storybookjs/storybook/issues/9209#issuecomment-1699329592

image

## 👩‍💻 Implementation

1. Update all Storybook-related dependencies to latest
    - This has the unfortunate side effect of printing a huge string of garbage to the console when you run `npm run storybook -w @ni/nimble-components`. I believe it's caused by https://github.com/storybookjs/storybook/issues/23675 but I'm not clear on how to apply the workaround since we don't appear to use a shared configuration.
3. Add storybook configuration to hide stories that start with "Internal" or "patterns" or "Tests" for the version hosted on nimble.ni.dev (shouldn't affect PR builds or dev builds)
4. Update MDX docs for hidden sections

## 🧪 Testing

I looked to see if there were any other notable features in their [release notes](https://github.com/storybookjs/storybook/releases) or [blogs](https://storybook.js.org/blog/storybook-7-1/) and didn't find much other than official support for the [Figma addon](https://storybook.js.org/blog/figma-plugin-for-storybook/). The only behavior change I noticed is a new badge on the Controls tab that shows how many API entry points we document 🤷‍♀️

1. Inspected local build and PR build of Storybook to ensure expected items are hidden
2. Poked around to see if anything else broke. 

I'll check the nimble.ni.dev/storybook after this deploys to ensure it's actually working, but I already verified the value of `window.location.hostname` in dev tools so I think it'll work.

## ✅ Checklist


packages/nimble-components/.storybook/manager.js
```diff
@@ -3,5 +3,15 @@ import theme from './theme';
 
 addons.setConfig({
     enableShortcuts: false,
+    sidebar: {
+        filters: {
+            patterns: item => {
+                return (
+                    !item.title.startsWith('Internal/')
```
rajsite, line 10: > I looked into conditionally hiding the Tests stories when deployed but couldn't think of a good way to do so. I guess we could add a filter based on the page URL being on nimble.ni.dev but that seemed brittle/hacky to me. Open to suggestions!

I think that sounds like a good option to me. Best balance of nimble dev and end user experience. Nimble devs can easily find the latest storybook link if they want to jump to that to see all the latest matrix tests (which would be hosted on chromatic.com) and removes a bunch of noise from the storybook for end users on `nimble.ni.dev`.

Could even consider only hiding all of them only on `nimble.ni.dev` so it's easier to poke around the internal stories if needed from a chromatic run.

packages/nimble-components/docs/internal.mdx
```diff
@@ -1,8 +0,0 @@
-import { Meta } from '@storybook/addon-docs';
```
rajsite, line 1: If we go with hiding these only on nimble.ni.dev then maybe we keep this around for devs. Maybe even add a patterns doc as well that contains the same note.

packages/nimble-components/.storybook/manager.js
rajsite: I think another option would be to create a global where the default value hides all the internal stories: https://storybook.js.org/docs/react/essentials/toolbars-and-globals

I think that we should be able to access the value of globals from manager.js but not 100% certain

jattasNI: I went with `window.location.hostname === 'nimble.ni.dev'` after trying it out in dev tools.

rajsite (approved):

PR #1493, opened 2023-09-06
Create and leverage shared doc string for column operation behavior

mollykreis:
# Pull Request

## 🤨 Rationale

This was an action item from https://github.com/ni/nimble/pull/1483#discussion_r1316503913

We should ensure that we have consistent docs for how grouping behaves for columns that display formatted values in their cells. This also applies to sorting.

## 👩‍💻 Implementation

Exported a new string from `table-column-stories-utils.ts` that explains that grouping & sorting is done on the data field value.

## 🧪 Testing

Verified the modified stories looked correct.

## ✅ Checklist


mollykreis:
@m-akinc, can I get a buddy review for these change?

packages/nimble-components/src/table-column/base/tests/table-column-stories-utils.ts
```diff
@@ -50,3 +50,5 @@ export const sharedTableActions = [
     'selection-change',
     'column-configuration-change'
 ] as const;
+
+export const columnOperationBehavior = 'Column operations, such as sorting and grouping, are preformed on the field values in the data records, not by the formatted values displayed within a cell.';
```
m-akinc, line 54:
```suggestion
export const columnOperationBehavior = 'Column operations, such as sorting and grouping, are performed on the field values in the data records, not on the formatted values displayed within the cells.';
```

m-akinc (approved):

rajsite (approved):

change/@ni-nimble-components-5e4aad3f-b6d2-47fa-8d61-7c6f4415f6b9.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: While you're at it, could you also update `table-column.stories.ts` to ensure that both the dedicated grouping and sorting stories have similar language? Looks like the sorting one already says "Sorting is based on the underlying field values in the column, not the rendered values" but `groupedRowsDescription` doesn't mention it.

jattasNI (approved):

Issue #1372, opened 2023-07-21
Create a new anchor pattern to share styles across nimble components


vivinkrishna-ni:
## 🧹 Tech Debt

Rationale: https://github.com/ni/nimble/pull/1335#discussion_r1268307649

There are instances where we aim to maintain uniform anchor styles across various Nimble components, especially when using links similar to the one mentioned in the previous comment. This is particularly relevant in situations where incorporating the actual `nimble-anchor` becomes challenging. Therefore, we can transfer the [default anchor styles](https://github.com/ni/nimble/blob/86f2f56c93053c132f96fd906bab2c7e535f866c/packages/nimble-components/src/anchor/styles.ts#L27C1-L54C6) to the [patterns](https://github.com/ni/nimble/tree/main/packages/nimble-components/src/patterns) and apply them consistently across components as required.

rajsite:
Not sure how the features are related but the prose mirror marks docs for links also discuss how to customize the html of marks of which links are one type of mark: https://prosemirror.net/examples/schema/#:~:text=Links

vivinkrishna-ni:
Thanks for sharing the references. We could possibly use `nimble-anchor` in both `nimble-rich-text-editor` and `nimble-rich-text-viewer` with the above TipTap configurations and updating the Prosemirror schema. It is incorporated in this PR: https://github.com/ni/nimble/pull/1496

However, the `nimble-anchor` in the editor is clickable by default whereas the `a` tag default behavior inside `contenteditable` div is non-clickable. For more details, refer to this issue: https://github.com/ni/nimble/issues/1502

vivinkrishna-ni:
In addition to the above-linked issue, we have captured a few more `nimble-anchor` behavior in `nimble-rich-text-editor` at the first pass of development that has to be fixed before rendering it in the editor component. Here is the tech debt to the issue: https://github.com/ni/nimble/issues/1516.

But we could use the `nimble-anchor` in `rich-text-viewer` as it is not an editable component.

PR #1492, opened 2023-09-06
Rich text editor | Move markdown functionalities into respective models

saikrishnan-ni:
# Pull Request

## 🤨 Rationale

- Created models for markdown functionalities to be used in rich text editor and text viewer components.

## 👩‍💻 Implementation

- Moved markdown parsing and serializing functionalities from the rich text editor and viewer components into respective models.
- `markdown-parser.ts` - Initialize the Prose Mirror markdown parse and expose the `parseMarkdownToDOM` method for parsing the markdown string into a corresponding document fragment, which can then be utilized by both the rich text components to load it in the UI.
- `markdown-serializer.ts` - Initialize the Prose Mirror markdown parse and expose the `serializeDOMToMarkdown` method to convert the rendered node in the editor to a markdown string.

Related discussion: https://github.com/ni/nimble/pull/1424#discussion_r1290998178

## 🧪 Testing

- Migrated all the markdown parsing unit test cases from the components into `markdown-parser.spec.ts`

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1496, opened 2023-09-07
Rich text editor & viewer | Support for absolute link

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

1. Add absolute links in the editor by either pasting or manually entering a URL (HTTP/HTTPS).
2. The input and output for absolute links in Markdown follow the [AutoLink ](https://spec.commonmark.org/0.30/#autolink) format in the CommonMark flavor i.e. `<URL>`.
3. Links in the viewer will render as `nimble-anchor`. Part of https://github.com/ni/nimble/issues/1372
4. Links in the editor will render as `<a>` as we have few constraints in rendering `nimble-anchor` in the editor. https://github.com/ni/nimble/issues/1516

#### Rich text editor:
1. Installed [@tiptap/extension-link](https://tiptap.dev/api/marks/link) to enable link in editor.
2. Configuring the following settings for the link in the editor:
    - Preventing users from opening a link in the editor by disabling the [`openOnClick`](https://tiptap.dev/api/marks/link#open-on-click)
    - Disallowing users from pasting a link to an already selected word by disabling [`linkOnPaste`](https://tiptap.dev/api/marks/link#link-on-paste), especially since hyperlinks are not supported for the initial pass.
    - Add Regex to [validate](https://tiptap.dev/api/marks/link#validate) that the entered/parsed URL conforms to the HTTPS/HTTP protocol. URLs not using HTTPS/HTTP will be displayed as plain text.
5. Adding custom link serializer for autolink in `markdown-serializer.ts`.
6. Preventing the application of additional formatting, such as bold or italics, to a link by using [`excludes: '_'`](https://prosemirror.net/docs/ref/#model.MarkSpec.excludes). This is because CommonMark's [autolink](https://spec.commonmark.org/0.30/#autolink) does not support the simultaneous application of another formatting within the link, making it impossible to render a specific part of the link in autolink markdowns, like  `<https://**nimble**.ni.dev>`.

#### Rich text viewer (or markdown-parser):
1. Rendering links as `nimble-anchor` by updating the schema in `markdown-parser.ts`.
2. [Validate](https://markdown-it.github.io/markdown-it/#MarkdownIt.prototype.validateLink) if the links in the input markdown string are HTTPS/HTTP.
3. [Normalize the link text](https://markdown-it.github.io/markdown-it/#MarkdownIt.prototype.normalizeLinkText) to render the link as is, instead of updating the link text if it contains any encoded or non-ASCII characters.

## 🧪 Testing
1. rich-text-editor.spec.ts:
    1. "Absolute link interactions in the editor" - These tests cover user interactions with the editor like typing or copying and pasting the links. This includes testing the `validate` in index.ts (only HTTP/HTTPS), assessing links within various nodes, and restricting the application of other marks (bold/italics) to the links within the editor.
    2. "Absolute link markdown tests" - These tests ensure the parsed links are rendered as `<a>` tags and not as `nimble-anchor` within the editor. This distinction arises because we have adjusted the schema of the link in the `markdown-parser.ts` to parse it as a `nimble-anchor`. It also covers `getMarkdown` is the same as `setMarkdown` for links similar to the tests pattern we followed for other marks and nodes. 
2. markdown-parser.spec.ts:
    1. "Absolute link" - These tests cover more in-depth of link formats that could possibly converted into links when parsed from a markdown string. It verifies `autolink`, `validateLink`, and `normalizeLinkText` in tokenizer rules and configurations set in the `markdown-parser.ts`.
3. markdown-serializer.spec.ts: 
    1. These tests ensure only the text content(not `href`) in the `<a>` tag is serialized to autolink markdown string.
    2. It also covers how other marks (bold/italics) are ignored when the link is wrapped within it.

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -188,6 +194,23 @@ export const styles = css`
...
+        ${
+            /**
+             * Restricting the pointer events for the following reasons:
+             * 1. nimble-anchor inside a "contenteditable" div is not working as native HTML anchor tag.
+             *    i.e. clicking on the link opens in the same tab whereas the default behavior of native HTML anchor
+             *    tag is not clickable inside "contenteditable" div.
+             * 2. Restricting the user from opening a link using the right-click context menu: If the user manually edits
+             *    the link, the 'href' attribute of the anchor tag will not be updated. If they attempt to open it using
+             *    the right-click context menu with 'Open in new tab/window,' it will still navigate to the link specified
+             *    in the 'href' attribute.
```
vikisekarNI, line 208:
```suggestion
             * 2. Restricting the user from opening a link using the right-click context menu: If the user manually edits
             *    the link's text content, the 'href' attribute of the anchor tag will not be updated. If they attempt to open it using
             *    the right-click context menu with 'Open in new tab/window,' it will still navigate to the link specified
             *    in the 'href' attribute, which may create unnecessary confusion while trying to open the link
             *  
```

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -46,10 +50,39 @@ export class RichTextMarkdownParser {
...
+
         return new MarkdownParser(
-            schema,
+            this.updatedSchema,
             supportedTokenizerRules,
             defaultMarkdownParser.tokens
         );
     }
+
+    private static getUpdatedSchema(): Schema {
```
vikisekarNI, line 62:
```suggestion
    private static getSchemaWithLinkConfiguration(): Schema {
```

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -51,7 +51,15 @@ export class RichTextMarkdownSerializer {
         };
         const marks = {
             italic: defaultMarkdownSerializer.marks.em!,
-            bold: defaultMarkdownSerializer.marks.strong!
+            bold: defaultMarkdownSerializer.marks.strong!,
+            // Autolink markdown in CommonMark flavor: https://spec.commonmark.org/0.30/#autolinks
```
vikisekarNI, line 55:
```suggestion
            // Autolink markdown in CommonMark flavor: https://spec.commonmark.org/0.30/#autolinks
            // Prosemirror model reference: https://github.com/ProseMirror/prosemirror-markdown/blob/c7210d0e55c82bfb0b2f7cba5dffe804575fafb3/src/to_markdown.ts#L3C1-L26C2
```

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -21,6 +21,24 @@ describe('Markdown parser', () => {
...
+        doc: DocumentFragment
+    ): Element | null | undefined {
+        let lastElement = doc.lastElementChild;
+
+        while (lastElement?.lastElementChild) {
+            lastElement = lastElement?.lastElementChild;
+        }
+        return lastElement;
+    }
+
```
vikisekarNI, line 41: For reviewer's info: Hoping this PR will take updates from [comment in another PR](https://github.com/ni/nimble/pull/1492#discussion_r1318953099) which directs to move these utility functions under `testing` folder

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -230,9 +248,185 @@ describe('Markdown parser', () => {
             ).toEqual(['Bulleted list in italics']);
         });
 
+        describe('Absolute link', () => {
+            it('absolute link markdown string to "a" tags with the link as the text content', () => {
```
vikisekarNI, line 252:
```suggestion
            it('absolute link markdown string to "nimble-anchor" tags with the link as the text content', () => {
```

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -51,11 +55,31 @@ describe('Markdown serializer', () => {
                 html: '<em>Italics</em>',
                 markdown: '*Italics*'
             },
+            {
+                name: 'Link',
+                html: '<p><a href="#">Link</a></p>',
```
vikisekarNI, line 60: Consider leaving a comment stating why we have `a` tag in serialization. I guess tiptap will convert this to `nimble-anchor` using `renderHTML`

vikisekarNI (approved):

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: I have moved these to `testing/markdown-parser-utils.ts` file

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
vivinkrishna-ni: Added a comment here

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -642,6 +642,160 @@ describe('RichTextEditor', () => {
                 'P'
             ]);
         });
+
+        describe('Absolute link interactions in the editor', () => {
```
m-akinc, line 646: Should have a test for mixed-case or uppercase link e.g. ``HTTP://foo``.

m-akinc, line 646: Should have a test for URL starting with just ``http`` (all these tests use ``https``).

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -77,6 +77,12 @@ export const styles = css`
         border: ${borderWidth} solid rgba(${borderRgbPartialColor}, 0.1);
     }
 
+    :host([disabled]) nimble-anchor::part(control) {
```
m-akinc, line 80: This should be moved next to (after) the other `nimble-anchor` styling.

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -350,11 +355,46 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                customLink.configure({
+                    // HTMLAttribute cannot be in camelCase as we want to match it with the name in Tiptap
+                    // eslint-disable-next-line @typescript-eslint/naming-convention
+                    HTMLAttributes: {
+                        rel: 'noopener noreferrer',
+                        target: null
+                    },
+                    openOnClick: false,
+                    linkOnPaste: false,
+                    validate: href => /^https?:\/\//.test(href)
```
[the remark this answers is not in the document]
vivinkrishna-ni: It should be case-sensitive. I have updated the same.

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -109,15 +111,18 @@ export class RichTextEditorPageObject {
     }
 
     public async setEditorTextContent(value: string): Promise<void> {
-        let lastElement = this.getTiptapEditor()?.lastElementChild;
-
-        while (lastElement?.lastElementChild) {
-            lastElement = lastElement?.lastElementChild;
-        }
+        const lastElement = this.getEditorLastChildElement();
         lastElement!.parentElement!.textContent = value;
```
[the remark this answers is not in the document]
vivinkrishna-ni: Done

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Added

vivinkrishna-ni: Done

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -642,6 +642,160 @@ describe('RichTextEditor', () => {
...
+            it('should change the text to "nimble-anchor" tag when it is a valid absolute link', async () => {
+                await pageObject.setEditorTextContent(
+                    'https://nimble.ni.dev/ '
```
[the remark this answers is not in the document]
vivinkrishna-ni: > It is kind of weird that you don't get a link until you've added whitespace after the URL.

We could see this is an expected behavior in every WYSIWYG editor we tried. Here is the [Tiptap issue](https://github.com/ueberdosis/tiptap/issues/3225) where they explicitly enabled the links only when whitespaces are added or by pressing the `Enter` key and not for focusing in and out the editor. Even in other editors like AzDo (only WYSIWYG editor, not markdown editor), enabling links only after they add a whitespace or a new line.

> But if I do `editor.setMarkdown('<http://abc>')`, I can get a rendered link without any whitespace after it.

This is also an expected behavior when we relate it with the markdown parser we use to convert the markdown string to DOM. In this case, when we do `editor.setMarkdown('<http://abc>')`, the string is already parsed into a link according to [Autolink](https://spec.commonmark.org/0.30/#autolinks) in CommonMark flavor. Whereas in the above case, the editor might not know whether the newly typed text is a valid link or not.

We also couldn't find a way to make any configurations to make the editor detect links for every keystroke and it might also create confusion or misinterpretation. For example, when the user wants to type URL `https://nimble.ni.dev`, if it is detected in every keystroke, it will change it to a link after it reaches `https://n`. Also, it becomes difficult for us to modify the `href` of the link based on every update. (As we don't have any solution for now to update `href` based on the `<a>` tag text contents)

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -350,11 +355,46 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+     * Extending the default link mark schema defined in the TipTap.
+     *
+     * "exclude": https://prosemirror.net/docs/ref/#model.MarkSpec.excludes
+     * "inclusive": https://prosemirror.net/docs/ref/#model.MarkSpec.inclusive
+     * "parseHTML": https://tiptap.dev/guide/custom-extensions#parse-html
+     * "renderHTML": https://tiptap.dev/guide/custom-extensions/#render-html
+     */
+    private getCustomLinkExtension(): Mark<LinkOptions> {
+        return Link.extend({
+            excludes: '_',
```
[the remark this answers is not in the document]
vivinkrishna-ni: This `excludes: '_'` is an intentional configuration due to the markdown output we support. As the [Autolink](https://spec.commonmark.org/0.30/#autolinks) does not allow other formatting within the link, we have to restrict adding other formatting to the absolute link in the editor too. 

> Would there be a way to force `nimble-anchor` to be the outermost element if we have a bold/italic link?

Yes, this is how the default priority of the links will be established in relation to other formatting options such as bold and italics. However, this can only be enabled once we have implemented support for [Links(CommonMark doc)](https://spec.commonmark.org/0.30/#links) which are in a format like this: `[Lin text](https://foo)`. At present, we don't support this in markdown and will be planned in future updates as mentioned in the [spec](https://github.com/ni/nimble/tree/main/packages/nimble-components/src/rich-text/specs#:~:text=Support%20for%20adding%20hyperlinks%20to%20the%20existing%20text%20in%20the%20editor.%20This%20allows%20users%20to%20add%20links%20to%20existing%20text%20in%20the%20editor.%20When%20the%20link%20button%20in%20the%20formatting%20options%20is%20clicked%2C%20a%20dialog%20opens%2C%20providing%20a%20space%20to%20enter%20the%20hyperlink%20for%20the%20selected%20text.). Once it is enabled, we can remove this `excludes: '_'` from both the editor config and the parser schema. At that time, if we set the link to a bold text the markdown output will be something like `[**text**](https://foo)`

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: @m-akinc Thanks for testing the link behavior in the editor and capturing some of the key points to be addressed. After discussing these points in detail internally and considering the timeline of this feature support, we have decided to render a native anchor instead of a `nimble-anchor` only in the editor which provides a fix to many of the linking problems stated above. 

Points `#1`, `#3`, `#4`, and `#5`, which you mentioned, will be resolved when we switch them back to using the native anchor tag. After rendering the native anchor tag,
1. We can get out of the ends of the link in the editor when the cursor is inside and can enter the link using the keyboard when the cursor is outside. 
2. There is no extra space around links as mentioned in the created bug.
3. If we copy and paste a portion of an existing URL, it will be displayed as plain text rather than as a link, which is the expected behavior. This could be attributed to the default paste rules of link settings when rendering native anchors in the Tiptap editor. However, when copying includes the trailing whitespace at the end of the link and then pasting it, the link is rendered as expected.
4. We can highlight the link by clicking and dragging with the mouse, as well as by using various keyboard methods to select the link.

We think that rendering `nimble-anchor` inside the editor might require further explorations to address all the issues that arise within the editor. So we thought of creating a tech debt issue to render `nimble-anchor` inside the editor. In the issue, we can capture all the above points and create a separate PR to resolve all these problems in the right place either in `nimble-anchor` or in the `nimble-rich-text-editor` (or Tiptap). Please let us know your thoughts.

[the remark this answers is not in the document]
vivinkrishna-ni: > 2. Putting the cursor at the beginning or end of the link and typing adds text outside the link. If you delete the character at the end of your link, you cannot easily add it back.

It is because of setting false to `inclusive` in the link configuration. As we did not find the best way to update the `href` of the link when the text content updates, we mostly restrict the user from editing the link. As we don't support the hyperlink feature, for now, we wanted to restrict the user in whatever way it is possible.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -687,6 +841,193 @@ describe('RichTextEditor', () => {
         ]);
     });
 
+    describe('Absolute link markdown tests', () => {
+        describe('asserting rendered links in the editor', () => {
```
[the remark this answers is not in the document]
vivinkrishna-ni: As mentioned in the [other comment](https://github.com/ni/nimble/pull/1496#discussion_r1321344449), we are planning to render a native anchor tag for the editor but the parser will have a `nimble-anchor` node. So we thought it would be better to check if it renders the native anchor tag here and in `markdown-parser.spec.ts`, the parsed content contains `nimble-anchor`. 

We can remove the redundant tests once and for all when we decide to render `nimble-anchor`.

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -188,6 +194,24 @@ export const styles = css`
...
+        white-space: normal;
+        ${
+            /**
+             * Restricting the pointer events for the following reasons:
+             * 1. nimble-anchor inside a "contenteditable" div is not working as native HTML anchor tag.
+             *    i.e. clicking on the link opens in the same tab whereas the default behavior of native HTML anchor
+             *    tag is not clickable inside "contenteditable" div.
+             *    Issue link: https://github.com/ni/nimble/issues/1502
+             * 2. Restricting the user from opening a link using the right-click context menu: If the user manually edits
+             *    the link's text content, the 'href' attribute of the anchor tag will not be updated. If they attempt to open it using
```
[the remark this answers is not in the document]
vivinkrishna-ni: Since there might be any number of links in the editor, we couldn't find a better way to update the `href` of the current link. Even the default linking behavior in Tiptap doesn't support updating the link. 
absolute-link-tiptap

When I am looking into some of the other WYSIWYG editors, they usually don't update the `href` when updating the text content. As the other editors might be supporting the link markdown format `[text](https://foo)`. So, until we enable `link` in markdown, we might have to have this workaround.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -46,10 +51,39 @@ export class RichTextMarkdownParser {
...
+                            anchorTag,
+                            {
+                                href: node.attrs.href as Attr,
+                                rel: node.attrs.rel as Attr
+                            }
+                        ];
+                    }
+                },
+                em: schema.spec.marks.get('em')!,
+                strong: schema.spec.marks.get('strong')!
```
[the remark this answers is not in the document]
vivinkrishna-ni: This isn't working as the `schema.spec.marks` is of type `OrderedMap<MarkSpec>`. The `OrderedMap` contains the `size` argument which is not accepted by the `marks` in the parent object. So created a separate mark here and used the `get()` method to get the mark in `MarkSpec` type and load it here.

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
m-akinc: > So we thought of creating a tech debt issue to render nimble-anchor inside the editor. In the issue, we can capture all the above points and create a separate PR to resolve all these problems in the right place either in nimble-anchor or in the nimble-rich-text-editor (or Tiptap)

Given that a native anchor does not have most of the issues I noted, I think it's reasonable to use that and create the tech debt item like you said.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -63,7 +63,7 @@ const component = (
         ${() => footerHiddenName} ${() => errorStateName} ${() => placeholderName} ${() => disabledName} 
     </p>
     <${richTextEditorTag}
-        style="margin: 5px 0px; width: 500px;"
+        style="margin: 5px 0px; width: 500px; height: 100px"
```
jattasNI, line 66: The changes to the component height in this PR are making it difficult to inspect the Chromatic diff and see what changed related to links. What do you think about splitting them into a separate PR so that the link changes are isolated? We could push that one through quickly first thing Tuesday morning (NIC time).

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -188,6 +189,31 @@ export const styles = css`
         color: ${controlLabelDisabledFontColor};
     }
 
+    .ProseMirror a {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the catch, I should've added disabled styles to the `a` tag after reverting the changes from rendering `nimble-anchor` to the `a` tag. I have updated the style now.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-matrix.stories.ts
vivinkrishna-ni: Created a PR for setting the heights in the storybook. https://github.com/ni/nimble/pull/1513

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: I have created a tech debt issue and captured all these points there: https://github.com/ni/nimble/issues/1516

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -210,9 +211,175 @@ describe('Markdown parser', () => {
...
+
+                expect(getTagsFromElement(doc)).toEqual(['P', 'NIMBLE-ANCHOR']);
+                expect(getLeafContentsFromElement(doc)).toEqual([
+                    'https://**nimble**.ni.dev/'
+                ]);
+                expect(getLastChildElementAttribute('href', doc)).toBe(
+                    'https://**nimble**.ni.dev/'
+                );
+            });
+
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for letting us know about different cases for links. I have updated the test cases in `markdown-parser.spec.ts` to test the behavior of the above-listed links are parsed as expected, considering we support only absolute links for now.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -350,11 +355,48 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+            inclusive: false,
+            parseHTML() {
+                return [{ tag: anchorTag }];
```
jattasNI, line 388: Is it correct that this is still using `nimble-anchor`? Or should it be symmetric with the `'a'` tag in `renderHTML`?

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
jattasNI: Could you update the PR description to capture the decision that links will use the `<a>` tag in the editor but `nimble-anchor` in the viewer? Right now it still says we're using the nimble anchor for both.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -642,6 +642,186 @@ describe('RichTextEditor', () => {
                 'P'
             ]);
         });
+
+        describe('Absolute link interactions in the editor', () => {
+            describe('various absolute links without other nodes and marks', () => {
```
jattasNI, line 647: Could you provide some background about how you decided which tests go in the element spec files, which go in the model spec files, and which go in both? The philosophy would be helpful to add to the Testing section of the PR description.

I haven't gone through thoroughly but I get a sense that we might be testing some functionality at the element level which could be tested at the model level and that we might be duplicating some cases between both. Ideally we'd be covering cases at the lowest possible level (model) and not duplicating cases, but I recognize that won't always be possible.

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -188,6 +189,26 @@ export const styles = css`
...
+        color: ${linkFontColor};
+        white-space: normal;
+        ${
+            /**
+             * Setting 'pointer-events: none;' to restrict the user from opening a link using the right-click context menu: If the user manually edits
+             * the link's text content, the 'href' attribute of the anchor tag will not be updated. If they attempt to open it using
+             * the right-click context menu with 'Open in new tab/window,' it will still navigate to the link specified
+             * in the 'href' attribute, which may create unnecessary confusion while trying to open the link.
```
rajsite, line 200: What's going to fix this behavior? Do we need to submit a bug to prose mirror / tiptap? Is there an associated nimble issue to link to?

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -642,6 +642,186 @@ describe('RichTextEditor', () => {
...
+                        // eslint-disable-next-line @typescript-eslint/no-loop-func
+                        async () => {
+                            await pageObject.setEditorTextContent(value.name);
+
+                            expect(pageObject.getEditorTagNames()).toEqual([
+                                'P',
+                                'A'
+                            ]);
+                            expect(pageObject.getEditorLeafContents()).toEqual([
+                                value.name.slice(0, -1)
```
rajsite, line 682:
```suggestion
                                // Name without the trailing space used by the editor to trigger conversion to a link
                                value.name.slice(0, -1)
```

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
rajsite: These comments don't explain why we can't just leverage the `defaultMarkdownSerializer.marks.link` and are forking it with our own implementation.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -46,10 +51,44 @@ export class RichTextMarkdownParser {
             'autolink'
         ]);
 
+        supportedTokenizerRules.validateLink = href => /^https?:\/\//i.test(href);
+
+        // In order to display encoded characters, non-ASCII characters, emojis, and other special characters in their original form,
+        // we bypass the default normalization of link text in markdown-it. This is done because we support only "AutoLink" feature in CommonMark flavor.
+        // "normalizeLinkText" method reference in markdown-it: https://github.com/markdown-it/markdown-it/blob/2b6cac25823af011ff3bc7628bc9b06e483c5a08/lib/index.js#L67C1-L86C2
+        supportedTokenizerRules.normalizeLinkText = url => url;
```
rajsite, line 59: So what's going to happen when we support more than the autolink feature? Won't this impact both autolink and normal link cases?

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Updated

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: I have updated the `testing` section of the PR description briefly about the approach followed in each file. Please let us know if you need more details or something we should consider changing in it.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: `parseHTML` should include `nimble-anchor` because the markdown parser is shared between both the editor and the viewer. After setting the markdown, the parsed content will contain only `nimble-anchor` and during the rendering process in the editor, `renderHTML` will change the link node tag to `<a>` tag.

If we update `a` in `parseHTML` now, the editor will not identify that it has a link because the parsed node will not have any `a` but instead, it will have `nimble-anchor`. So link will render as plain text in the editor while parsing.

Another advantage of having this right now is, that if we change it to `nimble-anchor` in the future it is a simple one-line code change in `renderHTML` and we can avoid modifying the entire schema in the parser.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
vivinkrishna-ni: I updated the comment with more details

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: This code needs an update when we enable normal link support. I have mentioned this along with other updates here to track the work: https://github.com/ni/nimble/issues/1527

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -664,6 +664,187 @@ describe('RichTextEditor', () => {
...
+                            expect(pageObject.getEditorLeafContents()).toEqual([
+                                // Name without the trailing space used by the editor to trigger conversion to a link
+                                value.name.slice(0, -1)
+                            ]);
+                        }
+                    );
+                }
+            });
+
+            it('should have the right attributes to the "a" tag', async () => {
```
jattasNI, line 712: nit: prefer specific descriptions rather than words like "right" or "correct" (we haven't been too strict about this but I happened to notice this one)

```suggestion
            it('the "a" tag should have href and rel attributes', async () => {
```

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
jattasNI: Thanks, that description is helpful.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -47,10 +52,44 @@ export class RichTextMarkdownParser {
...
+    private static getSchemaWithLinkConfiguration(): Schema {
+        return new Schema({
+            nodes: schema.spec.nodes,
+            marks: {
+                link: {
+                    attrs: {
+                        href: {},
+                        rel: { default: 'noopener noreferrer' }
+                    },
+                    inclusive: false,
```
rajsite, line 78: Same as above:
Will this inclusive configuration also need to change when hyperlink support is added?
If so can you add that in a comment?
If not can you describe why we are making this configuration change from the default?

rajsite (approved):

packages/nimble-components/src/rich-text/editor/styles.ts
rajsite: Discussed with the code owners and the two things that seemed important were to capture the tech debt being introduced to workaround hyperlink issues in an issue which was done and in source for all the workarounds that were added to include a link to the issue that when resolved should remove them.

I went through and placed a comment on all the code that seems like workarounds / tech debt added for the absolute url only case that will be cleaned up / removed when full hyperlink support is added. Please make sure all of them are commented. 👍

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Added a comment here

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -51,11 +55,33 @@ describe('Markdown serializer', () => {
                 html: '<em>Italics</em>',
                 markdown: '*Italics*'
             },
+            {
+                // All links will have `<a>` tag here as an input, since it is a mock editor.
+                // In the actual editor, links will render as `nimble-anchor` using `renderHTML`.
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, it renders `a` tag. I have removed this comment from here.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -350,11 +355,48 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+     *
+     * "excludes": https://prosemirror.net/docs/ref/#model.MarkSpec.excludes
+     * "inclusive": https://prosemirror.net/docs/ref/#model.MarkSpec.inclusive
+     * "parseHTML": https://tiptap.dev/guide/custom-extensions#parse-html
+     * "renderHTML": https://tiptap.dev/guide/custom-extensions/#render-html
+     */
+    private getCustomLinkExtension(): Mark<LinkOptions> {
+        return Link.extend({
+            excludes: '_',
+            inclusive: false,
```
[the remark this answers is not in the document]
vivinkrishna-ni: We will defer this decision on changing this configuration when hyperlink support is enabled. Therefore, I have added a comment here to revisit this when the hyperlink is added. Also updated the task list in the hyperlink support issue.

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -188,6 +189,26 @@ export const styles = css`
         color: ${controlLabelDisabledFontColor};
     }
 
+    .ProseMirror a {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Added

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -664,6 +664,187 @@ describe('RichTextEditor', () => {
                 'P'
             ]);
         });
+
+        describe('Absolute link interactions in the editor', () => {
+            describe('various absolute links without other nodes and marks', () => {
+                const supportedAbsoluteLink: { name: string }[] = [
+                    { name: 'https://nimble.ni.dev/ ' },
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have added the test cases here and in the markdown parser. However, when we add this link with emojis in the editor, the `href` attribute is not updated to Punycode encoded URL. In the parser, the URL is parsed as per your expectation and so added a test case with encoded value to the `href` attribute.

I also checked the other editors like AzDo, and CKEditor which render the same as here i.e. URLs with emoji not encoded URLs.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: As mentioned in the [other comment](https://github.com/ni/nimble/pull/1496#discussion_r1325915858), we will take the right decision when link support is added. I have updated the issue checklist too.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -22,7 +23,10 @@ describe('Markdown serializer', () => {
             OrderedList,
             ListItem,
             Bold,
-            Italic
+            Italic,
+            Link.extend({
```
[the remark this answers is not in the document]
vivinkrishna-ni: The tests are written to isolate the serializer from the component, yet they also require Tiptap's Node as an input to the `serializeDOMToMarkdown` method. However, this issue needs to be addressed, and in the same time we do not want to block this PR. We have created a tech debt issue, https://github.com/ni/nimble/issues/1533, and have started working on it to come up with a better testing strategy for the markdown-serializer. Please share your thoughts on the tech debt issue we've created.

> At this point can we just construct a `nimble-rich-text-editor` and use that with it's `setMarkdown` instead?

For links, these tests are already added in the describe block `asserting getMarkdown for rendered links` in the `nimble-rich-text-editor.spec.ts` file. We are also in the discussion of the right place to have these tests because constructing the actual `nimble-rich-editor` here is the same as writing tests in the component is what we thought.

> My concern is that if that's the case, what's the guarantee of stability of the HTML snippets captured in this file?

This is because we are using the `setContent` of Tiptap which accepts HTML strings in any form that can render as we wanted in the editor. For example, when I tried setting the `<em><strong>Bold</strong></em>` (which is not how it renders in editor DOM) through `setContent`, it renders as `<strong><em>Bold</em></strong>` based on the priority set in the Tiptap schema. 

> That's not a valid absolute link. What is this trying to even assert?

Yes, this should not be done in this way in the test files and I have updated all the text content used here to proper absolute links. The point we tried to test here is that the content of the `<a>` tag is just wrapped with the `<>` brackets when getting the markdown string to it due to the constraints we have with the prosemirror-markdown serializer. We added a comment about the rationale we went with that approach in `markdown-serializer.ts`.

To conclude, we will revisit and update the testing strategy for the serializer in a separate PR and we will consider all of the above and make sure to come up with the right possible way before we make major changes to the editor component.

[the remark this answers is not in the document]
rajsite: Sounds good. And a valid option may be that we find independently testing the markdown serializer to not be valuable. We never will pass HTML to it directly, only the TipTap editor will. So the test we want is that interactions with the tiptap editor produce expected markdown in the end. The HTML intermediate representation is a detail we don't really care about much as long as the same interactions in the editor produce the same markdown. But yes, continue discussion in the tech debt task.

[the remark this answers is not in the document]
vivinkrishna-ni: Sure, we will also see how we can test the anchors in Chromatic tests.

PR #1499, opened 2023-09-07
Add locale support to number-text column

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1011 

This PR adds support for the number-text column using the `lang` token on the theme provider to format numbers in a locale-aware way.

## 👩‍💻 Implementation

I followed the same pattern used in the date-text column to listen for changes to the `lang` theme provider token and pass that into the formatters for the number-text column.

## 🧪 Testing

- New unit test for the column that verifies the formatter is updated with the `lang` changes
- New unit tests for each formatter that verify the locale is used when formatting
    - Note: I updated the formatter tests to run all the test cases with an additional locale. On one hand this potentially seemed like "too many" locale tests, but I thought it was the best way to ensure that the locale was used in all internal formatters (e.g. the `default` formatter uses 3 formatters internally) and that we had coverage for a wide range of possible values to be formatted.

## ✅ Checklist


packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
```diff
@@ -2,81 +2,134 @@ import { getSpecTypeByNamedList } from '../../../../utilities/tests/parameterize
 import { DecimalFormatter } from '../decimal-formatter';
 
 describe('DecimalFormatter', () => {
+    const locales = ['en', 'de'] as const;
```
mollykreis, line 5: `en` and `de` are somewhat arbitrary choices for locales to test with. I also could have expanded this list to include a bunch more locales (or even just the ones that NI traditionally translates to). However, I decided to keep things relatively simple and test with 2 locales. I think this gives us the coverage we want, which is ensuring that all our formatters are using the specified locale without adding a bunch of extra tests that are only testing that the `Intl.NumberFormat` behaves the way we expect for each locale.

mollykreis:
@m-akinc, do you mind buddying these changes?

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -46,13 +50,21 @@ export class TableColumnNumberText extends TableColumnTextBase {
...
+
     public override connectedCallback(): void {
         super.connectedCallback();
+        lang.subscribe(this.langSubscriber, this);
         this.updateColumnConfig();
     }
 
     public override get validity(): TableColumnValidity {
         return this.validator.getValidity();
+        lang.unsubscribe(this.langSubscriber, this);
```
m-akinc, line 67: Is this a mistake?

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
m-akinc: Even this seems like overkill. I would have left the test cases using one locale, and just added one extra test that changes the locale to make sure the formatted value adjusts accordingly. Still, I don't think it's a big deal to over-test a bit.

m-akinc (approved):

packages/nimble-components/src/table-column/number-text/index.ts
mollykreis: Yes...this should be in the `disconnectedCallback` here. I'm not sure how that happened. 🤦‍♂️

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
mollykreis: I'll leave this discussion open so that owners can weigh in with their opinions. If no one has a strong opinion, I'll leave this as-is.

change/@ni-nimble-components-ac937d1b-bb11-4937-90be-1d29c446ca0b.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Let's add a note to the Storybook docs for the number column to mention that it respects the language token. I think Mert did something similar in the date column docs (possibly only visible in the story source right now) which you could copy from.

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
jattasNI: I went back and forth while reviewing the tests but ended up deciding I like this level of coverage. A carefully-chosen single test would ensure that at least one code path in the formatter is respecting the locale but this approach increases our odds of proving that all code paths in the formatter are respecting the locale (important for the default formatter which uses multiple Intl.NumberFormats).

jattasNI (approved):

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -31,7 +34,10 @@ async function setup(source: ColumnInstances): Promise<Fixture<Table<SimpleTable
                     Column 2
                 </${tableColumnNumberTextTag}>
             </nimble-table>`,
-        { source }
+        {
+            source,
+            parent: themeProvider
```
rajsite, line 39: I see this was a pattern introduced in date column for locale.

Could we instead include `nimble-theme-provider` in the template instead of imperatively constructing it and making it the parent?

I'm kinda concerned that it bypasses the fixture workaround we put here:

https://github.com/ni/nimble/blame/ac760e231e1a9b39e898c8753a0b4c95f7e0b41a/packages/nimble-components/src/utilities/tests/fixture.ts#L152-L157

table is viewport sensitive-ish so not sure if this could cause issues by the table being pushed out the viewport or not.

Maybe we should go as far as removing that as an option from the fixture helper..

mollykreis: I've updated these number-text column tests to include the theme-provider in the template.

I've we're happy with the way this looks, I can make a separate PR to apply the same change to the date-text column and figure out the next step towards getting rid of the `parent` option (either create tech debt issue or just do it).

rajsite: LGTM!

rajsite (approved):

Issue #265, opened 2022-01-20
Evaluate testing APIs that components should expose to client applications


fredvisser:
Copied from [User Story 1537054: Evaluate testing APIs that components should expose to client applications](https://ni.visualstudio.com/DevCentral/_workitems/edit/1537054)

(See also the comments in that US)

It should be easy for a client application to write a test which includes a component in the page, configures it, interacts with it, and inspects its visual state.

We've had good success following Angular Material's test harness/page object pattern and we had many challenges with the testability of Smart elements. We should evaluate what testing APIs the nimble components need to expose to be delightful for client applications to consume them.

The primary goal for now is Angular client applications, but we should see if we can write this layer at the nimble-components level so that other frameworks could take advantage of it too.

## Current status

[A branch exists](https://github.com/ni/nimble/tree/add-component-test-harnesses) that includes:
- a page object for the combobox
   - built using testing-library user events
   - includes unit tests that are passing
- re-export the page object for Angular
   - includes unit tests that are **not** passing 

[A not building branch exists](https://dev.azure.com/ni/DevCentral/_git/Skyline?version=GBusers/jmeyer/use-nimble-page-objects) that shows what it looks like to use that page object in an Angular app.

### Possible next steps
1. fix angular tests
2. write page objects for more components (see comments in this issue for some interesting ones)
3. figure out cases where page objects might need to expose read APIs (e.g. breadcrumb)
4. think about code sharing with playwright tests and how page objects would bubble up to apps. 
   - [see playwright testing library](https://github.com/testing-library/playwright-testing-library) for ideas about locators
   - think about whether we could single source page objects for both testing frameworks despite them having different models for input events

rajsite:
Some interesting topics:
- Can we create page objects that can be used across frameworks?
    - Page objects help create queries in the dom for elements and operate on elements. Can that be done in a way that can be leveraged from jasmine, playwright (older but potentially relevant vanillajs, puppeteer, testcafe, protractor, jest, cypress)
    - Interesting library may be https://testing-library.com/docs/ which supports multiple frameworks and ecosystems with a consistent api
    - playwright now has a "testing-library-like" api https://twitter.com/diegohaz/status/1578567655493193728
    - - could be related to [Improve/Share: Page Objects, CSS/Element Selectors, Test Harnesses](https://dev.azure.com/ni/DevCentral/_workitems/edit/2083041)
[...]

PR #1500, opened 2023-09-07
Create page object for nimble-select and add page object CONTRIBUTING docs

jattasNI:
# Pull Request

## 🤨 Rationale

As part of #265 we would like page objects for all of our components.

## 👩‍💻 Implementation

1. Created a nimble-components page object for nimble-select following standard naming and file location.
3. Added methods that seemed useful after scanning Nimble tests and SLE tests that reference the component
4. Added unit tests to ensure all page object methods are exercised
5. Expose from nimble-angular in a new entry point following standard naming and file location. (Note that I'm NOT migrating the select directives to their own new entry point)
6. Added documentation to CONTRIBUTING to describe a philosophy for what belongs in a page object and what doesn't

## 🧪 Testing

1. New unit tests in nimble-components. I didn't see anywhere in nimble-angular that could leverage it.
2. TODO: use in SLE

## ✅ Checklist

### DIFF
--- a/angular-workspace/projects/ni/nimble-angular/select/testing/ng-package.json
+++ b/angular-workspace/projects/ni/nimble-angular/select/testing/ng-package.json
@@ -1,0 +1,6 @@
+{
+    "$schema": "../../../../../../node_modules/ng-packagr/ng-package.schema.json",
+    "lib": {
+      "entryFile": "public-api.ts"
+    }
+}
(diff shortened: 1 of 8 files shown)


jattasNI:
This PR isn't likely to get attention soon so I'm abandoning it to reduce clutter, but it's still linked from the relevant issue if we ever want to revive it.

Issue #1422, opened 2023-08-10
Icon buttons used in toolbar is not getting 'tabindex' value '0' when clicked


vivinkrishna-ni:
# 🐛 Bug Report

When using buttons with icons within the `nimble-toolbar` and clicking on the icon, the button does not receive a `tabindex=0`. By default, the first button in the toolbar possesses a `tabindex=0`, whereas the remaining buttons have a `tabindex=-1`. Consequently, upon pressing the `Tab` key to shift focus from the previously selected field to the buttons, the focus does not transition to the button that was most recently clicked.
 
## 💻 Repro or Code Sample

1. Add nimble buttons with an icon in it to a `nimble-toolbar`. 
2. Inspect the buttons and verify the default `tabindex` value.
3. Click the icon in the button which has `tabindex=-1`.
4. Notice the `tabindex` value in the dev tools.

## 🤔 Expected Behavior

The button which is clicked should be updated to `tabindex=0`. Also, the button which has `tabindex=0` previously should be updated to `tabindex=-1`.

## 😯 Current Behavior

The button which is clicked is not updated to `tabindex=0` when clicking on the icon. However, when clicking the button and not on the icon, the `tabindex=0`. However, upon clicking outside the icon, the `tabindex` gets updated.
![tabindex-issue-toolbar](https://github.com/ni/nimble/assets/123377523/c6b1469b-0bf4-45cc-91f5-74b44bb8f560)

## 💁 Possible Solution

Add `pointer-events: none;` to `start` or `end` slots where the icon is placed.

## 🔦 Context

In `nimble-rich-text-editor` component, for footer formatting options we use `nimble-toolbar` with the `nimble-toggle-button` just the icon enabled with it. For more details, see: https://github.com/ni/nimble/pull/1416#discussion_r1282725591 (issue 2)

~Currently, `nimble-rich-text-editor` has the workaround fix as mentioned in the possible solution, and here is the link to the PR: https://github.com/ni/nimble/pull/1503. Once the actual fix is implemented, the workaround in the editor styles should be removed.~

## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser Google Chrome
* Version 115.0.5790.170

mollykreis:
[...]

PR #1503, opened 2023-09-08
Rich text editor | Fix for updating "tabindex" of button in toolbar when clicking on the icon

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Partial fix for https://github.com/ni/nimble/issues/1422

**Before fix**: `tabindex` value of the icon buttons in the toolbar not getting updated when clicking on the icon.
**After fix**: `tabindex` value of the icon buttons in the toolbar gets updated when clicked on the icon.

With the fix, the last clicked button(anywhere in the button) will get the focus when tabbing to the footer section next time.

## 👩‍💻 Implementation

As mentioned in this comment (https://github.com/ni/nimble/issues/1422#issuecomment-1707000396), adding `pointer-events: none;` to all the icons used inside the `nimble-toggle-button`.

## 🧪 Testing

- Verified the `tabindex` value after clicking on and around the icon in the buttons. 
- Written test cases to check if clicking the slot content actually updates the `tabindex` value of the button.

## ✅ Checklist


vikisekarNI (approved):

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -219,6 +219,10 @@ export const styles = css`
         place-items: center;
     }
 
+    nimble-toggle-button::part(start) {
```
vikisekarNI, line 222: Do we need a test case to check whether tab index gets updated?

vivinkrishna-ni: We already have a test case to check the tab index value with the click of every button (linked below). Given that clicking on the icon is a specific action and the test cases are written for the entire button, they passed earlier.

https://github.com/ni/nimble/blob/abc7972d954e3b5335aaac27084ccf4519c9910f/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts#L203C1-L205C31

vivinkrishna-ni: As discussed, I queried the slot element of the icon and checked the tab index value of the button on click.

change/@ni-nimble-components-f4a15d51-6bb8-4ee2-8980-eb1b3325b788.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for tabindex value change when clicked on icon of the buttons in rich text editor toolbar.",
```
m-akinc, line 3:
```suggestion
  "comment": "Update tabindex when clicking icon of a button in rich text editor toolbar.",
```

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -219,6 +219,10 @@ export const styles = css`
         place-items: center;
     }
```
jattasNI, line 221: I added @mollykreis as a reviewer since she's been the main person in the previous discussions on the linked issue.

From my reading of her comment, I understood the suggestion was to make this change in a more common location like our button styles, not just in the rich text editor styles.

vivinkrishna-ni: This is more of like a workaround fix for the `nimble-rich-text-editor` as the component is already integrated into the SLE, We thought of resolving some of the known issues if we have any workarounds in hand. Solving this workaround in the common location for buttons might not be the proper fix as the button slot contents can be anything, but we are targeted for only icons. Here in `rich-text-editor`, the buttons will only be used with icons placed at `slot="start"` as of now and once it is fixed in the actual component, we will remove this workaround. 

I have added a placeholder comment linking the issue and once it is fixed, we will remove this workaround.

mollykreis: My opinion is that the fix in this PR is too narrowly focused, even for a workaround. I don't have a problem adding `pointer-events: none;`, but as @jattasNI mentioned, I think we should add that style in a more common location, such as the shared styles for all buttons (https://github.com/ni/nimble/blob/main/packages/nimble-components/src/patterns/button/styles.ts).

jattasNI: Thanks for adding a code comment linking to the issue. If we go forward with this approach, could you also update the issue to mention that this workaround should be removed when the issue is fixed?

As for whether we *should* go forward with the workaround.... It's technically true that a client can slot anything into a button. But in our documentation we only describe adding icon content. And in a few spots in the code we apply icon-specific styling (e.g. setting icon colors). So for now it is our policy that the only supported content in that slot is an icon and I would be comfortable with adding more icon-specific code to the buttons. (Other reviewers, feel free to weigh in).

If you have the bandwidth to make the fix in the button code, that would be my preference (maybe in a separate PR so we can compare them). If not, I agree that fixing this case is better than doing nothing so I will approve it.

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/styles.ts
vivinkrishna-ni: I will wait for other reviewer's thoughts on this. If everyone is fine with this approach, I will update the issue to remove this workaround when the actual fix is implemented.

jattasNI: @rajsite and @mollykreis and I agreed at our team meeting today that it makes sense to proceed with this PR as-is since we don't have a super clear plan for what the actual fix will look like.

rajsite (approved):

rajsite:
Since we are in alignment and it's well covered you don't need to wait on @mollykreis's review

packages/nimble-components/src/rich-text/editor/styles.ts
vivinkrishna-ni: I have updated the description of this issue https://github.com/ni/nimble/issues/1422#issue-1844688796 to remove this workaround when the actual fix is implemented.

PR #1504, opened 2023-09-08
Rich text viewer | Fix horizontal scroll bar appearing issue for a single long word

aagash-ni:
# Pull Request

## 🤨 Rationale

Fix for [Bug 2516893](https://dev.azure.com/ni/DevCentral/_workitems/edit/2516893): Scroll bar appearing for a single large word.

Should restrict scroll bar when single word added beyond viewer's boundary, instead it should wrap to next line.

## 👩‍💻 Implementation

Added `overflow-wrap: anywhere` in viewer container CSS property. 

## 🧪 Testing

- Manually verified the wrapping behavior by entering long single word in the viewer.
- [Different Contents In Mobile Width](https://www.chromatic.com/pullrequest?appId=60e89457a987cf003efc0a5b&number=1504&activeElementId=comparison-test-key-64abd1e710a8504746e999f5-1200px&view=changes) UI Test already covering these changes. 

## ✅ Checklist


vikisekarNI (approved):

suseendran-ni (approved):

change/@ni-nimble-components-509eb87e-ffab-4277-bb4c-f9af5cd13a38.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add work-break for viewer container",
```
m-akinc, line 3:
```suggestion
  "comment": "Wrap long words to the next line in rich-text viewer",
```

m-akinc, line 3: Why is this only needed for the viewer? Should the editor be changed, too?

change/@ni-nimble-components-509eb87e-ffab-4277-bb4c-f9af5cd13a38.json
m-akinc, line 1: We should have a test for this. I think it would be sufficient to modify the existing `fitToContentTest` in `rich-text-viewer-matrix.stories.ts` by adding a very long word to the content. We should see that it wraps to the next line when the viewer width is constrained to 300px.

m-akinc (approved):

change/@ni-nimble-components-509eb87e-ffab-4277-bb4c-f9af5cd13a38.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jattasNI, line 2: Please update the Testing section of the PR description to include context about automated testing; there should always be justification if something is only being manually tested.

In this case it looks like you approved a Chromatic diff so you could mention that and note that this shows it was covered by existing auto tests.

jattasNI (approved):

packages/nimble-components/src/rich-text/viewer/styles.ts
```diff
@@ -25,6 +25,7 @@ export const styles = css`
         box-sizing: border-box;
         position: relative;
         color: inherit;
+        word-break: break-word;
```
vikisekarNI, line 28: Quickly skimmed through MDN docs for [`word-break`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) and [`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap). 

Looks like `word-break:break-word` is deprecated. `overflow-wrap: break-word` seems to be a good alternative

image

```suggestion
        overflow-wrap: break-word;
```

aagash-ni: `overflow-wrap: break-word;` seems to have a different behavior where it brings a scroll bar for a single long word instead of breaking it and breaks only if new words are added.
image

But `overflow-wrap: anywhere;` has the same behaviour as ` word-break: break-word;`

packages/nimble-components/src/rich-text/viewer/styles.ts
```diff
@@ -25,6 +25,7 @@ export const styles = css`
         box-sizing: border-box;
         position: relative;
         color: inherit;
+        overflow-wrap: anywhere;
```
rajsite, line 28: Make sure to update the PR implementation description to match the changes

rajsite (approved):

packages/nimble-components/src/rich-text/viewer/styles.ts
aagash-ni: Updated

PR #1505, opened 2023-09-08
Created new grid dimensions api attributes

munteannatan:
# Pull Request

## 🤨 Rationale

This change was specified in the wafer spec. It will allow for the user to specify the grid dimensions without the need to calculate them

## 👩‍💻 Implementation

Created the api, implemented the state change updates, fixed incompatibilities
Created validity api and checks

## 🧪 Testing

existing unit tests updated
storybook controls added
storybook matrix created

## ✅ Checklist


atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/tests/utilities.ts
```diff
@@ -53,12 +53,19 @@ export function getHighlightedValues(): string[] {
...
     return scaleBand<number>().domain(domain).range(range);
 }
 
+export function getHorizontalScale(): ScaleBand<number> {
+    return getScaleBand([2, 3, 4, 5, 6], [2, 7]);
+}
+
+export function getVerticalScale(): ScaleBand<number> {
+    return getScaleBand([1, 2, 3, 4, 5, 6], [1, 7]);
+}
```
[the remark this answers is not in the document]
munteannatan: they are only relevant to the die coordinates in the test data. I changed them to instances

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -261,10 +278,41 @@ const metadata: Meta<WaferMapArgs> = {
...
         dieHover: {
             name: 'die-hover',
             description:
                 'The event is fired whenever the mouse enters or leaves a die. In the event data, `detail.currentDie` will be set to the `WaferMapDie` element of the `dies` array that is being hovered or `undefined` if the mouse is leaving a die.'
+        },
+        validity: {
+            description: `Readonly object of boolean values that represents the validity states that the wafer map's configuration can be in.
+The object's type is \`WaferMapValidity\`, and it contains the following boolean properties:
+
+-   \`invalidGridDimensions \`: \`true\` when any of the \`gridMinX\`, \`gridMinY\`, \`gridMaxX\` or \`gridMaxY\` is \`undefined\`, but false when all of them are \`undefined\``,
```
jattasNI, line 314:
```suggestion
-   \`invalidGridDimensions \`: \`true\` when some of the \`gridMinX\`, \`gridMinY\`, \`gridMaxX\` or \`gridMaxY\` is \`undefined\`, but false when all of them are provided or all of them are  \`undefined\``,
```

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -261,10 +278,41 @@ const metadata: Meta<WaferMapArgs> = {
                 }
             }
         },
+        gridMinX: {
+            name: 'gridMinX',
```
jattasNI, line 282: It looks like this story is inconsistent about referring to its API by the property syntax (`gridMinX`) vs the attribute syntax (`grid-min-x`). Lots of our other stories are inconsistent too but generally we prefer the attribute syntax when an API supports both.

jattasNI (approved):

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+        return {
+            invalidGridDimensions: this.invalidGridDimensions
+        };
+    }
+
+    public isValid(): boolean {
+        return Object.values(this.getValidity()).every(x => x === false);
+    }
+
+    public validateGridDimensions(): boolean {
```
DStavilaNI, line 22: Could you please elaborate a little bit on the validation logic ? Unless I am missing something it does feel incomplete - there are a couple of cases which might need to be addressed, given that this is public API.   What if `gridMinX`  or  `gridMinY` is negative ?  `gridMinX` is negative and `gridMaxX` is positive ? Can the sign change based on the wafer orientation ? 
What if  `gridMaxX - gridMinX` is less than the die dimension ?

packages/nimble-components/src/wafer-map/modules/computations.ts
```diff
@@ -101,7 +101,11 @@ export class Computations {
             this._containerDimensions.width,
             this._containerDimensions.height
         );
-        const gridDimensions = this.calculateGridDimensions(this.wafermap.dies);
+        // the dimensions are valid if they are all undefined
+        const gridDimensions = !this.wafermap.validity.invalidGridDimensions
+            && this.wafermap.gridMinX !== undefined
+            ? this.calculateGridDimensionsFromBoundingBox()
+            : this.calculateGridDimensionsFromDies(this.wafermap.dies);
```
rajsite, line 108: Instead of a comment and somewhat complex conditional you could make a well named function
 something like:

```ts
        const gridDimensions = this.gridDimensionsValidAndDefined()
            ? this.calculateGridDimensionsFromBoundingBox()
            : this.calculateGridDimensionsFromDies(this.wafermap.dies);
```

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.invalidGridDimensions = false;
+        if (
+            this.wafermap.gridMinX === undefined
```
rajsite, line 25: A slightly stronger check would be checking that they are all `typeof` 'number' here and below.

There are known issues where removing an attribute results in `null` instead of `undefined` (i.e. `waferMap.removeAttribute('grid-min-x')`) even though the types don't show that. A stricter check using typeof number would prevent that issue.

packages/nimble-components/src/wafer-map/tests/computations.spec.ts
```diff
@@ -16,12 +16,17 @@ describe('Wafermap Computations module', () => {
         beforeEach(() => {
             const waferMock: Pick<
             WaferMap,
-            'dies' | 'originLocation' | 'canvasWidth' | 'canvasHeight'
+            | 'dies'
+            | 'originLocation'
+            | 'canvasWidth'
+            | 'canvasHeight'
+            | 'validity'
```
rajsite, line 23: Could you make a helper for creating the wafer mock for this file?

Edit: Actually looks like we should try and clean-up some of the other modules as well that create WaferMap Mocks. I don't think we should necessarily create a giant complex mock shared across all the tests but maybe one per test file. Or maybe a shared mock would still be cleaner.

Either way that shouldn't grow the scope of this PR. So can you create a tech debt task for the clean-up of WaferMap mocks and address it in a follow-up PR?

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
rajsite: Could you also create a test file for the validator that covers all these cases?

munteannatan: We support negative inputs, just as we support negative die coordinates.
These grid inputs refer to the virtual grid of dies that are provided as input in an array of points and values. they can have negative coordinates or range from negative to positive. We are mapping this grid to the canvas with positive die width and height in the prerendering module. 
The orientation does not influence this grid at all, the origin location will influence the direction in which they get drawn on the canvas (left -> right, top->down for example), but the virtual die grid is not influenced.
As they are represented as points, they do not have a width and a height, so in the case where the min and max are equal nothing is displayed.

packages/nimble-components/src/wafer-map/tests/computations.spec.ts
munteannatan: sure, I made it here [Technical Debt 2520619](https://dev.azure.com/ni/DevCentral/_workitems/edit/2520619): Clean up existing nimble tests

DStavilaNI (approved):

rajsite (approved):

Issue #1324, opened 2023-06-26
Nimble combobox list option is not opening the first time you type


codrutaserban:
# 🐛 Bug Report

In a slide-out the combobox is the first input and when it opens the focus is on the combobox and you can start writing. The problem is the option list doesn't open after you type. It works if the combobox lose focus and you click on it again.

## 💻 Repro or Code Sample

These are the inputs:
`
<nimble-combobox *ngIf="showAutocomplete"
                #id
                formControlName="..."
                autocomplete="none"
                appearance="underline"
                placeholder="Path *" 
                [error-text]="..."
                [error-visible]="abstractControl.touched && abstractControl.invalid">
`

## 🤔 Expected Behavior

The option list should open when you start typing.

## 😯 Current Behavior
When you open the slide-out, even that you can type right away the option list doesn't open.
The problem seems to be that the `open` attribute is not set when you start typing and it is the one that makes the nimble-anchored-region become visible.

![current behaviour](https://github.com/ni/nimble/assets/89146546/1d515e69-1a93-4808-906f-9f54d0a913c3)

## 🔦 Context
This happens when you open the slide-out and the component is already on focus (is the first input) and doesn't detect I am typing.

ipopa144:
This is reproduceable only when the combobox is used with `autocomplete="none"` option. When used with `autocomplete="both"`, it works as expected. See the `sl-workspace-selector` component for reference.

![workspace-selector](https://github.com/ni/nimble/assets/44372517/fef76baf-4680-4f76-922e-32effea99908)

atmgrifter00:
@codrutaserban @ipopa144, I want to point out that the current behavior of the Combobox when its `autocomplete` setting is either `none` or `inline` I would call "expected". As you can see in [FAST's Combobox implementation](https://github.com/microsoft/fast/blob/adaaeb653a37c0eec6ce5fbc57b73c24ca09f347/packages/web-components/fast-foundation/src/combobox/combobox.ts#L388), they are very explicit about which mode should open up the dropdown if it is currently closed while typing. 

While this isn't on its own proving that this is proper behavior, I believe we can find reason behind it from the purpose of the `autocomplete` settings. For the `both` and `list` options, they are actively performing filtering on a list of options, which would suggest a user would need to see the options being filtered as they type. For `inline` and `none` no filtering is taking place, and thus the "hints" that the dropdown is providing aren't necessary as they aren't changing as the user types.

[...]

PR #1507, opened 2023-09-08
Add open attribute to Combobox Angular directive.

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1324

## 👩‍💻 Implementation

As discussed in the comments of the linked issue, the problem behavior is actually expected behavior for the particular autocomplete mode that is being used. Unfortunately, the client use-case is providing a custom filtering behavior based on the user typing into the input field of the combobox (thus, subverting the native behavior).

To help mitigate the issue, and prevent the clients from having to access the `open` attribute on the native element of the combobox, we can at least expose the `open` attribute on the Angular directive.

## 🧪 Testing

Unit tests.

## ✅ Checklist


mollykreis (approved):
Looks like you're missing a change file but looks good other than that.

rajsite (approved):

PR #1512, opened 2023-09-11
Expose lang and validity from NimbleThemeProviderDirective

m-akinc:
# Pull Request

## 🤨 Rationale

We added a `lang` and `validity` properties as well as a `checkValidity()` function to the theme provider's API. We want to expose those from the Angular directive as well.

## 👩‍💻 Implementation

- added `lang`, `validity` and `checkValidity()` to `NimbleThemeProviderDirective`
- added unit tests for the same, plus pre-existing `theme` attribute

## 🧪 Testing

Unit tests pass

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/src/directives/theme-provider/tests/nimble-theme-provider.directive.spec.ts
```diff
@@ -11,4 +13,199 @@ describe('Nimble theme provider', () => {
...
+            expect(nativeElement.theme).toBe(Theme.dark);
+
+            fixture.componentInstance.theme = Theme.color;
+            fixture.detectChanges();
+
+            expect(directive.theme).toBe(Theme.color);
+            expect(nativeElement.theme).toBe(Theme.color);
+        });
+
+        it('can be configured with property binding for lang', () => {
```
mollykreis, line 142: Right now, this test is testing two different things:
1. can use property binding for lang
2. updating lang to an invalid value results in the element to be invalid

I think that would be better separate it into two different tests. To follow the patterns we already have in place, the "can be configured with property binding" should only test the binding aspect of the directive. My thought is that test should set a valid new value for 'lang'. Then, you could have a different test that setting an invalid value for 'lang' appropriately makes the element invalid.

angular-workspace/projects/ni/nimble-angular/src/directives/theme-provider/tests/nimble-theme-provider.directive.spec.ts
```diff
@@ -11,4 +13,199 @@ describe('Nimble theme provider', () => {
...
+            expect(nativeElement.theme).toBe(Theme.dark);
+
+            fixture.componentInstance.theme = Theme.color;
+            fixture.detectChanges();
+
+            expect(directive.theme).toBe(Theme.color);
+            expect(nativeElement.theme).toBe(Theme.color);
+        });
+
+        it('can be configured with attribute binding for lang', () => {
```
mollykreis, line 198: Same here. I think this test should just test the attribute binding with 'lang' being updated to a valid value.

mollykreis (approved):

rajsite (approved):

atmgrifter00 (approved):

Issue #1151, opened 2023-04-03
Address Table accessibility issues


atmgrifter00:
## 📌 User Story

The table needs the correct ARIA role.

Once the [table row grouping PR](https://github.com/ni/nimble/pull/1139) goes through, there will be some accessibility issues that will require some follow up work on (some of which will possibly be part of the [Table Interaction Keyboard](https://github.com/ni/nimble/issues/1137) work). Some of the issues raised include:
- Determine what the appropriate role for the `TableGroupRow` is ([ARIA guidance](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/rowgroup_role) strongly suggests that the `rowgroup` role is only appropriate for DOM elements that actually contain `row` (or other) elements, which our `TableGroupRow` does not).
- Determine what the appropriate focus interactions are for the expand/collapse button. We've had some direction that the expand/collapse button shouldn't be focusable, but this seems in conflict with some [expected ARIA behaviors](https://github.com/ni/nimble/pull/1139#discussion_r1152345869).
- ~~Assuming the expand/collapse button remains a button, we will need to create an API for users to supply it an accessible label, much like we did for the [action menu label](https://github.com/ni/nimble/blob/84efdc79ad9dc586b1a32cc3a5f4b39fbee18257/packages/nimble-components/src/table-column/base/index.ts#L32) (except likely exposed on the `Table`).~~ [Addressed in #1519]
- Configure `aria-expanded` on the group rows, but this requires the table itself to have a role of `treegrid`. Some questions about this:
    -  Should the table always be `treegrid` or should the role change based on whether or not a column is grouped and whether or not hierarchy is enabled?
    - Does `aria-expanded` make sense on group rows if the rows beneath it are not child elements in the DOM?
- ~~Add accessible label to the selection checkboxes in the multi-select case. This should probably be done in a consistent manner as adding a label to the expand/collapse button for group rows.~~ [Addressed in #1519]

Note: The localization concerns described above are also captured in #1090 

Note: When this is fixed, we can also address #1195.

atmgrifter00:
@leslieab, we will likely need to revisit some of the decisions surrounding focus behaviors for the expand/collapse button (see the description above).

atmgrifter00:
@leslieab, it would only raise warning for interactive elements like buttons. At the moment, the header has no such elements.

leslieab:
[...]

Issue #1510, opened 2023-09-11
setting aria-label on nimble icon causes accessibility issue


mollykreis:
# 🐛 Bug Report

Setting an `aria-label` on a nimble icon causes an accessibility issue: [Elements must only use allowed ARIA attributes](https://dequeuniversity.com/rules/axe/4.3/aria-allowed-attr?application=axeAPI). The `role` of the nimble icon is `generic`, and `generic` elements [are not allowed to have an ARIA label or description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/generic_role).

## 💻 Repro or Code Sample

This causes an accessibility issue any time `aria-label` is applied to an icon, which currently seems to be the recommended approach to making the icon accessible.

For example:
- Look at the [Icon column storybook story](https://nimble.ni.dev/storybook/?path=/story/incubating-table-column-icon--icon-column)
- Go to the "Accessibility" tab
- Go to the "Incomplete" tab
- Notice that all the icons in the table have accessibility issues associated with them because of the `aria-label` set on the icons

## 🤔 Expected Behavior

We should have a recommended way to set an accessible label on a nimble icon that doesn't cause its own accessibility issues

## 😯 Current Behavior

Our recommended way to set an accessible label on a nimble icon isn't correct because it causes accessibility issues

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

msmithNI:
Currently our docs state that icons should consider setting `title` but don't mention `aria-label` specifically. The icon column cell view does set aria-label though ([source link](https://github.com/ni/nimble/blob/9a5a98e8112d2ee1b94dcec3e01ff84bcebe08e1/packages/nimble-components/src/table-column/icon/cell-view/template.ts#L15)).

Based on the outcome of this issue we may need to update our docs about icon accessibility guidance:
- [Localization spec](https://github.com/ni/nimble/tree/main/specs/labels-and-localization)
- [icons.mdx](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/icon-base/tests/icons.mdx)

m-akinc:
I can think of three options:

[...]

Issue #1222, opened 2023-05-02
Table interactive sorting (via column header menus)


msmithNI:
See #854 which captures the other part of interactive sorting (via clicking the headers).

See #885 for the UX/IxD design for sorting via the header menus.

## 📌 User Story
- [ ] Interactive sorting via header menu (not started, header menu is also needed for #867 )

fredvisser:
Adding column header menus is not a near to medium priority for the nimble-table. In Stratus apps, we have other shared components to configure columns, so don't require this capability.

PR #1519, opened 2023-09-12
Add localized labels to remaining table sub-elements

mollykreis:
# Pull Request

## 🤨 Rationale

Add localized labels and titles to remaining elements within the nimble-table.

This is part of #1151 

## 👩‍💻 Implementation

Add new label tokens for required strings and use them within the table.

## 🧪 Testing

- Manually tested that the labels show up correctly in the accessibility tree*
- Manually tested that titles are appropriately shown on elements in storybook
- Created a branch that re-introduced initial rows in the Angular example app table and verified that lighthouse accessibility check now passes.

*As part of this manual check, I found that the labels on our group header icons don't work as expected due to #1510

## ✅ Checklist


packages/nimble-components/src/label-provider/table/label-token-defaults.ts
```diff
@@ -7,5 +7,10 @@ export const tableLabelDefaults: { readonly [key in TokenName]: string } = {
     tableGroupExpandLabel: 'Expand group',
     tableGroupsCollapseAllLabel: 'Collapse all groups',
     tableCellActionMenuLabel: 'Options',
-    tableColumnHeaderGroupedIndicatorLabel: 'Grouped'
+    tableColumnHeaderGroupedIndicatorLabel: 'Grouped',
+    tableColumnHeaderSortedAscendingIndicatorLabel: 'Sorted ascending',
+    tableColumnHeaderSortedDescendingIndicatorLabel: 'Sorted descending',
+    tableSelectAllLabel: 'Select all rows',
+    tableGroupSelectAllLabel: 'Select all rows in group',
+    tableRowSelectLabel: 'Select row'
```
mollykreis, line 15: I wasn't quite sure on all these "Select ..." labels. I thought about making the text something like "Toggle selection of ..." but I didn't think that would read well with a screen reader. My guess is that it would say something like "Toggle selection of all rows, checked", and it seemed confusing to me to saying that toggling something was checked.

I'm open to other suggestions here.

packages/nimble-components/src/table/components/header/template.ts
```diff
@@ -14,15 +18,27 @@ export const template = html<TableHeader>`
         @mousedown="${(_x, c) => !((c.event as MouseEvent).detail > 1)}"
     >
         <slot></slot>
-        ${'' /* Omit title attribute for sort indicators because aria-sort is set on the 1st sorted column */}
-        ${when(x => x.sortDirection === TableColumnSortDirection.ascending, html`
-            <${iconArrowUpTag} class="sort-indicator" aria-hidden="true"></${iconArrowUpTag}>
+        ${'' /* Set aria-hidden="true" on sort indicators because aria-sort is set on the 1st sorted column */}
```
mollykreis, line 21: This comment was a bit confusing to me. I'm not sure why we would omit the title, which could be helpful for sighted people, because we don't want the icon to be present in the accessibility tree. Since the sort icons already have `aria-hidden="true"` set on them, I didn't see any harm in applying a title. Let me know if I'm overlooking something.

mollykreis:
@msmithNI, could you buddy this PR for me?

packages/nimble-components/src/table/components/header/template.ts
msmithNI: [Here's the related discussion](https://github.com/ni/nimble/pull/1328#discussion_r1248069793) from my original label provider PR.

So the original thought was that setting title on the sort icons was redundant due to aria-sort. However I don't think I considered the fact that they were already set as `aria-hidden` so that didn't really apply.
I think your logic (could be useful for sighted people) makes sense so I'm fine adding title.

packages/nimble-components/src/label-provider/table/label-token-defaults.ts
msmithNI: I think these labels make sense.

Based on these APG doc pages for [buttons](https://www.w3.org/WAI/ARIA/apg/patterns/button/)/ [switches](https://www.w3.org/WAI/ARIA/apg/patterns/switch/)/ [checkboxes](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/), I think the label should describe the action/command, and the label should make sense regardless of checked state (since the text shouldn't change based on checked/unchecked). As you said adding 'Toggle' when that would be combined with 'unchecked'/'checked' would be redundant/confusing.

packages/nimble-components/src/table/template.ts
```diff
@@ -48,11 +51,13 @@ export const template = html<Table>`
                     <div class="header-row" role="row">
                         <span class="header-row-action-container" ${ref('headerRowActionContainer')}>
                             ${when(x => x.selectionMode === TableRowSelectionMode.multiple, html<Table>`
-                                <span role="columnheader" class="checkbox-container">
+                                <span role="gridcell" class="checkbox-container">
```
msmithNI, line 54: I'm curious if you have specific rationale for this change (other than that it's inconsistent with the role of the button below it).

I had a hard time trying to find guidance for the ARIA role of select-all buttons in a column header. The [AG Grid example](https://www.ag-grid.com/example/) still has role=columnheader on it's Select All Rows checkbox, but in their DOM structure, it's combined with the header text for the first column, so it's probably not a good example.

msmithNI (approved):
(Reminder that once this is checked in, the Angular + Blazor label providers need corresponding changes)

packages/nimble-components/src/table/template.ts
mollykreis: I made this change because it fixed an accessibility problem and made the checkbox consistent with the "Collapse all" button below. That probably wasn't the best rationale.

I chatted with @atmgrifter00 about this since he's been looking at the impact of various roles on our elements within the table. We decided that this approach definitely seems wrong. From what we can tell, a correct approach is to put the checkbox + collapse button in a single `columnheader` with a non-visible label of something like "Row actions". That allows the row's selection checkbox (and the collapse button that will eventually come with hierarchy on some rows) to be in that column. As-is, the row selection checkbox is in a `gridcell` that causes the number of cells in a row to be unequal to the number of columns the grid claims to have.

packages/nimble-components/src/table/styles.ts
```diff
@@ -186,6 +186,26 @@ export const styles = css`
     .row {
         position: relative;
     }
+
+    .accessibly-hidden {
+        ${
```
mollykreis, line 191: This styling has been copied & pasted a number of times throughout our codebase, so I think it warrants a shared style sourced from a single location. I plan to do that in a follow up PR.

packages/nimble-components/src/table/template.ts
mollykreis: I've updated the structure/roles with these elements.

mollykreis:
@atmgrifter00, do you mind doing a second buddy review on these changes? Specifically, can you provide a review on the changes to the way that we are optionally adding an additional "column" for the row selection checkbox and group all button?

atmgrifter00 (approved):

packages/nimble-components/src/table/styles.ts
mollykreis: Yep, that was my thought as well 👍

packages/nimble-components/src/table/index.ts
```diff
@@ -136,6 +136,13 @@ export class Table<
         return this.tableValidator.getValidity();
     }
 
+    public get showRowOperationColumn(): boolean {
```
rajsite, line 139: Should this be internal?

packages/nimble-components/src/table/template.ts
```diff
@@ -46,29 +50,34 @@ export const template = html<Table>`
             <div class="glass-overlay">
                 <div role="rowgroup" class="header-row-container">
                     <div class="header-row" role="row">
-                        <span class="header-row-action-container" ${ref('headerRowActionContainer')}>
+                        <span role="${x => (x.showRowOperationColumn ? 'columnheader' : '')}" class="header-row-action-container" ${ref('headerRowActionContainer')}>
+                            ${when(x => x.showRowOperationColumn, html<Table>`
```
rajsite, line 54: nit: trailing whitespace on line

rajsite (approved):

change/@ni-nimble-components-4d50372b-5cd2-4e13-9053-d023da4a2ebe.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: If this doesn't resolve #1151 and you don't plan to do so via follow-up PRs, could you make sure the description there is updated to capture what's done and what you know is left to do?

packages/nimble-components/src/label-provider/table/index.ts
```diff
@@ -44,6 +58,24 @@ export class LabelProviderTable
     @attr({ attribute: 'column-header-grouped-indicator' })
     public columnHeaderGroupedIndicator: string | undefined;
 
+    @attr({ attribute: 'column-header-sorted-ascending-indicator' })
+    public columnHeaderSortedAscendingIndicator: string | undefined;
+
+    @attr({ attribute: 'column-header-sorted-descending-indicator' })
+    public columnHeaderSortedDescendingIndicator: string | undefined;
```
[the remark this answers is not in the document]
jattasNI: The only new related strings I can predict are the menu items we may someday add for #1222. Could be good to try to predict their potential names and use that as input to the current names.

jattasNI (approved):

Issue #94, opened 2021-09-02
Resolve label accessibility errors


fredvisser:
The Select component is called out for not having ARIA values for several attributes, including labels –

![Screen Shot 2021-09-02 at 10 18 20 AM](https://user-images.githubusercontent.com/1458528/131891182-8c28a117-dd19-4aa7-b4a6-c51cc1460696.png)

However, FAST is still developing a strategy for labels within components (https://github.com/microsoft/fast/issues/4801), so we should resolve the accessibility & label issues when the above issue is resolved.

rajsite:
More discussion around having labels associated with controls as being a good default: https://twitter.com/claviska/status/1677469942394900480?s=20

jattasNI:
A couple clients mentioned wanting this recently: [LiMo in a Teams thread](https://teams.microsoft.com/l/message/19:wo8vmMKMsHfltKXxc0bczZo-X4JlQSV5VxpaRJdh13k1@thread.tacv2/1695654013481?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=9ee065d7-3898-4245-82f6-76e86084b8b1&parentMessageId=1695654013481&teamName=Design%20System&channelName=General&createdTime=1695654013481) and TestHub in a meeting.

jattasNI:
Fixing this should help prevent bugs like this, where the "Workspace" select is disabled but its label isn't. We should validate that this behaves correctly when we fix this issue. This example comes from the [SLE Edit Asset slidein](https://test.lifecyclesolutions.ni.com/assets/3f43450b-259d-4ade-95b7-b3cf64bd4703/history)
![image](https://github.com/ni/nimble/assets/10500124/195e0e7a-38cf-487d-86da-f3bca87e5f64)

msmithNI:
Select component has built-in label support as of #2315 - [Select story](https://nimble.ni.dev/storybook/?path=/story/components-select--select) no longer shows accessibility violations.

PR #1520, opened 2023-09-12
Select IxD Specification

leslieab:
# Pull Request

## 🤨 Rationale
For #1303 - Select and Combobox Refresh HLD

- Interaction design specification for select component

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

- [X] Submit project documentation
- [x] Review


packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+
+> NOTE: NEEDS VISUAL DESIGN FOR MULTIPLE FOCUS STATES?
+
+If the select component allows filtering, the filter bar has keyboard focus when the menu is opened.
+
+![ ](spec-images/select-24.png)
+
+> NOTE: NEEDS TECH WRITER FOR TEXT
+
+Typing into the filter bar filters the menu items in the menu (not pictured). The menu should indicate when there are no matches found. The "No matches" indicator is not interactive.
```
atmgrifter00, line 184: Does it bear mentioning that the filter text can match anywhere in the text for each option?

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+Keep the label text as concise as possible, limited to a single line of text. Use sentence casing.
+
+In a form, use the label to indicate whether the select is optional or required for the end-user to interact with. If most of the components in the form are required, indicate which components are optional. If most of the components in the form are optional, indicate which components are required. Required or optional indication in the label text should always follow the pattern "[Label] (optional)" or "[Label] (required)".
+
+![ ](spec-images/select-2.png)
+
+#### Helper Text
+> NOTE: NEEDS VISUAL DESIGN FOR TYPOGRAPHY
+
+Only include helper text to provide information that aids the end-user in choosing the right selection. Keep the helper text concise as possible, limited to a single line of text. Use sentence casing.
```
atmgrifter00, line 48: Just for clarity, is the helper text meant to be user-configurable, or reserved for the `Select` component to use when the situation calls for it? For example, one idea that was floated was we could use this area to provide the group name for the current selection if groups were being used.

If, instead, the idea was that the helper text was user-configurable, I would think this API would be available on _all_ Nimble input components. This would concern me.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+To allow end-users to multi-select menu items, use a [**tag picker**](/packages/nimble-components/src/tag-picker/specs/IxD.md).
+
+![ ](spec-images/select-alternative-picker.png)
+
+## Design
+### Configuration
+#### Label
+> NOTE: NEEDS VISUAL DESIGN FOR PLACEMENT
+
+The client-user should be able to specify whether they want the label to be positioned at the top or to the left of the select or not at all. Selects that don't have a label should still have an ARIA label.
```
atmgrifter00, line 78: I don't think this influences the content of this document, but I just wanted to mention that labels for select/combobox is still an ongoing issue in FAST, which we are tracking in part as part of [this issue](https://github.com/ni/nimble/issues/94).

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+#### Icons
+The client-user should be able to specify whether they want left icons on menu items or not. Icons show up in menu items and are displayed as part of the value in the field button. Images and other content not specified here should not be permitted.
+
+![ ](spec-images/select-7.png)
+
+#### Groups
+> QUESTION: IS THE SL "VIEW" PICKER A SELECT (SHOULD IT BE)? WHAT ABOUT GROUPS WITHOUT LABELS, WHAT ABOUT A SELECTION GROUP VS. AN ACTION GROUP OF MENU ITEMS (A.K.A MEGA-MENUS)?
+
+The client-user should be able to create groups for menu items in the menu. Groups are not interactable.
+
```
atmgrifter00, line 91: An important interaction decision to make about groups is whether or not they are part of the "filtering" interaction. One possible nicety is that a user can filter by the group, and then see all of the options under that group alone.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+#### Overflow Text
+> NOTE: NEEDS DEV & VISUAL DESIGN REVIEW
+
+All text in the select component should truncate, not wrap.
+
+![ ](spec-images/select-14.png)
+
+#### Progress & Loading
+When displaying menu items in the menu lags or is expected to take longer than normal, the menu should indicate that it's loading. The "Loading…" display is not interactive.
```
atmgrifter00, line 136: I'm not convinced that this is a concern that the `Select` component can adequately manage on its own (particularly when we consider the below feature of a "Load more" button). My preference would be that, if anything, the `Select/Combobox` offer a slot that allows users to populate the bottom of the dropdown with user-defined content.

My current conception of what the APIs might look like if we were to support this in the `Select/Combobox` directly would allow it to be configured to be in an awkward state, relying on the user to tightly manage so that it displays things properly, which I would rather avoid. While the slot API that allows general content also enables it to be configured awkwardly, the API itself isn't explicit about it, and is instead "explicit" in its expectation that it's up to the user to configure it appropriately.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+![ ](spec-images/select-15.png)
+
+> NOTE: IS THIS NEEDED? IF YES, NEEDS VISUAL DESIGN & TECH WRITER
+
+Menus with dynamically loading menu items may have a "Load more" button. Clicking the "Load more" button should resolve into the "Loading…" indication, then load more menu items and scroll the menu as needed.
+
+![ ](spec-images/select-16.png)
+
+#### Touch-Screen Devices
+On touch-screen devices, the select component should bring up the menu in a Nimble or native tray instead of a popover.
```
atmgrifter00, line 147: While I think it's good to give the proper guidance for mobile devices, this seems like it would be difficult to support. Note that one of the goals for our components was to behave like the native HTML components. Native selects do _not_ have this behavior in browsers on mobile. They open up a popup.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+
+![ ](spec-images/select-3.png)
+
+#### Field Button
+Selects can be empty by default or have a default selection from the menu. Empty selects should include placeholder text that propts the end-user to select an option from the menu. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Ellipses are not needed. Use sentence casing.
+
+![ ](spec-images/select-4.png)
+
+#### Menu & Menu Items
+A list of options to select, displayed as an open state. Order menu items alphabetically by default, unless another ordering will provide value to the end-user. Keep the menu item text concise as possible, limited to a single line of text. Use sentence casing.
```
atmgrifter00, line 58: I'm not sure the `Select` should intervene in the ordering of the items in the dropdown. Order should be provided via the DOM order of the options. If the user thinks alphabetical order is appropriate, they should order their options so.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+### Configuration
+#### Label
+> NOTE: NEEDS VISUAL DESIGN FOR PLACEMENT
+
+The client-user should be able to specify whether they want the label to be positioned at the top or to the left of the select or not at all. Selects that don't have a label should still have an ARIA label.
+
+![ ](spec-images/select-6.png)
+
+#### Icons
+The client-user should be able to specify whether they want left icons on menu items or not. Icons show up in menu items and are displayed as part of the value in the field button. Images and other content not specified here should not be permitted.
```
atmgrifter00, line 83: I'm concerned about the "Images and other content not specified here should not be permitted". I suspect we _need_ to support arbitrary visuals. I'm not even sure what the constraint would look like. Would we only allow elements that were prefixed with our curated list of supported elements (i.e. `nimble-icon-` , `nimble-avatar`, `nimble-flag`, and so on)? This is an example where I think the `Select` component, itself, should offer a general-purpose API, and any visual constraints placed upon it should happen at the app-level design spec.

packages/nimble-components/src/select/specs/IxD.md
leslieab: Yes, I think it is a requirement that most input components have client-user configurable helper text. This is to provide additional contextual information in a way that can still be accessed on touch devices, where the end-user cannot hover for help text or tool tips.

I do not think we should use this area to indicate the group name.

What are your concerns?

packages/nimble-components/src/select/specs/IxD.md
leslieab: You're right, I've updated this.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,254 @@
...
+
+![ ](spec-images/select-4.png)
+
+#### Menu & Menu Items
+A list of options to select, displayed as an open state. Order menu items alphabetically by default, unless another ordering will provide value to the end-user. Keep the menu item text concise as possible, limited to a single line of text. Use sentence casing.
+
+![ ](spec-images/select-5.png)
+
+### Related Components
+To allow end-users to enter a custom or create a new menu item, use a [**combobox**](/packages/nimble-components/src/combobox/specs/IxD.md).
```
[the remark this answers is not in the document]
leslieab: Sorry, I don't understand this comment. The functional difference between the select and the combo box is that the end-user can type to enter and commit a custom value in a combo box. I can remove the language "create a new menu item", if we don't plan on implementing that as a feature of the combo box (i.e. the user can enter a custom value, but that custom value will never be added as a menu item). But you don't type to enter a custom value on a select, because that's what a combo box is for.

packages/nimble-components/src/select/specs/IxD.md
leslieab: Okay, that's fair. I removed that line and added some icon information under Usage. My brain tends to default to being restrictive in the name of good UX. The right thing to do is make the components flexible and if it makes sense to add restrictions in the future than we can do that then.

packages/nimble-components/src/select/specs/IxD.md
atmgrifter00: The primary concern I have with this idea of helper text is essentially adding unbounded height to the component. As you mention the helper text seems analogous to a tooltip, which oftentimes can span multiple lines. If the motivation for adding this is enabling the components on mobile devices to better supply contextual information, I think there are other options to consider. One is providing an interactive icon that could be slotted in the component, which a user could touch to bring up a tooltip.

The secondary concern is the helper text's overlapping of the error text (which maybe we should mentioning in this document?). While I think it may be _okay_ to have the error text trump the helper text, there are other patterns, like the one I mention above that allow for available helper text, even in the presence of an error.

Another consideration is that we have to reconcile the recommendation of how group names are rendered with the Combobox (which I noticed its IxD document does not outline, but it will need the same behaviors as the `Select`.). While @NIbokeefe did provide a design option to display the group name within the input field, I don't think this is the ideal choice for something like a Combobox, where I think we should reserve the majority of that space for the actual interactive input element. I'll note that Brandon did also provide a design for displaying the group name in the upper right of the component (next to the label), which might satisfy my concern, if we didn't want to display it in the error text location.

packages/nimble-components/src/select/specs/IxD.md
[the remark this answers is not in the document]
atmgrifter00: I'm seeing this feature less as a "what a user can use as a value", and more of a "allow user to add new items to dropdown". I think it's odd to provide guidance to a user that if they were originally considering using a `Select`, which could have been motivated by them needing to use the `value/ngValue` APIs on the `list-option`, to use a `Combobox` if they also wanted to interactively create a new menu item (as the `Combobox` is still incompatible with their needs). There's a logical leap there that is grounded in the fact that the added menu item _must not_ have a value associated with it other than what was typed in.

Ultimately, this all could be moot if we agree that that feature should not be implemented on the `Combobox`, which I will make my case for on [its PR](https://github.com/ni/nimble/pull/1521). It's additionally moot in that this document is _not_ client-facing. I think the actual guidance we would be giving for why a user would choose a `Select` versus a `Combobox` would be fairly different.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,293 @@
...
+
+| Key                        | Description                                                                                                 |
+| -------------------------- | ----------------------------------------------------------------------------------------------------------- |
+| `ENTER` or `SPACE`         | Sets the value of the field button to the focused menu item, closes the menu, and moves focus to the select |
+| `DOWN ARROW` or `UP ARROW` | Moves visual focus to the next or previous menu item                                                        |
+| `ESC`                      | Closes the menu and moves visual focus to the select                                                        |
+| `HOME` or `END`            | Moves visual focus to the first or last menu item                                                           |
+| `PAGEUP` or `PAGEDOWN`     | Jumps visual focus up or down ten menu items (or to first or last menu item)                                |
+| _Printable characters_     | Moves visual focus to the first menu item that matches the typed character or string                        |
```
atmgrifter00, line 267: Is this how you want it to behave for a filterable select, versus opening the popup, and using the printable characters as the filter text?

packages/nimble-components/src/select/specs/IxD.md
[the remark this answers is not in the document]
leslieab: Sure, this is the user experience centric guidance meant as a first pass for content that will eventually (after being edited and combined with other guidance) live in Storybook. Any technical/functional guidance for choosing a component should be included in client-facing documentation, but I don't think the IxD spec is the correct place to include it.

packages/nimble-components/src/select/specs/IxD.md
leslieab: I'm not sure what you mean. The filter is in the popup.

atmgrifter00: I read the line of "Moves visual focus to the first menu item that matches the typed character or string" as: 

"If a user has the `Select` focused and then types a printable character that it will open the popup and focus the first item in the dropdown that either starts with or contains (this part is unclear) the typed in text". 

This is different than if the typed in text just immediately being forwarded to the filter input in the dropdown, if the `Select` _is_ filterable.

leslieab: Ah, I get it. No, I think it would be best to forward the text to the filter input in the dropdown. I've added that detail.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+Keep the label text as concise as possible, limited to a single line of text. Use sentence casing.
+
+In a form, use the label to indicate whether the select is optional or required for the end-user to interact with. If most of the components in the form are required, indicate which components are optional. If most of the components in the form are optional, indicate which components are required. Required or optional indication in the label text should always follow the pattern "[Label] (optional)" or "[Label] (required)".
+
+![ ](spec-images/select-2.png)
+
+#### Field Button
+
+Selects can be empty by default or have a default selection from the menu. Empty selects should include placeholder text that propts the end-user to select an option from the menu. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Ellipses are not needed. Use sentence casing.
```
mollykreis, line 53: When a select is empty by default, do we need a way for the user to be able to re-enter that default state after they've made a selection? Does that answer change based on whether filling in a value is optional or required?

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### Menu Height
+
+The menu height should be enough to display five to seven menu items before overflowing with a vertical scrollbar (not pictured).
+
+#### Overflow Text
+
+> NOTE: NEEDS DEV & VISUAL DESIGN REVIEW
+
+All text in the select component should truncate, not wrap.
```
mollykreis, line 151: I'm a bit confused by this screenshot along with the guidance above that "Ellipses are not needed". Could you provide some clarification?

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+![ ](spec-images/select-14.png)
+
+#### Progress & Loading
+
+> NOTE: HOW THIS (AND ANY) FEATURE IS INCLUDED IN IMPLEMENTATION IS UP TO DEVELOPERS
+
+When displaying menu items in the menu lags or is expected to take longer than normal, the menu should indicate that it's loading. The "Loading…" display is not interactive.
```
mollykreis, line 159: Wrong word?
```suggestion
When displaying menu items and the menu lags or is expected to take longer than normal, the menu should indicate that it's loading. The "Loading…" display is not interactive.
```

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+> NOTE: NEEDS VISUAL DESIGN FOR MULTIPLE FOCUS STATES?
+
+If the select component allows filtering, the filter bar has keyboard focus when the menu is opened.
+
+![ ](spec-images/select-24.png)
+
+> NOTE: NEEDS TECH WRITER FOR TEXT
+
+Typing into the filter bar filters the menu items in the menu (not pictured). The filter string can match anywhere in the text for each option. The menu should indicate when there are no matches found. The "No matches" indicator is not interactive.
```
mollykreis, line 209: I have a guess, but could you explicitly mention what happens to the entered filter string when the menu closes and opens again? Is it cleared, or does the component "remember" the most recently typed filter?

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+> NOTE: HOW THIS (AND ANY) FEATURE IS INCLUDED IN IMPLEMENTATION IS UP TO DEVELOPERS
+
+When displaying menu items in the menu lags or is expected to take longer than normal, the menu should indicate that it's loading. The "Loading…" display is not interactive.
+
+![ ](spec-images/select-15.png)
+
+> NOTE: IS THIS NEEDED? IF YES, NEEDS VISUAL DESIGN & TECH WRITER
+
+Menus with dynamically loading menu items may have a "Load more" button. Clicking the "Load more" button should resolve into the "Loading…" indication, then load more menu items and scroll the menu as needed.
```
jattasNI, line 165: I saw [a proposal in SLE](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/565673#1695761223) today for a dropdown (unclear if select or combobox) that is used to pick from hundreds/thousands of items so it initially loads only the first 100 items alphabetically. In that proposal, the only way to see more items was to start typing in the filter, which would trigger a request for items that match that filter text. (aside: I'm not sure what would happen if more than 100 items match the filter text...maybe users would have to refine their filter to see them.)

Do you think that proposed workflow aligns with everything in your spec? Any updates or recommendations we should add?

The main thing I can think to change in this document (if you like that workflow) would be to clarify in the filtering section that the filter should apply to all items, even ones that are on the server and aren't currently visible in the component.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+![ ](spec-images/select-32.png)
+
+`ESC` on a focused/selected menu item cancels the value and closes the menu.
+
+![ ](spec-images/select-33.png)
+
+## Future Considerations
+
+-   Should the select component allow multi-selection? Users on touch-screen devices may see a truncated list of comma-separated selections in the field button, without the ability to hover for a tooltip.
```
jattasNI, line 280: > Should the select component allow multi-selection?

Can we remove this as a Future Consideration since we recommend above that clients wanting multiselect should use the tag picker instead?

packages/nimble-components/src/select/specs/IxD.md
leslieab: Ellipses are not needed for the placeholder, unless long enough to truncate. I've added language to clarify this in the "Field Button" section.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### Icons
+
+The client-user should be able to specify whether they want left icons on menu items or not. Icons show up in menu items to the left of text and are displayed as part of the value in the field button.
+
+![ ](spec-images/select-7.png)
+
+#### Groups
+
+> QUESTION: IS THE SL "VIEW" PICKER A SELECT (SHOULD IT BE)? WHAT ABOUT GROUPS WITHOUT LABELS, WHAT ABOUT A SELECTION GROUP VS. AN ACTION GROUP OF MENU ITEMS (A.K.A MEGA-MENUS)?
```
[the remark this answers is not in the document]
leslieab: I'm going to move it to Future Considerations. But I essentially mean all the different iterations menus can take, for example, with static content, with selectable options, and with command buttons, etc.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,281 @@
...
+
+![ ](spec-images/select-5.png)
+
+> NOTE: NEED VISUAL DESIGN GUIDANCE FOR ICON SIZES, WHAT CAN BE USED AS AN ICON, ETC.
+
+In addition to text, menu items can include an icon to the left of the text, which also appears in the field button when selected. Icons should only be used when they provide more immediate communication to the user than the text. Do not include icons unless all menu items have an icon.
+
+![ ](spec-images/select-7.png)
+
+### Related Components
```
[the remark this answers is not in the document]
leslieab: Added menu button to "Rleated components". Updated to change to "list options". Do we still call the "menu" that the "list options" are in a "menu"?

packages/nimble-components/src/select/specs/IxD.md
leslieab: Sorry, I meant to update this section of the spec. What Meyer and I decided on is that we will provide a general API for a slot that could potentially be used by clients to develop the interaction design provided here is they want, and we may document the progress/loading design pattern in the future, but for now it doesn't make sense to include this at the component level. The proposal you describe should work if that's what they decide they want to do.

packages/nimble-components/src/select/specs/IxD.md
leslieab: Do you have a strong preference on whether it's cleared or "remembers"?

mollykreis: In my experience, most drop-downs with filters clear the filter when it re-opens, so that would be my initial preference (though I'm not sure how strong of a preference that is). I'm wondering if @atmgrifter00 has any thoughts from an implementation point of view.

packages/nimble-components/src/select/specs/IxD.md
leslieab: Except I'm not sure how you have an "optional" select when it requires a default value?

packages/nimble-components/src/select/specs/IxD.md
atmgrifter00: Yes, I think (fairly strongly) that the filter should be clear every time it reopens. Otherwise, if the user wants to search for something else, they would have to have the muscle memory to always press `Esc` to clear the filter, which might conflict with the muscle memory for dismissing the popup.

packages/nimble-components/src/select/specs/IxD.md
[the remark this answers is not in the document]
jattasNI: I don't think we have a standard term for it. Internally usually we say "popup" or "dropdown" to describe the container and "listbox" to describe the list that's inside the popup.

packages/nimble-components/src/select/specs/IxD.md
jattasNI: I suspect we will get requests for a special unset value. A classic use case is picking a user (e.g. assigning a work item owner in azdo) or a state/province/country where it's hard to pick a good default and you want to force the user to select something rather than potentially skip over the field and leave it at a bad default value.

If we get rid of the empty/placeholder state then I'm not sure what we'd tell those clients. Use a combobox? But we don't want to let them invent a state/country by typing in an arbitrary value.

In my experience with selects which allow an unset value, some of them never let you unset it after having set it and some of them do let you unset it, eg. by deleting all the text in the azdo assigned to field, (which is more similar to a combobox).

It's a sneaky problem. My vote would be either to design an answer for unset values in the select now or move it to Future Considerations, but not to forever disallow it.

leslieab: I've simply added the guidance that the empty state should only be used when there are no obvious default list options.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+![ ](spec-images/select-alternative-combobox.png)
+
+> NOTE: SEE FUTURE CONSIDERATIONS FOR MULTI-SELECTION
+
+To allow end-users to multi-select list options, use a [**tag picker**](/packages/nimble-components/src/tag-picker/specs/IxD.md).
+
+![ ](spec-images/select-alternative-picker.png)
+
+To allow end-users to select less than three options, use a [**radio group**](/packages/nimble-components/src/radio-group/specs/IxD.md).
```
jattasNI, line 83: nit: to avoid confusion vs multi-selection
```suggestion
To allow end-users to select from fewer than three options, use a [**radio group**](/packages/nimble-components/src/radio-group/specs/IxD.md).
```

jattasNI (approved):

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+> NOTE: HOW THIS (AND ANY) FEATURE IS INCLUDED IN IMPLEMENTATION IS UP TO DEVELOPERS
+
+When displaying list options and the dropdown lags or is expected to take longer than normal, the client-user should be able to indicate loading via a slot.
+
+![ ](spec-images/select-15.png)
+
+> NOTE: IS THIS NEEDED? IF YES, NEEDS VISUAL DESIGN & TECH WRITER
+
+Dropdowns with dynamically loading list options may have a "Load more" button. Clicking the "Load more" button should resolve into the "Loading…" indication, then load more list options and scroll the dropdown as needed.
```
rajsite, line 169: Kinda generally for the dynamically load more options case, I'd be interested in knowing the workflow for:

1. Showing how to request more loads (this one covers that ✔)
2. Showing that more is loading (this one covers that ✔)
3. Showing that loading failed (should a message be shown? Is it dismissible? would there be only one? Should there be a retry interaction or just click load more again?)
4. How to cancel loading / when loading should be cancelled (should closing the pop-up cancel loading or should it allow loading and updating while closed? How can the user tell loading is done or failed or maybe they should just reopen it to check?)

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+| `DOWN ARROW` or `UP ARROW` | Moves visual focus to the next or previous list option                                                                                                                      |
+| `ESC`                      | Closes the dropdown and moves visual focus to the select                                                                                                                    |
+| `HOME` or `END`            | Moves visual focus to the first or last list option                                                                                                                         |
+| `PAGEUP` or `PAGEDOWN`     | Jumps visual focus up or down ten list options (or to first or last list option)                                                                                            |
+| _Printable characters_     | Moves visual focus to the filter input (if filterable) with the characters typed or to the first list option (if not filterable) that matches the typed character or string |
```
rajsite, line 262: I feel like if focus moves to the filter input there will be an instinct to press tab to move out of the filter input to the option items. Is that a concern?

Is the interaction clear for typing a few characters, pressing down and up a bit, and then trying to edit the filter again? I think I'd expect after typing characters and having the focus jump to the filter input that pushing up would select the last itme, and pushing down would move focus back to the filter input

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+![ ](spec-images/select-alternative-picker.png)
+
+To allow end-users to select from fewer than three options, use a [**radio group**](/packages/nimble-components/src/radio-group/specs/IxD.md).
+
+To allow end-users to trigger an action instead of make a selection, use a [**menu button**](/packages/nimble-components/src/menu-button/specs/IxD.md).
+
+## Design
+
+### Configuration
```
rajsite, line 89: I don't see any discussion of error states. I'm assuming we keep the ability to put the control in an error state and show an error message below the control?

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+![ ](spec-images/select-10.png)
+
+#### Selection
+
+> NOTE: NOT SURE THIS IS A REAL USE CASE FOR SELECT COMPONENT, SEE FUTURE CONSIDERATIONS FOR MULTI-SELECTION
+
+The client-user should be able to specify whether they want to display selection with or without a check mark.
+
+![ ](spec-images/select-11.png)
```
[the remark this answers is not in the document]
leslieab: I was thinking of the table views menu as a use case, but I guess that's a button menu not a select. I'll move it under future work.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+![ ](spec-images/select-14.png)
+
+#### Progress & Loading
+
+> NOTE: HOW THIS (AND ANY) FEATURE IS INCLUDED IN IMPLEMENTATION IS UP TO DEVELOPERS
+
+When displaying list options and the dropdown lags or is expected to take longer than normal, the client-user should be able to indicate loading via a slot.
+
+![ ](spec-images/select-15.png)
```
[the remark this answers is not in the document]
leslieab: That is correct.

packages/nimble-components/src/select/specs/IxD.md
leslieab: Erring on the side of low-impact UX, I'd probably recommend having the loading indicator turn back into the "load more" button after a reasonable amount of time, even if it's closed (if possible, there seem like there could be some edge cases here). I can add this to the spec but I actually can't remember whether @atmgrifter00 and I agreed I should have moved this to Future Considerations or not, because the expectation is that the client devs would implement this if they wanted, and we would not build it into the component.

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -0,0 +1,286 @@
...
+
+Dropdowns with dynamically loading list options may have a "Load more" button. Clicking the "Load more" button should resolve into the "Loading…" indication, then load more list options and scroll the dropdown as needed.
+
+![ ](spec-images/select-16.png)
+
+#### Touch-Screen Devices
+
+On touch-screen devices, the select component should bring up the dropdown in a Nimble or native popover.
+
+![ ](spec-images/select-17.png)
```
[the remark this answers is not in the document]
leslieab: Great point! Do you have an opinion about which approach would be best? I'm going to look into what other filterable selects do for touch screen. 

I think we can agree that having an inline dropdown on a small touch screen device is always a bad experience. Not sure how more hybrid touchscreen computers handle it.

packages/nimble-components/src/select/specs/IxD.md
leslieab: I can't really picture what you're describing. Can you rephrase this a bit?

fredvisser: I've tweaked the proposal to have the fiter input maintaining the text focus to resolve the concerns about focus jumping between the filter bar and the list option. Also, with `Tab`, I'm proposing that it behave like `Esc` – i.e. not a valid way to select an item, but just cancel out of the dialog.

packages/nimble-components/src/select/specs/IxD.md
fredvisser: Moved to "Future Considerations"

PR #1521, opened 2023-09-12
Combobox IxD Specification

leslieab:
# Pull Request

## 🤨 Rationale

For #1303 - Select and Combobox Refresh HLD
- Interaction design specification for combobox component

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

- [X] Submit project documentation
- [x] Review


packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+- To present a pre-defined list of options
+- To allow end-users to add custom values and/or options
+- For entering settings or preference values that don't (yet) exist
+- For single selection
+
+When not to use:
+- When the options are well-known and straightforward
+- When there are fewer than three options
+- For complex and/or hierarchical data
+- For multi selection
```
atmgrifter00, line 23: I'm not sure I know what "When the options are well-known and straightforward" means. The guidance I would give on when to use a `Combobox` is "when the user wants to provide an input for text entry coupled with a popup that allows for quick selection and filtering". The key part of using the Combobox is that it is semantically the same as a `TextField` (and not a `Select`), but comes with "hints" for allowing the user to make a quick selection.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+#### Helper Text
+Only include helper text to provide information that aids the end-user in choosing the right selection. *See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
+
+#### Field Input
+Comboboxes can be empty by default or have a default selection from the menu. Empty comboboxes should include placeholder text that propts the end-user to enter a custom value or select an option from the menu. Placeholder text should always follow the pattern "Enter or select [thing(s)]", for example "Enter or select address". Ellipses are not needed. Use sentence casing.
+
+![ ](spec-images/combobox-2.png)
+
+#### Menu & Menu Items
+A list of options to select, displayed as an open state.*See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
```
atmgrifter00, line 49: It may be worth pointing out that the options used in the `Select` can have a value assigned to them that is distinct from the display value that has meaning to the `Select` (and the user). For the `Combobox` this is not the case, any value provided to an option is ignored.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Progress & Loading
+When displaying menu items in the menu lags or is expected to take longer than normal, the menu should indicate that it's loading. *See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
+
+#### Touch-Screen Devices
+On touch-screen devices, the combobox field input should bring up the native keyboard.
+
+![ ](spec-images/combobox-5.png)
+
+The combobox menu button should bring up the menu in a Nimble or native tray instead of a popover. *See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
```
atmgrifter00, line 109: This will have to be updated in the same was as the `Select` IxD document was.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+
+`ENTER` on the focused input field commits the typed, autocompleted, or selected value and closes the menu (if open). The caret remains visible after commit.
+
+![ ](spec-images/combobox-20.png)
+
+`ESC` clears any text in the field input.
+
+![ ](spec-images/combobox-21.png)
+
+Standard single line text editing keys, for example `DELETE` and `ALT` + `RIGHT ARROW`, should work when focus is in the field input (not pictured).
```
atmgrifter00, line 180: I don't believe `ALT` + `RIGHT ARROW` is a standard single line text editing key combination. Did you just mean `RIGHT ARROW` (which would just move the cursor one character to the right if the cursor isn't at the end of the text)?

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+
+![ ](spec-images/combobox-3.png)
+
+#### Autocomplete
+The client-user should be able to specify autocomplete behavior (inline, list, both, none). Autocomplete is accessed via the field input (not pictured). *See [current implementation](https://nimble.ni.dev/storybook/?path=/docs/components-combobox--docs) for more information.*
+
+#### Add New
+> NOTE: NEEDS DEV REVIEW, VISUAL DESIGNER FOR STATES, & TECH WRITER
+
+The client-user should be able to specify whether custom entries can be added as a menu item in the menu or not. In the former, the menu should display a menu item that new items can be added indicate that a new item can be added. In the later, the menu does not show up at all.
```
atmgrifter00, line 87: This feature I question the actual value of. In proper implementations within a UI framework like Angular, this addition of an item to the set of options in the dropdown will just happen automatically if a user commits a value that is currently not present in the set. This will happen in situations where the set of options are being generated from some observable, typically being populated by a service getting the set from some server state. It's not unreasonable to expect that when a user commits a new value that that value will makes its way to the server and ultimately update the observable that the `Combobox` is using to populate its options with.

In situations where the `Combobox` has some static set of options (so, not provided via an observable), then adding a new options to be used in the dropdown in successive operations isn't that useful, because it will ultimately be discarded on a page reload.

Ultimately, I'd like to hear the reasoning a client would have to want a feature like this, because at the moment I'm not sure what functionality it will truly provide.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,212 @@
...
+| Field input | Input that displays the current selected or entered option | 
+| Menu button | Click area that opens the menu | 
+| Menu | A list of options to choose from displayed as an *open* state | 
+| Menu item | A selection the end-user can make, shown with other options in the menu | 
+
+#### Label
+Comboboxes should always have a label, unless approved by an interaction designer or user researcher (usually in toolbar use cases). *See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
+
+#### Helper Text
+Only include helper text to provide information that aids the end-user in choosing the right selection. *See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information.*
```
atmgrifter00, line 41: See [comment on `Select` IxD](https://github.com/ni/nimble/pull/1520#discussion_r1330757227) for discussion on this feature.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: I think of a `combobox` as a `text field` and `select` combined. Without a popup menu, the `combobox` might as well be a `text field` with autocomplete and get rid of the dropdown button altogether. But a `combobox` with or without autocomplete, should have a popup menu that indicates selection the same way it does in a `select` component. But anyway, I updated this.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: Correct, I meant `SHIFT`.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: My thought process was that by default, the user can commit a value that isn't provided as an option in the menu, and while that value can be committed, it's not added as an option in the menu when that happens. I thought there is a use case for this when the end-user wants to assign a system (for example) to a workspace that doesn't exist yet, but that no longer appears to be accurate. 

I hadn't thought about it technically, but your points make sense. As I understand it, you are saying that the component itself is agnostic to whether the end-user can enter and commit a value that does not exist in the menu, because whether the commit is successful or errors is up to how it was implemented in an app by app developers. Is that correct? So there is no reason to have a specific "Add new" indicator in the dropdown, especially tied to some configuration on the component. But, like the loading indicator in the `select` spec, we may provide a slot in the menu and documented guidance about showing an "Add item" indicator.

atmgrifter00: I think you are understanding things correctly. However, given that, I still don't think we would provide any documented guidance about an 'Add item' indicator. My concern there is that we would be possibly leading developers down an unnecessary road about how to introduce a new option into the option set, and we would have to provide pretty detailed instructions/caveats of what a user should expect if they were to introduce a mechanism like that into the menu.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Secondary Text
+
+The client-user should be able to specify whether they want secondary text on menu items or not. _See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information._
+
+#### Prefix & Suffix
+
+> NOTE: NEEDS VISUAL DESIGN
+
+The client-user should be able to specify a prefix and/or suffix that appears in the field.
```
atmgrifter00, line 96: This feels like a pretty loaded feature, that I would suggest we move to the 'Future Considerations' section. There are a few concerns I have around this:
- This ultimately feels like a "universal" feature for various inputs to have like the `TextField` and `NumberTextField`. As a result, if we decide such a feature makes sense (which I think has room for debate) then we would need to be deliberate about its API
- There are some potential localization concerns particularly if we were promoting usages like putting a '$' as a prefix. Some currency types wouldn't put the symbol at the front, but at the end, however using the "suffix API" to do it might be inadequate because of spacing.
- Should we instead just promote a more general slotting API, where users can slot something at either the front and/or the end? While we lose some styling control here, it offers greater flexibility and avoids localization issues

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+![ ](spec-images/combobox-3.png)
+
+#### Autocomplete
+
+The client-user should be able to specify autocomplete behavior (inline, list, both, none). Autocomplete is accessed via the field input (not pictured). _See [current implementation](https://nimble.ni.dev/storybook/?path=/docs/components-combobox--docs) for more information._
```
atmgrifter00, line 102: We actually have an opportunity here to break from the FAST behavior of how autocomplete works, and offer something that I think is more in demand, which is the ability to filter against _any_ text within the option text, instead of just from the front. The only reason, that I know of, for why FAST limits the filter to the front of the text is because of how they do inline autocomplete, where they highlight the rest of the text that has been autocompleted, which is all done within the input section of the component.

So, I mention this in case you wanted to conceive of your ideal behavior for how autocomplete should work for the `Combobox`. The main interaction that I think we need to describe in detail, that is an unknown at the moment, is if the `Combobox` has an `autocomplete` of either `both` or `inline` (this results in the text in the input autocompleting, with untyped text currently being highlighted), what should happen as the user types, particularly if the text matches text _not_ at the front of the option text. The main alternative I can think of is that we don't highlight any of the autocompleted text, but we still update the text in the input field. The drawback to this is that your cursor could be jumping around if your match was triggering against the beginning at times and other places as you continue typing.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: I wasn't trying to be prescriptive with the prefix/suffix content, the image is just an example of what a prefix and suffix might look like. I was thinking of this as a general slotting API for the front and the end, and those slots have a specific visual design that distinguishes them from the displayed value. And agreed, there are other input components that would benefit from this feature.

atmgrifter00 (approved):

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,238 @@
...
+
+When the menu is open, clicking anywhere outside of a menu item and the text input closes the menu with no changes.
+
+![ ](spec-images/combobox-12.png)
+
+If the combobox allows autocomplete, typing in the field input will show menu items matching the typed value in the field and/or in the menu. The input field autocompletes the first matching menu item. The menu indicates selection on the first matching menu item.
+
+![ ](spec-images/combobox-13.png)
+
+Typing text that does not match any existing menu items either allows the end-user to add a new menu item to the menu or not (see options).
```
atmgrifter00, line 168:
```suggestion
Typing text that does not match any existing menu items either allows the end-user to add a new menu item to the menu or not (see Future Considerations).
```

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,238 @@
...
+When to use:
+
+-   To provide an input for text entry coupled with a popup for quick selection and filtering
+-   To allow end-users to enter custom values
+-   For entering settings or preference values that don't (yet) exist
+-   For single selection
+
+When not to use:
+
+-   When text entry is not needed for selection
```
mollykreis, line 25: I'm not sure what _When text entry is not needed for selection_ means. Could you clarify? For example, does that mean when the user doesn't need to be able to enter a value that isn't in the pre-populated list?

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,238 @@
...
+-   When there are fewer than three options
```
mollykreis, line 26: What would be the recommendation for when there are fewer than 3 options? Particularly because of how helpful the autocomplete is (i.e. the dropdown options) and the necessity to enter any value, I would think you'd still want to use a combobox regardless of how many options you have.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,238 @@
...
+| Menu        | A list of options to choose from displayed as an _open_ state           |
+| Menu item   | A selection the end-user can make, shown with other options in the menu |
+
+#### Label
+
+Comboboxes should always have a label, unless approved by an interaction designer or user researcher (usually in toolbar use cases). _See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information._
+
+#### Field Input
+
+Comboboxes can be empty by default or have a default selection from the menu. Empty comboboxes should include placeholder text that propts the end-user to enter a custom value or select an option from the menu. Placeholder text should always follow the pattern "Enter or select [thing(s)]", for example "Enter or select address". Ellipses are not needed. Use sentence casing.
```
mollykreis, line 48: >Ellipses are not needed. Use sentence casing.

Could you clarify whether these are guidelines for the placeholder or the value (or both)?

mollykreis (approved):

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: With fewer than three options, a radio group (or radio button group, which we don't really have a concept for) should be used. This way the user can see all the options immediately, without requiring a click for progressive disclosure. I've added this info to the "Related Components" section.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: Clarified in file.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Groups
+
+The client-user should be able to create groups for list items. _See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information._
+
+> NOTE: NEEDS VISUAL DESIGN
+
+The client should be able to specify whether they want the group name to be visible on the component when the dropdown is closed.
+
+![ ](spec-images/combobox-26.png)
```
rajsite, line 89: Unlike a select, the group of a combobox is not associated with a value. A selection is a unique selection of an item in a list. A combobox is a text input where you use the list of items to autocomplete the text. In a combobox there is no "selection", just autocompleting of some text.

For example, the options:
```
Texas
- Apple 
- Orange
Florida
- Cranberry
- Orange
```

For a select the option of orange under texas and florida are unique and different they may associate to unique numbers 2 and 4 for example that can be differentiated.
For a combobox clicking either option autocompletes the text as "Orange" there is no association to the group, just the text "Orange". You can't tell what group the text was autocompleted from when clicked.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Secondary Text
+
+The client-user should be able to specify whether they want secondary text on list items or not. _See [Select](/packages/nimble-components/src/select/specs/IxD.md) for more information._
```
rajsite, line 93: We could show secondary text in the options list, but similar to above there is no selection of an option, just autocompletion of text. so secondary text / group name would not appear in the text box of the combobox.

packages/nimble-components/src/combobox/specs/IxD.md
rajsite: I'd like to push this to future work unless we have specific known expected use cases today, i.e. this is used in exisiting designs. If so some examples would help because I also feel like this is a bit of a can of worms.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+> NOTE: NEEDS VISUAL DESIGN
+
+The client-user should be able to specify a prefix and/or suffix that appears in the field.
+
+![ ](spec-images/combobox-3.png)
+
+#### Autocomplete
+
+The client-user should be able to specify autocomplete behavior (inline, list, both, none). Autocomplete is accessed via the field input (not pictured). _See [current implementation](https://nimble.ni.dev/storybook/?path=/docs/components-combobox--docs) for more information._
```
rajsite, line 105: Are there actually clear use-cases where all of these modes are actually valuable? I feel like really only the list and none modes are needed.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+![ ](spec-images/combobox-alternative-select.png)
+
+To allow end-users to multi-select list items, use a [**tag picker**](/packages/nimble-components/src/tag-picker/specs/IxD.md).
+
+![ ](spec-images/combobox-alternative-picker.png)
+
+To allow end-users to select less than three options, use a [**radio group**](/packages/nimble-components/src/radio-group/specs/IxD.md).
+
+## Design
```
rajsite, line 69: I don't see any discussion of error states. I'm assuming we keep the ability to put the control in an error state and show an error message below the control?

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+![ ](spec-images/combobox-24.png)
+
+`RIGHT ARROW` or `LEFT ARROW` moves visual focus to the field input without closing the dropdown and moves the caret one character to the right or left.
+
+![ ](spec-images/combobox-25.png)
+
+## Future Considerations
+
+-   Helper text for input components (where it should be located, how it should be accessed).
+-   Do clients want an "Add new" list item option? Do we want to document how clients might show an "Add new" list item indicator in the dropdown?
```
rajsite, line 241: I feel like this is more relevant to the filterable select. The combobox is a text field, you can type whatever value you want. At the very least it applies to both combobx and filterable select.
It does open some questions like would that button / option exist under each group (i.e. how do you choose what group to add the option to)? Or would there just be a generic add option command at the end of the list and the client should show some other UI (dialog, slideout, etc) for choosing the group, etc.

packages/nimble-components/src/combobox/specs/IxD.md
```diff
@@ -0,0 +1,241 @@
...
+
+![ ](spec-images/combobox-12.png)
+
+If the combobox allows autocomplete, typing in the field input will show list items matching the typed value in the field and/or in the dropdown. The input field autocompletes the first matching list item. The dropdown indicates selection on the first matching list item.
+
+![ ](spec-images/combobox-13.png)
+
+Typing text that does not match any existing list items either allows the end-user to add a new list item to the dropdown or not (see Future Considerations).
+
+![ ](spec-images/combobox-16.png)
```
rajsite, line 173: If this is all future considerations can we move it there to prevent confusion?

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: @atmgrifter00 thoughts?

rajsite: There isn't a name for this new mode right? Let's call it `filter`. +1 for just `filter` unless y'all know of reasons for us to implement / keep the other modes.

packages/nimble-components/src/combobox/specs/IxD.md
atmgrifter00: I'd argue that this _can't_ be used for a filterable select. Ultimately in a select (at least with the typical usage of it in something like Angular) there is a model value associated with each option. There would be no good way to provide this feature that would allow a user to associate an appropriate model with a new option.

rajsite: Implementation-wise I think clicking that button just fires an event, I wouldn't expect it to mutate any configured state. It's just a purpose specific action button.

leslieab: How would an end-user add a new list option if they can't type the value into a select?

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: I thought clients were specifically looking for this feature in the combo box? To be clear, you're saying the combobox menu can show groups, but the component itself cannot display the group of the current selection?

atmgrifter00: As implemented (by FAST) @rajsite is correct in that there would be no good way for the combobox to display the group correctly when a user selects a value. All the combobox knows is that it has a specific string as its value.

packages/nimble-components/src/combobox/specs/IxD.md
atmgrifter00: It would have to be an auxiliary UI away from the select component itself. This is another reason that I'm not entirely keen on this as a feature. The user would not only have to supply the text to display in the dropdown, but also the model to use for that option.

jattasNI (approved):

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: Added, and you are correct.

packages/nimble-components/src/combobox/specs/IxD.md
leslieab: Alright, I've removed this specific text.

change/@ni-nimble-components-ffc88008-f921-4cf2-aa10-dd2e6c97f3e8.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @leslieab saw some recent pushes, just a reminder when ready for a re-review to hit the reset reviewer button for those who have not approved.

If the changes were significant enough that you think current approved reviewers should be reset to take another look you could chose to reset those reviewers for a re-review as well.

image

leslieab: Ooh, I didn't even know that existed. I will do that, thanks!

packages/nimble-components/src/combobox/specs/IxD.md
fredvisser: Closing the loop on this – I agree that we don't have a use case for anything more complex than `none` or `list` today. Our usage docs will reflect this, but we don't need to remove the other options today.

PR #1524, opened 2023-09-12
Remove fixture parent

mollykreis:
# Pull Request

## 🤨 Rationale

We decided that it wasn't a good idea to allow a test's fixture to set a custom parent element, and instead, the parent should explicitly be part of the fixture by putting it in the template. See discussion here: https://github.com/ni/nimble/pull/1499#discussion_r1322182443

## 👩‍💻 Implementation

- Update the `nimble-table-column-date-text` tests to include the `nimble-theme-provider` within the fixture's template rather than setting it as the parent of the fixture.
- Remove `parent` option on our fixtures to disallow setting a parent that isn't an element within the template
- Fix typo in `nimble-table-column-number-text` tests

## 🧪 Testing

Ran the unit tests

## ✅ Checklist


mollykreis:
@m-akinc, can you buddy this for me?

m-akinc (approved):

packages/nimble-components/src/utilities/tests/fixture.ts
```diff
@@ -31,12 +31,6 @@ export interface FixtureOptions {
      */
     document?: Document;
```
jattasNI, line 33: There's a comment at the top of the file tracking changes we've made versus the original FAST source. Who knows if we'll ever adopt their version (the linked issue is marked closed in the FAST Element 2.0 branch) but it couldn't hurt to add another comment mentioning this new diff.

jattasNI (approved):

rajsite (approved):

PR #1526, opened 2023-09-13
Rich text editor | Update dependency on prosemirror-markdown

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Fixes [Bug 2518928](https://dev.azure.com/ni/DevCentral/_workitems/edit/2518928): Rich text editor | Bold italics not updating as expected

In markdown-serializer, if the the markdown string contains leading or trailing whitespaces within marks, it will not parsed to an expected rich text content. For example, entering "**Bold *Italics*** **bold**" into the editor, the actual markdown serializer output is `**Bold *italics *bold**` which renders as below in viewer.
image

Expected output for the above example should be `**Bold *italics* bold**`

## 👩‍💻 Implementation

Bumped the `prosemirror-markdown` version from `1.11.0` to `1.11.2` which fixes [`expelEnclosingWhitespace`](https://github.com/ProseMirror/prosemirror-markdown/blob/c7210d0e55c82bfb0b2f7cba5dffe804575fafb3/src/to_markdown.ts#L21C3-L21C27) property to expel any leading or trailing whitespaces within a mark to outer nodes.

## 🧪 Testing

1. Manually tested by getting markdown string for different bold and italics mixed cases and loading it in the viewer.
2. Add test cases in the `markdown-serializer.ts` for different scenarios with the example mentioned above and a few other test cases that cause the leading and trailing whitespace. Please note that the input HTML string given to the tests is the same as how the `Tiptap` renders bold and italics cases. For example, strong has a higher priority than italics, so the HTML will be like
    1. Italics between bold (**Bold *italics* bold**):
        ```html
        <strong>
            Bold 
            <em>italics </em>
            bold
        </strong>
        ``` 
    2. Bold between italics (*Italics **bold** italics*):
        ```html
        <em>Italics </em>
        <strong>
             <em>bold </em>
        </strong>
        <em>italics</em>
        ``` 

## ✅ Checklist


change/@ni-nimble-components-20c7f83a-27bf-4fe3-aeb3-86e15e1561c5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update dependency on prosemirror-markdown",
```
vikisekarNI, line 3:
```suggestion
  "comment": "Mixed bold and italics formatting corner cases fixed. Update dependency on prosemirror-markdown",
```

vikisekarNI (approved):

jattasNI (approved):

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -56,6 +56,46 @@ describe('Markdown serializer', () => {
                 html: '<strong><em>Bold and Italics</em></strong>',
                 markdown: '***Bold and Italics***'
             },
+            {
+                name: 'Italics without spaces in between bold texts',
+                html: '<strong>Bold<em>italics</em>bold</strong>',
```
rajsite, line 61: As discussed in the following comment I have significant concerns with the usefulness of these tests: https://github.com/ni/nimble/pull/1496#discussion_r1324888037

As described in the PR description "Please note that the input HTML string given to the tests is the same as how the Tiptap renders bold and italics cases. " 

My concern is that these tests are manually created html strings and there is nothing to assert that when we update tiptap, etc that the HTML that would actually be generated won't change. The tests only reflect the current exact version of the dependencies used and I do not think are significantly useful for testing potential regressions or change in behavior.

I won't block this PR but we need to discuss an different testing strategy.

rajsite (approved):

rajsite:
You don't need to wait for @m-akinc's review for a straightforward dependency upgrade. We need additional discussion for tests separate from this PR.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
vivinkrishna-ni: Yes, we agree with your concerns whatever is posted here and in the other PR comment regarding the testing strategy followed for serializer logic. I have created a tech debt issue https://github.com/ni/nimble/issues/1533 to update the testing approach but we will try to take it immediately after this PR.

I am keeping these test cases as it is for now because these tests will check the fix for `expelEnclosingWhitespace` in the ProseMirror and might be captured if it breaks in future releases of ProseMirror. Also, when we update the testing approach, we will revisit the cases and add exactly what is required.

PR #1529, opened 2023-09-13
Allow hiding increment/decrement step buttons on numeric

### LINKED ISSUE #1528: Nimble Numeric field | Option to hide "+-" buttons
## 😯 Problem to Solve

In our app we got requirement to hide the "+-" buttons used to increment and decrement the number value. We have aligned with Jesse and came to know that there is an API already available for this, but its not documented. 

## 💁 Proposed Solution
- Option  to hide "+-" icon buttons, as its already available it would be better to have this documented.

jattasNI:
# Pull Request

## 🤨 Rationale

Fixes #1528

## 👩‍💻 Implementation

1. Storybook documentation for existing FAST attribute `hide-step` and some other attributes
2. Update matrix tests include `hide-step`
3. Angular directive and tests. Note that FAST defaults the attribute to `false` rather than `undefined`.
4. Blazor wrapper and test

## 🧪 Testing

1. Manually verified in Storybook
2. Autotests for Angular and Blazor follow existing patterns of other boolean attributes

## ✅ Checklist


jattasNI:
@atmgrifter00 could you be the initial buddy for this change please?

atmgrifter00:
While I am fine with the changes, it would be good to get @NIbokeefe to give his blessing. The interesting aspect for you, Brandon, I think, is that these changes will allow a user to configure a `NumberField` such that it will be visually identical to a `TextField`.

atmgrifter00 (approved):

NIbokeefe:
> While I am fine with the changes, it would be good to get @NIbokeefe to give his blessing. The interesting aspect for you, Brandon, I think, is that these changes will allow a user to configure a `NumberField` such that it will be visually identical to a `TextField`.

Sounds like it's a needed requirement, so I won't advise against it.  Like you mentioned the (+,-) is major indicator to the type of control, outside of the possibility of the label adding context.  

Have we considered disabling the buttons vs hiding them?  The value would still be enabled, which is a consistent point of input. So, I can't imagine the user would be any more or less confused by the buttons missing vs greyed out to interact directly with the value.

atmgrifter00:
> > While I am fine with the changes, it would be good to get @NIbokeefe to give his blessing. The interesting aspect for you, Brandon, I think, is that these changes will allow a user to configure a `NumberField` such that it will be visually identical to a `TextField`.
> 
> Sounds like it's a needed requirement, so I won't advise against it. Like you mentioned the (+,-) is a major indicator to the type of control, outside of the possibility of the label adding context.
> 
> Have we considered disabling the buttons vs hiding them? The value would still be enabled, which is a consistent point of input. So, I can't imagine the user would be any more or less confused by the buttons missing vs greyed out to interact directly with the value.

I don't think disabling is a good option for cases where there is no app state that would cause the buttons to become enabled. If things are _always_ disabled they should just not be shown (IMO).

jattasNI:
I'll bring this up for discussion at today's demo meeting to resolve the usage guidance but the current implementation is ready for review.

rajsite (approved):

msmithNI (approved):

PR #1532, opened 2023-09-13
Remove unexpected whitespace at start/end of anchor

### LINKED ISSUE #1508: nimble-anchor is wider than the native anchor in its shadow DOM
# 🐛 Bug Report

The `nimble-anchor` element's width includes some extra space on one or both of its sides, compared to the bounds of the native `a` element in the shadow root.
image
image

## 💻 Repro or Code Sample

Examine any `nimble-anchor` e.g. in Storybook.

## 🤔 Expected Behavior

Bounds of the `nimble-anchor` should be the same as the inner `a` element.

## 😯 Current Behavior

There is some kind of "padding" on one or both sides.

## 🔦 Context

Noticed when reviewing the link support in the rich-text editor.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1508

## 👩‍💻 Implementation

The issue is that whitespace inside the template was being rendered.

Inlined `startSlotTemplate` so that we could remove whitespace internal to it. (We had previously done the same thing to the `endSlotTemplate` for the same reason.) Removed all other whitespace in the template that was causing a problem.

## 🧪 Testing

Tested in Storybook. Created a new Chromatic test.

## ✅ Checklist


packages/nimble-components/src/anchor/template.ts
```diff
@@ -43,19 +41,26 @@ AnchorOptions
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
         ${ref('control')}
+    >${
+    /* Start and End slot templates inlined to avoid extra whitespace.
+        See https://github.com/microsoft/fast/issues/6557 */ ''
+}${
+    /* Whitespace intentionally avoided between tags for inline styles */ ''
+}<span
```
rajsite, line 49: Can you merge these two comments together? no compelling reason to have them separate back to back

rajsite, line 49: All the mixing of styles of no space between tags and space between tags makes this challenging to read. Can we use a consistent format of removing all the space between tags and an attribute per line

packages/nimble-components/src/anchor/tests/anchor-matrix.stories.ts
```diff
@@ -85,3 +85,12 @@ export const textWrapping: StoryFn = createStory(
     </p>
     `
 );
+
+export const inlineSpacing: StoryFn = createStory(
+    html`<style>
+        * {
+            font: var(${bodyFont.cssCustomProperty});
+        }
```
rajsite, line 93: stories aren't encapsulated in shadow roots and weird things can happen with too broad styles. Please write the style in a way that won't conflict with other things. Maybe add a wrapper span around the HTML with a class and target the class instead of targeting all elements.

jattasNI (approved):

packages/nimble-components/src/anchor/template.ts
m-akinc: I tried to format it more consistently. Hopefully it is easier to read now.

packages/nimble-components/src/anchor/tests/anchor-matrix.stories.ts
```diff
@@ -85,3 +85,12 @@ export const textWrapping: StoryFn = createStory(
     </p>
     `
 );
+
+export const inlineSpacing: StoryFn = createStory(
+    html`<style>
+        .inlineLink {
```
rajsite, line 91: we generally use kebab case for class names. If we need to capture that in the CSS docs or something we should. i.e. `.inline-link`

rajsite (approved):

packages/nimble-components/src/anchor/tests/anchor-matrix.stories.ts
m-akinc: Added a note to our coding conventions doc. That seemed like a better place than the CSS docs, because the CSS only references class names--it does not define them.

PR #1541, opened 2023-09-15
New icons for indent and outdent list level

### LINKED ISSUE #1531: Add 2 icons to Token Library for Rich Text Editor
arrow-left-list_16x16
arrow-right-list_16x16

SVG files are attached to implement icons into the Nimble Token Library.

This work pertains to:
https://github.com/ni/nimble/issues/1463

jattasNI:
# Pull Request

## 🤨 Rationale

Resolves #1531 

## 👩‍💻 Implementation

Followed the CONTRIBUTING instructions for adding icons. I also updated them to capture the results of a discussion about icon name.

I chose the file names "indent" and "outdent" based on FontAwesome's equivalent icons [indent](https://fontawesome.com/search?q=indent&o=r) and [outdent](https://fontawesome.com/search?q=outdent&o=r). I added synonyms for "increase list level" and "decrease list level" and for "indent left" and "indent right".

## 🧪 Testing

Verified icon severity affects the color as expected.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
+++ b/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
@@ -293,7 +293,7 @@ export const iconMetadata: {
         tags: ['history', 'timer']
     },
     IconIndent: {
-        tags: ['increase-list-level', 'right-arrow-three-lines']
+        tags: ['increase-list-level']
     },
     IconIndeterminantCheckbox: {
         tags: ['selection']
--- a/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
+++ b/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
@@ -293,7 +293,7 @@ export const iconMetadata: {
         tags: ['history', 'timer']
     },
     IconIndent: {
-        tags: ['increase-list-level']
+        tags: ['increase-list-level', 'indent-right']
     },
     IconIndeterminantCheckbox: {
         tags: ['selection']
@@ -359,7 +359,7 @@ export const iconMetadata: {
         tags: ['order']
     },
     IconOutdent: {
-        tags: ['decrease-list-level']
+        tags: ['decrease-list-level', 'indent-left']
     },
     IconPaste: {
         tags: ['clipboard']
--- a/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
+++ b/packages/nimble-components/src/icon-base/tests/icon-metadata.ts
@@ -292,6 +292,9 @@ export const iconMetadata: {
     IconHourglass: {
         tags: ['history', 'timer']
     },
+    IconIndent: {
+        tags: ['increase-list-level', 'right-arrow-three-lines']
+    },
     IconIndeterminantCheckbox: {
         tags: ['selection']
     },
@@ -354,6 +357,9 @@ export const iconMetadata: {
     },
     IconNumberList: {
         tags: ['order']
+    },
+    IconOutdent: {
+        tags: ['decrease-list-level']
     },
     IconPaste: {
         tags: ['clipboard']
--- a/packages/nimble-tokens/CONTRIBUTING.md
+++ b/packages/nimble-tokens/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Tokens are generated using the [Style Dictionary](https://amzn.github.io/style-d
 
 ### Icon naming
 
-Use [FontAwesome](https://fontawesome.com) names for icons. If FontAweseome doesn't contain an equivalent icon, minimize the use of metaphors; choose descriptive and unambiguous names instead.
+Use [Font Awesome](https://fontawesome.com) names for icons. If Font Awesome doesn't contain an equivalent icon, minimize the use of metaphors; choose descriptive and unambiguous names instead.
 
 | ✅ Descriptive name    | ❌ Metaphor            |
 |------------------------|------------------------|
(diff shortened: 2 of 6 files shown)


packages/nimble-components/src/icon-base/tests/icon-metadata.ts
```diff
@@ -292,6 +292,9 @@ export const iconMetadata: {
     IconHourglass: {
         tags: ['history', 'timer']
     },
+    IconIndent: {
```
rajsite, line 295: The naming convention is for the icon name to be a literal description of the visuals. Looking across the terms on our icons today I'd think triangle-right-bars and triangle-left-bars

I'd use "indent" and "outdent" as tags / higher-level concepts since they don't describe the visual itself.

jattasNI: I had the same convention in mind but was swayed by the [icon naming section of the CONTRIBUTING doc](https://github.com/ni/nimble/blob/main/packages/nimble-tokens/CONTRIBUTING.md#icon-naming) which says "Prefer [FontAwesome](https://fontawesome.com/) names for icons." 

Of course, it also goes on to say "Minimize the use of metaphors; choose descriptive and unambiguous names instead" which could be contradictory. I think our convention needs to be more clear - if the Font Awesome name uses a metaphor, should we follow it or write our own literal description? 

I think @fredvisser has the clearest vision here so I'd be happy to delegate the decision to him. To start the conversation, here's a possible tweak to the convention:

1. If Font Awesome has an equivalent icon, use its name. Small tweaks for clarity are allowed (e.g. modifiers to distinguish similar icons)
2. If Font Awesome doesn't have an equivalent icon, use a description of the icon's appearance:
   1. if the appearance is unambiguously a physical object, use that object's name (key, gear)
   2. if the appearance is a metaphor that's unambiguously used for one thing and one thing only, use that thing's name (wifi, list)
   3. if the meaning might be ambiguous, describe the appearance literally (three dots, triangle with line)

Alternatively our policy could be "ask Fred what the name should be" 😆

rajsite: Maybe what's catching me more off guard is that our icons don't look as unambiguously like indent and outdent icons as the FontAwesome examples. I could see ours representing "add to list" or "remove from list" or things along that nature as well.

rajsite: > If Font Awesome has an equivalent icon, use its name.

From the PR that made that contributing change the icons that were added were all named very literally based on the visual. Maybe at that time it was assumed most icons in FontAwesome would be named that way. 

Maybe the guidance could be "Use the FontAwesome name if the name is a literal description of the visual. If the name is a metaphor of the visual it can be used if the visual we create is unambiguously, universally, and singularly represents that metaphor. Otherwise if the new visual could be used to represent multiple metaphors then create a literal name".

jattasNI: Decided offline to match FontAwesome exactly (even when we don't love their name) to try to avoid having long debates about names. I'll update the CONTRIBUTING docs to say so.

I'll also reach out to Brandon to see if he can make the icon less ambiguous.

packages/nimble-tokens/CONTRIBUTING.md
```diff
@@ -30,15 +30,15 @@ Tokens are generated using the [Style Dictionary](https://amzn.github.io/style-d
 
 ### Icon naming
 
-Prefer [FontAwesome](https://fontawesome.com) names for icons. I.e. Minimize the use of metaphors; choose descriptive and unambiguous names instead.
+Use [FontAwesome](https://fontawesome.com) names for icons. If FontAweseome doesn't contain an equivalent icon, minimize the use of metaphors; choose descriptive and unambiguous names instead.
```
rajsite, line 33:
```suggestion
Use [Font Awesome](https://fontawesome.com) names for icons. If Font Awesome doesn't contain an equivalent icon, minimize the use of metaphors; choose descriptive and unambiguous names instead.
```

packages/nimble-tokens/CONTRIBUTING.md
```diff
@@ -30,15 +30,15 @@ Tokens are generated using the [Style Dictionary](https://amzn.github.io/style-d
...
 
 | ✅ Descriptive name    | ❌ Metaphor            |
 |------------------------|------------------------|
 | `key`                  | `access-control`       |
 | `cog`                  | `system-configuration` |
 | `arrow-left-from-line` | `logout`               |
 
-Add all appropriate metaphors and synonyms to the `icon-metadata.ts` file, so clients can quickly find icons in Storybook.
+Add all appropriate metaphors and synonyms to the `icon-metadata.ts` file, so clients can quickly find icons in Storybook. You can find ideas for synonyms in the [FontAwesome metadata](https://github.com/FortAwesome/Font-Awesome/blob/master/metadata/icons.yml).
```
rajsite, line 41:
```suggestion
Add all appropriate metaphors and synonyms to the `icon-metadata.ts` file, so clients can quickly find icons in Storybook. You can find ideas for synonyms in the [Font Awesome metadata](https://github.com/FortAwesome/Font-Awesome/blob/master/metadata/icons.yml).
```

rajsite (approved):
Approved naming and docs update, icon asset needs to be updated

packages/nimble-components/src/icon-base/tests/icon-metadata.ts
jattasNI: Updated with the versions of the icons that got team signoff.

PR #1543, opened 2023-09-15
Fix bug in table when reordering columns

### LINKED ISSUE #1536: CellViewTemplate and CellState get temporarily out of sync when reordering columns
# 🐛 Bug Report

When columns are reordered, the `cellViewTemplate` and `cellState` get temporarily out of sync with each other while all the bindings are binding updated on the cell. As a result, the `cellState` (such as the `columnConfig`) is handed to the wrong `cellViewTemplate`. This means that the `columnConfig` type that a given `cellViewTemplate` is expecting isn't what it is trying to use. This can lead to a number of different problems, most notably exceptions being thrown because types are different than expected at runtime.

## 💻 Repro or Code Sample

- Create a table with a number-text column and a date-text column.
- Programmatically swap the order of those two columns
- Notice exception in the console because the `formatter` in each `columnConfig` object was not the expected type

## 🤔 Expected Behavior

The table should gracefully reorder the columns without transitioning through an invalid or inconsistent state.

## 😯 Current Behavior

When reordering columns, the table (particularly the table cell views) transition through an invalid state, which can cause a number of different problems. In the end everything ends up in the correct state, but we may throw a bunch of exceptions and run into other problems on the way to that correct state.

## 💁 Possible Solution

Restructure the `nimble-table-row` template to ensure that the `cellViewTemplate` and `cellState` for a given `nimble-table-cell` can't get out of sync.

## 🔦 Context

I found this working on the Data Table grid. Specifically swapping a DateText column and a NumberText column causes an exception to be thrown because both columns have `columnConfig` objects with a `formatter`, but the `formatter` has different types based on the column.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1536 

## 👩‍💻 Implementation

The problem that was occurring with the bug is that `cellState` and `cellViewTemplate` binding on the `nimble-table-row` were happening one after another. Because of this the `cellState`, which includes the `cellRecord` and `columnConfig` could get out-of-sync from what the `cellViewTemplate` expected when the order of columns changes. For example, the `cellViewTemplate` for a date-text column could get passed a `cellRecord` and `columnConfig` for the number-text column that used to exist at a given column index.

To solve the problem, I decided the best approach was to have the row's template use a `repeat` directive over the `columns` with `recycle: false`. This means that if the set of columns changes, we won't attempt to reuse `nimble-table-cell` instances for columns that now may be changing types. In order to facilitate this change, I could no longer combine all state that cells need to know about into a single array of type `ColumnState`. Now, there are parallel arrays for `columns`, `cellIndentLevels`, and `cellStates`, which allows the row to distinguish between the cases when the whole cell should be recreated (i.e. the column has been changed) and when something about the cell should be updated (i.e. the indention, configuration, or value has been changed).

## 🧪 Testing

- Manually tested that swapping a number-text column and a date-text column in the DOM doesn't throw an exception anymore
- Unit tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-2c319a68-618f-4ea0-89c6-e3ef1bfbccd6.json
+++ b/change/@ni-nimble-components-2c319a68-618f-4ea0-89c6-e3ef1bfbccd6.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix bug in table when reordering columns",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table/components/row/index.ts
+++ b/packages/nimble-components/src/table/components/row/index.ts
@@ -59,6 +59,11 @@ export class TableRow<
     @observable
     public dataRecord?: TDataRecord;
 
+    /**
+     * @internal
+     * */
+    public columnNotifiers: Notifier[] = [];
+
     @observable
     public columns: TableColumn[] = [];
 
@@ -99,8 +104,6 @@ export class TableRow<
     // the selection checkbox 'checked' value should be ingored.
     // https://github.com/microsoft/fast/issues/5750
     private ignoreSelectionChangeEvents = false;
-
-    private columnNotifiers?: Notifier[] = [];
 
     @volatile
     public override get ariaSelected(): 'true' | 'false' | null {
@@ -215,7 +218,7 @@ export class TableRow<
     }
 
     private removeColumnObservers(): void {
-        this.columnNotifiers?.forEach(notifier => {
+        this.columnNotifiers.forEach(notifier => {
             notifier.unsubscribe(this);
         });
         this.columnNotifiers = [];
--- a/packages/nimble-components/src/table/components/row/index.ts
+++ b/packages/nimble-components/src/table/components/row/index.ts
@@ -22,7 +22,7 @@ import type {
 import type { TableColumn } from '../../../table-column/base';
 import type { MenuButtonToggleEventDetail } from '../../../menu-button/types';
 import { TableCell } from '../cell';
-import { ColumnInternals } from '../../../table-column/base/models/column-internals';
+import { ColumnInternals, isColumnInternalsProperty } from '../../../table-column/base/models/column-internals';
 
 declare global {
     interface HTMLElementTagNameMap {
@@ -68,11 +68,19 @@ export class TableRow<
     @attr({ attribute: 'menu-open', mode: 'boolean' })
     public menuOpen = false;
 
-    /** @internal */
+    /**
+     * @internal
+     * An array that parallels the `columns` array and contains the indent
+     * level of each column's cell.
+     * */
     @observable
     public cellIndentLevels: number[] = [];
 
-    /** @internal */
+    /**
+     * @internal
+     * An array that parallels the `columns` array and contains the cell state
+     * of each column's cell.
+     * */
     @observable
     public cellStates: (TableCellState | undefined)[] = [];
 
@@ -154,23 +162,11 @@ export class TableRow<
         if (
             source instanceof ColumnInternals
             && typeof args === 'string'
-            && (this.isColumnInternalsProperty(args, 'columnConfig')
-                || this.isColumnInternalsProperty(args, 'dataRecordFieldNames'))
+            && (isColumnInternalsProperty(args, 'columnConfig')
+                || isColumnInternalsProperty(args, 'dataRecordFieldNames'))
         ) {
             this.updateCellStates();
         }
-    }
-
-    private isColumnInternalsProperty(
-        changedProperty: string,
-        ...args: (keyof ColumnInternals<unknown>)[]
-    ): boolean {
-        for (const arg of args) {
-            if (changedProperty === arg) {
-                return true;
-            }
-        }
-        return false;
     }
 
     private emitActionMenuToggleEvent(
--- a/packages/nimble-components/src/table/components/row/index.ts
+++ b/packages/nimble-components/src/table/components/row/index.ts
@@ -1,4 +1,4 @@
-import { attr, observable, volatile } from '@microsoft/fast-element';
+import { Notifier, Observable, attr, observable, volatile } from '@microsoft/fast-element';
 import {
     Checkbox,
     DesignSystem,
@@ -16,17 +16,12 @@ import type {
 import type { TableColumn } from '../../../table-column/base';
 import type { MenuButtonToggleEventDetail } from '../../../menu-button/types';
 import { TableCell } from '../cell';
+import { ColumnInternals } from '../../../table-column/base/models/column-internals';
 
 declare global {
     interface HTMLElementTagNameMap {
         'nimble-table-row': TableRow;
     }
-}
-
-export interface ColumnState {
-    column: TableColumn;
-    cellState?: TableCellState;
-    cellIndentLevel: number;
 }
 
 /** Represents a single row (element) in the Table's data  */
@@ -69,6 +64,14 @@ export class TableRow<
 
     /** @internal */
     @observable
+    public cellIndentLevels: number[] = [];
+
+    /** @internal */
+    @observable
+    public cellStates: (TableCellState | undefined)[] = [];
+
+    /** @internal */
+    @observable
     public readonly selectionCheckbox?: Checkbox;
 
     /** @internal */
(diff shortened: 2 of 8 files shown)


packages/nimble-components/src/table/index.ts
```diff
@@ -687,7 +687,7 @@ export class Table<
                 column.columnInternals
             );
             notifierInternals.subscribe(this);
-            this.columnNotifiers.push(notifier);
+            this.columnNotifiers.push(notifierInternals);
```
mollykreis, line 690: Note: This isn't functionally required as part of this PR, but I found this mistake when I was writing the column notifiers for the row. Since it was a one-line fix and should be non-controversial, I decided to include it as part of this PR. However, I'd also be fine splitting it out into its own PR if others feel that would be more appropriate.

mollykreis:
@atmgrifter00, will you buddy this for me?

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -240,42 +254,89 @@ describe('TableRow', () => {
...
+            const cell1 = pageObject.getRenderedCell(1)!;
+            expect(cell1.cellViewTemplate).toEqual(
+                columnReferences.firstColumn.columnInternals.cellViewTemplate
+            );
+            expect(cell1.cellState?.columnConfig).toEqual(
+                columnReferences.firstColumn.columnInternals.columnConfig
+            );
+        });
+
+        it('updating column reuses cell', async () => {
```
mollykreis, line 318: I'm not a huge fan of these two tests that verify that we recycle cells in some cases vs creating cells in others. That behavior seems like more of an implementation detail, but I struggled to find a way to write a unit test that would catch the issue being fixed by this PR.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -165,6 +161,17 @@ export class TableRow<
         }
     }
 
+    public handleChange(source: unknown, args: unknown): void {
```
atmgrifter00, line 164: Mark `internal`?

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
atmgrifter00: I'm willing to defer to @rajsite or @jattasNI here. While I agree these tests seem a little too implementation-centric, I can see the potential usefulness of them.

atmgrifter00 (approved):

packages/nimble-components/src/table/components/row/index.ts
mollykreis: Yes, this is internal. We hadn't marked any of the internal functions as such in this file, so I went ahead and put that tag on all that it applies to.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -80,31 +100,7 @@ export class TableRow<
...
-                cellState = {
-                    cellRecord,
-                    columnConfig
-                };
-            }
-            const cellIndentLevel = i === 0 && this.nestingLevel > 0 ? this.nestingLevel - 1 : 0;
-            return { column, cellState, cellIndentLevel };
-        });
-    }
+    private columnNotifiers?: Notifier[] = [];
```
jattasNI, line 103: Does `columnNotifiers` need to be nullable? It looks like it always gets assigned a (possibly empty) array, not null or undefined.

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
jattasNI: No better ideas from me. Perhaps the real assert we want to make for the 'updating column reuses cell' test is actually a performance test that ensures the update rate is still fast. But we're not going to prioritize writing that now so I agree it's worth including these tests.

jattasNI (approved):

packages/nimble-components/src/table/components/row/index.ts
mollykreis: This is an interesting thing we've run into before -- the initial value of `[]` gets set in the constructor, along with all the other initial values. After each class member is initialized, its change handler is run before initializing the next class member & running its change handler. As a result, the `columns` change handler is called before `columnNotifiers` is initialized, which means that `removeColumnObservers` is called while `columnNotifiers` is still `undefined`. Therefore, I marked `columnNotifiers` with `?` since there are code paths that occur with `undefined` as its value.

One approach we've used in some places to avoid this problem is to look at the value of `this.$fastController.isConnected` in the change handler to not run the bulk of its implementation before the element is connected and then make sure everything is in the correct state in the `connectedCallback` since we hadn't been handling changes before then. Maybe that would be a better approach here as well. Any thoughts?

jattasNI: Ah, great explanation. It is definitely a mindset change to realize that something that appears to be initialized when it is defined is actually initialized after some of our code may have already run.

I'm not feeling a strong opinion either way. I don't think the broader type here is enough of a problem to justify much refactoring. But maybe if this keeps coming up we should form a policy around it and try to be consistent. For now I'm personally ok to leave it as-is.

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -240,42 +254,89 @@ describe('TableRow', () => {
...
+
+            const updatedCell = pageObject.getRenderedCell(0);
+            expect(originalCell).toBe(updatedCell);
+        });
+
+        it('reordering columns creates new cells', async () => {
+            const originalCell = pageObject.getRenderedCell(0);
+
+            // Swap the two columns
+            element.insertBefore(
```
rajsite, line 332: neat! and thanks for the comment!

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
rajsite: Yea I get the same kind of feeling from some of the virtualizer tests. But agree that it seems worth it to prevent accidental change in behavior.

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -200,13 +206,19 @@ describe('TableRow', () => {
...
         // prettier-ignore
-        async function setupTable(): Promise<Fixture<Table<SimpleTableRecord>>> {
+        async function setupTable(source: ColumnReferences): Promise<Fixture<Table<SimpleTableRecord>>> {
             return fixture<Table<SimpleTableRecord>>(
-                html`<nimble-table>
-                        <nimble-table-column-text id="first-column" field-name="stringData" column-id='foo'>Column 1</nimble-table-column-text>
-                        <nimble-table-column-text id="second-column" field-name="moreStringData" column-id='bar'>Column 2</nimble-table-column-text>
-                    </nimble-table>`
+                html`<${tableTag}>
+                        <${tableColumnTextTag} ${ref('firstColumn')} id="first-column" field-name="stringData" column-id='foo'>Column 1</${tableColumnTextTag}>
```
rajsite, line 218: nice cleanup! Can we remove the id attributes now?

packages/nimble-components/src/table/components/row/index.ts
rajsite: The other approach I've been pushing is to make it so the property initialization happens in the correct order to prevent change handler issues (and that may be a good idea here too).

That would mean making it public internal (to appease the lint rules) and moving `columnNotifiers` property initialization before `columns` property initialization. That way on columns property initialization and change handler execution `columnNotifiers` should be ready to go.

I kinda want to move away from the usage of isConnected except only for the case of avoiding unnecessary view rendering work until it is needed. Anything that can effect the state of the control should run even if not connected to the DOM. i.e. document.createElement and other imperative apis should behave well and properties on the element should report accurate info even if not attached to the DOM (same as native elements).

rajsite (approved):

change/@ni-nimble-components-2c319a68-618f-4ea0-89c6-e3ef1bfbccd6.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: not a very scientific test but was curious if all the extra notifiers could effect perf, but looks like on scrolling it might be a little better, maybe because we don't need to update all cellstate anymore

before (10-12 ms of work per frame)

image


after (7-9 ms of work per frame)

image

PR #1550, opened 2023-09-19
Rich text editor | Reverting back the workaround for the "tabindex" value update in footer buttons

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

We recognized that the workaround done in this PR https://github.com/ni/nimble/pull/1503/ creates an issue in iOS mobile browsers. In mobile/Tab iOS browsers, when the button is clicked, it does not automatically focus back on the editor, and the keypad closes.

Bug link: https://dev.azure.com/ni/DevCentral/_workitems/edit/2526025

With `pointer-events: none` to `slot="start"` of toggle button:

https://github.com/ni/nimble/assets/123377523/a0ec33d0-2040-4cc9-ac82-2e8c720f2bba

Without `pointer-events: none` to `slot="start"` of toggle button:

https://github.com/ni/nimble/assets/123377523/5c4b0e95-5ea4-4248-8559-7450bcdc2d04

After aligning with the leads, as a quicker resolution, we thought that this bug would create more impact for the mobile users than the other `tabindex` issue. 

## 👩‍💻 Implementation

Reverting all the changes that were implemented here: https://github.com/ni/nimble/pull/1503

## 🧪 Testing

Manually tested iOS devices(mobile and tab) in Safari and Chrome browsers.
    1. Opened [nimble.ni.dev](https://nimble.ni.dev/storybook/?path=/docs/incubating-rich-text-editor--docs) on an iOS device and clicked the buttons, the editor loses the focus.
    2. Opened the current [storybook publish](https://60e89457a987cf003efc0a5b-kqouhrfikv.chromatic.com/) on an iOS device and clicked the buttons, the editor got the focus as expected. (As shared in the above recordings)

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
+++ b/change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixes the focus back to the editor when the toolbar buttons are clicked in mobile iOS browsers",
+  "packageName": "@ni/nimble-components",
+  "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
+++ b/change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fixes the focus back to the editor when the toolbar buttons are clicked in mobile iOS browsers",
+  "comment": "Fixes keyboard closing when rich-text editor toolbar buttons are clicked in mobile iOS browsers",
   "packageName": "@ni/nimble-components",
   "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -250,16 +250,6 @@ export const styles = css`
         gap: ${standardPadding};
         place-items: center;
     }
-    ${
-        /**
-         * Restricting the pointer-events for the slot="start" where the icon is rendered.
-         * This can be removed after the below issue is fixed.
-         * https://github.com/ni/nimble/issues/1422
-         */ ''
-    }
-    nimble-toggle-button::part(start) {
-        pointer-events: none;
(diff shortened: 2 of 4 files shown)


vikisekarNI (approved):

vivinkrishna-ni:
We are making this PR ready for review ahead of a buddy review, as it involves the removal of the workaround code.

change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixes the focus back to the editor when the toolbar buttons are clicked in mobile iOS browsers",
```
m-akinc, line 3:
```suggestion
  "comment": "Fixes keyboard closing when rich-text editor toolbar buttons are clicked in mobile iOS browsers",
```

m-akinc (approved):
It's questionable to me that we would revert a general fix for the sake of the mobile experience. My understanding is that mobile support has been deprioritized, so we should not bother fixing those kinds of issues. However, the workaround being reverted does not seem super important (and we're still tracking the issue), so I'm not strongly opposed to this change.

change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: > It's questionable to me that we would revert a general fix for the sake of the mobile experience. My understanding is that mobile support has been deprioritized, so we should not bother fixing those kinds of issues. However, the workaround being reverted does not seem super important (and we're still tracking the issue), so I'm not strongly opposed to this change.

Further to this, the mobile bug doesn't seem too bad to me. I'm not terribly bothered if the keyboard is dismissed when I interact with something else on the page. But I also agree that it's your team's call which bug is more important.

By the way, I did test the Storybook builds on my iPhone both before and after this change and verified that it fixes the issue.

jattasNI (approved):

rajsite (approved):
revert looks fine deferring to rich text devs and nimble iphone users if it's the right direction

change/@ni-nimble-components-b0a59426-c9a2-444a-a9f7-1ccf79dd94b4.json
vivinkrishna-ni: Thanks for verifying the behavior. For the other issue https://github.com/ni/nimble/issues/1422, we will wait until the underlying issue in FAST gets fixed, if not we will find some other way to fix the tabindex behavior without affecting the mobile experience.

PR #1552, opened 2023-09-19
Hide icons from accessibility tree by default

### LINKED ISSUE #1510: setting aria-label on nimble icon causes accessibility issue

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1510

## 👩‍💻 Implementation

- icons set `aria-hidden` on their shadow DOM to exclude it from the accessiblity tree
- clients are now expected to set `role` (along with `aria-label`) on the icon if they wish to make it accessible
- icons no longer forward `aria-label` to the inner `svg`. 
- updated icon table column to set `role="img"` on the rendered icons (in addition to `aria-label`)
- updated table column grouping icon to set `role="img"`
- updated banner severity icons to set `role` and `aria-label`
- updated tests
- updated icon accessibility guidance in Storybook
- updated accessibility.md with developer guidance

## 🧪 Testing

Added/updated tests

## ✅ Checklist


m-akinc:
@mollykreis could you please buddy this change?

change/@ni-nimble-components-c29f703d-f01e-45c4-942f-c887470025b8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
mollykreis, line 2: > 🧪 Testing
Added/updated tests

You should also do some manual testing with these changes to make sure this approach doesn't have accessibility violations associated with it. Specifically, when I go to the "Icon Column" story in storybook, I see new accessibility violations associated with the icons & their underlying `svg` elements.

mollykreis (changes requested):

change/@ni-nimble-components-c29f703d-f01e-45c4-942f-c887470025b8.json
m-akinc: I had tested before explicitly adding `role="img"` to the inner `svg`s, and there were no issues. Interestingly, even though the `svg` element seems to have a default role of `img`, it is only when we explicitly set `role="img"` do we get violations. This is kind of a can of worms, because now every icon in our components needs to set `alt` text to avoid these violations. Looks like I'm going to have to add some new tokens for our label providers.

mollykreis: Should the expectation be that every icon has alt text no matter where it is used? For example, do we want a button that says "+  Add column" to bubble up in the accessibility tree alt text for the "+" icon within the button? Or, do you think this accessibility violation is an indication that marking our icons as an `img` is the wrong direction?

m-akinc: Did some more research. I think it is incorrect to always set the role to `img`. In your "+ Add column" example, or for the down arrow of a nimble-select (which is not a separate button), I think the correct role is `presentation`/`none`, since the image is purely decorational. It also appears that there is a more specific/appropriate role to use when the icon has meaning: `graphics-symbol`. This role was introduced in ARIA 1.1, so we can also specify `img` as a fallback role. We want to sometimes set the role to `presentation` and sometimes `graphics-symbol`. We would rather not leave it up to the client to set `role` and `aria-label` on the host `nimble-icon-*` element, so I suggest we use the presence of an `alt` text value to determine which role to set on the `svg`.

Note also that Storybook's Accessibility tab was giving us false negatives before I made the change in this PR. The default role of `svg` is `graphics-document` (even though Chrome shows it as `img`) which also requires an accessible name.

Sources:
https://www.w3.org/TR/graphics-aria-1.0/#graphics-symbol
https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs
https://www.w3.org/WAI/ARIA/apg/practices/hiding-semantics/

mollykreis: It doesn't look like your implementation/proposal follows the best practices included in the [accessible-svgs](https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs) page you linked. In the `Hiding SVGs` section, it says you shouldn't set the role to `presentation` -- instead, you should set `aria-hidden`. Is there a reason you're doing something different here?

m-akinc: Good catch. I had missed that detail.

m-akinc: Apparently, Chromatic's Accessibility checks don't like fallback roles, so now I'm just using `graphics-symbol` without a fallback role of `img`.

mollykreis: A few thoughts:
1. Do we have any idea of how widely supported ARIA 1.1 is? That's when you said `graphics-symbol` was added, right? If we can't do the `role` fallback, we should make sure we're picking a `role` that's well supported.
2. Perhaps we should have a larger discussion on the direction this has taken before getting into the details of the implementation in this PR. I'm not aware of any other component where we change its `role` based on a seemingly unrelated attribute, so I think it would be good for us to make sure we're happy with that as a new pattern here.

packages/nimble-components/src/icon-base/index.ts
```diff
@@ -25,27 +25,6 @@ export class Icon extends FoundationElement {
...
-        }
-        if (this.ariaLabel !== null && this.ariaLabel !== undefined) {
-            svg.setAttribute('aria-label', this.ariaLabel);
-        } else {
-            svg.removeAttribute('aria-label');
-        }
-    }
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
```
mollykreis, line 30: Do we still need to have `Icon` extend and use the `ARIAGlobalStatesAndProperties` mixin?

packages/nimble-components/src/table-column/enum-base/models/mapping-icon-config.ts
```diff
@@ -9,6 +9,7 @@ export interface IconView {
 const createIconTemplate = (icon: string): ViewTemplate<IconView> => html`
     <${icon}
         title="${x => x.text}"
+        role="graphics-symbol"
```
mollykreis, line 12: Can you also add a role to the grouped icon in the table header? https://github.com/ni/nimble/blob/ab995f09749e33f9fdcb5a5975a8c1d84d8aca00/packages/nimble-components/src/table/components/header/template.ts#L40

mollykreis (approved):

packages/nimble-components/src/table-column/enum-base/models/mapping-icon-config.ts
m-akinc: Any idea why we don't also make the ascending/descending sort indicators accessible?

packages/nimble-components/src/icon-base/index.ts
m-akinc: I don't think so.

packages/nimble-components/src/icon-base/template.ts
```diff
@@ -3,6 +3,6 @@ import type { Icon } from '.';
 
 export const template = html<Icon>`
     <template>
```
rajsite, line 5: can remove the wrapping template if it has no bindings on it.
think we should also give it the no whitespace in the template treatment and a comment stating that because it is used inline.

rajsite (approved):

change/@ni-nimble-components-c29f703d-f01e-45c4-942f-c887470025b8.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
m-akinc: > Do we need to update any of those?

Mostly, no, because their role or other ARIA attributes (e.g. `aria-expanded`, `aria-checked`) already communicate the meaning of the icons. I was on the fence but decided to add `role` and `aria-label` to the banner severity glyphs.

> Can we document this guidance in CONTRIBUTING.md or accessibility.md?
Added to accessibility.md.

packages/nimble-components/src/banner/template.ts
```diff
@@ -35,13 +35,13 @@ export const template = html<Banner>`
     >
         <div class="icon">
             ${when(x => x.severity === BannerSeverity.error, html`
-                <${iconExclamationMarkTag}></${iconExclamationMarkTag}>
+                <${iconExclamationMarkTag} role="img" aria-label="Error"></${iconExclamationMarkTag}>
```
jattasNI, line 38: I think we should offer apps a way of localizing these labels by setting them via the locale provider instead of hard-coded strings.

packages/nimble-components/src/icon-base/tests/icons.mdx
```diff
@@ -21,4 +21,4 @@ import * as iconsStories from './icons.stories';
 
 ## Accessibility
 
-When using an icon standalone, the `title` attribute should usually be set, to provide accessible text (which will also show as a tooltip).
+When using an icon standalone, and it has a meaning that is not otherwise communicated by adjacent text, set the `role` attribute to `img` and the `aria-label` attribute to a descriptive label that should be used as the accessible name. When slotting an icon into another component, if that component's accessible name fully describes its semantics, it is not neccessary to give the icon its own role and accessible name. For example, when slotting a "+" icon into a button labeled "Add File", the button's label serves as the accessible name and fully describes the button's semantics. In that case, the "+" icon is purely decorative. However, if the button's label was just "File", it would be neccessary to set the icon's role and its `aria-label` to "Add".
```
jattasNI, line 24: How come we're getting rid of the guidance to set `title`? If the icon is being used completely standalone (i.e. not slotted into a button or other input), wouldn't we still want a way for mouse users to learn its meaning?

Perhaps this is a moot point since we'd almost never recommend placing an icon in the middle of nowhere with no other context. The only use case I can think of is an icon table column, but that already has guidance to set user-visible text for tooltips and group labels via its custom API.

How about something like this?

```suggestion
An icon's image should never be the sole way for a user to understand its meaning. Always provide additional context via adjacent visible text or via a `title` for sighted mouse users and a `role` and `aria-label` for users of accessibility technology. Set the `role` attribute to `img` and the `aria-label` and `title` attributes to a descriptive label.

For example, when slotting a "+" icon into a button labeled "Add File", the button's label serves as the accessible name and fully describes the button's semantics. In that case, the "+" icon is purely decorative. However, if the button's label was just "File", it would be neccessary to set the icon's role and its `aria-label` to "Add".
```

m-akinc: I had removed the guidance about `title` because if we are telling them to use `aria-label`, it is no longer an accessibility concern to set `title`. It may still be a good idea to set title (for sighted users), but I didn't think we needed to call that out in our accessiblity section.

I'm fine with your version, though.

packages/nimble-components/src/banner/template.ts
m-akinc: Good point. Added.

jattasNI (approved):

Issue #1117, opened 2023-03-14
Allow app to configure nimble-dialog size CSS attributes


jattasNI:
## 😯 Problem to Solve

As discussed in #1102, the nimble-dialog currently has a fixed size and applications can only change it by setting internal styles. This prevents apps from using it for use cases like a full-screen wizard.

## 💁 Proposed Solution

We don't have clear guidance about pre-set sizes we want to support yet, so we'd like to allow apps to set an arbitrary size to begin with. This would be via normal CSS attributes like `width` and `height`. This is similar to the approach taken by the spinner.

## 📋 Tasks

- [ ] Update dialog `styles.ts` to respect `width` and `height` set by clients
- [ ] Figure out what to do about other size attributes like `min-width`, `max-height`
- [ ] Add design tokens for the known pre-set sizes (the current default size plus the sizes in the PR linked above)
- [ ] Add storybook documentation that shows how to set the size
- [ ] Add a storybook matrix test for each pre-set size

jattasNI:
This came up again in an [SLE HLD about a modal over the software grid](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/470673?_a=files&path=/docs/design-documents/Systems-Management/Install-Software-UI.md). That team is hoping to contribute a fix.

jattasNI:
This came up again in a request from an NI partner developing SystemLink plugins using Nimble. Their use case was for a wizard-style dialog. I believe the current proposal would meet their needs.

kroeschl:
I would also like to see this feature (or at least a workaround). My use case is a simple dialog with a single text area and a couple of controls where users sometimes want to enter large amounts of text. In that case, I'd like to allow them to resize the text area and the containing dialog by setting
```css
width: revert;
min-width: 400px;
max-height: revert;
```

PR #1553, opened 2023-09-20
 Expose width and height of nimble-dialog & add tokens

stefanc18:
# Pull Request

## 🤨 Rationale

#1117

## 👩‍💻 Implementation

- Performed the tasks that are enumerated in the issue linked above.
- Create 2 tokens for the existing (small) dialog: width & max-height
- Create 2 tokens for the large dialog: width & height
- Have the small dialog as the default (width & max-height of small dialog; height set to `fit-content`, this was the default height that was applied to the dialog until now)
- Added storybook documentation.

## 🧪 Testing

Manually thorugh the storybook tests & matrix tests in chromatic. The sizes are configurable in the storybook similarly to how the width is configurable for the nimble drawer. For the matrix tests, I added 2 stories: one for the small dialog and one for the large dialog: https://www.chromatic.com/build?appId=60e89457a987cf003efc0a5b&number=7776

## ✅ Checklist


packages/nimble-components/src/dialog/styles.ts
```diff
@@ -25,14 +27,21 @@ import { themeBehavior } from '../utilities/style/theme';
...
         flex-direction: column;
         background-color: ${applicationBackgroundColor};
         border: none;
         box-shadow: ${elevation3BoxShadow};
         padding: 0px;
-        width: 400px;
-        max-height: 600px;
+        width: inherit;
+        height: inherit;
+        max-height: inherit;
```
rajsite, line 44: Nothing size related should be relying on inherit. See the discussion in this thread and the related threads it links to:
https://github.com/ni/nimble/pull/1446#discussion_r1309430889

packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -306,6 +306,18 @@ export const iconSize = DesignToken.create<string>(
...
+).withDefault('400px');
+export const dialogSmallMaxHeight = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.dialogSmallMaxHeight)
+).withDefault('600px');
+export const dialogLargeWidth = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.dialogLargeWidth)
+).withDefault('1024px');
+export const dialogLargeHeight = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.dialogLargeHeight)
+).withDefault('680px');
```
rajsite, line 320: Where are these values coming from?

packages/nimble-components/src/dialog/styles.ts
stefanc18: In the second link that you provided, I see that the approach was to set the width to 100% and then configure it from the parent component, which would be like a something like a div wrapper. But this approach doesn't seem to work in this case. I've tried setting width, height, max-height of the dialog to 100% and then set them from the host component, but that did not work.

The only other way that I see working is having 2 new tokens, like $ni-nimble-internal-dialog-width, that clients would override, instead of setting the width property on the nimble-dialog. This way, the width & height wouldn't be "inherit", but that token. Does this approach feel suitable to you? If not, could you please provide another suggestion?

packages/nimble-components/src/theme-provider/design-tokens.ts
stefanc18: - The width and the max height for the small ones are exactly the same ones as those that are [hardcoded now](https://github.com/ni/nimble/blob/43176dfe4924fd4a40af5d014fee7961b0022ad9/packages/nimble-components/src/dialog/styles.ts#L34C1-L35C27). I followed @jattasNI 's indication of creating tokens for this one, as he mentioned in the issue that I linked.
- The width and height for the large component have been decided by Leslie for the software install workflow in SLE. We followed up with our PO, TJ, and the decision was to have these values as a standard for all large dialogs

packages/nimble-components/src/dialog/styles.ts
jattasNI: I'm guessing that setting the width of the `dialog` to `100%` causes it to fill the entire screen width rather than have the same width as the `nimble-dialog`. This is probably because the modal `dialog` has `position: fixed` which means its [containing block is the entire viewport](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block) rather than its parent DOM element. There's a bit more context [in this SO post](https://stackoverflow.com/questions/73925255/how-can-the-html-dialog-element-be-positioned-according-to-normal-flow-when-op) but I don't like their suggestion to put the `dialog` in a `div` just for positioning.

The reason `inherit` works is because [inherit is always relative to the parent element, not the containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit). But that doesn't solve the risks that Milan mentioned.

I don't love the suggestion to ask clients to set the values of Nimble token. While it's technically possible, we usually try to discourage clients changing the token value; it's easy to mess it up and impact other components.

I haven't been able to think of other ideas so at the moment my vote would be to use `inherit` in the `dialog` but be explicit about setting each of the corresponding attributes on the `:host` (as you're doing), which would mean (I believe) a client would have to actively unset them to cause problems. But it'll be good to see if Milan has other ideas (he was OOO on Friday so let's discuss more next week).

rajsite: I didn't get a chance to explore this more thoroughly today but may be able to tomorrow.

One idea that could be prototyped is having the `nimble-dialog` itself handle it's positioning and override the styles that the `dialog` is doing. The dialog has preset styling but it's nothing special and we partially override it already. I think we could make the `nimble-dialog` the sized and positioned item and then not have to deal with the user trying to size something hidden in the shadow root.

The next not great idea would be to not hide the thing in a shadow root, i.e. expose the internal dialog as a part that could be styled directly. But that seems unnecessary. Seems like the user should just size the `nimble-dialog` and have it be the size visualized.

stefanc18: @rajsite any news on the preferred approach? I don't understand how the first idea would be implemented without having the `dialog` inherit the styles of the parent, which is the `nimble-dialog`. Or do you prefer the second approach that you suggested?

rajsite: The first idea works for normal elements. The host element is sized by users and a child element in the shadow root just says make my size width:100%, height:100% to fill the container. Other options are to use display: contents to say layout is delegated to the child.

I prototyped it a bit but ran into issues as the dialog in the shadow root pops out of the context and moves to the top layer. The spec says that if the dialog element is not position fixed the containing block is the [initial containing block](https://drafts.csswg.org/css-position-4/#top-styling) which I believe means the direct parent (nimble-dialog) should control the sizing for something like width:100%, height:100% but that doesn't seem to be the case. 

Regardless, my option 1 suggestion doesn't make much sense because the internal dialog is what moves into the top-layer, not the nimble-dialog wrapper. So we wouldn't want to do all the fixed position styling, etc, on something not in the top layer.


So I think the second approach is what makes sense. We expose the internal dialog as a shadow dom part that can be styled. We should also do something to make the `nimble-dialog` less of a target. Something like set it to visibility:hidden and make it size 0x0, etc. The open questions are the API name choices, ie what to call the shadow part we expose for the dialog. Haven't thought of what to call it but exposing the dialog as a shadow dom part does seem like the probable direction to go.

stefanc18: @rajsite Thank you for the detailed explanation. I've updated the implementation to expose the shadow as a part of the nimble-dialog. It can be now styled like this:
image

Thus, the need of inheriting the width and height is gone. I hope I understood correctly about the desired approach, let me know if this needs any other modifications.

jattasNI:
@stefanc18 based on your recent updates it looks like this is ready for reviewers to take another look. In future you can signal this by re-requesting review (I did this already in this case but let us know if you're still working on it).

image

packages/nimble-components/src/dialog/tests/dialog.stories.ts
```diff
@@ -16,6 +26,9 @@ interface DialogArgs {
     includeFooterButtons: boolean;
     preventDismiss: boolean;
     content: ExampleContentType;
+    width: DialogSizeOptions;
+    height: DialogSizeOptions;
+    maxHeight: DialogSizeOptions;
```
rajsite, line 31: I think showing this as three separate properties is misleading. You wouldn't want to mix small and large tokens. Instead you would use all small or all large tokens and then either height or max-height.

Can you follow a pattern more like Spinner `size` entry in storybook?
I think the states are small fixed, small growable, large fixed, and large growable.
And the usage details should cover targeting the `part(dialog)` CSS Part of the nimble-dialog

packages/nimble-components/src/dialog/styles.ts
```diff
@@ -12,7 +12,9 @@ import {
     smallPadding,
     subtitleFont,
     subtitleFontColor,
-    elevation3BoxShadow
+    elevation3BoxShadow,
+    dialogSmallWidth,
+    dialogSmallMaxHeight
```
rajsite, line 17: MaxHeight is a new token suffix and should be added to the list: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/theme-provider/design-token-names.ts#L243

Note the comment at the top, it should come before Height in the list.

packages/nimble-components/src/dialog/template.ts
```diff
@@ -6,6 +6,7 @@ export const template = html<Dialog>`
         <dialog
             ${ref('dialogElement')}
             role="dialog"
+            part="dialog"
```
rajsite, line 9: Let's name it `control` that's how [FAST named the part of their dialog with the actual role dialog](https://github.com/microsoft/fast/blob/507d2fef836f75b072d0c05c9e37b7c59959ff51/packages/web-components/fast-foundation/src/dialog/dialog.template.ts#L30) so we can align with that.

packages/nimble-components/src/dialog/styles.ts
rajsite: FYI, that will result in you needing to add an entry here as well: https://github.com/ni/nimble/blob/73b899d1d36e3c04219e36b989a7367bd3800a87/packages/nimble-components/src/theme-provider/tests/tokens.stories.ts#L100

stefanc18: added

packages/nimble-components/src/dialog/tests/dialog.stories.ts
stefanc18: I added as suggested, but only "small growable" and "large fixed". I don't think the others make sense:
- small fixed: we don't have a small fixed height, when the dialog is small, it is set to "fit-content"
- large growable: we could set the width to be large and the max height to be the large one, and the height to "fit-content".. but then we'll just have a very wide dialog with a small height, that would only grow with the content. That doesn't look like a desired use case..

Let me know your thoughts

jattasNI: Adding some more guidance about how the client specifies the size would be helpful. Right now they can toggle between the options and reverse engineer the setting by looking at the "Show code" view but we can be more explicit.

A good pattern to follow is what the table and spinner docs do where they have a dedicated section called Usage Docs >> Sizing ([example](https://nimble.ni.dev/storybook/?path=/docs/components-spinner--docs#sizing)). To do this you'll need to turn off `autodocs` in the story and add a `.mdx` file that controls the structure of the document ([example](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/spinner/tests/spinner.mdx)). In the sizing section you'd describe which design tokens we recommend and how to configure the dialog to use them.

stefanc18: Sounds good, thanks for providing the examples! I've added usage docs similarly to the spinner. This is what they look like:

image

Let me know if it needs other modifications

packages/nimble-components/src/dialog/tests/dialog-matrix.stories.ts
```diff
@@ -28,6 +44,33 @@ const component = html`
     </${dialogTag}>
 `;
 
+const dialogSizingTestCase = (
+    [widthLabel, widthStyle]: [string, string],
+    [heightLabel, heightStyle]: [string, string],
+    [maxHeightLabel, maxHeightStyle]: [string, string]
+): ViewTemplate => html`
+    <p style="font: var(${cssPropertyFromTokenName(
```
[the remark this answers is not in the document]
stefanc18: Added, there's a space now underneath the dialog: https://www.chromatic.com/build?appId=60e89457a987cf003efc0a5b&number=8190

packages/nimble-components/src/dialog/tests/dialog.mdx
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+
+For larger dialogs, for example a wizard-like dialog, the following style configuration should be used:
+
+```scss
+nimble-dialog::part(control) {
+    width: $ni-nimble-dialog-large-width;
+    height: $ni-nimble-dialog-large-height;
+    max-height: $ni-nimble-dialog-large-height;
```
rajsite, line 44: We should encapsulate these specifics in the design system. For example that height and max-height are the same value or that height is the value 'fit-content'.

So I'd propose we export the following tokens:
```
$ni-nimble-dialog-small-width
$ni-nimble-dialog-small-height
$ni-nimble-dialog-small-max-height

$ni-nimble-dialog-large-width
$ni-nimble-dialog-large-height
$ni-nimble-dialog-large-max-height
```

packages/nimble-components/src/dialog/tests/dialog.stories.ts
```diff
@@ -154,6 +192,20 @@ const metadata: Meta<DialogArgs> = {
...
             }
         },
+        size: {
+            description: sizeDescription,
+            options: [
+                DialogSizeOptions.smallGrowable,
+                DialogSizeOptions.largeFixed
+            ],
+            control: {
+                type: 'select',
```
rajsite, line 202: can you make this a radio instead? makes it easier to see the available options and the list is small so we can show them all

packages/nimble-components/src/dialog/tests/dialog.stories.ts
```diff
@@ -49,9 +56,33 @@ const content = {
     [ExampleContentType.longContent]: longContent
 } as const;
 
+const sizeDescription = `
+Size of a nimble dialog.
+
+See the Sizing section below for information on controlling the size of the dialog.
```
rajsite, line 62: When on the canvas page you can't see the usage docs so this could be confusing

```suggestion
See the Sizing section of the Usage Docs for information on controlling the size of the dialog.
```

packages/nimble-components/src/dialog/tests/dialog.stories.ts
```diff
@@ -49,9 +56,33 @@ const content = {
...
+const widths = {
+    [DialogSizeOptions.smallGrowable]: dialogSmallWidth.getValueFor(
+        document.body
+    ),
+    [DialogSizeOptions.largeFixed]: dialogLargeWidth.getValueFor(document.body)
+} as const;
+
+const heights = {
+    [DialogSizeOptions.smallGrowable]: 'fit-content',
+    [DialogSizeOptions.largeFixed]: dialogLargeHeight.getValueFor(document.body)
```
rajsite, line 74: We shouldn't get the calculated value for all these, we should just leverage the design token css custom property. Browser should do the work for calculating it. i.e. `var(${dialogLargeHeight.cssCustomProperty})`

packages/nimble-components/src/dialog/tests/dialog-matrix.stories.ts
```diff
@@ -60,3 +103,48 @@ export const dialogDarkThemeBlackBackground: StoryFn = createFixedThemeStory(
...
+`);
+
+smallDialogSizing.play = playFunction;
+
+export const largeDialogSizing: StoryFn = createStory(html`
+    ${createMatrix(dialogSizingTestCase, [
+        [
+            [
+                `Width ${dialogLargeWidth.getValueFor(document.body)}`,
+                `width: ${dialogLargeWidth.getValueFor(document.body)}`
```
rajsite, line 132: Same as below, for all these use the css property name and don't worry about calculating the value

```suggestion
                `Width Large Dialog`,
                `width: var(${dialogLargeWidth.cssCustomProperty})`
```

packages/nimble-components/src/dialog/tests/dialog-matrix.stories.ts
```diff
@@ -28,6 +44,33 @@ const component = html`
...
+    <p style="font: var(${cssPropertyFromTokenName(
+        tokenNames.bodyFont
+    )}); padding-bottom: 1000px;">${() => widthLabel}; ${() => heightLabel} ${() => maxHeightLabel}</p>
+    <style>
+        ${dialogTag}::part(control) {
+            ${() => widthStyle};
+            ${() => heightStyle};
+            ${() => maxHeightStyle};
+        }
+    </style>
```
rajsite, line 61: can you add a class name to that `p` tag like `spacer` and then move the styles into the style tag, also note the pattern used for the css custom property:

```
.spacer {
   padding-bottom: 1000px;
}
```

rajsite (commented):
Thanks for following up on the feedback, it's getting close!

packages/nimble-components/src/dialog/tests/dialog.stories.ts
stefanc18: changed 
image

packages/nimble-components/src/dialog/tests/dialog-matrix.stories.ts
```diff
@@ -60,3 +105,53 @@ export const dialogDarkThemeBlackBackground: StoryFn = createFixedThemeStory(
...
+            ]
+        ],
+        [
+            [
+                'Max Height Small Dialog',
+                `width: var(${dialogSmallMaxHeight.cssCustomProperty})`
+            ]
+        ]
+    ])}
+`);
```
rajsite, line 130: These stories aren't correct, notice they are all setting the `width:` css prop. But also it's a little odd to use the matrix helpers to render a single case.

I'd instead recommend using the createFixedThemeStory helper following a pattern like menu-button stories: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/menu-button/tests/menu-button-opened-matrix.stories.ts

stefanc18: Apologies for this one.. copy paste error :D

It should be okay now, I updated to use `createFixedThemeStory` as you suggested

rajsite (approved):

PR #1554, opened 2023-09-20
Icon table with metadata

rajsite:
# Pull Request

## 🤨 Rationale

Create a table for icons in storybook that shows the tag names and metaphors.

## 👩‍💻 Implementation

- Rendered the icons using the nimble-table
- Show metaphors in a separate column that is not sortable
- Set the height of the table large enough so that all the rows render in the DOM. That way the page is searchable with normal ctrl+f search and there are no nested scrollbars. Can also use normal browser zoom to see the icons bigger.

## 🧪 Testing

Manually checked in storybook

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/icon-base/tests/icons.stories.ts
+++ b/packages/nimble-components/src/icon-base/tests/icons.stories.ts
@@ -18,9 +18,9 @@ import { tableColumnTextTag } from '../../table-column/text';
 import { iconMetadata } from './icon-metadata';
 
 type IconName = keyof typeof nimbleIconComponentsMap;
-const data = Object.values(nimbleIconComponentsMap).map(klass => ({
-    tag: DesignSystem.tagFor(klass),
-    metaphor: iconMetadata[klass.name as IconName].tags.join(', ')
+const data = Object.values(nimbleIconComponentsMap).map(iconClass => ({
+    tag: DesignSystem.tagFor(iconClass),
+    metaphor: iconMetadata[iconClass.name as IconName].tags.join(', ')
 }));
 
 type Data = (typeof data)[number];
--- a/packages/nimble-components/src/icon-base/tests/icons.stories.ts
+++ b/packages/nimble-components/src/icon-base/tests/icons.stories.ts
@@ -39,10 +39,10 @@ const metadata: Meta<IconArgs> = {
                 component: `Nimble icons can be slotted into other components or used independently. Each icon is available as a custom element. For example, \`<${iconAddTag}></${iconAddTag}>\``
             },
             source: {
-                code: null,
+                code: null
             }
         }
-    },
+    }
 };
 
 export default metadata;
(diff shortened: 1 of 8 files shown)


packages/nimble-components/src/icon-base/tests/icons.stories.ts
```diff
@@ -1,22 +1,33 @@
...
 } from '../../theme-provider/design-token-names';
+import { Table, tableTag } from '../../table';
+import { tableColumnIconTag } from '../../table-column/icon';
+import { mappingIconTag } from '../../mapping/icon';
+import { tableColumnTextTag } from '../../table-column/text';
+import { iconMetadata } from './icon-metadata';
 
-const nimbleIconComponents = Object.values(nimbleIconComponentsMap);
+type IconName = keyof typeof nimbleIconComponentsMap;
+const data = Object.values(nimbleIconComponentsMap).map(klass => ({
```
fredvisser, line 21: Just curious `klass` is an unreserved keyword version of `class`? Or something else?

fredvisser (approved):

packages/nimble-components/src/icon-base/tests/icons.stories.ts
rajsite: yep 😅 something deep inside my finger memory, but I wouldn't let that through review these days so good catch, Updated to a normal name of iconClass

packages/nimble-components/src/icon-base/tests/icons.stories.ts
```diff
@@ -1,22 +1,33 @@
...
+}));
+
+type Data = (typeof data)[number];
 
 interface IconArgs {
     severity: keyof typeof IconSeverity;
+    tableRef: Table<Data>;
 }
 
 const metadata: Meta<IconArgs> = {
```
jattasNI, line 33: I was finding it slightly jarring to see a huge table with no context. How about a bit of docs above it like this?
```suggestion
const metadata: Meta<IconArgs> = {
    description: 'Nimble icons can be slotted into other components or used independently. Each icon is available as a custom element. For example, `<nimble-icon-add></nimble-icon-add>`'.
```

jattasNI (approved):

packages/nimble-components/src/icon-base/tests/icons.stories.ts
```diff
@@ -52,26 +47,57 @@ Set severity on the element to switch between the theme-aware color options.
...
     render: createUserSelectedThemeStory(html`
-        <style class="code-hide">
-            .container > * {
-                padding: 5px;
-            }
-        </style>
-        <div class="container">
-            ${repeat(() => nimbleIconComponents, iconTemplate)}
-        </div>
+        <${tableTag}
```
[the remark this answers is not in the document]
rajsite: Disabled show code for icons and the tokens stories via `{ parameters: { docs: { source: { code: null } } } }` similar to the matrix stories

Issue #1527, opened 2023-09-13
Rich text editor | Support for hyperlinks along with absolute links


vivinkrishna-ni:
## 📌 User Story
**Context** - We are envisioning that after adding hyper link support, only pasted or typed absolute URLs might use `<autolink>` syntax and rest of them will likely use normal links syntax `[link text](url)`. However this is subjected to discussion on rich text components spec to decide the intended behavior after research. 

**Scope of this user story** - As part of adding hyperlink support, we may need to revisit some restrictions on supporting only absolute links and consider removing/replacing them. This user story will capture those tasks for future reference. 

Hyperlink (termed as [Links in CommonMark Spec](https://spec.commonmark.org/0.30/#links)) support for the `nimble-rich-text-editor` component will be added along with current absolute link (termed as [Autolinks in CommonMark Spec](https://spec.commonmark.org/0.30/#autolinks)) support

- Hyperlink markdown syntax - `[link text](url)`
- Absolute link markdown syntax - `<autolink>`

_Note that any detailed implementation plan for hyperlink support will be updated in spec HLD for rich text components._ 

### 😮 Current Support

We support only **Absolute links** for the initial pass which is by typing or pasting a full URL into the editor and not adding a link to the existing texts.

We also made the following changes to support only absolute links in the editor and autolink in the markdown which should be revisited when we support normal hyper links with text:
1. [ ] Added `pointer-event: none;` to all the links in the editor which will restrict the user from opening the link from the editor in any form.
2. [ ] In `markdown-serializer.ts`, [prosemirror-markdown](https://github.com/ProseMirror/prosemirror-markdown/blob/c7210d0e55c82bfb0b2f7cba5dffe804575fafb3/src/to_markdown.ts#L3C1-L26C2) implements logic for both hyperlink and autolink in `defaultMarkdownSerializer`. Therefore, implemented our own implementation to support only autolink in markdown output.
3. [ ] In `markdown-parser.ts`, bypass the default [normalization link text](https://github.com/markdown-it/markdown-it/blob/2b6cac25823af011ff3bc7628bc9b06e483c5a08/lib/index.js#L67C1-L86C2) provided by markdown-it to display the link as is in the editor and viewer. This needs to be removed.
[...]

PR #1559, opened 2023-09-21
Rich text editor | Support for pasting absolute links when copied from other sources

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

When links are copied from other sources and pasted into the editor, they initially render as plain text. However, they will be transformed into links by adding whitespace or by pressing the Enter key.

Additionally, when hyperlinks (a link behind different text content) are copied and pasted into the editor, only the text content of the link (not the actual URL) is pasted by default or using Ctrl + V. When pasting as plain text or using Ctrl + Shift + V, it still pastes the same text content.

Fixes the following bugs:
1. [Bug 2525823](https://dev.azure.com/ni/DevCentral/_workitems/edit/2525823): Rich text editor | Copying a link from Edge browser which is not pasting as a link
3. [Bug 2525827](https://dev.azure.com/ni/DevCentral/_workitems/edit/2525827): Rich text editor | Absolute links requires space or enter after pasting the link

## 👩‍💻 Implementation

1. By utilizing [`transformPasted`](https://prosemirror.net/docs/ref/#view.EditorProps.transformPasted), if the pasted content includes a link mark, the text content of that node is replaced with the `href` value of the link mark. However, if the URL within the `href` attribute doesn't start with `https` or `http` protocols, then the text content is pasted as plain text within the same node.
2. In addition to the `nimble-anchor` tag in `parseHTML`, added the `<a>` tag. This is necessary because when copying a link from an external source, the link will typically be enclosed within an `<a>` tag, which helps the editor identify it as a link and render it within an `<a>` tag.

## 🧪 Testing

1. To simulate the pasting of links into the editor, include the `pasteHTMLToEditor` function in the page objects.
2. I have created unit tests to verify that when a valid anchor element is pasted, it is displayed as a link. Additionally, when a hyperlink from an external source is pasted, it remains an absolute link.
4. I have unit tests to check that when invalid links are pasted, they should be rendered as plain text.

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +335,35 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
         const customLink = this.getCustomLinkExtension();
+        const validLinkRegex = /^https?:\/\//i;
+
+        /**
+         * @param htmlString contains the html string of the copied content. If the content is a link, the `htmlString` contains anchor tag and a href value.
+         * ProseMirror reference for `transformPastedHTML`: https://prosemirror.net/docs/ref/#view.EditorProps.transformPastedHTML
+         */
+        const transformPastedHTML = function (htmlString: string): string {
+            const tempElement = document.createElement('div');
+            tempElement.innerHTML = htmlString;
```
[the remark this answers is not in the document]
vikisekarNI: On second thought, is it handled internally and rendering it just as plain text as we switched off HTML? I know this happens while parsing markdown but just wanted to make sure that it also happens after this function hands over the string back to Prosemirror/Tiptap?

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -393,10 +427,17 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
             // See: https://github.com/ni/nimble/issues/1527
             inclusive: false,
             parseHTML() {
-                // To load the `nimble-anchor` from the HTML parsed content by markdown-parser as links in the
-                // Tiptap editor, the `parseHTML` of Link extension should return `anchorTag`. This is because the
-                // link mark schema in `markdown-parser.ts` file uses `<nimble-anchor>` as anchor tag and not `<a>`.
-                return [{ tag: anchorTag }];
+                // To load the `nimble-anchor` from the HTML parsed content by markdown-parser as links in the Tiptap editor, the `parseHTML`
+                // of Link extension should return nimble `anchorTag` in addition to `<a>` tag as it is used when pasting a link from external
+                // source. This is because the link mark schema in `markdown-parser.ts` file uses `<nimble-anchor>` as anchor tag and not `<a>`.
```
vikisekarNI, line 432: Did you meant? 

```suggestion
                // To load the `nimble-anchor` from the HTML parsed content by markdown-parser as links in the Tiptap editor, the `parseHTML`
                // of Link extension should return nimble `anchorTag`.
                // This is because the link mark schema in `markdown-parser.ts` file uses `<nimble-anchor>` as anchor tag and not `<a>`. 
                // ---
                // `<a>` tag is added here to support when pasting a link from external
                // source.
```

vivinkrishna-ni: Yes, I think the suggested change is more clear.

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Since the `transformPastedHTML` accepts and returns HTML in the form of a string, it is already ignoring tags like `script` and I think they are handled within the ProseMirror. Here is the snapshot when pasting a `<script>alert('attack')</script>` string and trying to extract the input in this method,
image

As in the image, the tag values are escaped for the angular brackets of script tags and paste only the text content of it.

Also, I tried to modify the code by creating a script element just like the above example and returning its `innerHTML` it doesn't cause any issues. Since the HTML string is passed through this method every time and we just have it in the HTML element returning the same `innerHTML` and modifying only for anchor elements, it seems to be working as expected. I also wrote a test case to add script tags in the anchor tags and verified it renders as plain text.

vikisekarNI (approved):

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
rajsite: After searching around for a bit I think relying on the template element and it's document context should give an inert context for the HTML manipulation:


I'm having trouble finding documentation that manipulating the dom nodes of a template in a document fragment in this way actually preserves the safety guarantee of the template element..

```ts
const validAbsoluteLinkRegex = /^https?:\/\//i;
document.createElement('template');
template.innerHTML = '<a href="HTTPS://NIMBLE.NI.DEV">Nimble</a>';
const templateDocument = template.content.ownerDocument;
template.content.querySelectorAll('a').forEach(anchorElement => {
        const href = anchorElement.getAttribute('href');
        // When pasting a link, the `href` attribute of the anchor element should be a valid HTTPS/HTTP link;
        // else, it should be rendered as plain text in a paragraph element.
        if (href && validAbsoluteLinkRegex.test(href)) {
            anchorElement.textContent = href; // Modifying the anchor element text content with its href
        } else {
            const paragraphElement = templateDocument.createElement('p');
            paragraphElement.textContent = anchorElement.textContent;

            // TODO under what circumstance will the parentNode be nullish?
            // Do you have a test case for this?
            anchorElement.parentNode?.replaceChild(
                paragraphElement,
                anchorElement
            );
        }
    });

console.log(template.innerHTML);
```

rajsite (changes requested):

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Regarding point 3, we have restricted the image pasting event by utilizing the [`handlePaste`](https://prosemirror.net/docs/ref/#view.EditorProps.handlePaste) method and prevent the event when the image or any files are uploaded to the editor.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,53 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                    let tempAnchorElement: HTMLAnchorElement | Anchor = anchorElement;
+                    if (
+                        anchorElement.parentElement
+                        && anchorElement.parentElement.tagName === 'NIMBLE-ANCHOR'
+                    ) {
+                        tempAnchorElement = anchorElement.parentElement as Anchor;
+                    }
+                    const href = anchorElement.getAttribute('href');
+                    // When pasting a link, the `href` attribute of the anchor element should be a valid HTTPS/HTTP link;
+                    // else, simply rendered as a plain text in the same node by creating a span element and replaced it with anchor element.
```
vikisekarNI, line 362:
```suggestion
                    // When pasting a link, the `href` attribute of the anchor element should be a valid HTTPS/HTTP link;
                    // else, simply rendered as a plain text in the same node by creating a span element and replace it with anchor element.
```

vikisekarNI (approved):

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,53 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+        /**
+         * @param htmlString contains the html string of the copied content. If the content is a link, the `htmlString` contains anchor tag and a href value.
+         * ProseMirror reference for `transformPastedHTML`: https://prosemirror.net/docs/ref/#view.EditorProps.transformPastedHTML
+         */
+        const transformPastedHTML = (htmlString: string): string => {
+            const templateElement = document.createElement('template');
+            templateElement.innerHTML = htmlString;
+            const templateDocument = templateElement.content.ownerDocument;
+
+            templateElement.content
```
jattasNI, line 350: I'd personally be surprised if a hyperlink (text backed by a link) was replaced by the URL on paste. I would expect the text contents to be pasted and the link to be stripped away. I'm basing this on the behavior I see in plain text editors like VSCode.

What's the basis for the currently implemented behavior? Is there a PO request or customer workflow that requires it?

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,53 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
 
     private createTiptapEditor(): Editor {
         const customLink = this.getCustomLinkExtension();
+        const validAbsoluteLinkRegex = /^https?:\/\//i;
+
+        /**
+         * @param htmlString contains the html string of the copied content. If the content is a link, the `htmlString` contains anchor tag and a href value.
+         * ProseMirror reference for `transformPastedHTML`: https://prosemirror.net/docs/ref/#view.EditorProps.transformPastedHTML
+         */
+        const transformPastedHTML = (htmlString: string): string => {
```
jattasNI, line 345: What is the expected behavior for plain text URLs embedded within a larger paragraph? For example:
```
The URL https://www.ni.com takes you to NI's homepage.
```

I'm seeing that these URLs remain plan text unless you move the cursor to the end of the URL and type space/Enter.

Is that expected (or at least a known limitation)? I think my personal expectation would be for the editor to detect valid URLs within the entire pasted text. That seems to be the expectation in the reported bug too, since it is complaining about having to type space/enter after pasting a link (though it's unclear if that was a plain text link or a hyperlink). [Bug 2525827](https://dev.azure.com/ni/DevCentral/_workitems/edit/2525827): Rich text editor | Absolute links requires space or enter after pasting the link

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
...
+                    },
+                    {
+                        name: 'Hyperlink(http) within paragraph node',
+                        input: '<p>Anchor between <a href="http://nimble.ni.dev">Nimble</a> text</p>',
+                        url: 'http://nimble.ni.dev',
+                        textContent: 'Anchor between http://nimble.ni.dev text'
+                    },
+                    {
+                        name: 'Absolute link(https) as nimble-anchor within paragraph node',
+                        input: '<p>Anchor between <nimble-anchor href="https://nimble.ni.dev"><a href="https://nimble.ni.dev">https://nimble.ni.dev</a></nimble-anchor> text</p>',
```
m-akinc, line 1162: I'm confused by what is being tested here. Firstly, nesting `<a>` inside `<nimble-anchor>` is a strange and not particularly relevant use case, it seems (i.e. why would we ever encounter this?). Secondly, if you are going to test nested anchors, why aren't there any test cases that have `<nimble-anchor>` inside `<a>`? Thirdly, what if the `href` values of the `<a>` and `<nimble-anchor>` are different? Any why aren't there any test cases (anywhere in this file) of just `<nimble-anchor>` without a nested `<a>`? I'm not suggesting you add more test cases. I'm just trying to understand what about these cases that you're written are relevant. Would it be sufficient to just test `<nimble-anchor>` without the nested `<a>`?

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
...
+                        url: 'http://nimble.ni.dev',
+                        textContent: 'Anchor between http://nimble.ni.dev text'
+                    }
+                ] as const;
+
+                parameterizeNamedList(
+                    differentProtocolLinks,
+                    (spec, name, value) => {
+                        spec(
+                            `${name} renders as absolute link(href and text content should be same) in editor`,
```
m-akinc, line 1178: I just want to point out that these tests would pass even if the returned HTML was `<p><a href="...">Anchor between ... text</a></p>`, which is not what you expect. I'm not saying you need to change these tests, but it may be worth being more precise when writing future tests.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
...
+                                        'href'
+                                    )
+                                ).toBe(value.url);
+                            }
+                        );
+                    }
+                );
+            });
+
+            describe('pasting various links between nodes should render as absolute links between nodes ', () => {
```
m-akinc, line 1146:
```suggestion
            describe('pasting various links within text should render as absolute links within text', () => {
```

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
                     );
                 }
             });
+
+            describe('pasting various valid(https/http) links should render as absolute links', () => {
+                const differentProtocolLinks = [
```
m-akinc, line 1057: nit: I don't think you need to test every permutation. It would probably be sufficient to have a couple cases for absolute vs. hyperlink, then maybe, for example, one testing an uppercase https, and one testing a mixed case http.

This is mostly a suggestion for the future. If you want to leave all these test cases, you can.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
...
+                                    )
+                                ).toBe(value.url);
+                            }
+                        );
+                    }
+                );
+            });
+
+            describe('pasting various valid(https/http) nimble-anchor links should render as absolute HTML anchors', () => {
+                const differentProtocolLinks = [
```
m-akinc, line 1204: Again, I'm confused by the nested links, and I'm skeptical that all these permutations are necessary.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,395 @@ describe('RichTextEditor', () => {
...
+                            expect(pageObject.getEditorLeafContents()).toEqual([
+                                value.text
+                            ]);
+                        });
+                    }
+                );
+            });
+
+            describe('pasting various not supported nimble-anchor links should render as plain text', () => {
+                const differentProtocolLinks = [
```
m-akinc, line 1360: No need to separate these test cases from the set above. The test function is exactly the same. But I'd also question (again) what is really being tested here with the nested links, and if all of these cases are necessary.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,53 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+        };
+
+        // To prevent the pasting of files/images into the editor and avoid making any network requests.
+        // ProseMirror handlePaste: https://prosemirror.net/docs/ref/#view.EditorProps.handlePaste
+        // The below function needs an update when image support is added
+        const handlePaste = (
+            _view: EditorView,
+            event: ClipboardEvent
+        ): void => {
+            if (event.clipboardData && event.clipboardData.files.length >= 1) {
```
m-akinc, line 382: Are you confident that everything that could result in a side effect (like making network requests) will be represented by a file in the `clipboardData`?

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/index.ts
rajsite: I think we need to push back on this whole direction. 
1. It seems like the whole thing is a hack that will be removed when full link support is added based on the comment "Pasting rules need not to be transformed when hyperlink support added".
2. It was not a trivial fix. The assumption was incorrect. Safely handling HTML has been difficult. First it was done unsafely and then I suggested an alternative that I think might be safe and asked to find a way to confirm that but did not receive a compelling confirmation. 
    
    The concern was not that upstream tiptap will safely handle pasted HTML upstream inside the editor. The concern is that we are safely doing it which deep diving into the editor does not confirm.
3. It seems to require duplicating work tiptap, etc are handling upstream (sanitization, parsing paste results, etc).
4. It is exposing us to a whole new domain I don't know how to deal with correctly. I see a whole bunch of logic added to handle nimble-anchor tags in the pasted html I do not understand and was not cited in comments. Turns out that copying and pasting a rendered nimble-anchor results in a html string chunk like the following:
    ```
    <html>
    <body>
    <!--StartFragment--><span style="color: rgb(0, 0, 0); font-family: &quot;Source Sans Pro&quot;, &quot;Source Sans Pro Fallback&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;"><span> </span>the<span> </span></span><nimble-anchor href="https://nimble.ni.dev" appearance="default" style="display: inline; box-sizing: border-box; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-numeric: ; font-variant-east-asian: ; font-variant-alternates: ; font-variant-position: ; font-weight: 400; font-stretch: ; font-size: 14px; line-height: ; font-family: &quot;Source Sans Pro&quot;, &quot;Source Sans Pro Fallback&quot;; font-optical-sizing: ; font-kerning: ; font-feature-settings: ; font-variation-settings: ; color: rgb(0, 0, 0); letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a class="control" part="control" href="https://nimble.ni.dev/" style="color: var(--ni-nimble-link-font-color); text-decoration: underline;"><span class="content" part="content"><slot>link</slot></span></a></nimble-anchor><span style="color: rgb(0, 0, 0); font-family: &quot;Source Sans Pro&quot;, &quot;Source Sans Pro Fallback&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;"><span> </span>to n</span><!--EndFragment-->
    </body>
    </html>
    ```
    
    Some takeaways: 
        - The copy and paste results in the internal shadow dom of the nimble-anchor being exposed in the html chunk and all the internal html like slots, etc. Which is kinda nuts to deal with.
        - Styling gets inlined in some way I don't understand
        - I have no idea how cross-browser consistent any of this is

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: As we mentioned in the other group, we quickly aligned with Josh and it is also a request from him to paste it as a link when copied from the Edge browser. 

Due to the limitations of supporting only the absolute link without the hyperlink, we could come up with this approach when pasting only the URL when the hyperlinks are copied and pasted. Since this is impacting directly the user's point of view who might be using Edge browsers they struggle to identify what is pasted here in the editor.

We are currently capturing all the hyperlink workaround removal in a separate issue. We will remove this workaround when the support is added as we aligned before.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: The expected behavior that we are trying to achieve here is if the URL is already a link when copied from the source, it should render as a link; if it is plain text, it can remain plain text. Here is the short GIF explaining this scenario copied from a viewer and pasting it in the editor. Even the AzDo Board comments work the same.
copy-paste-link-in-paragraph

packages/nimble-components/src/rich-text/editor/index.ts
vikisekarNI: This behavior was implemented for [Bug 2525823](https://dev.azure.com/ni/DevCentral/_workitems/edit/2525823): Rich text editor | Copying a link from Edge browser which is not pasting as a link

### **Edge browser**
1. When copying URL from Edge browser's URL bar, we have both types `text/plain` and `text/html` available. 
image

2. If we look at the options in edge browser for pasting - link paste seems to be default. Below is a snip of edge browser with Google search loaded and I was trying to copy and paste the link from URL bar
image


### **Chrome & Firefox browser**
In chrome and firefox, while copying, it copies only `text/plain` which leaves us to paste only plain text
image

### **Copying nimble anchor**
However copying `nimble-anchor` is adding additional inline styles which we might need to investigate further through a separate issue for `nimble-anchor`
image

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Basically, what am trying to test here is when copying a `nimble-anchor` from the viewer or any other source, in the clipboard, it stores as text/html as below.

<details><summary>HTML string when copied <code>nimble-anchor</code> from viewer</summary>
<p>

&lt;html&gt; &lt;body&gt; &lt;!--StartFragment--&gt;&lt;span style=&quot;color: rgb(0, 0, 0); font-family: &amp;quot;Source Sans Pro&amp;quot;, &amp;quot;Source Sans Pro Fallback&amp;quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;&quot;&gt;&lt;span&gt; &lt;/span&gt;the&lt;span&gt; &lt;/span&gt;&lt;/span&gt;&lt;nimble-anchor href=&quot;https://nimble.ni.dev&quot; appearance=&quot;default&quot; style=&quot;display: inline; box-sizing: border-box; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-numeric: ; font-variant-east-asian: ; font-variant-alternates: ; font-variant-position: ; font-weight: 400; font-stretch: ; font-size: 14px; line-height: ; font-family: &amp;quot;Source Sans Pro&amp;quot;, &amp;quot;Source Sans Pro Fallback&amp;quot;; font-optical-sizing: ; font-kerning: ; font-feature-settings: ; font-variation-settings: ; color: rgb(0, 0, 0); letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;&quot;&gt;&lt;a class=&quot;control&quot; part=&quot;control&quot; href=&quot;https://nimble.ni.dev/&quot; style=&quot;color: var(--ni-nimble-link-font-color); text-decoration: underline;&quot;&gt;&lt;span class=&quot;content&quot; part=&quot;content&quot;&gt;&lt;slot&gt;link&lt;/slot&gt;&lt;/span&gt;&lt;/a&gt;&lt;/nimble-anchor&gt;&lt;span style=&quot;color: rgb(0, 0, 0); font-family: &amp;quot;Source Sans Pro&amp;quot;, &amp;quot;Source Sans Pro Fallback&amp;quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(244, 244, 244); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;&quot;&gt;&lt;span&gt; &lt;/span&gt;to n&lt;/span&gt;&lt;!--EndFragment--&gt; &lt;/body&gt; &lt;/html&gt;

</p>
</details> 

In the above HTML string, we can clearly see that the `<a>` tag is nested inside a `<nimble-anchor>`. So as to simulate the same pasting behavior of `nimble-anchor`, I just added a `<a>` within nimble-anchor which will fail if we remove the `transformPasted` logic from this code. 

Also, the case about having different `href` for nimble anchor and HTML anchor is technically not possible we did write that case.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Removed duplicate permutations of tests

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Removed these permutations of tests

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: I have mentioned the reason for a nested link in the other comment and removed this repeated test case for the nimble-anchor case

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: We have completely removed this method as getting data:url is not an issue as we discussed over a call yesterday.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,45 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
 
     private createTiptapEditor(): Editor {
         const customLink = this.getCustomLinkExtension();
+        const validAbsoluteLinkRegex = /^https?:\/\//i;
+
+        /**
+         * @param slice contains the Fragment of the copied content. If the content is a link, the slice contains anchor text node with link mark.
+         * ProseMirror reference for `transformPasted`: https://prosemirror.net/docs/ref/#view.EditorProps.transformPasted
+         */
+        const transformPasted = (slice: Slice): Slice => {
```
jattasNI, line 345: nit: my personal preference would be for `transformPasted` and `updateLinkNodes` to be `private` methods on the class rather than arrow functions. I believe that should be a trivial change (I don't see any state that they're capturing) but if there's some reason that the current approach is necessary or preferable then I don't feel too strongly.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
jattasNI: As discussed offline let's delete the nimble-anchor cases and instead file a tech debt item to track writing an integration test which instantiates a viewer and an editor, and copies from one to the other. That will be more robust if the format of the copied nimble anchor ever changes.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1052,6 +1052,259 @@ describe('RichTextEditor', () => {
...
+                                        'href'
+                                    )
+                                ).toBe(value.url);
+                            }
+                        );
+                    }
+                );
+            });
+
+            describe('pasting various not supported links should render as plain text', () => {
```
jattasNI, line 1228: Could you add one case that pastes a plain text URL and validates that renders as plain text (unless you already cover that somewhere and I missed it)?

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: As discussed offline, the current implementation is better than using the HTML strings so I am resolving this comment.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Yes, I have removed the test cases related to `nimble-anchor` and created a tech debt here: https://github.com/ni/nimble/issues/1568

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: I have updated the assertion method as `getEditorTextContents` which returns the array of text contents in the editor. So in this case `<p><a href="...">Anchor between ... text</a></p>`, this newly added method in the page object will return an array with a whole content in both the index.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: I have removed the `nimble-anchor` cases from the test files.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
vivinkrishna-ni: Added a test case.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: We were able to move the `updateLinkNodes` to be a private method but moving `transformPasted`, creates an issue. When we tried moving this to a private method, the function was not added to the [`transformPasted`](https://prosemirror.net/docs/ref/#view.EditorProps.transformPasted) of `editorProps` and renders the hyperlink as such and not transformed to an absolute link.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,6 +336,45 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                        });
+                    }
+                });
+            };
+            const fragmentAsJson = (slice.content.toJSON() as NodeObj[]) ?? [];
+            updateLinkNodes(fragmentAsJson);
+            const modifiedFragment = Fragment.fromJSON(
+                this.tiptapEditor.schema,
+                fragmentAsJson
+            );
```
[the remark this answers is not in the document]
vivinkrishna-ni: Initially, we tried using the prose-mirror model to manipulate the fragment, but the [text content](https://prosemirror.net/docs/ref/#model.Node.text) of any node is a read-only property. 

The other approach we also tried is, after finding the `link` mark in a `text` node, thinking of creating a new `text` node with the same marks and replacing it in the same object. However, we felt that this operation involves modifying the node replacements which might cause an issue with the other supported formatings. We also couldn't find a way to just update the text content without affecting the existing nodes. So we decided to utilize the JSON formats to update the text nodes without altering the existing nodes.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -333,9 +335,43 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
     }
 
+    private readonly updateLinkNodes = (nodeArray: NodeObj[]): void => {
+        nodeArray.forEach((node: NodeObj) => {
+            if (node.type !== 'text' && node.content) {
+                this.updateLinkNodes(node.content);
+            } else {
+                node.marks?.forEach((mark, index: number) => {
+                    if (mark.type === 'link' && mark.attrs) {
+                        if (this.validAbsoluteLinkRegex.test(mark.attrs.href)) {
```
rajsite, line 345: Why do we need to manually do this validation here? Shouldn't the Link mark validation be handling this for us downstream?

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -350,6 +386,12 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
             // Lists do not have any default paste rules, they have only input rules, so disabled paste rules
             // https://tiptap.dev/api/editor#enable-paste-rules
             enablePasteRules: false,
+            editorProps: {
+                // Validating whether the links in the pasted content belongs to the supported scheme (HTTPS/HTTP),
+                // and rendering it as a link in the editor. If not, rendering it as a plain text.
+                // Also, updating the link text content with its href as we support only the absolute link.
+                transformPasted
```
rajsite, line 393: We should be able to remove the entire transformPasted when full hyperlink support is added right? Why was that removed?

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
rajsite: If this workaround is not temporary it seems like it would be better to understand the prose-mirror model behavior better. I would be surprised if the model cannot handle updating the text for a node.

If this is all a temporary workaround that will not need to be maintained long-term across prose mirror updates that will be scheduled to removed soon then the weaker typing being done here may be a an ok temporary implementation.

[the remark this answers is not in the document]
vivinkrishna-ni: After exploring deep into the ProseMirror model, found that we can utilize the given APIs to update the Link node value. I have updated the logic without converting to JSON formats and retains all the type information that ProseMirror offers.

We also tested all different kinds of content copied from external sources and verified it worked as expected. So we believe this can be more cleaner approach than the previous one. This method can be updated/removed once we support hyperlinks. I have added the necessary code comments too.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Actually, the validation added to the Tiptap editor in the Link mark is not working for the pasted content containing links. The validation is added only for the input rule i.e. by typing into the editor and pressing any special keys to convert it as a link. We found that there is an open issue in Tiptap explaining that the validation is not added for the pasted links. 

I captured this here - https://github.com/ni/nimble/issues/1527 and once the underlying Tiptap issue is resolved, we can remove this manual validation from here. 
image

I also saw the discussion in the above linked Tiptap where one mentioned it is resolved with the latest version `2.1.11` of Tiptap but when I locally bumped to the latest version, I could still reproduce the issue. And also noticed there are some other issues with the `linkOnPaste` behavior with the latest version so dropped the plan of updating the version as of now.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Yesterday, we thought it was needed for validating even for hyperlinks, but seems this is an issue with Tiptap. So I have added it back and captured the information in the hyperlink user story.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: As discussed offline, this bug is needed for this cycle release in SLE and so we have to add this workaround. We also aligned on all the above points and captured different issues in respective issues for tracking the work in the future. So I am closing this for now.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -333,9 +335,66 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+     * @param fragment Fragment containing the pasted content. [Fragment](https://prosemirror.net/docs/ref/#model.Fragment)
+     * @returns modified fragment from the `updatedNode` after updating the valid link text with its href value.
+     */
+    private readonly updateLinkNodes = (fragment: Fragment): Fragment => {
+        const updatedNodes: Node[] = [];
+
+        fragment.forEach(node => {
+            if (node.isText && node.marks.length > 0) {
+                let textNode = node;
+                node.marks.forEach(mark => {
```
rajsite, line 352: Do you really expect to have multiple `link` marks for a specific node? Is that possible? If so what tests covers those cases?

If not could this be written to show the assumption that a node only has one link mark, ie an algorithm like:

```
const mark = node.marks.find(mark => mark.type.name === 'link' && mark.attrs);
if (mark) {
  if (this.validAbsoluteLinkRegex.test(mark.attrs.href as string)) {
      updatedNodes.push()
  } else {
      updatedNodes.push()
  }
} else {
    updatedNodes.push(node)
}
```

Unless I'm misunderstanding it doesn't seem like iterating across all the marks is valuable.

rajsite (approved):

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Yes, you're correct. It is not possible to have multiple `link` marks in a single `text` node and it is unnecessary to loop through the marks just to identify and update the `link` mark. So I have made changes to the logic as you suggested and it works as expected 👍

PR #1560, opened 2023-09-21
Add reusable css for styling an element as visually hidden while keeping it available to the accessibility tree

mollykreis:
# Pull Request

## 🤨 Rationale

We have the same style copy/pasted in a few different places to make text visually hidden while still available in the accessibility tree. Therefore, it would be good to have that style single-sourced.

## 👩‍💻 Implementation

Create shared `cssPartial` style that can be used in other css files for components.
Replace existing duplicated code with new style.

## 🧪 Testing

Manually verified there aren't accessibility violations with the modified components.
Verified no changes in chromatic.

## ✅ Checklist


mollykreis:
@m-akinc, can you buddy these changes?

m-akinc (approved):

rajsite (approved):

Issue #1562, opened 2023-09-21
Resizing columns in table sometimes adjusts other column widths that should be unaffected


atmgrifter00:
# 🐛 Bug Report

There are particular scenarios where sizing a pair of columns which, as part of the interaction, do not adjust a subset of the other columns, but when the user finishes the resize (releases the mouse button), the columns that were unaffected by the user resize all of the sudden have their sizes changed.

The current implementation results in this behavior:
![TableColumnSizeBug](https://github.com/ni/nimble/assets/26874831/99c5c23d-53bf-43b5-8ef6-ce65c73b213c)

You can see that after the user releases the mouse, the column that was grown snaps back to a smaller size.

## 💻 Repro or Code Sample

[Angular example app](https://nimble.ni.dev/storybook/example-client-app/#/customapp)

## 🤔 Expected Behavior

The columns do not change their sizes upon release of the mouse after a column sizing operation.,

## 😯 Current Behavior

See gif above.

## 💁 Possible Solution

The problem is rooted in how we are trying to [calculate the new fractional widths for columns](https://github.com/ni/nimble/blob/10471e6b15de319c075a0656c2ae4c8ad0bdcb75/packages/nimble-components/src/table/models/table-layout-manager.ts#L239C1-L239C1) after an interactive resize.

Currently, in an effort to try and maintain existing fractional widths of columns, we are adjusting the fractional widths of the resized columns by multiplying the original fractional width by '(new pixel width)/(original pixel width)' (see above link). This results in the buggy behavior, because if one of the columns had a large enough min pixel width, the layout system will ultimately, after trying to grow that column, take some of that size back and distribute it amongst the other columns.

Instead, when growing any column, we can just set all current fractional widths to be a ratio of the column's current width against the total pixel width of all columns. The one side-effect of this

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #1563, opened 2023-09-21
Table column sizing bug fix

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1562

## 👩‍💻 Implementation

As described by the possible solution section in the linked issue, we can provide better behavior if we re-calculate all fractional widths after a user resize to be the ratio of the current width of the column over the total width of all columns. This implementation also re-normalizes these values, by dividing by the average column width, instead of the total width. This will keep the resulting fractional width values closer to their original values if they were untouched by the interactive resize.

There are a couple of consequences to this change that are worth calling out:
1. After a column resize, it is likely that any column that, for instance, had a fractional width of 1, would then have a smaller fractional width. One consequence of this is that if you were to add a new column, also having a fractional width of 1, it would now be larger than the other columns that also started off with a fractional width of 1. With the normalization, however, this difference shouldn't be very large.
2. Since all fractional widths are now being adjusted to be representative of what their current pixel width compared to the total is, this means that if you were to grow the table, _all_ columns would then being to grow proportionally. Before, this wasn't the case, as fractional widths of all columns were not updated, just the ones that were interactively resized. So, a column that had a larger minimum width than what its fractional width would grow the column to wouldn't necessarily adjust in size if the table width grew.

Ultimately, these changes in behavior seem like okay concessions in the face of the noted bug, and there doesn't seem to be a good way to have _all_ of the good behaviors.

## 🧪 Testing

Added unit test.

## ✅ Checklist


atmgrifter00:
@mollykreis can you provide a buddy fix for this?

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -222,10 +222,13 @@ export class TableLayoutManager<TData extends TableRecord> {
     }
 
     private resetGridSizedColumns(): void {
-        if (!this.gridSizedColumns) {
+        if (!this.gridSizedColumns || this.gridSizedColumns.length === 0) {
             return;
         }
 
+        const gridSizeColumnTotal = this.getTotalColumnFixedWidth();
```
mollykreis, line 229: `this.getTotalColumnFixedWidth()` doesn't return the _grid size_ column total, right? I believe function will return the total of _all_ columns. Is that what we want here? I'm thinking it is not what we want because it would lead to fairly small fractional widths being set on the fractional width columns. Potentially all fractional width columns could have a `currentFractionalWidth` set on them of well less than `1` if there are a lot of fixed width columns.

mollykreis (changes requested):

packages/nimble-components/src/table/models/table-layout-manager.ts
atmgrifter00: Oh that's right...will fix. I think this means I should write another test too.

atmgrifter00: @mollykreis, I've now added a couple of tests to verify the existence of a fixed width column doesn't cause any unexpected behaviors. I'm not entirely happy with the descriptions I've used, nor am I sure we're covering things as completely as possibly, and I'm open to suggestions.

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -222,10 +222,17 @@ export class TableLayoutManager<TData extends TableRecord> {
     }
 
     private resetGridSizedColumns(): void {
-        if (!this.gridSizedColumns) {
+        if (!this.gridSizedColumns || this.gridSizedColumns.length === 0) {
```
mollykreis, line 225: If you want, you could simplify this to one check. I don't have a strong preference either way.

```suggestion
        if (!this.gridSizedColumns?.length) {
```

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -441,6 +495,12 @@ describe('Table Interactive Column Sizing', () => {
                 `${columnSizeTest.name}`,
                 // eslint-disable-next-line @typescript-eslint/no-loop-func
                 async () => {
+                    element.columns.forEach((column, i) => {
```
mollykreis, line 498: It might be worth writing a test (or potentially updating these tests) to verify that each column has the expected fractional width after the resize operation. I could see that potentially being a pain because of the result being a seemingly arbitrary value with a lot of decimal places, but I think it would be beneficial to have at least one unit test that verifies that the "normalization" logic is working as expected.

atmgrifter00: We agreed that [this test](https://github.com/ni/nimble/blob/ab995f09749e33f9fdcb5a5975a8c1d84d8aca00/packages/nimble-components/src/table/tests/table-column-sizing.spec.ts#L492) adequately covers this concern along with the other new tests.

mollykreis (approved):

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -222,10 +222,17 @@ export class TableLayoutManager<TData extends TableRecord> {
...
-        if (!this.gridSizedColumns) {
+        if (!this.gridSizedColumns?.length) {
             return;
         }
 
+        let gridSizeColumnTotal = 0;
+        for (const gridSizedColumn of this.gridSizedColumns) {
+            gridSizeColumnTotal
+                += gridSizedColumn.columnInternals.currentPixelWidth!;
+        }
```
jattasNI, line 233: nit: I think we should prefer array methods over loops, but that's a preference not a coding convention. Something like this (you should check my work).
```suggestion
        const gridSizeColumnTotal = this.gridSizedColumns.reduce((sum, col) => sum + col.columnInternals.currentPixelWidth!, 0);
```

jattasNI (approved):
I played around with storybook and the resize behaviors felt intuitive to me

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -222,10 +222,16 @@ export class TableLayoutManager<TData extends TableRecord> {
     }
 
     private resetGridSizedColumns(): void {
-        if (!this.gridSizedColumns) {
+        if (!this.gridSizedColumns?.length) {
```
rajsite, line 225: nit / pref: Be explicit on size checks, feel like truthiness on size checks can be easily overlooked

```suggestion
        if (!this.gridSizedColumns || this.gridSizedColumns.length === 0) {
```

rajsite (approved):

PR #1564, opened 2023-09-22
Fix undefined number sort

### LINKED ISSUE #1561: Nimble Table number column sorts undefined/missing values by 0, instead of as the smallest value
# 🐛 Bug Report

The Nimble table currently sorts undefined/missing values right after 0, for ascending sort. This is different behavior than the string/text column, which sorts undefined as the smallest values (i.e. first for ascending sort, last for descending sort)

We also need the behavior of sorting undefined/missing values as smallest for [AzDo feature 2396309 (Sort results grid interactively on numeric custom property columns)](https://dev.azure.com/ni/DevCentral/_workitems/edit/2396309).

## 💻 Repro or Code Sample

Run Storybook on a local nimble-components dev build

Open `table-column-number-text.stories.ts` and change the sample data:
```ts
const simpleData = [
    {
        firstName: 'Homer',
        lastName: 'Simpson',
        age: 45.2358734623,
        favoriteNumber: Math.PI
    },
    {
        firstName: 'Marge',
        lastName: 'Simpson',
        age: 42.918275125,
        favoriteNumber: 28729375089724643
    },
    {
        firstName: 'Bart',
        lastName: 'Simpson',
        age: 13.5689,
        favoriteNumber: 1000
    },
    {
        lastName: 'MissingFirstName',
        age: 22,
        favoriteNumber: 55
    },
    {
        firstName: 'Test',
        lastName: 'MissingFavoriteNumber',
        age: 33
    },
    {
        firstName: 'Test2',
        lastName: 'UndefinedFavoriteNumber',
        age: 44,
        favoriteNumber: undefined
    },
    {
        firstName: 'Maggie',
        lastName: 'Simpson',
        age: 1.238957645,
        favoriteNumber: 0
    },
    {
        firstName: 'Milhouse',
        lastName: 'Van Houten',
        age: 14.1,
        favoriteNumber: -0.00000064532623
    }
];
```

## 🤔 Expected Behavior

Undefined/missing values are sorted as smallest (first for ascending, last for descending)

## 😯 Current Behavior

NimbleTableNumberUndefined

## 💁 Possible Solution

Switch from using `TableColumnSortOperation.basic` to one that handles undefined values specifically (or potentially change the behavior of that sort)

## 🔦 Context

As mentioned above, behavior needed for [AzDo feature 2396309 (Sort results grid interactively on numeric custom property columns)](https://dev.azure.com/ni/DevCentral/_workitems/edit/2396309)

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1561 

## 👩‍💻 Implementation

The TanStack `basic` sort doesn't sort `undefined` and `null` the way we want our basic sort to work. Therefore, I implemented our own `basicSortFunction` to use when the sort operation is `basic` rather than relying on TanStack's behavior.

Additionally, I realized that we had very few tests for our sorting functions. The tests we had were in `sort-operations.spec.ts`, which tests sorting at a table level. To avoid too much coupling in our tests between concrete column types and sort operations, I decided to write some tests that only exercise the sort functions.

## 🧪 Testing

- Wrote a bunch of new unit tests for sorting
- Manually verified that `undefined` and `null` do not get sorted with `0` for the numeric column

## ✅ Checklist


packages/nimble-components/src/table/tests/table-sorting.spec.ts
```diff
@@ -252,24 +252,6 @@ describe('Table sorting', () => {
         expect(getRenderedRecordIds()).toEqual(['4', '2', '3', '1']);
     });
 
-    it('performs case sensitive string sort', async () => {
```
mollykreis, line 255: I removed the tests from this file that were sort-type specific. I thought it would be better to have them in `sort-operations.spec.ts` and leave these tests to be the table behavior around sorting rather than the actual sort algorithm.

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    ): TanStackRow<TableFieldValue> {
+        const row = jasmine.createSpyObj<TanStackRow<TableFieldValue>>(
+            'TanStackRow',
+            ['getValue']
+        );
+        row.getValue.and.returnValue(value);
+        return row;
+    }
+
+    const basicNumericSortOrder = [
```
mollykreis, line 18: Each of these `*SortOrder` arrays is ordered from smallest value to largest value. The tests call the sort operation on each index of each array with every index in the same array and verify that it returns the correct value. I had originally gone down the path of writing a bunch of individual tests cases for things like _undefined is equal to undefined_, _undefined is less than null_, _null is greater than undefined_, etc., but the list of tests was getting really long, which made it difficult to make sure we covered all the cases we wanted.

atmgrifter00 (approved):

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            testCases: [localeAwareCaseSensitiveSortOrder]
+        }
+    ] as const;
+    for (const sortOrderTestCase of sortOrderTestCases) {
+        for (const expectedSortOrder of sortOrderTestCase.testCases) {
+            for (let i = 0; i < expectedSortOrder.length; i++) {
+                for (let j = 0; j < expectedSortOrder.length; j++) {
+                    const sortOrderA = expectedSortOrder[i]!;
+                    const sortOrderB = expectedSortOrder[j]!;
+                    it(`comparing ${sortOrderA.name} to ${sortOrderB.name} returns expected value`, () => {
```
jattasNI, line 76: Could you add more context to the test case names to make them unique (include sort operation and test case name)?

You could also consider rewriting to use the [parameterize utility](https://github.com/ni/nimble/blob/b70ce309d6c03a356b5abaaea7b2199a61f7fbd5/packages/nimble-components/src/utilities/tests/parameterized.ts#L73). This is mostly a suggestion along the lines of "when your favorite tool is a hammer, everything looks like a nail". In fact it may be a bad suggestion. I suspect that would make the tests harder to read in this case: you'd still need just as many loops to generate all the test cases, plus the loop inside `paramaterize` to iterate over them. And since the test cases would be generated rather than static, it might still be hard to `xit` or `fit` on individual cases.

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+                        let expectedSortResult: number;
+                        if (i === j) {
+                            expectedSortResult = 0;
```
jattasNI, line 79: No action required, just documenting a train of thought I had about our position on [stable sorting](https://en.wikipedia.org/wiki/Category:Stable_sorts) (items with the same value being sorted in a consistent order)...

I was going to suggest that we update each sort order array to include a duplicate value and then verify that they are sorted in a stable order. But I guess since these tests cover the comparison function rather than the result of the sort, that wouldn't be any different than the existing case where `i === j` (and in fact those cases would break the assumption that `array[i] < array[i+1]`).

So to properly test stable sorting we'd need to ensure that equal values return a sort result of 0 (which these tests handle) and then also ensure that the TanStack sort function maintains sort order when the sort result is 0. The latter would presumably be the job of TanStack (or at least a configuration we pass to TanStack) but I'm not sure if they even support stable sort. I found a couple of issues requesting it that were closed with no action.
https://github.com/TanStack/table/issues/2188
https://github.com/TanStack/table/issues/913

Anyway, for the scope of this PR I think we're fine. Just recording this for discussion.

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
jattasNI: >Each of these *SortOrder arrays is ordered from smallest value to largest value. The tests call the sort operation on each index of each array with every index in the same array and verify that it returns the correct value.

I think that would be a useful comment somewhere in this file.

jattasNI (approved):

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
mollykreis: I restructured the tests. I'm hoping they are easier to read now. The downside of the restructuring is that now all the `expect` statements for a given sort operation + type is in one test, but I think the readability of the test outweighs the fact that the test name itself doesn't show exactly which comparison caused the test to fail. (Plus, as you pointed out, that wasn't the case before anyway since there were duplicate test names).

rajsite: This format is pretty good. Just to say it, to be on brand completely I'd expect each of the testCases to be separate describe blocks and the names of the specs generated as, for example `undefined equals undefined`:

```ts
describe('Sort Operations', () => {
    describe('basic sorting with numbers', () => {
        // generated via parameterize
        it('undefined equals undefined' , () => {
            // Then either duplicate the test code or parameterize a test function and call it
           // Trying to skip the test structure to avoid a small bit of duplication isn't worth it
        })
    });
});
```

That would uniquify the test names and still be very readable but I don't think it's particularly high value. This PR looks good but it's not a practice I'd advocate generally. If rationalizing maybe it's just for this lots of "mildly" interesting `expect` use cases.

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // The `expectedSortOrder` arrays in the test cases are ordered from smallest value
+    // to largest value. The tests call the sort operation on each index of each array
+    // with every index in the same array and verify that it returns the correct value.
+    const testCases = [
+        {
+            name: 'basic sorting with numbers',
+            sortOperation: TableColumnSortOperation.basic,
+            expectedSortOrder: [
+                undefined,
```
rajsite, line 27: I think we should also be testing `NaN` since it is a weird one as `NaN !== NaN` so that case should behave well

rajsite (commented):
Waiting to see how `NaN` is handled

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
mollykreis: I agree that this isn't generally the structure we'd want for our tests. I originally had gone down the path of creating individual tests for the various values we wanted to compare, but it became very difficult very quickly to figure out whether or not I had coverage for all the comparisons we wanted. That's how I landed on having a single array that defined the order and then looping through the array to make sure comparisons between each entry resulted in the correct value. I felt that in the case of these tests, it was better for the tests to more obviously handle all the cases we want to cover even if it required diverging a little from our typical patterns.

packages/nimble-components/src/table/models/tests/sort-operations.spec.ts
mollykreis: I've added handling and testing for `NaN`. I am sorting it between `null` and `NEGATIVE_INFINITY`, so let me know if you think it would be more appropriately sorted somewhere else.

rajsite: Seems like a good spot! The most number like not a number.

rajsite: I guess if we are also being complete with the IEEE754 edge cases we should show how `-0` will be sorted (a slightly smaller zero :P)

rajsite (approved):

packages/nimble-components/src/table/models/sort-operations.ts
```diff
@@ -46,3 +46,41 @@ function localeAwareCaseSensitiveSortFunction<TData>(
     }
     return 1;
 }
+
+/**
+ * A function to perform a basic sort of two rows from TanStack for a given column.
+ * The function sorts `undefined` followed by `null` before all other values that are
```
jattasNI, line 52: Don't forget to either mention NaN in the comment or change it to be more generic

PR #1566, opened 2023-09-25
File size table column spec

m-akinc:
# Pull Request

## 🤨 Rationale

Spec for file size table column

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


m-akinc:
@atmgrifter00 could you buddy this please? I guess it's more of a "preliminary" review, since you are a required reviewer anyway.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+A Blazor wrapper will be created for the component.
+
+### Visual Appearance
+
+Will use same text styling as the `nimble-table-column-text`.
+
+## Implementation
+
+```ts
+class TableColumnFileSize extends TableColumnTextBase {
```
atmgrifter00, line 71: Much of the implementation included here doesn't seem to pertain to the unique aspects of this column type. I would vote to remove any typical/boilerplate-ish code. In fact, I'm not sure there's much value in even providing the snippet in addition to the API description above. Most of the other implementation here is just a mirror of other patterns we have for dealing with how to update internal state.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+
+-   Configurable number of decimal places
+-   Configurable units (e.g. display all values in KB)
+-   Grouping by value ranges/buckets
+
+### Features
+
+-   Converts the byte value to largest unit (e.g. KB, MB, GB) that results in a value of >=1
+-   Option to use 1024-based units (e.g. KiB) or 1000-based units (e.g. KB)
+-   Displays up to one decimal place
```
atmgrifter00, line 21: Any reason why we aren't doing 2? "Quarters" are a fairly common mental unit that seems like it would be good to represent.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+-   Sorts/groups by numeric value (number of bytes) rather than display string
+-   Localized number and unit strings
+
+### Risks and Challenges
+
+### Prior Art/Examples
+
+Size column in [SLE Files grid](https://dev.lifecyclesolutions.ni.com/files/)
+
+Windows explorer (details view)
```
atmgrifter00, line 31: I would just include a picture.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+-   Option to use 1024-based units (e.g. KiB) or 1000-based units (e.g. KB)
+-   Displays up to one decimal place
+-   Sorts/groups by numeric value (number of bytes) rather than display string
+-   Localized number and unit strings
+
+### Risks and Challenges
+
+### Prior Art/Examples
+
+Size column in [SLE Files grid](https://dev.lifecyclesolutions.ni.com/files/)
```
atmgrifter00, line 29: Might be good to include a picture, just to avoid situations where someone clicks the link and doesn't have permissions.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+    <nimble-table-column-file-size field-name="size">Size</nimble-table-column-file-size>
+</nimble-table>
+```
+
+### API
+
+`nimble-table-column-file-size`
+
+-   `field-name` - name of the record field containing the data. Data must be a `number` byte count.
+-   `unit-type` - "binary" (KiB, MiB, GiB, TiB, PiB) or "decimal" (KB, MB, GB, TB, PB); defaults to "decimal"
```
atmgrifter00, line 50: What's influencing our decision to default to `decimal`?

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: You mean two decimal places? By "Quarters," are you saying we should be able to show "1.25"? I doubt the usefulness of showing precision to the hundreths place. SLE currently only shows up to one decimal place (though that's not a definitive argument). I'll leave this open to see if others have an opinion.

atmgrifter00: Yes, sorry, two decimal places. The usefulness, to me, is intrinsic in that .25 is a half of a half, which isn't an uncommon value to represent, particularly when dealing in concepts that are powers of 2. I would think seeing something like 1.3 MiB is odder than seeing something like 1.25 MiB.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,294 @@
...
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created.
+
+### Blazor integration
+
+A Blazor wrapper will be created for the component.
+
+### Visual Appearance
+
+Will use same text styling as the `nimble-table-column-text`.
```
atmgrifter00, line 66: Despite the presence of a unit, this column still feels more like a numeric text column in spirit. One thing I could imagine a client wanting to do is have all of the values right aligned, and possibly having them all have the same amount of fractional digits.

Ultimately, if feels like we're missing some design decisions that might dictate some necessary APIs.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I've removed most of the code snippets, except for the implementation of the conversion/formatting function. I added a couple of examples of markup and rendered output, as you suggested.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: KB/MB/GB seem much more prevalent KiB/MiB/GiB. For example, all of the prior art examples I gave, plus some below use the decimal-type units:
GitHub
image
Gmail
image
npmjs.com
image

I couldn't find any examples of KiB/MiB/GiB being used in the wild. Granted, it's not clear if every example I found was actually using a 1000 conversion factor, but I don't have a reason believe otherwise.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I agree that fundamentally, we're displaying a numeric value, but I don't think that necessarily brings along any additional requirements compared to a text column (assuming that both are read-only).

I've added a justification for left-aligning values in the **Visual Appearance** section. That argument hinges on us not using a consistent number of decimal digits and not using a consistent unit. If we think it is worth providing those options, then right-alignment could make sense. I don't think we currently have any clients asking for that, but there's nothing keeping us from adding that support at a later time.

Let's see what the rest of the team thinks.

atmgrifter00: I'll just note that in all of the prior art you provided, save SLE, columns displaying file size were right-aligned :). This may raise a question of whether or not we should consider that a "user expectation". Ultimately, I agree, that these are features that can be added over time. Given that, I think it might make sense to word your explanation for doing left-alignment a bit differently, in that we shouldn't say this column _won't_ do that, but that if we decide that we should allow users this flexibility we can add those features in a non-breaking way.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+`nimble-table-column-file-size`
+
+-   `field-name` - name of the record field containing the data. Data must be a `number` byte count.
+-   `unit-type` - "binary" (KiB, MiB, GiB, TiB, PiB) or "decimal" (KB, MB, GB, TB, PB); defaults to "decimal"
+
+The component will extend `TableColumnTextBase`, thereby including the APIs for a **groupable** and **fractional-width** column.
+
+### Examples
+
+Given a data value of 2856 (bytes):
```
atmgrifter00, line 56: <nit> The grokking of what is going on here might be helped if the value was simply 1024 (or some integer multiple of that).

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+    return `${formatter.format(currentSize)} ${unitStrings[unitIndex]}`;
+}
+```
+
+Group header view will extend `TableColumnTextGroupHeaderViewBase` and use its template.
+
+### States
+
+N/A
```
jattasNI, line 140: Are there any invalid states? What would we show for negative numbers?

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
mollykreis: I'll also point out that the existing SLE code shows no decimal place for bytes and a maximum of 1 decimal place for everything else.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Features
+
+-   Converts the byte value to largest unit (e.g. KB, MB, GB) that results in a value of >=1
+-   Option to use 1024-based units (e.g. KiB) or 1000-based units (e.g. KB)
+-   Displays up to one decimal place
+-   Sorts/groups by numeric value (number of bytes) rather than display string
+-   Localized number and unit strings
+
+### Risks and Challenges
```
mollykreis, line 25: I'm not sure if this should be listed here or not, but definitely a risk for integration within SLE is that we'll now have two totally different implementations for how to convert a number into a formatted file size -- one for when the value is displayed in the table and one for when the value is displayed anywhere else (i.e. just using the existing Angular pipe).

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+![Chrome Dev Tools Network tab](../spec-images/fileSizeChromeDevTools.png)
+
+---
+
+## Design
+
+### API
+
+`nimble-table-column-file-size`
```
mollykreis, line 47: You mentioned above that this doesn't have to be used specifically for file sizes. Would it make sense to make the name a bit more generic? At least at this point in time, I don't have any good suggestions, but I thought I'd at least start a discussion.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+-   `field-name` - name of the record field containing the data. Data must be a `number` byte count.
+-   `unit-type` - "binary" (KiB, MiB, GiB, TiB, PiB) or "decimal" (KB, MB, GB, TB, PB); defaults to "decimal"
```
mollykreis, line 50: I'll point out that this is a difference in behavior from what exists in SLE -- particularly, file sizes there are displayed in `KB`, `MB`, `GB`, but they are determined using 1024 division into the original "bytes" value.

Referring back to a potential risk I mentioned in a different comment, I think we need to be careful about diverging too much from existing code since that code can be used in other places within an applications. Some thoughts about how to resolve this:
- Make sure POs are aligned with our behavior change (i.e. either starting to use different units or starting to show values converted to upper units using 1000 rather than 1024)
- Update the `format-bytes` pipe to match what the column SLE will use so that you don't see one value in the table and a different value when looking at the details of a file

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+The component will extend `TableColumnTextBase`, thereby including the APIs for a **groupable** and **fractional-width** column.
```
mollykreis, line 52: To confirm, do we think it makes sense for this column to be groupable? That doesn't seem particularly useful unless we think there's a use case for this column that would only have a discrete set of file sizes.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
mollykreis: I agree with @m-akinc -- the rationale we were given by visual designers was that right-aligned values made sense if all values had the same number of decimals and the same length of units.

Perhaps, this is really a question for the visual designers to see what their expectation is for this column's alignment.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+**Cons:** Only supports a fixed set of languages. Up to us to find accurate translations. Unsure if subtags other than the language (e.g. region) could be relevant to the translation.
+
+2. Use Nimble label provider. Add label tokens for the units, and rely on clients to provide translations.
+
+**Pros:** Supports any language a client cares to provide translations for.
+
+**Cons:** Less convenient for clients. Does not honor the `lang` setting on the page or on `nimble-theme-provider`.
+
+I suggest we go with option 1, primarily because I am hesitant to add twelve new label provider strings that clients are expected to localize. I suggest initially supporting the following languages:
```
mollykreis, line 172: Given those two options, I think option 2 is better. I don't think maintaining a map of localized strings within nimble is a good idea. I'm not sure how you came up with the list of locales we should initially support, but I don't think it should be the responsibility of nimble to add explicit support for any language that a client wants to use the product in.

We've established the pattern of using the label provider, and I don't think we should avoid using we think it is inconvenient for clients.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Documentation
+
+This component will be documented via a new story in Storybook.
+
+---
+
+## Open Issues
+
+-   Is there a better name for this column type? E.g. `nimble-table-column-memory-size`.
```
mollykreis, line 222: I was wondering the same thing, particularly because you mentioned that it doesn't need to be used specifically for file sizes. Unfortunately, I don't have a better name to suggest at this point in time.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
m-akinc: I can't believe I missed that. However, after just a little experimentation, I'm pretty unhappy with some of the formatting and translations.
English:
image
- No space betwen number and units
- "B" instead of "byte" or "bytes"
- "BB" instead of "GB"

French:
image
- inconsistent spacing between number and units
- "o" instead of "octet" or "octets"
- "Mdo" instead of "Go"
- "Bno" instead of "To"

German:
image
- these all seem weird, especially the lack of KB

Spanish
image
- no GB, and others seem questionable

I could go on, but in general, I have a lot of doubts about letting `Intl.NumberFormat` handle the unit translation.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I don't think there are any. The only configuration option is `unit-type`, which will be strongly typed to disallow invalid values. For invalid values like `Inf`, we just show an empty cell rather than putting the whole column (and table) in an invalid state. I think we should support negative numbers in case a client wants to use this column to represent some kind of size delta.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I can't think of any practical way to avoid that. I don't think we want to let the client provide a formatting function, or to export our formatting function for a client to use elsewhere. Luckily, I don't think our formatting will be so different from the Angular pipe that it will be a glaring problem.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: That's a fair point. I wonder, though, is it worth removing the ability to group from this column, when every (?) other column type allows it? I guess I very slightly lean towards making it non-groupable.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
mollykreis: As proposed now, I think it will be a problem. We're either going to have different units or a different value with the same units compared to the existing pipe.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
m-akinc: For that specific example, I highly doubt we would have a progress-bar-like column that is _only_ meant to show progress of files/file sizes. I'm having a hard time coming up with another presentation (for file/memory sizes, specifically) that is plausible. I'm not putting my foot down, but I think this may be a case where we don't need to tack `-text` onto the end of the element name.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: True, we will need to change either the units or the factor used by the current pipe, as you point out in your other comment.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
jattasNI: I agree that Nimble shouldn't pick a subset of languages to support and I agree that Nimble shouldn't do the work to translate every language, so between those options I'd also vote for the label provider approach.

If we say using the Intl.NumberFormat translations (https://github.com/ni/nimble/pull/1566/files#r1338793620) is option 3 then one more idea for option 4 is to pull in a library that does the translation for us. It'd have to be better than the browser in some way, standards-based, and not too big or slow, so I dunno if we'd find one that meets our needs. But one that came up in my searching was [format.js](https://formatjs.io/). I'm not sure if it has any capabilities beyond Intl.NumberFormat though.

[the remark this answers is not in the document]
m-akinc: I have added options 3 and 4 that are variations on using `Intl.NumberFormat` just for unit translation. Both require giving up support for the KiB-type unit labels in one way or another.

Because of the cons of the `Intl.NumberFormat` approaches, I'm now advocating for option 2. I'm waiting to hear others' votes, though.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I've added this concern to the section, as suggested.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: I will figure out which of these approaches to take when uptaking the new column in SLE.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: format.js [uses `Intl.NumberFormat`](https://formatjs.io/docs/intl#formatnumber), so that doesn't provide anything different. I could not find any other libraries that can do the translation for us. I did add two new options that are based on using `Intl.NumberFormat`. I still think option 2 is preferable, though. I'm waiting to hear others' opinions.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,222 @@
+# File Size Table Column
```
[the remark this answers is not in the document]
m-akinc: Done

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
atmgrifter00: We had an out-of-thread discussion regarding limiting grouping in the API and decided that it was sufficient for clients to simply leverage the [`groupingDisabled` API](https://github.com/ni/nimble/blob/990e415d40fa3abe36414c1736b2133f7cd0cd03/packages/nimble-components/src/table-column/mixins/groupable-column.ts#L20C16-L20C33) if they wanted to prevent users from grouping by this column.

I could imagine a scenario where a client wanted to use this column to represent products that had a very discrete set of values for byte size (i.e. hard drives or memory sticks), and it would make a ton of sense for those values to be grouped by.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+N/A
+
+### Documentation
+
+This component will be documented via a new story in Storybook.
+
+---
+
+## Open Issues
```
jattasNI, line 230: One alternative that we should consider: a single column type that supports displaying a numeric value and a unit for a fixed set of units. To start with it would be used for file size and elapsed time, but eventually someone could add support for V, degrees C, etc.

A possible API would be a single attribute like `unit = 'byte-binary' | 'byte-decimal' | 'elapsed-time' | 'volt' | ...`

Or maybe `unit = 'byte' | 'second' | 'volt'` and `byte-format = 'binary' | 'decimal'` and `second-format = 'h:m:s' | 'h,m,s'`.

Maybe worth talking about in person with the team to move the discussion along.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
jattasNI: I'm also liking option 2 best right now.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
mollykreis: As a point of reference, it seems like Windows explorer (which shows units as KB, MB, GB), uses a conversion factor of 1024 (although sometimes the math really looks like they are using 1025, which is a bit perplexing).

In this example from the "Properties" window for a file, you can see that the KB is not the bytes divided by 1000:
image

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Localization of the unit label
+
+This is more difficult, as there does not seem to be a great API for getting a file size unit string for a given locale. The only implemenation I've found is `Intl.NumberFormat`. This function can also do the numeric conversion, but we would not use it for that, because it only supports converting by using a factor of 1000, not 1024. We could instead do the number conversion ourselves and just ask `Intl.NumberFormat` to translate the unit string we need. However, we would have to work around several limitations:
+
+-   Does not have translations for KiB/MiB/etc.
+-   Does not translate plural "bytes".
+-   Throws error for locales with lang subtag `zh_CN` and `zh_TW` -- expects just `zh`.
+-   Requires creating separate `Intl.NumberFormat` instances for each unit we wish to translate.
+-   Missing translations for "byte" in French (gives "o" instead of "octet"), Japanese ("バイト"), Chinese ("字节" in simplified or "位元組" in traditional), Korean ("바이트"), Russian ("Б" instead of "байт"), and probably others.
```
mollykreis, line 166: I'm not convinced that there are actually missing translations. I think you just need to use `unitDisplay: 'long'` in the case that you set `unit: 'byte'`. Then, I believe you get all the expected strings, including the English distinction of `byte` vs `bytes` (which is called out as a problem a few bullet points above).

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+For this we will use `Intl.NumberFormat`, passing in the value of the `lang` design token as the locale.
+
+#### Localization of the unit label
+
+This is more difficult, as there does not seem to be a great API for getting a file size unit string for a given locale. The only implemenation I've found is `Intl.NumberFormat`. This function can also do the numeric conversion, but we would not use it for that, because it only supports converting by using a factor of 1000, not 1024. We could instead do the number conversion ourselves and just ask `Intl.NumberFormat` to translate the unit string we need. However, we would have to work around several limitations:
+
+-   Does not have translations for KiB/MiB/etc.
+-   Does not translate plural "bytes".
+-   Throws error for locales with lang subtag `zh_CN` and `zh_TW` -- expects just `zh`.
```
mollykreis, line 164: I might be missing something here, but don't we generally use `-` rather than `_` in the locales? Passing a language of `zh-CN` or `zh-TW` appears to work just fine with `Intl.NumberFormat`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+-   Requires creating separate `Intl.NumberFormat` instances for each unit we wish to translate.
```
mollykreis, line 165: I'm not sure if having to create separate `Intl.NumberFormat` instances for each unit is really a limitation, but I don't have a huge issue with calling it out as something we'd have to do if we went with this approach. We aren't talking about a huge number of units (I believe 6?).

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Pros:** Supports any language a client cares to provide translations for.
+
+**Cons:** Less convenient for clients. Does not honor the `lang` setting on the page or on `nimble-theme-provider`.
+
+3. Use `Intl.NumberFormat` and drop support for units based on a factor of 1024.
+
+**Pros:** No translation work for the client or Nimble team. Honors `lang` value.
+
+**Cons:** No support for 1024-based units. Shows "byte" unit even for multiple bytes. Missing several translations for "byte". Unclear if missing translations for other units (KB, MB, etc.).
```
mollykreis, line 186: As mentioned above, I don't think the unit translations is an issue here, nor is the plural "bytes".

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+4. Use `Intl.NumberFormat` and support both 1024 or 1000 based conversions, but just always use KB/MB/GB/TB/PB as the units. There is precedent for this, so it may not be surprising to clients. This is what SLE's current implementation does.
+
+**Pros:** No translation work for the client or Nimble team. Honors `lang` value.
+
+**Cons:** Potentially misleading units label for 1024-based values. Shows "byte" unit even for multiple bytes. Missing several translations for "byte". Unclear if missing translations for other units (KB, MB, etc.).
```
mollykreis, line 192: As mentioned above, I don't think the unit translations is an issue here, nor is the plural "bytes".

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+### Documentation
+
+This component will be documented via a new story in Storybook.
+
+---
+
+## Open Issues
+
+-   Is there a better name for this column type? E.g. `nimble-table-column-memory-size`.
+-   Should we support grouping?
```
mollykreis, line 233: I believe we decided that we should support grouping.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+-   What is our unit localization approach?
```
mollykreis, line 234: I think another open question related to unit localization is what the expected/acceptable behavior is from our POs. Some of the tradeoffs for localization have to do with showing `MB` rather than `MiB`, and we probably need someone to decide if that's an acceptable tradeoff to make.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
rajsite: > progress-bar-like column

 I could imagine that. 
image

And a text representation of progress
image

And both from the example given below:
image

Particularly for things like systems grid representing software being installed to devices with limited space. Could imagine that being a two segment progress bar where current used space is filled, how much this package adds, and total space is the whole bar. Or one segment where the whole bar is just the available space and segment is how much the package will use.

> If so maybe we should update those guidelines to describe that case more generally.

I think what that case is describing is that at the time it was expected that the same column might have multiple visualizations (text or icon+text). Instead of having separate columns for those visualizations. I dunno if that was a great choice and if we would actually implement that. It would probably be a new column element depending on the requirement (current requirements sound like they want custom messages per mapping, ie a warning status with a row specific message, so that would be a pretty different API and likely a different column). I think that was a bad degree of flexibility to leave in because I think it's very likely that different visualizations have different visualization specific api configuration that would be conflated together if configured on a single column.

Some examples based on this (explicit visuals):

- `nimble-table-column-file-size-text`
- `nimble-table-column-file-size-progress`
- `nimble-table-column-file-size-progress-text`

vs not explicit visuals:

- `nimble-table-column-file-size` (renders as text, but not clear)
- `nimble-table-column-file-size-progress` (renders as progress bar or text but not clear)

I think I lean towards the explicit set of:

- `nimble-table-column-file-size-text` (variant file-size)
- `nimble-table-column-file-size-progress` (variant file-size)
- `nimble-table-column-file-size-progress-text` (variant file-size and variant progress)
- `nimble-table-column-number-text` (variant number)
- `nimble-table-column-progress` (implicit variant number)
- `nimble-table-column-progress-text` (implicit variant number, variant progress)

With the thoughts:
The progress* columns probably have similar apis for configuring the data set for file size progress (configuring the row size and total size) distinct from text
The progress visual column probably has a distinct api for configuring the visual from text

And I might suggest the opposite for a docs clarification. Shoving or anticipating multiple visuals for the same column is probably not the right direction.

So I think a naming update could be:

- Always have the visual in the column name (mapping is out of compliance with the updated rule and should be mapping-text, but may be a low ROI to update or this might be the **best** time to fix it since most usage of nimble-table is under our control in sl-lib 🤷‍♂️)
- Add the concept of "implicit variants" where variants can be implicit if they are simple / primitive types / intuitive defaults for a given visual

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
jattasNI: One downside of option 2 is that we still don't have a good story for localizing Nimble strings in Blazor apps. That's been less noticeable for tooltips and accessible labels but might become more noticeable with visible column text.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
rajsite, line 192: nit / fyi add indent spacing so the pros and cons show up under the items in the list

```suggestion
1. Find translations of each of our unit labels ("byte", "bytes", "KB", "MB", "GB", "TB", "PB", "KiB", "MiB", "GiB", "TiB", "PiB") for a fixed set of languages we wish to support. Maintain a mapping of language codes (e.g. "fr", "de", "zh_CN") to arrays of those localized unit labels. Given a locale to use, look up its language subtag in our map. If not found, fall back to English.
   
   **Pros:** No work for clients.
   
   **Cons:** Only supports a fixed set of languages. Up to us to find accurate translations. Unsure if subtags other than the language (e.g. region) could be relevant to the translation.
   
2. Use Nimble label provider. Add label tokens for the units, and rely on clients to provide translations.
   
   **Pros:** Supports any language a client cares to provide translations for.
   
   **Cons:** Less convenient for clients. Does not honor the `lang` setting on the page or on `nimble-theme-provider`.
   
3. Use `Intl.NumberFormat` and drop support for units based on a factor of 1024.
   
   **Pros:** No translation work for the client or Nimble team. Honors `lang` value.
   
   **Cons:** No support for 1024-based units. Shows "byte" unit even for multiple bytes. Missing several translations for "byte". Unclear if missing translations for other units (KB, MB, etc.).
   
4. Use `Intl.NumberFormat` and support both 1024 or 1000 based conversions, but just always use KB/MB/GB/TB/PB as the units. There is precedent for this, so it may not be surprising to clients. This is what SLE's current implementation does.
   
   **Pros:** No translation work for the client or Nimble team. Honors `lang` value.
   
   **Cons:** Potentially misleading units label for 1024-based values. Shows "byte" unit even for multiple bytes. Missing several translations for "byte". Unclear if missing translations for other units (KB, MB, etc.).
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Cons:** Less convenient for clients. Does not honor the `lang` setting on the page or on `nimble-theme-provider`.
+
+3. Use `Intl.NumberFormat` and drop support for units based on a factor of 1024.
+
+**Pros:** No translation work for the client or Nimble team. Honors `lang` value.
+
+**Cons:** No support for 1024-based units. Shows "byte" unit even for multiple bytes. Missing several translations for "byte". Unclear if missing translations for other units (KB, MB, etc.).
+
+4. Use `Intl.NumberFormat` and support both 1024 or 1000 based conversions, but just always use KB/MB/GB/TB/PB as the units. There is precedent for this, so it may not be surprising to clients. This is what SLE's current implementation does.
```
rajsite, line 188: Good precedent? Seems like legal disputes over these formats isn't a great precedent to carry forward https://en.wikipedia.org/wiki/Binary_prefix#Legal_disputes

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
rajsite: Gah... I think I'm leaning towards option 1...

Option 2 pushes a fair bit of integration work on everyone who tries to use the column. How much benefit are they really getting from the column if they have to juggle and possibly mess-up or be inconsistent about digital vs decimal translations across languages. Seems extra unsatisfying work. I can imagine wanting to use a theme-provider to override labels for different increment / decrement buttons within a page, but I can't really imagine a scenario where that flexibility is good here.

And it's a bummer that the standards groups are pushing back on the EIC formats of mebibytes, etc from following these threads: https://github.com/tc39/ecma402/issues/32#issuecomment-468172074

I think we should support those for the accuracy.. I don't think anyone would actually be completely lost by us using KiB and MiB for digital. Think they'll figure it out and maybe learn something. Would be nice to be able to help build some consistency for units at a T&M company, maybe this can be our push for that :) (related wonder if the NI brand guidelines cover any of this)

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
rajsite: Devtools research on trying to be consistent: https://goo.gle/devtools-si

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
rajsite: Devtools research on trying to be consistent: https://goo.gle/devtools-si 

and the relevant xkcd lol
image

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
m-akinc: Ok, I will change to `nimble-table-column-file-size-text` and look at updating the naming guidelines.

> mapping is out of compliance with the updated rule and should be mapping-text

We don't have a column named "mapping" -- it's `nimble-table-column-enum-text`, so I think we're compliant.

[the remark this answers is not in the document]
m-akinc: Now renaming to `nimble-table-column-byte-text`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: Sounds like we're okay with up to one decimal place.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: This is not as straightforward as I thought, as it seems that 1024 is a better default conversion for file sizes, even though most places incorrectly use the decimal units (KB, etc).

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: Changed to `nimble-table-column-byte-text`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: This is great. You're right. Using `unitDisplay: 'long'` does the trick.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: You're right. I thought that those were different language subtags for traditional and simplified Chinese, but those represented both language and region (and someone had just used '_' instead of '-').

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
m-akinc: Yeah, I guess it's pretty inconsequential, so I'll remove it.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+Unit tests:
+
+-   standard component tests
+-   renders blank for invalid input values: `Inf`, `-Inf`, `NaN`, and non-number
+-   for an interesting range of values, value conversion results in correct values and units (using English `lang`)
+-   correct units used for both `unit-type`s
+-   standard cell text/group header text ellipsizing tests
+-   honors `lang` value and responds to changes to `lang` value
+-   uses English labels when unsupported locale is given
```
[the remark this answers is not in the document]
m-akinc: I'll make sure to update it if we decide on a different approach.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
```diff
@@ -0,0 +1,234 @@
...
+
+-   Converts the byte value to largest unit (e.g. KB, MB, GB) that results in a value of >=1
+-   Option to use 1024-based units (e.g. KiB) or 1000-based units (e.g. KB)
+-   Displays up to one decimal place
+-   Sorts/groups by numeric value (number of bytes) rather than display string
+-   Localized number and unit strings
+
+### Risks and Challenges
+
+The byte formatting pipe currently used in SLE converts using a factor of 1024, but uses unit labels KB/MB/GB. In order to avoid showing different values or units for the same file outside of an `sl-table`, we should modify the SLE pipe to be consistent with our implementation. If there are other clients that would have the same problem, they will just display the file sizes inconsistently.
```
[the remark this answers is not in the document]
m-akinc: It seems a little strange to be turning to your design system to get a function for file size formatting, but maybe? If we do, it should be exported from utilities and not from anything table-related. I feel like it might be more natural for clients to provide _us_ a formatting function (to override our default). I'm not sure it's worth providing the support, though.

change/@ni-nimble-components-df572d28-5858-4799-8684-fabcd595dfab.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Think there are open questions @fredvisser and @jattasNI were following up on:

1. PO requirements on if  both decimal vs binary display of byte sizes, do we need both or can we standardize on one?
2. If showing binary display, can we use the NIST / ISO / ANSI [binary prefixes](https://en.wikipedia.org/wiki/Binary_prefix), i.e. MiB, GiB and migrate all usages in SystemLink to be standardized (other pipes, etc) 
3. Do we have tech writing / ni.com / other guidance on decimal vs binary sizing we can align on to drive consistency across all applications?

Mert is OOO this week and with the open questions I think we are blocked on reviews for now so moving back to draft. Feel free to re-open if there are parts to review ATM.

m-akinc:
Closing. We're changing direction to extend the functionality of the number-text column instead of creating a separate column.

packages/nimble-components/src/table/specs/table-column-specs/table-column-file-size.md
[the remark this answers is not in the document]
rajsite: Another meme, this one from reddit (we did the full curve!):
image

The comments are interesting: https://www.reddit.com/r/ProgrammerHumor/comments/17r2bqx/aretheystupid/

Someone from Azure discussed there approach for defining column units in a table: https://www.reddit.com/r/ProgrammerHumor/comments/17r2bqx/aretheystupid/k8h2fw7/?context=10000

Direct link to Azure column handling docs: https://github.com/Azure/portaldocs/blob/main/portal-sdk/generated/declarative-assets.md#custom-columns

Issue #861, opened 2022-12-05
Table hierarchical data support


nate-ni:
## 😯 Problem to Solve
The table needs to support displaying a hierarchical data source.
From table spec: "Provide hierarchy via the data"

Excludes [leaf mode](https://github.com/ni/nimble/issues/1759) and [dynamic child loading](https://github.com/ni/nimble/issues/1758), which are tracked separately.

## 💁 Proposed Solution
 - [x] #890

## 🖼️ References/Prior Art
![Image](https://user-images.githubusercontent.com/109235103/200389498-b34805b6-07f3-4b07-a783-c0696d56e139.png)

jattasNI:
We'd really like to get a fix for #1162 before proceeding with this issue. Without it we'd have to put large TanStack workarounds into our implementation.

jattasNI:
A client using the old grid [mentioned](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/548655#1692975724) that they want larger data sizes for hierarchy than what we've claimed support for to date. The specs grid can contain up to 5000 parent nodes, each with up to 10 children, meaning 55000 rows of data.

atmgrifter00:
Rough prototype from innovation day: [https://github.com/ni/nimble/tree/table-pinned-columns](https://github.com/ni/nimble/tree/table-pinned-columns) ([Storybook](https://60e89457a987cf003efc0a5b-cohvfnnpop.chromatic.com/?path=/story/incubating-table-column-configuration--grouping)).

Things of note:
- Using 'sticky' for the frozen entities seems like the only viable implementation route (as opposed to something like separate containers for pinned and unpinned elements).
- Using display of `inline-grid` seems necessary to get sticky behavior to work properly (at least for containers that were using `grid` before).
- Offsetting header row using CSS `left` for scrolling purposes interferes with setting CSS `left` for making header content sticky (the latter of which seems necessary). Instead of setting the CSS `left` property, we can instead set the `scrollLeft` of the element containing all of the header content (requiring implementation in code-behind). This seems like it also might require its CSS `overflow` to be explicitly be set to `hidden`.
- Cell elements will likely require having their background explicitly set so that appropriate visual clipping occurs when a scrollable column is dragged passed a pinned column.
- Assumption is that all pinned columns will be rendered in order before unpinned columns (not implemented in prototype)

Visual things to finalize:
- Should there be any explicit visual indicator denoting the border between pinned and unpinned columns?
[...]

PR #1567, opened 2023-09-26
Data hierarchy hld

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #861


packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+### `Table` API
+
+```ts
+public Table() {
+    ...
+    // This attribute determines which field in the record specifies the id that is the
+    // parent row. If it is not set, the row has no parent.
+    @attr({ attribute: 'parent-id-field-name' })
+    public parentIdFieldName?: string;
```
mollykreis, line 32: I think we should include the validity checks that will be part of this new API. For example, if `parentIdFieldName` is set but `idFieldName` is not set, that should be invalid. Probably the same for `hasChildrenFieldName` being set without `parentIdFieldName` being set.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+    // This field would be an option for how we allow a user to specify the field
+    // name where they will provide a value indicating whether or not that row
+    // is a parent row. See the below discussion on lazy-loading. If it is not set,
+    // the row is presumed child-less (and thus will show no expand/collapse button).
+    @attr({ attribute: 'has-children-field-name' })
+    public hasChildrenFieldName?: string;
```
mollykreis, line 39: I'm wondering if we should make this name more explicitly tied to lazy-loaded children so that it's more obvious looking at the API that you don't need to set this for all hierarchy cases. Maybe something like `hasDelayedChildrenFieldName`? I could also be convinced that it isn't worth making the name longer and that it's better to keep it as-is.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+-   [Prototype with user data being hierarchical](https://github.com/ni/nimble/tree/data-hierarchy-prototype) with [Storybook](https://60e89457a987cf003efc0a5b-yncupvnoes.chromatic.com/?path=/story/incubating-table--table&args=data:LargeDataSet)
+
+## Implementation / Design
+
+Users will provide hierarchical data as a flat list (just as they do for non-hierarchical data), but will have to include a field in their data that is representative of the "parent Id" for that record of data.
+
+### `Table` API
+
+```ts
+public Table() {
```
mollykreis, line 27: I think we may also need a way for a client to define whether or not they want parent rows to be expanded by default or not. The alternative would be to auto expand if the children aren't lazy-loaded and not auto expand if the children are lazy-loaded, but I think it would likely be better to allow a client to explicitly specify their desired behavior.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+    // name where they will provide a value indicating whether or not that row
+    // is a parent row. See the below discussion on lazy-loading. If it is not set,
+    // the row is presumed child-less (and thus will show no expand/collapse button).
+    @attr({ attribute: 'has-children-field-name' })
+    public hasChildrenFieldName?: string;
+
+    // This attribute will control the behavior of what happens when a user selects a parent
+    // row.
+    @attr({ attribute: 'parent-selects-children'})
+    public parentSelectsChildren?: boolean;
```
mollykreis, line 44: As another point of reference, JQX calls this property `selectionByHierarchy`.

mollykreis, line 44: We should be sure to think through what the behavior is with all combinations of `parentSelectsChildren` + `selectionMode`. I think the particularly unclear case might be single-selection + parentSelectsChildren=true. My initial thought is that parent rows wouldn't be selectable with single-selection, but we should make sure that makes sense and  probably include a short section in this HLD to describe the expected behavior in each case.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+The `Table` will also provide a `row-expand-toggle` event for when a row is expanded/collapsed that will provide details to the client including which row was expanded/collapsed, and what its new state is:
+
+```ts
+interface TableRowExpandedEventDetail {
+    newState: boolean;
+    recordId?: string;
```
mollykreis, line 53: As mentioned in another comment, I think we need to have validation that we have recordIds when dealing with hierarchy. Therefore, I think the type of recordId should just be `string` (not optional).

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+    @attr({ attribute: 'parent-selects-children'})
+    public parentSelectsChildren?: boolean;
+}
+```
+
+The `Table` will also provide a `row-expand-toggle` event for when a row is expanded/collapsed that will provide details to the client including which row was expanded/collapsed, and what its new state is:
+
+```ts
+interface TableRowExpandedEventDetail {
+    newState: boolean;
```
mollykreis, line 52: We've been including `newState` and `oldState` in all our `toggle` events, so we should do that here as well.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Translating flat list to Tanstack-understandable hierarchy
+
+Tanstack provides APIs for us to implement that allow it to return the rows in a hierarchical fashion where child rows are provided as a property on a row called `subRows`. For this to work as expected it is required that a flat list of data (that contains implicit hierarchy) be transformed into a hierarchical data structure.
+
+A third party library called [`performant-array-to-tree`](https://www.npmjs.com/package/performant-array-to-tree) offers an easy, and as the package says, performant means of doing this (in O(n) time). It comes with an MIT license, and is apparently fully tested, so it seems like it would be suitable to use for this purpose.
+
+### Group rows with data hierarchy
+
+Tanstack allows limited support of data modeling where there is both grouping and parent-child row relationships. Essentially, when there is grouping present, only rows at the top-level of a hierarchical set of data will be grouped. All further child rows under the top-level parent row will not be grouped.
```
mollykreis, line 82: I'm wondering if we can clarify the last sentence of this paragraph. Maybe something like _Child rows will continue to be shown under their parent row rather than being grouped based on their own data._

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+-   It is up to the user to manage whether or not the children of a row has already been loaded in order to avoid recreation of their data and calling `setData` on the `Table` unnecessarily.
+
+### Translating flat list to Tanstack-understandable hierarchy
+
+Tanstack provides APIs for us to implement that allow it to return the rows in a hierarchical fashion where child rows are provided as a property on a row called `subRows`. For this to work as expected it is required that a flat list of data (that contains implicit hierarchy) be transformed into a hierarchical data structure.
+
+A third party library called [`performant-array-to-tree`](https://www.npmjs.com/package/performant-array-to-tree) offers an easy, and as the package says, performant means of doing this (in O(n) time). It comes with an MIT license, and is apparently fully tested, so it seems like it would be suitable to use for this purpose.
+
+### Group rows with data hierarchy
```
mollykreis, line 80: Is there anything that's worth pointing out about sorting with data hierarchy?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+I see no reason to explicitly disable this behavior. One behavior we will need to ensure, however, is that the count value we display in the group row, is _only_ the number of immediate children in the group row. It would be odd for the number to change just because a row was expanded (such as in the lazy loading case).
+
+#### Managing expanded state
+
+Currently, the `Table` defaults to expanding all rows by setting its Tanstack expanded state to the singleton `true`. This means that all group rows will be expanded by default. The Nimble `Table` also provides an implementation of `getIsRowExpanded` which is consumed by `TanStack` that will ultimately result in the per-row state of whether the row is expanded. It is in this implementation, where even if the Tanstack state is set to `true`, we can mark all parent rows, by default, as collapsed, which is required for the lazy-loading scenario.
```
mollykreis, line 88: I mentioned this in a section above, but I think we're going to need configuration for "expand parent rows automatically". There are definitely cases where we expand the parent rows by default, and I don't think we'll want to change that behavior.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+After a user clicks on a parent row to expand it, and that parent does not have children yet, there could be a noticeable delay between when the user clicked the expand button, and when the children show up in the table. This could happen for a variety of reasons, but commonly might occur due to a slow network request.
+
+It would be helpful for the `Table` to visually represent the state that it is currently waiting on data to be loaded from the client. It can accomplish this fairly easily using existing state with no additional APIs or new components.
+
+Prototype visual:
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+This ultimately may put the burden on the client to ensure that the `Table` is updated as needed to get rid of any displayed progress indicator, including in scenarios where the expansion of a parent row failed to load any children (possibly due to some client-side error). The `Table` may only guarantee that the progress indicator is shown when a parent row is expanded and it currently has no children, and that it will be removed once children are present.
```
mollykreis, line 104: I think it would be good to harden the proposal here (or maybe just the language associated with it). Specifically, I think we should state when the table _will_ (not _may_) put a loading affordance for the row. Then, maybe you could talk about how in the expected workflow the loading spinner would be removed automatically and how a client could remove it in some unexpected cases (like a failed network request or lazy-loading the children determined there aren't actually children like was initially thought).

Sorry if this seems nit-picky, but it's a little unclear to me right now whether this is actually the proposal or just one possible solution to show that we have options.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+    public hasChildrenFieldName?: string;
+
+    // This attribute will control the behavior of what happens when a user selects a parent
+    // row.
+    @attr({ attribute: 'parent-selects-children'})
+    public parentSelectsChildren?: boolean;
+}
+```
+
+The `Table` will also provide a `row-expand-toggle` event for when a row is expanded/collapsed that will provide details to the client including which row was expanded/collapsed, and what its new state is:
```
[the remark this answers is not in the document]
atmgrifter00: As far as I know I thought the `beforetoggle` events were created out of necessity. At the moment, I'm not sure what the necessity would be for such an event to allow users to achieve what they need. Maybe we should wait to create such an event until a need is made clear?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I'll leave this thread open for discussion, but after chatting with @leslieab, the set of valid behaviors does _not_ include one where `selectionMode` is `single` _and_ a parent can select its children. So a possible alternative to this API is one where we provide a new option for `selectionMode` that allows multi-select but does _not_ allow a parent to select its children. Perhaps `multiWithNoAutoChildSelect`? Otherwise, we will have an API that allows for an invalid configuration, that we would likely just have to document what the expected behavior would be.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I'll leave this thread open, and ask for feedback from others.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I'll also leave this thread open. I think what we want to decide on whether an API to accomplish what Molly suggests should be present in this initial HLD or not.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+5. The user must then create a set of data with rows where they provide a field with the name specified by the `parendIdFieldName` attribute that has the value of the `recordId` value supplied in the event details.
+6. The user then sets the data on the `Table` again with the `setData` method.
+
+Notes:
+
+-   It is up to the user to manage whether or not the children of a row has already been loaded in order to avoid recreation of their data and calling `setData` on the `Table` unnecessarily.
+
+### Translating flat list to Tanstack-understandable hierarchy
+
+Tanstack provides APIs for us to implement that allow it to return the rows in a hierarchical fashion where child rows are provided as a property on a row called `subRows`. For this to work as expected it is required that a flat list of data (that contains implicit hierarchy) be transformed into a hierarchical data structure.
```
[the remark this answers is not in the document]
atmgrifter00: I've added a note to this section suggesting that we would likely want to create the data using a fairly specific prefix.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I've added a brief section on sorting.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I've made the wording a bit stronger and added some workflows.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,121 @@
...
+
+A third party library called [`performant-array-to-tree`](https://www.npmjs.com/package/performant-array-to-tree) offers an easy, and as the package says, performant means of doing this (in O(n) time). It comes with an MIT license, and is apparently fully tested, so it seems like it would be suitable to use for this purpose.
+
+### Group rows with data hierarchy
+
+Tanstack allows limited support of data modeling where there is both grouping and parent-child row relationships. Essentially, when there is grouping present, only rows at the top-level of a hierarchical set of data will be grouped. All further child rows under the top-level parent row will not be grouped.
+
+I see no reason to explicitly disable this behavior. One behavior we will need to ensure, however, is that the count value we display in the group row, is _only_ the number of immediate children in the group row. It would be odd for the number to change just because a row was expanded (such as in the lazy loading case).
```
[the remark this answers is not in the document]
atmgrifter00: I've added a section on Selection that proposes a behavior.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I'll leave this thread open as well, and update this section as needed pending any decisions we make.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
jattasNI: I had a similar thought before seeing this thread. My first idea was to add a new selection mode: `selectionMode = 'none' | 'single' | 'multiple' | 'multiple-with-children'` or something like that.

Another idea: `selectionMode` is ignored for hierarchical tables and we add a new attribute `hierarchySelectionMode = 'none' | 'single' | 'multiple' | 'multiple-with-children' | ...`? 

I could imagine wanting other modes in future like `'single-leaf' | 'multiple-leaf'`.

I think all of these options open the door to invalid combinations and we just need to live with that and rely on our existing patterns to report it or ignore it. With the `parentSelectsChildren` attribute it'd be invalid/ignored if it's set when `selectionMode = 'none'` or `selectionMode = 'single'`. With the new selection modes it'd be invalid if the table isn't hierarchical but a client sets a hierarchical selection mode.

My initial vote would be to add new mode(s) to the existing `selectionMode` property.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,158 @@
...
+    @attr({ attribute: 'has-children-field-name' })
+    public hasChildrenFieldName?: string;
+
+    // This attribute will control the behavior of what happens when a user selects a parent
+    // row.
+    @attr({ attribute: 'parent-selects-children'})
+    public parentSelectsChildren?: boolean;
+
+    // The set of rows the user would like to toggle the expand/collapse state of
+    public async expandCollapseRows(rowIds: string[]): Promise<void> {
```
jattasNI, line 47: Would this toggle the state of all of the provided `rowIds`, possibly resulting in some of them becoming expanded and others collapsed? If so, that seems somewhat messy to me because calling this requires knowing the state of every row first. I wonder if it might be more convenient to have one of these:
1. `expandRows()` and `collapseRows()` which don't toggle but just set the state on the rows you provide, leaving other rows unchanged
2. `collapseAllAndExpandRows()` where you provide the entire set of expanded rows and everything else ends up collapsed

I'm struggling to weigh the options here. Do you have ideas of what these would be used for?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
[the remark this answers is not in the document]
jattasNI: Just curious, what are the proposed use cases for the `row-expand-toggle` event? Just for dynamically loading children?

Part of the reason I ask is I wonder if we'd also want a way to query the current expand/collapse state of every row, outside of the moment when an event is fired. But I can't think of a use case for that so I'm ok to add it later/never unless we see an obvious need now.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
jattasNI: I think I like `hasDelayedChildrenFieldName` better than `hasChildrenFieldName` for the reasons Molly mentioned.

My only other idea is something like `forceExpandableFieldName` but I don't like that any more or less.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: This may be pedantic, but I wouldn't consider a `selectionMode` set to something that semantically is aligned with hierarchical data invalid when the table currently isn't displaying hierarchical data. It would simply be not used to its full extent. Whereas, having the two separate APIs of `parentSelectsChildren` (or whatever name we want for that) along with `selectionMode` _does_ enable an invalid state.

Ultimately, I think we are in violent agreement that we should have a single API, and just add a new mode to `selectionMode`.

atmgrifter00: I guess the question front of us is do we want a mode that allows multiple selection _AND_ where selecting a parent doesn't auto-select children. If so, that might be semantically hard to capture in a new mode (I mean we can, but it might be fairly verbose).

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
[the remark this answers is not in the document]
atmgrifter00: Yes, at the moment, the primary use case in mind for this event is to enable lazy loading of children.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: > I guess the question front of us is do we want a mode that allows multiple selection AND where selecting a parent doesn't auto-select children.

Why wouldn't we want that mode? It seems reasonable to me to be able to select multiple items in a hierarchical table and have it select each row individually regardless of where they exist in the hierarchy. Am I missing something? Wouldn't that just be the `'multiple'` mode? And then `'multiple-with-children'` has the different behavior when selecting a row with children?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: I agree that it seems like a questionable API for a "toggle row expanded" function when I would think a client would want to be explicitly expanding or collapsing (rather than toggling) a row. I also think "expand" might be an interesting case -- would we expand all ancestors and the row specified? Maybe that's rationale for just having a "collapse" function for now (although maybe it's too inconsistent to have collapse and not expand).

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I think you are right. Not sure how I tripped myself up in my head.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+In addition to just supporting data hierarchy with a fully provided set of data, we also need to support the scenario where a row is presented as a parent, but none of its children are currently represented in the data. Upon expanding the row, the client is then expected to load all of the data for that parent and provide it to the table (via the standard `setData` method).
+
+## Links To Relevant Work Items and Reference Material
+
+-   [#890: Design for hierarchical data support in tables](https://github.com/ni/nimble/issues/890)
+-   [#861: Table hierarchical data support](https://github.com/ni/nimble/issues/861)
+-   [Prototype using flat list that includes a parentId](https://github.com/ni/nimble/tree/data-hierarchy-flat-list-prototype) with [Storybook](https://60e89457a987cf003efc0a5b-qzwoshcidz.chromatic.com/?path=/story/incubating-table--table&args=data:LargeDataSet)
+-   [Prototype with user data being hierarchical](https://github.com/ni/nimble/tree/data-hierarchy-prototype) with [Storybook](https://60e89457a987cf003efc0a5b-yncupvnoes.chromatic.com/?path=/story/incubating-table--table&args=data:LargeDataSet)
+
+## Implementation / Design
```
jattasNI, line 20: It would be good to capture our plans for interactions:
1. collapse all button is updated to collapse hierarchical rows in addition to group rows
2. what is the mouse interaction to expand/collapse a row: click anywhere on the row or only on certain parts of the row? If you can click anywhere, would a selectable row still become selected when you click on it?
3. what are the expected keyboard interactions for expand/collapse? (or add this as a decision / open issue on #1506)

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+
+Tanstack will sort children within each parent row by the same column that the parent is being sorted by. So, it will not be possible to sort children by a different column value than the one its parent is sorted by.
+
+### Selection
+
+One scenario worth calling out is the situation where a parent row is present with no children, and the user has a selection mode set such that selecting a parent should select its children. In such a scenario where the parent is selected, the children will _not_ be selected upon opening the parent. This is the same for when a group row is expanded and a user decided to use the toggle event to load even more data into the table that would introduce more rows under the group. The `Table` will make no attempt to decipher when and when not to select rows that have been introduced through another call to `setData`.
+
+### ARIA
+
+A parent row would have the same ARIA expectations of any child row, with the additional need to supply the `aria-expanded` attribute when it is expanded.
```
jattasNI, line 155: Some other things to consider:
1. accessibility attributes (role/states/label/title) of the progress indicator when expanding a lazy loaded row
2. accessibility attributes of the expand/collapse icon toggle button (for groups we currently show a tooltip; would this need one too?)

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: The driving force behind this API was the need for a user to easily collapse a row (without explicit interaction) when they hit an error case when wanting to populate a parent row with children (they expanded a parent row, and now the 'row loading' indicator is present which they want to remove).

From the suggestions, I think I'm happiest with the `expandRows()` and `collapseRows()` APIs. Can we provide an additional parameter that controls whether the action should apply to all children:

```ts
public expandRows(rowIds: string[], expandChildren?: boolean): Promise<void>{}
```

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+    ...
+    // This attribute determines which field in the record specifies the id that is the
+    // parent row. If it is not set, the row has no parent.
+    @attr({ attribute: 'parent-id-field-name' })
+    public parentIdFieldName?: string;
+
+    // This field would be an option for how we allow a user to specify the field
+    // name where they will provide a value indicating whether or not that row
+    // is a parent row. See the below discussion on lazy-loading. If it is not set,
+    // the row is presumed child-less (and thus will show no expand/collapse button).
```
rajsite, line 37: If not set and there are rows pointing to this row as a parent then is it still presumed childless?

I feel like generally / commonly the lazy loading is an edge case. Would be nice to not force everyone to deal with toggling an extra flag if it's not needed.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+    @attr({ attribute: 'has-children-field-name' })
```
rajsite, line 38: Should we follow a model like the row selection api? It's a similar kind of concept, marking certain rows a certain way out-of-band with the rest of the actual dataset.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
rajsite, line 37: I can imagine a lazy loading scenario where we know the size of things in the hierarchy but don't want to load the full data / leaf nodes. i.e. I know I have 40 children. Is that a scenario we could expect today / should be able to accommodate?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+4. User handles event that row was expanded
+5. User sets data on table that has children for the row that was expanded
+6. "Row loading" indicator is removed and child rows are now displayed
+
+Error workflow:
+
+1. Same as steps 1-4 above
+2. Error occurs retrieving data for child rows
+3. User calls `expandCollapseRows(...)` on table instance passing parent row id that was previously expanded.
+4. Row is collapsed and "row loading" indicator is no longer displayed.
```
rajsite, line 143: Have we validated this workflow with UX? I imagine it could be surprising as written for the row to just collapse itself without any indicator / message / retry button.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+This ultimately will put the burden on the client to ensure that the `Table` is updated as needed to get rid of any displayed progress indicator, including in scenarios where the expansion of a parent row failed to load any children (possibly due to some client-side error). The `Table` will only guarantee that the progress indicator is shown when a parent row is expanded and it currently has no children, and that it will be removed once children are present.
+
+Expected user workflow:
+
+1. User loads data into table that specifies some rows as parents, but has no rows that indicate that row as a parent.
+2. User clicks on a parent row
+3. Row expands showing "row loading" indicator
```
rajsite, line 133: Guess that means multiple rows can be be expanded and show loading at the same time (and potentially fail to load for the same reason)

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+4. User handles event that row was expanded
+5. User sets data on table that has children for the row that was expanded
+6. "Row loading" indicator is removed and child rows are now displayed
+
+Error workflow:
```
rajsite, line 138: Guess we also need to document a cancellation workflow that apps are expected to implement. i.e. if the row is collapsed maybe cancel the request? Or maybe don't cancel it?
Seems like it would be weird to pop open after it was collapsed once data loads.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: So, the idea here is that the table doesn't do anything in this scenario automatically (i.e. the row would _not_ collapse itself). The user is responsible for updating the display state of the row to be collapsed if they recognized that an error occurred that would prevent the table from updating with child data. Perhaps I'm not understanding the situation you're describing.

atmgrifter00: Good point. I'll talk this through with @leslieab.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: Yeah, I already had removed that comment locally.

atmgrifter00: I guess I'm not fully understanding this scenario. Why wouldn't the proposed API handle this?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: Let me see if I understand the proposed workflow. Would it be something like this:
1. Client calls `setData` on table where some rows will eventually be parents (table has no attributes for flagging a row that it will be a parent, and thus rows won't show an expand/collapse button at this point).
2. Client calls something like `setLazyLoadedParents(rowIds)`. This will then mark the provided rows as parents, which will result in them showing expand/collapse buttons.
3. Same workflow described already once a user expands a parent row.

Pros:
- Clients no longer have to provide an extra field in their data to represent this state, which is pretty nice since that state only has transient value.

Cons:
- Possible UI oddity in that rows will have the expand/collapse button "pop in" to existence.

If I have it right, I think I might agree with you in that I would prefer the UI hiccup to making clients provide an extra field in their data that most rows don't need.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
rajsite: Was hypothesizing about presenting that to user in a way they can see the count of children before expanding, or for having selection counts on lazy loaded sections be accurate

atmgrifter00: My understanding is that parents would not show child counts. Additionally, group rows that had hierarchical data under them would only show counts for top-level entities. @leslieab, can you provide checks on me here? If I am right, does this satisfy your concern @rajsite?

leslieab: Yes, that's correct. We can revisit it, but that's the decision we landed on.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,168 @@
...
+### `TableRowSelectionMode` API
+
+We need the ability for a client to configure what happens when a user selects a parent row, where it could either only select the parent or would select both the parent and all of its children.
+
+```ts
+export const TableRowSelectionMode = {
+    none: undefined,
+    single: 'single',
+    multiple: 'multiple',
+    multipleWithChildren: 'multipleWithChildren' // NEW MODE!!!
```
rajsite, line 57: Some questions about this mode:
- What should happen when the user checks a row that has lazy children and then the user expands the row and the lazy children load. In the multipleWithChildren mode wouldn't the user expect those children to be checked?
- Are there examples where the operations on the parent are distinct from selection of parent + children. i.e. in a filesystem analogy selecting a parent folder and deleting it, moving it, etc, moves all the children as well without "selection" of the children.
   
   It seems like useful operations on the parent vs parent + children would be identical (the operations possible on a selection of folders + files or on states + cities). The distinguishing case would be selecting all the children, but not the parent.
   
   The question behind the question is do we actually need the multipleWithChildren mode as a distinction from the multiple mode? If so can we capture the use cases and behavior more precisely for this mode in the spec.
- Will it be common that a user just wants an easy way to select all children of an item in the hierarchy? i.e. they don't to select parent + children but just the children of a parent. If that's a common workflow, how could a user unselect the parent without unselecting all the children?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
rajsite: After discussing offline I think my questions are resolved here. But there were follow-on questions around selection mode I captured in a separate thread. I think this thread is resolved.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+#### Leaf-only mode
+
+There are existing SLE grid variants (namely the Tags grid) that provide a type of hierarchical display that doesn't align exactly with either the proposed 'data hierarchy' outlined in this HLD, nor the grouping capability. Instead, while the parent rows may possibly be presented as a normal row (containing values in some subset of columns that the children have values for), they will not be able to be acted upon in the way child rows can, such as allowing an action menu to be defined on them, or being selected individually.
+
+It is expected that this mode will involve creating a new API on the table in order to fully enable, but it should be able to use the other APIs defined in this HLD to support it, and leverage capabilities such as lazy-loading.
+
+Here are a few of the considerations that were made with respect to this mode:
+
+-   grouping should group on leaf items, not parents (could be cumbersome to accomplish with Tanstack)
+-   action menu only on leafs
```
rajsite, line 177: I *think* the following issue may be leaf mode specific: https://github.com/ni/nimble/issues/1609
I'm asking them to confirm which table they are working on to see if they would be using hierarchical leaf mode or if they have some other requirement.

mollykreis: That's an interesting case because I believe they want an action menu on a different column based on whether or not a given row is a leaf row. So, I don't think leaf mode by itself would solve their problem.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+-   single selection mode can only select leafs
+-   selection state (when set to `multiple`) of parents (if selected) should become indetermine when new rows loaded (e.g. via lazy loading)
+-   selection counts should ignore parents
+-   (maybe?) leaf-mode + multi-selection + lazy loading is considered an invalid confiuguration. This could mean that the Table API of `forceExpandCollapseFieldName` should be more semantically associated with lazy loading.
+    -   (alternative?): Could we instead just hide the selection checkbox for parent rows that have no children?
+
+#### Auto-expand-parent API
+
+Currently, the proposed default behavior of parent rows is that they will display as collapsed initially (unlike groups). This is primarily done to support the lazy-loading case where we never want a row that has no children to default to being expanded. We could add an API that allows a user to determine what the default state of parents should be, which would ultimately not apply to rows that have no children (i.e. rows with children could be auto expanded, but any parent rows without children would remain collapsed).
```
mollykreis, line 186: I don't remember seeing the rationale for this being "future work" (but it's quite possible I just missed that discussion). I'm fairly certain this is something we need, so why not include it as part of the initial HLD? Are there open questions/concerns about this functionality that we are trying not to solve right now?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+public Table() {
+    ...
+    // This attribute determines which field in the record specifies the id that is the
+    // parent row. If it is not set, the data is presented without hierarchy as a flat list.
+    @attr({ attribute: 'parent-id-field-name' })
+    public parentIdFieldName?: string;
+
+    // This field would be an option for how we allow a user to specify the field
+    // name where they will provide a value indicating whether or not that row
+    // should always contain an expand-collapse button.
```
fredvisser, line 36: I had to read this several times to understand – how about…
```suggestion
    // This field lets the user choose a column name that determines whether 
    // each row has an expand-collapse button or not.
```

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+### Validation
+
+Tha table will be invalid if the user has set the `parentIdFieldName` attribute but not the `idFieldName`. Additionally, the table will be invalid if the user has set the `forceExpandCollapseFieldName` attribute but not the `parentIdFieldName`.
+
+### Lazy Loading
+
+The APIs noted above will enable the client to lazy load data into the `Table`. This will essentially be accomplished with the following steps:
+
+1. Providing a field in a record of the table data that indicates whether that row of data is intended to be a parent row. Records that are intended to be parents must set the value of the field, whose name is specified by the `forceExpandCollapseFieldName` attribute, to `true`.
+2. After providing the current data to the `Table` via the `setData` method, all rows that have a value of `true` in the field specified by the `forceExpandCollapseFieldName` attribute will display and expand/collapse button.
```
jattasNI, line 85:
```suggestion
2. After providing the current data to the `Table` via the `setData` method, all rows that have a value of `true` in the field specified by the `forceExpandCollapseFieldName` attribute will display an expand/collapse button.
```

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. Same as steps 1-4 above
+2. Error occurs retrieving data for child rows
+3. User calls `collapseRows(...)` on table instance passing parent row id that was previously expanded.
+4. Row is collapsed and "row loading" indicator is no longer displayed.
+5. Client should also provide appropriate error messaging within the application
+
+### Sorting
+
+Tanstack will sort children within each parent row by the same column that the parent is being sorted by. So, it will not be possible to sort children by a different column value than the one its parent is sorted by.
```
jattasNI, line 149: To be very explicit about it: sorting has lower precedence than hierarchy. So sorting will only affect the order of rows at each level of hierarchy and only within each parent.

A consequence of this is that sorting won't necessarily show you the largest value at the top/bottom of the table.

I think all of that is sane, but it might be good to mention it explicitly.

jattasNI (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+2. After providing the current data to the `Table` via the `setData` method, all rows that have a value of `true` in the field specified by the `forceExpandCollapseFieldName` attribute will display and expand/collapse button.
+3. Users must register a handler for the `row-expand-toggle` event on the `Table` instance, and will receive that event upon clicking the expand/collapse button.
+4. The details of the handled event will include the id for the row that was expanded.
+5. The user must then create a set of data with rows where they provide a field with the name specified by the `parendIdFieldName` attribute that has the value of the `recordId` value supplied in the event details.
+6. The user then sets the data on the `Table` again with the `setData` method.
+
+_Note: It is up to the user to manage whether or not the children of a row has already been loaded in order to avoid recreation of their data and calling `setData` on the `Table` unnecessarily._
```
fredvisser, line 91:
```suggestion
_The client is responsible for checking if the row’s children have already been loaded. This can prevent unnecessary data recreation and `setData` calls on the `Table`.._
```

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I've added an open question with respect to the Steps grid, but also calling it 'Resolved' stating that the Nimble table is not going to add features for what appears to be a sub-optimal UX.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: I've now added to this to the proposed API section.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Translating flat list to Tanstack-understandable hierarchy
+
+Tanstack provides APIs for us to implement that allow it to return the rows in a hierarchical fashion where child rows are provided as a property on a row called `subRows`. For this to work as expected it is required that a flat list of data (that contains implicit hierarchy) be transformed into a hierarchical data structure.
```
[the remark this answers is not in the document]
atmgrifter00: I've reworded this section to be less confusing.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+
+The following are various expected mouse and keyboard interactions related to parent rows:
+
+1. The collapse-all button in the header will also collapse parent rows.
+2. Clicking on a parent row (not on the expand/collapse button) will select the parent row. If the table `selectionMode` is set to `none`, then if a user clicks on a parent row (again _not_ on the expand/collapse button) nothing occurs.
+3. Keyboard interactions are being defined as part of the [Table Keyboard Navigation HLD](https://github.com/ni/nimble/pull/1506).
+
+### Validation
+
+Tha table will be invalid if the user has set the `parentIdFieldName` attribute but not the `idFieldName`. Additionally, the table will be invalid if the user has set the `forceExpandCollapseFieldName` attribute but not the `parentIdFieldName`.
```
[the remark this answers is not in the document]
atmgrifter00: Not sure. At the moment I feel somewhat indifferent to validating in this situation. Seems like, at worst, we will simply represent the child rows as parents themselves (not parented under anything).

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
atmgrifter00: Currently using `forceExpandableFieldName`. We can always iterate on the name in the implementation PR, as long as its purpose is well understood.

fredvisser (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,199 @@
...
+
+    // This field would be an option for how we allow a user to specify the field
+    // name where they will provide a value indicating whether or not that row
+    // should always contain an expand-collapse button.
+    @attr({ attribute: 'force-expand-collapse-field-name' })
+    public forceExpandCollapseFieldName?: string;
+
+    // The set of rows the user would like to expand. Pass 'true' for `expandChildren` if all
+    // children rows parented under any specified in rowIds should also be expanded.
+    public async expandRows(rowIds: string[], expandChildren?: boolean): Promise<void> {
```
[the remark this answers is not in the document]
mollykreis: My vote would be to remove this until we have actual use cases so that we don't end up with a hard-to-maintain API that never served a real purpose.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,228 @@
...
+
+It would be helpful for the `Table` to visually represent the state that it is currently waiting on data to be loaded from the client. It can accomplish this fairly easily using existing state with no additional APIs or new components.
+
+Prototype visual:
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+This ultimately will put the burden on the client to ensure that the `Table` is updated as needed to get rid of any displayed progress indicator, including in scenarios where the expansion of a parent row failed to load any children (possibly due to some client-side error). The `Table` will only guarantee that the progress indicator is shown when a parent row is expanded and it currently has no children, and that it will be removed once children are present.
+
+Expected user workflow:
```
mollykreis, line 154: nit: Can you update these two "workflow" sections to appropriately distinguish "user" vs "client"?

mollykreis, line 154: It doesn't look like these workflows reflect the new "row state" APIs that you've proposed.

atmgrifter00: Updated.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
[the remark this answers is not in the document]
atmgrifter00: I've now removed all APIs related to updating row state (including expand/collapse), and moved the API concerns to the Open Questions section.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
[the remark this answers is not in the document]
atmgrifter00: After chatting off-line, it became clear that we want to handle the cases where a client provides the table improperly configured data related to the parentId field, which we can catch with the 3rd party library that converts the flat list to a hierarchical form. Notably, if the client either provides circular parentId relationships between records, or if a record declares a parentId, and no record contains that id. In either case we will mark the table as invalid.

mollykreis (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,201 @@
...
+    // Note that for parent rows that have no children yet, they will always default to
+    // collapsed, regardless of the state of this attribute.
+    @attr({ attribute: 'auto-expand-parents' })
+    public autoExpandParents?: boolean;
+}
+```
+
+### `TableRowExpandToggleEventDetail` API
+
+The `Table` will also provide a `row-expand-toggle` event for when a row is expanded/collapsed that will provide details to the client including which row was expanded/collapsed, and what its new state is:
```
rajsite, line 50: based on the other suggestions `row-expansion-toggle` and `TableRowExpansionToggleEventDetail`?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -0,0 +1,201 @@
...
+
+1. Providing a field in a record of the table data that indicates whether that row of data is intended to be a parent row. Records that are intended to be parents must set the value of the field, whose name is specified by the `forceExpandCollapseFieldName` attribute, to `true`.
+2. After providing the current data to the `Table` via the `setData` method, all rows that have a value of `true` in the field specified by the `forceExpandCollapseFieldName` attribute will display an expand/collapse button.
+3. Clients must register a handler for the `row-expand-toggle` event on the `Table` instance, and will receive that event upon clicking the expand/collapse button.
+4. The details of the handled event will include the id for the row that was expanded.
+5. The client should first call `setRowState()` in their event handler code flagging the relevant rows to be lazy loading (this will cause the lazy loading indicator to appear).
+6. The client must then add the child records (with their `parentIdFieldName` value set to the parent's recordId) to the data set and call `setData()` on the Table.
+7. The client then sets the data on the `Table` again with the `setData` method.
```
rajsite, line 86: Accidental duplicate I assume?

rajsite (approved):
Approved with standup discussion decisions

PR #1574, opened 2023-09-27
Table to stable 🎉

rajsite:
# Pull Request

## 🤨 Rationale

Switch the table documentation to specifying that it is stable.

## 👩‍💻 Implementation

- Removed the incubating warning from stories
- Moved the stories to `/Incubating`
- Updated the component status table (also snuck in some wafer map corrections)
- Made sure each table story exposes either `autodocs` or `mdx` docs
- Got clever and changed the `Table Column - ` prefix to `Table Column:` so they come after `Table Column Configuration` alphabetically as `:` is before `space` in ASCII. Doesn't show up in the URL so think it's fine.

## 🧪 Testing

Manually looked through the stories.

## ✅ Checklist


jattasNI (approved):
🪩🪅🍾🎈

PR #1577, opened 2023-09-29
Vertical scale and test mocks updates

munteannatan:
# Pull Request

## 🤨 Rationale

The HTML canvas has its origin in the [top-left corner](https://www.w3schools.com/graphics/canvas_coordinates.asp).
When rendering the dies, we need to take this into consideration and flip the vertical scale so that the origin location will be respected.

Another improvement is the wafer mock utilities for each module.
## 👩‍💻 Implementation

switched the ends of the scale range and updated tests.
replaced individual mocks with a utility method

## 🧪 Testing

tests are passing locally

## ✅ Checklist


atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/modules/computations.ts
```diff
@@ -249,13 +249,15 @@ export class Computations {
             .paddingOuter(0)
             .align(0)
             .round(false);
+        // html canvas has top-left origin https://www.w3schools.com/graphics/canvas_coordinates.asp
```
rajsite, line 252: nit/pref, prefer mdn docs

```suggestion
        // html canvas has top-left origin https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Drawing_shapes#the_grid
```

packages/nimble-components/src/wafer-map/modules/computations.ts
```diff
@@ -264,15 +266,17 @@ export class Computations {
         containerHeight: number
     ): ScaleQuantile<number, number> {
         const scale = scaleQuantile().domain([0, containerHeight]);
+        // html canvas has top-left origin https://www.w3schools.com/graphics/canvas_coordinates.asp
```
rajsite, line 269:
```suggestion
        // html canvas has top-left origin https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Drawing_shapes#the_grid
```

rajsite (approved):

jattasNI:
Note: there was a spurious diff in the label provider chromatic test. It was comparing against an old baseline from before we added the latest 3 labels. I've seen this in other PRs too so I approved it to unblock this.

zszilagy (approved):

Issue #1569, opened 2023-09-26
Elapsed time table column


m-akinc:
## 😯 Problem to Solve

SLE apps need a column that displays elapsed time.

See AzDO feature: https://ni.visualstudio.com/DevCentral/_workitems/edit/2526116

## 📋 Tasks

- [x] [nimble-duration-text-column spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md)
- [x] nimble-duration-text-column custom element
- [x] nimble-duration-text-column unit tests
- [x] nimble-duration-text-column Storybook tests
- [x] nimble-duration-text-column Storybook docs
- [x] nimble-duration-text-column Angular wrapper
- [x] nimble-duration-text-column Blazor wrapper
- [x] Add to Angular example app
- [x] Add to Blazor example app

msmithNI:
#1619 added Blazor/Angular integration, and added the duration column to the all-components bundle.

Issue #783, opened 2022-10-17
Theme-aware tokens for graphs


jattasNI:
## 😯 Problem to Solve

The PDV app in SLE is going to add a Plotly graph to their app. It would be good for Nimble to help them style it with theme-aware colors/fonts/spacing/line thicknesses.

![image](https://user-images.githubusercontent.com/10500124/196209932-c68aa72b-695a-4c9b-9a18-c50a950e3bd2.png)

## 💁 Proposed Solution

Ideas:
1. Provide dedicated theme-aware tokens for the graph
2. Update our documentation for existing generic theme-aware tokens to describe how they apply to graphs
3. Provide an example of applying tokens to a plotly graph
4. A dedicated `nimble-graph` component that wraps plotly in a FAST template and applies the tokens automatically
    - in past discussions we've leaned away from this approach because the ROI seems low...it would take a lot of work to expose all the APIs for the graph in a custom element and apps tend to want to do a lot of customization of graphs

rajsite:
I'm wondering if Plotly has the right hooks to be-aware of CSS Custom Properties to use for styling.

I expect that we actually need programmatic hooks into the theme changing and for querying the theme-aware token values programmatically from JS. FAST gives those APIs but we should find the right way to expose that to apps.

jattasNI:
If this issue is accurate then Plotly still doesn't support styling via CSS https://github.com/plotly/plotly.js/issues/1195#issuecomment-262793638

ni-ramya:
Is it possible to expose variables for fonts and font sizes as well? This will be required for styling the legends, graph titles, axis names, tec.

jattasNI:
[Here is the first PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/393894?path=/Web/Workspaces/TestInsights/src/app/shared/components/graph/constants/graph.constants.ts&_a=files) that wants to use these tokens. For now I'm suggesting using the tokens in `design-tokens.ts` to read the current value of certain tokens.

This also revealed a desire to export this API from `nimble-angular` (and presumably `NimbleBlazor` too someday) instead of just `nimble-components`.

When we fix this issue we should go and update that code to use the new API.

jattasNI:
[Here's a PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/405799?discussionId=3933510&_a=files) pulling base token values (unaware of themes) from nimble-tokens.

ni-ramya:
[...]

PR #1579, opened 2023-09-29
Elapsed time column hld

atmgrifter00:
# Pull Request

## 🤨 Rationale

#1569 

[AzDO Work Item: SLE elapsed time table column](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/ASW%20Design%20System/Features/?workitem=2526116)


m-akinc (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+### Non-goals
+
+-   APIs to configure individual aspects of the display (i.e. fractional digits for seconds display)
+-   APIs to format the elapsed time in a variety of ways
+    -   Show elapsed time as digital display (00:01:20.323)
+    -   Show elapsed time with symbols indicating a quantitative value (i.e. "< 1 sec")
+
+### Risks and Challenges
+
+Formatting the elapsed time units in the exact way that SLE presents them would require custom logic that differs from what the browser provides natively using APIs like `Intl.Numberformat.format()`. The preference here is to align with current browser behaviors for unit formatting, which includes localization.
```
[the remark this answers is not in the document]
atmgrifter00: The problem with using "narrow" is that it also removes the spacing between the numeric value and the unit, which I don't believe we want to do.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+1. When and if we are able to adopt the `Intl.DurationFormat` to format values, we will no longer be hard-coding the commas in the string result as that object will produce the entirety of the display text, which may not match what we produce in our implementation.
+
+### API
+
+_*Props/Attrs*_
+
+-   `field-name`: string
+
+This would result in a column whose value would only show up to the 'hours' unit. We would show up to two fractional digits as needed. Formatting to change slightly from what SLE offers now using "hr" instead of "h", and "sec" instead of "s".
```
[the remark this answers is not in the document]
atmgrifter00: While I don't disagree, the proposal is aimed at a solution that creates as little difference between what we get for a display and what is currently there in SLE.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+<nimble-table>
+    <nimble-table-column-elapsed-time field-name="elapsedTime">Elapsed Time</nimble-table-column-elapsed-time>
+</nimble-table>
+```
+
+The displayed value for elapsed time will always take the form of "\<hour display>, \<minute display>, \<second display>.\<fractional seconds>", where the commas are non-localized delimiters <sup>1</sup>.
+
+1. When and if we are able to adopt the `Intl.DurationFormat` to format values, we will no longer be hard-coding the commas in the string result as that object will produce the entirety of the display text, which may not match what we produce in our implementation.
+
+### API
```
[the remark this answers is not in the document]
atmgrifter00: So, the current working theory is that _all_ text-based columns will be groupable (see comments about this on your file size PR). So, there should be no difference in behavior between this column and a numeric or date time column.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+Below is an example of how the `nimble-table-column-elapsed-time` would be used within a `nimble-table`:
+
+```HTML
+<nimble-table>
+    <nimble-table-column-elapsed-time field-name="elapsedTime">Elapsed Time</nimble-table-column-elapsed-time>
+</nimble-table>
+```
+
+The displayed value for elapsed time will always take the form of "\<hour display>, \<minute display>, \<second display>.\<fractional seconds>", where the commas are non-localized delimiters <sup>1</sup>.
```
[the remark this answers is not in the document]
m-akinc: I bet you $5 they don't care. 😉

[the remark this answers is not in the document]
atmgrifter00: I'll leave the thread open for a wider audience to have input on.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+Formatting the elapsed time units in the exact way that SLE presents them would require custom logic that differs from what the browser provides natively using APIs like `Intl.Numberformat.format()`. The preference here is to align with current browser behaviors for unit formatting, which includes localization.
+
+### Prior Art/Examples
+
+![ElapsedTime SLE](../spec-images/ElapsedTimeSLE.PNG)
+
+## Design
+
+Below is an example of how the `nimble-table-column-elapsed-time` would be used within a `nimble-table`:
```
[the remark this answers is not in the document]
atmgrifter00: I'll leave thread open for other ideas. I do not oppose `duration`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+The displayed value for elapsed time will always take the form of "\<hour display>, \<minute display>, \<second display>.\<fractional seconds>", where the commas are non-localized delimiters <sup>1</sup>.
+
+1. When and if we are able to adopt the `Intl.DurationFormat` to format values, we will no longer be hard-coding the commas in the string result as that object will produce the entirety of the display text, which may not match what we produce in our implementation.
+
+### API
+
+_*Props/Attrs*_
+
+-   `field-name`: string
```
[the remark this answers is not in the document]
atmgrifter00: I've added this information in the top 'Overview' section as this fact isn't really an aspect of the API.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
+# Elapsed Time Table Column
+
+## Overview
+
+The elapsed time table column will provide a way to visualize numeric data that represents a timespan. A 'timespan' is simply a numeric value representing a total number of seconds (i.e. a value of 5 means '5 seconds').
+
+### Background
+
+[Work Item](https://ni.visualstudio.com/DevCentral/_workitems/edit/2526116)
```
mollykreis, line 9: You should change this link (or add a second link) to the nimble issue rather than the azdo issue: https://github.com/ni/nimble/issues/1569

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Standard unit testing. Test cases of interest:
+
+-   unusual number values (NaN, -∞, ∞, negative values)
```
mollykreis, line 162: I'm not sure if it belongs in the "Test Plan" section or some other section above, but I think it would be good to define the behavior for these "unusual number values". Will they all be rendered (with or without units)? Will some/all of them just display blank cells?

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Documentation
+
+This component will be documented via a new story in Storybook.
+
+---
+
+## Open Issues
+
+-   Should we expand set of displayed time units to include days and/or years? Without introducing an accompanying API (i.e. would show days and years as needed by default), this might result in a different display than what SLE currently results in.
```
mollykreis, line 177: I personally don't see an issue showing 'days' as a unit larger than 'hours', but I think extending it up to 'years' may open a can of worms. With leap year, I'm not sure we can state that all years have the same number of seconds. Plus, I could see that by the time something has lasted a year (or multiple years), we might stop caring so much about the number of seconds for that particular value. I could see including years really being year+days rather than getting all the way down to something like `3 years, 269 days, 19 hours, 12 minutes, 36.86 seconds`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+
+Below is an example of how the `nimble-table-column-elapsed-time` would be used within a `nimble-table`:
+
+```HTML
+<nimble-table>
+    <nimble-table-column-elapsed-time field-name="elapsedTime">Elapsed Time</nimble-table-column-elapsed-time>
+</nimble-table>
+```
+
+The displayed value for elapsed time will always take the form of "\<hour display>, \<minute display>, \<second display>.\<fractional seconds>", where the commas are non-localized delimiters <sup>1</sup>.
```
mollykreis, line 36: Unless I missed it, I didn't see anything that stated how many `fractional seconds` would be shown. I assume it isn't configurable since I didn't see an API to control that. Are you thinking that it'll be like currently implemented in SLE, which is a max of 2 digits but could be less if not necessary for a given value (i.e. `maximumFractionDigits: 2` on the [NumberFormatter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat))?

atmgrifter00: I made mention of it in the paragraph in the 'API' section. To summarize I was thinking we would just default to showing at most 2 fractional digits, just like in SLE today, and forgo the configurable API for now.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+
+### API
+
+_*Props/Attrs*_
+
+-   `field-name`: string
+
+This API would result in a column whose value would only show up to the 'hours' unit (i.e. it would show hours, minutes, and seconds). We would show up to two fractional digits as needed. Formatting to change slightly from what SLE offers now using "hr" instead of "h", and "sec" instead of "s".
+
+Examples:
```
jattasNI, line 48: The examples illustrate some of this but it might be worth making it explicit: larger units won't be displayed if their value is 0 but smaller ones will (I'm guessing).

**Not:**
- 0 hr, 0 min, 2 sec
- 1 hr, 3 sec
- 1 hr

**Yes:**
- 2 sec
- 1 hr, 0 min, 3 sec
- 1 hr, 0 min, 0 sec

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
jattasNI: I also strongly support adding 'days' since SLE will likely have durations like test execution times which are greater than 24 hours. And that will somewhat alleviate the clunkiness of including commas as both thousands separators and unit separators (`5,000 hr, 2 min, 1 sec` becomes `208 days, 8 hr, 2 min, 1 sec`).

I'm ok with not supporting weeks/months/years for now due to unclear requirements and complexity about conversion factors.

And I don't think we need to add a configuration option to turn off days formatting initially. We can do that (and your other configuration alternatives) if a client asks for it.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+
+1. When and if we are able to adopt the `Intl.DurationFormat` to format values, we will no longer be hard-coding the commas in the string result as that object will produce the entirety of the display text, which may not match what we produce in our implementation.
+
+### API
+
+_*Props/Attrs*_
+
+-   `field-name`: string
+
+This API would result in a column whose value would only show up to the 'hours' unit (i.e. it would show hours, minutes, and seconds). We would show up to two fractional digits as needed. Formatting to change slightly from what SLE offers now using "hr" instead of "h", and "sec" instead of "s".
```
jattasNI, line 46: Could you add some examples in other NI-supported languages? Do any of them require lots of additional width or do they all offer good abbreviations? Do the translations provided by the browser match what SLE currently does?

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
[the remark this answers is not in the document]
jattasNI: FYI to other reviewers, this thread [is continued](https://github.com/ni/nimble/pull/1579/files#r1344141560) in the Open Issues section.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
```diff
@@ -0,0 +1,177 @@
...
+### Non-goals
+
+-   APIs to configure individual aspects of the display (i.e. fractional digits for seconds display)
+-   APIs to format the elapsed time in a variety of ways
+    -   Show elapsed time as digital display (00:01:20.323)
+    -   Show elapsed time with symbols indicating a quantitative value (i.e. "< 1 sec")
+
+### Risks and Challenges
+
+Formatting the elapsed time units in the exact way that SLE presents them would require custom logic that differs from what the browser provides natively using APIs like `Intl.Numberformat.format()`. The preference here is to align with current browser behaviors for unit formatting, which includes localization.
```
jattasNI, line 20: I asked a similar question on the file size HLD: can / should we expose an API that clients can use to format time in the same way as the column does without instantiating a column? Ideally the `ElapsedTimePipe` would have the same formatting logic as the elapsed time column.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
atmgrifter00: So that's now how the current `ElapsedTimePipe` is implemented. A value for a particular time unit is only included in the display if it has a non-zero value.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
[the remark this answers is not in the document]
atmgrifter00: FYI to all reviewers, this thread is now being continued here, in that I am resolving it to include "days" now.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
atmgrifter00: Added days, but no years.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,200 @@
...
+}
+```
+
+### Angular integration
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created.
+
+### Blazor integration
+
+A Blazor wrapper will be created for the component.
```
jattasNI, line 83: Now that you've added `DurationFormatter` (thanks!), what framework integrations would we want for it?

I'm guessing for Angular we'd just re-export it from nimble-angular so that clients don't need to depend on nimble-components to access it?

For Blazor my gut is that it's not necessary initially but would be easy to add if a use case came up. 

Either way, let's document our proposals here.

And I ass

change/@ni-nimble-components-292ee238-87d4-40bc-b76c-941992c1f31e.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Let's add link to the new spec from the table column and formatted text specs, similar to what Mert's doing in #1566.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
jattasNI: I think I'm fine with that as a default. Thanks for updating the examples.

It does introduce a corner case for the value 0. Would we display blank? 0 sec?

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
atmgrifter00: I've added a table showing the expected results in 3 other languages (German, French, and Chinese).

atmgrifter00: Done. Apparently, the translations will _not_ match what SLE currently does, as the units are not currently localized, only the numeric values.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
atmgrifter00: I've updated the example to just use "0 sec" instead of "2 sec" to demonstrate that we _will_ show a value for 0.

change/@ni-nimble-components-292ee238-87d4-40bc-b76c-941992c1f31e.json
atmgrifter00: Done, but note that only the formatted text spec was altered to point to the new spec in Mert's changes. I've followed suit.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,202 @@
...
+
+```ts
+// inherits field-name attribute
+public class TableColumnDurationText : TableColumnTextBase { }
+
+```
+
+This API would result in a column whose value would only show up to the 'days' unit (i.e. it would show days, hours, minutes, and seconds). We would show up to two fractional digits as needed. Formatting to change slightly from what SLE offers now using "hr" instead of "h", and "sec" instead of "s".
+
+Examples:
```
mollykreis, line 54: Interestingly, most of these localized values don't align with what SLE displays for non-English today. I'm wondering if that's fine, or if we need to pull the thread a bit more to verify that we're ok with a change here. I'm not even sure who we'd check with -- localization engineers?

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,202 @@
...
+
+We will export the class that is responsible for formatting the duration values so that we can use it for formatting values outside of the `Table` (a requirement in SLE).
+
+```ts
+export class DurationFormatter {
+    public constructor(private readonly lang: string) {}
+
+    // formats a value in the seconds domain into a localized display value that can
+    // consist of days, hours, minutes, and seconds.
+    public format(value: TableNumberFieldValue): string {}
```
mollykreis, line 73: Should we decouple this from table types and not use `TableNumberFieldValue` here?

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,202 @@
...
+}
+```
+
+### Angular integration
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created. The `DurationFormatter` will be re-exported from Angular.
```
mollykreis, line 79: Any thoughts on what entry point we would export the `DurationFormatter` from in Angular?

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Accessibility
+
+### Mobile
+
+N/A
+
+### Globalization
+
+For now we do not intend to alter behavor for `ltr` vs `rtl` settings.
```
mollykreis, line 165:
```suggestion
For now we do not intend to alter behavior for `ltr` vs `rtl` settings.
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
jattasNI: Posting [on this channel](https://teams.microsoft.com/l/channel/19%3Acd7187eeafd24fc3a3d98256ba0c55c2%40thread.skype/Prj_SystemLink?groupId=0af69736-3f49-48ec-8c7e-d122f54130d5&tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e) would be a good way to ask for input from localization engineers. It'd be good to get confirmation that the strings are acceptable, but I think we should stay firm about strongly preferring the browser translations unless they're outright wrong. Let's try to avoid extra work if people suggest tweaks that are hard to support but not essential.

atmgrifter00: I've [posted](https://teams.microsoft.com/l/message/19:cd7187eeafd24fc3a3d98256ba0c55c2@thread.skype/1696521241285?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=null&parentMessageId=1696521241285&teamName=Localization%20Projects&channelName=Prj_SystemLink&createdTime=1696521241285) on the channel.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,202 @@
...
+
+N/A
+
+### Dependencies
+
+None
+
+### Test Plan
+
+Standard unit testing. Test cases of interest:
```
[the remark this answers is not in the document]
atmgrifter00: I've added a 'Considered Corner Cases' section where I capture this concern.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,206 @@
...
+... // do similar work for minutes and seconds
+```
+
+Other implementation details will conform to existing patterns all other table columns handle, such as `CellView` and `GroupHeaderView` implementations.
+
+_Note: For number values that can't be translated into a representable time (e.g. `NaN`, `∞`, `-∞`, negative values), we will simply show a blank cell._
+
+### Considered Corner Cases
+
+Currently in SLE, there are corner cases that can result in a display of something like "25 min, 60 sec", which is non-ideal. This happens when the value being formatted has a large fractional value close to 1 (e.g. .999999997). The better behavior would be to either show a formatted value of "26 min", or even "25 min, 59.99 sec", though the former would be a more accurate representation if the value was 59.99999997.
```
rajsite, line 110: How will really small durations be shown? like lifetime of muon decay in atmosphere, Planck time, etc.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
rajsite: I'd probably avoid a name that generic on the public API (we have an `internal-utilities` entrypoint not intended to be publicly consumed). Maybe `formatters`, so like `import {DurationFormatter} from '@ni/nimble-angular/formatters';`?

Guess it's encroaching on the component names, so wouldn't have  `<nimble-formatters>` elements 🤷

So our entrypoints would eventually look like:
```
/<elements>
/internal-utilities
/formatters
/testing (not there now, but eventually)
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Alternative Implementations / Designs
```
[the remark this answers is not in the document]
atmgrifter00: This is a great point. The reasoning behind the current proposal was simply with alignment with what SLE is currently expecting. However, I agree that it's totally valid for us to question that as an API. At the moment, I think I would agree with you and prefer that we should expect values to be in the milliseconds unit.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,206 @@
...
+| "3 days, 23 hr, 2 min" | "3 Tg., 23 Std., 2 Min."  | "3 j, 23 h, 2 min" | "3 天, 23 小时, 2 分钟" |
+| "0 sec"                | "0 Sek."                  | "0 s"              | "0 秒"                  |
+
+#### `DurationFormatter`
+
+We will export the class that is responsible for formatting the duration values so that we can use it for formatting values outside of the `Table` (a requirement in SLE).
+
+```ts
+export class DurationFormatter {
+    public constructor(private readonly lang: string) {}
```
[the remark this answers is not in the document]
atmgrifter00: I guess the only reason I have to push back on this is that the native APIs for `Intl.NumberFormat` and `Intl.DurationFormat` provide the locales as a .ctor parameter much like I have here, and then the rest of the options in a separate parameter.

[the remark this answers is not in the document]
rajsite: Re-reading the proposal I gave I think it does have some gaps. It would need to take an element reference and also some kind of callback function to fire and notify that the locale has changed.

From that perspective and with your point seems like this would be two separate APIs if comparing to DOM apis. A `DurationFormatter(lang: string, opts)` and a `LangObserver` (similar to ResizeObserver, IntersectionOberserver, etc) which kinda fits in the [ThemeObserver discussion](https://github.com/ni/nimble/issues/783#issuecomment-1287160658).

The `LangObserver` may not be needed today yet since I don't think any Angular apps have different localized subsections or need to respond to lang changes programmatically as they just refresh the page (I don't know if those two assertions are true). If all that's true it could be a future work maybe (assuming such an observer isn't needed for Blazor apps that are localized today)

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
rajsite: Just landed on this from a completely unrelated context but found the W3C Web Platform Design Principles guidelines and still working through it but it had a comment on time units always being milliseconds: https://w3ctag.github.io/design-principles/#milliseconds

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,206 @@
...
+// inherits field-name attribute
+public class TableColumnDurationText : TableColumnTextBase { }
+
+```
+
+This API would result in a column whose value would only show up to the 'days' unit (i.e. it would show days, hours, minutes, and seconds). We would show up to two fractional digits as needed. Formatting to change slightly from what SLE offers now using "hr" instead of "h", and "sec" instead of "s".
+
+Examples:
+
+| English                | German                    | French             | Chinese                 |
```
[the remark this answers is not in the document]
atmgrifter00: I was really just trying to be cautious in my statements about what _could_ happen. The [docs for `Intl.DurationFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/format#using_format) strongly suggest that the individual parts will be comma separated.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
atmgrifter00: It seems the proposed direction is acceptable with the localization folks.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,206 @@
...
+    result.push(formattedHours);
+}
+... // do similar work for minutes and seconds
+```
+
+Other implementation details will conform to existing patterns all other table columns handle, such as `CellView` and `GroupHeaderView` implementations.
+
+_Note: For number values that can't be translated into a representable time (e.g. `NaN`, `∞`, `-∞`, negative values), we will simply show a blank cell._
+
+### Considered Corner Cases
```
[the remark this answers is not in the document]
atmgrifter00: Proposed we format extremely large numbers (what the cutoff value for "extremely large" is t.b.d.) as scientifically formatted seconds.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
atmgrifter00: Also propose we format extremely small numbers as seconds in scientific format, until we have a more configurable API.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,221 @@
...
+
+    // formats a value in the seconds domain into a localized display value that can
+    // consist of days, hours, minutes, and seconds.
+    public format(value: number): string {}
+}
+```
+
+### Angular integration
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created. The `DurationFormatter` will be re-exported from Angular from a new `formatters` entrypoint.
```
rajsite, line 79: Should we also ship an ElapsedTimePipe that calls into DurationFormatter?

atmgrifter00: This makes sense to me.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
jattasNI: Here are some examples from Safari 16.6. Looks like the comma separators are language-dependent, which matches what our localization engineers suggested.

```
> const duration = {
  hours: 1,
  minutes: 46,
  seconds: 40,
};


> new Intl.DurationFormat("fr-FR", { style: "long" }).format(duration);
< "1 heure, 46 minutes et 40 secondes"

> new Intl.DurationFormat("fr-FR", { style: "short" }).format(duration);
< "1 h, 46 min et 40 s"

> new Intl.DurationFormat("fr-FR", { style: "narrow" }).format(duration);
< "1h 46min 40s"

> new Intl.DurationFormat("ja", { style: "short" }).format(duration);
< "1時間 46分 40秒"

> new Intl.DurationFormat("zh-CN", { style: "short" }).format(duration);
< "1小时46分钟40秒"

> new Intl.DurationFormat("DE", { style: "short" }).format(duration);
< "1 Std., 46 Min. und 40 Sek."
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,221 @@
...
+
+Ultimately, we will expect to be able to configure a duration column to show values down to the nanoseconds unit once we can leverage `Intl.DurationFormat`, and we expose the right formatting APIs on the column. Until then, we should behave as we do for really large numbers, and if the value is below a particular threshold we will format the value in scientific in the seconds unit.
+
+#### Large fractional parts
+
+Currently in SLE, there are corner cases that can result in a display of something like "25 min, 60 sec", which is non-ideal. This happens when the value being formatted has a large fractional part close to 1 (e.g. .999999997). The better behavior would be to either show a formatted value of "26 min", or even "25 min, 59.99 sec", though the former would be a more accurate representation if the value was 59.99999997.
+
+#### Negative Zero
+
+`-0` will simply render as "0".
```
jattasNI, line 124: nit: either "`-0` will render as "0 sec"" or "`-0` will render the same as 0 does". As this reads now it seems like 0 will render as "0 sec" but -0 will render as "0", which would be confusing.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
jattasNI: Given this, I would vote to stick with the current proposal to use commas for all languages until DurationFormat is well supported. When DurationFormat is well supported we can switch to it, which will probably change the comma behavior for some languages (assuming Safari's implementation is standard)

That seems cleaner than trying to be more forward compatible by adding special cases that omit commas for ja and zh. There are more such cases than seems worthwhile to me (in addition to varying commas there are also varying spaces and inconsistency about including "and" before seconds).

jattasNI (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
atmgrifter00: Updated spec to state we will use milliseconds as the unit of the provided values.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
atmgrifter00: Made a 'Future Considerations' section where I mention the possibility of needing an observer class.

packages/nimble-components/src/table/specs/table-column-specs/table-column-elapsed-time-hld.md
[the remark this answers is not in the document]
atmgrifter00: I've now changed the spec to say we will display up to 3 decimal places (in addition to showing values in scientific notation for certain situations).

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,227 @@
...
+    // formats a value in the seconds domain into a localized display value that can
+    // consist of days, hours, minutes, and seconds.
+    public format(value: number): string {}
+}
+```
+
+### Angular integration
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created. The `DurationFormatter` will be re-exported from Angular from a new `formatters` entrypoint.
+
```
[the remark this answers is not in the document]
rajsite: Think we had chatted about it being a `pipes` entrypoint since that is what angular would actually export.
Not sure if we should re-export the formatter as well (if no use-case for using it directly then maybe not yet).

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,227 @@
...
+Other implementation details will conform to existing patterns all other table columns handle, such as `CellView` and `GroupHeaderView` implementations.
+
+_Note: For number values that can't be translated into a representable time (e.g. `NaN`, `∞`, `-∞`, negative values), we will simply show a blank cell._
+
+### Considered Corner Cases
+
+#### Really large numbers
+
+At some point numbers that can be formatted as a duration become too large to be practically represented in such a way. We will have a cutoff value (in days) where when the value exceeds that limit, we will revert to displaying the duration as a scientific representation of the value in the seconds unit.
```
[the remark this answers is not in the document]
rajsite: For very large numbers it seemed more intuitive than saying 3.264E61 days
Curious about the concern since I just recommended the same thing to @m-akinc for very large and very small numbers we don't have localized prefixes for.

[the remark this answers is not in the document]
atmgrifter00: I don't have a strong opinion here, but I'll leave it as is until I can be swayed in the other direction.

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,227 @@
...
+    // consist of days, hours, minutes, and seconds.
+    public format(value: number): string {}
+}
+```
+
+### Angular integration
+
+An Angular directive will be created for the component. The component will not have form association, so a `ControlValueAccessor` will not be created. The `DurationFormatter` will be re-exported from Angular from a new `formatters` entrypoint.
+
+We will also export a `DurationPipe` that will use the `DurationFormatter` under the hood.
```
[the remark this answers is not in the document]
atmgrifter00: Nah. Updated in new [PR](https://github.com/ni/nimble/pull/1597).

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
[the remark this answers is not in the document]
atmgrifter00: Corrected this in the [new PR](https://github.com/ni/nimble/pull/1597).

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,227 @@
...
+
+#### `DurationFormatter`
+
+We will export the class that is responsible for formatting the duration values so that we can use it for formatting values outside of the `Table` (a requirement in SLE).
+
+```ts
+export class DurationFormatter {
+    public constructor(private readonly lang: string) {}
+
+    // formats a value in the seconds domain into a localized display value that can
```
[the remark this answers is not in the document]
atmgrifter00: Corrected in [new PR](https://github.com/ni/nimble/pull/1597).

packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md
```diff
@@ -0,0 +1,227 @@
...
+### Non-goals
+
+-   APIs to configure individual aspects of the display (i.e. fractional digits for seconds display)
+-   APIs to format the duration in a variety of ways
+    -   Show duration as digital display (00:01:20.323)
+    -   Show duration with symbols indicating a quantitative value (i.e. "< 1 sec")
+
+### Risks and Challenges
+
+Formatting the duration units in the exact way that SLE presents them would require custom logic that differs from what the browser provides natively using APIs like `Intl.Numberformat.format()`. The preference here is to align with current browser behaviors for unit formatting, which includes localization.
```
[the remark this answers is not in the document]
atmgrifter00: Corrected in [new PR](https://github.com/ni/nimble/pull/1597).

Issue #1573, opened 2023-09-26
Number-text table column unit support


m-akinc:
## 😯 Problem to Solve

SLE needs a table column to display file sizes.
See AzDO Feature: https://ni.visualstudio.com/DevCentral/_workitems/edit/2526114

## 📋 Tasks

- [x] spec
- [ ] element updates, tests, and docs
- [ ] update Angular wrapper, tests
- [ ] update Blazor wrapper, tests
- [ ] Add to Angular example app
- [ ] Add to Blazor example app

PR #1596, opened 2023-10-12
Update numeric column spec for unit support

m-akinc:
# Pull Request

## 🤨 Rationale

Spec for #1573 

## 👩‍💻 Implementation

See spec.

## 🧪 Testing

NA

## ✅ Checklist


m-akinc:
@mollykreis can I get a buddy on this?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,22 +161,32 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+-   `decimal-maximum-digits` - when format is `decimal`, a number that controls the maximum number of digits shown to the right of the decimal separator. Possible values are from 0 to 20. Defaults to 3. If `decimal-digits` is also specified, exactly that number of digits are shown and `decimal-maximum-digits` is ignored. Formats other than `decimal` ignore `decimal-maximum-digits`.
+-   `unit-family` - when specified, displays a unit label from among a family of related units. Values are converted from a source unit (e.g. bytes) to the smallest related unit that can represent that value with magnitude >= 1 (e.g. KB, MB, etc). Set this attribute to the tag name of an element representing a family of related units, e.g. `'nimble-unit-byte'`. These unit elements are not rendered but serve to allow selective loading of data for only the needed units. The source data for the column is expected to be given in the unit specified in the tag name, e.g. for `'nimble-unit-byte'`, a source value should be a number of bytes. Supported units will be:
```
mollykreis, line 176: Any thoughts on when we'd add support for more units? Would we wait for requests or add the ones that the NumberFormatter gives us translations for right away?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,22 +161,32 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
--   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 if unspecified. Formats other than `decimal` ignore `decimal-digits`.
+-   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 unless `decimal-maximum-digits` is specified, in which case a number of digits between 0 and `decimal-maximum-digits` are shown. Formats other than `decimal` ignore `decimal-digits`.
+-   `decimal-maximum-digits` - when format is `decimal`, a number that controls the maximum number of digits shown to the right of the decimal separator. Possible values are from 0 to 20. Defaults to 3. If `decimal-digits` is also specified, exactly that number of digits are shown and `decimal-maximum-digits` is ignored. Formats other than `decimal` ignore `decimal-maximum-digits`.
```
mollykreis, line 175: Using `decimal-maximum-digits` and/or `unit-family` (specifically `unit-family` when not all values result in the same unit) will mess up the right-alignment that gets applied automatically with the `decimal` format. We probably need to update the logic for the `default` alignment to account for this.

mollykreis (changes requested):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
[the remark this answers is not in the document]
m-akinc: Thanks, you're right.

> ...should this really be a validation error?

Yeah, I guess it should.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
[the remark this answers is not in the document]
m-akinc: Whoops. Yes, that's what I meant.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,22 +161,32 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+-   `unit-family` - when specified, displays a unit label from among a family of related units. Values are converted from a source unit (e.g. bytes) to the smallest related unit that can represent that value with magnitude >= 1 (e.g. KB, MB, etc). Set this attribute to the tag name of an element representing a family of related units, e.g. `'nimble-unit-byte'`. These unit elements are not rendered but serve to allow selective loading of data for only the needed units. The source data for the column is expected to be given in the unit specified in the tag name, e.g. for `'nimble-unit-byte'`, a source value should be a number of bytes. Supported units will be:
+    -   `'nimble-unit-byte'` - Unit labels in this family are `byte`/`bytes`, `KB`, `MB`, `GB`, `TB`, `PB`
```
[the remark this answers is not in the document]
m-akinc: Yes, it is an elment name:
> Set this attribute to the tag name of an element representing a family of related units, e.g. `'nimble-unit-byte'`

I believe we have to take the element name directly (not an enum value) so that the client imports the tag name constant, rather than us importing all of those constants (and associated elements).

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,22 +161,32 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
 
 This column will trigger `invalidColumnConfiguration` on the table's validity state and will include flags in the column's validity state if its configuration can't be translated to a valid `Intl.NumberFormat` object. To provide better developer feedback about what's wrong with the configuration, the column could expose a public method like `createNumberFormat()` which would use the column's configuration to construct the formatter but allow any exceptions to be thrown.
 
+The `'human-readable'` format will rely on `Intl.NumberFormat` to translate unit strings (for the [units that it supports](https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers)) and localize the number (for comma/decimal). We will include our own logic for converting between unit values. For a family of units not supported by `Intl.NumberFormat`, we will provide our own translations (for some subset of languages) in a `nimble-unit-<name>` element. If the client requests a translation for one of these units in a language we don't support, we will fall back to English.
```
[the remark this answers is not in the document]
m-akinc: Whoops. That was from the previous revision of this spec where I was making unit support a separate format mode (called `'human-readable'` at Milan's request). I'll update this to say something like "When `unit-family` is specified, the column will..."

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,22 +161,32 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+Other libraries considered for number conversion and/or unit label localization include [globalizejs](https://github.com/globalizejs/globalize), [convert](https://convert.js.org/), [convert-units](https://github.com/convert-units/convert-units), and [iLib](https://github.com/iLib-js/iLib). None of these provided sufficient functionality to justify including them as a dependency.
+
+We will use `unit-family='nimble-unit-byte'` to display file sizes in SLE tables. Currently, SLE displays these values with the common `KB`/`MB`/`GB` unit labels, but uses a factor of 1024 to convert between units (which is not uncommon, but technically incorrect). Our formatting will use the same unit labels, but use a conversion factor of 1000. This will result in slightly different values being displayed. For SLE's internal consistency, we will also change their file size pipe to use a conversion factor of 1000 instead of 1024.
```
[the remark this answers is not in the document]
m-akinc: I just wanted to include a straw-man plan here with the understanding that we will potentially change it based on PO input. I'll leave this comment thread open until we have a resolution.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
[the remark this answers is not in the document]
m-akinc: I'll try to make this more clear. The idea is that _every_ supported unit family will have a corresponding `nimble-unit-<name>`. Even for units supported by `Intl.NumberFormat`, having corresponding elements allows us to keep a consistent API and use a consistent abstraction in the implementation.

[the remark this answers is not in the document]
m-akinc: I hadn't worked out all the implementation details at the time of writing, but in my prototyping so far, I've come up with this:
```TS
interface Unit {
    public conversionFactor: number;
    public format(value: number): string;
}

interface UnitFamily {
    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[];
}

// supported by Intl.NumberFormat
class UnitFamilyByte extends FoundationElement implements UnitFamily { 
    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[] {
        // returns implementations of Unit that wrap Intl.NumberFormat instances configured for a specific locale and a specific unit (e.g. 'fr-FR' and 'kilobyte')
    }
}

// manually translated unit
class UnitFamilyVolt extends FoundationElement implements UnitFamily {
    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[] {
        // returns implementations of Unit that contain a shared Intl.NumberFormat (for formatting the number) and a specific translated unit string to append
    }
}
```

> What does it look like to support smaller units (e.g. amps converted to mA or μA)?

Source values are still expected to be given in amps. We just have conversion factors <1:
mA: `unit.conversionFactor = 0.001`
μA: `unit.conversionFactor = 1e^6`

> How does a given unit specify what magnitudes of conversions it supports?

The `getSupportedUnits` implementation returns a `Unit` for each supported unit/conversion. A `Unit` has a conversion factor and can provide one specific, translated unit label.

I'm not sure how much of these details belong in this spec. I can include the code above if that would help.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: I was thinking we would wait for requests. I want to have at least one example of a `Intl.NumberFormat`-supported unit and one manually-translated unit to serve as examples and to ensure a good design. If there are other units that we know up-front will be useful, I'm open to adding them.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: That makes sense.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -200,6 +239,15 @@ This column will trigger `invalidColumnConfiguration` on the table's validity st
...
     </nimble-table-column-number-text>
+
+    <nimble-table-column-number-text
+        field-name="fileSize"
+        format="decimal"
+        decimal-maximum-digits="1"
+        unit-family="nimble-unit-byte"
+    >
+        File Size
+    </nimble-table-column-number-text>
```
mollykreis, line 250: This should have the units element within it, right?

```suggestion
        File Size
        <nimble-unit-byte></nimble-unit-byte>
    </nimble-table-column-number-text>
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,60 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
 
-This column will trigger `invalidColumnConfiguration` on the table's validity state and will include flags in the column's validity state if its configuration can't be translated to a valid `Intl.NumberFormat` object. To provide better developer feedback about what's wrong with the configuration, the column could expose a public method like `createNumberFormat()` which would use the column's configuration to construct the formatter but allow any exceptions to be thrown.
+This column will trigger `invalidColumnConfiguration` on the table's validity state and will include flags in the column's validity state if both `decimal-digits` and `decimal-maximum-digits` are specified, or its configuration can't be translated to a valid `Intl.NumberFormat` object. To provide better developer feedback about what's wrong with the configuration, the column could expose a public method like `createNumberFormat()` which would use the column's configuration to construct the formatter but allow any exceptions to be thrown.
```
mollykreis, line 182: In this paragraph (or perhaps the documentation for `unit-family`), you should mention that setting a `unit-family` that doesn't correspond to a registered element that implements `UnitFamily` will also result in a validation error.

mollykreis (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: I wasn't planning it that way. Just like the icon mapping column, we can just dynamically create the unit element. We don't even need it in the DOM.

rajsite: A child element might be more appropriate here. Part of the problem with icons was that they render and they have their own configuration settings as a visual component that might not be relevant to table column configuration so we made the column configuration capture only what it cared about (and it may have been a bit of an overcorrection from the flexibility / complication of the template approach).

In this case these elements are just for configuration so it may be worth keeping as children. I think it's also worth diverging from the icon pattern because we found that added some complexities (you need to import components you don't actually use in DOM and we need a reliable / typesafe way to get the tag names in Blazor we didn't figure out yet).

I can't think of any other reason (beyond pattern consistency with icon column which has issues) that these elements used only for config shouldn't be actual DOM children here.

m-akinc: Changed API to take `nimble-unit-<name>` as content of the column element.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+A unit for the column may be configured by providing a `nimble-unit-<name>` element as content (in addition to the column label). Unit elements represent families of related units, e.g. `nimble-unit-byte` represents bytes, KB, MB, etc. Values are converted from a source unit (e.g. bytes) to the largest related unit (e.g. KB, MB, etc.) that can represent that value with magnitude >= 1. The source data for the column is expected to be given in the base unit specified in the tag name, e.g. for `nimble-unit-byte`, a source value should be a number of bytes. Note that unit family elements are not rendered but serve to allow selective loading of data for only the needed units. The initial set of unit elements are:
+
+-   `nimble-unit-byte` - Unit labels in this family are `byte`/`bytes`, `KB`, `MB`, `GB`, `TB`, `PB`
+-   `nimble-unit-byte-1024` - Unit labels in this family are `byte`/`bytes`, `KiB`, `MiB`, `GiB`, `TiB`, `PiB`
```
jattasNI, line 184: Since we're always using child elements to pick the unit, another option instead of a dedicated element for binary byte conversions would be to use one element for both types of byte conversion and configure the behavior via an attribute on the element:
```
<nimble-unit-byte conversion="binary">
```

An obvious con of this approach is that now clients would pull in the `KiB` strings even if they're only using the `KB` unit. But otherwise I like the idea of configuring unit-specific behavior on the unit element. Some other examples (just brainstorming, not advocating these be included in the spec):

Include units that aren't powers of 1000 (see thread below)
```
<nimble-unit-meter include-centi include-deci> 
```

Pick currency
```
<nimble-unit-currnecy currency='peso'>
```

Configure "kilobytes" vs "KB"
```
<nimble-unit-byte unit-display='long'>
```

I'm personally leaning towards fewer unit elements with more configuration, even at the cost of some additional strings for the element. But open to more input or stronger opinions in other directions.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+// unit family with Nimble-provided unit translations
+class UnitFamilyVolt extends FoundationElement implements UnitFamily {
+    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[] {
+        // returns implementations of Unit that contain a shared Intl.NumberFormat (for formatting the number) and a specific translated unit string to append
+    }
+}
+```
+
+Other libraries considered for number conversion and/or unit label localization include [globalizejs](https://github.com/globalizejs/globalize), [convert](https://convert.js.org/), [convert-units](https://github.com/convert-units/convert-units), and [iLib](https://github.com/iLib-js/iLib). None of these provided sufficient functionality to justify including them as a dependency.
```
jattasNI, line 222: This paragraph on alternate libraries could move to the alternatives section.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -331,7 +386,7 @@ Nimble already has a mechanism for clients to provide custom columns by deriving
 
 ### Additional unit APIs
 
-We considered a few different options for displaying units within a cell, but it was unclear if any of them would satisfy the requirements of our clients. Because there are a number of open questions regarding the client requirements, we decided to defer any work involving units.
+We considered a few different options for displaying units within a cell, but it was unclear if any of them would satisfy the requirements of our clients. Because there are a number of open questions regarding the client requirements, we initially decided to defer any work involving units. Later, we revisited unit support as an alternative to implementing a separate file size column type.
 
 Some of the options considered are described below:
```
jattasNI, line 391: nit: I don't think the historical context is useful anymore
```suggestion
We considered some alternative approaches for configuring units:
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+-   `nimble-unit-byte` - Unit labels in this family are `byte`/`bytes`, `KB`, `MB`, `GB`, `TB`, `PB`
+-   `nimble-unit-byte-1024` - Unit labels in this family are `byte`/`bytes`, `KiB`, `MiB`, `GiB`, `TiB`, `PiB`
+-   `nimble-unit-volt` - Unit labels in this family are `volt`/`volts`, plus `V` prefixed by all supported metric prefixes.
+
+Supported metric prefixes are f (femto), p (pico), n (nano), μ (micro), m (milli), c (centi), d (deci), k (kilo), M (mega), G (giga), T (tera), P (peta), and E (exa).
```
jattasNI, line 187: I think the use of prefixes that aren't base 1000 (centi and deci and also [hecto and deca](https://en.wikipedia.org/wiki/Metric_prefix)) will need to be configurable per unit. For example I'm guessing `cm` is common but `cV` is not. I'm not sure if it'll need to be configurable by clients of the column or just implementers of the unit elements, but hopefully our API could be extended to allow either.

I assume we expect all unit elements to be defined within the Nimble repo so the contract to define a unit isn't part of the public API and if we wanted to change that contract it wouldn't be a breaking change? If that's true then we can iterate on that contract as we add more units and don't need to nail it down now.

Given that, we might consider removing the code sample below for now (or marking it as an example) and working that out during implementation. And also making this line less perscriptive.

```suggestion
Supported metric prefixes could include f (femto), p (pico), n (nano), μ (micro), m (milli), c (centi), d (deci), k (kilo), M (mega), G (giga), T (tera), P (peta), and E (exa).
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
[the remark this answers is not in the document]
jattasNI: @fredvisser FYI this thread is waiting for your conversation with POs to happen

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: I think your suggestion makes sense. The extra translations shouldn't be a concern, especially since I doubt we'll actually have to translate the prefixes anyway.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
jattasNI: To help clients maintain consistent formatting outside of the table, in the duration column we ended up deciding to expose a TS class with the formatting logic and an Angular pipe that wraps it. I think in your previous HLD you mentioned that might be low ROI but I'd like to revisit that in a quest for consistency with the duration column.

The challenge here is that there is potentially more configuration to specify when formatting a string. How would a client specify the unit family and any additional configuration to the format?

One idea is a static `format()` method on each unit element class which takes a locale and a configuration object. Nimble may not expose that via a pipe but clients could write their own pipe.

Thoughts?

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,65 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+class UnitFamilyVolt extends FoundationElement implements UnitFamily {
+    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[] {
+        // returns implementations of Unit that contain a shared Intl.NumberFormat (for formatting the number) and a specific translated unit string to append
+    }
+}
+```
+
+We will use `nimble-unit-byte` to display file sizes in SLE tables. Currently, SLE displays these values with the common `KB`/`MB`/`GB` unit labels, but uses a factor of 1024 to convert between units (which is [not uncommon](https://en.wikipedia.org/wiki/JEDEC_memory_standards#Unit_prefixes_for_semiconductor_storage_capacity), but [technically incorrect](https://physics.nist.gov/cuu/Units/binary.html)). Our formatting will use the same unit labels, but use a conversion factor of 1000. This will result in slightly different values being displayed. For SLE's internal consistency, we will also change their file size pipe to use a conversion factor of 1000 instead of 1024. We will need to search for other places in the product where byte values are being converted and update them similarly.
```
jattasNI, line 222: To help clients maintain consistent formatting outside of the table, in the duration column we ended up deciding to expose a TS class with the formatting logic and an Angular pipe that wraps it. I think in your previous HLD you mentioned that might be low ROI but I'd like to revisit that in a quest for consistency with the duration column.

The challenge here is that there is potentially more configuration to specify when formatting a unit value. How would a client specify the unit family and any additional configuration to the format?

I don't have great ideas at the moment. If others don't either then maybe we can defer this until we start uptaking this in SLE.

m-akinc: Essentially duplicate of https://github.com/ni/nimble/pull/1596#discussion_r1364236902

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
jattasNI: At first glance these ideas all feel a bit clunky to me (I actually hesitated to suggest it at all, hence the near-duplicate comment thread with this suggestion removed that I've now deleted). Either we duplicate a bunch of format configuration API or we require instantiating an element just to format a string. I don't want this question to hold up progress on the column work, so would be ok with leaving this as an open question and revisiting it when we start integrating into SLE.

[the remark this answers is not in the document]
jattasNI: Here is Fred's ADR discussing it: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/582048

jattasNI (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: Ok, I moved the bit I added about creating an external formatter to the "Open Issues" section.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,65 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+A unit for the column may be configured by providing a `nimble-unit-<name>` element as content (in addition to the column label). Unit elements represent families of related units, e.g. `nimble-unit-byte` represents bytes, KB, MB, etc. Values are converted from a source unit (e.g. bytes) to the largest related unit (e.g. KB, MB, etc.) that can represent that value with magnitude >= 1. The source data for the column is expected to be given in the base unit specified in the tag name, e.g. for `nimble-unit-byte`, a source value should be a number of bytes. Note that unit family elements are not rendered but serve to allow selective loading of data for only the needed units. The initial set of unit elements are:
```
atmgrifter00, line 181: Feels like there should be a 'Units' sub-heading for this section.

atmgrifter00, line 181: I'm not sure I understand what you mean by "Note that unit family elements are not rendered but serve to allow selective loading of data for only the needed units." Do you mean the custom element (e.g. `nimble-unit-<name>`) isn't a rendered entity? If so, perhaps this can be better said as "Unit elements serve as configuration components only, and have no visual representation on their own".?

m-akinc: Given the choice, I preferred to insert a 'Units' sub-heading.

m-akinc: The two points I was trying to communicate were that (1) they have no visual, and (2) the motivation for using an element-based API is to avoid always loading all translations for all languages for all units. I've reworded the sentence to hopefully make that clearer.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,65 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+    public format(value: number): string;
+}
+
+// a set of related units, e.g. {byte, kilobyte, megabyte, gigabyte, terabyte, petabyte}
+interface UnitFamily {
+    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[];
+}
+
+// unit family supported by Intl.NumberFormat
+class UnitFamilyByte extends FoundationElement implements UnitFamily {
```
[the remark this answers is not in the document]
m-akinc: > Is this class analogous to `nimble-unit-byte`?

Yes.

> should it be `nimble-unit-family-byte`?

That name would be more consistent with the terminology and type names I'm using in this spec, but I don't want to codify "unit family" into official terminology. I think a client will think of this element as "the unit" to use, so I think that's the name we should use.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): Unit[] {
```
rajsite, line 211: We have the NumberFormatter pattern. I think these elements should just return a NumberFormatter class that can be constructed instead of coming up with a new pattern. Maybe there would be a shared subsclass so there is a consistent interface

```ts
abstract class UnitScaleFormatter extends NumberFormatter {
    constructor (lang, options) {}
    format(number): string
}

class ByteScaleFormatter extends UnitScaleFormatter {}
```

And then the element class name can align more closely with the element name and an abstract class to define the middle interface:
```ts
class Unit extends FoundationElement {
    abstract getFormmater()
}

class UnitByte extends Unit{
    getFormmater() {return ByteScaleFormatter}
}

// somewhere in code:
const FormatterClass = el.getFormatter();
const formatter = new FormatterClass(lang, opts);
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
     -   This could be extended to other pre-configured formats in future. Their configuration attributes would be prefixed with the name of the format mode.
     -   **Note:** all of the above will be implemented using a `Intl.NumberFormat` formatter. Nimble will configure the formatter with defaults to match the [visual design spec](https://github.com/ni/nimble/issues/887). The exception is that we will set `useGrouping: true` to achieve `1,000` rather than `1000` because this styles the values in a way that is more human readable.
--   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 if unspecified. Formats other than `decimal` ignore `decimal-digits`.
+-   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 unless `decimal-maximum-digits` is specified, in which case a number of digits between 0 and `decimal-maximum-digits` are shown. It is invalid to specify both `decimal-digits` and `decimal-maximum-digits`. Formats other than `decimal` ignore `decimal-digits`.
```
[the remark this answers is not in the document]
m-akinc: Yes. Looks like we document this in Storybook, but not here. I'll add it.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
+
+```TS
+// a specific unit, e.g. kilobyte, millivolt, etc.
+interface Unit {
+    public conversionFactor: number;
+    public format(value: number): string;
+}
+
+// a set of related units, e.g. {byte, kilobyte, megabyte, gigabyte, terabyte, petabyte}
+interface UnitFamily {
```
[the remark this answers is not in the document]
m-akinc: So `UnitFamily` -> `UnitScale` and `Unit` -> `ScaledUnit`? That's fine with me.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: One difficulty I anticipate is that a `UnitScaleFormatter` is not independent of the existing `NumberFormatter`s we have today. We have to handle formatting both decimal or default modes, with or without units. In my prototype, I accomplish this by composing `NumberFormatter`s: `DefaultFormatter` and `DecimalFormatter` internally create a `UnitScaleFormatter` to use. I'll try to see if your suggestion can work, though. The line before the code section already says that it is only "an exmple of the abstractions/APIs that might be used," so it's understood that it is subject to change.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -161,21 +161,67 @@ Follows the [Column Type Philosophy](/packages/nimble-components/src/table/specs
...
--   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 if unspecified. Formats other than `decimal` ignore `decimal-digits`.
+-   `decimal-digits` - when format is `decimal`, a number that controls how many digits are shown to the right of the decimal separator. Defaults to 2 unless `decimal-maximum-digits` is specified, in which case a number of digits between 0 and `decimal-maximum-digits` are shown. It is invalid to specify both `decimal-digits` and `decimal-maximum-digits`. Formats other than `decimal` ignore `decimal-digits`.
+-   `decimal-maximum-digits` - when format is `decimal`, a number that controls the maximum number of digits shown to the right of the decimal separator. This has the same effect as `decimal-digits` except that trailing zeros are omitted. Possible values are from 0 to 20. It is invalid to specify both `decimal-digits` and `decimal-maximum-digits`. Formats other than `decimal` ignore `decimal-maximum-digits`.
```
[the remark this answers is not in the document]
m-akinc: I've changed it mostly the way you had it, except that I think it's important to point out that `decimal-digits` does not default to 2 when `decimal-maximum-digits` is specified.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: A problem I ran into was the return type of `Unit.getFormatter()`. My best guess is that it should be this:
```TS
abstract class Unit extends FoundationElement {
    public abstract getFormatter(): typeof UnitScaleFormatter;
}
```
But because `UnitScaleFormatter` is abstract, I get the TS error "Cannot create an instance of an abstract class" anywhere I do this:
```TS
const FormatterClass = el.getFormatter();
const formatter = new FormatterClass(lang, opts);
```

rajsite: You use the `new` keyword for typing constructors. Can see examples in the mixins. [playground example](https://www.typescriptlang.org/play?#code/FASwdgLgpgTgZgQwMZQAQFUwggZSQgGygDEB7GAWwQmhgHkAHCEUsAZ1QG9hVfUA3QgC5UbCDHABzANzAAvsAQAjMTGQRUSAgjYdM2PIRLkqNWFx58krVQFckEcgApSTFuxH7c+ImUrVaRmYbAEouBT5UZVV1VDgTaicQkVUpWQUtHQ4AIQBPaD9TWlQoAA9oMAATPSxvI0KA825I+P8IJJTxKQtIyJgoCFsYMFQAcgALKAICUlHZSIUFCFyGNC9DXwSzGABhVnF7RxhUAF5UMCgAd1QXNxtPWo3jNsC79jCTgD4MR59notgsms7A0kgGDW2pxuyR+Bj+ENoe0gMEO5FO32afH6g2GqDyBS2tHSwGBYlQCPMZzBEApMCSQJsGlaAOOZwu11pTk4ggIIgmIFGchCDPYpCIADoZpInMzGjBxbL2iFhcAgA)

rajsite: Not sure how the response meant this was resolved so re-opening.
The response didn't sound like a resolution but continued discussion... I'm not sure what the intention of resolving it was.

m-akinc: Yeah, based on the short chat we had about it last week, my take was that we didn't need to settle on implementation details in this thread/PR, as long as it was understood that the API I had in the spec was not the decided upon approach -- just an example. I am still making an effort to use your suggested approach in my prototype.

rajsite: Gotcha, my concerns where:
- the comment sounded like something was tried and didn't work but didn't say conversation would continue anywhere else
- my follow-up response with an example was not acknowledged, the thread was just resolved without discussion

Would be helpful to make sure it's clear how a thread is resolved (thumbs up or comment, etc).

rajsite (approved):

atmgrifter00 (approved):

PR #1598, opened 2023-10-13
Duration text column

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1569

## 👩‍💻 Implementation

Fairly straightforward table column implementation, modeled closely to others like the `TableColumnNumberFormat` column. Main thing of interest is the new `DurationFormatter`, which uses several distinct `Intl.NumberFormat` instances to produce the appropriate, localizable value and its unit for each relevant time section of the duration.

## 🧪 Testing

Added unit tests. Created a new page object for the `TableColumnDurationText` that was identical in purpose to the `TableColumnDateTextPageObject`, so I created a new abstract base class to contain the shared logic between the two page object classes.

## ✅ Checklist


packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+    {
+        firstName: 'Montgomery',
+        lastName: 'Burns',
+        swearWordCadence: 3.78e12
+    }
+];
+
+const metadata: Meta<SharedTableArgs> = {
+    title: 'Incubating/Table Column - Duration Text',
```
m-akinc, line 47:
```suggestion
    title: 'Components/Table Column: Duration Text',
```

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export default metadata;
+
+interface TextColumnTableArgs extends SharedTableArgs {
+    fieldName: string;
+    checkValidity: () => void;
+    validity: () => void;
+}
+
+const durationTextColumnDescription = 'The `nimble-table-column-duration-text` column is used to display duration fields as text in the `nimble-table`. The duration values must be of type `number` and each represent a total number of milliseconds. Values that represent more than 100 days or less than a millisecond will be formatted in scientific notation as seconds only. All sub-second values will be represented with up to 3 digits of precision.';
```
m-akinc, line 77: If we end up supporting negative values, the language should be updated to say "values with a magnitude of". If we don't support negatives, that should be documented here.

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+                story: durationTextColumnDescription
+            }
+        }
+    },
+    // prettier-ignore
+    render: createUserSelectedThemeStory(html<TextColumnTableArgs>`
+        ${incubatingWarning({
+        componentName: 'table',
+        statusLink: 'https://github.com/orgs/ni/projects/7/views/21'
+    })}
```
m-akinc, line 92:
```suggestion
```

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+        >
+            <${tableColumnTextTag}
+                field-name="${x => x.fieldName}"
+            >
+            Name
+            </${tableColumnTextTag}>
+            <${tableColumnDurationTextTag}
+                field-name="swearWordCadence"
+            >
+            Last swear word
```
m-akinc, line 105: nit:
```suggestion
            Time since last swear word
```

packages/nimble-components/src/table-column/duration-text/cell-view/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { TableColumnTextCellViewBase } from '../../text-base/cell-view';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-table-column-duration-text-cell-view': TableColumnDurationTextCellView;
+    }
+}
+
+/**
+ * A cell view for displaying date/time fields as text
```
m-akinc, line 17:
```suggestion
 * A cell view for displaying duration fields as text
```

packages/nimble-components/src/table-column/duration-text/group-header-view/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import { styles } from '../../text-base/group-header-view/styles';
+import type { TableColumnDurationTextColumnConfig } from '..';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-table-duration-text-group-header': TableColumnDurationTextGroupHeaderView;
+    }
+}
+/**
+ * The group header view for displaying date/time fields as text.
```
m-akinc, line 14:
```suggestion
 * The group header view for displaying duration fields as text.
```

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+        ...sharedTableArgs(simpleData)
+    }
+};
+
+export default metadata;
+
+interface TextColumnTableArgs extends SharedTableArgs {
+    fieldName: string;
+    checkValidity: () => void;
+    validity: () => void;
```
m-akinc, line 74:
```suggestion
```

m-akinc (approved):

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,86 @@
...
+            notation: 'scientific',
+            maximumFractionDigits: 3
+        });
+    }
+
+    public format(value: number | null | undefined): string {
+        if (
+            value === null
+            || value === undefined
+            || value < 0
```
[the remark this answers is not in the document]
atmgrifter00: I'm not sure. Ultimately, a difference between two timestamps can always be represented as a positive value, it just depends on which timestamp you subtract from which. Representing that difference as a negative value I don't think has any intrinsic value over representing it as the positive version of the difference, and semantically it just seems to make more sense as a positive value. A "duration" is a period of time _since_ a start time (which always moves forward). Like saying "the time since 0 seconds has been -2 seconds" is nonsensical.

I can certainly be convinced otherwise, but at the moment I don't think I feel compelled to support negative values.

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,86 @@
...
+        const minutes = Math.floor((value / 60000) % 60);
+        if (minutes) {
+            const formattedMinutes = this.minutesFormatter.format(minutes);
+            result.push(formattedMinutes);
+            remainingTime -= 60000;
+        }
+
+        const valueInSeconds = remainingTime / 1000;
+        const seconds = valueInSeconds === 0 ? 0 : valueInSeconds % 60;
+        if (value < 1 && valueInSeconds !== 0 && result.length === 0) {
```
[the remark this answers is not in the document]
atmgrifter00: Not quite. So, for starters, `value` can both be less than 1 _and_ 0, and if it _is_ 0, we don't want to use the scientific formatter. That said, there _is_ an unnecessary check here, which is the `result.length === 0` check, which will _always_ be true when `value < 1` (by virtue of the preceding logic in the function). Note that we want to use the `valueInSeconds !== 0` check instead of `milliseconds !== 0` to handle the case when `milliseconds` is actually a really small value that results in a rounded value of `0` for `valueInSeconds`.

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        const result = [];
+        const fractionalDays = value / 86400000;
+        let remainingTime = value;
+        const days = Math.floor(fractionalDays);
+        if (days <= 100 && days > 0) {
+            const formattedDays = this.daysFormatter.format(days);
+            result.push(formattedDays);
+            remainingTime -= 86400000;
```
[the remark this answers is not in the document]
atmgrifter00: You are right, however I couldn't think of a new test that would have caught the incorrect previous algorithm. Ultimately, I think the previous algorithm worked because of how it is ultimately always taking a modded value to work from, so it didn't matter what multiple of each "time segment" was used to subtract from the remaining time.

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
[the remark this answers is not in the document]
m-akinc: I'm not sure why I wasn't considering the possibility that `value` was 0. Makes sense.

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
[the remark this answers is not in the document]
m-akinc: I don't have a concrete use case, but a user's data contains columns for timestamp `a`, timestamp `b`, and `b` - `a`, then you can end up with negative intervals if some `b`s preceed their `a`. Since I don't have a particular use case in mind, though, I'm satisfied with only supporting non-negative values now (and potentially adding support later if requested).

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.spec.ts
```diff
@@ -0,0 +1,310 @@
...
+        expect(
+            document.createElement('nimble-table-column-duration-text')
+        ).toBeInstanceOf(TableColumnDurationText);
+    });
+
+    it('reports column configuration valid', () => {
+        expect(column.checkValidity()).toBeTrue();
+    });
+
+    describe('displays blank when', () => {
```
jattasNI, line 79: Could you move many of these tests to be next to `duration-formatter.ts` instead? That way we know the formatter is being tested if someone is using it independently of the column. Then the column tests could just verify things that require the element.

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export default metadata;
+
+interface TextColumnTableArgs extends SharedTableArgs {
+    fieldName: string;
+}
+
+const durationTextColumnDescription = 'The `nimble-table-column-duration-text` column is used to display duration fields as text in the `nimble-table`. The duration values must be of type `number` and each represent a positive total number of milliseconds. Values that represent more than 100 days or less than a millisecond will be formatted in scientific notation as seconds only. All sub-second values will be represented with up to 3 digits of precision.';
+
```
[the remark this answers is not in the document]
atmgrifter00: I prefer the latter.

jattasNI (approved):

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        const result = [];
+        const millisecondsPerSecond = 1000;
+        const millisecondsPerMinute = millisecondsPerSecond * 60;
+        const millisecondsPerHour = millisecondsPerMinute * 60;
+        const millisecondsPerDay = millisecondsPerHour * 24;
+        const fractionalDays = milliseconds / millisecondsPerDay;
+        let remainingTime = milliseconds;
+        const days = Math.floor(fractionalDays);
+        if (days <= 100 && days > 0) {
```
rajsite, line 55: nit: Can we capture the `100` days as a constant?

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,93 @@
...
+            const formattedDays = this.daysFormatter.format(days);
+            result.push(formattedDays);
+            remainingTime -= days * millisecondsPerDay;
+        } else if (days > 100) {
```
rajsite, line 59: Was there a discussion on where 100 days came from? Maybe we wait for feedback and don't reopen now but I think durations in the couple year category would be pretty possible and "600 days" would be more useful than the value in seconds. I'd advocate for a value like 999 (the rollover to 4 digits) as the big number to abandon ship.

packages/nimble-components/src/table-column/duration-text/cell-view/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private cellRecordChanged(): void {
+        this.updateText();
+    }
+
+    private updateText(): void {
+        if (this.columnConfig) {
+            this.text = this.columnConfig.formatter.format(this.cellRecord?.value)
+                ?? '';
```
rajsite, line 34: Doesn't the formatter always return a string?

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
```diff
@@ -0,0 +1,93 @@
...
+            unit: 'second',
+            notation: 'scientific',
+            maximumFractionDigits: 3
+        });
+    }
+
+    public format(milliseconds: number | null | undefined): string {
+        if (
+            milliseconds === null
+            || milliseconds === undefined
```
rajsite, line 40: could simplify with a `typeof milliseconds !== 'number'` check

rajsite (approved):

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -0,0 +1,111 @@
...
+        firstName: 'Ned',
+        lastName: 'Flanders',
+        swearWordCadence: Number.NaN
+    },
+    {
+        firstName: 'Montgomery',
+        lastName: 'Burns',
+        swearWordCadence: 3.78e12
+    }
+];
```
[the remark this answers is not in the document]
atmgrifter00: Ultimately, I think this has to do with the typing of the `Table` in the storybook context, which is just typed to `TableRecord` in the [`SharedTableArgs`](https://github.com/ni/nimble/blob/260ad74bc906cc5b0b3861f9df376593ce3ce732/packages/nimble-components/src/table-column/base/tests/table-column-stories-utils.ts#L6). I thin k to make the `as const` part to work, we would need to create an interface that extends `TableRecord` and type both the data and the table to that interface. It seems like that would involve either a severance from the shared mechanisms between the table stories or a more advanced API for the shared part, and I'm not sure what the ROI is.

I'll go ahead and create a tech debt issue to track it though.

packages/nimble-components/src/table-column/duration-text/models/duration-formatter.ts
atmgrifter00: Sold!

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
atmgrifter00: No push back on decision to not support negative values.

PR #1605, opened 2023-10-16
Allow Renovate to create PRs for vulnerabilities

jattasNI:
# Pull Request

## 🤨 Rationale

As part of #1604 we noticed there are several dependencies with non-critical vulnerabilities. Renovate claims to be able to fix these automatically.

## 👩‍💻 Implementation

Enable [Renovate vulnerabilityAlerts setting](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts). We had previously disabled it to be conservative and reduce noise. 

## 🧪 Testing

The docs say it will create PRs immediately, ignoring our schedule setting (currently monthly). Once this goes in we'll see what it does. Hopefully something similar to #1603.

## ✅ Checklist

### DIFF
--- a/angular-workspace/package.json
+++ b/angular-workspace/package.json
@@ -57,6 +57,6 @@
     "ng-packagr": "^14.1.0",
     "playwright": "^1.30.0",
     "rollup": "^3.10.1",
-    "typescript": "~4.7.4"
+    "typescript": "~4.8.2"
   }
 }
--- a/angular-workspace/projects/ni/nimble-angular/CHANGELOG.json
+++ b/angular-workspace/projects/ni/nimble-angular/CHANGELOG.json
@@ -2,6 +2,21 @@
   "name": "@ni/nimble-angular",
   "entries": [
     {
+      "date": "Mon, 16 Oct 2023 21:31:42 GMT",
+      "version": "17.6.2",
+      "tag": "@ni/nimble-angular_v17.6.2",
+      "comments": {
+        "patch": [
+          {
+            "author": "beachball",
+            "package": "@ni/nimble-angular",
+            "comment": "Bump @ni/nimble-components to v20.6.3",
+            "commit": "not available"
+          }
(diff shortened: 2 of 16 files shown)


.github/renovate.json
```diff
@@ -8,9 +8,6 @@
   "schedule": [
     "monthly"
   ],
```
rajsite, line 10: Is there a way to configure it to only look at production dependencies? And what severity level does it care about?

Our current npm ci reports 19 vulnerabilities: https://github.com/ni/nimble/actions/runs/6537395490/job/17751139631?pr=1605#step:7:67

Is merging this going to create 19 PRs?

rajsite: I see the default strategy will update the lock file only so it should be able to merge immediately if it can: https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts

aight lets try it out

rajsite: Oh wait, should we do in-range-only? https://docs.renovatebot.com/configuration-options/#rangestrategy

Sounds like the default of update-lockfile might create package.json updates if needed which then need change files. Guess we can try either.

rajsite (approved):

.github/renovate.json
jattasNI: I think I'll keep it at update-lockfile since that's what we use for our existing Renovate configuration for non-security updates. I'm hoping it'll bias towards only touching the lockfile but will still create package.json updates if needed. But if it gets too aggressive about touching the package file then I agree in-range-only would be a good idea.

jattasNI, line 10: FYI the [Renovate instructions](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts) say that we should enable Dependabot alerts in our repo settings. I will do this when I complete the PR (won't turn on Dependabot updates though).

https://github.com/ni/nimble/settings/security_analysis


image

PR #1611, opened 2023-10-18
Add decimal-maximum-digits attr to number-text column

m-akinc:
# Pull Request

## 🤨 Rationale

This feature is needed to eventually render file sizes in the Nimble table in a way consistent with how SLE renders them today. Specifically, it renders them with one decimal place, except if it would be a zero, in which case it renders without any decimal places.

## 👩‍💻 Implementation

The `Intl.NumberFormat` used by the number-text column has options `maximumFractionDigits` and `minimumFractionDigits`. We had always been setting both options to the value of `decimal-digits` set by the client. Now, if `decimal-maximum-digits` is set rather than `decimal-digits`, we use that for `maximumFractionDigits` and pass 0 for `minimumFractionDigits`.

Additionally:

- The default alignment of a `decimal`-formatted column is not always `right` anymore. If `decimal-maximum-digits` is set, then the default alignment is `left`, because decimal points are no longer guaranteed to align.
- `decimal-maximum-digits` is mutually exclusive with `decimal-digits`, and a validation error is set if both are set.
- A validation error is set if the value of `decimal-maximum-digits` is set outside the range of 0-20.
- Updated Storybook documentation
- Added unit tests
- Updated Blazor wrapper and added one test
- Updated Angular directive and added tests

## 🧪 Testing

Unit tests pass

## ✅ Checklist


m-akinc:
@msmithNI do you mind buddying this?

msmithNI (approved):

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
```diff
@@ -120,6 +120,7 @@ describe('DecimalFormatter', () => {
                 () => {
                     const formatter = new DecimalFormatter(
                         locale,
```
jattasNI, line 122: I think we should have test coverage in this file for the case where min and max digits are different.

jattasNI (approved):

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -171,7 +174,13 @@ export const numberTextColumn: StoryObj<NumberTextColumnTableArgs> = {
...
             name: 'decimal-digits',
             description:
-                "The number of decimal places to format values to when the column's `format` is configured to be `decimal`. When not configured, a default value of `2` is used. The value must be in the range 0 - 20 (inclusive)."
+                "The number of decimal places to format values to when the column's `format` is configured to be `decimal`. If neither `decimal-digits` or `decimal-maximum-digits` are set, a default value of `2` is used. `decimal-digits` and `decimal-maximum-digits` cannot both be set at the same time. The value must be in the range 0 - 20 (inclusive)."
+        },
+        decimalMaximumDigits: {
+            name: 'decimal-maximum-digits',
+            description:
+                "The maximum number of decimal places to format values to when the column's `format` is configured to be `decimal`. This differs from `decimal-digits` in that trailing zeros are omitted. `decimal-digits` and `decimal-maximum-digits` cannot both be set at the same time. The value must be in the range 0 - 20 (inclusive).",
+            control: { type: 'none' }
```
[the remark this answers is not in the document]
m-akinc: I can use a select for both properties.

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -111,7 +130,12 @@ export class TableColumnNumberText extends TableColumnTextBase {
             case NumberTextFormat.decimal:
                 return new DecimalFormatter(
                     lang.getValueFor(this),
-                    this.decimalDigits ?? defaultDecimalDigits
+                    typeof this.decimalMaximumDigits === 'number'
+                        ? 0
+                        : this.decimalDigits ?? defaultDecimalDigits,
+                    this.decimalMaximumDigits
+                        ?? this.decimalDigits
+                        ?? defaultDecimalDigits
```
atmgrifter00, line 138: Maybe move the logic for these last two parameters into some consts above the `return` statement?

atmgrifter00, line 138: Is there an issue if `decimalMaximumDigits` is less than `defaultDecimalDigits` (i.e. `decimalMaximumDigits` is 1). It seems that would result in the formatter being given a value for `maximumFractionDigits` that is smaller than `minimumFractionDigits`.

packages/nimble-blazor/NimbleBlazor/Components/NimbleTableColumnNumberText.razor.cs
```diff
@@ -54,8 +54,14 @@ public partial class NimbleTableColumnNumberText : NimbleTableColumn, IFractiona
     public NumberTextAlignment? Alignment { get; set; }
 
     /// <summary>
-    /// The number of digits to display after the decimal place when the column's format is decimal.
+    /// The number of digits to display after the decimal place when the column's format is decimal. Mutually exclusive with DecimalMaximumDigits.
```
atmgrifter00, line 57: It might be helpful to make it clear that setting both `DecimalDigits` and `DecimalMaximumDigits` results in an invalid state (instead of just saying "Mutually exclusive" for both.

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/index.ts
m-akinc: No, because if `decimalMaximumDigits` is defined (`typeof this.decimalMaximumDigits === 'number'`), we always pass `0` as the minimum.

atmgrifter00: Oh right...duh. 🤦‍♂️

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -107,11 +126,20 @@ export class TableColumnNumberText extends TableColumnTextBase {
...
 
     private createFormatter(): NumberFormatter {
+        let minimumDigits;
+        let maximumDigits;
         switch (this.format) {
             case NumberTextFormat.decimal:
+                minimumDigits = typeof this.decimalMaximumDigits === 'number'
+                    ? 0
+                    : this.decimalDigits ?? defaultDecimalDigits;
+                maximumDigits = this.decimalMaximumDigits
```
rajsite, line 136: Can't these just be const?
```suggestion
                const minimumDigits = typeof this.decimalMaximumDigits === 'number'
                    ? 0
                    : this.decimalDigits ?? defaultDecimalDigits;
                const maximumDigits = this.decimalMaximumDigits
```

rajsite (approved):

packages/nimble-components/src/table-column/number-text/index.ts
m-akinc: No, it results in this eslint violation: https://eslint.org/docs/latest/rules/no-case-declarations

rajsite: @m-akinc re-opening as the linked doc shows the correct implementation. Surround it in a block.

image

m-akinc: Duh, of course. This was re-opened after the PR was merged, so I'll add this change to my next PR.

Issue #296, opened 2022-01-28
nimble-card Component


m-akinc:
## 😯 Problem to Solve

Need to create a new Nimble component to display other components within a card. This was requested by the following clients but could be useful in other places:
1. Transportation BU (1st screenshot below; contact Jonathan Meyer or Deborah Bryant for more info).
2. [SLE landing page](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/258105) (2nd screenshot below; contact Fred Visser, Mark Black, or Michael Brandt for more info)

## 💁 Proposed Solution

- FAST component: https://explore.fast.design/components/fast-card
- Visual Design: not requested yet

![image](https://user-images.githubusercontent.com/7282195/151621674-4135f8af-03fb-4733-a715-cff616a867df.png)
![image](https://user-images.githubusercontent.com/10500124/151622134-c9c197dc-3a6f-4d92-8436-f07cb4741378.png)
![image](https://github.com/ni/nimble/assets/10500124/7fabefbf-b9e6-4dc7-bd63-22acd606b3eb) from [this BTS wireframe](https://lucid.app/lucidspark/6669223f-be64-404c-9374-6971288459b5/edit?page=0_0&invitationId=inv_1b26b2fa-159b-4ce0-b696-e5ad27f6f2b7)

### Requirements/features
- visual styling like drop shadows
- slots to include arbitrary content
- future: higher level card layouts as a mobile-friendly representation of a data grid?

## 📋 Tasks

- [x] nimble-card visual design issue
- [x] nimble-card visual design spec
- [x] nimble-card spec
- [x] nimble-card custom element
- [x] nimble-card unit tests
- [x] nimble-card Storybook tests
- [x] nimble-card Storybook docs
- [x] nimble-card Angular wrapper
- [x] nimble-card Blazor wrapper
- [x] Add to Angular example app
- [x] Add to Blazor example app
- [x] Update Component Status table in README and incubating status
- [ ] nimble-card pattern/usage documentation

jattasNI:
Remaining open questions are around usage guidance for cards vs accordions. Should cards ever be collapsible? If so, would they just be replaced by accordion or are there still independent purposes? Fred can start driving this as he has bandwidth and works more on usage docs in the coming months.

PR #1618, opened 2023-10-23
Spec for nimble-card component

kjohn1922:
# Pull Request

## 🤨 Rationale

We want to introduce a nimble-card component for the new Routines UX in SLE. See the initial feature HLD here:
[Feature Design: No code Routines creation](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=%2Fdocs%2Fdesign-documents%2FRoutines%2FNo-code-Routines-creation.md&version=GBmaster&_a=preview)

This type of component has previously been requested for other use cases as well:
[GitHub Issue #296: nimble-card component](https://github.com/ni/nimble/issues/296)

This is a draft of a spec doc

## 👩‍💻 Implementation

Added README under `nimble-components/src/card/specs`.

See prototype branch here: https://github.com/ni/nimble/tree/users/kelseyj/card-prototype

## 🧪 Testing

N/A

## ✅ Checklist

~~- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.~~


packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+-   _Styling: Does FAST provide APIs to achieve the styling in the visual design spec?_
+    -   No additional requirements expected here.
+-   _Testing: Is FAST's coverage sufficient? Should we write any tests beyond Chromatic visual tests?_
+    -   No additional requirements
+-   _Documentation: Any requirements besides standard Storybook docs and updating the Example Client App demo?_
+    -   No additional requirements
+-   _Tooling: Any new tools, updates to tools, code generation, etc?_
+    -   No additional requirements
+-   _Accessibility: keyboard navigation/focus, form input, use with assistive technology, etc._
+    -   We will set the ARIA role to `region`, based on the intended usage of the component. It may make sense to have clients set
```
mollykreis, line 93: A few thoughts here on the ARIA roles:

- It doesn't look like the [fast-card](https://github.com/microsoft/fast/blob/b78c921ec4e49ec9d7ec980f079ec114045df42e/packages/web-components/fast-foundation/src/card/card.template.ts) or [mat-card](https://material.angular.io/components/card/examples) apply a role to their elements. What's the rationale for deciding that the nimble-card should have a role?
- Since the fast-card doesn't apply an ARIA role, if nimble is going to automatically set a role on the component, we'll need to fork their template for this component. That should be called out in this document (assuming I'm understanding your proposal correctly).
- If we use the `region` role, then it is also required for the element to have a label. We should make sure it's easy to for clients to set this label and avoid having accessibility issues within their apps.
- Do you have any thoughts on when a client would need to use a different role (such as `group` as you mentioned) rather than `region`? We've tried to avoid making clients apply ARIA roles on our components, so we should try to understand if there really is a use case for different roles. If there are, we should see if there is a way to encapsulate that within the nimble component.

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+component. See the [Angular Material `mat-card` component](https://v5.material.angular.io/components/card/overview)
+for an example. We could add similar content in the future, but it is not required at this time.
+
+Since we're not providing any built-in content, the API surface area for the `nimble-card` is very small. The template
+contains only a slot, and there are no configurable properties on the component. Here is an example usage for
+the `nimble-card` component:
+
+```
+<nimble-card>
+    <div>
```
mollykreis, line 34: I wouldn't expect the content of the `nimble-card` to be wrapped in a single element (i.e. the `div` here). Can the `div` be removed to make the `h2`, `p`, and `nimble-button` direct children of the `nimble-card`?

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+The `nimble-card` will not include any built-in content (title, footer, actions, etc.). Some component libraries
+do provide a card component with built-in, pre-styled content or pre-styled child components to use inside the card
+component. See the [Angular Material `mat-card` component](https://v5.material.angular.io/components/card/overview)
+for an example. We could add similar content in the future, but it is not required at this time.
+
+Since we're not providing any built-in content, the API surface area for the `nimble-card` is very small. The template
+contains only a slot, and there are no configurable properties on the component. Here is an example usage for
+the `nimble-card` component:
+
+```
```
mollykreis, line 32: nit: Add the code snippet type will give syntax highlighting in the rendered markdown

```suggestion
```html
```

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+    <div>
+        <h2>Heading</h2>
+        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>
+        <nimble-button>Button</$nimble-button>
+    </div>
+</nimble-card>
+```
+
+While we don't have a finalized visual design spec, we expect the style will include a background color and a
+border, potentially with rounded corners. We will add design tokens for the background and border. The GitHub
```
mollykreis, line 43: > We will add design tokens for the background and border.

Unless we know that the background and border will require different design tokens than our existing nimble components, we should hold off on adding new tokens.

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+    -   _(none)_
+-   _Methods_
+    -   _(none)_
+-   _Events_
+    -   _(none)_
+-   _CSS Classes and Custom Properties that affect the component_
+    -   width
+    -   height
+    -   elevation
+        -   We don't currently plan to provide different styles based on elevation. This could be added in the future.
```
mollykreis, line 65: I see that these were pulled over from the FAST docs (which ordinarily is a reasonable thing to do), but it doesn't look like they actually have any special support via CSS classes or custom properties for width, height, or elevation. Therefore, I'd just remove these and set the value to "none" like you have in the other cases.

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+We are designing the `nimble-card` component to address that use case.
+
+---
+
+## Design
+
+The `nimble-card` will not include any built-in content (title, footer, actions, etc.). Some component libraries
+do provide a card component with built-in, pre-styled content or pre-styled child components to use inside the card
+component. See the [Angular Material `mat-card` component](https://v5.material.angular.io/components/card/overview)
+for an example. We could add similar content in the future, but it is not required at this time.
```
mollykreis, line 26: It's interesting that you call having additional built-in content as not required at this time, but your code example includes a header (via `h2` element) and the figma document linked has headers in each section. I'm leaning towards saying we should have a "title" slot that makes it easy to put a title in the section with the correct font, size, color, spacing, etc.

That suggestion does require creating our own copy of the component's template, but in this case, I think the deviation from what FAST provides is probably worth it in order to help enforce consistency across usages. However, since we don't have designs or UX guidance on how the card would be used more broadly, maybe it's better to track this in the "Open Issues" section for now. That'll give other reviewers/owners a chance to weigh in.

change/@ni-nimble-components-6aa0f191-427b-4745-9425-f602170a1266.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add spec for nimble-card, no code changes yet",
+  "packageName": "@ni/nimble-components",
+  "email": "kelsey.johnson@ni.com",
```
mollykreis, line 5: Most of us don't expose our NI email address in the change files. If you don't want to have your address here, you can follow the [instructions in the CONTRIBUTING doc](https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#beachball-change-file) to configure github to keep your email address private.

change/@ni-nimble-components-6aa0f191-427b-4745-9425-f602170a1266.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add spec for nimble-card, no code changes yet",
```
mollykreis, line 3: nit: Since the "type" already is "none", you don't need to mention in the comment that there aren't any code changes.

```suggestion
  "comment": "Add spec for nimble-card",
```

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,111 @@
...
+        <nimble-button>Button</$nimble-button>
+    </div>
+</nimble-card>
+```
+
+While we don't have a finalized visual design spec, we expect the style will include a background color and a
+border, potentially with rounded corners. We will add design tokens for the background and border. The GitHub
+issue mentions several use cases for a card component that might call for different visual treatments, but we
+are only solving for the Routines use case at this time (grouping related configuration in a slide-out). To
+support other use cases in the future, we could potentially add appearance variants or make use of the `elevation`
```
mollykreis, line 46: > or make use of the `elevation` property of the component.

I mentioned this in a different comment, but I didn't actually see an `elevation` property on the component. I'd take this last part of the sentence off and just leave it at "To
support other use cases in the future, we could potentially add appearance variants."

packages/nimble-components/src/card/specs/README.md
kjohn1922: I hadn't thought about the necessity of forking the template to set the role, but that makes sense, I see why that would be necessary.

I got the idea of setting the role from the `mat-card` documentation here: https://material.angular.io/components/card/overview#group-region-and-landmarks

> There are several ARIA roles that communicate that a portion of the UI represents some semantically meaningful whole. Depending on what the content of the card means to your application, you can apply one of [role="group"](https://www.w3.org/TR/wai-aria/#group), [role="region"](https://www.w3.org/TR/wai-aria/#region), or [one of the landmark roles](https://www.w3.org/TR/wai-aria/#landmark) to the <mat-card> element.
>
>You do not need to apply a role when using a card as a purely decorative container that does not convey a meaningful grouping of related content for a single subject. In these cases, the content of the card should follow standard practices for document content.

The reasoning there seemed like it would apply to our `card` usage as well.

Reading more into [group](https://www.w3.org/TR/wai-aria/#group), though, I actually can't think of a case where it would make sense to have a card to use that role. Also looking at [region](https://www.w3.org/TR/wai-aria/#region), these containers may not rise to the level of "sufficiently important that users will likely want to be able to navigate to the section easily and to have it listed in a summary of the page".

kjohn1922: I'll defer to your expertise here. Just semantically, it makes sense to my mind that a container for grouping related configuration would be a "region", but I also think it would not be worth forking the fast-card template just for the sake of setting the role, when they made a determination that the card did not need a role.

packages/nimble-components/src/card/specs/README.md
kjohn1922: Ah, you're correct, the `div` isn't necessary. I was seeing some unexpected layout behavior without the `div` in my prototype, but it looks like that was the result of some other things I was messing with in the styles.

packages/nimble-components/src/card/specs/README.md
kjohn1922: That's fair. Can you tell me more about when it is appropriate to create design tokens? It wasn't clear to me from the documentation.

change/@ni-nimble-components-6aa0f191-427b-4745-9425-f602170a1266.json
kjohn1922: Fixed. Thanks for the tip!

packages/nimble-components/src/card/specs/README.md
kjohn1922: Removed.

packages/nimble-components/src/card/specs/README.md
mollykreis: We create component-specific design tokens when a given "global" token doesn't apply. For example, we have tokens such as `ni-nimble-border-hover-color` that applies to a number of different controls. So, in that case we don't have tokens for `ni-nimble-button-border-hover-color`, `ni-nimble-select-border-hover-color`, `ni-nimble-combobox-border-hover-color`, etc.

In this case, I'm wondering whether or not our existing `ni-nimble-section-background-color` makes sense for this card (or possibly the card is transparent with just a border indicating the grouping). This may or may not require a card-specific token, but that'll be based on a visual designer telling us that this will be a different pattern than other components. Same for the border color -- we may use an existing token like `ni-nimble-divider-background-color`, or it may be a new color.

packages/nimble-components/src/card/specs/README.md
kjohn1922: The h2 in the example is just because I was copying the fast-card's example content in the explorer: https://explore.fast.design/components/fast-card

I would also lean toward adding a title, though. I think it makes sense and adds a lot of value to the card. In the initial design conversations, Mark and Katie suggested that we not have it because there might be cases where we'd want to use the card without a title. Seems like it would be easy enough to support that possibility by making the title optional, though.

I'll add it to the Open Issues for now, to have more discussion.

mollykreis: That sounds good. We have some components (dialog and banner at least) that have a title that can optionally be hidden. Depending on where we land on the ARIA role, being able to set a title, even if it isn't visually shown, will be important to make sure the card as an accessible label.

packages/nimble-components/src/card/specs/README.md
mollykreis: I agree with your assessment that I'm not sure of a case where we'd want to set the role to `group`. It's also hard to say whether or not usages of the card will be "sufficiently important". I personally lean towards just doing what the fast-card is doing unless we have a compelling reason to diverge.

kjohn1922: That's reasonable. I removed the sentence about setting a role. We'll follow the fast-card's example and leave the nimble-card without a role.

packages/nimble-components/src/card/specs/README.md
kjohn1922: I added an item under open issues about this question.

I also updated the [prototype](https://github.com/ni/nimble/tree/users/kelseyj/card-prototype) to have a title slot (no title-hidden option, I'll save that for if we decide to do it).

packages/nimble-components/src/card/specs/README.md
kjohn1922: Thanks for the information, that makes sense. We'll wait for visual design input before adding new tokens. (The prototype branch has tokens for background and border, but that was just so I could learn how they work.)

I added a sentence about not creating tokens yet.

mollykreis (approved):

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,116 @@
+# Nimble Card
```
jattasNI, line 1: We should get the team's opinions on calling this "card" vs "container" or something else. We can always change our minds and rename it but let's make our best guess now.

My initial thought was that we might decide to have different components:
- "container" is meant for grouping related controls. One might think of it as a non-collapsible [accordion](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/accordion/specs/IxD.md).
- "card" is meant for showing summary or status information

But in a quick scan through a few other design systems I only saw cards. Maybe that (plus FAST's name choice) is enough to go with card to start with, but I'd like to do a bit more reading and see if others have thoughts

kjohn1922: I went through a similar thought process. We called it a "container" in the feature definition, but looking at other design systems, "card" seems like a common name for a display container grouping related content, whether that content is configuration or information. That's why I started calling it a card, although obviously I'm still open to suggestions. Container also makes sense for our use case.

The "configuration container" and "info card" use cases could be appearance variants of the same Card control. Having it as two separate components might lead to confusion about when to use which one, especially if both controls end up with their own appearance variants.

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The `nimble-card` is a container that is designed to contain arbitrary content specified by a client
+application. The `nimble-card` is intended for grouping related content.
+
+### Background
+
+[GitHub Issue #296: nimble-card component](https://github.com/ni/nimble/issues/296)
+
+An IxD and Visual Design spec for the `nimble-card` component have not yet been provided. We will add some
+preliminary styling in the interim.
```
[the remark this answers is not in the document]
kjohn1922: Will do, I'll add a sentence to the spec. I have the component status marked as incubating in my prototype already. For the task list -- is there a defined procedure for keeping that up to date? I'm not that familiar with GitHub workflows. Is it just a matter of checking the boxes?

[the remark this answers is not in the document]
jattasNI: Yep, just checking the boxes is the simplest way to do it. It doesn't need to happen with every PR but it's good to do at least at feature / cycle boundaries.

If you want to get more automated, there's a button next to each item in the task list to convert it to a github issue. Then you could link those issues to PRs and use the [special keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to automatically resolve the issues when the corresponding PRs complete.

packages/nimble-components/src/card/specs/README.md
jattasNI: I looked at a few more design systems today and didn't see a compelling consensus about containers and cards being different from each other. So I'm ok with picking one term and I'm ok with that term being "card".

Let's leave the thread open for other reviewers but if no one objects we can resolve this with no changes.

packages/nimble-components/src/card/specs/README.md
jattasNI: I agree that an optional title seems useful.

jattasNI (approved):

atmgrifter00 (approved):

packages/nimble-components/src/card/specs/README.md
kjohn1922: I updated the spec to say that we will include an optional title. Since we're including a title, I assume we also want to set `aria-label` to that title and and add a `title-hidden` attribute (similar to nimble-banner), so I added those to the spec as well.

mollykreis: It doesn't sound like we're giving the `nimble-card` an ARIA role, so we don't actually have anything to label with the title (no role on the card means that it won't be part of the accessibility tree, and ARIA labels provide a label for items in the accessibility tree).

packages/nimble-components/src/card/specs/README.md
```diff
@@ -0,0 +1,114 @@
...
+-   _Documentation: Any requirements besides standard Storybook docs and updating the Example Client App demo?_
+    -   No additional requirements
+-   _Tooling: Any new tools, updates to tools, code generation, etc?_
+    -   No additional requirements
+-   _Accessibility: keyboard navigation/focus, form input, use with assistive technology, etc._
+    -   If the user supplies a title, we will set `aria-label` to that value.
+    -   The `nimble-card` itself does not receive keyboard focus and tabbing jumps to the first focusable component inside the `nimble-card`.
+-   _Mobile: small screens, touch interactions, mobile-specific integrations_
+    -   No additional requirements. By default, the `nimble-card` fits its height to its content and grows/shrinks horizontally to
+        fit its parent container. Other mobile-friendly considerations are up to the client and the settings on the child components.
```
rajsite, line 102: Sounds like this is saying it's default width will be 100%. I think a more expected default would be `auto`, i.e. it'll size to it's contents instead of filling the space of the parent container. i.e. `display: flex` / `display: block` default behavior

rajsite (approved):

packages/nimble-components/src/card/specs/README.md
rajsite: Is it semantically like a section? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

kjohn1922: Ok, that makes sense. I removed the note about a label and the title-hidden/accessibly-hidden details, based on our conversation.

rajsite: I think we would still need title-hidden (I made an update to my comment so maybe that was missed)
i.e. visually if you don't want the header visible. Or are you saying with a card the header is always visible?

kjohn1922: Sorry, my "that makes sense" was to Molly -- I hadn't refreshed the page to see your comment. The idea behind removing "title-hidden" was that, without the need to set the accessibility attributes, we can achieve collapsing/hiding the title completely without an attribute for it, if there is no content for the title slot.

rajsite: Think I'm advocating for it to show up in the accessibility tree as a section. If that's the direction, we would want users to always put something in the title slot so some context is given to the accessibility tree of this visual group of stuff (even if visually the title is hidden).

packages/nimble-components/src/card/specs/README.md
kjohn1922: Since there were no objections to this, I'm resolving this issue. Thanks for the discussion!

packages/nimble-components/src/card/specs/README.md
kjohn1922: This was describing the behavior I was seeing with the[ `flex-card` example](https://explore.fast.design/components/fast-card), which I believe is using `display: block`. They set an explicit width in the example, and if you remove the `width` then the card stretches horizontally to fill its parent.

In my prototype, both `display: flex` and `display: block` also make the card stretch horizontally to fill the parent container. If we want the default to be `auto` behavior, I can set `${display('inline-block')}` in the style. Is that what you're suggesting?

rajsite: Ah yea, brain fart, block level elements would try and fill the width.
I was suggestion inline-block but after taking a look at the openui card research page: https://open-ui.org/components/card.research/
looks like most design systems are treating cards as block level elements rather than inline-block, so I'll retract the statement. A block level layout like flex would be good 👍

packages/nimble-components/src/card/specs/README.md
kjohn1922: I updated the spec again, to include the title slot. After further discussion, we don't want to include the title-hidden property for now. We don't intend to provide styling for the no-title case and it's possible the designers will say it always needs a title in the end, so we can wait to provide the option to hide the title until we have an actual use case for it.

Issue #1551, opened 2023-09-19
Migrate to the parameterize and parameterizeNamedList test helpers


rajsite:
## 🧹 Tech Debt

The PR #1379 introduced new patterns for writing parameterized tests which addresses a bunch of short comings with the existing patterns and deprecated the older `getSpecTypeByNamedList`.

This tech debt task should:
- Migrate all existing usages of `getSpecTypeByNamedList` to `parameterizeNamedList`
- Migrate tests in such a way as they take advantage of the strict typing (using `as const` on the test cases) and consistent names for the test function parameters (`spec`, `name`, and `value`)
- Remove the deprecated `getSpecTypeByNamedList` helpers

We also have parameterized tests in the nimble-angular package. These tests generally just are written with `forEach` (as opposed to the deprecated `getSpecTypeByNamedList`, which is internal to nimble-components). We should also look at making `parameterize`/`parameterizeNamedList` exported from nimble-components so it can be used for our Angular tests too.

PR #1619, opened 2023-10-23
Angular/Blazor integration for duration table column

msmithNI:
# Pull Request

## 🤨 Rationale

Angular/Blazor integration for duration table column. Part of #1598 

## 👩‍💻 Implementation

- Standard Angular/Blazor integration. The duration column has no inputs specific to that column type so the API is just the shared properties for grouping/column widths/etc.
- Added new `DurationPipe` at new `pipes` entrypoint in `nimble-angular`
- Include duration column in nimble-components `all-components` bundle (Blazor example apps use that bundle so that's when we usually catch omissions in it)

## 🧪 Testing

- Tested in local builds
- Added autotests
- Added duration column to component demo page in Angular/Blazor example apps

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+    name: 'duration'
+})
+export class DurationPipe implements PipeTransform {
+    private readonly durationFormatter;
+
+    public constructor(@Inject(LOCALE_ID) locale: string) {
+        this.durationFormatter = new DurationFormatter(locale);
+    }
+
+    public transform(timeInMilliseconds: string | number | null | undefined): string {
```
msmithNI, line 17: Note: I made this pipe take both `string` and `number` since that's what the [ElapsedTimePipe does](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/src/pipes/elapsed-time.pipe.ts&version=GBmaster&line=18&lineEnd=19&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), but I don't feel strongly about it.

I wanted to make it easier for existing `ElapsedTimePipe` usages to switch to this, but in practice it won't be a drop-in replacement, since our duration formatter takes milliseconds as input, whereas ElapsedTimePipe takes seconds.

I also thought about adding a secondary `transformSeconds()` method that took seconds as an input. That would let usages like [in sl-test-results-grid](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemLinkShared/projects/systemlink-test-results-lib/src/components/sl-test-results-grid/sl-test-results-grid.component.ts&version=GBmaster&line=165&lineEnd=166&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) to just call a single method and not need to multiply by 1000. However once the switchover to `sl-table` is complete and they're using `nimble-table-column-duration-text`/`sl-table-column-duration-text` directly, they'd still need to convert their model values to milliseconds explicitly anyway. So long term I'm not sure a `transformSeconds` method would be that useful.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
```diff
@@ -0,0 +1,29 @@
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
+import { DurationFormatter } from '@ni/nimble-components/dist/esm/table-column/duration-text/models/duration-formatter';
+
+/**
+ * A pipe that transforms a number of milliseconds (string/number) into a string formatted like "1 day, 2 hr, 3 min, 4.567 sec"
+ */
+@Pipe({
+    name: 'duration'
+})
+export class DurationPipe implements PipeTransform {
```
msmithNI, line 10: Note: I just followed [the HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-column-specs/table-column-duration-text-hld.md?plain=1#L81) for the pipe name and entrypoint of `pipes`.

msmithNI:
@m-akinc can you buddy this PR?

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { DurationPipe } from './duration.pipe';
+
+describe('DurationPipe', () => {
+    describe('basic formatting (English)', () => {
+        let pipe: DurationPipe;
+
+        beforeEach(() => {
+            pipe = new DurationPipe('en');
```
m-akinc, line 8: Could be reused by all tests, since it should not have any state.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+                expected: '1 day, 2 hr, 3 min, 4.56 sec'
+            },
+            {
+                name: '1d 2h 3m 4.56s (as string "93784560")',
+                value: '93784560',
+                expected: '1 day, 2 hr, 3 min, 4.56 sec'
+            },
+        ];
+
+        testCases.forEach(test => {
```
m-akinc, line 34: Could consider using `parameterizeNamedList` from `nimble-components`.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
m-akinc, line 1: Do we want this in a `tests` directory?

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+    private readonly durationFormatter;
+
+    public constructor(@Inject(LOCALE_ID) locale: string) {
+        this.durationFormatter = new DurationFormatter(locale);
+    }
+
+    public transform(timeInMilliseconds: string | number | null | undefined): string {
+        if (timeInMilliseconds == null) {
```
m-akinc, line 18: Does this imply that `DurationFormatter.format` handles `undefined`, but not `null`? Just seems a little unexpected that we're only special-casing `null`.

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
```diff
@@ -115,6 +118,8 @@ public class SimpleTableRecord
         public int StatusCode { get; }
         public string Result { get; }
         public double Number { get; }
+
+        public double Duration { get; }
```
m-akinc, line 122:
```suggestion
        public double Duration { get; }
```

m-akinc (approved):

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
msmithNI: Updated

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
msmithNI: Looked at `DurationFormatter` again and it handles both `null` and `undefined` (just not `string`), so removed this `if` check.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
msmithNI: I'll wait for @rajsite to weigh in on this one. The Readme beside `parameterizeNamedList` says "to be used only by nimble-component library authors", but since this is within the same repo it might be fine.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
msmithNI: Moved to new `tests` subdirectory.

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
msmithNI: Updated

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
+import { DurationFormatter } from '@ni/nimble-components/dist/esm/table-column/duration-text/models/duration-formatter';
+
+/**
+ * A pipe that transforms a number of milliseconds (string/number) into a string formatted like "1 day, 2 hr, 3 min, 4.567 sec"
+ */
+@Pipe({
+    name: 'duration'
+})
+export class DurationPipe implements PipeTransform {
```
jattasNI, line 10: It would be good to mention the pipe in our docs somewhere. The simplest option would be to add an Angular section to the duration column storybook (then we'd do the same for the numeric column unit pipes later). I guess another option would be a dedicated story for Angular pipes. I'd be fine with that if someone strongly prefers it but I'd vote for adding it to the table column for now.

angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+            },
+            {
+                name: '1d 2h 3m 4.56s (as string "93784560")',
+                value: '93784560',
+                expected: '1 day, 2 hr, 3 min, 4.56 sec'
+            },
+        ];
+
+        testCases.forEach(test => {
+            it(`${test.name} transforms to ${test.expected}`, () => {
```
jattasNI, line 31: Can you use the [parameterize pattern](https://github.com/ni/nimble/blob/6db2e6b2d558cdda6457ba8744ab52f8055dd8eb/packages/nimble-components/src/utilities/tests/parameterized.ts#L73)? Makes it easier to focus/skip out a test.

jattasNI (approved):

atmgrifter00 (approved):

angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
jattasNI: Just saw the same comment in a different thread: https://github.com/ni/nimble/pull/1619#discussion_r1369357665

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
rajsite: > but since this is within the same repo it might be fine

Nah, boundary interface between projects should be what is exposed from packages. We shouldn't have hidden coupling between packages  just because they are in the same repo (nimble-angular should be able to be developed outside the repo without changes).

The options would be:

1. Move it to [`testing`](https://github.com/ni/nimble/tree/main/packages/nimble-components/src/testing) and maybe add a comment on the function saying it's internal. Create a tech debt issue to find the right spot to put it.
2. Copy and paste it to the angular lib. Create a tech debt issue to find the right spot and remove the duplicate.
3. Create a more concrete proposal for sharing it. Right now it seems like it could be as a new package as part of the style guide. See https://github.com/ni/javascript-styleguide/issues/125#issuecomment-1631682110

I think I lean to option 2 surprisingly even to me. I'm a bit concerned that option 1 would lead to nimble-components dependencies being added for the wrong reason and maybe even weird circular ones (like tokens depending on components just for the test helper). But option 1 also sounds good if it can be relatively short-term. Thoughts @jattasNI ?

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
rajsite: I'd vote for adding it to the duration column storybook as usage docs. Looks like an mdx page would be needed.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
jattasNI: I agree that 2 is a better interim place to be than 1 and that 3 shouldn't be part of this PR, so 2 is the best of those options.

I guess there's also option 0: leave it as-is and file tech debt to create a shared library. Then we'd update this along with every other place that doesn't use parameterize today.

I think 2 puts us in a better place but if it blows up the scope I don't mind 0.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
rajsite: > I made this pipe take both string and number

Looks like Angular's numbery pipes do that too so seems inline with common patterns:
https://angular.io/api/common/DecimalPipe#input-value
https://angular.io/api/common/PercentPipe#input-value
https://angular.io/api/common/CurrencyPipe#input-value

> I also thought about adding a secondary transformSeconds()

That wouldn't actually implement the behavior of the PipeTransform interface right? Think it would have to be a separate pipe that maybe extends / uses this one. But yea even if something like that is useful it would be systemlink-specific and on that side of the fence

angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { DurationPipe } from '../duration.pipe';
+
+describe('DurationPipe', () => {
+    describe('basic formatting (English)', () => {
+        const pipe = new DurationPipe('en');
+
+        const testCases = [
+            {
```
rajsite, line 8: think we should have the ieee754 edge case tests -0, 0, NaN, Inf, -Inf, 

maybe also some boundary values max safe int, min safe int, max value, min value, epsilon

And the string behavior is new here, should probably exercise that more thoroughly. Angular seems pretty strict about not actual numbers being passed in. We should probably follow their behavior: https://github.com/angular/angular/blob/16.2.10/packages/common/src/pipes/number_pipe.ts#L293

rajsite (approved):

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
msmithNI: Correct, the secondary method wouldn't be part of the PipeTransform API. On the SLE side, some code instantiates the pipe directly and calls `transform()` from TypeScript code, so it would work for that purpose.  
But good point that we could add something on the SystemLink side if needed. So I'll keep this as-is.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.spec.ts
msmithNI: I think my preference is not to copy/paste `parameterize` in this PR. The copied version would either be test-less, or we'd need to copy/paste the tests too. Since it's being used from the `pipes` entrypoint, I guess it would need to be exported from the `internal-utilities` entrypoint. We would also need to tweak the `tsconfig` files to allow the Jasmine types for the file it was in.

I updated #1551 to say that we should look at sharing `parameterize` with our Angular tests too (and noted that those just use `forEach` not `getSpecTypeByNamedList`).

angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
msmithNI: Added additional tests for numeric edge cases, and some additional string tests.

It looks like the Angular pipe [returns null](https://github.com/angular/angular/blob/16.2.10/packages/common/src/pipes/number_pipe.ts#L250) for `null` and `undefined`, but I think returning empty string as this pipe currently does makes more sense for our use cases (it's consistent with `DurationFormatter`, and the behavior of other pipes we have in SLE).  
Non-numeric string values will turn into empty string (`parseFloat` turns them into `NaN`, `DurationFormatter` makes `NaN` empty string). So since non-numeric strings are covered, and `transform` input is typed as `number|string|null|undefined`, I don't think there's any case we need to `throw` for.

angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
msmithNI: Added mdx page for the duration column with Usage Docs/Angular Usage/Duration Pipe section.

PR #1620, opened 2023-10-23
Unit support for number-text column

m-akinc:
# Pull Request

## 🤨 Rationale

See #1573 

## 👩‍💻 Implementation

Clients configure a unit for a number-text column by providing a unit element (e.g. `nimble-unit-byte`) as content. The unit element indicates which unit-specific `UnitScale` to use. The `UnitScale` is used by the column's formatter to pick an appropriate scaled unit for a given number, and it provides a `UnitFormatter` to do the ultimate formatting.

There are two categories of `UnitScale`s: those that rely on `Intl.NumberFormat` to provide translated unit labels, and those that rely on translations built directly into Nimble.

Initial unit elements are `nimble-unit-byte` (which can be configured to use either 1000-based or 1024-based units) and `nimble-unit-volt`. The intent was to implement at least one of each category of `UnitScale` to serve as examples for future unit implementations.

## 🧪 Testing

Unit tests created.

## ✅ Checklist


m-akinc:
Sorry, this PR is much bigger than I would like. I could not come up with a great way to piece it apart, after the fact.

m-akinc:

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,96 @@
...
+        lang: string,
+        formatterOptions: Intl.NumberFormatOptions
+    ): ScaledUnit[];
+
+    protected override format(number: number): string {
+        const unit = this.pickBestUnit(number);
+        return unit.format(number / unit.conversionFactor);
+    }
+
+    private pickBestUnit(number: number): ScaledUnit {
```
mollykreis, line 47: It looks like every number that gets formatted calls this function twice -- once when calling `getValueForBestUnit` and once when calling `format`. Instead, is there a way to rewrite part of this code so that doesn't have to happen. In most cases, it seems pretty straightforward, but the `alwaysUseBaseUnit` code path might make things too complicated.

packages/nimble-components/src/table-column/number-text/models/unit-prefix.ts
```diff
@@ -0,0 +1,26 @@
...
+ * A prefix that represents a scaling factor when applied to a base unit
+ */
+export class UnitPrefix {
+    public constructor(
+        public readonly factor: number,
+        public readonly text: string
+    ) {}
+}
+
+// The same prefixes are used for English, French, German, Japanese, and Chinese (all currently supported languages).
```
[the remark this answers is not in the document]
m-akinc: I confirmed with Eiko from localization that all the abbreviated units for voltage (e.g. mV, cV, dV, kV, ...) were the same as English for the other four languages we support. I don't think that is unique to voltage. From my experimenting, I see that `Intl.NumberFormat` does use "kg" for kilograms in Chinese:
image
Did you maybe pass "gram" as the unit? `Intl.NumberFormat` does weird things (like, for some languages, giving "1k" instead of "1,000") if you only pass the base unit.

I did see that using Google translate for "kg" in Chinese produced "公斤". However, it looks like that's just the the translation for the word "kilogram". The [Wikipedia page for SI units in Chinese](https://zh.wikipedia.org/wiki/%E5%9B%BD%E9%99%85%E5%8D%95%E4%BD%8D%E5%88%B6) shows the same unit abbreviations and prefix symbols as the English page. That by itself doesn't guarantee that they use SI representations, but the page also has a picture of a highway sign with distances in "km". In any case, based on what `Intl.NumberFormat` does and what Eiko said, I'm pretty confident that we're using reasonable (if not exclusively used) representations.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
m-akinc: > is there a way to rewrite part of this code so that doesn't have to happen. In most cases, it seems pretty straightforward...

I may be dense, but I'm not seeing a straightforward way. I really dislike that I had to add `getValueForBestUnit()` to the API and run `pickBestUnit()` twice, but it seems necessary. The only thing I can think of is to keep a cache of calculated best units for given input numbers, but that is too much. Let me know if you had something else in mind.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -152,6 +177,12 @@ export const numberTextColumn: StoryObj<NumberTextColumnTableArgs> = {
             <${tableColumnNumberTextTag} field-name="favoriteNumber" format="${x => NumberTextFormat[x.format]}" alignment="${x => NumberTextAlignment[x.alignment]}" decimal-digits="${x => x.decimalDigits}" decimal-maximum-digits="${x => x.decimalMaximumDigits}">
                 Favorite Number
             </${tableColumnNumberTextTag}>
+            <${tableColumnNumberTextTag} field-name="measurement" format="${x => NumberTextFormat[x.format]}" alignment="${x => NumberTextAlignment[x.alignment]}" decimal-digits="${x => x.decimalDigits}">
```
mollykreis, line 180: I think this column should also use the `decimalMaximumDigits` from the story.

mollykreis (approved):

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
mollykreis: The case in the `decimal` formatter seems fairly straightforward, but I agree that the `default` formatter definitely makes things more difficult.

I don't have a huge issue with the code as-is as long as it isn't causing performance problems. It just seems like there is a lot of work being done in order to render a single cell value.

packages/nimble-components/src/units/volt/index.ts
```diff
@@ -0,0 +1,28 @@
+import { DesignSystem } from '@microsoft/fast-foundation';
+import { template } from '../base/template';
+import { UnitScale } from '../base/unit-scale';
+import type { UnitScaleFormatterConstructor } from '../../table-column/number-text/models/unit-scale-formatter';
+import { VoltScaleFormatter } from '../../table-column/number-text/models/volt-scale-formatter';
```
rajsite, line 5: number-text isn't responsible for all formatters, just shared base classes. formatters provided by the unit elements should be organized with them
`/units/volt/models/volt-scale-formatter`

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,96 @@
...
+        Object.defineProperty(this, 'supportedUnits', {
+            value: supportedUnits,
+            writable: false,
+            configurable: false
+        });
+        Object.defineProperty(this, 'baseUnit', {
+            value: baseUnit,
+            writable: false,
+            configurable: false
+        });
```
rajsite, line 93: This seems really wonky, why aren't we doing a normal lazy loading pattern like (conceptually, didn't try to make actual implementation):

```ts
    private get supportedUnits(): ScaledUnit[] {
        if (this._supportedUnits === undefined) {
            this.setSupportedUnitsAndBaseUnit();
        }
        return this._supportedUnits;
    }
```

packages/nimble-components/src/table-column/number-text/models/byte-scale-formatter.ts
```diff
@@ -0,0 +1,70 @@
+import { IntlNumberFormatScaledUnit } from './intl-number-format-scaled-unit';
+import type { ScaledUnit } from './scaled-unit';
+import { UnitScaleFormatter } from './unit-scale-formatter';
+
+/**
+ * Formatter for numbers with byte units (1000-based)
+ */
+export class ByteScaleFormatter extends UnitScaleFormatter {
```
rajsite, line 8: There are some basic naming consistency issues around all these. Like this is arbitrarily leaving off `Unit` I'd vote for `ByteUnitScaleFormatter`. The thing that this is is not a "ScaleFormatter" (there isn't such a thing), it is a "UnitScaleFormatter"

packages/nimble-components/src/table-column/number-text/models/no-unit-scale-formatter.ts
```diff
@@ -0,0 +1,20 @@
+import { IntlNumberFormatScaledUnit } from './intl-number-format-scaled-unit';
+import type { ScaledUnit } from './scaled-unit';
+import { UnitScaleFormatter } from './unit-scale-formatter';
+
+/**
+ * Degenerate UnitScaleFormatter for formatting without units
+ */
+export class NoUnitScaleFormatter extends UnitScaleFormatter {
```
rajsite, line 8: No is a weird description. What kind of unit is this, a "No" unit. I'd vote for AbsentUnitScaleFormatter or EmptyUnitScaleFormatter or UnitlessUnitScaleFormatter

packages/nimble-components/src/table-column/number-text/models/volt-scale-formatter.ts
```diff
@@ -0,0 +1,25 @@
+import { ManuallyTranslatedUnitScaleFormatter } from './manually-translated-unit-scale-formatter';
+import { UnitPrefix, metricPrefixes } from './unit-prefix';
+import { UnitTranslation } from './unit-translation';
+
+/**
+ * Formatter for numbers with voltage units
+ */
+export class VoltScaleFormatter extends ManuallyTranslatedUnitScaleFormatter {
```
rajsite, line 8: I'd vote for this being VoltUnitScaleFormatter, it being ManuallyTranslatted is an implementation detail, not what this is implementing (a UnitScaleFormatter)

packages/nimble-components/src/table-column/number-text/models/tests/unit-scale-formatter.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+                }
+            };
+        });
+
+        protected override getSupportedUnits(): ScaledUnit[] {
+            return this.units;
+        }
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-loop-func
```
rajsite, line 101: There isn't a loop here... can you double check the eslint disables?

rajsite (commented):
Some initial high-level feedback

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
m-akinc: This was just an implementation I saw suggested when I was searching for whether JS had a standard type/pattern for lazy loading. We can just use private fields instead, though.

packages/nimble-components/src/table-column/number-text/models/tests/manually-translated-unit-scale-formatter.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        );
+        return translations;
+    }
+
+    protected override getSupportedPrefixes(): UnitPrefix[] {
+        return [new UnitPrefix(5, '5.'), new UnitPrefix(10, '10.')];
+    }
+}
+
+describe('ManuallyTranslatedUnitScaleFormatter', () => {
```
jattasNI, line 24: Would it be worthwhile to add a test that you can provide just the language portion of the locale but not the region portion? e.g (`fr` instead of `fr-FR`)

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -55,6 +63,14 @@ export class TableColumnNumberText extends TableColumnTextBase {
     })
     public decimalMaximumDigits?: number;
 
+    /** @internal */
```
jattasNI, line 66: I'm still wrapping my head around this PR so might have missed something, but are there tests for the logic in this file? I was expecting changes in `table-column-number-text.spec.ts` to validate things like adding/removing/changing attributes of unit elementsand the new case for left/right default alignment.

m-akinc: Good catch. I missed adding some tests there. I've added cases for:
- adding/removing a unit element
- setting an attribute of a unit element
- configuring a unit in default formatting mode
- a few new alignment scenarios

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,91 @@
...
+        lang: string,
+        formatterOptions: Intl.NumberFormatOptions
+    ): ScaledUnit[];
+
+    protected override format(number: number): string {
+        const unit = this.pickBestUnit(number);
+        return unit.format(number / unit.conversionFactor);
+    }
+
+    private pickBestUnit(number: number): ScaledUnit {
```
rajsite, line 53: Think we can add a shortcut case for this.supportedUnits.length === 1

rajsite (commented):
Some more incremental feedback

packages/nimble-components/src/unit/base/unit-scale.ts
```diff
@@ -0,0 +1,9 @@
+import { FoundationElement } from '@microsoft/fast-foundation';
+import type { UnitScaleFormatterConstructor } from '../../table-column/number-text/models/unit-scale-formatter';
+
+/**
+ * An element representing a set of related units which could be used to represent the same value
+ */
+export abstract class UnitScale extends FoundationElement {
```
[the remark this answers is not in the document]
rajsite: Based on other comments I'd push for `Unit`

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,91 @@
...
+                return unit;
+            }
+        }
+        return this.supportedUnits[this.supportedUnits.length - 1]!;
+    }
+
+    // This is an implemenation of lazy initialization.
+    // Ideally, we could initialize supportedUnits and baseUnit in the constructor,
+    // but they depend on an abstract method and potentially other state that isn't
+    // available until the derived class is finished being constructed.
```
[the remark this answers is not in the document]
m-akinc: I'm leaving the part that explains why we need to lazy load in the first place.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,91 @@
...
+            || this.alwaysUseBaseUnit
+        ) {
+            return this.baseUnit;
+        }
+        for (const unit of this.supportedUnits) {
+            if (magnitude >= unit.conversionFactor) {
+                return unit;
+            }
+        }
+        return this.supportedUnits[this.supportedUnits.length - 1]!;
```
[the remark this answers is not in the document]
m-akinc: I'm not sure it's worth it when we'll be dealing with arrays on the order of a dozen elements (especially since there's apparently no built-in JS implementation), but I guess we can. We do run `pickBestUnit` twice for each cell, so maybe that's enough reason.

jattasNI (commented):
Another batch of feedback. About halfway through now.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
rajsite: yea... you're right, with small sets the optimizations may not always pan out. May not be worth it unless we measure.

We should probably have a test with a large grid of numbers like at least 10x10 visible and rendering (if not much more like excel spreadsheet level) and make sure there are no regressions and that there aren't any specific bottlenecks happening.

packages/nimble-components/src/table-column/number-text/models/manually-translated-unit-scale-formatter.ts
```diff
@@ -0,0 +1,62 @@
...
+        super(lang, formatterOptions);
+        this.unitTranslations = this.getUnitTranslations();
+        this.supportedPrefixes = this.getSupportedPrefixes();
+        if (!this.unitTranslations.get('en')) {
+            throw new Error('English translations must exist');
+        }
+    }
+
+    protected override getSupportedUnits(
+        lang: string,
```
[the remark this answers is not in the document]
rajsite: > I'm trying to figure out if the lang parameters in this class hierarchy should actually be called locale instead

If we get technical its a full bcp47 descriptor, which is referred to as a language-tag and is more than just script and region:
image

JS tends to use the term locale for the strings, HTML tends to use lang for them. Nimble codebase seems to use `lang`  (187) more often than locale (87).

Maybe since most of these objects are wrappers around the Intl.NumberFormat apis and that uses `locale` then we can stick to locale for the Formatters and UnitScale objects. But stuff that touches the elements and the lang DesignToken should use lang.

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -102,6 +140,22 @@ export class TableColumnNumberText extends TableColumnTextBase {
         this.updateColumnConfig();
     }
 
+    private unitElementsChanged(): void {
+        void this.updateColumnConfigFromUnitElements();
+    }
+
+    private async updateColumnConfigFromUnitElements(): Promise<void> {
```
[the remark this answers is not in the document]
m-akinc: I guess it's more thorough to put the table in an invalid state, so I'll do that.

jattasNI (commented):
Nearly there, just a few more files

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -102,6 +140,14 @@ export class TableColumnNumberText extends TableColumnTextBase {
         this.updateColumnConfig();
     }
 
+    private unitElementsChanged(): void {
+        this.unitScale = this.unitElements.find(
```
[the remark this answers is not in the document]
m-akinc: Moved it into a utility function called `waitUntilCustomElementsDefinedAsync`.

packages/nimble-components/src/table-column/number-text/models/manually-translated-unit-scale-formatter.ts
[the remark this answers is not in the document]
m-akinc: > Taking a step back, what use cases do we want to claim support for today or might we want to support in the future?
>
>1. specifying different unit translations for different regions (e.g. en-us meter vs en-ca metre)
>2. specifying different value formatting for different regions (I can't think of an example but it seems plausible since other values like dates are region-specific)
>3. falling back to a default translation for a language when the client requests a locale for that language but a different region (e.g. making sure fr-ca resolves to our fr or fr-fr translations and not our en translations)

We can support all three. I've made the necessary changes to support 1 (2 and 3 were already supported) and added test cases.

packages/nimble-components/src/unit/byte/models/tests/byte-1024-unit-scale-formatter.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { Byte1024UnitScaleFormatter } from '../byte-1024-unit-scale-formatter';
+
+describe('Byte1024UnitScaleFormatter', () => {
```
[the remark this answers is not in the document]
m-akinc: > Could we add cases beyond the range of PiB like you did for volts?

Yes. Added.

> do we have unit-agnostic test coverage that ensures invalid values formatted as units display correctly?

Kind of. With or without units, the number portion gets formatted the same way. We have existing tests (for the DecimalFormatter and DefaultFormatter) that validate the formatting of interesting values, like NaN, -0, Infinity, etc. But none of those values format as an empty string. So if you format NaN without a unit, it's "NaN"; with the bytes unit, it's "NaN bytes". But we don't explicitly test for the latter, since the unit's presence doesn't affect the formatting of the number.

packages/nimble-components/src/unit/byte/tests/unit-byte.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    it('should export its tag', () => {
+        expect(unitByteTag).toBe('nimble-unit-byte');
+    });
+
+    it('can construct an element instance', () => {
+        expect(document.createElement('nimble-unit-byte')).toBeInstanceOf(
+            UnitByte
+        );
+    });
+
```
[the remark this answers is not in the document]
m-akinc: The 'returns ByteScaleFormatter when "binary" is false' test was _almost_ testing that case, except that I was explicitly setting the `binary` property to `false`. I changed the test setup, so that I'm instead just leaving the `binary` attribute off. Now it covers the case you mentioned.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
m-akinc: Do you mean an automated performance test? That would be our first one, and I'm not sure how we would capture the rendering time. Or were you suggesting a manual test via a Storybook story or addition to the example app?

More generally, we probably need to have some performance-based acceptance testing of the table and all its various features at some point. Not sure if that would be automated, manual, or a mix.

change/@ni-nimble-components-e9fe99af-bb91-40c3-80e3-f5b41e2b0ff6.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I imagine it will be common for teams to want to contribute their own units so I think it would be useful to have some documentation to point them to. It doesn't have to be comprehensive but it would be good to at least be able to point them at an example implementation / checklist / overview.

Could you take a stab at curating this? Could be in a follow up PR if that's easier. I'm not too picky about where it lives as long as it's linkable but some options include:
- update the spec to ensure all this info is up to date with the current implementation and gathered in one section
- add a CONTRIBUTING doc in one of these folders and link to it from the spec and/or root CONTRIBUTING doc and/or the column docs in storybook

packages/nimble-components/src/unit/byte/models/tests/byte-1024-unit-scale-formatter.spec.ts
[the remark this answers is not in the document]
jattasNI: How about the cases that result in an empty number like `undefined` and `null`? Will those show " bytes" or empty string? I'd advocate for the latter (and for test coverage that validates it).

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
rajsite: Example app is not a test bed so that's not the right spot.
A story could be possible.

My specific concern with this PR is avoiding regressions to numeric performance as is. Want to make sure if you are not using units that we aren't paying extra overhead. Think the general concern is separate and true.

change/@ni-nimble-components-e9fe99af-bb91-40c3-80e3-f5b41e2b0ff6.json
m-akinc: I've added a section to the spec with instructions for implementing a new unit.

packages/nimble-components/src/unit/byte/models/tests/byte-1024-unit-scale-formatter.spec.ts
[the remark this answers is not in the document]
m-akinc: Yes, that is handled just by deriving from `NumberFormatter`, but I've added testing for it to `unit-scale-formatter.spec.ts` anyway.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
m-akinc: I added a new story that is a 30x20 table of numbers. We can check the responsiveness of this table for obvious regressions when we make significant changes to the table or number column. To compare against pre-unit performance, I added the same story in my local `main` branch. The good news is there's no percievable difference. The bad news is that even before my change, upon initial load, the table and the storybook controls (i.e. unit selection radio buttons) are unresponsive for about five seconds. When I pause execution during the unresponsive period, the call stack is in `axe.js`, so my guess is that Storybook's accessibility plug-in is to blame.

packages/nimble-components/src/table-column/number-text/models/tests/unit-scale-formatter.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            const formatter = new TestUnitScaleFormatter('', {});
+            const formattedValue = formatter.formatValue(value.value);
+            expect(formattedValue).toEqual(value.expectedFormattedValue);
+        });
+    });
+
+    parameterizeNamedList(testCases, (spec, name, value) => {
+        spec(`${name} (getValueForBestUnit)`, () => {
+            const formatter = new TestUnitScaleFormatter('', {});
+            const formattedValue = formatter.getScaledNumber(value.value);
```
atmgrifter00, line 112:
```suggestion
            const scaledValue = formatter.getScaledNumber(value.value);
```

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public constructor(
+        private readonly locale: string,
+        private readonly formatterOptions: Intl.NumberFormatOptions
+    ) {
+        super();
+    }
+
+    public getScaledNumber(number: number): number {
```
atmgrifter00, line 38: Just out of curiosity, what is the reasoning behind making this a public method? If it's just for tests, could you mark it `internal`?

jattasNI (approved):

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
m-akinc: No, it is called from `DecimalFormatter` and `DefaultFormatter`. Those formatters make decisions based on what the ultimate displayed value will be, so they need to ask the `UnitScaleFormatter` for it.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -227,6 +227,18 @@ class UnitVolt extends FoundationElement implements UnitScale {
...
+    -   If it is not, your formatter will be responsible for providing translated unit strings for a set of languages you choose to support. Your formatter should extend `ManuallyTranslatedUnitScaleFormatter`. You must implment `getUnitTranslations()`, which will return a `Map` of [language tags](https://en.wikipedia.org/wiki/IETF_language_tag) to `UnitTranslation` objects. If you need different translations for the same language, based on region (e.g. "meter" for `en-US` vs. "metre" for `en-CA`), you may provide a locale string that includes the region (e.g. `en-CA`) as a key into your `Map`. You must also implement `getSupportedPrefixes()`, which returns an array of `UnitPrefix`. These prefixes define which scaled units (e.g. "kg", "mm", "GB") your unit is capable of converting to/displaying. For a standard set of SI prefixes, you may return `metricPrefixes`. See `src/unit/volt/models/volt-unit-scale-formatter.ts` for an example.
```
atmgrifter00, line 238:
```suggestion
    -   If it is not, your formatter will be responsible for providing translated unit strings for a set of languages you choose to support. Your formatter should extend `ManuallyTranslatedUnitScaleFormatter`. You must implement `getUnitTranslations()`, which will return a `Map` of [language tags](https://en.wikipedia.org/wiki/IETF_language_tag) to `UnitTranslation` objects. If you need different translations for the same language, based on region (e.g. "meter" for `en-US` vs. "metre" for `en-CA`), you may provide a locale string that includes the region (e.g. `en-CA`) as a key into your `Map`. You must also implement `getSupportedPrefixes()`, which returns an array of `UnitPrefix`. These prefixes define which scaled units (e.g. "kg", "mm", "GB") your unit is capable of converting to/displaying. For a standard set of SI prefixes, you may return `metricPrefixes`. See `src/unit/volt/models/volt-unit-scale-formatter.ts` for an example.
```

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
rajsite: > When I pause execution during the unresponsive period, the call stack is in axe.js, so my guess is that Storybook's accessibility plug-in is to blame.

The axe plugin can be disabled for stories that create a lot of DOM, see matrix test configs: https://github.com/ni/nimble/blob/a844a186ce955d3f25ea1b58f469f70ff378916a/packages/nimble-components/src/utilities/tests/matrix.ts#L7

> The good news is there's no perceivable difference.

I'd like a stronger indication that there is no change in performance with and without the change. The described test doesn't give much confidence for that.

One idea is to add a mode to the story that updates the data set on a short interval. That way when running a performance profile the initial load is not captured and multiple executions can be sampled. I'd do something that pregenerate a couple data sets that can be cycled between.

Also open to other ideas to give a more confident measurement of that rendering perf.

atmgrifter00 (approved):

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
m-akinc: I added a play function to the story that swaps data every second. Unfortunately, I couldn't use a story argument to turn this on/off, because I couldn't get the current argument values in the play function. Anyway, I copied the same changes to main for comparision. It looks like both with and without my changes, it takes about 25 ms to render. (The whole 30x20 table was visible.)
image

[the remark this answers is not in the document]
m-akinc: > The axe plugin can be disabled for stories that create a lot of DOM, see matrix test configs:

It turned out that we weren't actually disabling the a11y plugin anywhere, because we were setting `disabled` rather than `disable`. I fixed those.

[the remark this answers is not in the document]
rajsite: Thanks for putting that together! That layout pass is pretty darn long, almost a full frame of time to calculate layout is surprising but not introduced by this change so looks good 👍

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
```diff
@@ -125,33 +128,47 @@ describe('DecimalFormatter', () => {
...
+                    value.expectedFormattedValue[locale]
+                );
+            });
+        });
     }
+
+    describe('with unit', () => {
+        class TestUnitScaleFormatter extends UnitScaleFormatter {
+            public override getSupportedScaledUnits(): ScaledUnit[] {
+                return [1, 2, 4].map(scaleFactor => {
```
rajsite, line 152: I feel like I would expect a whole lot more test cases similar to the decimal formatter test cases, particularly around rounding scenarios + scaleFactor boundaries

For example, 999.999 with 0 decimal digits will round up to 1000, on metric uunits will this render as 1K?

packages/nimble-components/src/unit/base/models/manually-translated-unit-scale-formatter.ts
```diff
@@ -0,0 +1,65 @@
...
+                    commonPluralRules,
+                    label,
+                    label
+                )
+            );
+        }
+        return supportedUnits;
+    }
+
+    protected abstract getUnitTranslations(): Map<string, UnitTranslation>;
```
rajsite, line 63: This should be a ReadonlyMap type

packages/nimble-components/src/unit/volt/models/volt-unit-scale-formatter.ts
```diff
@@ -0,0 +1,25 @@
+import { ManuallyTranslatedUnitScaleFormatter } from '../../base/models/manually-translated-unit-scale-formatter';
+import { UnitPrefix, metricPrefixes } from '../../base/models/unit-prefix';
+import { UnitTranslation } from '../../base/models/unit-translation';
+
+/**
+ * Formatter for numbers with voltage units
+ */
+export class VoltUnitScaleFormatter extends ManuallyTranslatedUnitScaleFormatter {
+    protected override getUnitTranslations(): Map<string, UnitTranslation> {
+        const unitTranslations = new Map<string, UnitTranslation>();
```
rajsite, line 10: We can make this map once since it's all constants. Can use a static initialization block to set it up: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks

Or even more direct use the array syntax to initialize with a single constant: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/Map#creating_a_new_map

packages/nimble-components/src/unit/byte/models/byte-1024-unit-scale-formatter.ts
```diff
@@ -0,0 +1,31 @@
+import { ManuallyTranslatedUnitScaleFormatter } from '../../base/models/manually-translated-unit-scale-formatter';
+import { UnitPrefix } from '../../base/models/unit-prefix';
+import { UnitTranslation } from '../../base/models/unit-translation';
+
+/**
+ * Formatter for numbers with byte units (1024-based)
+ */
+export class Byte1024UnitScaleFormatter extends ManuallyTranslatedUnitScaleFormatter {
+    protected override getUnitTranslations(): Map<string, UnitTranslation> {
+        const unitTranslations = new Map<string, UnitTranslation>();
```
rajsite, line 10: same as the other lets make more statics and readonly typings

packages/nimble-components/src/unit/base/models/unit-prefix.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export class UnitPrefix {
+    public constructor(
+        public readonly factor: number,
+        public readonly text: string
+    ) {}
+}
+
+// The same prefixes are used for English, French, German, Japanese, and Chinese (all currently supported languages).
+export const metricPrefixes: UnitPrefix[] = [
```
rajsite, line 12: remove the manuual type and make it `as const`

packages/nimble-components/src/unit/byte/models/byte-1024-unit-scale-formatter.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    protected override getSupportedPrefixes(): UnitPrefix[] {
+        return [
+            new UnitPrefix(1024, 'Ki'),
+            new UnitPrefix(1024 ** 2, 'Mi'),
+            new UnitPrefix(1024 ** 3, 'Gi'),
+            new UnitPrefix(1024 ** 4, 'Ti'),
+            new UnitPrefix(1024 ** 5, 'Pi')
+        ];
```
rajsite, line 29: `as const`

rajsite (commented):
another round of feedback

rajsite:
General comment before the next round is to take a pass and improve static-ness and const-ness of all the different objects we are making that are lookups

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
m-akinc: All the other test cases above for the decimal formatter validate interaction with the `UnitScaleFormatter` via the `EmptyUnitScaleFormatter`. This covers most logic, except that the `EmptyUnitScaleFormatter` only has a single scaled unit, so we need additional tests to cover the interaction of unit selection with rounding. I think that's the only thing that was missing.

Good catch though, because we weren't handling that case properly! I've added logic to `UnitScaleFormatter` to handle rounding, and associated test cases to `unit-sale-formatter.spec.ts`.

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
```diff
@@ -125,33 +119,47 @@ describe('DecimalFormatter', () => {
...
-    const disabled: string[] = [];
     for (const locale of locales) {
-        for (const testCase of testCases) {
-            const specType = getSpecTypeByNamedList(
-                testCase,
-                focused,
-                disabled
-            );
+        parameterizeNamedList(testCases, (spec, name, value) => {
             // eslint-disable-next-line @typescript-eslint/no-loop-func
```
rajsite, line 124: remove

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
rajsite: I'm skeptical about the algorithm put in.. I'm concerned that:

1. We are picking out one specific setting from formatterOptions to handle but our configuration allows arbitrary numeric formatter settings through. Can any of those other settings impact the rounding behavior? Like significant digits on exponentials might cause rounding right 1255 with two significant digits is 1.3e3.
2. Does this algorithm actually work for arbitrary scales? Like at the KiB to MiB boundary like 1,048.5 KiB with 0 decimals should round up to 1MiB due to the 1,048,576 byte boundary

rajsite:
FYI there is an open thread: https://github.com/ni/nimble/pull/1620#discussion_r1382293007

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
m-akinc: 1. You're right. 999 bytes with `maximumSignificantDigits=2` and `notation='scientific'` would be formatted as `1E3 bytes` rather than `1E0 kB` as it should. Moreover, there are other formatter options (including `roundingMode`, `roundingPriority`, and `roundingIncrement`) that affect the rounding behavior. The only approach I can think of to address this is to call `formatter.format()` for each unit, parse the number out of the resulting string, then test if it is >=1. I'm afraid that formatting and parsing O(n) times for every cell (even if n is a dozen or less) may be too inefficient. I should probably profile this. Another approach could be to put guardrails on our implementation by throwing errors if certain formatting options are passed into the `UnitScaleFormatter` constructor. What are your thoughts?
2. Yes, it works. Rounding behavior is unaffected by the conversion factor between units. So for example, if we start with 1,048,500 bytes, we test whether that can be represented as MiB by doing `1,048,500 / 1024^2 = 0.9999275`, then we see that rounds up to 1, so we pick MiB as the unit. The tests I have for this in `unit-scale-formatter.spec.ts` actually use a test formatter with non-uniform scaling factors 0.5, 1, 5, and 10.

rajsite: >  The only approach I can think of to address this is to call formatter.format() for each unit, parse the number out of the resulting string, then test if it is >=1. I'm afraid that formatting and parsing O(n) times for every cell (even if n is a dozen or less) may be too inefficient. I should probably profile this. 

I don't think I follow what algorithm is being proposed that results in `formatting and parsing O(n) times for every cell`. My first thought is it would be 1 or 2 times? (if it matches the current scale or the next up scale) Or maybe it's a worst case O(n) but average is 1 or 2?

> Another approach could be to put guardrails on our implementation by throwing errors if certain formatting options are passed into the UnitScaleFormatter constructor.

Sounds like advocating for a separate column at that point where only supported options are exposed. I'd probably not lean back in that direction.

> The tests I have for this in unit-scale-formatter.spec.ts actually use a test formatter with non-uniform scaling factors 0.5, 1, 5, and 10.

I'd appreciate some additional tests that more directly show real world examples. The invented scaling is probably useful but really hard to reason about.

packages/nimble-components/src/table-column/number-text/models/unit-scale-formatter.ts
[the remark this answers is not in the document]
m-akinc: In order to reliably handle the rounding effects of all possible `Intl.NumberFormat` options (as opposed to just `maximumFractionDigits`), I had to make significant changes involving additional calls to `Intl.NumberFormat.format()`. My initial, straightforward approach increased the render time from ~25ms to ~35ms. I felt that was too much, so I optimized the algorithm and got render times back down to ~27ms:
image

packages/nimble-components/src/table-column/number-text/models/tests/decimal-formatter.spec.ts
m-akinc: See my [response to an earlier thread about performance](https://github.com/ni/nimble/pull/1620#discussion_r1388383086). There are contrived scenarios where rounding do more than just picking the next unit up. I've implemented an algorithm (in `UnitScaleFormatter.format()`) that generally will result in two calls to `Intl.NumberFormat.formatToParts()`. In the worst case, it may call it N times, where N is the number of possible scaled units, though that would require configuration that we will almost surely never use.

> I'd appreciate some additional tests that more directly show real world examples (behaviors of real world scales).

I changed the `UnitScaleFormatter` tests so that most tests use byte scaled units, and only a few cases that can't be exercised by a byte scale are using my invented scaling.

jattasNI:
@m-akinc I moved this back to draft given our plan to do another round of refactoring.

m-akinc:
As we discussed, I am removed logic that was necessary to consider rounding effects when picking a unit. I also moved the responsibility of formatting out of the `ScaledUnit`s and back into the `Decimal/DefaultFormatter`. This hopefully makes things clearer. The logic in `Decimal/DefaultFormatter` did get more complicated as a result, though. They now have to manage caching `Intl.NumberFormat` instances for different scaled units. Furthermore, they have to deal with the fact that some units require passing additional options into `Intl.NumberFormat`, whereas others require appending the unit label to the formatted number string. This ugliness can't be avoided unless we let the `ScaledUnit`s do the formatting.

I reviewed performance again, using the "large table" story. I discovered that manually-translated units were slowing the render down because I was creating new instances of `Intl.PluralRules` for every cell. I am now caching a single instance in the formatter, which solves the performance issue.

I considered moving some files into the `utilities` directory as was suggested, but I'm not sure it's a better place for them.

packages/nimble-components/src/utilities/wait-until-custom-elements-defined-async.ts
```diff
@@ -0,0 +1,8 @@
+export async function waitUntilCustomElementsDefinedAsync(
```
rajsite, line 1: Note: need to update rich-text base class to use this after merging

packages/nimble-components/src/utilities/number-formatter/unit-scale/models/tests/manually-translated-unit-scale.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    const unitScale = new TestManuallyTranslatedUnitScale();
+
+    it('automatically includes base unit', () => {
+        expect(unitScale.baseScaledUnit).toBeDefined();
+    });
+
+    it('creates expected units from prefixes', () => {
+        const units = unitScale.supportedScaledUnits.sort((a, b) => (a.scaleFactor < b.scaleFactor ? -1 : 1));
```
mollykreis, line 26: `unitScale.supportedScaledUnits` is already sorted, right? Is there a reason you need to call `sort` here?

packages/nimble-components/src/utilities/wait-until-custom-elements-defined-async.ts
m-akinc: Also in `RichTextMentionListBox`, once that goes in.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -126,6 +134,23 @@ To improve the ability for users to visually scan values, applications should se
...
+<details>
+    <summary>Unit Elements</summary>
+
+    <ul>
+        <li>\`nimble-unit-byte\`: Labels in this unit scale are \`byte\`/\`bytes\`, \`KB\`, \`MB\`, \`GB\`, \`TB\`, and \`PB\`. Translations exist for all languages supported by the runtime environment.
+            <ul>
+                <li>\`binary\` - boolean attribute that indicates a binary conversion factor of 1024 should be used rather than 1000. The resulting unit labels are \`byte\`/\`bytes\`, \`KiB\`, \`MiB\`, \`GiB\`, \`TiB\`, and \`PiB\`. Translations exist for English, French, German, Japanese, and Chinese.</li>
+            </ul>
+        </li>
+        <li>\`nimble-unit-volt\`: Labels in this unit scale are \`volt\`/\`volts\`, \`fV\`, \`pV\`, \`nV\`, \`μV\`, \`mV\`, \`cV\`, \`dV\`, \`kV\`, \`MV\`, \`GV\`, \`TV\`, \`PV\`, and \`EV\`. Translations exist for English, French, German, Japanese, and Chinese.
```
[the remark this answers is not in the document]
m-akinc: Okay (though it's after `dV`, as that is decivolts).

packages/nimble-components/src/utilities/number-formatter/unit-scale/models/tests/manually-translated-unit-scale.spec.ts
m-akinc: Being sorted is an implementation detail. Only `UnitScale` itself knows that/relies on that. So if my test needs it sorted, it has to sort it.

mollykreis (approved):

change/@ni-nimble-components-e9fe99af-bb91-40c3-80e3-f5b41e2b0ff6.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: FYI @jattasNI @m-akinc working on some tweaks in this branch: https://github.com/ni/nimble/tree/numeric-column-units-structure-tweaks

I'll handle merge from main, the open comments in the PR, and addressing some TODOs I added before merging that branch into this PR for a final lookover

jattasNI: I haven't looked at this PR since the December refactors. I'll take another look once Milan finishes those changes.

packages/nimble-components/src/utilities/wait-until-custom-elements-defined-async.ts
rajsite: updated the rich-text base class and doesn't look like the RichTextMentionListBox used that pattern.

change/@ni-nimble-components-e9fe99af-bb91-40c3-80e3-f5b41e2b0ff6.json
rajsite: @m-akinc @jattasNI ready for review. I primarily reviewed the utilties/unit-format code, I reset myself to look at the element / table integration.

Major changes:
- Moved the related high-level overview docs to utilities/uunit-format/README.md. Would recommend reading iit to get a lay of the land for reviewing.
- Renamed the top-level classes to *UnitFormat, i.e. DefaultUnitFormat and DecimalUnitFormat. This was to more closely align with the Intl.NumberFormat/DateFormat/ListFormat/DataTimeFormat convention
- Updated the API surface to align with the Intl.NumberFormat API (use the same parameter names and defaults)
- Re-organized the layout to emphasize and split UnitScales and their associated objects from ManuallyTranslatedUnitScales and made naming align
- Generally switched to concrete classes and subclassing over interfaces where possible. Helped understand the different objects and responsibilities better with consistent naming.

packages/nimble-components/src/utilities/unit-format/README.md
```diff
@@ -0,0 +1,56 @@
+# UnitFormat
+
+## Overvew
+
+The `UnitFormat` library provides a way to format numbers with units. The library:
+
+-   Provides opionated number formatters that are well-suited for:
```
m-akinc, line 7:
```suggestion
-   Provides opinionated number formatters that are well-suited for:
```

packages/nimble-components/src/utilities/unit-format/README.md
```diff
@@ -0,0 +1,56 @@
...
+});
+console.log(formatter.format(1000));
+// Output: '1 kB'
+```
+
+## Concepts
+
+### UnitFormat
+
+There are multiple `UnitFormat`s that represent different configurations for formatting numbers. These represent wrappers around `Intl.NumberFormat` with opionated configurations. For example, the `DefaultUnitFormat` will represent "very large" and "very small" numbers with an exponential notation and numbers between those ranges as decimal. The different `UnitFormat`s take an optional `unitScale` that provides units and automatic conversions of numbers for that unit scale.
```
m-akinc, line 30:
```suggestion
There are multiple `UnitFormat`s that represent different configurations for formatting numbers. These represent wrappers around `Intl.NumberFormat` with opinionated configurations. For example, the `DefaultUnitFormat` will represent "very large" and "very small" numbers with an exponential notation and numbers between those ranges as decimal. The different `UnitFormat`s take an optional `unitScale` that provides units and automatic conversions of numbers for that unit scale.
```

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
```diff
@@ -196,48 +196,9 @@ If a value with a unit would be formatted with exponential notation, it will alw
...
-class UnitVolt extends FoundationElement implements UnitScale {
-    public getSupportedUnits(lang: string, formatterOptions: Intl.NumberFormatOptions): ScaledUnit[] {
-        // returns implementations of ScaledUnit that contain a shared Intl.NumberFormat (for formatting the number) and a specific translated unit string to append
-    }
-}
-```
-
-We will use `nimble-unit-byte` to display file sizes in SLE tables. Currently, SLE displays these values with the common `KB`/`MB`/`GB` unit labels, but uses a factor of 1024 to convert between units (which is [not uncommon](https://en.wikipedia.org/wiki/JEDEC_memory_standards#Unit_prefixes_for_semiconductor_storage_capacity), but [technically incorrect](https://physics.nist.gov/cuu/Units/binary.html)). Whether SLE chooses to standardize on the default 1000-based byte units or the 1024-based ones, it will be a change from the current behavior. To ensure consistency, we will update SLE's file size pipe and search for other places where byte values are being converted so they can be updated accordingly.
-
-##### Creating new units
```
m-akinc: Do we not want this section? I'm indifferent--I think it's simple enough to follow the examples of existing units--but I think @jattasNI had asked for this to be documented.

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/tests/volt-unit-scale.spec.ts
```diff
@@ -102,20 +102,36 @@ describe('VoltUnitScale', () => {
...
-                expect(
-                    unit.unitFormatterFactory(locale, {}).format(scaledValue)
-                ).toEqual(value.formatted[index]!);
-            }
+    parameterizeNamedList(testCases, (spec, name, value) => {
+        spec(`gets expected unit for ${name}`, () => {
+            const { scaledValue, scaledUnit } = voltUnitScale.scaleNumber(
+                value.number
+            );
+            expect(
```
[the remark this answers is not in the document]
rajsite: Couple of things:

The array indices are hard-coded but type checked. With the loop TypeScript wasn't able to understand the type and a non-null assertion was needed. Unrolled TypeScript validates the index is valid for each entry, no compiler ovverrides needed.

The bigger reason is we should avoid expects in loops generally. A concrete example I ran into and changed the pattern was the following. For this snippet:
```
for (const [index, locale] of [
                'en',
                'fr',
                'de',
                'ja',
                'zh'
            ].entries()) {
                expect( // line 115
                    unit.unitFormatterFactory(locale, {}).format(scaledValue)
                ).toEqual(value.formatted[index]!);
            }
```

I had introduced a problem that failed in one of the languages. Looking just at the CI results I couldn't tell which language. Even though the test was parameterized, this wasn't in the parameterization so in the test name I couldn't tell which case failed. The stack trace doesn't help because it just says it failed on line 115 but that corresponds to 5 different expects. In this case it was pretty easy to tell what string didn't match but for numbers or booleans it's not easy.

So in general avoid expects in loops. Some alternatives:
1. Break it out into the parametrization so it is part of the test name
2. Break it into clear pre and post conditions steps. Calculate an array of values before and compare it to an array of expects. Use the jasmine matchers to help write that test:
    ```ts
    const expected = [true, true, true];
    const actual = obj.dynamicValues();
    expect(actual).toEqual(jasmine.arrayWithExactContents(expected));
    ```
3. Unroll the loop and make each expect explicit

I chose 3 for this case, 2 would have been fine too.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -0,0 +1,76 @@
...
+            unitScale = passthroughUnitScale
+        }: {
+            minimumFractionDigits?: number,
+            maximumFractionDigits?: number,
+            unitScale?: UnitScale
+        } = {
+            minimumFractionDigits: 0,
+            maximumFractionDigits: 3,
+            unitScale: passthroughUnitScale
+        }
```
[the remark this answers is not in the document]
rajsite: Yep, it's the destructuring assignment language feature. Reduces a lot of boilerplate for configuration objects as you described. MDN is a bit terse but this gives some discussion https://javascript.info/destructuring-assignment#smart-function-parameters
The named types instead of the inline ones also make it a little clearer

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -0,0 +1,76 @@
+import { UnitFormat } from './base/unit-format';
+import type { ScaledUnitFormat } from './unit-scale/base/scaled-unit-format';
+import type { UnitScale } from './unit-scale/base/unit-scale';
+import { passthroughUnitScale } from './unit-scale/passthrough-unit-scale';
+
+/**
+ * Format for decimal numbers with units.
+ */
+export class DecimalUnitFormat extends UnitFormat {
+    /** Resolved UnitScale */
```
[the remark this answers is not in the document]
rajsite: (this comment resolves multiple discussions)
Goal from the change was to align with `Intl.NumberFormat` api as stated in the README so that leads to two changes:
- constructor takes two params, locale and options object, so that was done for both UnitFormats even if today it only has one option: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#parameters (added some types to emphasize that)
- the constructor params are resolved to different values and you can observe those changes in the public api (added resolvedOptions to match `Intl.NumberFormat` api instead of individual properties)

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: Didn't find the unit element discussion very necessary, They are pretty trivial.

I tweaked the unit-format README to be a bit clearer about the different unit-scale types. I think it captures everything that was here but I've also looked at it too long so up for adding additional docs if @jattasNI can think of some that are useful.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -126,6 +134,23 @@ To improve the ability for users to visually scan values, applications should se
 </details>
 `;
 
+const unitDescription = `A unit for the column may be configured by providing a \`nimble-unit-<name>\` element as content (in addition to the column label). Unit elements represent a set of related, scaled units, e.g. \`nimble-unit-byte\` represents bytes, KB, MB, etc. Values are converted from a source unit (e.g. bytes) to the largest scaled unit (e.g. KB, MB, etc.) that can represent that value with magnitude >= 1. The source data for the column is expected to be given in the base unit specified in the tag name, e.g. for \`nimble-unit-byte\`, a source value should be a number of bytes.
+
+<details>
```
jattasNI, line 139: I think I might advocate for us to move the unit docs to a separate section in an mdx file for this column and/or a separate story. As we add units it'll eventually be a lot of info to cram into a details section of a parameter description.

But that doesn't need to happen in this PR, it can happen as we add more units. I'm just throwing it out now for feedback / awareness.

jattasNI (commented):
Finished reviewing elements and docs. Looking nice and clean! I'll review the unit library next week.

rajsite:
Reset @jattasNI since they stated in the latest comment the library is still in review

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text-matrix.stories.ts
```diff
@@ -82,3 +100,87 @@ tableColumnNumberTextThemeMatrix.play = async (): Promise<void> => {
...
+    for (let i = 0; i < rowsLargeData; i++) {
+        const row: KeyedObject = {};
+        for (let j = 0; j < columnsLargeData; j++) {
+            row[`col${j}`] = i * j + j + set * 100;
+        }
+        largeDataSets[set]!.push(row);
+    }
+}
+
+export const largeTable: StoryObj<NumberTextColumnTableArgs> = {
```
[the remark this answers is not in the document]
rajsite: I had requested the large table so we know that it did not regress the performance of number columns without units compared to those with, i.e. that unit support was not adding a general tax to all number rendering since every number does go through the UnitFormat system.

It accomplishes that but after playing with it I'd like to see more configuration for the different UnitFormats, for it to use numeric values that cause different scale factors to actually be rendered (currently it is never rendering anything other than the base scale factor for the units selected), for more column types than just numeric, and I'd like to see a stress test of the limits of update performance for profiling (better understand where the costs of table updates actually are).

I'm going to split this out into a separate PR to play with organizing it a bit for those goals.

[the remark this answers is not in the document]
rajsite: lol [relevant question](https://teams.microsoft.com/l/message/19:0c59ac4a4cc1438894cb42f4b21703f9@thread.skype/1705423311927?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=41626d4a-3f1f-49e2-abdc-f590be4a329d&parentMessageId=1705423311927&teamName=ASW%20SystemLink&channelName=UI%20Working%20Group&createdTime=1705423311927) in teams. Yea while we did perf tests in prototyping, we haven't really validated it now that we are further along with more complex columns and with fairly complex rendering implemented with column sizing, etc. Would be good to have confidence working with clients

packages/nimble-components/src/utilities/unit-format/unit-scale/tests/byte-unit-scale.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { parameterizeNamedList } from '../../../tests/parameterized';
+import { byteUnitScale } from '../byte-unit-scale';
+
+describe('ByteUnitScale', () => {
```
jattasNI, line 4: I think we should have tests in at least one more language here. We don't need all languages since we're relying on NumberFormatter for translations but we should test one non-English language to ensure we don't hardcode an `en` somewhere in the implementation.

packages/nimble-components/src/utilities/unit-format/unit-scale/base/unit-scale.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.supportedScaledUnits = supportedScaledUnits;
+        this.baseScaledUnit = baseScaledUnit;
+    }
+
+    // Note that for the sake of reducing complexity in the implementation,
+    // we do NOT consider the effects of rounding when picking the unit to
+    // use for a given value. If formatting results in rounding, a value
+    // may be shown with an unexpected unit. Examples:
+    // - 999 bytes with two significant digits => "1000 bytes" (instead of "1 kB")
+    // - 0.00000000000000001 volts (= 0.01 fV) with one fractional digit => "0 fV" (instead of "0 volts")
```
jattasNI, line 40: Could you make sure we have test cases that verify the examples provided here (or something equivalent)?

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/base/manually-translated-scaled-unit-format.ts
```diff
@@ -0,0 +1,74 @@
...
+            ? this.unitTranslation.singular
+            : this.unitTranslation.plural;
+        return `${formatted} ${unitLabel}`;
+    }
+
+    private getTranslationToUse(locale: string): UnitTranslation {
+        const localeObject = new Intl.Locale(locale ?? '');
+        const language = localeObject.language;
+        const region = localeObject.region;
+        const languageAndRegion = `${language}-${region?.toUpperCase() ?? ''}`;
```
jattasNI, line 67: Can you use `localeObject.toLocaleString()` here instead? It'd be good to avoid any custom logic about how locale strings are constructed.

For example right now I think this handles the region being upper or lower case but the language is assumed to be lower case. In my testing `Intl.Locale` handles that by normalizing `language` to lower, `region` to upper, and `toLocaleString()` to `lower-UPPER`.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
rajsite: Agreed. I think most of the options in the table have too many details inline that are worth having but should be moved to discussion in the MDX. I don't think we need to track that explicitly but likely would predicate adding additional units after this PR on adding an mdx and cleaning up units at least.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -185,6 +216,11 @@ export const numberTextColumn: StoryObj<NumberTextColumnTableArgs> = {
             options: [undefined, 0, 1, 2, 3, 20],
             control: { type: 'select' }
         },
+        unit: {
+            description: unitDescription,
+            options: ['default', 'byte', 'byte (1024)', 'volt'],
```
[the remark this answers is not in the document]
rajsite: Went with volt ⚡ (partially because the current setup makes for not great byte values, i.e. bytes usually aren't continuous and -6.45326E-7 bytes which the table will show right now doesn't make sense, but if interested we can clean-up in a follow-up)

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/base/manually-translated-scaled-unit-format.ts
m-akinc: Using `toLocaleString()` does seem better, but it looks like what I had would have worked reliably anyway (since `Intl.Locale` already normalizes the values of `language` and `region`). The `toUpperCase()` call was apparently redundant.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
rajsite: @jattasNI did you want to make a comment on this specifically?

change/@ni-nimble-components-e9fe99af-bb91-40c3-80e3-f5b41e2b0ff6.json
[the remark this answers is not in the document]
m-akinc: The unit tests already run and pass in Chrome and Firefox by default. I also ran them locally against WebKit. I believe that should give us the confidence that the NumberFormat API is giving the same results in all browsers we care about. Let me know if you intended something else (i.e. "inspect the results"?).

packages/nimble-components/src/utilities/unit-format/unit-scale/base/unit-scale.ts
m-akinc: Since this is a behavior that we are settling for, rather than one that we explicitly want, I didn't think it made sense to assert it via tests. But I defer to your judgment.

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/base/manually-translated-scaled-unit-format.ts
m-akinc: I had to revert back to the original logic (minus the unnecessary `toUpperCase()` call), because `localeObject.toLocaleString()` can include subtags from the input locale string that we don't care about (e.g. in "fr-Latn-CA-u-hc-h12", we don't care about anything other than "fr" and "CA"). In other words, if we have translations for "fr-CA", we want to use them whether the locale is "fr-CA" or "fr-Latn-CA" or "fr-CA-u-hc" or so on.

packages/nimble-components/src/utilities/unit-format/unit-scale/base/unit-scale.ts
rajsite: Makes sense to test our edge cases so the behavior is known and it is clear if it changes rather than untested and unknown

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/base/manually-translated-scaled-unit-format.ts
```diff
@@ -0,0 +1,74 @@
...
+            ? this.unitTranslation.singular
+            : this.unitTranslation.plural;
+        return `${formatted} ${unitLabel}`;
+    }
+
+    private getTranslationToUse(locale: string): UnitTranslation {
+        const localeObject = new Intl.Locale(locale ?? 'en');
+        const language = localeObject.language;
+        const region = localeObject.region;
+        const languageAndRegion = `${language}-${region ?? ''}`; // do not include any other subtags that might be in the given locale string
```
rajsite, line 67: Won't this result in the string `en-` if region is undefined? Is that right?

rajsite (approved):

packages/nimble-components/src/utilities/unit-format/manually-translated-unit-scale/base/manually-translated-scaled-unit-format.ts
m-akinc: Yes, so `en-` will fail the lookup, but next we try the language by itself, which will succeed. For clarity, I've tweaked it so we never attempt to lookup a value like `en-`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-formatted-text.md
m-akinc: If you're okay with the deletion, resolve this thread, so the auto-merge can complete.

rajsite: @jattasNI I'll go ahead and resolve for now to move the PR, can add additional docs on a follow up if you have additional feedback

PR #1623, opened 2023-10-25
Prevent navigation due to routerLink directive when host is disabled

### LINKED ISSUE #1612: [Angular] Disabled elements using nimbleRouterLink navigate on click
# 🐛 Bug Report

In Angular, if an anchor element (e.g. nimble-anchor-button) is disabled, clicking on it will still perform navigation if the route is specified using `nimbleRouterLink`. This is because the router link directive uses a `click` `@HostBinding` that doesn't perform a check for whether or not the element is disabled.

## 💻 Repro or Code Sample

In an Angular app, create a disabled `nimble-anchor-button` that has its route specified using a `nimbleRouterLink` binding. Notice that clicking on it navigates to the route specified by the router link even though the button is disabled.

## 🤔 Expected Behavior

If an anchor element (e.g. nimble-anchor-button) is disabled, clicking on it should do nothing.

## 😯 Current Behavior

If an anchor element (e.g. nimble-anchor-button) is disabled, clicking on it navigates as though it were not disabled.

## 🌍 Your Environment

Not environment specific

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1612

## 👩‍💻 Implementation

Overriding `RouterLinkWithHref`'s `onClick` handler and short-circuiting if the host element is `disabled`.

Affects anchor button, anchor menu item, anchor tree item, and anchor tab.

## 🧪 Testing

Added one new test for each of the four affected components.

## ✅ Checklist


m-akinc:
@msmithNI could you please buddy this PR?

msmithNI (approved):

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-button/nimble-anchor-button-router-link-with-href.directive.ts
```diff
@@ -20,4 +22,16 @@ export class NimbleAnchorButtonRouterLinkWithHrefDirective extends RouterLinkWit
     public set nimbleRouterLink(commands: never[] | string | null | undefined) {
         this.routerLink = commands;
     }
+
+    public constructor(injector: Injector, private readonly elementRef: ElementRef<AnchorButton>) {
```
rajsite, line 26: We have a directive base for anchors generally: https://github.com/ni/nimble/blob/main/angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/nimble-anchor-base.directive.ts

Seems like we could have a `DisableableRouterLinkWithHref extends RouterLinkWithHref`  base all these could extend from?

You'll probably have to make the ElementRef type generic, something like `T extends {disabled: boolean}` and can then do something like `NimbleAnchorButtonRouterLinkWithHrefDirective extends DisableableRouterLinkWithHref<AnchorButton>`, etc.

atmgrifter00 (approved):

PR #1630, opened 2023-10-27
Add incubating `nimble-card` component

kjohn1922:
# Pull Request

Initial component code for the `nimble-card` component.

## 🤨 Rationale

This adds the initial `nimble-card` component, storybook, and tests.

I will add a "title" slot to the template in a future PR. Future PRs will also add Angular and Blazor support.

The styling in the component right now is intended as a placeholder. We will get styling from our designer and update the styles before releasing the Routines UI with the card component.

## 👩‍💻 Implementation

- Added myself as a code owner for the incubating component, per the contributing instructions
- Marked the component as incubating in the component-status table
- Added index, styles, and template files for the new component, called `nimble-card`

## 🧪 Testing

- Added a storybook page under Incubating with an example
- Added a spec file with a basic unit test

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-a3500224-7314-4022-aa82-2f4cf29f7866.json
+++ b/change/@ni-nimble-components-a3500224-7314-4022-aa82-2f4cf29f7866.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add an incubating `nimble-card` component",
+  "packageName": "@ni/nimble-components",
+  "email": "35350751+kjohn1922@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/card/index.ts
+++ b/packages/nimble-components/src/card/index.ts
@@ -1,4 +1,7 @@
-import { DesignSystem, Card as FastCard } from '@microsoft/fast-foundation';
+import {
+    DesignSystem,
+    Card as FoundationCard
+} from '@microsoft/fast-foundation';
 import { styles } from './styles';
 import { template } from './template';
 
@@ -11,11 +14,11 @@ declare global {
 /**
  * A nimble-styled card
  */
-export class Card extends FastCard {}
+export class Card extends FoundationCard {}
 
 const nimbleCard = Card.compose({
     baseName: 'card',
-    baseClass: FastCard,
+    baseClass: FoundationCard,
     template,
     styles
 });
--- a/packages/nimble-components/src/card/index.ts
+++ b/packages/nimble-components/src/card/index.ts
@@ -1,0 +1,24 @@
+import { DesignSystem, Card as FastCard } from '@microsoft/fast-foundation';
+import { styles } from './styles';
+import { template } from './template';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-card': Card;
+    }
+}
+
+/**
+ * A nimble-styled card
+ */
+export class Card extends FastCard {}
+
+const nimbleCard = Card.compose({
+    baseName: 'card',
+    baseClass: FastCard,
+    template,
+    styles
+});
+
+DesignSystem.getOrCreate().withPrefix('nimble').register(nimbleCard());
+export const cardTag = DesignSystem.tagFor(Card);
--- a/packages/nimble-components/src/card/styles.ts
+++ b/packages/nimble-components/src/card/styles.ts
@@ -5,8 +5,8 @@ import {
     bodyFontColor,
     borderColor,
     borderWidth,
-    mediumPadding,
-    sectionBackgroundColor
+    sectionBackgroundColor,
+    standardPadding
 } from '../theme-provider/design-tokens';
 
 export const styles = css`
@@ -17,6 +17,6 @@ export const styles = css`
         color: ${bodyFontColor};
         background-color: ${sectionBackgroundColor};
         border: ${borderWidth} solid ${borderColor};
-        padding: ${mediumPadding};
+        padding: ${standardPadding};
     }
 `;
--- a/packages/nimble-components/src/card/styles.ts
+++ b/packages/nimble-components/src/card/styles.ts
@@ -10,13 +10,15 @@ import {
 } from '../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('block')}
+    ${display('flex')}
 
     :host {
+        flex-direction: column;
+        gap: ${standardPadding};
+        padding: ${standardPadding};
+        border: ${borderWidth} solid ${borderColor};
         font: ${bodyFont};
         color: ${bodyFontColor};
         background-color: ${sectionBackgroundColor};
-        border: ${borderWidth} solid ${borderColor};
-        padding: ${standardPadding};
     }
 `;
--- a/packages/nimble-components/src/card/styles.ts
+++ b/packages/nimble-components/src/card/styles.ts
@@ -1,48 +1,23 @@
 import { css } from '@microsoft/fast-element';
 import { display } from '@microsoft/fast-foundation';
-import { Theme } from '../theme-provider/types';
-import { themeBehavior } from '../utilities/style/theme';
 import {
     bodyFont,
     bodyFontColor,
-    buttonFillPrimaryColor
+    borderColor,
+    borderWidth,
+    mediumPadding,
+    sectionBackgroundColor
 } from '../theme-provider/design-tokens';
 
+// TODO AB#2542845 Replace placeholder style
 export const styles = css`
     ${display('block')}
 
     :host {
         font: ${bodyFont};
         color: ${bodyFontColor};
-        margin: 10px;
-        padding: 10px;
+        background-color: ${sectionBackgroundColor};
+        border: ${borderWidth} solid ${borderColor};
+        padding: ${mediumPadding};
     }
-`.withBehaviors(
-    themeBehavior(
-        Theme.light,
-        css`
-            :host {
-                background-color: #f0f0f0;
-                border: 2px solid #cccccc;
-            }
-        `
-    ),
-    themeBehavior(
-        Theme.dark,
-        css`
-            :host {
-                background-color: #333333;
-                border: 2px solid #666666;
-            }
-        `
-    ),
-    themeBehavior(
-        Theme.color,
-        css`
-            :host {
-                border: 2px solid ${buttonFillPrimaryColor};
-                background-color: ${buttonFillPrimaryColor};
-            }
-        `
-    )
-);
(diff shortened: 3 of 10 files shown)


packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { Meta, StoryObj } from '@storybook/html';
+import { loremIpsum } from '../../utilities/tests/lorem-ipsum';
+import {
+    createUserSelectedThemeStory,
+    incubatingWarning
+} from '../../utilities/tests/storybook';
+import { buttonTag } from '../../button';
+import { cardTag } from '..';
+
+const overviewText = `The \`nimble-card\` is a container that is designed to contain arbitrary content that is specified by a client
```
kjohn1922, line 11: I modeled this overview text off of the nimble-card-button's overview. Do we typically have a tech writer review these strings?

packages/nimble-components/src/card/styles.ts
```diff
@@ -0,0 +1,23 @@
...
+import {
+    bodyFont,
+    bodyFontColor,
+    borderColor,
+    borderWidth,
+    mediumPadding,
+    sectionBackgroundColor
+} from '../theme-provider/design-tokens';
+
+// TODO AB#2542845 Replace placeholder style
```
kjohn1922, line 12: Let me know if there's a preferred way to mark placeholders or todo items. This is what I would do in the Skyline repo, but I don't know if it's appropriate here.

packages/nimble-components/src/card/index.ts
```diff
@@ -0,0 +1,24 @@
+import { DesignSystem, Card as FastCard } from '@microsoft/fast-foundation';
```
mollykreis, line 1: We generally rename the fast component as `FoundationXYZ`, so in this case, you'd do `Card as FoundationCard`.

packages/nimble-components/src/card/tests/card.stories.ts
mollykreis: No, we've never gotten a tech writer to review these descriptions. A lot of time we'll pull text from the W3C description of what a specific component is, but I don't think that applies here. I think what you have here is fine, and I'm sure other reviewers will feel free to weigh in if they have changes they'd like to see.

packages/nimble-components/src/card/styles.ts
mollykreis: We don't often link TODOs like this, but if we do, they should link to a nimble issue rather than an azdo work item. If it's short-lived, then you probably don't need a TODO, and we'll just know that it needs to be addressed before closing the issue for the card component. If it's going to stay around for a while, then we often create tech debt issues within nimble to track incomplete work or cleanup that needs to happen.

packages/nimble-components/src/card/template.ts
```diff
@@ -0,0 +1,8 @@
+import { html } from '@microsoft/fast-element';
+import type { Card } from '.';
+
+export const template = html<Card>`
+    <template>
+        <slot></slot>
+    </template>
```
mollykreis, line 7: Since you aren't modifying the `<template>` element at all by adding a role, event handlers, attributes, css classes, etc, you don't need it here.

```suggestion
    <slot></slot>
```

packages/nimble-components/src/tests/component-status.stories.ts
```diff
@@ -112,9 +112,9 @@ const components = [
         componentName: 'Card',
         issueHref: 'https://github.com/ni/nimble/issues/296',
         issueLabel: 'Issue',
-        componentStatus: ComponentFrameworkStatus.doesNotExist,
-        angularStatus: ComponentFrameworkStatus.doesNotExist,
-        blazorStatus: ComponentFrameworkStatus.doesNotExist
+        componentStatus: ComponentFrameworkStatus.incubating,
+        angularStatus: ComponentFrameworkStatus.incubating,
+        blazorStatus: ComponentFrameworkStatus.incubating
```
mollykreis, line 117: The Angular and Blazor status should still be `doesNotExist` until support is added in those frameworks for the card.

```suggestion
        angularStatus: ComponentFrameworkStatus.doesNotExist,
        blazorStatus: ComponentFrameworkStatus.doesNotExist
```

packages/nimble-components/src/tests/component-status.stories.ts
```diff
@@ -112,9 +112,9 @@ const components = [
         componentName: 'Card',
```
mollykreis, line 112: Now that you have a storybook page for the component, you can link it here as well.

```suggestion
        componentName: 'Card',
        componentHref: './?path=/docs/incubating-card--docs',
```

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+    render: createUserSelectedThemeStory(html`
+        ${incubatingWarning({
+        componentName: 'card',
+        statusLink: 'https://github.com/ni/nimble/issues/296'
+    })}
+        <${cardTag}>
+            <h2>Title</h2>
+            <p>${loremIpsum}</p>
+            <${buttonTag}>Button</${buttonTag}>
```
mollykreis, line 33: This might seem a bit nitpicky, but we like to have our examples be something that we'd recommend clients doing. So, some things that we'd want to change about this example are:
- Don't use an `<h2>` tag for the title. Instead, you'd want to use a `<span>` or `<div>` styled with nimble tokens. I think it would also be acceptable to remove the title from the example for now since any storybook-specific styling you do now will be negated by the `title` slot that will be coming soon.
- Don't wrap the text in a `<p>`. Instead, use a `<span>` or `<div>` since `<p>` tags have default top/bottom margins that are not tied to nimble tokens.
- Space the elements within the card out using nimble tokens.

I think this could look something like:
```html
<style>
    .title {
        margin-bottom: var(${standardPadding.cssCustomProperty});
        font: var(${titlePlus1Font.cssCustomProperty});
        color: var(${titlePlus1FontColor.cssCustomProperty});
    }

    .body {
        margin-bottom: var(${standardPadding.cssCustomProperty});
    }
</style>
<${cardTag}>
    <div class="title">Title</div>
    <div class="body">${loremIpsum}</div>
    <${buttonTag}>Button</${buttonTag}>
</${cardTag}>
```

packages/nimble-components/src/card/index.ts
```diff
@@ -0,0 +1,24 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-card': Card;
+    }
+}
+
+/**
+ * A nimble-styled card
+ */
+export class Card extends FastCard {}
```
mollykreis, line 14: You'll need to add an import of this component in `all-components.ts`.

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -0,0 +1,40 @@
+import { html } from '@microsoft/fast-element';
```
mollykreis, line 1: Can you also add "matrix" stories for the new component? You can look at some of the existing matrix tests in the repository as an example. The one you create (at least initially) won't need as many test dimensions as most of our existing components, which test for all the various enabled/disabled state, appearances, etc.

`card-button-matrix.stories.ts` might be a good reference since it is somewhat similar to what you'll be adding.

kjohn1922: Added matrix stories with the theme matrix and the hidden state
image
Theme:
image
Hidden:
image

packages/nimble-components/src/card/template.ts
kjohn1922: Done. Prettier moved it all onto one line: `export const template = html<Card>`\<slot\>\</slot\>`;`

packages/nimble-components/src/card/tests/card.stories.ts
kjohn1922: Not nitpicky at all, thanks for the information! I went ahead and removed the title from the example for now and made the other suggested changes.

packages/nimble-components/src/card/styles.ts
kjohn1922: I removed the TODO, since we do plan to update this within the next iteration

mollykreis (approved):

jattasNI (approved):

change/@ni-nimble-components-a3500224-7314-4022-aa82-2f4cf29f7866.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: For your implementation PRs a helpful thing to do is to mention the implementation issue #296 in the PR description. GitHub automatically adds links in both directions so it can make it easier to look a the issue and see all the dev work that's associated with it.

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -0,0 +1,55 @@
...
+    render: createUserSelectedThemeStory(html`
+    ${incubatingWarning({
+        componentName: 'card',
+        statusLink: 'https://github.com/ni/nimble/issues/296'
+    })}
+        <style>
+            .card-content {
+                display: flex;
+                flex-direction: column;
+                gap: var(${standardPadding.cssCustomProperty});
```
rajsite, line 36: This seems like reasonable default styling to place in the card styles itself. i.e. 

```
${display('flex')}

: host {
    flex-direction: column;
    gap: ${standardPadding};
}
```
Know why we wouldn't make that the default?

kjohn1922: Done!

kjohn1922: This gets a little more complicated when we add the `title` slot and add the `section` to the template for accessibility, but I think we can sort through that in the next PR that adds the title slot.

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+    tags: ['autodocs'],
+    parameters: {
+        docs: {
+            description: {
+                component: overviewText
+            }
+        },
+        actions: {}
+    },
+    render: createUserSelectedThemeStory(html`
```
rajsite, line 26: FYI the changes were small but wanted to see how the changes impacted the show code view. I like that the example has less custom styling / is simpler usage. I'm guessing we will want to show a bit more about sizing patterns but we'll save that for when it's closer to the finished implementation

image

rajsite (approved):

PR #1632, opened 2023-10-30
Add sizing to spec templates

jattasNI:
# Pull Request

## 🤨 Rationale

We've had several discussions lately about how existing and new components should be sized, including questions about default sizes, intrinsic sizes to fit all of a component's data, and respecting min/max sizes. Many of these conversations were triggered by bugs or feature requests. It would be good to think them through earlier in the component development process.

## 👩‍💻 Implementation

Added a checklist item about sizing (and more generally about response to all css properties) to the new component spec templates.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/specs/templates/custom-component.md
+++ b/specs/templates/custom-component.md
@@ -50,7 +50,7 @@
 - *Methods*
 - *Events*
 - *CSS Classes and CSS Custom Properties that affect the component*
-- *Built-in CSS Properties that affect the component. In particular, how is it sized?*
+- *How native CSS Properties (height, width, etc.) affect the component*
 
 *Consider high and low-level APIs. Attempt to design a powerful and extensible low-level API with a high-level API for developer/designer ergonomics and simplicity.*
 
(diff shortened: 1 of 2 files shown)


atmgrifter00 (approved):

mollykreis (approved):

specs/templates/custom-component.md
```diff
@@ -50,6 +50,7 @@
 - *Methods*
 - *Events*
 - *CSS Classes and CSS Custom Properties that affect the component*
+- *Built-in CSS Properties that affect the component. In particular, how is it sized?*
```
rajsite, line 53: Think mentioning a couple specific CSS Properties (like width and height) makes it clearer. Also chose "native" over "built-in" but other ideas could be `Browser-native CSS Properties` or `browser built-in CSS Properties`

```suggestion
- *Response to configuring native CSS Properties that affect the component (height, width, etc.)?*
```

rajsite (approved):

Issue #1533, opened 2023-09-14
Rich text editor | Need an update to the testing strategy followed in markdown-serializer 


vivinkrishna-ni:
## 🧹 Tech Debt

Relevant discussions:

https://github.com/ni/nimble/pull/1496#discussion_r1324888037
https://github.com/ni/nimble/pull/1526#discussion_r1325206227

In the rich text editor component, the custom-written markdown serializer's unit tests need an update in the testing strategy. Currently, it creates a mock TipTap editor with configurations not aligned with the actual editor and is unstable moving forward. We need to find a better approach so that the test written is aligned with the production code and also avoid writing manual HTML inputs as same as the Tiptap editor renders.

jattasNI:
@aagash-ni could you update the description of this issue or leave a comment describing what work is remaining to close the issue now that #1633 is complete?

vivinkrishna-ni:
@jattasNI This issue is resolved with this https://github.com/ni/nimble/pull/1633 PR completion. The Chromatic tests that I mentioned earlier are already taken into consideration for absolute links in viewer and editor matrix tests and also for other rich text formatting.

This issue is primarily to resolve [this comment](https://github.com/ni/nimble/pull/1496#discussion_r1324888037) on the markdown-serializer testing strategy and not includes the integration testing like copy-pasting from other rich text component. For integration testing, we have a separate issue created https://github.com/ni/nimble/issues/1568 and will be tracking from there. We would like to get your thoughts/acknowledgment to close this issue.

Issue #1568, opened 2023-09-26
Rich text | Integration test for nimble rich text editor and viewer components


vivinkrishna-ni:
## 🧹 Tech Debt

As the `nimble-rich-text-editor` and `nimble-rich-text-viewer` components are mostly used together for a feature like `comments` and also share and align the markdown formats between them, integration tests will capture more common cases. For example, the tests can include,
1. Getting a markdown output from the editor and loading it to the viewer
2. Copying a text or link from the viewer and pasting it in the editor. Related discussion: https://github.com/ni/nimble/pull/1559#discussion_r1337518612

Our initial thought is to instantiate both the components and utilize their page object methods to interact between them.

vivinkrishna-ni:
> As the second task, an update is needed in the implementation of the sliceEditorContent() method used in the rich-text-mention spec file. Currently, when testing the removal of characters from the editor, the content is manipulated and replaced. However, a more accurate approach would involve removing text by emitting actual keyboard events such as delete or backspace. Related Discussion: https://github.com/ni/nimble/pull/1683#discussion_r1417601642

This part of the issue is resolved by using the actual Tiptap utilities to delete a text instead of removing the elements from the DOM of the editor. We have used the [`deleteRange()`](https://tiptap.dev/api/commands/delete-range) method from Tiptap to delete a particular range from the editor by passing a position to it. We have updated the `sliceEditorContent()` of the editor's page object method in https://github.com/ni/nimble/pull/1694. Related comment in the same PR: https://github.com/ni/nimble/pull/1694#discussion_r1423991292

rajsite:
List of tests that were deemed too challenging to write as unit tests that may be writeable as integration tests:
- Mention editing interactions (in particular backspace interactions, see #1716 )

PR #1633, opened 2023-10-30
Rich text editor | Update testing strategy in markdown-serializer

aagash-ni:
# Pull Request

## 🤨 Rationale
Testing strategy in the markdown serializer need to be updated as it is generating a mock editor currently to test the behavior instead of implementing actual editor. So, it needs to be updated with actual editor to test the production configuration.

Issue link: #1533 

## 👩‍💻 Implementation

- Removed the mock editor and constructed the actual editor `nimble-rich-text-editor` to get the editor document.
- Removed all the html as input and updated it with `setMarkdown()` method as in production flow only there will be markdown interaction with the editor and not with the html.

## 🧪 Testing

- Manually ran the test and verified.

## ✅ Checklist


aagash-ni:
@m-akinc Kindly review this PR

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -1,293 +1,303 @@
...
+                inputMarkdown: '`code`',
+                outputMarkdown: '\\`code\\`'
             },
-            { name: 'Code', html: '<code>Code</code>', plainText: 'Code' },
             {
                 name: 'CodeBlock',
-                html: '<pre><code>CodeBlock</code></pre>',
-                plainText: 'CodeBlock'
+                inputMarkdown: '```\nCodeBlock\n```',
+                outputMarkdown: '\\`\\`\\` CodeBlock \\`\\`\\`'
```
m-akinc, line 264: For all these tests where the output has added characters, it would be good to have additional tests that use the `outputMarkdown` as inputs. We should verify that re-loading the resulting markdown doesn't keep adding characters every time.

m-akinc (approved):

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
aagash-ni: Added testcases to check the character addition markdown output when reloading.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -1,293 +1,354 @@
...
+        ({ element } = await setup());
+    });
 
     afterAll(() => {
-        mockEditor.destroy();
+        element.tiptapEditor.destroy();
     });
 
-    describe('various supported nodes should be serialized to a markdown output', () => {
+    describe('Should serialize editor document to a markdown output', () => {
```
jattasNI, line 24: Can you help me understand the value of these tests now? What codepath are they executing and what do they cover that unit tests don't?

It seems like a few of them are testing that certain characters are escaped or stripped from the output while the bulk of them are testing that the input and output markdown are identical.

My initial thought is that the tests that have different output for a given input are likely valuable cases. If they are valuable here, they should be split off into a separate block from the other set of tests and ideally given test names that describe what they are covering. e.g. "bold is removed from links". But I'm not clear on whether they need to be integration tests like this or whether we have similar coverage elsewhere. Can you comment on that?

The tests with identical input and output might have some value if we think there's a chance they'd catch issues as we change our rule configuration. If that's the case then what set of tests do we need to gain that confidence? Should we be testing every combination of every markdown syntaxes? Unsupported syntaxes? Weird and malicious strings? Or are those covered elsewhere?

rajsite: Yea the [threads linked from the issue](https://github.com/ni/nimble/pull/1496#discussion_r1326133305) state that editor interactions would be valuable:

> So the test we want is that interactions with the tiptap editor produce expected markdown in the end. The HTML intermediate representation is a detail we don't really care about much as long as the same interactions in the editor produce the same markdown.

The setMarkdown + getMarkdown tests seem different. Are they trying to check that strings can be round tripped without change? Is that a requirement?

rajsite: @aagash-ni haven't seen much discussion on the issue, I'll move it back to draft until you have a chance to respond

aagash-ni: @jattasNI @rajsite From the previous conversations on this testing strategy, we discussed about 

- Using the setContent() method to directly set HTML tags in the editor which may not be effective in a production flow where html won't be in play.
-  Using the `setMarkdown()` method to set editor content, which will be testing the whole integration along with the parser stuff. So deferred that approach as also this test has already been covered in editor spec file.

So, as per your suggestion in the current approach we are testing it out by adding contents to editor through interactions and asserting the serialized markdown. Kindly look into it and let us know your thoughts.

change/@ni-nimble-components-81243789-c09a-42e2-a4e8-bd0ae9f69b2f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Update serializer test cases",
```
jattasNI, line 3:
```suggestion
  "comment": "Update rich text component serializer test cases",
```

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -1,415 +1,306 @@
...
+            await pageObject.clickFooterButton(ToolbarButton.italics);
+            await pageObject.setEditorTextContent('Bold and Italics');
+            expect(element.getMarkdown()).toEqual('***Bold and Italics***');
+        });
+
+        it('Italics without spaces in between bold texts', async () => {
+            await pageObject.clickFooterButton(ToolbarButton.bold);
+            await pageObject.setEditorTextContent('Bold');
+            await pageObject.clickFooterButton(ToolbarButton.italics);
+            await pageObject.setEditorTextContent('italics');
```
jattasNI, line 62: A suggestion to improve readability would be to rename this page object method to `appendEditorTextContent()` since `set` implies (to me) that it replaces the contents. The other method `replaceEditorTextContent()` is pretty clear already.

But I recognize this would cause churn in other test files (possibly even client tests if they're using this page object) so it could be a separate PR or something we decide isn't worthwhile.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -1,415 +1,306 @@
...
+            await pageObject.clickFooterButton(ToolbarButton.italics);
+            await pageObject.clickFooterButton(ToolbarButton.italics);
```
jattasNI, line 64: Why is the test pressing the italics button twice here? Shouldn't that result in it remaining toggled on, so the final segment would also be both bold and italics?

Same for tests below.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
jattasNI: These tests seem valuable and easy to read to me. Great job!

My remaining question is what level of coverage this style of test should strive for. I think you did a good job of covering each styling mode and their combinations. Here are some other cases to consider. (I haven't evaluated whether they are already covered in other test suites so I'm open to pushback that they don't belong here)
1. ensure that toggling each styling mode *off* results in subsequent appended text *not* using that style
2. other ways of changing styling modes (e.g. keyboard shortcuts)
3. pasting rich text, either from another rich text editor, another rich text viewer, or arbitrary rich text

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
aagash-ni: This is working in some weird way, while looking into it the state toggling of the formatting buttons is managed by two methods, firstly on interaction with toolbar directly and the second way, the button state will be automatically toggled based on the cursor placement in the editor as the function `updateEditorButtonsState()` is binded to every transaction of the editor. So the second way is making these behaviors here.

> await pageObject.clickFooterButton(ToolbarButton.italics);

The above code will toggle the italics to active state.

> await pageObject.setEditorTextContent('italics');

Now adding text contents will be in italics and here the transaction event been triggering `updateEditorButtonsState()` method which making the italics to off state again. I am predicting it is happening because in the test case environment the cursor is not updated properly once after text content been appended. So, to make it work as real flow, I just called it twice in both cases to make it deactivates the formatting.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
aagash-ni: we will take up this work in separate PR, tracking this task [here](https://dev.azure.com/ni/DevCentral/_backlogs/backlog/ASW%20SystemLink%20LIMS/Features/?workitem=2580851)

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
aagash-ni: > ensure that toggling each styling mode off results in subsequent appended text not using that style

For this it's already been covered in few of the existing test cases. Like `Italics without spaces in between bold texts` and also in editor spec file.

> other ways of changing styling modes (e.g. keyboard shortcuts)

For this in editor spec file button states are monitored for the keyboard shortcuts and the toggle functionality has been covered in the current test cases. So, it seems like indirectly checking those cases. Like to hear your thoughts on this?

> pasting rich text, either from another rich text editor, another rich text viewer, or arbitrary rich text

This approach has been already tracked [here](https://github.com/ni/nimble/issues/1568) for integration test. Will be covering this approach there.

change/@ni-nimble-components-81243789-c09a-42e2-a4e8-bd0ae9f69b2f.json
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
jattasNI: >ensure that toggling each styling mode off results in subsequent appended text not using that style
>For this it's already been covered in few of the existing test cases.

I'm not seeing coverage for cases like ending a list by toggling the list button off after adding a few items. But I was just skimming test names so may have missed it.

Your other two responses sound good, thanks.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
rajsite: If going that way other toggle apis in the browser are represented as toggle(force?:boolean) so I'd vote for `toggleFooterButton(button: ToolbarButton, force?: boolean)`

aagash-ni: Added `toggleFooterButton()` method with the `force parameter` and updated the description as well.

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
aagash-ni: Added cases to check the toggling off the list as well.

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -98,6 +98,24 @@ export class RichTextEditorPageObject {
         await waitForUpdatesAsync();
     }
 
+    // In testing environment, when clicking on the footer button, it may not persist in the same state if any editor transaction occurs in between.
+    // This behavior is likely due to dynamic modifications of formatting button states based on cursor positions during editor transactions.
+    // By setting the "force" parameter to true in our method, will ensure the attainment of our intended state by interacting with it twice.
+    public async toggleFooterButton(
```
jattasNI, line 104: I think @rajsite's proposal was to make `force` behave like browser APIs [like this one](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute#force).

- if not specified at all, "toggles" the value by clicking it once
- if true, sets the value to true
- if false, sets the value to false

For the true and false cases we'd click the button once if it needs to change its value. If it's already at the current value we could either do nothing or click it twice if that's required by the quirk you mention in your comment.

Once we do that, this method could replace the `clickFooterButton()` method since it would have identical behavior as calling this with `force` unset. You could delete that method in this PR or [mark it deprecated](https://jsdoc.app/tags-deprecated) and remove it in a future PR.

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
aagash-ni: Updated it. If `force` sets true/false will check the button state and update it based on its active state. If unset it will just click one time (same behavior).

rajsite: Since all usages of clickFooterButton seem to be in this file please go ahead and switch to toggleFooterButton. It's literally just a find and replace.

I don't see clickFooterButton used here or [in azdo](https://dev.azure.com/ni/_search?text=clickFooterButton&type=code&lp=custom-Collection&filters=&pageSize=25) so I say we should remove it instead of deprecating it.

aagash-ni: Removed `clickFooterButton` method

PR #1638, opened 2023-11-01
Add a title slot to the Card component

kjohn1922:
# Pull Request

Add a title slot for the `nimble-card` component.

## 🤨 Rationale

The `title` slot in the `nimble-card` component will  make it easy for clients to add a title with the correct styling and to enforce consistency across usages.

The style itself is still a placeholder. We'll apply styles from a visual design in a future PR, before releasing the Routines UI with the card component.

## 👩‍💻 Implementation

- Add a `slot` with `name="title"` to the `nimble-card` template
- Add a style to apply a title font to the title slot (placeholder, may require additional styling from visual design)
- For accessibility, put a `section` at the root of the component template
    - Use `display: contents` to allow the `host` display settings to apply directly to the slotted content.
- Add a title to the matrix stories
- Add a configurable title to the storybook page.
   - Modeled after the  `nimble-banner` with similar description text for the title.

## 🧪 Testing

Updated storybook and matrix tests:

image
image

## ✅ Checklist


packages/nimble-components/src/card/template.ts
```diff
@@ -1,4 +1,9 @@
 import { html } from '@microsoft/fast-element';
 import type { Card } from '.';
 
-export const template = html<Card>`<slot></slot>`;
+export const template = html<Card>`
+    <section aria-labelledby="titleSlot">
```
kjohn1922, line 5: Here's how this appears in the accessibility tree for the storybook example:
image

kjohn1922: Is there anything here that I can/should add a unit test for? I looked for unit tests in the other components for confirming the `role` or the behavior of `aria-labledby` but I didn't see anything that looked similar.

packages/nimble-components/src/card/styles.ts
```diff
@@ -21,4 +22,12 @@ export const styles = css`
         color: ${bodyFontColor};
         background-color: ${sectionBackgroundColor};
     }
+
+    ::slotted([slot='title']) {
+        font: ${titleFont};
```
mollykreis, line 27: You should also set the color here to `titleFontColor`. It happens to be the same color right now, but that isn't a guarantee.

mollykreis, line 27: I'm not sure if it's what we want or not, but I'll point out that the title currently wraps when it gets too long. In some of our components, we limit the title to only being one line and show ellipsis if it gets too long. Since we get this current behavior for free & the card is incubating, I personally don't have an issue leaving the style as-is. However, it would be good to get an answer from designers on this before we consider the styles final.

image

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -37,9 +42,18 @@ const metadata: Meta = {
                 <${listOptionTag} value="3">Option 3</${listOptionTag}>
             </${selectTag}>
         </${cardTag}>
-    `)
+    `),
+    argTypes: {
+        title: {
+            description:
+                'Text displayed as a title inside the card. Cards should **always include a title**. The title is used to provide an accessible name to assistive technologies.<br><br>Provide the title in an `inline` element such as `<span>` that is targeted to the `title` slot.'
```
mollykreis, line 49: > Provide the title in an `inline` element such as `<span>` that is targeted to the `title` slot.

The places where we give this guidance today have style dependencies on `inline` elements being slotted (e.g. messes up ellipsis if a block element is slotted for the title). Right now, I don't believe that's the case for the title of the card. I lean towards removing that statement, at least for now.

mollykreis (approved):

packages/nimble-components/src/card/template.ts
mollykreis: I can't think of anything worthwhile to test here.

For some components, we test that the correct `role` is applied, but that's coming for free for the card because of using the `section` element. We also tend to write tests for dynamic ARIA attributes or ARIA attributes that need to get proxied from the host element to a different element within the template, but neither of those are the case for the card.

packages/nimble-components/src/card/styles.ts
kjohn1922: I brought this up with our interaction designer, and she's going to discuss it with her group. I'll add a note to my user story for finalizing the styling to include this if we decide we need to.

packages/nimble-components/src/card/tests/card.stories.ts
kjohn1922: Thanks for pointing this out. I shortened this sentence to: "Provide the title in an element targeted to the `title` slot."

jattasNI (approved):

packages/nimble-components/src/card/styles.ts
```diff
@@ -21,4 +23,13 @@ export const styles = css`
         color: ${bodyFontColor};
```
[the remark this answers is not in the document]
kjohn1922: It does still work! The screenshot in the description shows it in action. If it weren't working, the titles and the controls would all be on one horizontal line.

Setting `display: contents;` on the `section` (line 33) is what allows it to work. Elements with `display: contents` are treated as if their children are in their place, which allows the `display: flex`, `flex-direction`, and `gap` attributes to apply directly to the `slot` elements. The `slot`, in turn, also uses `display: contents` by default, which is how styles get applied to the contents inside the slots in the first place. (Thanks goes to Milan for teaching me all that this morning.)

packages/nimble-components/src/card/tests/card.stories.ts
```diff
@@ -29,6 +33,7 @@ const metadata: Meta = {
         statusLink: 'https://github.com/ni/nimble/issues/296'
     })}
         <${cardTag}>
+            <span slot="title">${x => x.title}</span>
```
rajsite, line 36: In the storybook docs page you'll notice that the select pop-up does not pop all the way out like you would expect and looks stuck.

image

This is a known issue and you can fix it by using this helper at the start of the story template:

https://github.com/ni/nimble/blob/f159a084527c7e53bac99dd43182c10c2f2f628f/packages/nimble-components/src/combobox/tests/combobox.stories.ts#L53

And looks like it causes issues in chrome now too! They might be behaving more consistently. Can you update the line in the helpers to remove mentions of specific browsers?

https://github.com/ni/nimble/blob/f159a084527c7e53bac99dd43182c10c2f2f628f/packages/nimble-components/src/utilities/tests/storybook.ts#L166

Also, handy fyi, if you are trying to debug (or in my case take a snippet) of an element that is only shown on hover / focus, you can use the [emulate a focused page](https://developer.chrome.com/docs/devtools/rendering/apply-effects/#emulate-a-focused-page) feature to keep them visible when clicking outside the window

packages/nimble-components/src/card/template.ts
```diff
@@ -1,4 +1,9 @@
...
+        <slot name="title" id="titleSlot"></slot>
```
rajsite, line 6: nit for classes and ids prefer kebab case names, ie `title-slot`

packages/nimble-components/src/card/template.ts
```diff
@@ -1,4 +1,9 @@
...
+        <slot></slot>
+    </section>
```
rajsite, line 8: noticed in the [full page accessibility tree view](https://developer.chrome.com/blog/full-accessibility-tree/#full-accessibility-tree-in-devtools) that the increment and decrement buttons for the number field are showing up as focusable which they aren't (can't tab to them or have a focus state on them). It's separate from this PR but can you create an issue for those:
image

packages/nimble-components/src/card/styles.ts
```diff
@@ -21,4 +23,13 @@ export const styles = css`
...
         background-color: ${sectionBackgroundColor};
     }
+
+    ::slotted([slot='title']) {
+        font: ${titleFont};
+        color: ${titleFontColor};
+    }
+
+    section {
```
rajsite, line 32: There aren't many styles yet but get used to organizing styles in document order: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

image

packages/nimble-components/src/card/template.ts
kjohn1922: Created an issue here: #1644

packages/nimble-components/src/card/styles.ts
kjohn1922: Moved the section style above the slot. Thanks for the link!

packages/nimble-components/src/card/tests/card.stories.ts
kjohn1922: Added `disableStorybookZoomTransform` to the card storybook and updated the comment in the utility to remove the mention of Firefox.

>Also, handy fyi, if you are trying to debug (or in my case take a snippet) of an element that is only shown on hover / focus, you can use the [emulate a focused page](https://developer.chrome.com/docs/devtools/rendering/apply-effects/#emulate-a-focused-page) feature to keep them visible when clicking outside the window

This will be super useful, thanks for this!

packages/nimble-components/src/card/template.ts
kjohn1922: Fixed. The banner has `titleSlot` in camel case [here](https://github.com/ni/nimble/blob/f159a084527c7e53bac99dd43182c10c2f2f628f/packages/nimble-components/src/banner/template.ts#L53). I can update that in this PR, too, if you don't mind pulling in unrelated files.

rajsite (approved):

Issue #1639, opened 2023-11-01
Storybook navigation broken on mobile


rajsite:
## 🧹 Tech Debt

Looks like for a bit the navigation in storybook has not worked on mobile. We should report it as an issue in storybook / find if an existing issue has been made.

Context: https://github.com/ni/nimble/pull/1635#discussion_r1378957171

m-akinc:
Blocked on [Storybook bug](https://github.com/storybookjs/storybook/issues/24431)

fredvisser:
Looks like it's resolved in the new mobile UI for Storybook 8.0 - https://github.com/storybookjs/storybook/issues/23723

PR #1642, opened 2023-11-01
Update fast-foundation dependency

### LINKED ISSUE #1422: Icon buttons used in toolbar is not getting 'tabindex' value '0' when clicked

### LINKED ISSUE #1423: Wrong toolbar button getting focused when the toolbar receives focus
# 🐛 Bug Report

In `nimble-toolbar`, button focusing operates based on the `tabindex` value. However, when clicking (mouse down) on another button with `tabindex=-1`, the button with `tabindex=0` receives focus during the mouse down event. This makes the `nimble-button` being focused when clicking another button.

_Note: This issue is reproducible only in the Chrome browser._

## 💻 Repro or Code Sample

1. Adding another control for the `nimble-toolbar` (in the below example added textarea)
2. In the click event, add a focus event to the other control (say textarea)
3. Click the button that is not recently focused (i.e. not the one with `tabindex=0`)

## 🤔 Expected Behavior

No button in the toolbar should receive focus.

## 😯 Current Behavior

The button with the `tabindex=0` receives focus.
259184551-24c12137-3fce-43a8-a96f-3b4af0ad6a3f

## 🔦 Context

It is needed in the `nimble-rich-text-editor` as we focus back on the editor when a formatting button in the `nimble-toolbar` is clicked.
For more details, see https://github.com/ni/nimble/pull/1416#discussion_r1282725591 (Issue 1)
## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser Google Chrome
* Version 115.0.5790.170

mollykreis:
# Pull Request

## 🤨 Rationale

The latest version of fast-foundation has fixes for two of our toolbar issues, so this PR pulls in the latest version.

Resolves #1422 
Resolves #1423 

## 👩‍💻 Implementation

Bump version in package file
Update the rich-text-editor page object to adjust for a change in the toolbar code

## 🧪 Testing

Used storybook to verify that the two toolbar issues are resolved

## ✅ Checklist


jattasNI (approved):

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -92,6 +92,8 @@ export class RichTextEditorPageObject {
 
     public async clickFooterButton(button: ToolbarButton): Promise<void> {
         const toggleButton = this.getFormattingButton(button);
+        const event = new Event('mousedown', { bubbles: true });
+        toggleButton!.dispatchEvent(event);
         toggleButton!.click();
```
rajsite, line 97: Do you know why the click behavior needs to fire the mousedown event?
From what I can tell the template listens for click events:
https://github.com/ni/nimble/blob/f159a084527c7e53bac99dd43182c10c2f2f628f/packages/nimble-components/src/rich-text/editor/template.ts#L52
and invoking the click method does emit those click events: https://jsbin.com/bohelazuwu/edit?html,console,output

It's not clear to me why this change is needed

mollykreis: It is specific to the focus behavior of the buttons in the toolbar. The _clicking buttons should update the checked state of the toggle button with focus_ collection of tests fail without it because they assert that the `tab-index` of each button is correct after clicking on it, but the fast toolbar updates the `tab-index` based on `mousedown` due to the ordering of mouse, click, and focus events.

rajsite: Ahh I see the discussion here: https://github.com/microsoft/fast/pull/6836

And a real interactive click goes through the mousedown, mouseup, click, progression. Same for a touch-based click, it should fire the same mouse events I think.

Maybe at some point we should have some helpers used in the page object that simulate real clicks better. Not something we need in this PR.

Could you sanity check that the focus behavior is working  / no weird issues on mobile? I vaguely remember issues with the rich text buttons on ios safari, I wonder if those behaviors are impacted @jattasNI

rajsite (approved):

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
jattasNI: I don't recall the issues we saw in the past. I tried on iOS with storybook builds from main and from this PR and didn't notice any behavior differences. (Note if others want to try, #1639 makes it hard but I got it to work in landscape mode). The experience kinda sucks because clicking the buttons gives keyboard focus to the text field which causes the keyboard to show and overlap the buttons. But if you hide the keyboard the buttons show the states that I'd expect.

mollykreis: I verified that it works correctly on Windows desktop browsers (Chrome, Edge, and Firefox) and that it works correctly on iOS safari. There is a small lag on iOS safari with the buttons being pressed and the styling updating, but that doesn't seem to be related to this change because it is also there on the version of storybook published to nimble.ni.dev.

rajsite: Yea we should do a pass on our quick interaction things (like increment / decrement) and se touch-action: manipulation for a better mobile experience: https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#manipulation

Can you make an issue for that?

PR #1646, opened 2023-11-02
Add Angular support for the nimble-card

kjohn1922:
# Pull Request

Add Angular support for the `nimble-card` component.

## 🤨 Rationale

We want to be able to use the `nimble-card` in Angular applications.

## 👩‍💻 Implementation

- Add and export a new Angular directive and module for the `nimble-card` component
- Update the component status table to "incubating" for Angular support

## 🧪 Testing

- Added a basic unit test
- Added an example to example-client-app:
image

## ✅ Checklist


kjohn1922:
"Performance Audit with Lighthouse" is failing the build. I don't see anything about Lighthouse in the README or CONTRIBUTING docs. Is there something I need to do that I'm missing?

change/@ni-nimble-components-258183d7-7c0f-4208-978c-61b6089e7796.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add Angular support for the nimble-card component",
```
mollykreis, line 3: Since the only change you made to nimble-components is updating the component-status story, you can mark the change type here as `none`.

```suggestion
  "type": "none",
  "comment": "Update the Angular status of the nimble-card component",
```

angular-workspace/projects/ni/nimble-angular/src/directives/card/nimble-card.directive.ts
```diff
@@ -0,0 +1,14 @@
+import { Directive } from '@angular/core';
```
mollykreis, line 1: We've started putting each component in their own entry point (though all the existing ones are still in the main entry point since it would be a big breaking change to move them all and we haven't decided when/how to stage that work).

So, this should be in its own entry point under `nimble-angular/card`. You can look at some of our newer components as an example if you need to (rich-text editor/viewer, table, or label providers).

change/@ni-nimble-angular-fb64a60a-b575-4965-bea7-df33b3911b9f.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: @kjohn1922, I'm moving this question to make it easier to have a discussion:

> "Performance Audit with Lighthouse" is failing the build. I don't see anything about Lighthouse in the README or CONTRIBUTING docs. Is there something I need to do that I'm missing?

I haven't seen that before, and when I run Lighthouse against the Angular app, I see a significantly higher performance score. I don't have any great suggestions other than re-running the pipeline to see if it was just some fluke. Perhaps others have a better suggestion -- @jattasNI @rajsite

change/@ni-nimble-components-258183d7-7c0f-4208-978c-61b6089e7796.json
kjohn1922: Cool, thanks, I was wondering about this. I'll update this in both PRs.

change/@ni-nimble-angular-fb64a60a-b575-4965-bea7-df33b3911b9f.json
rajsite: I've seen similar very low frequency intermittent low lighthouse scores before (usually from chromatic server issues), a new commit was pushed and is running so lets see what happens

kjohn1922: @mollykreis @rajsite It's still failing after running it a couple of times this morning. I'm a bit at a loss on how to troubleshoot the issue.

rajsite: Looks like in the lighthouse score the accessibility score is failing to calculate altogether after the change:
image

Some ideas:
- Try and reverse engineer what change is causing the issue. You can run make changes locally and re-run light house with `npm run performance` or by starting the Angular dev server and running lighthouse in chrome devtools: https://developer.chrome.com/docs/lighthouse/overview/#devtools
- It looks like an error in lighthouse itself is causing the failure. Maybe updating it could fix the issue? From `npm outdated` you can see the version we are using is a bit out of date: 
image

rajsite: @jattasNI @kjohn1922 my merge from main did two things, update lighthouse and change the accessibility status check to warn (because the upgrade caused it to go below 88 with new checks).

After merging in this branch it does not look like it fixed the error being introduced in this PR that causes the accessibility score to fail to calculate altogether: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1699290639438-32861.report.html

So I'm still interested in seeing if you are able to narrow down what causes the score to fail to calculate from the changes. You might try reducing to as simple a nimble-card as possible and incrementally adding child elements until you can identify the issue.

While we have allowed a lowering of the accessibility score to be allowed through, I'm not sure about allowing something through that completely breaks the calculation. Thoughts?

kjohn1922: Nice! Thanks for pushing the update to my branch.
image

It's still getting some questionable results:
image
I see low scores if I run it locally against `main`, but I've only seen the `NaN` result in my branch, for some reason. I get that NaN even if I remove the new card component from the example app altogether, though, so I don't think it's anything I'm doing that's causing the issue.

rajsite: @kjohn1922 yep, I think the NaN result corresponds to my comment saying accessibility fails to calculate (might be another case of simultaneous posting if you haven't seen my comment :P)

kjohn1922: Yep, I posted a reply without seeing your latest comment. Not sure how that keeps happening, haha. I'll look into the NaN value some more.

kjohn1922: Still debugging, but I'm not sure what to make of this: If I remove `aria-labelledby="title-slot"` from the `section` in the card template, the issue goes away and I get a value instead of NaN from the accessibility check.

rajsite: Huh, wonder if the library  reporting the error is making bad assumptions about ID formats. Looks like other IDs snuck by formatted as camelCase instead of kebab-case. Could try switching back to that, ie titleSlot

kjohn1922: Tried camel case, that was not the problem.

It gets a value, though, if I explicitly set `role="region"` on the `section`. The Lighthouse report was actually trying tell me what was wrong, I think:
image

A `section` implicitly has a `role` of `region`, but I think it's looking for it to be set explicitly. I feel like I'm trying to read between the lines, though, and I'm not sure why that results in a NaN value rather than a hit to the score. The other issues listed in the report don't have the "Error!" text, though, and that error goes away if I add `role="region"`.

rajsite: >  I feel like I'm trying to read between the lines

It's possible the axe extension would give more info, built on the same library: https://www.deque.com/axe/browser-extensions/

> The other issues listed in the report don't have the "Error!" text, though, and that error goes away if I add role="region".

That sounds like an okay workaround to unblock the PR! Can leave a comment and reference to the issue in the template.

packages/nimble-components/src/card/template.ts
```diff
@@ -2,7 +2,8 @@ import { html } from '@microsoft/fast-element';
 import type { Card } from '.';
 
 export const template = html<Card>`
-    <section aria-labelledby="title-slot">
+    <!-- Explicitly set role to work around Lighthouse error. See https://github.com/ni/nimble/issues/1650. -->
```
rajsite, line 5: On a phone right now but double check the comment style in other templates

kjohn1922: Updated the comment to use a template literal.

change/@ni-nimble-angular-fb64a60a-b575-4965-bea7-df33b3911b9f.json
kjohn1922: Pushed a change with the role set and a code comment pointing to the issue.

change/@ni-nimble-components-258183d7-7c0f-4208-978c-61b6089e7796.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
rajsite, line 2: There are changes in the nimble package now
```suggestion
  "type": "patch",
```

rajsite (approved):

angular-workspace/projects/example-client-app/src/app/app.module.ts
```diff
@@ -10,6 +10,7 @@ import { NimbleTextAreaModule, NimbleTextFieldModule, NimbleNumberFieldModule, N
     NimbleCardButtonModule, NimbleDialogModule, NimbleRadioGroupModule, NimbleRadioModule, NimbleSpinnerModule,
```
rajsite, line 10: @kjohn1922 looks ready to move out of draft, dunno if you were waiting on anything else

kjohn1922: I was waiting for sign-off from @mollykreis to move it out of draft, but I can go ahead and do that with your approval

mollykreis (approved):

rajsite:
Don't need to wait on @jattasNI looks well covered 👍 thanks for digging in and finding out what was breaking axe core in lighthouse. If you're feeling ambitious can try to reproduce and report it upstream to https://github.com/dequelabs/axe-core

PR #1649, opened 2023-11-03
Update radio button documentation for Angular difference

m-akinc:
# Pull Request

## 🤨 Rationale

Addresses AzDO bug: https://dev.azure.com/ni/DevCentral/_workitems/edit/2563164

Copied from bug:
The Angular `RadioControlValueAccessor` and associated `RadioControlRegistry` handle selection/deselection of related radio buttons based on `name` or `formControlName` and parent form. These look specifically at the `name` input property on `RadioControlValueAccessor`, not the `name` attribute/property on the `Radio` element itself. While the `RadioGroup` element propagates `name` to the child `Radio` elements, that does not cause `RadioControlValueAccessor`'s `name` to be set. So unless `name` is explicitly set in the template on the `Radio` elements, the directives will all have undefined values and all act as if they are part of one radio group.

Inititally, I tried using a `MutationObserver` to update the CVA's `name` property when the host element's `name` changed. Unfortunately, I could not create the observer before the CVA initialized the checked states of the radio buttons. So initially there could still only be at most one radio button checked (among multiple different radio groups), even though subsequent interaction would allow one checked button per group.

I decided that just documenting the need to set `name` on every radio button (in Angular) was reasonable, especially since it mirrors the requirement to put `formControlName` on every radio button when using reactive forms.

## 👩‍💻 Implementation

See latest comment in the linked bug for rationale for a documentation change rather than a behavior change.

While investigating, I found an incorrect type in the `NimbleRadioDirective`, so I fixed that, too.

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/radio-group/tests/radio-group.stories.ts
```diff
@@ -75,7 +75,7 @@ const metadata: Meta<RadioGroupArgs> = {
         },
         name: {
             description:
-                'Radio buttons whose values are mutually exclusive should set the same `name` attribute. Setting the name on the group sets it on all child radio buttons.'
+                'Radio buttons whose values are mutually exclusive should set the same `name` attribute. Setting the name on the group sets it on all child radio buttons. When using radio buttons in an Angular form, you must explicitly set either `name` or `formControlName` on each radio button rather than only setting `name` on the group.'
```
m-akinc, line 78: I considered creating an `.mdx` and putting this in a separate Angular usage section, but I thought it would be harder to miss here in the property doc.

jattasNI, line 78: Could you update the Angular example app to follow this guidance?

change/@ni-nimble-angular-eeee6b1e-cb23-4a3c-ab87-8f1b30f413fe.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: nit: could you copy a bit more of the rationale from the AzDO issue into the PR description? Saves clicks now and if exploring history in the future and makes it more accessible to people without AzDO access.

jattasNI (approved):

atmgrifter00 (approved):

packages/nimble-components/src/radio-group/tests/radio-group.stories.ts
rajsite: Also not 100% clear on the guidance, can't tell if it's saying Angular needs to set the same name on the group and all the children or just the children. I think I would suspect just the children.

m-akinc: I've updated the example and changed the guidance to explicitly say that setting the name on the group doesn't work in the Angular forms case.

rajsite (approved):

PR #1594, opened 2023-10-10
Spec update to support @mention in rich text components

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Spec document/HLD update to support `@mention` for `nimble-rich-text-editor` and `nimble-rich-text-viewer`.

[Epic 2464517](https://dev.azure.com/ni/DevCentral/_workitems/edit/2464517): Support for @mention in comments

## 👩‍💻 Implementation

Here is the [storybook build](https://60e89457a987cf003efc0a5b-fvlbopmfvm.chromatic.com/) to try testing out the `@mention` prototype in rich text components. 

## 🧪 Testing

N/A

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1652, opened 2023-11-06
Rich Text Editor | Add Configuration component for @mention support

aagash-ni:
# Pull Request

## 🤨 Rationale
The list of items needs to be populated in the mention popup of the editor when entering configured characters (mentioning feature), in our case user list need to be populated when entering `@` character in the editor. To get the configuration of the users from the client and pass on to the editor, Configuration components are needed.

Link to the Feature: [Feature 2351134](https://dev.azure.com/ni/DevCentral/_workitems/edit/2351134): Support user @ mention and send a plain text email notification

## 👩‍💻 Implementation

- Added user mapping component `nimble-mapping-user` by extending the mapping base class. Added attribute `display-name` for textual representation of user name.
- Added `RichTextMention` abstract class, which will hold
         -  `mentionInternals` contains mention configuration, icon, character and pattern data which will be used in the editor.
         -  `checkValidity()` and `validity` getter - will return the validity of the configuration.
         - Generated configuration from mapping element

- Added `MentionValidator` and `MentionInternals` model classes.
- Added `RichTextMentionBaseValidator` base validator to check the below cases
          - unsupportedMappingType
          - duplicateMappingMentionHref
          - missingMentionHrefValue
          - unsupportedMentionHrefValue
          - missingDisplayNameValue
          - missingPatternAttribute
          - unsupportedPatternValue

- Added `nimble-rich-text-mention-users` component derived from the base class which will hold definitions for the abstract methods to configure user mentioning in editor. 

## 🧪 Testing

- Added unit tests for the component.
- Added basic storybook doc, will be linked to editor and viewer mdx files once created.

## ✅ Checklist


packages/nimble-components/src/rich-text-mention/mention-base/models/rich-text-mention-base-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+            mappingClass => mapping instanceof mappingClass
+        ));
+        this.setConditionValue('unsupportedMappingType', !valid);
+    }
+
+    private validateUniqueMentionHref(
+        mentionHref: (string | undefined)[]
+    ): void {
+        const typedKeys = mentionHref.map(x => x?.toString() ?? undefined);
+        const invalid = new Set(typedKeys).size !== typedKeys.length;
```
m-akinc, line 60:
```suggestion
        const invalid = new Set(mentionHref).size !== mentionHref.length;
```

packages/nimble-components/src/rich-text-mention/mention-base/models/rich-text-mention-base-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    private validateNoMissingMentionHref(mappings: MappingMentionBase[]): void {
+        const invalid = mappings.some(
+            mapping => mapping.mentionHref === undefined
+        );
+        this.setConditionValue('missingMentionHrefValue', invalid);
+    }
+
+    private validateHref(
+        mappings: (string | undefined)[],
```
m-akinc, line 72: rename `mappings` to `mentionHrefs`

packages/nimble-components/src/rich-text-mention/mention-base/models/rich-text-mention-base-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+        pattern: string | undefined
+    ): void {
+        const regexPattern = new RegExp(pattern!);
+        const valid = mappings.some(href => (href ? regexPattern.test(href) : false));
```
m-akinc, line 76:
```suggestion
        const valid = mappings.every(href => href === undefined || regexPattern.test(href));
```
Don't mark invalid just because some `href` is missing (that triggers `missingMentionHrefValue`), but make sure all are `undefined` or match the pattern.

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
m-akinc, line 1: The story for the user mentions element should be like [the one for mapping spinner](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/mapping/spinner/tests/mapping-spinner.stories.ts). The `title` should be `Internal/Rich Text`, so that it's not visible in the tree. The story will not render anything but will provide documentation for the element and the configuration options. We then should create `.mdx` files for the rich text editor and viewer documentation, which will pull in the documentation for the  configuration elements. See the [.mdx](https://github.com/ni/nimble/blob/880581c89efa6711349f29a99ba9b5c8110c3b09/packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.mdx#L27) for our enum-text table column for an example of how that works.

packages/nimble-components/src/mapping/mention-user/index.ts
m-akinc, line 1: This element should also have a Storybook file that is like [the one for mapping spinner](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/mapping/spinner/tests/mapping-spinner.stories.ts). The documentation should be pulled in by `.mdx` files for the rich text editor and viewer (which will need to be created).

packages/nimble-components/src/rich-text-mention/mention-users/index.ts
```diff
@@ -0,0 +1,77 @@
...
+export class RichtextMentionUsers extends RichTextMentionBase<
+RichTextMentionConfig,
+RichTextMentionUsersValidator
+> {
+    private readonly character = '@';
+
+    private readonly icon = '';
+
+    public override createValidator(): RichTextMentionUsersValidator {
+        return new RichTextMentionUsersValidator(this.mentionInternals);
```
m-akinc, line 34:
```suggestion
        return new RichTextMentionBaseValidator(this.mentionInternals, baseValidityFlagNames, [MappingMentionUser]);
```
I don't think you need a separate `RichTextMentionUsersValidator` class, since it doesn't add anything to the base validator. Just remove `abstract` from the base validator class so you can construct it. (Then remove the file for the other validator.)

packages/nimble-components/src/rich-text-mention/mention-base/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    Subscriber,
+    ViewTemplate
+} from '@microsoft/fast-element';
+import type { MappingConfig } from './models/mapping-config';
+import { MappingMentionBase } from '../../mapping/mention-base';
+import type { RichTextMentionBaseValidator } from './models/rich-text-mention-base-validator';
+import type { ListOption } from '../../list-option';
+import { RichTextMention } from '../base';
+
+export type MappingConfigs = Map<string, MappingConfig>;
```
m-akinc, line 14:
```suggestion
export type MappingConfigs = ReadonlyMap<string, MappingConfig>;
```

m-akinc (approved):

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
rajsite: Ahh I see it happened here: https://github.com/ni/nimble/pull/1594#discussion_r1363906239

I got too focused on the suggested attribute name and didn't think back to the mapping API.

@jattasNI is this thread compelling enough to use the api `<nimble-mapping-user key="http://user/blah" text="John Doe">` in order to be compatible with the Mapping Base class (so we can re-use mappings in other contexts like table in the future)

or should we fork and have a specific category of mention-mapping that is a distinct base class with `mention-href` and `display-name`?. I lean toward aligning with the mapping base class I think. Would make it easier to align with the table but not deal breaking either, the mapping elements are pretty well decoupled from the columnConfig sent to the view so we could probably make either work, but Mapping would be more inline.

packages/nimble-components/src/rich-text-mention/base/index.ts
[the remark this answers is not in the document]
aagash-ni: While referring table, all the configuration components are kept in separate folder `table-column`. Is that only for configuration component or column specific directory? Shall we this folder under `rich-text/mention` ? Like to hear your thoughts? @m-akinc

packages/nimble-components/src/rich-text-mention/mention-base/models/mapping-user-config.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export class MappingUserConfig extends MappingConfig {
+    public constructor(
+        mentionHref: string | undefined,
+        displayName: string | undefined
+    ) {
+        super(mentionHref, displayName);
+    }
+
+    public override createListView(): ViewTemplate<ListOption> {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-base/index.ts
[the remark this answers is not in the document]
aagash-ni: Just for the typing concern in editor I made it into two separate classes. Updated it to a single base class and moved all to the `base` directory

packages/nimble-components/src/rich-text-mention/mention-users/index.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Rich Text Mention that will map user url and name
+ */
+export class RichtextMentionUsers extends RichTextMentionBase<
+RichTextMentionConfig,
+RichTextMentionUsersValidator
+> {
+    private readonly character = '@';
+
+    private readonly icon = '';
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    @attr
+    public pattern?: string;
+
+    public checkValidity(): boolean {
+        return this.mentionInternals.validConfiguration;
+    }
+
+    public get validity(): RichTextMentionValidity {
+        return {};
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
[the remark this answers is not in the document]
aagash-ni: Combined both validators on a single class.

packages/nimble-components/src/rich-text-mention/mention-base/models/rich-text-mention-base-validator.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text-mention/mention-users/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-base/index.ts
aagash-ni: Updated

packages/nimble-components/src/mapping/mention-user/index.ts
aagash-ni: Once the alignment on mapping base class implementation for mention-mapping, I will update the storybook file.

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
aagash-ni: Updated the storybook file. I will link this in `.mdx` file of editor and viewer once created.

packages/nimble-components/src/rich-text-mention/mention-base/models/rich-text-mention-base-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+    'unsupportedMappingType',
+    'duplicateMappingMentionHref',
+    'missingMentionHrefValue',
+    'unsupportedMentionHrefValue',
+    'missingDisplayNameValue',
+    'missingPatternAttribute'
+] as const;
+
+/**
+ * Validator for RichTextmentionBase
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
jattasNI: I agree we should share the base class. And I may be missing something but if we're ok with a string type for the mention href then I'm good with using the existing `key` of type `string` rather than making the class generic.

But the name/text behavior is making me think through the purpose of the base class. In theory we could map from a key to any type of value (in fact, we do map to icon elements and to spinner elements in some of the table column mappings), so why does the value on the base class have name `text` and type `string`? I think it's because the `text` attribute is not the value, it's a textual summary of the value. In table mappings we always need a textual summary of the value to display it in group headers and tooltips. But will all mappings (even non-table ones) need a textual summary?

If not, maybe we need a new mapping base class that doesn't include the text summary and a table column mapping base class that does.

If all mappings will need a text summary then we could keep it as-is and agree to call it `text`. Or, if the summary might need to be derived from several other attributes (name, email, department), then we could change the base property `text` not to be an attribute and instead be an abstract member which concrete mapping classes must provide by computing its value from their attributes. In that scenario this element would continue to use a `display-name` attribute and it would return its value in the `text` override.

Right now I'm leaning towards using the base class as-is but keeping the door open to my last proposal if we find mappings need that additional information to contribute to the display text.

packages/nimble-components/src/rich-text-mention/base/index.ts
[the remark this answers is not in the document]
m-akinc: That is a good point. I'm now thinking it may be fine as it is. I would say leave it like this, and let's see if Milan or Jesse suggest to move them.

packages/nimble-components/src/rich-text-mention/mention-base/index.ts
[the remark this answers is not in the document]
m-akinc: > Just for the typing concern in editor I made it into two separate classes.

Can you clarify this? I'm still not sure what you mean. But I want to know, because it may have been a good reason.

m-akinc:
I accidentally approved the PR by habit, even though we decided we would just leave a message approving, so that official approval wasn't given until the tests had been added and reviewed. I will keep an eye out for the tests and review those as well.

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
```diff
@@ -0,0 +1,30 @@
+import { html } from '@microsoft/fast-element';
+import type { Meta, StoryObj } from '@storybook/html';
+import { createUserSelectedThemeStory } from '../../../utilities/tests/storybook';
+import { hiddenWrapper } from '../../../utilities/tests/hidden';
+
+const metadata: Meta = {
+    title: 'Internal/Rich Text',
```
jattasNI, line 7: It's best to keep the titles in sync with the file names
```suggestion
    title: 'Internal/Rich Text Mention Users',
```

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export default metadata;
+
+export const richTextMentionUsers: StoryObj = {
+    render: createUserSelectedThemeStory(hiddenWrapper(html`<style></style>`)),
+    argTypes: {
+        pattern: {
+            description:
+                'Using this regex, the user ID will be extracted from the user href during the parsing and serializing of comments in the editor.',
+            control: { type: 'none' }
```
jattasNI, line 26: Can we make this a string control with a useful default value?

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export default metadata;
+
+export const richTextMentionUsers: StoryObj = {
+    render: createUserSelectedThemeStory(hiddenWrapper(html`<style></style>`)),
+    argTypes: {
+        pattern: {
+            description:
+                'Using this regex, the user ID will be extracted from the user href during the parsing and serializing of comments in the editor.',
```
jattasNI, line 25:
```suggestion
                'A regex that is used to extract the user ID from user hrefs during the parsing and serializing of markdown in the editor.',
```

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
jattasNI: We discussed this today and Milan was kind enough to implement what we decided in #1656. You'll need to adapt this PR in response to that change (either by merging from main if it goes in today or from its branch if he doesn't have time to wrap up the testing gap until tomorrow).

packages/nimble-components/src/rich-text-mention/base/index.ts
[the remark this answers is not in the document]
rajsite: Chatted with Jesse and seems okay to keep it like this to make it easier to parallel the table-column patterns it is aligning with.

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
rajsite: Yea I didn't get to finish quite yet, I don't think the implementation will change significantly, I primarily need to adjust tests and angular/blazor bindings.

The PR changes to generic key and removes the text attribute from the mapping base class.
The user mapping class should extend that base class and provide string as the key type and a display-name attribute.

packages/nimble-components/src/rich-text-mention/mention-users/index.ts
```diff
@@ -0,0 +1,82 @@
+import { DesignSystem } from '@microsoft/fast-foundation';
```
rajsite, line 1: The path should correspond directly to the name
rich-text-mention-users -> rich-text-mention/users/index.ts

packages/nimble-components/src/rich-text-mention/base/models/mapping-user-config.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Mapping configuration corresponding to a user mapping
+ */
+export class MappingUserConfig extends MappingConfig {
+    public override createListView(): ViewTemplate<ListOption> {
+        return html`<${listOptionTag} value="${this.mentionHref ?? ''}">${
+            this.displayName ?? ''
+        }</${listOptionTag}>`;
+    }
```
rajsite, line 13: I'm a bit concerned about this. Firstly we should not dynamically construct templates. They should be statically defined, usually by being in a separate file and imported. 

I can't provide a more concrete alternative without knowing more about the use case though. Is the concern about how flexible the `@mention` drop down picker is? I think initially we can keep it simple and say it's list options for all mention types, and the only state needed is the mentionHref and displayHref which is in the MappingConfig base class. MappingUserConfig should just be empty.

```
MappingUserConfig extends MappingConfig {}
```

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
aagash-ni: Merged the changes of `generic-mapping-base-class` branch and extended the mapping base class with string type and added `display-name` attribute. Retargeted the current PR to `generic-mapping-base-class` branch. I will change again to main once https://github.com/ni/nimble/pull/1656 PR is merged.

change/@ni-nimble-components-876e3564-bb4e-4e2e-a165-f2efd10f9986.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add configuration component for @mention support",
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+import { createUserSelectedThemeStory } from '../../../utilities/tests/storybook';
+import { hiddenWrapper } from '../../../utilities/tests/hidden';
+
+const metadata: Meta = {
+    title: 'Internal/Rich Text',
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    'The `nimble-rich-text-mention-users` element will use the `nimble-mapping-mention-user` element as its user list content and utilize the pattern attribute to parse user href links, generating configuration for the mention list. This element will also include various validation checks to ensure the content format and pattern are correctly verified.'
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mapping-user-config.ts
aagash-ni: Removed `createListView()` method and only having `mentionHref` and `displayName` in base class.

packages/nimble-components/src/rich-text-mention/mention-users/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/mention-users/tests/rich-text-mention-users.stories.ts
aagash-ni: Updated

packages/nimble-components/src/mapping/mention-user/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { MappingMentionBase } from '../mention-base';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-mapping-mention-user': MappingMentionUser;
+    }
+}
+
+/**
+ * Extended base class, as mentioning users requires only mentionHref and displayName
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mapping-config.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Common state shared across Mapping Config
+ */
+export abstract class MappingConfig {
+    public constructor(
+        public readonly mentionHref: string | undefined,
+        public readonly displayName: string | undefined
```
jattasNI, line 7: Should `displayName` be on the derived class in the state objects just like it is in the config element objects?

packages/nimble-components/src/rich-text-mention/users/index.ts
```diff
@@ -0,0 +1,73 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-rich-text-mention-users': RichtextMentionUsers;
+    }
+}
+
+/**
+ * Rich Text Mention that will map user url and name
+ */
+export class RichtextMentionUsers extends RichTextMention<
```
jattasNI, line 26: Use consistent capitalization: `RichText`, not `Richtext`. Same for the variable further down this file.

packages/nimble-components/src/rich-text-mention/users/models/rich-text-mention-users-validator.ts
```diff
@@ -0,0 +1,39 @@
...
+    RichTextMentionValidator
+} from '../../base/models/mention-validator';
+
+const usersValidityFlagNames = [
+    ...baseValidityFlagNames,
+    'missingDisplayNameValue'
+] as const;
+
+/**
+ * Validator for RichtextMentionUsers
```
jattasNI, line 16:
```suggestion
 * Validator for RichTextMentionUsers
```

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -0,0 +1,109 @@
...
+        const invalid = mappings.some(mapping => mapping.key === undefined);
+        this.setConditionValue('missingMentionHrefValue', invalid);
+    }
+
+    private validateHref(
+        mentionHrefs: unknown[],
+        pattern: string | undefined
+    ): void {
+        const regexPattern = new RegExp(pattern!);
+        const valid = mentionHrefs.every(
```
jattasNI, line 102: Is this logic correct? It looks like we're saying it's valid to provide an `href` that's `undefined` or not a string. Why would we allow that?

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private validateUniqueMentionHref(
+        mentionHref: (string | undefined | unknown)[]
+    ): void {
+        const invalid = new Set(mentionHref).size !== mentionHref.length;
+        this.setConditionValue('duplicateMappingMentionHref', invalid);
+    }
+
+    private validateNoMissingMentionHref(mappings: Mapping<unknown>[]): void {
```
jattasNI, line 92: Could you change this to accept `mappingHrefs` like other helpers that only access the mapping keys?

packages/nimble-components/src/rich-text-mention/users/models/rich-text-mention-users-validator.ts
```diff
@@ -0,0 +1,39 @@
...
+    public override validate(
+        mappings: Mapping<MentionHref>[],
+        pattern: string
+    ): void {
+        super.validate(mappings, pattern);
+        this.validateNoMissingDisplayName(mappings);
+    }
+
+    private validateNoMissingDisplayName(mappings: MappingMentionUser[]): void {
+        const invalid = mappings.some(
```
jattasNI, line 34: I think this will fail if there are no mappings but we want to support that case, for example while the app is fetching users to map.

When you write tests please add coverage for that case.

packages/nimble-components/src/rich-text-mention/users/index.ts
```diff
@@ -0,0 +1,73 @@
...
+import { RichTextMentionUsersValidator } from './models/rich-text-mention-users-validator';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-rich-text-mention-users': RichtextMentionUsers;
+    }
+}
+
+/**
+ * Rich Text Mention that will map user url and name
```
jattasNI, line 24:
```suggestion
 * Rich Text user mention configuration element which will have MappingMentionUser elements as children
```

jattasNI (commented):
Looks ready for testing to me

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
m-akinc: The conditions here should be a compliment to those used in `validateNoMissingMentionHref`, i.e. if `validateNoMissingMentionHref` catches `undefined` hrefs, `validateHref` should ignore `undefined` hrefs. However, it does seem like hrefs that are defined but not a string should trigger the `unsupportedMentionHrefValue` flag.

packages/nimble-components/src/rich-text-mention/base/models/mapping-config.ts
aagash-ni: Currently we will support only list options for all the mention types, the basic states for the list options are `mentionHref` and `displayName`. So, the base is holding both states. Is this ok to have these two states in the base class `MappingConfig` or we need to move the `displayName` to the derived class `MappingUserConfig` like below. Like to know your thoughts?

```
export class MappingUserConfig extends MappingConfig {
    public constructor(mentionHref: string | undefined, public readonly displayName: string | undefined) {
        super(mentionHref);
    }
}
```

packages/nimble-components/src/mapping/mention-user/index.ts
aagash-ni: Added storybook file for user mention mapping element.

packages/nimble-components/src/rich-text-mention/users/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/users/models/rich-text-mention-users-validator.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/users/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: As the `test()` method will expect only `string` as input as well as the `url` should be of string type, `href` need to be validated for `string` type. So in that case even a defined `href` of other types will also be validated under `unsupportedMentionHrefValue` flag.

packages/nimble-components/src/rich-text-mention/base/models/mapping-config.ts
jattasNI: I thought about it some more and decided it makes sense to keep on this `MappingConfig` base class and not on `MappingUserConfig`. It will be a requirement of every mapping type to be able to set the `displayName`. 

They may each provide that differently. `MappingMentionUser` gets it via the `display-name` attribute but a new mapping might provide it with a different attribute or by computing it from multiple attributes.

packages/nimble-components/src/mapping/mention-user/index.ts
```diff
@@ -0,0 +1,25 @@
+import { attr } from '@microsoft/fast-element';
```
rajsite, line 1: We should just refer to this as `mapping-user`
It's just how nimble will represent a user generally (there is nothing mention specific, this is just user data)
So in the file paths and file contents of mapping-user we should clean out references to mentions

packages/nimble-components/src/mapping/base/types.ts
```diff
@@ -1 +1,4 @@
 export type MappingKey = string | number | boolean;
+
+/** For user mapping, String-type keys are preferred since they are typically used for URLs. */
+export type MentionHref = string;
```
rajsite, line 4:
```suggestion
/** For user mapping, a string formatted as a URL. */
export type MappingUserKey = string;
```

packages/nimble-components/src/mapping/mention-user/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import { template } from '../base/template';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-mapping-mention-user': MappingMentionUser;
+    }
+}
+/**
+ * Maps key (url) to user display name
+ * Non-visible configuration element for mapping rich text user mention keys to the user name to display
```
rajsite, line 14: User mappings could contain other info in the future
```suggestion
 * Maps key (url) to a user
 * Non-visible configuration element for mapping rich text user mention keys to a user
```

packages/nimble-components/src/mapping/mention-user/tests/mapping-mention-user.stories.ts
```diff
@@ -0,0 +1,35 @@
+import { html } from '@microsoft/fast-element';
```
rajsite, line 1: Make sure to clean-up references to mention in the filename / contents here too

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+import { Mapping } from '../../mapping/base';
+
+export type MappingConfigs = ReadonlyMap<string, MappingConfig>;
+
+export interface RichTextMentionConfig {
+    mappingConfigs: MappingConfigs;
+}
+
+/**
+ * The base class for Mention mapping configuration
```
rajsite, line 27:
```suggestion
 * The base class for Mention configuration
```

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export abstract class RichTextMention<
+    TMentionConfig extends RichTextMentionConfig,
+    TValidator extends RichTextMentionValidator<[]>
+>
+    extends FoundationElement
+    implements Subscriber {
+    public readonly mentionInternals: MentionInternals<TMentionConfig> = new MentionInternals(this.getMentionInternalsOptions());
```
rajsite, line 35: Mark internal

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    @attr
+    public pattern?: string;
```
rajsite, line 38: Add a doc comment

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    /** @internal */
+    public validator = this.createValidator();
```
rajsite, line 41: Mark it readonly and move it up so it is defined right after mentionInternals is defined.

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    public checkValidity(): boolean {
+        return this.mentionInternals.validConfiguration;
+    }
+
+    public get validity(): RichTextMentionValidity {
+        return this.validator.getValidity();
+    }
+
+    public handleChange(source: unknown, args: unknown): void {
```
rajsite, line 58: mark it internal

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (source instanceof Mapping && typeof args === 'string') {
+            this.updateMentionConfig();
+        }
+    }
+
+    public abstract createValidator(): TValidator;
```
rajsite, line 64: mark protected

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -0,0 +1,36 @@
...
+    readonly pattern: string;
+}
+
+/**
+ * Internal mention state
+ */
+export class MentionInternals<TMentionConfig> {
+    /**
+     * The relevant, static configuration which will hold mentioning info of view, character, icon and pattern
+     */
```
rajsite, line 15: this doc comment is on mentionConfig which is not the static configuration

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -0,0 +1,36 @@
...
+     * Whether this mention has a valid configuration.
+     */
+    @observable
+    public validConfiguration = true;
+
+    public readonly icon: string;
+
+    public readonly character: string;
+
+    public readonly pattern: string;
```
rajsite, line 29: Add doc comments

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -0,0 +1,109 @@
...
+        this.setConditionValue('missingMentionHrefValue', invalid);
+    }
+
+    private validateHref(
+        mentionHrefs: unknown[],
+        pattern: string | undefined
+    ): void {
+        const regexPattern = new RegExp(pattern!);
+        const valid = mentionHrefs.every(
+            href => href === undefined
```
rajsite, line 103: all the hrefs need to validate with the `new URL()` constructor, ie, not throw

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -0,0 +1,109 @@
...
+            this.untrack(name);
+        }
+        this.updateMentionInternalsFlag();
+    }
+
+    private updateMentionInternalsFlag(): void {
+        this.mentionInternals.validConfiguration = this.isValid();
+    }
+
+    private validatePattern(pattern: string | undefined): void {
```
rajsite, line 71: pattern should be tested against the RegExp constructor which can throw: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp#exceptions

packages/nimble-components/src/mapping/mention-base/index.ts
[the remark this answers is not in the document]
rajsite: Don't think it's blocking but I haven't finished #1656 , plan to finish it tomorrow

packages/nimble-components/src/mapping/mention-user/index.ts
aagash-ni: Updated as `mapping/user`.

packages/nimble-components/src/mapping/base/types.ts
aagash-ni: Updated

packages/nimble-components/src/mapping/mention-user/tests/mapping-mention-user.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: Updated the validation with `new URL(url)` to check href

packages/nimble-components/src/mapping/mention-user/index.ts
aagash-ni: Updated

change/@ni-nimble-components-876e3564-bb4e-4e2e-a165-f2efd10f9986.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I can't recall if we agreed on a workflow for these PRs that don't have tests, but I'd propose we move it back to Draft once the initial implementation review is approved. That will keep it out of reviewers' view until the tests are added. It looks like it's pretty much ready for that but I'll let @rajsite comment on his status once he looks at the latest response to his feedback

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+describe('RichTextMentionUsers', () => {
+    let element: RichTextMentionUsers;
+    let connect: () => Promise<void>;
+    let disconnect: () => Promise<void>;
+
+    async function setup(
+        mappings: BasicUserMentionMapping[],
+        pattern = ''
+    ): Promise<Fixture<RichTextMentionUsers>> {
+        return fixture<RichTextMentionUsers>(html`
```
rajsite, line 24: Do a prettier-ignore to make the template readable, this whitespace is very confusing

change/@ni-nimble-components-876e3564-bb4e-4e2e-a165-f2efd10f9986.json
rajsite: Yea it looks like there are a few tests but guessing there will be more around testing out state updates / dynamic changes so yea looks ready to finish adding tests

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
aagash-ni, line 1: I have added few cases to cover the dynamic update of `pattern` and `mappings` and its respective configuration updates. Please let us know if there are any other suggestions/methods, we can test this? Like to hear your thoughts?

jattasNI: @m-akinc can you take a first pass at evaluating the test coverage? I'll re-request your review.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+                { key: 'user:1', displayName: 'user' },
+                { key: 'user:1', displayName: 'user' }
+            ]));
+            await connect();
+            expect(element.checkValidity()).toBeFalse();
+            expect(element.validity.duplicateMappingMentionHref).toBeTrue();
+            expect(element.mentionInternals.mentionConfig).toBe(undefined);
+        });
+
+        it('is invalid with mismatch pattern', async () => {
```
m-akinc, line 241: This doesn't seem like a different case than the next test ('is invalid with invalid mismatching key'). This one's pattern says "invalidPattern", but it's not really invalid, is it? It's just that the keys don't match it as a regex? Unless I'm missing something, I think you can remove this test.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+                [
+                    { key: 'user:1', displayName: 'user' },
+                    { key: 'user:2', displayName: 'user' }
+                ],
+                '(invalid'
+            ));
+            await connect();
+            expect(element.checkValidity()).toBeFalse();
+            expect(element.validity.unsupportedPatternValue).toBeTrue();
+            expect(element.validity.unsupportedMentionHrefValue).toBeTrue();
```
m-akinc, line 293: If it's not too much trouble, I think we should not be setting `unsupportedMentionHrefValue` when we know the pattern is invalid.

m-akinc (approved):

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
aagash-ni: Updated it to avoid setting the `unsupportedMentionHrefValue` when the pattern is invalid.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
aagash-ni: Removed this case.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    it('should have empty Map as mappingConfigs when there is no mappings', async () => {
+        ({ element, connect, disconnect } = await setup([]));
+        await connect();
+        const mappingConfig = new Map([]);
+        expect(element.mentionInternals.mentionConfig?.mappingConfigs).toEqual(
+            mappingConfig
+        );
+    });
+
+    it('Should have undefined mentionConfig when an invalid pattern is assigned', async () => {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    ): MappingUser[] {
+        return mappings.map(mapping => {
+            const mappingUser = new MappingUser();
+            mappingUser.displayName = mapping.displayName;
+            mappingUser.key = mapping.key;
+            return mappingUser;
+        });
+    }
+
+    function getMappingConfig(
```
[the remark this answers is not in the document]
aagash-ni: Removed `getMappingConfig()` method instead asserted the map size, keys, and the respective values for keys.

jattasNI (approved):

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -0,0 +1,147 @@
...
+            : mentionHrefs.some(href => {
+                return (
+                    href === undefined
+                      || typeof href !== 'string'
+                      || this.isInvalidUrl(href)
+                      || !RegExp(pattern!).test(href)
+                );
+            });
+        this.setConditionValue('unsupportedMentionHrefValue', invalid);
+    }
```
rajsite, line 127: I find the logic of this confusing. This should be modified in two ways:
1. Don't rely on implicit validation from other code paths. Here it is assuming that some other functions have run first and set some flags. Don't access the other flags
2. I'd expect each code path to result in `unsupportedMentionHrefValue` being set. As written there are code paths that don't set the flag at all.

This is the error reporting. If I told you the error is `unsupportedMentionHrefValue` you should be able to give a succinct answer as to why that is the case. Right now it would be hard to explain.

Maybe instead of `missingMentionHrefValue` and `unsupportedMentionHrefValue` we should have `mentionHrefNotValidUrl`, i.e. it fails to parse as a URL (because it is an invalid string, such as empty string) and `mentionHrefDoesNotMatchPattern`, i.e. the URL does not match the pattern.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+describe('RichTextMentionUsers', () => {
+    let element: RichTextMentionUsers;
+    let connect: () => Promise<void>;
+    let disconnect: () => Promise<void>;
+
+    // prettier-ignore
+    async function setup(
+        mappings: BasicUserMentionMapping[],
+        pattern = ''
+    ): Promise<Fixture<RichTextMentionUsers>> {
```
rajsite, line 41: Can you do the same type of pattern I did here for changing to a configuration object: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts#L51-L55

It makes it easier to see the exact configuration change between tests

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+        const mappingConfigs = element.mentionInternals.mentionConfig!.mappingConfigs;
+        expect(mappingConfigs?.size).toEqual(2);
+        expect(Array.from(mappingConfigs.keys())).toEqual(['user:1', 'user:2']);
+        expect(mappingConfigs.get('user:1')?.displayName).toEqual('user');
+        expect(mappingConfigs.get('user:2')?.displayName).toEqual('user');
+        expect(mappingConfigs.get('user:1')?.mentionHref).toEqual('user:1');
+        expect(mappingConfigs.get('user:2')?.mentionHref).toEqual('user:2');
+    });
+
+    it('should update mappingConfigs when mapping elements changed', async () => {
```
rajsite, line 99: we should have tests for when observable properties change. Such as display-name and mention-href updating (both from valid to valid and valid to invalid)

rajsite:
Looking really good, will take a quick look after minor updates made and conflicts resolved.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: Added `mentionHrefDoesNotMatchPattern` and `mentionHrefNotValidUrl` and removed `mentionHrefNotValidUrl`.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.spec.ts
aagash-ni: Updated as per your suggestion

rajsite (approved):

Issue #1516, opened 2023-09-12
Render "nimble-anchor" in rich text editor


vivinkrishna-ni:
## 🧹 Tech Debt

`nimble-rich-text-editor` currently uses a native anchor tag to render the links entered into the editor or by parsed from markdown(`autolink`) string to the editor. The actual plan to render all the links in the rich text editor as `nimble-anchor` as mentioned here: https://github.com/ni/nimble/issues/1372#issuecomment-1658622984.

However, we couldn't render `nimble-anchor` for the following reasons: (copied from https://github.com/ni/nimble/pull/1496#discussion_r1320335836)

1. https://github.com/ni/nimble/issues/1502
2. If you put the cursor in a link, you cannot left/right arrow past the ends of the link. (And if the cursor is outside a link, you cannot use the keyboard to get into the link.).
3. https://github.com/ni/nimble/issues/1508.
4. If you copy and paste part of an existing link's URL, it will create a new link with the href set to the full URL, but just the partial text. I would have expected it to paste as plain text (if it didn't start with "http(s)://")
5. If you click and drag from within a link to highlight, you cannot highlight beyond the ends of the link. Oddly, you can do this with shift+Home/End.

These issues might be due to either the `nimble-anchor` behavior in `contenteditable` div or in the `Tiptap` configurations. Once we have the fix for the above issues, the native anchor tag in the rich text editor should be replaced with `nimble-anchor`.

Code for rendering `nimble-anchor` in `contenteditable` div: https://typescript-l63dbq.stackblitz.io/

vivinkrishna-ni:
We have observed another behavior of the `nimble-anchor` when copying from the viewer and pasting it in the editor while implementing https://github.com/ni/nimble/pull/1559. Whenever the link is copied from rich text viewer by selecting from the beginning of the text to the end of the link and pasted into the editor, it doesn't render as a link. Instead, when pasted into the editor, the clipboard stores only the text content within a `span` tag and so pasted as plain text. I think in this case, only the slotted content in the `<span>` tag is copied when copying only the link.

<details><summary>HTML string when copied <code>nimble-anchor</code> from viewer</summary>
<p>
[...]

PR #1653, opened 2023-11-06
Rich text mention users view component bring up

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

A `nimble-rich-text-mention-users-view` component is to display the mentioned users in the `nimble-rich-text-editor` and `nimble-rich-text-viewer` components.

[UI Mockup for `@mention` in comments feature](https://www.figma.com/file/Q5SU1OwrnD08keon3zObRX/SystemLink-orig?type=design&node-id=7248-114254&mode=design&t=FvNdiZL9EFQPjIZr-0)
 
## 👩‍💻 Implementation

1. Create a base class `MentionView` extended from `FoundationElement`
2. For user mentions, derived from the above base class and create `nimble-rich-text-mention-users-view` component
3. `mentionFont` tokens are added to the nimble-theme-providers
    The mention token follows the naming rules as mentioned in the [doc](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#token-naming)
    1. element - mention
    2. part - N/A
    3. state - disabled
    4. token_type - font, font-color, font-family, font-weight, font-size, font-line-height, fallback-font-family

## 🧪 Testing

1. Written unit tests for the new component addition and tested the rendering text
2. Storybook matrix test

## ✅ Checklist


packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from '@microsoft/fast-element';
+import type { RichTextMentionUsersView } from '.';
+
+export const template = html<RichTextMentionUsersView>`<span
+    class="control"
+    part="control"
+    mention-href="${x => x.mentionHref}"
+    mention-label="${x => x.mentionLabel}"
```
rajsite, line 8: Why are you binding these attributes to the span element in the shadow root? The span element does not have those attributes, what is the goal here?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
```diff
@@ -0,0 +1,11 @@
...
+>
+    <slot></slot>
```
rajsite, line 10: What is the component trying to abstract? I though the attributes would be the only configuration, i.e. the mention-href and the mention-label. And those attributes would be used to populate the internal contents, i.e. show the `@` character and the user name. Why is the slot needed?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from '@microsoft/fast-element';
+import type { RichTextMentionUsersView } from '.';
+
+export const template = html<RichTextMentionUsersView>`<span
```
rajsite, line 4: When doing display inline make sure there is no whitespace between elements in the template. Use the [anchor template](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/anchor/template.ts) as a reference and notice the pattern for opening and closing tags it uses to prevent any whitespace between tags

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,13 @@ export const comments: { readonly [key in TokenName]: string | null } = {
...
-    graphGridlineColor: 'Gridline color for graphs'
+    graphGridlineColor: 'Gridline color for graphs',
+    atMentionFont: 'Font shorthand for at mention views',
+    atMentionFontColor: 'Font color for at mention views',
+    atMentionDisabledFontColor: 'Disabled font color for at mention views',
+    atMentionFontFamily: 'Font family for at mention views',
+    atMentionFontSize: 'Font size for at mention views',
+    atMentionFontWeight: 'Font weight for at mention views',
+    atMentionFontLineHeight: 'Font line height for at mention views',
+    atMentionFallbackFontFamily: 'Fallback font family for at mention views'
```
rajsite, line 304: Make sure to include a comment in the PR describing how this follows the token naming rules: https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#token-naming

This looks like:
element: at-mention
part: n/a
state: n/a, disabled
token: various

I think `at-mention` is too specific, I expect we would style other mentions with a similar font. So I'd advocate for element: mention instead

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+
+interface RichTextMentionUsersViewArgs {
+    mentionHref: string;
+    mentionLabel: string;
+}
+
+const richTextMentionUsersViewDescription = 'The rich text mention users view component is used to render the at(@) mentioned users in the nimble rich text editor and nimble rich text viewer.';
+
+const metadata: Meta<RichTextMentionUsersViewArgs> = {
+    title: 'Incubating/Rich Text View: Mention Users Node',
```
rajsite, line 21: This is not a user facing component right? End users will not construct these directly. This is just an implementation detail. If so move it under Tests/

```suggestion
    title: 'Tests/Rich Text Mention: Users',
```

vivinkrishna-ni: Yes, I have moved it to Tests/ as suggested

packages/nimble-components/src/theme-provider/design-token-comments.ts
vivinkrishna-ni: Updated the name to `mention`

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: Thanks, I have updated the template without whitespace between elements.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: Used the slot approach for a couple of reasons:
1. Tiptap typically renders the mention node in the default slot of the span element through the [`renderLabel`](https://tiptap.dev/api/nodes/mention#render-label) configuration by accessing the node values. So I tried to make our custom component behave the same as the default `span` element does.
2. A workaround for the Firefox bug: The bug is related to the selection of text within the shadow root element. Links to the bugs,
    1. https://bugzilla.mozilla.org/show_bug.cgi?id=1455893 
    2. https://bugzilla.mozilla.org/show_bug.cgi?id=1545367
    3. https://bugzilla.mozilla.org/show_bug.cgi?id=1860292

   All the above bugs are related which states that there is an issue in selecting the texts when the content is within a shadow root. In our case, it affects in below scenarios when opened in the Firefox browser,
    1. In the editor, when we select the mention node, the blue selection background is not shown
    2. In the viewer, not able to select both the @mention view node and other texts in the viewer at the same time

The above issues are resolved when using a slot to render the name from the parent component.

However, this slot approach creates other new issues in the selection of texts in the editor and they are:
1. Selecting the mention view node using the keyboard (Shift + left/right arrow keys) doesn't work as expected
2. If there is no whitespace/any other character at the beginning and at the end of the mention view node in the editor, keyboard selection doesn't work but mouse selection works

1 is resolved when populating the `mention-label` as text content with `@`. But 2 is not resolved in both cases. So both approach has some issues, thought that the Firefox issue was specific to a single browser and the other one was generic, so I have updated the code by populating the attribute value. We are finding it a bit hard to render a shadow root component within a `contenteditable` div across different browsers that creates a different behavioral change like the issues we faced when rendering [`nimble-anchor` within the editor](https://github.com/ni/nimble/issues/1516).

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: I thought that the Tiptap requires these custom attributes to be part of internal control when rendering in the editor and getting and setting the markdown strings but it seems it is not required. This view element itself will act as a mention node for all purposes. So I have removed them.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.stories.ts
m-akinc, line 1: Since this is a test story rather than a documentation story, we should make a few changes:
- use `createMatrixThemeStory` instead of `createUserSelectedThemeStory`, so that we test it in all themes
- rename the file to `rich-text-mention-users-view-matrix.stories.ts`
- simplify/remove most of the content by following the example of [the card's matrix stories](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/card/tests/card-matrix.stories.ts).

packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -772,6 +772,26 @@ export const [
...
+    mentionFontWeight,
+    mentionFontSize,
+    mentionFontLineHeight,
+    mentionFallbackFontFamily
+] = createFontTokens(
+    tokenNames.mentionFont,
+    (element: HTMLElement) => getColorForTheme(element, DigitalGreenDark, PowerGreen, PowerGreen),
+    (element: HTMLElement) => hexToRgbaCssColor(getDefaultFontColorForTheme(element), 0.3),
+    BodyFamily,
+    BodyWeight,
```
m-akinc, line 789: Just want to double check that we want the regular body text weight and not something like `BodyEmphasizedWeight`.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
```diff
@@ -0,0 +1,19 @@
...
+    ${display('inline-block')}
+
+    :host {
+        box-sizing: border-box;
+        font: ${mentionFont};
+    }
+
+    .control {
+        color: ${mentionFontColor};
+    }
```
m-akinc, line 18: Should we also add styling for the disabled state? (`mentionDisabledFontColor`)

m-akinc (commented):
Looks good to me! I'll approve after tests have been added. (I believe that should only be a couple standard tests for the `nimble-rich-text-mention-users-view` element.)

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.stories.ts
vivinkrishna-ni: Updated the storybook as suggested

packages/nimble-components/src/theme-provider/design-tokens.ts
vivinkrishna-ni: Since we have the option to change the text formatting to bold, we thought the text could be in body text weight instead of emphasized text.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
vivinkrishna-ni: Added styling for disabled state and included in the theme matrix test too.

vivinkrishna-ni:
@m-akinc Written test cases for the `nimble-rich-text-mention-users-view` element. Kindly review the latest changes.

m-akinc (approved):

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
rajsite: I'm not clear on a couple things. My primary question is trying to understand what view is rendered while the mention is being entered vs after the entering is complete.

For example while entering I expect the characters the user is entering to be emitted into the slot but after an option is selected maybe there would be a completely custom view.

For example like in GitHub I can type the characters `#94` and select the corresponding option but it gets rendered with a custom view: 
image

The current component being proposed doesn't make it clear to me how the view after entered would be customized.
A thought is maybe something like:

During editing:
```
<rich-text-mention-users-view is-editing></rich-text-mention-users-view>
```
Behavior:
Text is inserted into the slot, all the element does is style the text in the slot.

After mention is commited:
```
<rich-text-mention-users-view mention-href="url" mention-label="John Doe"></rich-text-mention-users-view>
```
Behavior:
Text in the slot is configured as display:none, The configured attributes are used to render the content in the shadow dom template using the `@` icon and formatted text. In the future may be hover buttons, card, etc that display a lot of content or maybe it becomes a link, etc.

vivinkrishna-ni: > the mention is being entered vs after the entering is complete.

This is possible by modifying the [`decorationTag`](https://tiptap.dev/api/utilities/suggestion#decoration-tag) during editing with the `editing` boolean attribute and [`renderHtml`](https://tiptap.dev/guide/custom-extensions/#render-html) for rendering the view node after mention is committed without the `editing` boolean attribute in the Titpap editor.

I have updated the code as suggested by having a `@` icon in the view mode. I just updated the color of the icon by updating the `--ni-nimble-icon-color` to match the text color. Please let me know if that is fine.

rajsite: Or can it be inverted? Looks like render html does let you set attributes, so we could make it `disable-editing`

vivinkrishna-ni: Yes, `renderHtml` allows us to set attributes. I will update the code shortly.

vivinkrishna-ni: I have updated the code with `disable-editing` attribute and verified its working in the editor. Kindly review the same.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
```diff
@@ -0,0 +1,15 @@
+import { html, when } from '@microsoft/fast-element';
+import type { RichTextMentionUsersView } from '.';
+import { iconAtTag } from '../../../../icons/at';
+
+export const template = html<RichTextMentionUsersView>`<template>
```
rajsite, line 5: I'd say simplify the template to just
```
<span class="control"><${iconAtTag}></${iconAtTag}>${x => x.mentionLabel}</span><slot contenteditable="true"></slot>
```

Notes:
- Don't expose the `part` unnecessarily, FAST does it a lot in their templates but fundamentally Nimble is opionated. We don't want to expose unnecessary flexibility which would cause applications to diverge in style.
- At most one instance of the view will have editing enabled, every other instance will have editing disabled. So the template logic overhead is generally extra work / template complexity. Just hide `.control` when editing is disabled with CSS.
- Remove the outer `template` element. It's only needed if you want to bind something to the host. It's not being used for anything so remove it.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+            document.createElement('nimble-rich-text-mention-users-view')
+        ).toBeInstanceOf(RichTextMentionUsersView);
+    });
+
+    it('should set the "control" class on the internal control', async () => {
+        await connect();
+        expect(
+            element.shadowRoot?.firstElementChild?.classList.contains('control')
+        ).toBeTrue();
+    });
```
rajsite, line 47: This test isn't useful. Those classes are private to the shadow root. Nothing external should be aware of them

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    it('should export its tag', () => {
+        expect(richTextMentionUsersViewTag).toBe(
+            'nimble-rich-text-mention-users-view'
+        );
+    });
+
+    it('can construct an element instance', () => {
+        expect(
+            document.createElement('nimble-rich-text-mention-users-view')
```
rajsite, line 38: use the tag constant

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { html } from '@microsoft/fast-element';
+import { RichTextMentionUsersView, richTextMentionUsersViewTag } from '..';
+import { type Fixture, fixture } from '../../../../../utilities/tests/fixture';
+import { waitForUpdatesAsync } from '../../../../../testing/async-helpers';
+
+async function setup(): Promise<Fixture<RichTextMentionUsersView>> {
+    return fixture<RichTextMentionUsersView>(
+        html`<nimble-rich-text-mention-users-view
```
rajsite, line 8: use the tag constant

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+            mention-href="users:1"
+            mention-label="John Doe"
+            disable-editing
+            >@John Doe</nimble-rich-text-mention-users-view
```
rajsite, line 12: use the tag constant

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+            element.shadowRoot?.firstElementChild?.classList.contains('control')
+        ).toBeTrue();
+    });
+
+    it('should set the `part` attribute to "control" on the internal control', async () => {
+        await connect();
+        expect(
+            element.shadowRoot?.firstElementChild?.part.contains('control')
+        ).toBeTrue();
+    });
```
rajsite, line 54: If the part is no longer exported this can be removed

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+        ({ element, connect, disconnect } = await setup());
+    });
+
+    afterEach(async () => {
+        await disconnect();
+    });
+
+    it('should export its tag', () => {
+        expect(richTextMentionUsersViewTag).toBe(
+            'nimble-rich-text-mention-users-view'
```
rajsite, line 32: fyi, all other usages should be the constant. This one stays as is since it is validating thee constant

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    :host {
+        box-sizing: border-box;
+        font: ${mentionFont};
+        white-space: normal;
+        --ni-nimble-icon-color: ${mentionFontColor};
+    }
+
+    .control {
+        color: ${mentionFontColor};
```
rajsite, line 21: Can just move this style under `:host`

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    :host([disabled]) .control {
+        color: ${mentionDisabledFontColor};
+        --ni-nimble-icon-color: ${mentionDisabledFontColor};
+    }
+
+    ${iconAtTag} {
+        position: relative;
+        right: -1.6px;
+        vertical-align: bottom;
```
rajsite, line 32: instead of this positioning can you use `host: {display :inline-flex;}` and then use flax layout for positioning?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+    .control {
+        color: ${mentionFontColor};
+    }
+
+    :host([disabled]) .control {
+        color: ${mentionDisabledFontColor};
+        --ni-nimble-icon-color: ${mentionDisabledFontColor};
+    }
+
+    ${iconAtTag} {
```
rajsite, line 29: While I had the icon in my example of editing vs non-editing views I'm not use if we should actually use the icon for the non-editing view or if we should have the text inserted by the template. i.e. when copy and paste as plaintext should the @ symbol copy as text.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const component = ([
+    disabledName,
+    disabled
+]: DisabledState): ViewTemplate => html`
+    <style class='code-hide'>
+        .mention-container {
+            font: var(${bodyFont.cssCustomProperty});
+            color: var(${bodyFontColor.cssCustomProperty});
+            margin: 4px;
```
rajsite, line 35: smallPadding

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+    </style>
+    <div class="mention-container">
+        <${richTextMentionUsersViewTag} mention-href="user:1" mention-label="John Doe" ?disabled="${() => disabled}" disable-editing>
```
rajsite, line 39: for both of these can you put some text before and after so that one can test out selection across the control.
Maybe something like:
```
[Mention View]@John Doe[Mention View Disabled]
```

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
```diff
@@ -0,0 +1,15 @@
...
+    ${when(
+        x => x.disableEditing,
+        html<RichTextMentionUsersView>`
+    <span
+        class="control"
+        part="control"
+><${iconAtTag}></${iconAtTag}>${x => x.mentionLabel}</span>
+`
+    )}<slot contenteditable="true"></slot
```
rajsite, line 14: Is setting contenteditable required? It doesn't seem to do anything in storybook, i.e. I can't edit the text in the slot.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
vivinkrishna-ni: As per the [mockup](https://www.figma.com/file/JYi8RTEDGARU3YJf22zXkH/SystemLink-LIMS---BTS?node-id=7248%3A114263&mode=dev), we just want the green font only after a selection. So can this be styled as-is?
image

rajsite: I think that intent would be more clearly expressed as:
```
:host {
   color: bodyFontColor or whatever it is
}

.control {
   color: mentionFontColor
}
```

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: Yes, it is required in editing mode as the slot element is within a shadow root, the parent contenteditable is not inherited when `@` is entered into the editor.

rajsite: Can you describe the expected behavior in storybook? On the editing enabled story: https://60e89457a987cf003efc0a5b-qxfzssfcgg.chromatic.com/?path=/story/tests-rich-text-mention-user--rich-text-mention-user-view-edit-enabled-theme-matrix

I can't type anything into the slot / edit the john doe text. The html looks like:
image

If I change the html in devtools to the following:
image

Then I can edit the content in the page. The element becomes editable

vivinkrishna-ni: I have to update the storybook on this test so that we can check how it works in edit mode. I actually meant it is required in the editor which is a contenteditable div.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
vivinkrishna-ni: I believe when copy-pasting as plain text the `@` should be pasted along with the text at least for the user mention. I just quickly tried with other editors like AzDo, and GitHub which were pasted with the @ symbol. So it is better to go with the text instead of an icon for `@mention`. Later, for other types of mentions, we could use icons just like GitHub renders an icon for issues. Will that work?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: Updated the template

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
vivinkrishna-ni: I have removed this test

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
vivinkrishna-ni: Removed

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
vivinkrishna-ni: Since I have removed the icon here, removed all the styles related to icon.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: I added a contenteditable div to the matrix tests around the element to try editing the component.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    :host {
+        box-sizing: border-box;
+        font: ${mentionFont};
+        color: ${bodyEmphasizedFontColor}
+        white-space: normal;
+        --ni-nimble-icon-color: ${mentionFontColor};
+    }
+
+    :host(:not([disable-editing])) .control {
```
rajsite, line 21: Avoid a bunch of `:not` selectors and rely on the cascade: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes

```
.control {
   display: none;
}

:host([disable-editing]) {
   display: whatever it is
}
```

vivinkrishna-ni: Thanks for sharing the doc. I have updated the style as suggested.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: Since we couldn't find a better approach to solve the problem, we will have the same solution which is to have a contenteditable in slot element. It would be really helpful if you could guide us in a way to find a solution for this issue. Otherwise, the only option is to keep the element editable that functions well in all browsers as it is now. Please let me know your thoughts.

rajsite (commented):
Almost done! Just minor tweaks left

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
```diff
@@ -0,0 +1,69 @@
...
+    disabled
+]: DisabledState): ViewTemplate => html`
+    <style class='code-hide'>
+        .mention-container {
+            font: var(${bodyFont.cssCustomProperty});
+            color: var(${bodyFontColor.cssCustomProperty});
+            margin: var(${smallPadding.cssCustomProperty});
+        }
+    </style>
+    <div class="mention-container">
```
jattasNI, line 42: When first looking at the matrix test in storybook, the `[Mention View]` and `[View Disabled]` text looked odd to me. Can you comment on its purpose and whether there's a reason to use the [square bracket] formatting right next to the control?

I agree it is valuable to include text that describes the purpose of each test. But I would vote to format it nicely and either make it plain text with no square brackets (like in the spinner matrix test) or put it in the component itself (many of our tests use the label or content; in this case you could set the description in `mention-label`)

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export default metadata;
+
+const component = ([
+    disabledName,
+    disabled
+]: DisabledState): ViewTemplate => html`
+    <style class='code-hide'>
+        .mention-container {
```
jattasNI, line 36: Let's not set font styles on something that contains the mention element. That makes it ambiguous whether the test is validating the default appearance of the element or the appearance caused by custom storybook styles.

If the goal is to style the descriptive text next to the element you could put that in a `<span>` and target the span. Or you could follow my other suggestion and use the mention-label for the descriptive text and delete this styling.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    it('can construct an element instance', () => {
+        expect(
+            document.createElement(richTextMentionUsersViewTag)
+        ).toBeInstanceOf(RichTextMentionUsersView);
+    });
+
+    it('should set the `mention-label` attribute to the internal control text content', async () => {
+        await connect();
+        expect(element.shadowRoot?.firstElementChild?.textContent).toBe(
```
jattasNI, line 44: Could you move this into a new page object class with just a single method like `getTextContent(): string`?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+            '@John Doe'
+        );
+        element.setAttribute('mention-label', 'Name Change');
+
+        await waitForUpdatesAsync();
+        expect(element.shadowRoot?.firstElementChild?.textContent).toBe(
+            '@Name Change'
+        );
+    });
+});
```
jattasNI, line 54: Could you add some paramaterized tests that ensure the text content is correct for the [Nimble wacky strings](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/utilities/tests/wacky-strings.ts) list?

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
rajsite: Goal is to have some text around the inline text so you can play with selection, etc. Similar to the [anchor story](https://60e89457a987cf003efc0a5b-higxxqmccn.chromatic.com/?path=/docs/components-anchor--docs)

Removing the brackets would be fine, just should have something around and not just inside

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { template } from './template';
+import { styles } from './styles';
+
+/**
+ * A nimble styled rich text mention users view
+ */
+export class RichTextMentionUsersView extends RichTextMentionView {}
+
+const nimbleRichTextMentionUsersView = RichTextMentionUsersView.compose({
+    baseName: 'rich-text-mention-users-view',
```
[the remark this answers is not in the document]
vivinkrishna-ni: Got it! Updated the folder structure as suggested.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
vivinkrishna-ni: I removed the brackets and formatted them as meaningful texts.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view-matrix.stories.ts
vivinkrishna-ni: Yeah, makes sense. Since we don't want the other string to be in the mention-label, I just wrapped it in span, added styles, and removed it from the container class.

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
vivinkrishna-ni: Added tests with wacky strings

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/tests/rich-text-mention-users-view.spec.ts
vivinkrishna-ni: Moved it to a page object file

packages/nimble-components/src/rich-text/components/mention-view/mention-users-view/template.ts
vivinkrishna-ni: The root cause of the issue is not identified yet so we can't strongly say that this issue is related to the Chrome browser or is due to the Tiptap editor handling the decorative node. So I have created a tech debt issue as you mentioned and added it as a comment in the template. Link to the issue: https://github.com/ni/nimble/issues/1659

change/@ni-nimble-components-8eeeed78-306a-426a-a070-ad528135dcf8.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Milan's comment about selection made me try it out in various browsers in Storybook on my machine (macOS 13.5.2). It works well in Safari, Edge, and Chrome but unfortunately Firefox is stopping the selection operation when I drag over the mention view. I can select text or the mention view, but not both.


https://github.com/ni/nimble/assets/10500124/674e7832-f9ae-461d-bd00-b2d442e405db

packages/nimble-components/src/rich-text-mention/base/view/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export class RichTextMentionView extends FoundationElement {
+    /**
+     * Stores the unique URL of the mentioned user matching the pattern
+     *
+     * @public
+     * HTML Attribute: mention-href
+     */
+    @attr({ attribute: 'mention-href' })
+    public mentionHref?: string;
```
jattasNI, line 15: I saw in the AzDO HLD that the proposal is to render the user ID if the user name hasn't loaded yet. Is that accurate? If so, what layer will be responsible for that? I could imagine the view doing it, though it would need access to the ID string but right now it has the full href so it would need to extract the ID.

This should probably be added in a follow-on PR, but it makes me question what the purpose of the `mention-href` attribute is. I don't see it used for anything. If it doesn't have a clear planned purpose then I'd vote to remove it until we implement that feature.

change/@ni-nimble-components-8eeeed78-306a-426a-a070-ad528135dcf8.json
vivinkrishna-ni: We already came across this issue and I commented on it in [the other comment](https://github.com/ni/nimble/pull/1653#discussion_r1384680808) in detail. This is a known issue in Firefox and here is the link to the issue: https://bug1455893.bmoattachments.org/attachment.cgi?id=9289349

The only workaround to the issue is to have all the text content in the slot instead of constructing ourselves in the template. In that way, the Firefox issue resolves, but creates other issues like selecting using the keyboard when the view node is rendered into the editor. Since this issue is specific only to Firefox we thought of having this issue over other generic issues that are on all other browsers.

packages/nimble-components/src/rich-text-mention/base/view/index.ts
vivinkrishna-ni: Yes, the user ID will be in place of the name if the mapping element is not given and also if the link in the markdown matches the pattern correctly. I also mentioned it in the Nimble HLD [here](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text/specs/mention-hld.md#user-mention-view-visible-ui-element:~:text=the%20mentioned%20user-,mention%2Dlabel,-%3A%20string%20%2D%20the%20user) for your reference.

> what layer will be responsible for that?

This will be handled only in the markdown-parser if the given markdown string contains mention node (in autolink format) that matches the pattern. For example, if `<users:1>` markdown string is given as an input and the pattern is also the same as `users:.*` then the `@1` will be shown by updating the `mention-label` of this element in the parser.

Once the mapping element is updated and the `users:1` name is provided the ID will be replaced with the mapped name by getMarkdown + setMarkdown in the editor.

> what the purpose of the `mention-href` attribute is.

It is used only in the markdown serialization where we get the current state of the Tiptap, read the `mention-href` attribute value of this mention view node, and construct the markdown string.

rajsite: Yea this stuff is complex, it's used for the copy-paste workflow. When you copy and paste a chunk of view between rich text editors the rendered mention has the mention-href that goes into the system and is treated as a link node and then should go down the same code path as mentions for links.

Makes me think of a test case of copying between rich comments that have different mention features  / no mention features configured. The mentions copied should just become links?

And I'm only 80% sure that's accurate. I think it would be preferable on copy to turn the mention views to anchors so that stuff never is dealt with externally. But I don't think that was how the prototypes behaved. And I know keyboards can get multiple representations, not sure how tiptap is dealing with that.

change/@ni-nimble-components-8eeeed78-306a-426a-a070-ad528135dcf8.json
rajsite: Oh interesting, didn't realize there was still an issue with Firefox. One thing to try is in normal mode don't have the slot element in the template and then have the slot conditionally when editing-disabled is set (so we can set it display none)

rajsite (approved):

change/@ni-nimble-components-8eeeed78-306a-426a-a070-ad528135dcf8.json
vivinkrishna-ni: Created an issue in Nimble for the issue to track: https://github.com/ni/nimble/issues/1662

packages/nimble-components/src/rich-text-mention/base/view/index.ts
vivinkrishna-ni: > The mentions copied should just become links?

Currently, we don't know how to copy just a link when mention node is copied. In the prototypes, we modified the content only when pasting the node as you said. However, when quickly surfing around the Prosemirror we could find some ways to transform the copied content like [transformCopied](https://prosemirror.net/docs/ref/#view.EditorProps.transformCopied). When we implement it, we will see how can we copy just a link at least in the editor. 

But for the viewer, since it is not using any external libraries like Prosemirror/Tiptap to render the content in view, we might have to look for some native options to restrict copying the node.

jattasNI (approved):

packages/nimble-components/src/rich-text-mention/base/view/index.ts
rajsite: Both of those sound like potentially useful future enhancements assuming they aren't considered requirements for the first release (which seems unlikely giving the time pressure)

vivinkrishna-ni:
I have approved the storybook tests myself, hope you have all reviewed the UI as well.

vivinkrishna-ni:
> @vivinkrishna-ni if all the comments are resolved and you have received the feedback from owners you need you can merge the PR. Not sure if you are waiting on any feedback or for @suseendran-ni's review

We aligned internally within the team earlier. So, I will merge the PR now.

PR #1654, opened 2023-11-07
Rich text editor | Nimble listbox component bring up for mention popup

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

The `nimble-rich-text-mention-list-box` is the component used to populate the list of mention options in the editor. To enable some of the interactions within the popup, it requires FAST's listbox component. As the listbox has not been created in Nimble yet, this PR introduces a new `nimble-listbox` component.

For more details about mention list box component, see [mention-hld](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text/specs/mention-hld.md#mention-popup-visible-ui-element).

## 👩‍💻 Implementation

1. Create a new `nimble-listbox` component from `fast-listbox`
2. Add to the `Tests/` section in the storybook as it is not intended for end users usage and it is for the implementation of `nimble-rich-text-mention-list-box` for mention popup.
3. Create `menuMinWidth` token to share the min-width between `menu` and `listbox`

## 🧪 Testing

1. Basic test cases are written for the new `nimble-listbox` component
2. Storybook matrix test

## ✅ Checklist


packages/nimble-components/src/listbox/tests/listbox.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+        <${listOptionTag} value="3">Option 3</${listOptionTag}>
+        <${listOptionTag} value="4">Option 4</${listOptionTag}>
+        <${listOptionTag} value="5">Option 5</${listOptionTag}>
+    </${listboxTag}>
+    `)
+};
+
+export default metadata;
+
+export const listbox: StoryObj = {};
```
m-akinc, line 33: For test stories, we do them a little differently than documentation stories.

```suggestion
import { ViewTemplate, html } from '@microsoft/fast-element';
import type { Meta, StoryFn} from '@storybook/html';
import {
    createMatrix,
    sharedMatrixParameters
} from '../../utilities/tests/matrix';
import { createMatrixThemeStory } from '../../utilities/tests/storybook';
import { listboxTag } from '..';
import { listOptionTag } from '../../list-option';

const metadata: Meta = {
    title: 'Tests/Listbox',
    parameters: {
        ...sharedMatrixParameters()
    }
};

export default metadata;

const component = (): ViewTemplate => html`
    <${listboxTag}>
        <${listOptionTag} value="1">Option 1</${listOptionTag}>
        <${listOptionTag} value="2">Option 2</${listOptionTag}>
        <${listOptionTag} value="3">Option 3</${listOptionTag}>
        <${listOptionTag} value="4">Option 4</${listOptionTag}>
        <${listOptionTag} value="5">Option 5</${listOptionTag}>
    </${listboxTag}>
    `)
};

export const listboxThemeMatrix: StoryFn = createMatrixThemeStory(
    createMatrix(component)
);
```

packages/nimble-components/src/listbox/tests/listbox.stories.ts
m-akinc, line 1: Rename file to `listbox-matrix.stories.ts`

change/@ni-nimble-components-46cd01a3-0093-40b0-94b4-d5773a0a0582.json
m-akinc, line 1: I'm questioning whether we actually need a `nimble-listbox`, or whether the @-mention feature should use `nimble-menu` instead. I have done some googling and haven't been able to find anything that makes it clear when to use one vs the other. Unless we have a reason not to, I think we should prefer to use what we already have (i.e. a menu).

packages/nimble-components/src/listbox/styles.ts
```diff
@@ -0,0 +1,35 @@
...
+} from '../theme-provider/design-tokens';
+
+export const styles = css`
+    ${display('grid')}
+
+    :host {
+        background: ${applicationBackgroundColor};
+        border: ${borderWidth} solid ${popupBorderColor};
+        margin: 0;
+        min-width: 176px;
```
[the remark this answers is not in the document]
vivinkrishna-ni: Created a `menuMinWidth` token and updated here and in menu styles

change/@ni-nimble-components-46cd01a3-0093-40b0-94b4-d5773a0a0582.json
vivinkrishna-ni: @m-akinc Here are the following reasons for not utilizing the `nimble-menu` and `nimble-menu-item` for the dropdown list and choosing to write our custom component `nimble-rich-text-mention-list-box` in the first place,
1. The typical role of the mention popup is `listbox` which shows the list of users to select from and typing texts to filter the dropdown similar to `combobox`. The `nimble-combobox` internally uses the `nimble-list-option` as a default slot element. Other editors like GitHub and AzDo also use the `listbox` role for their mention popups.
2. The `nimble-menu` uses the `tabindex` value to receive focus on the list of options, which results in losing focus from the editor when the popup is opened. Consequently, keyboard interactions are handled either for the editor or for the mention popup at the same time. On the other hand, if we use `nimble-listbox`, we can handle keyboard navigation using the `selected` boolean attribute. This way, both the editor and the dropdown have focus, resulting in the use of specified key events like arrow keys and the enter key for the dropdown, while other keys like alphabet keys for the editor. 

Initially, when writing the HLD, we prototyped and checked the UI interactions with `nimble-menu` but it didn't seem to work as we expected. The list box provides the necessary APIs to handle different types of interactions specifically for the mention popup so decided to go with the custom mention popup as we aligned in the HLD. Please let us know your thoughts.

m-akinc (approved):

packages/nimble-components/src/listbox/tests/listbox-matrix.stories.ts
```diff
@@ -0,0 +1,32 @@
...
+        ...sharedMatrixParameters()
+    }
+};
+
+export default metadata;
+
+const component = (): ViewTemplate => html`
+    <${listboxTag}>
+        <${listOptionTag} value="1">Option 1</${listOptionTag}>
+        <${listOptionTag} value="2">Option 2</${listOptionTag}>
```
jattasNI, line 23: Does the listbox need to support disabling an item? If so let's disable one of these to ensure the styling is correct.

packages/nimble-components/src/listbox/tests/listbox-matrix.stories.ts
```diff
@@ -0,0 +1,32 @@
...
+    title: 'Tests/Listbox',
+    parameters: {
+        ...sharedMatrixParameters()
+    }
+};
+
+export default metadata;
+
+const component = (): ViewTemplate => html`
+    <${listboxTag}>
```
jattasNI, line 21: We should set the selected item to be one of the options so that the test exercises the styling of a selected item.

packages/nimble-components/src/listbox/tests/listbox-matrix.stories.ts
```diff
@@ -0,0 +1,32 @@
+import { ViewTemplate, html } from '@microsoft/fast-element';
```
jattasNI, line 1: Please add a "Hidden Listbox" story following the pattern of the other component matrix stories

packages/nimble-components/src/listbox/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+    borderWidth,
+    popupBorderColor,
+    elevation2BoxShadow,
+    bodyFont,
+    bodyFontColor,
+    smallPadding,
+    menuMinWidth
+} from '../theme-provider/design-tokens';
+
+export const styles = css`
```
jattasNI, line 15: How should sizing behave for this component? It looks like it currently attempts to grow its height to fit the number of options and grow its width to fit the length of the longest option. If you constrain it with a small `max-height` it renders options outside the bounds of the listbox and with a `max-width` it wraps option text.

I think the height behavior is OK if we state / assume that we expect clients to put the listbox in a scrolllable parent if needed.

The text wrapping seems a bit weirder to me since it's very possible that user names will be longer than what we design for and I don't think fixed height components should wrap, they should ellipsize text. Thoughts on adding styles to ellipsize? 

If we do then we should add matrix test coverage for that case.

packages/nimble-components/src/listbox/tests/listbox-matrix.stories.ts
m-akinc: Sorry, I had excluded the "Hidden Listbox" story from my suggested change, because I didn't think we needed them for "internal" elements.

jattasNI: I can see that argument if our internal usages never intend to hide the component. I lean towards adding it now because I suspect we may someday want to make this a public component or find an internal use case that wants to hide it and it's easier to remember to add the test coverage now.

vivinkrishna-ni: Added the `hidden listbox` test

packages/nimble-components/src/listbox/styles.ts
vivinkrishna-ni: Yeah, we could have ellipsis text especially when there is a long user name on a small mobile screen, instead of hiding the popup we could show the ellipsis text. To achieve this, I have updated the content element in `nimble-list-option` that is given by a client to this component.

jattasNI: Thanks for looking into it. I hadn't considered that it would be a list-option change, which means it impacts other components like select and combobox too. We should probably add test coverage to those components' matrix tests for this case. It also brings up an accessibility question: how does the user access the full text? Do we need to add a tooltip?

Given this complexity I would actually vote to do this in a separate PR or file an issue to track it. This PR can keep the previous behavior of wrapping which is not great but an OK behavior while we're still developing.

jattasNI (approved):

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -70,6 +70,7 @@ export const comments: { readonly [key in TokenName]: string | null } = {
     dialogLargeWidth: 'Standard width for large dialogs.',
     dialogLargeHeight: 'Standard height for large dialogs.',
     dialogLargeMaxHeight: 'Standard maximum height for large dialogs.',
+    menuMinWidth: 'Standard menu min width for menu popup.',
```
rajsite, line 73: minWidth needs to be added to the suffix list, can you do it similar to maxHeight as discussed here: https://github.com/ni/nimble/pull/1553#discussion_r1374822884

packages/nimble-components/src/listbox/styles.ts
vivinkrishna-ni: Created an issue to track the work: https://github.com/ni/nimble/issues/1658

packages/nimble-components/src/theme-provider/design-token-comments.ts
vivinkrishna-ni: Added as mentioned in the linked comment

rajsite (approved):

PR #1655, opened 2023-11-07
Angular 15 upgrade

### LINKED ISSUE #732: Nimble Angular extends the Angular public API
# 🐛 Bug Report

[NimbleSelectControlValueAccessorDirective](https://github.com/ni/nimble/blob/8174d40b98d5c9841076b867f8c32c193bcbd1e3/angular-workspace/projects/ni/nimble-angular/src/directives/select/nimble-select-control-value-accessor.directive.ts#L24) and [NimbleSelectListOptionDirective](https://github.com/ni/nimble/blob/8174d40b98d5c9841076b867f8c32c193bcbd1e3/angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-select-list-option.directive.ts#L13) extend classes from @angular/forms.

Update:
There seem to be many more cases where we extend Angular classes that do not seem to be explicitly marked as extendable/non-final. This is not a comprehensive list:
- `NimbleNumberFieldControlValueAccessorDirective` extends `NumberValueAccessor`, whose [API doc](https://angular.io/api/forms/NumberValueAccessor) does not explicitly say it is not final/can be extended.
- `NimbleRadioControlValueAccessorDirective` extends `RadioControlValueAccessor`, whose [API doc](https://angular.io/api/forms/RadioControlValueAccessor) does not explicitly say it is not final/can be extended.
- `NimbleToggleButtonControlValueAccessorDirective` and `NimbleSwitchControlValueAccessorDirective` extend `CheckboxControlValueAccessor`, whose [API doc](https://angular.io/api/forms/CheckboxControlValueAccessor) does not explicitly say it is not final/can be extended.
- `NimbleTextFieldControlValueAccessorDirective` and `NimbleTextAreaControlValueAccessorDirective` extend `DefaultValueAccessor`, whose [API doc](https://angular.io/api/forms/DefaultValueAccessor) does not explicitly say it is not final/can be extended.

This is [not supported](https://github.com/angular/angular/blob/main/docs/PUBLIC_API.md) unless explicitly documented.
> We explicitly don't consider the following to be our public API surface:
> - constructors of injectable classes (services and directives) - please use DI to obtain instances of these classes
> - extending any of our classes unless the support for this is specifically documented in the API docs
>
> All classes in Angular's public API are final (they should not be extended) unless explicitly stated in the API documentation.
> Extending such final classes is not supported, since protected members and internal implementation may change outside of major releases.

## 💻 Repro or Code Sample

See [NimbleSelectControlValueAccessorDirective](https://github.com/ni/nimble/blob/8174d40b98d5c9841076b867f8c32c193bcbd1e3/angular-workspace/projects/ni/nimble-angular/src/directives/select/nimble-select-control-value-accessor.directive.ts#L24) and [NimbleSelectListOptionDirective](https://github.com/ni/nimble/blob/8174d40b98d5c9841076b867f8c32c193bcbd1e3/angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-select-list-option.directive.ts#L13).

## 🤔 Expected Behavior

N/A

## 😯 Current Behavior

N/A

## 💁 Possible Solution

This is just a warning that extending the public API can lead to breakages and isn't recommended. If they are documented as supporting such, this can be closed. My search was not extensive. An alternative might be to use composition.  Except for the constructor,  NimbleSelectListOptionDirective is probably not a high risk. However, NimbleSelectControlValueAccessorDirective appears to depend on [undocumented API](https://github.com/ni/nimble/blob/8174d40b98d5c9841076b867f8c32c193bcbd1e3/angular-workspace/projects/ni/nimble-angular/src/directives/select/nimble-select-control-value-accessor.directive.ts#L17).

## 🔦 Context

This was investigated as a possible cause to #729. It was not.

## 🌍 Your Environment

* Version: @ni/nimble-angular@8.4.10

### LINKED ISSUE #1017: ngModelChange fires twice when the value for nimble-text-field is updated
# 🐛 Bug Report

Noticed that I was logging twice in my ngModelChange handler when typing new values into a nimble-text-field

## 💻 Repro or Code Sample

https://stackblitz.com/edit/angular-ivy-ayatnv?file=src/app/app.component.ts

```
<nimble-text-field [ngModel]="value" (ngModelChange)="onValueChange($event)"></nimble-text-field>
```

```
public value = '';

onValueChange(value: string) {
    console.log(value)
}
```

## 🤔 Expected Behavior

ngModelChange should call my handler once per character change

## 😯 Current Behavior

ngModelChange calls my handler twice per character change

image
*Behavior as demonstrated by the stackblitz link*

## 💁 Possible Solution

N/A

## 🔦 Context

Trying to update a service value with the text field's new value

## 🌍 Your Environment

OS: Windows, Stackblitz
Browser: Microsoft Edge, FireFox

### LINKED ISSUE #1079: Support Angular 15
## 🧹 Tech Debt

nimble-angular currently supports Angular 14 but the latest version is 15. We need to support the latest.

~This is currently blocked on #810.~

Once that is fixed we have two options:
1. migrate Nimble to only support 15. This is what we have done historically.
2. claim support for both 14 and 15. This should be possible now that both versions are on Ivy and it would make it easier for our clients to upgrade on their own cadence. But there are open questions:
  - would we compile using 14 and claim it's forwards compatible? or compile using 15 and claim it's backwards compatible? are both of these possible?
  - what testing do we need to do with each version? Probably both due to the issue below
  - will this be made more difficult because of us extending internal Angular APIs (#732)

@TrevorKarjanis did some research on these options and believes that 2 is possible but he won't have bandwidth to drive progress for a while.

### LINKED ISSUE #1570: RouterLink support for anchor elements in Angular 15
## 📌 User Story

The [`RouterLink`](https://github.com/angular/angular/blob/296bc993ea3fde9e912381d6ffcae95fb536a166/packages/router/src/directives/router_link.ts#L122) and [`RouterLinkWithHref`](https://github.com/angular/angular/blob/296bc993ea3fde9e912381d6ffcae95fb536a166/packages/router/src/directives/router_link.ts#L309) directives from Angular 14 (and earlier) have been combined into a single [`RouterLink`](https://github.com/angular/angular/blob/ce1b915868e654cdb679e9381db9d3bd3d68d5c4/packages/router/src/directives/router_link.ts#L122) directive in Angular 15. Our Nimble directives for anchor-based elements derived from `RouterLinkWithHref`, but in Angular 15, the equivalent behaviors are provided by `RouterLink`, but [only when it is applied to an `<a>` or `<area>` element.](https://github.com/angular/angular/blob/ce1b915868e654cdb679e9381db9d3bd3d68d5c4/packages/router/src/directives/router_link.ts#L190)

We need to come up with a plan for maintaining `RouterLink`-style support after a switch to Angular 15.

### LINKED ISSUE #1665: Update all forked Angular directives to be based on Angular 15
## 🧹 Tech Debt
As part of the Angular 15 update, we should resolve differences between the Angular 14 directives that were previously copied into nimble-angular and the new upstream versions of those directives for Angular 15.

There is work to be done for the Radio CVA, but we should also verify that all other copied files are based on the Angular 15 version.

mollykreis:
# Pull Request

## 🤨 Rationale

Update to Angular 15.2 for nimble-angular.

Resolves #1079 
Resolves #1017 
Resolves #1570
Resolves #732
Resolves #1665 

## 👩‍💻 Implementation

- Update Angular version to 15.2 using Angular's upgrade utility
- Create forks of some of Angular's CVAs and Angular's RouterLink directive under `angular-workspace\projects\ni\nimble-angular\src\thirdparty\directives\` with modifications to get them to work as expected as base classes for nimble directives
- Add explicit dependency on `source-map-loader` to `nimble-components` to ensure the tests continue to run as they previously did

## 🧪 Testing

- Ran the Angular app and verified it worked as expected
- Existing unit tests pass
- Copied & adapted Angular tests for CVAs and RouterLink directive into `angular-workspace\projects\ni\nimble-angular\src\thirdparty\directives\tests\`
- Added new unit tests for each nimble CVA that verifies that `ngModelChange` is only called once when a control's value changes
- Added new unit tests for each routerLink directive that verifies that the href was sanitized
- Used the built package to verify it works correctly in SystemLinkShared

## ✅ Checklist


mollykreis:
@m-akinc, can you buddy these changes for me?

angular-workspace/projects/ni/nimble-angular/src/thirdparty/utils/coercion.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * [Nimble]
+ * Copied from https://github.com/angular/angular/blob/035aee01089b9f9d4b5b6af66a74002e07723fba/packages/core/src/util/coercion.ts
+ * with the no modifications so that the `coerceToBoolean` function can be used by the forked directive
```
m-akinc, line 4:
```suggestion
 * with no modifications so that the `coerceToBoolean` function can be used by the forked directive
```

package-lock.json
m-akinc, line 1: The diff is pretty extensive and hard to distill what actually changed. I assume we just have to assume it is correct, but I thought I'd call this out to make sure the diff isn't surprising to you.

change/@ni-nimble-components-f628da4e-838f-45bf-9e54-469aa8273f75.json
m-akinc, line 1: I trust there's a good reason, but I'm curious what this does, and why the upgrade caused us to make this change.

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/disableable-router-link-with-href.directive.ts
```diff
@@ -1,14 +1,23 @@
...
  * Base class for Nimble router link directives that go on disableable elements
  */
 @Directive()
-export class DisableableRouterLinkWithHrefDirective<T extends { disabled: boolean }> extends RouterLinkWithHref {
-    public constructor(injector: Injector, private readonly elementRef: ElementRef<T>) {
-        super(injector.get(Router), injector.get(ActivatedRoute), injector.get(LocationStrategy));
+export class DisableableRouterLinkWithHrefDirective<T extends { disabled: boolean }> extends RouterLink {
+    public constructor(
+    @Inject(Router) router: Router,
+        @Inject(ActivatedRoute) route: ActivatedRoute,
```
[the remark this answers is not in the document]
mollykreis: I agree that this indentation is a bit odd, but I didn't think it was worth fighting with the linter on this. And, the linter complains with that line intended. If we think it's worth it, I can disable the indentation lint rule for this line.

angular-workspace/projects/ni/nimble-angular/src/directives/breadcrumb-item/nimble-breadcrumb-item-router-link-with-href.directive.ts
```diff
@@ -17,14 +18,22 @@ import type { BreadcrumbItem } from '@ni/nimble-components/dist/esm/breadcrumb-i
...
     @Input()
     public set nimbleRouterLink(commands: never[] | string | null | undefined) {
         this.routerLink = commands;
     }
 
-    public constructor(injector: Injector, private readonly elementRef: ElementRef<BreadcrumbItem>) {
-        super(injector.get(Router), injector.get(ActivatedRoute), injector.get(LocationStrategy));
+    public constructor(
+    @Inject(Router) router: Router,
+        @Inject(ActivatedRoute) route: ActivatedRoute,
```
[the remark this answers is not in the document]
mollykreis: I agree that this indentation is a bit odd, but I didn't think it was worth fighting with the linter on this. And, the linter complains with that line intended. If we think it's worth it, I can disable the indentation lint rule for this line.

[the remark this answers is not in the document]
m-akinc: Seems like a bug in the linter. I don't think it's worth fighting the linter about it, though.

change/@ni-nimble-components-f628da4e-838f-45bf-9e54-469aa8273f75.json
mollykreis: It's not entirely clear to me why things worked with Angular 14, but we are using the `source-map-loader` `loader` in [karma.conf.js for nimble-components](https://github.com/ni/nimble/blob/47fc4c3463a2a2419d5a2be76e24b46eda90ff2f/packages/nimble-components/karma.conf.js#L112). That requires the `source-map-loader` npm package to be installed. That npm package happens to be a dependency of `@angular-devkit/build-angular`, which has been good enough to make nimble-components happy. However, it isn't anymore, so the correct thing seemed to be to explicitly install it as a dev dependency of nimble-components, since it is needed to run the nimble-components tests.

package-lock.json
mollykreis: I scrolled through it, and nothing in it looked crazy to me. But, I'm not really sure that I'd catch something that was wrong in such a large package-lock diff.

angular-workspace/projects/ni/nimble-angular/src/directives/anchor/tests/nimble-anchor-router-link-with-href.directive.spec.ts
```diff
@@ -98,4 +105,8 @@ describe('Nimble anchor RouterLinkWithHrefDirective', () => {
             expect(routerNavigateByUrlSpy).not.toHaveBeenCalled();
         }));
     });
+
+    it('sanitizes href', () => {
+        expect(sanitizer.sanitize).toHaveBeenCalledWith(SecurityContext.URL, '/page1?param1=true');
+    });
```
rajsite, line 111: Are these tests requiring that other tests in the file have already run?

mollykreis: No, this test doesn't have a dependency on the other tests in the file. The sanitization happens as part of the component being run the and the directive bindings being process, which all happens in the `beforeEach` of the tests.

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/checkbox_value_accessor.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * [Nimble]
+ * Copied from https://github.com/angular/angular/blob/035aee01089b9f9d4b5b6af66a74002e07723fba/packages/forms/src/directives/checkbox_value_accessor.ts
```
rajsite, line 3: Can you use the url that has the tag this was pulled from? Makes it a lot easier to see which version of Angular we last copied changes from.

angular-workspace/package.json
```diff
@@ -21,24 +21,24 @@
     "format": "ng lint --fix"
   },
   "dependencies": {
```
rajsite, line 23: Any thoughts come to mind on how the directive upgrade process should work?

Initial thought:

1. navigate to each file in third party based on tag, i.e.: https://github.com/angular/angular/blob/15.2.0/packages/forms/src/directives/checkbox_value_accessor.ts
2. Take the file contents and do a diff against what is in nimble to merge and changes. Use our [Nimble] comment annotations
3. Update the URL at the top of the file in nimble to know what version of Angular the file was last compared against

angular-workspace/projects/ni/nimble-angular/src/thirdparty/testing/matchers.ts
```diff
@@ -0,0 +1,237 @@
+/**
+ * [Nimble]
+ * Copied from https://github.com/angular/angular/blob/6070c9ddcff88d4ad4bcf73a2dd1874920661d93/packages/platform-browser/testing/src/matchers.ts
+ * with the following modifications:
+ * - Update imports
+ * - Comment out everything other than what is needed to use `toHaveText` matcher
```
rajsite, line 6: Can you mention why? Guessing to avoid polluting the jasmine scope with extra matchers only needed for thirdparty?

angular-workspace/projects/ni/nimble-angular/src/directives/anchor/tests/nimble-anchor-router-link-with-href.directive.spec.ts
rajsite: Maybe the test name could be `sanitized initial bound href`
The active voice of sanitizes makes me expect something to happen in the test

angular-workspace/projects/ni/nimble-angular/src/thirdparty/testing/matchers.ts
m-akinc: I had added this file. More generally, I wanted to minimize impact to our codebase. For example, in the future, if we merge changes to the commented-out code, it will be immediately clear that it does not affect us.

mollykreis: @m-akinc, you did add this file, but I believe I was actually the one that comment out most of the implementation 😄 

There were a lot of things in this file that were using private Angular APIs, and since we didn't need anything other than `toHaveText`, I didn't see the need to find alternatives to the private APIs. I also agree with the rationale that it's easy to tell that commented out code changing doesn't have an impact to our codebase.

mollykreis: I've updated the comment.

angular-workspace/package.json
mollykreis: The process you outlined makes sense to me. It may also be helpful in some cases to diff the two unmodified Angular files against each other to understand what the changes were made upstream that need to be integrated into nimble.

angular-workspace/projects/ni/nimble-angular/src/thirdparty/testing/matchers.ts
m-akinc: > @m-akinc, you did add this file, but I believe I was actually the one that comment out most of the implementation 😄

😳 Whoops! I assumed I had done it, and then gave the rationale I thought I used! 😂

change/@ni-nimble-angular-040b8f50-1a12-4746-b73d-85a5a4b97f3a.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: Note: We won't want to merge this change until we've done the necessary app work to be ready to update the whole platform to Angular 15.

angular-workspace/package.json
jattasNI: Note that we have some docs about the directive upgrade process already [in the README for the `thirdparty` folder](https://github.com/ni/nimble/blob/main/angular-workspace/projects/ni/nimble-angular/src/thirdparty/README.md). At the very least its link needs to be updated but incorporating these instructions would be good too.

One other thought: if Angular ever refactors their tests then it's possible we would no longer need the files that this PR is adding to `utils` or `testing`. It might be good to add a step to either scan the Angular repo for deletions or the Nimble repo for orphaned files.

jattasNI (approved):

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/disableable-router-link-with-href.directive.ts
```diff
@@ -1,14 +1,23 @@
...
+        // eslint-disable-next-line @angular-eslint/no-attribute-decorator
+        @Attribute('tabindex') tabIndexAttribute: string | null | undefined,
```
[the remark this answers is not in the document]
mollykreis: From code inspection, `tabIndexAttribute` won't be used at all in our version of `router_link.ts` because it is only used with `isAnchorElement` is `false`, and we've hardcoded `isAnchorElement` to `true`. If we want to make it really clear that this won't be used, I could pass `undefined` into the `super()` constructor. Do you think that would be preferred?

atmgrifter00 (approved):

angular-workspace/package.json
mollykreis: It looks like the docs were already pretty good, so I updated the link and added an additional step about making sure all the files were still required.

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/disableable-router-link-with-href.directive.ts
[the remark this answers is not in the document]
rajsite: Yea I think I have a small preference for setting it to undefined. Seems like we can remove the eslint disable and  makes it clearer it's not utilized

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/control_value_accessor.ts
```diff
@@ -2,6 +2,7 @@
  * [Nimble]
  * Copied from https://github.com/angular/angular/blob/14.2.6/packages/forms/src/directives/control_value_accessor.ts
```
rajsite, line 3: Was this file updated to 15.2.0? The default_value_accesor file was and is depending on classes from this file (BaseControlValueAccessor) so seems like it may be a good idea to do that now?

rajsite (approved):

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/control_value_accessor.ts
mollykreis: I've confirmed that there aren't any changes that need to be pulled in from the 15.2.0 version, and I updated this path.

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/disableable-router-link-with-href.directive.ts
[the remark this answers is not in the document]
rajsite: Didn't check all the RouterLink implementations but looks like at least NimbleBreadcrumbItemRouterLinkWithHrefDirective needs the same treatment

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/router_link.ts
```diff
@@ -0,0 +1,404 @@
+/**
+ * [Nimble]
+ * Copied from https://github.com/angular/angular/blob/15.2.0/packages/router/src/directives/router_link.ts
+ * with the following modifications:
+ * - Hardcode `isAnchorElement` to `true` so that the directive will correctly set the `href` on elements within nimble that represent anchors
+ * - Make `href` a `@HostBindinding` to avoid using Angular's private sanitization APIs
```
rajsite, line 6:
```suggestion
 * - Make `href` a `@HostBindinding` to avoid using Angular's private sanitization APIs. Implementations leveraging RouterLink should have a test ensuring sanitization is called.
```

rajsite: Seems like we are relying on Angular to do some magic on attributes named href? Is there anything to link to documenting that? If we can't find that can we say something explicitly about relying on that observed behavior?

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-base/disableable-router-link-with-href.directive.ts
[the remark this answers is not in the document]
mollykreis: Ah, good catch. I've now checked for all places we had been creating our own constructor taking in `Attribute('tabindex')` and removed them.

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/radio_control_value_accessor.ts
```diff
@@ -165,6 +174,10 @@ export class RadioControlValueAccessor extends BuiltInControlValueAccessor imple
    */
   @Input() value: any;
 
+  // [Nimble]: Can't override default behavior by injection token, because it is not exported. Inlining value of setDisabledStateDefault.
+  private callSetDisabledState = 'always';
```
m-akinc, line 178: @mollykreis I want to see if you agree this is the right way to handle integrating these changes introduced by the 15.2 version of the file. Seems like the result is that apps won't be able to opt out of a [breaking change in Angular 15](https://angular.io/guide/update-to-version-15#setdisabledstate-is-always-called-when-a-controlvalueaccessor-is-attached) (which, tbh, I did not totally understand). If we really needed to support it, I guess we could define our own token and module config option, [like Angular does](https://github.com/angular/angular/blob/c3e9a2b7cc659ef387d3bba01b6848a9d052a04f/packages/forms/src/form_providers.ts#L46) for the `FormsModule` and `ReactiveFormsModule`. I'm assuming that's not necessary, though.

mollykreis: It certainly looks like the most straightforward way to integrate the changes. I'm not a huge fan of the fact that the `nimble-radio` will have a slightly different behavior than other nimble form controls, but I don't see a way to avoid that.

It [also doesn't look like we have very many usages of the radio button in Angular code](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=nimble-radio&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/RoutineManager/src/app/components/routine-events/routine-events.component.html), so it's probably not a big deal if we don't support the legacy mode.

rajsite: Looks like the `SetDisabledStateOption` is a public api, can we use that type here?
```suggestion
  private callSetDisabledState: SetDisabledStateOption = 'always';
```

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/radio_control_value_accessor.ts
```diff
@@ -204,6 +217,33 @@ export class RadioControlValueAccessor extends BuiltInControlValueAccessor imple
     };
   }
 
+  /** @nodoc */
+  override setDisabledState(isDisabled: boolean): void {
+    /**
```
rajsite, line 222: I didn't try and fully grasp what this is going on about, but is there something we should capture as Angular usage guidance for radio buttons from what is discussed here?

angular-workspace/projects/ni/nimble-angular/src/thirdparty/testing/matchers.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace jasmine {
```
rajsite, line 1: Is this file just created by us? Can you comment at the top if so and anything that needs to be done to update it?

angular-workspace/projects/ni/nimble-angular/src/thirdparty/testing/browser_util.ts
```diff
@@ -0,0 +1,161 @@
+/**
+ * [Nimble]
+ * Copied from https://github.com/angular/angular/blob/6070c9ddcff88d4ad4bcf73a2dd1874920661d93/packages/platform-browser/testing/src/browser_util.ts
```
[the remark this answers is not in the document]
m-akinc: No, I think I mistakenly pulled both this file and matchers.ts from a recent `main` head revision instead of the 15.2 branch. I'll update them accordingly. Since these files both only affect the copied tests, it should not invalidate any previous testing of the actual upgrade (other than those automated tests).

angular-workspace/projects/ni/nimble-angular/src/thirdparty/README.md
```diff
@@ -3,8 +3,9 @@ This folder contains forked third-party source.
 - Each file in this folder should begin with a comment explaining where the file comes from and listing any modifications from the original.
 - All comments made by Nimble developers should be prefixed with the tag "`[Nimble]`".
 
-The `directives` folder corresponds to the [`packages/forms/src/directives` folder](https://github.com/angular/angular/tree/14.2.6/packages/forms/src/directives) of the Angular repo. **When updating the version of Angular** used by the `nimble-angular` package:
+The `directives` folder corresponds to the [`packages/forms/src/directives` folder](https://github.com/angular/angular/tree/15.2.0/packages/forms/src/directives) of the Angular repo. **When updating the version of Angular** used by the `nimble-angular` package:
```
rajsite, line 6: It looks like the latest 15.2 tag is actually 15.2.10: https://github.com/angular/angular/blob/15.2.10/packages/forms/src/directives/number_value_accessor.ts

Thoughts on validating against the latest changes in 15.2? Looking briefly at the changes from 15.2.0 -> 15.2.10 it doesn't look like many of the files we copied are updated but may be worth double checking and updating these references: https://github.com/angular/angular/compare/15.2.0...15.2.10

We also may want to point out using the GitHub compare feature as a hint during upgrades to see potential changes easier.

m-akinc: None of our forked files changed between 15.2.0 and 15.2.10, so I updated all the links at the tops of those files. I also updated the README to suggest using GitHub compare to do updates. I also updated the peer dependency versions in nimble-angular's package file.

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/radio_control_value_accessor.ts
m-akinc: After spending a good chunk of time trying to understand the consequences of our radio CVA not respecting `callSetDisabledState = 'whenDisabledForLegacyCode'`, I've come to the conclusion that it's not even worth documenting. What they're doing in the radio CVA doesn't seem to make any sense. Below is a matrix of the behaviors in different cases:
image


- The pre-v15 (buggy) behavior was that `setDisabledState` wouldn't be called during initialization if the initial value of `disabled` was `false`. 
- The [fixed logic](https://github.com/angular/angular/blob/b76b3c619bc2ff9ffdd436559a7550398a810b10/packages/forms/src/directives/shared.ts#L78) additionally calls `setDisabledState` when `disabled` is `false`. This depends on a `callSetDisabledState` flag, whose value defaults to `'always'`.
- If a client configures `FormsModule` or `ReactiveFormsModule` with `callSetDisabledState = 'whenDisabledForLegacyMode'`, they get the "legacy" behavior, i.e. `setDisabledState` is called for the control only "when disabled."
- For reasons, the radio CVA could not use the new, fixed behavior. It was important for it to continue _NOT_ calling `setDisabledState` if the initial value of `disabled` was `false`. To accomplish this, they overrode `setDisabledState` to [make it a no-op](https://github.com/angular/angular/blob/b76b3c619bc2ff9ffdd436559a7550398a810b10/packages/forms/src/directives/radio_control_value_accessor.ts#L222) on the initial call, if `disabled` was `false`. In other words, the radio CVA uses legacy mode behavior.
- However, if the legacy mode flag (`'whenDisabledForLegacyMode'`) is set, the radio CVA's `setDisabledState` is _not_ a no-op when `disabled` is `false`. Essentially the legacy mode flag causes it to use non-legacy behavior. This makes zero sense, and there is no stated use case for wanting the radio CVA to ever use non-legacy behavior. I can only assume this is a bug (with extremely limited impact).

angular-workspace/projects/ni/nimble-angular/src/thirdparty/README.md
mollykreis: @m-akinc, why did the peer dependency versions need to change? Is there a requirement that clients can only use 15.2.10 or higher of Angular?

angular-workspace/projects/ni/nimble-angular/src/thirdparty/directives/radio_control_value_accessor.ts
rajsite: So configuring `[SetDisabledStateOption](https://angular.io/api/forms/SetDisabledStateOption)` is something users can do, like this example: https://stackoverflow.com/a/75209669

Sounds like we ignore that configuration. Is that not worth capturing in the Usage Docs with a brief explanation (i.e. Nimble Radio Button just supports mode `always`)?

mollykreis: Angular's `NgForm` class handles the `whenDisabledForLegacyCode` option for all CVAs other than Radio CVA. Based on the comment in Angular's Radio CVA, the Radio behaves differently than all other CVAs because the form models the entire group of radio buttons as a single `FormControl`, which means there isn't a way to control the disabled state for individual radios.

Therefore, to answer your questions:
- There already is a test in our forked `value_accessor_integration.spec.ts` for `callSetDisabledState: 'whenDisabledForLegacyCode'`.
- I expect that mode to work just fine for our other CVAs

mollykreis: I've converted the radio group story to an MDX file and added a statement about the CVA behavior.

angular-workspace/projects/ni/nimble-angular/src/thirdparty/README.md
mollykreis: We've decided to stick with 15.2.10.

PR #1656, opened 2023-11-07
Mapping base class make generic and remove text property

rajsite:
# Pull Request

## 🤨 Rationale

We would like to use the mapping base class in more contexts. Conceptually the mapping base class represents just a key and specific implementations represent any additional configuration and what types the key supports.

The mapping base class was modified to support generic key types and remove properties not necessarily needed for all mappings. For example the upcoming user mapping represents a key (href: string) and implementation details of a display-name today and others in the future (username, email, etc.)/

## 👩‍💻 Implementation

- Make the key generic on mapping
- Remove text from the mapping base class
- Update the table enum usages for the types

## 🧪 Testing

Relying on existing tests. Did not add additional tests around generic key types. Existing validation in table column implementations are thorough. Addressed testing gap with spinner validation.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f7f47d9d-f5d5-4ad0-b573-a1ab40586bfd.json
+++ b/change/@ni-nimble-components-f7f47d9d-f5d5-4ad0-b573-a1ab40586bfd.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update internal mapping base class to be generic",
+  "packageName": "@ni/nimble-components",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
+++ b/packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
@@ -46,7 +46,9 @@ export class TableColumnEnumTextValidator extends TableColumnEnumBaseValidator<
     }
 
     private validateMappingTypes(mappings: Mapping<unknown>[]): void {
-        const valid = mappings.every(TableColumnEnumTextValidator.isSupportedMappingElement);
+        const valid = mappings.every(
+            TableColumnEnumTextValidator.isSupportedMappingElement
+        );
         this.setConditionValue('unsupportedMappingType', !valid);
     }
 }
--- a/packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
+++ b/packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
@@ -1,11 +1,16 @@
+import type { Mapping } from '../../../mapping/base';
 import { MappingText } from '../../../mapping/text';
 import type { ColumnInternals } from '../../base/models/column-internals';
 import {
     TableColumnEnumBaseValidator,
     enumBaseValidityFlagNames
 } from '../../enum-base/models/table-column-enum-base-validator';
+import type { MappingKeyType } from '../../enum-base/types';
 
-const enumTextValidityFlagNames = [...enumBaseValidityFlagNames] as const;
+const enumTextValidityFlagNames = [
+    ...enumBaseValidityFlagNames,
+    'missingTextValue'
+] as const;
 
 /**
  * Validator for TableColumnEnumText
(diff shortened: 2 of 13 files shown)


packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
```diff
@@ -23,22 +24,35 @@ export class TableColumnIconValidator extends TableColumnEnumBaseValidator<
...
         const invalid = mappings
-            .filter(isMappingIcon)
+            .filter(TableColumnIconValidator.isSupportedMappingElement)
             .some(mappingIcon => mappingIcon.resolvedIcon === undefined);
         this.setConditionValue('invalidIconName', invalid);
     }
+
+    private validateNoMissingText(mappings: Mapping<unknown>[]): void {
+        const invalid = mappings
+            .filter(TableColumnIconValidator.isSupportedMappingElement)
```
rajsite, line 54: This is a bug, icon columns support IconMapping and SpinnerMapping

Doesn't look like spinner mappings have any tests unfortunately. This was about to introduce a bug that spinner mapping text are no longer validated. Adding some tests.

jattasNI (approved):
Thanks for doing this, and knocking it out so quickly!

change/@ni-nimble-components-f7f47d9d-f5d5-4ad0-b573-a1ab40586bfd.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Angular doesn't have a shared base class for mappings and probably should
Blazor does have a shared base class and should be updated to move Text out

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
```diff
@@ -23,22 +24,43 @@ export class TableColumnIconValidator extends TableColumnEnumBaseValidator<
...
+        ] as const);
+    }
+
+    private static isIconMappingElement(
+        mapping: Mapping<unknown>
+    ): mapping is MappingIcon {
+        return mapping instanceof MappingIcon;
+    }
+
+    private static isSupportedMappingElement(
```
m-akinc, line 36: Can we make this non-`static` so it can use `supportedMappingElements` from the base? We would also have to change that from `private` to `protected`. Should have single source of truth.

packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
```diff
@@ -16,4 +21,25 @@ export class TableColumnEnumTextValidator extends TableColumnEnumBaseValidator<
     public constructor(columnInternals: ColumnInternals<unknown>) {
         super(columnInternals, enumTextValidityFlagNames, [MappingText]);
     }
+
+    private static isSupportedMappingElement(
```
m-akinc, line 25: Same as other comment about single-sourcing supported mappings.

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
m-akinc: What is the bug? And what tests are missing for spinner mappings? They have just the usual element construction and exported tag tests.

rajsite: There is no test validating spinners have text

rajsite: And the bug is in my change, which misses validating that spinners have text because it only looks at icons.

rajsite: Before it was pushed. It was a note for me to address. I pushed the change that now checks spinners but need tests

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
rajsite: Like the sentiment of single sourcing the knowledge of supported elements. I tried to make the typing stronger for giving the base class a list of inferred supported Mappings but that was tricky. I instead went the other way and just made each specific implementation responsible for managing their supported mappings. Think it looks pretty good and isn't a concerning amount of duplication. I'll resolve but reset you in-case there are concerns.

packages/nimble-components/src/table-column/enum-text/models/table-column-enum-text-validator.ts
rajsite: Resolved with alternate pattern discussed above

change/@ni-nimble-components-f7f47d9d-f5d5-4ad0-b573-a1ab40586bfd.json
rajsite: Created a tech debt task for this: https://github.com/ni/nimble/issues/1666

packages/nimble-components/src/table-column/icon/models/table-column-icon-validator.ts
rajsite: Added tests

m-akinc (approved):

Issue #1137, opened 2023-03-27
Table keyboard interaction


jattasNI:
## 😯 Problem to Solve

We want users to be able to perform common table interactions via keyboard as well as via mouse. 

## 💁 Proposed Solution

- [Updated HLD](https://github.com/ni/nimble/blob/856f21dfc0630cc2267bd278b03a627c7e684dce/packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md)
- PR: #2172 

Earlier resources: #877 ; [interaction spec - currently has password error](https://xd.adobe.com/view/fa09e396-dbb9-40b8-547f-1cf9eab35a0b-8c38) (UXDesign1).

## 📋 Tasks
- [x] Demo prototype visuals to Brandon / get design feedback
- [x] Header/row/cell navigation via Up/Down/Left/RightArrow keys
- [x] Navigation within current row / header row with Tab/Shift-Tab
- [x] Return to navigation mode with Esc after action menu / cell contents are focused
- [x] Ensure keyboard navigation behaves sanely with mouse / other table interactions
- [x] Finalize impact to table / cell view APIs, if any
- [x] Get/address feedback on implementation via draft PR
- [x] Additional keys (Home/End/PgUp/PgDn/CtrlEnter/F2/etc)
- [x] Additional work identified in HLD

atmgrifter00:
As part of the work in [this PR](https://github.com/ni/nimble/pull/1679), we are hiding the expand-collapse button for both group rows and parent rows from the accessibility tree. For parent rows this is meant to be a temporary state until we provide keyboard interactions for the table, at which point the expand-collapse button for parent rows should be focusable (but _not_ for group rows).

It's also worth noting that by hiding the expand-collapse button the accessibility score actually drops, seemingly because it expects the wrapped `button` element to have its `tabindex` set to `-1`, the same as the FAST component, but `tabindex` is [not passed through](https://github.com/microsoft/fast/blob/fd65254531ac3a593b32528d81b44ebf82c8e4ab/packages/web-components/fast-foundation/src/button/button.template.ts#L13) in the template (while `aria-hidden` is).

m-akinc:
[HLD is in PR](https://github.com/ni/nimble/pull/1506), but needs some attention to keep moving forward.

PR #1663, opened 2023-11-14
Table data hierarchy

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #861 

## 👩‍💻 Implementation

Here are the notable aspects of the implementation:

#### Creating the hierarchical data
- Adding a new `parentIdFieldName` attribute to the `Table` that determines what the hierarchical structure will be
- Adding a new library (`performant-array-to-tree`) that handles the conversion of the flat table data into a hierarchical structure that Tanstack needs.
- Updating the typing of the table data we use internally to support hierarchy
- Handle variable order of setting data, `idFieldName` and `parentIdFieldName`, which can require the rebuilding of the data into the proper hierarchical form.
- Creation of a `DataHierarchyManager` class to encapsulate the processing of the data into its hierarchical form.

#### Visualizing the hierarchical data
- Updating the `Row` component to show an expand/collapse button.
- Creation of a `ExpansionManager` class that provides APIs to manage the expanded state of all collapsible rows in the table.

## 🧪 Testing

- Added suite of unit tests that ensure the correct number of rows are rendered when moving between various states of the `Table` (when setting various combinations of `idFieldName` and `parentIdFieldName`)
- Added validation tests
- Added unit tests that verify hierarchy works as expected when used in combination with one or more of grouping/selection/sorting
- Added matrix tests that test a variety of scenarios that result in varying indentations.

## ✅ Checklist


packages/nimble-components/src/table/types.ts
```diff
@@ -123,6 +136,12 @@ export interface TableRowSelectionEventDetail {
     selectedRecordIds: string[];
 }
 
+export interface TableRowExpandToggleEventDetail {
+    oldState: boolean;
+    newState: boolean;
+    recordId?: string;
```
mollykreis, line 142: `recordId` doesn't need to be optional, does it? Shouldn't any expandable row have a defined record id?

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -64,6 +88,35 @@ export class TableCell<
...
+
+    public onRowExpandToggle(event: Event): void {
+        const expandEventDetail: TableRowExpandToggleEventDetail = {
+            oldState: this.expanded,
+            newState: !this.expanded,
+            recordId: this.recordId
+        };
+        this.$emit('row-expand-toggle', expandEventDetail);
+        event.stopImmediatePropagation();
+        // To avoid a visual glitch with improper expand/collapse icons performing an
```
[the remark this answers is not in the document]
atmgrifter00: I had the same reservations as you on this, but at the moment I haven't come up with a better way to share this implementation. I'll try to re-think this.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -32,6 +33,18 @@ export class TableCell<
...
     public recordId?: string;
 
+    @observable
+    public isTopLevelRow = false;
+
+    @observable
+    public isParentRow = false;
+
+    @observable
+    public isFirstCell = false;
```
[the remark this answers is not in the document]
atmgrifter00: We use `isParentRow` in isolation in the template code for the `Row`. Perhaps there's still a way to consolidate the two concepts into one, but I figured I would just point that out.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -32,6 +33,18 @@ export class TableCell<
     @observable
     public recordId?: string;
 
+    @observable
+    public isTopLevelRow = false;
```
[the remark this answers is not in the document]
atmgrifter00: I'll admit I think my thinking at the time of writing it was likely faulty, which was that if it wasn't an API I wouldn't be leveraging directly in CSS, but simply for template code, then observables were sufficient. However, who's to say that we would never want to create CSS around these APIs? So they should just be attributes. I'll change them.

packages/nimble-components/src/table/components/cell/template.ts
```diff
@@ -6,11 +6,33 @@ import {
...
+import { mediumPadding } from '../../../theme-provider/design-tokens';
 
 // prettier-ignore
 export const template = html<TableCell>`
-    <template role="cell" style="--ni-private-table-cell-nesting-level: ${x => x.nestingLevel}">
+    <template role="cell" style="
+        --ni-private-table-cell-nesting-level: ${x => x.nestingLevel};
+        --ni-private-table-cell-view-padding: ${x => (x.isParentRow && x.isFirstCell && !x.isTopLevelRow ? mediumPadding.getValueFor(x) : '0px')};
+    ">
+        ${when(x => x.isParentRow && x.isFirstCell && (x.nestingLevel !== 0 || !x.isTopLevelRow), html<TableCell>`
```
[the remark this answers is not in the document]
atmgrifter00: A top level row is a data row only, but always at the top. So, a group-row can have a nesting level of 0, but it is _not_ a top-level row. Additionally, a row that isn't a top-level row could have a `nestingLevel` of 0. That said, this logic still doesn't need to deal with the `nestingLevel` state, as `isTopLevel` is enough (in addition to the other existing checks) to determine if we need to show the expand/collapse button.

packages/nimble-components/src/table/index.ts
```diff
@@ -953,22 +1105,24 @@ export class Table<
...
+    ): boolean => {
+        const isGroupRow = row.getIsGrouped();
+        if (!isGroupRow && !row.subRows.length) {
             return false;
         }
 
         const expandedState = this.table.options.state.expanded;
-
-        if (expandedState === true) {
+        if (expandedState === true && isGroupRow) {
```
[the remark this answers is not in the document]
atmgrifter00: I was having parent rows _not_ default to expanded for now. Do you feel they should default to expanded?

packages/nimble-components/src/table/index.ts
```diff
@@ -1085,6 +1253,22 @@ export class Table<
 
         return tanstackSelectionState;
     }
+
+    private updateExpandedStateToDefault(): void {
+        const expandedState: { [key: string]: boolean } = {};
```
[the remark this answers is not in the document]
atmgrifter00: Perhaps this can be called something better. The intent of this method is to collapse the expanded state to just that of the current available groupRows. So, there's nothing to remove, as all we're doing is adding to a default empty state.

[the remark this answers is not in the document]
mollykreis: I think there may have been confusion about my question and where the line of code I chose to add it on. My question isn't about whether this function needed to be doing more. My question is about whether we needed to be doing more processing of the expanded state when data is updated to make sure the expanded state isn't trying to claim that removed rows are expanded.

packages/nimble-components/src/table/index.ts
```diff
@@ -469,9 +452,8 @@ export class Table<
     }
 
     /** @internal */
-    public handleGroupRowExpanded(rowIndex: number, event: Event): void {
-        this.toggleGroupExpanded(rowIndex);
-        event.stopPropagation();
```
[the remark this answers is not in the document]
atmgrifter00: Yep. Fixed.

packages/nimble-components/src/table/index.ts
```diff
@@ -603,6 +625,109 @@ export class Table<
...
+                if (isSelected) {
+                    selectedRecordIds.push(recordId);
+                }
+            }
+        );
+        return selectedRecordIds;
+    }
+
+    private trackFlatDataOrder(data: readonly TData[]): void {
+        if (this.idFieldName) {
```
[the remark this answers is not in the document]
atmgrifter00: Now happening in the `idFieldNameChanged` method.

packages/nimble-components/src/table/index.ts
```diff
@@ -754,9 +886,19 @@ export class Table<
...
-            updatedOptions.data = [...this.table.options.data];
+        if (
+            this.tableUpdateTracker.requiresTanStackDataReset
+            && !this.tableUpdateTracker.updateRowParentIds
+        ) {
+            if (!this.parentIdFieldName) {
+                // Perform a shallow copy of the data to trigger tanstack to regenerate the row models and columns.
+                updatedOptions.data = [...this.table.options.data];
+            } else {
+                const flatList = this.convertHierarchicalDataToFlatList();
```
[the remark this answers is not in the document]
atmgrifter00: So we need to reprocess the data in the situation where:
1. `parentId` is set
2. `id` is not set
3. table is given data (this results in a flat data set in Tanstack, since there was no `id` set)
4. `id` is then set

The final step results in `requiresTanStackDataReset` being true but `updateRowParentIds` being false, and we still need to reprocess the data to convert it to its hierarchical form. It seems like there should be clearer boolean logic to handle all of the needed cases, but I haven't come up with it yet.

packages/nimble-components/src/table/types.ts
atmgrifter00: This has been updated as part of [this PR](https://github.com/ni/nimble/pull/1679).

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -60,7 +66,8 @@ export class TableValidator<TData extends TableRecord> {
 
     public validateRecordIds(
         data: TData[],
-        idFieldName: string | undefined
+        idFieldName: string | undefined,
+        parentIdFieldName?: string
```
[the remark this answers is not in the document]
atmgrifter00: This was actually a misguided change to begin with. This has been effectively reverted.

packages/nimble-components/src/table/index.ts
```diff
@@ -995,16 +1149,26 @@ export class Table<
...
+        }
         // must update the collapsedRows before toggling expanded state
         if (wasExpanded) {
             this.collapsedRows.add(row.id);
         } else {
             this.collapsedRows.delete(row.id);
         }
         row.toggleExpanded();
+        if (row instanceof TableRow) {
+            this.$emit('rowExpanded');
```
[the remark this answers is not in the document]
atmgrifter00: Correct. This has been removed. We are now simply letting the event emitted from the `Row` component bubble up as expected.

packages/nimble-components/src/table/index.ts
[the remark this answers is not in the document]
atmgrifter00: This code has been reworked to have parent rows default to being expanded.

packages/nimble-components/src/table/components/cell/template.ts
```diff
@@ -6,11 +6,33 @@ import {
...
+            <${buttonTag}
+                    appearance="${ButtonAppearance.ghost}"
+                    content-hidden
+                    class="expand-collapse-button"
+                    tabindex="-1"
+                    @click="${(x, c) => x.onRowExpandToggle(c.event)}"
+                >
+                    <${iconArrowExpanderRightTag} ${ref('expandIcon')} slot="start" class="expander-icon ${x => x.animationClass}"></${iconArrowExpanderRightTag}>
+                    ${x => (x.expanded ? tableGroupCollapseLabel.getValueFor(x) : tableGroupExpandLabel.getValueFor(x))}
```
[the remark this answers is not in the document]
atmgrifter00: This has been addressed as part of [this PR](https://github.com/ni/nimble/pull/1679).

packages/nimble-components/src/table/components/cell/template.ts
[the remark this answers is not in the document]
atmgrifter00: The `isTopLevelRow` API has been removed (not needed).

packages/nimble-components/src/table/components/cell/index.ts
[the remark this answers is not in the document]
atmgrifter00: All of the new APIs are now denoted as `@attr`. See [this PR](https://github.com/ni/nimble/pull/1679).

packages/nimble-components/src/table/components/cell/index.ts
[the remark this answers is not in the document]
atmgrifter00: These APIs nave been [removed completely](https://github.com/ni/nimble/pull/1679).

packages/nimble-components/src/table/components/cell/template.ts
```diff
@@ -6,11 +6,33 @@ import {
...
+import { buttonTag } from '../../../button';
+import { iconArrowExpanderRightTag } from '../../../icons/arrow-expander-right';
+import { mediumPadding } from '../../../theme-provider/design-tokens';
 
 // prettier-ignore
 export const template = html<TableCell>`
-    <template role="cell" style="--ni-private-table-cell-nesting-level: ${x => x.nestingLevel}">
+    <template role="cell" style="
+        --ni-private-table-cell-nesting-level: ${x => x.nestingLevel};
+        --ni-private-table-cell-view-padding: ${x => (x.isParentRow && x.isFirstCell && !x.isTopLevelRow ? mediumPadding.getValueFor(x) : '0px')};
```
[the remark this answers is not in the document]
atmgrifter00: [This has been removed](https://github.com/ni/nimble/pull/1679).

packages/nimble-components/src/table/components/cell/index.ts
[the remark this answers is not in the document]
atmgrifter00: For now, I have created a shared style, and thankfully, the `TableCell` no longer is responsible for showing an expand-collapse button, so it is now only a duplicate implementation between `Row` and `GroupRow`, but that are at least now sharing the same style.

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -45,21 +47,30 @@ export class TableValidator<TData extends TableNode> {
...
 
+        if (
+            !this.idFieldNameNotConfigured
+            && parentIdFieldName !== undefined
+            && idFieldName === undefined
+        ) {
+            this.idFieldNameNotConfigured = true;
+        }
+
         return !this.idFieldNameNotConfigured;
```
mollykreis, line 69: What would you think about rewriting this function a bit to make it more clear what's going on? Maybe something like this:

```ts
const idFieldNameRequired = selectionMode !== TableRowSelectionMode.none || typeof parentIdFieldName !== 'string';
if (idFieldNameRequired) {
    this.idFieldNameNotConfigured = typeof idFieldName !== 'string';
} else {
    this.idFieldNameNotConfigured = false;
}
return !this.idFieldNameNotConfigured;
```

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -72,27 +83,26 @@ export class TableValidator<TData extends TableNode> {
...
-        for (const record of data) {
-            if (
-                !Object.prototype.hasOwnProperty.call(
-                    record.clientRecord,
-                    idFieldName
-                )
-            ) {
-                this.missingRecordId = true;
-                continue;
+        if (typeof idFieldName === 'string') {
```
mollykreis, line 86: I don't think this outer if-statement is providing any value. I should probably be removed.

packages/nimble-components/src/table/models/selection-managers/selection-manager-base.ts
```diff
@@ -99,10 +101,7 @@ export abstract class SelectionManagerBase<TData extends TableRecord> {
             return [];
         }
 
-        return row
-            .getLeafRows()
-            .filter(leafRow => leafRow.getLeafRows().length === 0)
-            .map(leafRow => leafRow.id);
+        return row.getLeafRows().map(leafRow => leafRow.id);
```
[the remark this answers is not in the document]
atmgrifter00: You are right, and your suggested change fixes the issue (that I replicated in Storybook). I was attempting to write a test for this, but I believe we're missing some pageObject mechanisms to make this a straightforward test (and perhaps the right data setup), so this is still missing.

packages/nimble-components/src/table/index.ts
```diff
@@ -760,13 +803,29 @@ export class Table<
             );
         }
         if (this.tableUpdateTracker.requiresTanStackDataReset) {
```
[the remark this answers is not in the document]
atmgrifter00: I updated this to also reset for `updateRowIds` but not `updateRowParentIds`, because I don't think that technically conflict with the current expanded state. I could definitely be convinced otherwise.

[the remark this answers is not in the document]
mollykreis: If I'm thinking about this correctly, then I think it would be good to reset the expanded state when parent row IDs changes since: (1) I think that will be necessary when we have lazy-loaded hierarchy, (2) I'm not sure if makes sense from a user's perspective to keep some rows expanded when they are hierarchically ordered in a different way, and (3) the set of rows that `isRowExpandable` may change, which might cause issues with the expansion manager (we'd have to think through if there really are problems here).

packages/nimble-components/src/table/models/selection-managers/selection-manager-base.ts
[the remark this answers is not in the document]
mollykreis: I've added a test for this in `table-selection.spec.ts`. I verified it failed without your fix and passes with your fix.

packages/nimble-components/src/table/models/table-validator.ts
mollykreis: I went ahead and removed it.

packages/nimble-components/src/table/index.ts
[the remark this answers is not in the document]
atmgrifter00: This logic has been updated and now is managed by an `ExpansionManager` class.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -201,12 +201,14 @@ export class TableRow<
     }
 
     public onRowExpandToggle(event: Event): void {
-        const expandEventDetail: TableRowExpandToggleEventDetail = {
-            oldState: this.expanded,
-            newState: !this.expanded,
-            recordId: this.recordId!
-        };
-        this.$emit('row-expand-toggle', expandEventDetail);
+        if (this.recordId) {
```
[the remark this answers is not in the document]
atmgrifter00: Logic has simply changed to:
```
if (typeof this.recordId === 'string') { ... }
```
and, yes, this was just to make TS happy where we fill out the fields of the event detail object.

packages/nimble-components/src/table/tests/table.stories.ts
```diff
@@ -104,11 +205,20 @@ The attribute is invalid in the following conditions:
 -   A record was found that did not have a field with the name specified by \`id-field-name\`. This will cause \`validity.missingRecordId\` to be \`true\`.
 -   A record was found where \`id-field-name\` did not refer to a value of type \`string\`. This will cause \`validity.invalidRecordId\` to be \`true\`.`;
 
+const parentIdFieldNameDescription = `An optional string attribute that specifies the field name within a row's record to use as a row's parent ID, which,
+when used in combination with the \`idFieldName\` attribute, will display the table data in a hierarchical fashion. If the attribute is not specified, the
+data in the table will always be presented without hierarchy.
+
+The attribute is invalid in the following conditions:
+-   When this attribute is set, but \`id-field-name\` is unset. This will cause \`validity.idFieldNameNotConfigured\` to be \`true\`.
+-   When there are circular references between records discovered based on field values of \`parend-id-field-name\` for one record and \`id-field-name\` of another. This will cause \`validity.invalidParentIdConfiguration\` to be \`true\`.
```
mollykreis, line 214:
```suggestion
-   When there are circular references between records discovered based on field values of \`parent-id-field-name\` for one record and \`id-field-name\` of another. This will cause \`validity.invalidParentIdConfiguration\` to be \`true\`.
```

packages/nimble-components/src/table/index.ts
```diff
@@ -853,28 +915,41 @@ export class Table<
...
         const rows = this.table.getRowModel().rows;
         this.tableData = rows.map(row => {
-            const isGrouped = row.getIsGrouped();
+            const isGroupRow = row.getIsGrouped();
+            const hasParentRow = isGroupRow ? false : row.getParentRow();
+            // we check row.original.subRows below because row.subRows is populated for group rows
+            // which we don't want to include
+            const isParent = row.original.subRows !== undefined
+                && row.original.subRows.length > 0;
+            const isGroupRowChildWithNoHierarchy = !isGroupRow
```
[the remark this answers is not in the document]
atmgrifter00: Changed name to `isChildOfGroupRowWithNoHierarchy` for clarity.

change/@ni-nimble-components-af9b983d-43db-46f7-a204-c271899c2559.json
```diff
@@ -1,5 +1,5 @@
 {
-  "type": "minor",
+  "type": "major",
   "comment": "Introducing data hierarchy to table",
```
mollykreis, line 3: It might be helpful to be more specific about what the breaking change was in each of these change files. Something like _renamed `groupsCollapseAll` on the table label provider to `collapseAll`_. That will make it easier for someone scanning the changelog to see if they need to worry about a given breaking change.

packages/nimble-components/src/table/tests/table-matrix.stories.ts
```diff
@@ -37,46 +37,44 @@ const data = [
...
+        id: '',
         firstName: 'Marge',
         lastName: 'Simpson',
         age: 35,
         quote: "Oh, I've Always Wanted To Use Rosemary In Something!",
-        id: '',
         parentId: '3'
     },
     {
+        id: '7',
```
mollykreis, line 73: nit: Is there a reason to jump from an id of 4 to an id of 7 here? I don't think it's necessarily a problem, but if there isn't a reason to, this fake data will probably be more maintainable if we use consecutive id values.

mollykreis (approved):

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -201,12 +201,14 @@ export class TableRow<
...
     public onRowExpandToggle(event: Event): void {
-        const expandEventDetail: TableRowExpandToggleEventDetail = {
-            oldState: this.expanded,
-            newState: !this.expanded,
-            recordId: this.recordId!
-        };
-        this.$emit('row-expand-toggle', expandEventDetail);
+        if (typeof this.recordId === 'string') {
+            const expandEventDetail: TableRowExpansionToggleEventDetail = {
+                oldState: this.expanded,
```
jattasNI, line 206: In a very quick scan, I didn't see anything setting `this.expanded = !this.expanded`. Can you make sure something is doing that and that we have test coverage to verify the event data for both expand and collapse?

packages/nimble-components/src/table/index.ts
```diff
@@ -439,11 +417,7 @@ export class Table<
 
     /** @internal */
     public handleCollapseAllGroupRows(): void {
```
jattasNI, line 419: I didn't trace where this is used, but can you double check whether it should be renamed not to include "group" in its name as you did for other APIs?

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -864,6 +870,465 @@ describe('Table', () => {
...
+                await waitForUpdatesAsync();
+                expect(pageObject.getRenderedRowCount()).toBe(5);
+
+                pageObject.clickDataRowExpandCollapseButton(0); // collapses parent
+                await waitForUpdatesAsync();
+
+                expect(pageObject.getRenderedRowCount()).toBe(3);
+            });
+
+            it('maintains expanded state between data updates', async () => {
```
jattasNI, line 1051: What is the behavior if a record is in the data and expanded, then a data update deletes it, then another data update restores it? Is its original expanded state maintained? If "yes" then we should probably have a test for it. If "no" then we probably don't need one for that case, but we should make sure that we have a test asserting that new data is always rendered in the same default expanded state (presumably collapsed).

jattasNI (commented):
About halfway through. Looking good so far!

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -864,6 +870,465 @@ describe('Table', () => {
                 verifyRenderedData(sortedData);
             });
         });
+
+        describe('hierarchical data', () => {
```
jattasNI, line 874: I scanned the table docs for features that might be impacted by hierarchy. The only one I could think of that I didn't see tested was column order. If I change which column is first, it should affect which column shows the expand/collapse UI. I verified that this works in your Storybook build; should we have a unit test for it? Hopefully there are existing column re-order tests you could leverage from.

packages/nimble-components/src/table/tests/table-selection.spec.ts
```diff
@@ -2363,4 +2464,284 @@ describe('Table row selection', () => {
...
+
+            it('can interactively select child row', async () => {
+                const rowIndex = getUnsortedExpandedIndexOfRecord('2.0');
+                await pageObject.clickRow(rowIndex);
+                const currentSelection = await element.getSelectedRecordIds();
+                expect(currentSelection).toEqual(['2.0']);
+            });
+        });
+
+        describe('with multi-selection', () => {
```
jattasNI, line 2536: Could you add/update tests to validate the "select all rows" button:
 - should show indeterminate state if all top level rows are selected but some leafs aren't
 - should select/deselect all rows in the hierarchy when clicked

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -864,6 +870,465 @@ describe('Table', () => {
...
+                    stringData2: 'Parent 1 Child 2',
+                    numericData: 0,
+                    moreStringData: 'foo',
+                    id: '6',
+                    parentId: '0',
+                    parentId2: 'Parent 1',
+                    parentId3: 'Parent 2'
+                }
+            ];
+            it('shows collapse all button with hierarchical data', async () => {
```
jattasNI, line 938: Could you make sure we have tests that verify the behavior of clicking the collapse all button with hierarchical data?

packages/nimble-components/src/table/tests/table.stories.ts
```diff
@@ -104,11 +205,20 @@ The attribute is invalid in the following conditions:
...
+const parentIdFieldNameDescription = `An optional string attribute that specifies the field name within a row's record to use as a row's parent ID, which,
+when used in combination with the \`idFieldName\` attribute, will display the table data in a hierarchical fashion. If the attribute is not specified, the
+data in the table will always be presented without hierarchy.
+
+The attribute is invalid in the following conditions:
+-   When this attribute is set, but \`id-field-name\` is unset. This will cause \`validity.idFieldNameNotConfigured\` to be \`true\`.
+-   When there are circular references between records discovered based on field values of \`parent-id-field-name\` for one record and \`id-field-name\` of another. This will cause \`validity.invalidParentIdConfiguration\` to be \`true\`.
+-   When an id specified by \`parent-id-field-name\` is not discovered in any record. This will cause \`validity.invalidParentIdConfiguration\` to be \`true\`.`;
+
 const validityDescription = `Readonly object of boolean values that represents the validity states that the table's configuration can be in.
```
jattasNI, line 217: Don't forget to add `invalidParentIdConfiguration` to the list of validity flags too.

packages/nimble-components/src/table/models/data-hierarchy-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+        parentIdFieldName: string | undefined
+    ): TableNode<TData>[] {
+        let hierarchicalData: TableNode<TData>[];
+        if (
+            typeof idFieldName === 'string'
+            && typeof parentIdFieldName === 'string'
+        ) {
+            try {
+                // The call to arrayToTree will perform a deep copy of the data, but it does allow a
+                // configuration that will do shallow copies, and thus it's signature doesn't support
```
jattasNI, line 28:
```suggestion
                // configuration that will do shallow copies, and thus its signature doesn't support
```

jattasNI (approved):

change/@ni-nimble-components-af9b983d-43db-46f7-a204-c271899c2559.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: What's the expected situation for keyboard interactions to expand/collapse rows in this PR? If it's not planned to be implemented yet, could you make sure the work is tracked in #1137?

jattasNI: I'd completely missed it because I only looked at the Tasks list in the description. I added another item to that list that points to your comment so now I'm satisfied (feel free to edit further if you wish).

packages/nimble-components/src/table/tests/table.spec.ts
atmgrifter00: So, in this particular test new data is being added to the existing set of data, and we ultimately assert two separate things:
1. the new data added to the existing data defaults to expanded (which is the expected default expanded state...not collapsed)
2. that the previous row that was expanded (after first collapsing everything), stays expanded after adding the new records.

To answer your other question, yes, if a record is removed and then re-added it will be rendered with the default expanded state (which is expanded). I think there actually is value in writing a test for that.

packages/nimble-components/src/table/tests/table.spec.ts
atmgrifter00: So, column order does _not_ affect the expand/collapse UI as it is now being completely handled by the `Row` component (despite my original implementation).

packages/nimble-components/src/table/index.ts
```diff
@@ -1006,15 +1071,11 @@ export class Table<
...
-            this.collapsedRows.add(row.id);
-        } else {
-            this.collapsedRows.delete(row.id);
-        }
+    private toggleRowExpanded(rowIndex: number): void {
+        const rows = this.table.getRowModel().rows;
+        const row = rows[rowIndex]!;
+        // Must update the expansionManager before toggling expanded state
+        this.expansionManager.toggleRowExpansion(row);
         row.toggleExpanded();
```
rajsite, line 1079: now we have two parallel systems tracking the expansion state of rows :/
Is that necessary? Can we just rely on the tanstack state? is there a path towards having just tanstack represent that state? (feature request? bug?) may be worth a discussion

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -201,12 +201,14 @@ export class TableRow<
     }
 
     public onRowExpandToggle(event: Event): void {
-        const expandEventDetail: TableRowExpandToggleEventDetail = {
-            oldState: this.expanded,
-            newState: !this.expanded,
-            recordId: this.recordId!
-        };
-        this.$emit('row-expand-toggle', expandEventDetail);
+        if (typeof this.recordId === 'string') {
```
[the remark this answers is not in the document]
atmgrifter00: So, to be clear, there was no row expansion prior to this PR from a user's perspective. That said, this seems like an ultimately unnecessary check. The only way this event is triggered is 1) the row is considered a parent row, which means the data was already validated, which then means that 2) the type of `recordId` has already been confirmed to be of type `string`. Will revert.

packages/nimble-components/src/table/models/data-hierarchy-manager.ts
```diff
@@ -0,0 +1,92 @@
+import { arrayToTree } from '../../utilities/thirdparty/performant-array-to-tree/arrayToTree';
+import type { TableNode, TableRecord } from '../types';
+import type { TableValidator } from './table-validator';
+
+/**
+ * Manages data hierarchy within the table, including converting between a flat list of
+ * data and hierarchical data.
+ */
+export class DataHierarchyManager<TData extends TableRecord> {
+    private isDataFlat = false;
```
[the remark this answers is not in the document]
atmgrifter00: All of the data returned from the `getTableNodes` function has the `indexField` added on to it, so I don't think there is a trivial means of looking at the data and "knowing" that is flat or not. 

I'm not sure I'm following the suggestion, as this state is only of interest to this particular manager class (used in the `getAllRecords` implementation). If we returned this state, then it would become private state held on to the table and then passed into the `getAllRecords` method, which seems unnecessarily roundabout.

change/@ni-nimble-components-af9b983d-43db-46f7-a204-c271899c2559.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "major",
```
rajsite, line 2: Chatted with meyers, some high level things to look into:
- Can DataHierarchyManager be the owner of the data (i.e. the constructor receives the data, calculates isDataFlat). May mean the validator needs to update to understand DataHierarchyManager instead of data references.
- Similar to selection manager, can the expansion manager own configuring expansion state on TanStackTable (right now the top-level table owns updating both separately). Can we document why we need to persist collapseRows state outside of tanstack today and is there a path to tanstack completely owning that state (we submit a bug / PR, doesn't block this PR but we work towards it)
- Can we use generics on arrayToTree to more strongly type the config, i.e. each record must have fields for id and parentid that are keyof the generic

packages/nimble-components/src/table/models/expansion-manager.ts
```diff
@@ -0,0 +1,76 @@
+import type { Row as TanStackRow } from '@tanstack/table-core';
+import type { TableNode, TableRecord } from '../types';
+
+/**
+ * Manages the expanded/collapsed state of group rows and parent rows in the table. We must track
```
mollykreis, line 5: Feel free to take/reject any number that you think make sense.

_Manages the expanded/collapsed state of rows in the table._

_We must track the expansion state separately from TanStack because:_
_1. TanStack does not support having a different initial expansion state per row unless explicitly specified for each row by ID. This causes problems in the nimble-table because we need to have a different initial expansion state for group rows, parent rows, and parent rows with lazy loaded children._
_2. TanStack does not remove entries from its expanded state when those rows are no longer present in the data. This is not ideal because the object maintaining the expansion state can grow unbounded._

packages/nimble-components/src/table/template.ts
```diff
@@ -113,13 +113,13 @@ export const template = html<Table>`
...
                             ${repeat(x => x.virtualizer.visibleItems, html<VirtualItem, Table>`
-                                ${when((x, c) => (c.parent as Table).tableData[x.index]?.isGrouped, html<VirtualItem, Table>`
+                                ${when((x, c) => (c.parent as Table).tableData[x.index]?.isGroupRow, html<VirtualItem, Table>`
                                     <${tableGroupRowTag}
                                         class="group-row"
                                         :groupRowValue="${(x, c) => c.parent.tableData[x.index]?.groupRowValue}"
                                         ?expanded="${(x, c) => c.parent.tableData[x.index]?.isExpanded}"
                                         :nestingLevel="${(x, c) => c.parent.tableData[x.index]?.nestingLevel}"
-                                        :leafItemCount="${(x, c) => c.parent.tableData[x.index]?.leafItemCount}"
+                                        :immediateChildCount="${(x, c) => c.parent.tableData[x.index]?.immediateChildCount}"
```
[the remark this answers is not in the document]
atmgrifter00: Removing the optional chaining leads to test failures, that at the moment are unexplainable. Since this is the current working pattern, I am opting to leave it as-is and then follow-up with an investigation on why we are seeing errors when using the non-null assertion.

change/@ni-nimble-components-af9b983d-43db-46f7-a204-c271899c2559.json
atmgrifter00: So here are the changes (or non-changes) that I have made for each of the mentioned issues/concerns and some accompanying reasoning:
- Went ahead and refactored the `DataHierarchyManager` as discussed. Nothing in the validator needed to be changed.
- Added documentation to top of `ExpansionManager` class to outline the reasons we need to maintain separate state from TanStack. Added a reference to `TanStackTable` on the `ExpansionManager` which allows it call appropriate functions directly like `toggleAllRowsExpanded`.
    - I'm not sure there is a good path to request the kinds of behaviors from TanStack that we are implementing for expanded state. For now, I think our auxiliary state is fairly easy to reason about and hooking it into the TanStack pipeline is fairly trivial.
- Made `arrayToTree` generic, strongly-typed, and less configurable (allowing us to remove a swath of tests we didn't care about).

packages/nimble-components/src/table/models/data-hierarchy-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    public constructor(
+        private readonly tableValidator: TableValidator<TData>
```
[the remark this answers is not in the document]
atmgrifter00: Removed this coupling. See [this comment thread](https://github.com/ni/nimble/pull/1663#discussion_r1456538472).

packages/nimble-components/src/table/models/data-hierarchy-manager.ts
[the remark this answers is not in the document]
atmgrifter00: See [this comment thread](https://github.com/ni/nimble/pull/1663#discussion_r1456538472).

packages/nimble-components/src/table/index.ts
atmgrifter00: See [this comment thread](https://github.com/ni/nimble/pull/1663#discussion_r1456538472).

packages/nimble-components/src/table/models/expansion-manager.ts
```diff
@@ -0,0 +1,68 @@
+import type { Row as TanStackRow } from '@tanstack/table-core';
+import type { TableNode, TableRecord } from '../types';
+
+/**
+ * Manages the expanded/collapsed state of group rows and parent rows in the table.
+ */
+export class ExpansionManager<TData extends TableRecord> {
+    private isInDefaultState = true;
```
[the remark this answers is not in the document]
atmgrifter00: Added comment.

packages/nimble-components/src/table/models/expansion-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export class ExpansionManager<TData extends TableRecord> {
+    // This field represents whether or not the expanded state of **all** rows is in the default expanded
+    // state or not. Note that the default expanded state for a particular row type (group vs parent) can
+    // potentially be different (e.g. expanded for groups and collapsed for parent rows).
+    private isInDefaultState = true;
+    private collapsedRows = new Set<string>();
+
+    public constructor(
+        private readonly table: TanStackTable<TableNode<TData>>
```
rajsite, line 26: nit, can we name this reference `tanStackTable`, i.e. the same name used in selection manager and makes it easier to disambiguate from the nimble table

packages/nimble-components/src/utilities/models/array-to-tree.ts
```diff
@@ -0,0 +1,121 @@
+// Modified from https://github.com/philipstanislaus/performant-array-to-tree/blob/v1.11.0/src/arrayToTree.ts
+// Copyright (c) 2022 philipstanislaus
+// SPDX-License-Identifier: MIT
+
+export interface TreeItem<Item> {
+    clientRecord: Item | undefined;
+    subRows: TreeItem<Item>[];
+    originalIndex: number | undefined;
+}
```
rajsite, line 9: Why not just return a TreeNode as defined in table/types? Can then avoid the `as TreeNode` cast in the table code

packages/nimble-components/src/utilities/models/array-to-tree.ts
```diff
@@ -0,0 +1,121 @@
+// Modified from https://github.com/philipstanislaus/performant-array-to-tree/blob/v1.11.0/src/arrayToTree.ts
```
rajsite, line 1: Since this file is super integrated now and table specific we can just move it to table/models

packages/nimble-components/src/utilities/models/array-to-tree.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    // stores all item ids that have not been added to the resulting unflattened tree yet
+    // this is an opt-in property, since it has a slight runtime overhead
+    const orphanIds: null | Set<string | number> = new Set();
+
+    // idea of this loop:
+    // whenever an item has a parent, but the parent is not yet in the lookup object, we store a preliminary parent
+    // in the lookup object and fill it with the data of the parent later
+    // if an item has no parentId, add it as a root element to rootItems
+    // [Nimble] Convert to a for-loop to have access to the item's index in the flat list
```
rajsite, line 39: Don't really need the `[Nimble]` marker here or below in other comments, it's all  nimble now :)

packages/nimble-components/src/utilities/models/array-to-tree.ts
```diff
@@ -0,0 +1,121 @@
...
+
+            // add the current item to the parent
+            lookup[parentId]!.subRows.push(treeItem);
+        }
+    }
+
+    if (orphanIds?.size) {
+        const orphans = Array.from(orphanIds.values()).join(',');
+        throw new Error(
+            `The items array contains orphans that point to the following parentIds: [${orphans}]. These parentIds do not exist in the items array. Hint: prevent orphans to result in an error by passing the following option: { throwIfOrphans: false }`
```
rajsite, line 98: Can you remove the "hint" part of the message, doesn't seem to apply anymore

packages/nimble-components/src/table/index.ts
```diff
@@ -609,6 +629,23 @@ export class Table<
         this.tableUpdateTracker.trackColumnInstancesChanged();
     }
 
+    private getCurrentSelectedRecordIds(): string[] {
```
rajsite, line 632: nit: maybe this logic could be moved to selection manager base as a public api method?
selection manager has the tanstack table reference which seems to be able to lookup the same thing, i.e. tanstackTable.options.state.rowSelection

rajsite (commented):
Implementation looking good, just some minor tweaks, still need to give a pass of the tests but looking solid!

packages/nimble-components/src/utilities/models/array-to-tree.ts
atmgrifter00: This might require a redesign of the algorithm which requires setting a couple of these fields to `undefined` in a temporary state, which `TableNode` at the moment does not allow (and I think would be awkward if it did). For now I would prefer to leave as-is, unless you have an suggestion I'm not seeing to resolve this.

rajsite: You can have a type used in the file as, `type TreeNodePartial = Partial<TreeNode>` and then at return do `return rootItems as TreeNode`.

rajsite: Mainly have that connection to the TreeNode type of what fields are defined and encapsulate the knowledge that arrayToTree satisfies the TreeNode contract in the method (instead of the callers of the method having to assert that)

packages/nimble-components/.prettierignore
```diff
@@ -27,5 +27,6 @@ CHANGELOG.*
 src/utilities/tests/tests/fixture.spec.ts
 src/utilities/tests/fixture.ts
 src/utilities/style/direction.ts
+src/utilities/thirdparty/*
```
rajsite, line 30: not needed anymore

packages/nimble-components/src/.eslintrc.js
```diff
@@ -68,7 +68,9 @@ module.exports = {
         // Enabled to prevent accidental usage of async-await
         '@typescript-eslint/require-await': 'error'
     },
-    ignorePatterns: ['.eslintrc.js'],
+    ignorePatterns: [
+        '.eslintrc.js'
+    ],
```
rajsite, line 73: is this change needed?

rajsite (approved):

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -134,6 +139,7 @@ describe('Table', () => {
 
         beforeEach(async () => {
             ({ element, connect, disconnect } = await setup());
+            jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
```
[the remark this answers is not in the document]
atmgrifter00: Uhhhh...pretty sure that snuck in as part of some debugging code. Removed.

msmithNI (approved):

PR #1664, opened 2023-11-14
Support for @mention in Rich Text Viewer

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

[User Story 2536828](https://dev.azure.com/ni/DevCentral/_workitems/edit/2536828): FE | Add @ mention support in Nimble rich-text-viewer and in markdown parser

## 👩‍💻 Implementation

1. Support for the `nimble-rich-text-mention-users` configuration element in the viewer component
2. Handle dynamic loading of child elements and their property value changes in the viewer
3. Update `markdown-parser` to add support for mention node
    1. Update `toDOM(node)` method in autolink schema to render as mention view element if it matches the pattern and display name is available
    2. New `markdown-parser-mention-configuration` model class to generate an object that is required for the parser during the initialization
4. Render `nimble-rich-text-mention-users-view` as a mention node in viewer
5. Remove the restriction of the HTTP/HTTPS scheme for absolute links in the markdown; render the link without `href` as aligned in the HLD PR: https://github.com/ni/nimble/pull/1594
6. Added usage docs for user mentions in `nimble-rich-text-viewer`

## 🧪 Testing

1. Added tests for `markdown-parser` to check the input markdown string containing mention node and render a custom view element for user mention
2. Added tests to check the dynamic loading of child elements into the viewer component

## ✅ Checklist


packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -3,6 +3,7 @@ import { observable } from '@microsoft/fast-element';
 export interface MentionInternalsOptions {
```
vivinkrishna-ni, line 3: Change file will be added once the predecessor PRs are merged.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -81,4 +89,209 @@ describe('RichTextViewer', () => {
...
+
+            const mappingUser = document.createElement(
+                mappingUserTag
+            ) as MappingUser;
+            mappingUser.key = 'user:1';
+            mappingUser.displayName = 'username1';
+            userMention.appendChild(mappingUser);
+
+            element.appendChild(userMention);
+            await waitForUpdatesAsync();
```
m-akinc, line 158: To reduce duplication, I would pull this out into a helper function that you call from each of these tests:
```TS
async function appendUserMentionConfiguration(element: RichTextViewer, userKey: string | undefined, displayName: string | undefined) {
   const userMention = document.createElement(
        richTextMentionUsersTag
    ) as RichTextMentionUsers;
    userMention.pattern = '^user:(.*)';

    if (userKey || displayName) {        
        const mappingUser = document.createElement(
            mappingUserTag
        ) as MappingUser;
        mappingUser.key = userKey ?? '';
        mappingUser.displayName = displayName ?? '';
        userMention.appendChild(mappingUser);
    }
    element.appendChild(userMention);
    await waitForUpdatesAsync();
}

File: packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts [hunk 3]
```diff
@@ -534,70 +553,56 @@ describe('Markdown parser', () => {
...
+                            expect(
+                                lastChildElementHasAttribute('href', doc)
+                            ).toBeFalse();
                         }
                     );
-                }
+                });
             });
 
             describe('various unsafe characters in an absolute link', () => {
```
m-akinc, line 576:
```suggestion
            describe('malformed or unsupported absolute links', () => {
```

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -534,70 +553,56 @@ describe('Markdown parser', () => {
...
-                            expect(getLeafContentsFromElement(doc)).toEqual([
-                                value.name
+                            expect(getTagsFromElement(doc)).toEqual([
+                                'P',
+                                'NIMBLE-ANCHOR'
                             ]);
+                            expect(
+                                lastChildElementHasAttribute('href', doc)
+                            ).toBeFalse();
                         }
```
m-akinc, line 571: Would be good to also validate the value of `getLeafContentsFromElement(doc)`.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -923,4 +928,317 @@ describe('Markdown parser', () => {
...
+
+            expect(getTagsFromElement(doc)).toEqual([
+                'P',
+                `${richTextMentionUsersViewTag}`.toUpperCase()
+            ]);
+            expect(getLastChildElementAttribute('mention-label', doc)).toEqual(
+                'username1'
+            );
+        });
+
```
m-akinc, line 1068:
```suggestion
```
Not sure this is an interesting test case.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -923,4 +928,317 @@ describe('Markdown parser', () => {
...
+            expect(getTagsFromElement(doc)).toEqual([
+                'P',
+                `${richTextMentionUsersViewTag}`.toUpperCase()
+            ]);
+            expect(getLastChildElementAttribute('mention-label', doc)).toEqual(
+                '1234-5678'
+            );
+        });
+
```
m-akinc, line 1089:
```suggestion
```
This doesn't seem materially different from the earlier test case where the user id does not match any of the mapping elements.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -923,4 +928,317 @@ describe('Markdown parser', () => {
...
+            expect(lastChildElementHasAttribute('href', doc)).toBeFalse();
+        });
+
+        it('should get anchor element when pattern is not a valid Regex', async () => {
+            ({ element, connect, disconnect } = await setup(
+                [
+                    { key: 'user:1', displayName: 'username1' },
+                    { key: 'user:2', displayName: 'username1' }
+                ],
+                'abc'
```
m-akinc, line 1200: 'abc' is a valid regex. Really this test case is covering the case where the absolute link is not http/https, and it doesn't match the regex. Maybe just change the test name.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -923,4 +928,317 @@ describe('Markdown parser', () => {
...
+            );
+
+            expect(getTagsFromElement(doc)).toEqual([
+                'P',
+                `${anchorTag}`.toUpperCase()
+            ]);
+            expect(lastChildElementHasAttribute('href', doc)).toBeFalse();
+        });
+
+        it('should get anchor element when key is not a valid URL', async () => {
```
m-akinc, line 1219:
```suggestion
        it('should get anchor element when keys do not match the pattern', async () => {
```

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -65,16 +132,101 @@ export class RichTextMarkdownParser {
...
+        // that renders as an alternative to the display name if missing.
+        return regexpArray?.[1] ?? '';
+    }
+
+    // Validates the mention Href in the markdown string with the input configuration values
+    private static validateMentionPattern(
+        mention: MentionInternals<RichTextMentionConfig>,
+        mentionHref: string
+    ): boolean {
+        const regexPattern = new RegExp(mention.pattern ?? '');
```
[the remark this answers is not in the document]
vivinkrishna-ni: No, it should return false when `mention.pattern` is undefined. It can't be null as its type is `string | undefined`

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -53,7 +62,65 @@ export class RichTextMarkdownParser {
...
+            markdown => {
+                // `mention` rule is added before `autolink` rule to get the highest preference when getting the DOM output
+                // MarkdownIt before: https://markdown-it.github.io/markdown-it/#Ruler.before
+                markdown.inline.ruler.before(
+                    'autolink',
+                    'mention',
+                    (state, _silent) => {
+                        const mentionHref = this.getMentionHref(state);
+
+                        if (mentionHref && mentionsMap?.size) {
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated as per the suggestion

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -4,27 +4,34 @@ import {
...
      * This function takes a markdown string, parses it using the ProseMirror MarkdownParser, serializes the parsed content into a
      * DOM structure using a DOMSerializer, and returns the serialized result.
      * If the markdown parser returns null, it will clear the viewer component by creating an empty document fragment.
      */
     public static parseMarkdownToDOM(
-        value: string
+        value: string,
+        mentionsMap?: Map<string, MentionInternals<RichTextMentionConfig>>
     ): HTMLElement | DocumentFragment {
+        this.markdownParser = this.initializeMarkdownParser(mentionsMap);
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, I have used it as a local variable

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -65,16 +132,101 @@ export class RichTextMarkdownParser {
...
+                    mentionLabel: { default: '' },
+                    disableEditing: { default: 'true' }
+                },
+                group: 'inline',
+                inline: true,
+                content: 'inline*',
+                toDOM(node) {
+                    const { mentionHref, mentionLabel, disableEditing } = node.attrs;
+                    return [
+                        RichTextMarkdownParser.viewElement,
```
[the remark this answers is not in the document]
vivinkrishna-ni: Even `getCustomSchemaConfiguration()` is defined during initialization, the `toDOM()` callback function will be called only if the above `mention` tokenizer rule (that sets the `viewElement`) returns true. So the `viewElement` is always defined here.

Also, we have to set the `viewElement` only in the `mention` tokenizer rule as the markdown input can contain different types of mentions (in the future). For example, if the markdown string is like `<user:1> <issue:1>`, assuming we support the issue mention too, for each node, the view element might be different which can be defined only within the tokenizer rule (within `initializeMarkdownParser()`).

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -44,10 +67,80 @@ export class RichTextViewer extends FoundationElement {
...
+            this.updateMentionsMap();
+        }
+    }
+
+    private childItemsChanged(): void {
+        void this.updateMentionsFromChildItems();
+    }
+
+    private async updateMentionsFromChildItems(): Promise<void> {
+        const definedElements = this.childItems.map(async item => (item.matches(':not(:defined)')
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the heads up. I will keep this comment open to not miss updating if your PR goes in first.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts
```diff
@@ -32,6 +34,13 @@ const metadata: Meta<RichTextViewerArgs> = {
     <${richTextViewerTag}
         :markdown="${x => x.markdown}"
     >
+        <${richTextMentionUsersTag} pattern="^user:(.*)">
```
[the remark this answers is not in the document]
vivinkrishna-ni: While we currently use this pattern exclusively for user mentions, it holds potential for future scenarios. For instance, as Milan pointed out in [this comment](https://github.com/ni/nimble/pull/1594#discussion_r1364736273), links directing to an issue are presently absolute links but will be converted into issue mention view elements once support is added. 

Also, the user mentions lack of a specific use case for having a pattern, but looking ahead, if a client wishes to implement a user link that, upon clicking, navigates to the user's profile, we may require a pattern allowing the client to determine the link in configuration element.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -44,10 +67,80 @@ export class RichTextViewer extends FoundationElement {
...
+        this.mentionNotifiers = [];
+    }
+
+    private updateMentionsMap(): void {
+        this.mentions.forEach(mention => {
+            if (
+                mention.mentionInternals.pattern
+                && mention.mentionInternals.mentionConfig
+            ) {
+                this.mentionsMap.set(
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the catch. I have re-initialized the map every time it needs an update.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: I moved this to a helper method and used whenever needed as suggested.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: Added

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: Removed

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: Removed this test case.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: Updated the test name

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -65,16 +132,101 @@ export class RichTextMarkdownParser {
...
+                        {
+                            'mention-href': mentionHref as string,
+                            'mention-label': mentionLabel as string,
+                            'disable-editing': disableEditing as string
```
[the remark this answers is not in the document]
m-akinc: I think I understand the gist of your response, but my concern was that `disable-editing` is a boolean attribute, and we seem to be assigning a string value. I've looked at the documentation for this and I couldn't determine whether we can instead set a boolean value, i.e.:
```TS
attrs: {
    mentionHref: { default: '' },
    mentionLabel: { default: '' },
    disableEditing: { default: true },
    viewElement: { default: '' }
},
...
toDOM(node) {
    ...
    return [
        viewElement,
        {
              'mention-href': mentionHref as string,
              'mention-label': mentionLabel as string,
              'disable-editing': disableEditing as boolean
         }
```
Or if we always want the view element to have `disable-editing` set here, it would be clearer to do:
 ```TS
    return [
        viewElement,
        {
              'mention-href': mentionHref as string,
              'mention-label': mentionLabel as string,
              'disable-editing': 'true'
         }
```
I'm just concerned that the way it is currently written, it looks like if `disableEditing` was the string `"false"`, we would enable editing on that view element, but that is not true.

File: packages/nimble-components/src/rich-text/models/markdown-parser.ts (hunk 10 above)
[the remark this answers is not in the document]
Comment (m-akinc) [reply packages/nimble-components/src/rich-text/models/markdown-parser.ts]: nit: Sorry, can you also invert the `if (currentMention)` below (line 92)?

File: packages/nimble-components/src/rich-text/models/markdown-parser.ts (hunk 12 above)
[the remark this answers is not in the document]
Comment (m-akinc) [reply packages/nimble-components/src/rich-text/models/markdown-parser.ts]: Thanks for the explanation. I do like setting the element tag in the token rather than storing it on the parser.

File: packages/nimble-components/src/rich-text/models/markdown-parser.ts (hunk 10 above)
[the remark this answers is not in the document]
Comment (vivinkrishna-ni) [reply packages/nimble-components/src/rich-text/models/markdown-parser.ts]: Sorry, I missed it. Now updated to reduce the nesting.

File: packages/nimble-components/src/rich-text/models/markdown-parser.ts (hunk 16 above)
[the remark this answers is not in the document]
Comment (vivinkrishna-ni) [reply packages/nimble-components/src/rich-text/models/markdown-parser.ts]: Sorry, I misunderstood. As you suggested, the second approach is clearer as it will always be in view mode, so I have implemented the same.

File: packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (m-akinc) [reply packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts:36]: > links directing to an issue are presently absolute links but will be converted into issue mention view elements once support is added.

Sure, but we would not re-use the `nimble-rich-text-mention-users` element for issues, so I'm still unsure why we would need to configure the pattern on the user mentions element.

> we may require a pattern allowing the client to determine the link in configuration element.

This sounds like a new feature that we would have to add support for in Nimble, and we could update the pattern used to something like `^user:([^:]*)(::.*)?` to support appending an optional url. I'm just thinking about this from a client's perspective, and I'm confused about why the pattern is configurable instead of just documented. Would I ever want to use a different pattern in different viewer/editor instances? Can I just use the pattern `'user'` or even `''`? What are the implications of my pattern choice?

At the least, it would probably be good to export a constant for `'^user:(.*)` and recommend using that.

@rajsite I'm curious to hear your thoughts on this.

**Review by m-akinc — APPROVED** [review]

File: packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (rajsite) [reply packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts:36]: The idea was not to be coupled to a specific system's representation of user hrefs.
SystemLink doesn't have user profile pages but if it did, at like `https://test.lifecyclesolutions.ni.com/users/milan` then you would want a regex that can match that url pattern, something like `http:\/\/.*/users\/.*`
On the other side of the fence they are choosing to use `user:` I think, but that's not a requirement of all `rich-text-mention-users` users.
We are also not precluded from making a SystemLink specific `rich-text-systemlink-mention-users` type thing in the future that keeps the pattern hard coded and internal.

File: packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts [hunk 17]
```diff
@@ -2,13 +2,43 @@ import { html } from '@microsoft/fast-element';
 import { RichTextViewer, richTextViewerTag } from '..';
 import { fixture, type Fixture } from '../../../utilities/tests/fixture';
 import { RichTextViewerPageObject } from '../testing/rich-text-viewer.pageobject';
+import { anchorTag } from '../../../anchor';
```
jattasNI, line 5: We should probably also add test cases using the pattern in #1633

jattasNI (commented):
I only had a chance to play with Storybook today, not review the implementation. Will look more tomorrow.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
aagash-ni: I think the PR you mentioned above is testing strategy PR for serializer, In that what pattern you are meaning @jattasNI ?

jattasNI: Sorry, left that comment in a rush at the end of the day. We will want to write interaction tests like that when we add @mention support to the rich text editor but it's not needed for the rich text viewer.

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
jattasNI: @m-akinc think we should split that PR out from the rest of #1620?

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts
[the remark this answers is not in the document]
jattasNI: Could we add some test coverage for the case where the mention regex matches certain links that start with http:// or https://? It'd be good to include this in both storybook and unit tests. In storybook it could be a selector to set different match patterns.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -2,13 +2,43 @@ import { html } from '@microsoft/fast-element';
...
+import { richTextMentionUsersViewTag } from '../../../rich-text-mention/users/view';
+import { MappingUser, mappingUserTag } from '../../../mapping/user';
 
 async function setup(): Promise<Fixture<RichTextViewer>> {
     return fixture<RichTextViewer>(
         html`<nimble-rich-text-viewer></nimble-rich-text-viewer>`
     );
 }
 
+async function appendUserMentionConfiguration(
```
jattasNI, line 20: I think we should have some test coverage for:
1. mention users are supplied via template rather than appended later
2. multiple mapping users

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+# Usage Docs
+
+## User Mention Usage
+
+To enable support for `@mention` in the `nimble-rich-text-viewer` component, the following configuration and mapping components should be added as children to convert the input markdown into a mention node, rendering a user name. Also, see [Client Usage Guidance on Filtered Users](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text/specs/mention-hld.md#client-usage-guidance-on-filtered-users-1) for instructions on loading the mapping elements into the viewer component. Follow the document for specific APIs in the below configuration and mapping elements.
+
+## User Configuration
+
+<Description of={mentionUserStories} />
```
jattasNI, line 29: This description includes some text that's specific to the editor rather than the viewer (talks about the @mention button and keystroke). We should either make the shared text more general or export separate descriptions for the editor and the viewer.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
```diff
@@ -0,0 +1,35 @@
...
+<Description of={viewerStories} />
+
+<DocsStory of={viewerStories.richTextViewer} expanded={false} />
+<Controls of={viewerStories.richTextViewer} />
+
+# Usage Docs
+
+## User Mention Usage
+
+To enable support for `@mention` in the `nimble-rich-text-viewer` component, the following configuration and mapping components should be added as children to convert the input markdown into a mention node, rendering a user name. Also, see [Client Usage Guidance on Filtered Users](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text/specs/mention-hld.md#client-usage-guidance-on-filtered-users-1) for instructions on loading the mapping elements into the viewer component. Follow the document for specific APIs in the below configuration and mapping elements.
```
jattasNI, line 25: I think we should provide the guidance about which users to include inline within this doc rather than linking to the spec. It can also be more succinct: simply state that every user mentioned in the markdown should have a corresponding user mapping.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+## User Configuration
+
+<Description of={mentionUserStories} />
+<Controls of={mentionUserStories.richTextMentionUsers} />
```
jattasNI, line 30: This contains an interactive control that lets you change the regex pattern but it doesn't affect the rich text viewer within this page. We should either make it work (possibly with a fixed set of pre-selected patterns as I mentioned in another comment thread) or not make it a control.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -27,6 +32,24 @@ export class RichTextViewer extends FoundationElement {
...
 
+    /**
+     * @internal
+     */
+    public mentions: RichTextMention[] = [];
+
+    /**
+     * @internal
+     */
+    public mentionsMap: Map<string, MentionInternals<RichTextMentionConfig>> = new Map();
```
jattasNI, line 43: What do you think about giving these names that distinguish the type of mention object in addition / instead of the data structure that stores the mention object? Maybe `mentionElements` and `mentionConfigs`/`mentionIdToConfigMap`/`mentionIdToMentionInternalsMap`? 

Just a half-formed thought. I'm having some trouble following all the various types but that might be because of the branches on top of branches making it harder to browse source right now.

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
m-akinc: If I do, that will require a follow-on to my PR to use the new utility function. I think I'd rather just leave it in my PR and take it on myself to make sure this code gets updated when possible.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -53,7 +64,69 @@ export class RichTextMarkdownParser {
...
+
+                        const currentMention = Array.from(
+                            mentionsMap.values()
+                        ).find(
+                            mention => mention.validConfiguration
+                                && this.validateMentionPattern(
+                                    mention,
+                                    mentionHref
+                                )
+                        );
```
rajsite, line 90: Wait... what's the point of creating the mentionsMap if you just turn it into an array?? It's doesn't look like anything tried to lookup stuff in the map at all lol.

Also wouldn't this code run frequently? All this validation and stuff should have happened way before parseMarkdownToDOM is called and ideally only once.

What the updateMentionsMap function needs to do is capture all the stuff needed for the markdown parser to run at that time. Check validity, create regex objects, etc. Create a configuration that markdown-parser can use without coupling to all these other parts of the system.

rajsite (commented):
It may be a good idea to set-up a meeting for tomorrow to discuss these changes

packages/nimble-components/src/rich-text/viewer/index.ts
aagash-ni: Updated `mentions` to `mentionElements`. Changed the `mentionsMap` to array form instead of Map and renamed to `mentionInternalsList` as the array been generated from the Map on parser.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -44,10 +67,81 @@ export class RichTextViewer extends FoundationElement {
...
+        this.updateMentionsMap();
+    }
+
+    private observeMentions(): void {
+        this.removeMentionObservers();
+
+        for (const mention of this.mentions) {
+            const notifier = Observable.getNotifier(mention);
+            notifier.subscribe(this);
+            this.mentionNotifiers.push(notifier);
```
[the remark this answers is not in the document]
aagash-ni: Removed subscription on all attributes other than `mentionInternals`.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -44,10 +67,81 @@ export class RichTextViewer extends FoundationElement {
...
     }
 
+    /**
+     * @internal
+     */
+    public handleChange(source: unknown, args: unknown): void {
+        if (
+            (source instanceof RichTextMention
+                || source instanceof MentionInternals)
+            && typeof args === 'string'
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -27,6 +32,24 @@ export class RichTextViewer extends FoundationElement {
...
+     */
+    public mentionsMap: Map<string, MentionInternals<RichTextMentionConfig>> = new Map();
+
+    /**
+     * @internal
+     */
+    @observable
+    public readonly childItems: Element[] = [];
+
+    private mentionNotifiers: Notifier[] = [];
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
aagash-ni: Removed its control.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Added tests for https:// and http:// links as mention node and also as a link in markdown-parser.spec.ts file.
Included in the storybook docs as a selector between `https://user/.*` and `user:.*`

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -53,7 +64,69 @@ export class RichTextMarkdownParser {
...
-        supportedTokenizerRules.validateLink = href => /^https?:\/\//i.test(href);
+        // MarkdownIt use: https://markdown-it.github.io/markdown-it/#MarkdownIt.use
+        supportedTokenizerRules.use(
+            markdown => {
+                // `mention` rule is added before `autolink` rule to get the highest preference when getting the DOM output
+                // MarkdownIt before: https://markdown-it.github.io/markdown-it/#Ruler.before
+                markdown.inline.ruler.before(
+                    'autolink',
+                    'mention',
+                    (state, _silent) => {
```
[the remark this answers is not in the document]
vivinkrishna-ni: This entire tokenizer rule logic is removed as per the concern raised by Milan in [the other comment](https://github.com/ni/nimble/pull/1664#discussion_r1396496176) and leveraged the link schema to render the view element.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: Added the above mentioned tests in rich-text-viewer.spec.ts

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
vivinkrishna-ni: Updated the description to be more generic in the user mention storybook. Please let me know if there are any corrections in the doc.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
vivinkrishna-ni: Updated as per the suggestion. Also, I feel keeping the link to the spec doc as is in the storybook so that the client can find more details about how can we get the mentioned users list from the markdown string. Please let me know if we should not link the doc to the storybook.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -65,16 +138,106 @@ export class RichTextMarkdownParser {
...
+        ) {
+            position += 1;
+        }
+
+        if (position === max && state.src.charCodeAt(position) !== closingTag) {
+            return undefined;
+        }
+
+        return state.src.slice(state.pos + 1, position);
+    }
```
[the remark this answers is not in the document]
vivinkrishna-ni: As discussed offline, I have removed the entire logic of parsing the markdown string to identify the mention node instead I made an update in the `toDOM()` method of link schema to differentiate between mention view element and the absolute link in the markdown-parser.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -23,9 +24,14 @@ export class RichTextMarkdownParser {
...
     public static parseMarkdownToDOM(
-        value: string
+        value: string,
+        mentionsMap?: ReadonlyMap<
+        string,
+        MentionInternals<RichTextMentionConfig>
+        >
     ): HTMLElement | DocumentFragment {
-        const parsedMarkdownContent = this.markdownParser.parse(value);
+        const markdownParser = this.initializeMarkdownParser(mentionsMap);
```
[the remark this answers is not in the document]
vivinkrishna-ni: I have updated the code as suggested.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: > what's the point of creating the mentionsMap if you just turn it into an array?

I have updated it to an array and identified what type of mention string is in the markdown input in markdown-parser.ts file.

As discussed offline over a call, we aligned to have a separate model class with only the necessary details of what the markdown parser requires. Also checked the validation at the time of creating the parser configuration instead of handling it within the parser. This way only when the mention internals change, all the regex pattern objects are created and the parser just gets the necessary details like the display name when the mention href in the markdown maps to.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.mdx
jattasNI: Linking to the spec for additional information is ok.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -88,10 +95,28 @@ export class RichTextMarkdownParser {
...
+                                    'disable-editing': true
+                                }
+                            ];
+                        }
+
                         return [
                             anchorTag,
                             {
-                                href: node.attrs.href as Attr,
+                                // Absolute links whose scheme is not HTTP/HTTPS will render as anchor tag without `href` attribute
```
jattasNI, line 118: I'm not caught up on the discussions around this. What's the rationale for rendering other schemes as an anchor with no `href`? Alternatives might be:
1. rendering them as text as they are today. e.g. by returning a `<div>` or `<span>`
2. rendering them with an href so users can click them. (I think I recall this being a conscious decision to limit the schemes we start out supporting?)

According to [this thread](https://stackoverflow.com/questions/5637969/is-an-empty-href-valid), clicking an anchor with an empty href results in the page reloading which seems undesirable to me. Though when I test this in the browser with an `ftp://ni.com` link I currently see it render as plain text content, so I don't think I understand how to hit this code path.

jattasNI (commented):
Reviewed everything except the two `.spec.ts` files. I'll cover those once we settle the implementation questions.

Other than the comments I left, the responses to Milan's feedback look good to me.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
```diff
@@ -3,13 +3,18 @@ import type { Meta, StoryObj } from '@storybook/html';
 import { createUserSelectedThemeStory } from '../../../utilities/tests/storybook';
 import { hiddenWrapper } from '../../../utilities/tests/hidden';
 
+const patternDescription = `A regex serves purpose of validation against the input mention markdown string. It is also used to extract the user ID from user hrefs during the parsing of markdown in the rich text components. 
+
+* To extract the user ID, the pattern should be grouped to render as mention view element if the user name is not identified in the mapping elements but matches the pattern. For example, the pattern should be \`user:(.*)\` to extract the user ID which is next to \`user:\`. If the pattern does not have grouping regex (\`user:.*\`), also if the mapping element for the particular user href is not found, the given input will render as plain text in the rich text components.
```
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for the improved description. Regarding displaying the user key instead of the display name, it occurs only when neither the pattern has a group _nor_ the mapping element is missing; it must be the absence of both conditions for rendering as plain text. I have updated the same with that correction in the doc as well.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.stories.ts
```diff
@@ -32,16 +49,38 @@ const metadata: Meta<RichTextViewerArgs> = {
...
+        },
+        data: {
+            name: 'pattern and href',
+            description:
+                'Pattern and href selector for the user mention configuration and mapping element(s).',
+            options: Object.values(exampleDataType),
+            control: {
+                type: 'radio',
+                labels: {
+                    [exampleDataType.userPattern]: 'User Pattern (user:.*)',
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, it should be there. I have updated the same.

packages/nimble-components/src/rich-text/viewer/styles.ts
```diff
@@ -41,4 +45,11 @@ export const styles = css`
     li > p:empty {
         display: none;
     }
+
+    ${
+        /* Forcing the font color of the anchor to be in the default link color irrespective of whether it has `href` attribute or not  */ ''
```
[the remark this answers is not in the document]
vivinkrishna-ni: Added a code comment above about the rationale for changing the default font color when `href` is removed from `nimble-anchor`. I have replied to the other comment in markdown-parser on what is the rationale for this change in the absolute link behavior.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -39,15 +60,93 @@ export class RichTextViewer extends FoundationElement {
...
+    private hasDuplicateConfigurationElement(): boolean {
+        const mentionChars = new Set<string>();
+
+        return this.mentionElements.some(mention => {
+            const hasDuplicateMentionChar = mentionChars.has(
+                mention.mentionInternals.character
+            );
+            mentionChars.add(mention.mentionInternals.character);
+            return hasDuplicateMentionChar;
+        });
```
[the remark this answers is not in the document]
vivinkrishna-ni: Updated the same.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -39,15 +60,93 @@ export class RichTextViewer extends FoundationElement {
...
+            .map(
+                mention => new MarkdownParserMentionConfiguration(
+                    mention.mentionInternals
+                )
+            );
+
+        this.updateView();
+    }
+
+    private hasDuplicateConfigurationElement(): boolean {
```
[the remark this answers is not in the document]
vivinkrishna-ni: It is because all the validation handled in the mention-validator holds specific to the configuration element and mapping elements, not on the number of configuration elements within the children of the viewer component.

The problem we see in not including it in the mention-validator is that the creation of mention configs happens only based on the valid state of the elements exactly at [this part of the code](https://github.com/ni/nimble/blob/d5330e74cfc855856bed26fc6b6f2fe3dde2eadd/packages/nimble-components/src/rich-text-mention/base/index.ts#L106). Also, we can identify if there are duplicate configuration elements only in the parent component (viewer/editor) so we are not sure how can utilize the mention-validator for this case because each configuration has its own validation that might not be able to determine what element in the children is duplicate.

We are validating this at the top layer (`childItemsChanged`) if there are duplicate configurations found and avoid other processes like observing the mention internals change or updating the mention configurations. Only after the duplicates are removed from the children, the subscriptions are added, then listen to the mention internals state.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: This decision was taken during one of the discussions in the HLD and here is the link to the conversation: https://github.com/ni/nimble/pull/1594#discussion_r1377876092.

The summary of the discussion is all the links in the rich text components should be loaded as `anchor` tags irrespective of their scheme because the absolute link and mention share the same markdown format. So the goal for us is to render mention nodes even if it is not actually a mention in the first place but an absolute link with the matching pattern of mention. 

For instance, initially, if the markdown string contains `<user:1>` but the configuration element is not added, it renders it as plain text (assuming non-HTTPS/HTTP links) in the editor within the `<p>` tag. Now when the configuration element is added dynamically it has the pattern of `user:.*` so it is a mention now. Here when performing getMarkdown + setMarkdown, the string will not be in autolink format as the DOM rendered within the `<p>` node so it won't be converted into a mention node.

To address this scenario, we decided to render all the supported autolink within `anchor` tag so that when getMarkdown is called, we get the string in autolink format and setMarkdown will parse the string and convert the same link to a mention node when configuration element added dynamically.

But non-HTTPS/HTTP links should not be clickable links as per the requirement of absolute links, we decided to remove the `href` as suggested by Milan [here](https://github.com/ni/nimble/pull/1594#discussion_r1379256654) and seems it is a right way to do it.

> I don't think I understand how to hit this code path.

The default behavior of all the HTML `<a>` elements renders as plain text if there is no `href`. But the thread you linked above seems to have an empty string to the `href` attribute but in our case we completely removed the attribute if it is not a HTTPS/HTTP link.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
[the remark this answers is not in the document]
jattasNI: I'm afraid I'm still confused about the behavior and I think the description is missing some cases. Could you fill in the *Renders* column with cases like "key", "display name", "plain text or URL" for the cases below (I took a first pass but you should double check my guesses)? Then we can come up with a description that covers the cases succinctly. And then we should also ensure that we have test coverage for the relevant cases.

| Regex has group | Mapping exists with key | Display name set on mapping | Renders |
|--|--|--|--|
| Y | Y | Y | display name |
| Y | Y | N |  ? |
| Y | N | Y | N/A |
| Y | N | N |  ? |
| N | Y | Y |  ? |
| N | Y | N |  plain text or URL |
| N | N | Y |  N/A |
| N | N | N |  key |

Here's the markdown table syntax for you to copy:
```
| Regex has group | Mapping exists with key | Display name set on mapping | Renders |
|--|--|--|--|
| Y | Y | Y | display name |
| Y | Y | N |  ? |
| Y | N | Y | N/A |
| Y | N | N |  ? |
| N | Y | Y |  ? |
| N | Y | N |  plain text or URL |
| N | N | Y |  N/A |
| N | N | N |  key |
```

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -88,10 +100,40 @@ export class RichTextMarkdownParser {
...
+                            ];
+                        }
+
                         return [
                             anchorTag,
                             {
-                                href: node.attrs.href as Attr,
+                                /**
+                                 * Both mention and absolute link markdown share the autolink format in CommonMark flavor.
+                                 * However, absolute links without HTTPS/HTTP should render as plain text, while mentions are
```
jattasNI, line 125:
```suggestion
                                 * However, absolute links without HTTPS/HTTP should render as plain text (anchor with no href), while mentions are
```

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -88,10 +100,40 @@ export class RichTextMarkdownParser {
...
+                                 * displayed as view elements.
+                                 *
+                                 * To achieve this, configuration elements matching already rendered absolute links in the rich
+                                 * text components should be converted to mention nodes. We should preserve markdown for
```
jattasNI, line 129: This part doesn't seem right. By "configuration elements" do you mean `rich-text-mention-users` and `nimble-mapping-user`? Those don't know anything about the markdown content, so they aren't converted to mentions.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
jattasNI: Thanks for the explanation, that makes sense. My main concern is ensuring we have test coverage for all these cases. I haven't looked at the spec files yet though.

packages/nimble-components/src/rich-text/viewer/styles.ts
```diff
@@ -41,4 +45,20 @@ export const styles = css`
...
+    ${
+        /**
+         * When an absolute link is not HTTPS/HTTP, the anchor tag renders without an `href`, appearing as plain text.
+         * However, the `nimble-anchor` displays differently in color when the `href` attribute is absent.
+         * To ensure a consistent appearance, the font color is forced to the default link color regardless of the `href`
+         * attribute's presence.
+         *
+         * This visual consistency stems from the similarity between mention markdown and absolute links, with the distinction
+         * that mention markdown does not restrict the link scheme. When converting absolute links without HTTPS/HTTP to mention
+         * views, preserve the markdown in autolink format for proper rendering.
```
jattasNI, line 58: Thanks for adding this comment. I might actually say it's *too* detailed, leaving it more likely to get out of date as the code evolves. Let's go with something more focused on the styling and less on the other details.
```suggestion
         *
         * See models/markdown-parser.ts where link elements are emitted for more info.
```

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
jattasNI: Gotcha, that makes sense that an individual mention configuration can't validate this state because it depends on its siblings. But I think the right place to validate it is in a validator for the rich text viewer. This is preferable to your current approach because it allows clients to inspect the validator to see why their mentions aren't rendering.

This pattern is similar to how [the table has a validator](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/models/table-validator.ts) which handles table validity and aggregates validity of its columns (see `invalidColumnConfiguration`). But now the rich text viewer will have a validator which handles this case and also aggregates the validity of the mention config elements.

Doing that would be a good sized PR on its own. Ideally it would be a separate PR that precedes this one. But I think it'd also be acceptable to remove the code from this PR, leave a TODO comment describing this case, and do something sane for now like pick the first config element that matches the mention character. Then in a follow up PR you could implement the validation I described above and remove the TODO.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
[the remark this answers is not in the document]
vivinkrishna-ni: Thanks for sharing the markdown syntax for the table. Here is the updated table for the below cases as per the current implementation:

| Regex has group | Mapping exists with key | Display name set on mapping | Renders |
|--|--|--|--|
| Y | Y | Y | display name |
| Y | Y | N |  plain text or URL |
| Y | N | Y | N/A (key) |
| Y | N | N |  key |
| N | Y | Y |  display name |
| N | Y | N |  plain text or URL |
| N | N | Y | N/A (plain text or URL)  |
| N | N | N |  ~key~ plain text or URL  |

In the above table, I am assuming the "key" is the user ID in `@mention` and not the entire mention-href.

The plain text or URL depends on the scheme of the autolink in markdown i.e. HTTPS/HTTP links. Also, in the second case, you might be wondering why it is "plain text or URL" instead of "key", it is because if the `display-name` is missing from the mapping element then it is a failed validation captured in mention-validator of user mentions so displaying as a link and not mention.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: I meant the "configuration element" is `rich-text-mention-users` and updated the same in detail there. Also, I meant the URL rendered in the components should be converted to mention neither the configuration element nor the markdown string.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: I have written test cases to cover all these cases in the `markdown-parser.spec.ts` file.

packages/nimble-components/src/rich-text/viewer/styles.ts
vivinkrishna-ni: Updated as per the suggestion.

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: It makes sense to have it as a validator for the viewer component like the table validator. Thanks for sharing the references. 

As suggested, I have removed the validation from here, added a TODO in the files needed, and will handle this validation in the follow-up PR.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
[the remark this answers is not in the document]
jattasNI: Very helpful, thanks! To summarize it by grouping some of the cases, mention links will display:
1. the display name if the regex pattern matches, a mapping is found for that key, and that mapping has a display name
2. the key if the regex pattern matches and the regex has a group but no mapping is found
3. a URL or plain text otherwise, depending on whether the mention is HTTP/HTTPS

Could you update the storybook docs with a summary like that?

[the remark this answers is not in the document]
vivinkrishna-ni: I have updated this summary in the pattern description.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -923,4 +931,324 @@ describe('Markdown parser', () => {
...
+
+            expect(getTagsFromElement(doc)).toEqual([
+                'P',
+                `${anchorTag}`.toUpperCase()
+            ]);
+            expect(lastChildElementHasAttribute('href', doc)).toBeFalse();
+            expect(getLeafContentsFromElement(doc)).toEqual(['user:1']);
+        });
+
+        it('should get anchor element with href when autolink markdown format is HTTP but does not match with the pattern', async () => {
```
jattasNI, line 1170: I think we're missing a case where the mention `pattern` matches certain HTTP or HTTPS URLs but not others. For example, if the pattern is `^http://user/(.*)` and the contents are `<http://ni.com>` then it should render as a link, not a mention.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -2,13 +2,57 @@ import { html } from '@microsoft/fast-element';
...
+        html`<${richTextViewerTag}>
+            <${richTextMentionUsersTag} pattern="^user:(.*)">
+                <${mappingUserTag} key="user:1" display-name="John Doe"></${mappingUserTag}>
+                <${mappingUserTag} key="user:2" display-name="Mary Wilson"></${mappingUserTag}>
+            </${richTextMentionUsersTag}>
+        </${richTextViewerTag}>`
+    );
+}
+
+async function appendUserMentionConfiguration(
```
jattasNI, line 32: nit: could you make `userKeys` and `displayNames` [optional parameters](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters) instead of setting their type to include `undefined`? Then the callers wouldn't need to explicitly pass `undefined` for those parameters and could instead call with `appendUserMentionConfiguration(element)`.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -81,4 +125,268 @@ describe('RichTextViewer', () => {
...
+                `${richTextMentionUsersViewTag}`.toUpperCase()
+            ]);
+            expect(
+                pageObject.getRenderedMarkdownLastChildAttribute(
+                    'mention-label'
+                )
+            ).toEqual('updated-name');
+        });
+
+        it('updating valid `key` in mapping mention should update it to a mention view if it is a absolute link before', async () => {
```
jattasNI, line 306: Could you add a test for the opposite of this case? i.e. change the key so that a mention which was previously unmatched becomes matched, converting it from a link to a mention view.

jattasNI (approved):
Once you address my last few minor comments, I'm OK with you completing this PR. If @rajsite has additional feedback you can cover it in follow up PRs.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
vivinkrishna-ni: @rajsite I have updated the parser configuration as per our discussion last week and I hope is current implementation is what you would want us to make. If you have any concerns with the change, we will either update it in the next follow-up PRs for @mention or create a new PR to address your feedback. For now, I am closing this comment.

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
vivinkrishna-ni: I will close this comment for now and will make changes related to this once the linked PR is merged.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
vivinkrishna-ni: Added the suggested test case.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: Made them as optional parameters and removed calling the `undefined` in the function calls within the tests

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: Added a test case for the opposite case.

PR #1668, opened 2023-11-15
Introduce table hierarchical datatype

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #861

This is a staging PR in that it will lay the type groundwork that we can ultimately leverage for representing hierarchical data in the table. There are _no_ functional changes in this PR.

## 👩‍💻 Implementation

Creating a new datatype to represent hierarchical data, which Tanstack requires in order to visualize hierarchical data properly. This type is used for the internal representation of the table data, and thus does not affect our public facing datatype used in the `setData` method. This new datatype basically wraps each record in another object, and also provides another field typed as an array of the new datatype (providing hierarchy).

## 🧪 Testing

Ensured existing tests pass.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-376b2ed8-09ef-409f-9acb-e7f04b283d39.json
+++ b/change/@ni-nimble-components-376b2ed8-09ef-409f-9acb-e7f04b283d39.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Introduce table hierarchical datatype.",
+  "packageName": "@ni/nimble-components",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table/index.ts
+++ b/packages/nimble-components/src/table/index.ts
@@ -34,7 +34,7 @@ import { TableValidator } from './models/table-validator';
 import { styles } from './styles';
 import { template } from './template';
 import {
-    InternalTableRecord,
+    TableNode,
     TableActionMenuToggleEventDetail,
     TableColumnConfigurationChangeEventDetail,
     TableColumnSortDirection,
@@ -212,8 +212,8 @@ export class Table<
     @observable
     public documentShiftKeyDown = false;
 
-    private readonly table: TanStackTable<InternalTableRecord<TData>>;
-    private options: TanStackTableOptionsResolved<InternalTableRecord<TData>>;
+    private readonly table: TanStackTable<TableNode<TData>>;
+    private options: TanStackTableOptionsResolved<TableNode<TData>>;
     private readonly tableValidator = new TableValidator();
     private readonly tableUpdateTracker = new TableUpdateTracker(this);
     private readonly selectionManager: InteractiveSelectionManager<TData>;
@@ -268,10 +268,10 @@ export class Table<
         await this.processPendingUpdates();
 
         const data = newData.map(record => {
-            return { data: { ...record } } as InternalTableRecord<TData>;
+            return { clientRecord: { ...record } };
         });
         const tanStackUpdates: Partial<
-        TanStackTableOptionsResolved<InternalTableRecord<TData>>
+        TanStackTableOptionsResolved<TableNode<TData>>
         > = {
             data
         };
@@ -736,7 +736,7 @@ export class Table<
 
     private updateTanStack(): void {
         const updatedOptions: Partial<
-        TanStackTableOptionsResolved<InternalTableRecord<TData>>
+        TanStackTableOptionsResolved<TableNode<TData>>
         > = {};
         updatedOptions.state = {};
 
@@ -804,7 +804,7 @@ export class Table<
         this.validateWithData(this.table.options.data);
     }
 
-    private validateWithData(data: InternalTableRecord[]): void {
+    private validateWithData(data: TableNode[]): void {
         this.tableValidator.validateRecordIds(data, this.idFieldName);
         this.canRenderRows = this.checkValidity();
     }
@@ -856,7 +856,7 @@ export class Table<
         const rows = this.table.getRowModel().rows;
         this.tableData = rows.map(row => {
             const rowState: TableRowState<TData> = {
-                record: row.original.data,
+                record: row.original.clientRecord,
                 id: row.id,
                 selectionState: this.getRowSelectionState(row),
                 isGrouped: row.getIsGrouped(),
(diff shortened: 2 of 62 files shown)


packages/nimble-components/src/table/index.ts
```diff
@@ -211,8 +212,8 @@ export class Table<
     @observable
     public documentShiftKeyDown = false;
 
-    private readonly table: TanStackTable<TData>;
-    private options: TanStackTableOptionsResolved<TData>;
+    private readonly table: TanStackTable<InternalTableRecord<TData>>;
```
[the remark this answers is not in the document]
atmgrifter00: Sure. I'm also open to just a more succinct name for this type. Any suggestions?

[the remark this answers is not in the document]
mollykreis: I'm struggling to come up with a name that makes sense and isn't too easily confused with our existing types. It's not much more succinct, but maybe something like `TypedInternalTableRecord` would at least help avoid the nested generics.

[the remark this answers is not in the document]
atmgrifter00: Turns out we can't easily create a type alias within a class definition, so we're just moving ahead with what's there for now.

jattasNI (approved):

change/@ni-nimble-components-376b2ed8-09ef-409f-9acb-e7f04b283d39.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Any specs that need updating to capture this concept?

packages/nimble-components/src/table/types.ts
```diff
@@ -40,6 +40,18 @@ export interface TableRecord {
...
+/**
+ * InternalTableRecord describes a hierarchical data structure that is used for
+ * the internal representation of the data, and allows us to represent data with
+ * parent-child relationships within Tanstack.
+ */
+export interface InternalTableRecord<
+    TRecord extends TableRecord = TableRecord
+> {
+    subRows?: InternalTableRecord<TRecord>[];
+    data: TRecord;
```
rajsite, line 52: how would you feel about calling this `originalRecord`, `userRecord`, `providedRecord`?
I think I like `userRecord` the most. It's a lot clearer to me what the context is when reading a test like the following:
```ts
 const data = [
      { data: { stringField: 'value-1', numberField: 10 } },
      { data: { stringField: 'value-2', numberField: 11 } }
  ];

// vs

 const data = [
    { userRecord: { stringField: 'value-1', numberField: 10 } },
    { userRecord: { stringField: 'value-2', numberField: 11 } }
];
```
Also could we do some typing around saying it is readonly?

packages/nimble-components/src/table/types.ts
```diff
@@ -40,6 +40,18 @@ export interface TableRecord {
     [key: TableFieldName]: TableFieldValue;
 }
 
+/**
+ * InternalTableRecord describes a hierarchical data structure that is used for
+ * the internal representation of the data, and allows us to represent data with
+ * parent-child relationships within Tanstack.
+ */
+export interface InternalTableRecord<
```
rajsite, line 48: I'm guessing we can save index on InternalTableRecords in the future right? like `userIndex` for the `userRecord`?

packages/nimble-components/src/table/index.ts
```diff
@@ -267,9 +268,11 @@ export class Table<
         await this.processPendingUpdates();
 
         const data = newData.map(record => {
-            return { ...record };
+            return { data: { ...record } } as InternalTableRecord<TData>;
```
rajsite, line 271: shouldn't need the as cast, the fn provided to the map should just specify a return type

packages/nimble-components/src/table/types.ts
jattasNI: I like the idea of a more descriptive name but have a minor preference not to use the term "user" since I'm on a fruitless crusade to use language that differentiates between end-users and clients. I'd vote `originalRecord` or `clientRecord`.

packages/nimble-components/src/table/types.ts
jattasNI, line 48: Another possible name for this type would be `TableNode` or `TableRecordNode` to convey that it's a node in a tree. I think I like that slightly better because it conveys its purpose rather than its visibility (internal). But it's your call and I'm fine with the current name.

packages/nimble-components/src/table/types.ts
rajsite: Think in this context I'd lean more to `clientRecord`. Seems more indicative of the thing that came from outside than `originalRecord`

packages/nimble-components/src/table/types.ts
rajsite: Based on the other convo `clientRecord` and `clientRecordIndex` :)

atmgrifter00: I don't think we can do that through this mechanism unfortunately. This interface needs to mirror what the 3rd party `arrayToTree` method produces. I handled tracking the index in the prototype by keeping a map that associated row Id with a set of metadata, one piece of which was the original index in the array.

Assuming there is more to discuss on this topic, my understanding is that this doesn't need to hold up this PR.

change/@ni-nimble-components-376b2ed8-09ef-409f-9acb-e7f04b283d39.json
atmgrifter00: Added some docs to the table data API spec.

packages/nimble-components/src/table/types.ts
atmgrifter00: Went with `TableNode` as that saved me four whole syllables!

packages/nimble-components/src/table/types.ts
```diff
@@ -40,6 +40,16 @@ export interface TableRecord {
     [key: TableFieldName]: TableFieldValue;
 }
 
+/**
+ * InternalTableRecord describes a hierarchical data structure that is used for
```
jattasNI, line 44:
```suggestion
 * Describes a hierarchical data structure that is used for
```

jattasNI (approved):

jattasNI:
@rajsite @mollykreis I told @atmgrifter00 we can bypass your final review on this one since you're OOO and there will be opportunities for more feedback on the follow up PRs

PR #1673, opened 2023-11-17
Enable Renovate for nuget dependencies

jattasNI:
# Pull Request

## 🤨 Rationale

We have a few Nuget dependencies that would be good to keep up to date. Some of them have security vulnerabilities as revealed by Snyk PRs like #1670 but we didn't like the way those PRs looked as we'd prefer they be updated in a group.

## 👩‍💻 Implementation

1. Configure Renovate to update Nuget dependencies in addition to npm ones. It [doesn't look like](https://docs.renovatebot.com/configuration-options/#rangestrategy) Renovate supports [lockfile only updates for Nuget](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/#:~:text=To%20enable%20the%20use%20of%20lock%20file%20with,%E2%80%93%20packages.lock.json%20file%20at%20the%20project%20root%20directory.) so I think owners will have to push beachball change files to these PRs.
2. Update CONTRIBUTING docs to mention that Renovate might modify `.csproj` files and thus might need beachball change files.

## 🧪 Testing

When this PR completes I'll poke Renovate and see what PRs it creates.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1675, opened 2023-11-20
Fix nuget Renovate config

### LINKED ISSUE #1674: Action Required: Fix Renovate Configuration
There is an error with this repository's Renovate configuration that needs to be fixed. As a precaution, Renovate will stop PRs until it is resolved.

Location: `.github/renovate.json`
Error type: The renovate configuration file contains some invalid settings
Message: packageRules[3]: Each packageRule must contain at least one match* or exclude* selector. Rule: {"groupName":"Nuget dependencies","rangeStrategy":"auto","enabled":true}

jattasNI:
# Pull Request

## 🤨 Rationale

After #1673, Renovate created #1674 saying the config I set was invalid. The issue is that the `packageRule` I created for nuget didn't contain any `match*` or `exclude*` rules so Renovate can't figure out what it applies to.

Resolves #1674 (hopefully)

## 👩‍💻 Implementation

I had only created that rule to try to set the update strategy to `auto` for nuget. Other than that we want to start out with the default configuration of updating all nugets. So I 
1. removed the place where we were globally changing update strategy to `lockfile-only`
2. set it on npm package rules instead (and also made it more explicit that those were npm rules)
3. deleted the nuget package rule

## 🧪 Testing

We'll see what Renovate does with this after I submit.

## ✅ Checklist

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,7 +4,6 @@
     "config:base"
   ],
   "enabledManagers": ["npm", "nuget"],
-  "rangeStrategy": "update-lockfile",
   "rebaseWhen": "conflicted",
   "schedule": [
     "monthly"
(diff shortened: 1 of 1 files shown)


jattasNI:
@rajsite I'm completing this PR without your review since you're OOO and it (hopefully) fixes a Renovate error.

PR #1677, opened 2023-11-20
Hide inc/dec buttons on numeric from a11y tree

### LINKED ISSUE #1644: Accessibility tree for number field claims increment/decrement buttons are focusable
# 🐛 Bug Report

In the accessibility tree, the increment/decrement buttons  of a number field are shown as focusable. They are not focusable.

## 💻 Repro or Code Sample

Seems to affect all number fields, so you can see the issue on the [Number Field storybook page](https://nimble.ni.dev/storybook/?path=/docs/components-number-field--docs). The issue is visible in both the regular accessibility tree and the full page accessibility tree view with Chrome dev tools.

image
image

## 🤔 Expected Behavior

The "focusable" attribute should be "false" in the accessibility tree for the increment/decrement buttons of a number field.

The [`fast-number-field` example in the explore view](https://explore.fast.design/components/fast-number-field) doesn't show the increment/decrement buttons in the accessibility tree at all.

## 😯 Current Behavior

The "focusable" attribute is "true" in the accessibility tree for the increment/decrement buttons of a number field.

## 💁 Possible Solution

Unsure of the solution. The increment/decrement buttons have tabindex="-1" set [here](https://github.com/ni/nimble/blob/f159a084527c7e53bac99dd43182c10c2f2f628f/packages/nimble-components/src/number-field/index.ts#L71).

## 🔦 Context

Mentioned in PR comment in #1638 [here](https://github.com/ni/nimble/pull/1638#discussion_r1379435727).

## 🌍 Your Environment

* OS & Device: Windows 10 on PC
* Browser Google Chrome 118.0.5993.90
* Version [e.g. 1.8.0]

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1644

## 👩‍💻 Implementation

Added `aria-hidden="true"` to both inc/dec buttons. As these buttons are not focusable, and there are other keyboard interactions (up/down arrow) for incrementing/decrementing the value, these buttons are not interesting to assistive technologies. The `<button>` element implicitly has `focusable=true` for a11y, and this cannot be changed, even if it is removed from the tab order with `tabIndex=-1`. If we want to keep it as an unfocusable button, we must use `aria-hidden`.

## 🧪 Testing

Checked in Storybook.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-9da58386-e822-45d9-a191-56fc2717b88b.json
+++ b/change/@ni-nimble-components-9da58386-e822-45d9-a191-56fc2717b88b.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Hide inc/dec buttons on numeric from a11y tree",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/number-field/index.ts
+++ b/packages/nimble-components/src/number-field/index.ts
@@ -74,6 +74,7 @@ const nimbleNumberField = NumberField.compose<NumberFieldOptions>({
             appearance="ghost"
             content-hidden
             tabindex="-1"
+            aria-hidden="true"
         >
             ${x => numericDecrementLabel.getValueFor(x)}
             <${iconMinusWideTag}
@@ -88,6 +89,7 @@ const nimbleNumberField = NumberField.compose<NumberFieldOptions>({
             appearance="ghost"
             content-hidden
             tabindex="-1"
+            aria-hidden="true"
         >
             ${x => numericIncrementLabel.getValueFor(x)}
             <${iconAddTag}
--- a/packages/nimble-components/src/number-field/tests/number-field.spec.ts
+++ b/packages/nimble-components/src/number-field/tests/number-field.spec.ts
@@ -23,9 +23,31 @@ describe('NumberField', () => {
     });
 
     it('can construct an element instance', () => {
-        expect(document.createElement('nimble-number-field')).toBeInstanceOf(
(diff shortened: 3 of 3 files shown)


packages/nimble-components/src/number-field/index.ts
```diff
@@ -74,6 +74,7 @@ const nimbleNumberField = NumberField.compose<NumberFieldOptions>({
             appearance="ghost"
             content-hidden
             tabindex="-1"
+            aria-hidden="true"
```
jattasNI, line 77: I am surprised we don't already have tests to validate these attributes. Would you mind at least adding coverage for `tabindex` and `aria-hidden` to assert the intended behavior that you describe in the PR description?

change/@ni-nimble-components-9da58386-e822-45d9-a191-56fc2717b88b.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I tested in Edge and it successfully prunes the accessibility tree. In Firefox's accessibility tree, however, I still see the buttons and they have a warning saying that clickable elements must be focusable. I found [many many Mozilla issues](https://bugzilla.mozilla.org/show_bug.cgi?id=aria-hidden) about this. It seems there has been a decade-long debate about whether it's correct to remove these from the tree. I couldn't actually figure out what they concluded, if anything. Some of the issues seem to suggest they are now supposed to be pruned, but that's not happening for me.

I'm mostly mentioning this for posterity and in case you want to dig in further. If not I think it's OK to move forward with this as-is since it improves behavior in Chromium browsers (Safari doesn't have a dev tools a11y tree so I didn't test there).

jattasNI (approved):
This seems simple enough that it's OK to complete with just my review. We can follow up on any feedback from @rajsite if he reviews after he's back.

PR #1678, opened 2023-11-20
Change Nimble Wafer Map Die Highlighting

Razvan1928:
The details can be found in the spec file


packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,41 @@
...
+N\A
+
+## Implementation / Design
+
+The current highlight logic has the benefit of being simple and easy to use. We wish to build on top of how highlight is made.
+
+1. Add a boolean field on each die called "isHighlighted" this will be used when parsing the wafer to chose the color of the die.
+![isHighlighted](./resources/isHighlighted.PNG)
+
+2. Add a field on the wafer map "HighlightBy", which will be used to decide if the highlight should be done by "Value" (current approach) or by "Coordinates" our proposal.
```
atmgrifter00, line 26: I think `HighlightMode` might be a better name, as this does essentially seem like a "mode" that the `WaferMap` is in.

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,41 @@
...
+
+index.ts
+
+![highlightBy](./resources/HighlightBy.PNG)
+
+types.ts
+
+![highlightByEnum](./resources/HighlightedByEnum.PNG)
+
+This approach will let us to highlight any die we want, without any restrictions. The logic that decides which dies should be highlight can be moved completely out of the wafer map component, leaving the component the duty of only highlighting the given dies using the "isHighlighted" field.
```
atmgrifter00, line 36:
```suggestion
This approach will let us to highlight any die we want, without any restrictions. The logic that decides which dies should be highlighted can be moved completely out of the wafer map component, leaving the component the duty of only highlighting the given dies using the "isHighlighted" field.
```

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,41 @@
...
+## Links To Relevant Work Items and Reference Material
+
+N\A
+
+## Implementation / Design
+
+The current highlight logic has the benefit of being simple and easy to use. We wish to build on top of how highlight is made.
+
+1. Add a boolean field on each die called "isHighlighted" this will be used when parsing the wafer to chose the color of the die.
+![isHighlighted](./resources/isHighlighted.PNG)
```
atmgrifter00, line 24: nit: I think when you are including what are essentially code snippets, it's probably preferable to just use the markdown code snippet syntax, instead of images.

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Links To Relevant Work Items and Reference Material
+
+N\A
+
+## Implementation / Design
+
+The current highlight logic has the benefit of being simple and easy to use. We wish to build on top of how highlight is made.
+
+1. Add a boolean field on each die called "isHighlighted" this will be used when parsing the wafer to chose the color of the die.
```
atmgrifter00, line 23: Any reason the `isHighlighted` field can't be nullable?

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
Razvan1928: good point!

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
Razvan1928: good one

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,64 @@
...
+   Value,
+   Coordinates
+}
+```
+
+This approach will let us to highlight any die we want, without any restrictions. The logic that decides which dies should be highlighted can be moved completely out of the wafer map component, leaving the component the duty of only highlighting the given dies using the "isHighlighted" field.
+
+## Alternative Implementation / Design
+
+1. Add a boolean field on each die called "isHighlighted" this will be used when parsing the wafer to choose the color of the die.
```
munteannatan, line 63: I think you should remove this because this is the proposed implementation, this section is just for alternatives to it

munteannatan (approved):

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
Razvan1928: You're god damn right son

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,59 @@
...
+
+![Highlighted Values](resources/highlighted-values.png)
+
+The current approach is limited. We can either highlight all the dies having 2s and 3s or neither of them. The customer needs exceeds this approach. The customer may see an anomaly on the dies having DieX: 15 and Soft Bin: 2 and wishes to highlight all these values, but the current code is not capable of this.
+
+## Links To Relevant Work Items and Reference Material
+
+N\A
+
+## Implementation / Design
```
jattasNI, line 21: I thought of a couple other potential APIs and I'm curious for your opinions on whether they would be preferable:

1. remove `HighlightMode` and instead highlight the union of dies from `highlightedValues` and `isHighlighted`. That way:
    - `HighlightMode.Nothing` option could be achieved by leaving both `highlightedValues` and `isHighlighted` unset
    -  `HighlightMode.Value` option could be achieved by leaving `isHighlighted` unset
    - `HighlightMode.Coordinates` option could be achieved by leaving `highlightedValues` unset
2. Remove `highlightedValues` entirely and instead require clients to update `isHighlighted` on every `WaferMapDie` with a particular value

From an API design perspective I think both options are preferable to the current proposal because the API surface area is smaller so there are fewer cases to test for (with 2 being even simpler than 1, obviously). But I don't have a sense for the impact on clients or the complexity/performance implications.

Razvan1928: Good ideas, I will go with the first one

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,81 @@
...
```
jattasNI, line 21: Reminder that, in addition to the code changes and corresponding tests, it would be good to update Storybook to document the new field and include an example. One way to do this would be to add a new option to the `dies` control that includes a set of values with some of them highlighted.

image

jattasNI (approved):
Heads up that Milan and Molly are out of office this week so won't be able to review until next week. I think it would be good to wait for Milan's review especially since he's been more involved with the wafer map API design than me.

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
rajsite: It kinda feels like we are conflating data with presentation a bit in the api / conflating incidental view state with the data. Feels similar to the selection state of rows on the table.

Some thoughts:
- I think the main problem with highlightedValues is that it can't address independent wafers uniquely
- Updating the data is an expensive operation. Lets say you wanted to update the highlighted state interactively with mouse moves. We shouldn't need to reset the whole data set, just the highlighted state.
- Maybe we want additional incidental state in the future like `selected` or `loading` states for specific wafers

Maybe we should do something more like the table:
- Include an `id?: string` in `WaferMapDie`
- Keep configuration of the highlighted wafers separate, i.e. `highlightedDies: string[] = []`

Concretely, the alternate proposals are:
1. Add an optional `id` to `WaferMapDie`. Validate that all ids are unique or all are are absent. Add `highlightedDies` which takes an array of `id`s. If any ids in `highlightedDies` don't match the data in `dies` just ignore them.
2. Just use the index in `dies` as the identifier for a wafer and don't add an explicit `id`  field. `highlightedDies` just takes an array of indexes.

In all of these proposals I think we should get rid of `highlightedValues`. I don't think there is much value in having both `highlightedValues` and `highlightedDies:[]` / `[{isHighLighted}]`.

munteannatan: The concerns you raised make sense; we would not want to update the data just for the highlight change.

But I also have a concern regarding the id/index approach. The wafer map use-case, differently from the table example you gave, will not have just a small number of dies highlighted. In some cases, we would have most of the dies selected based on the user selecting some die traits. I worry that this will impact the performance of the renderer when it must find the specific id/indexes in the provided list. Do you think this is a valid concern?

As an alternative we are thinking of using secondary values in the metadata which can be used for highlighting along with a secondary highlighting list. 

```
dies: [
   {
       x: 1,
        y: 1,
        value: 12,
        metadata: {
            secondaryValue: 2
        }
    },
    ...
]

secondaryHighlightList: [{secondaryValue: 2}, {secondaryValue: 3} ...]
```
This can cover the scenario you presented if the user will use ids for secondary values and a small list of ids in the secondary highlight list and it can also cover the scenario when we have different die traits from the displayed value that we may want to highlight.

We can also make this list the only highlight list, and when the metadata is absent it will use the actual values.

rajsite: What if instead of unique ids we did tags?

```
dies: [{
  x:1,
  y:1,
  value: 12,
  tags: ['a', 'b']
}]
highlightedDies: ['a']
```

Then you have a lot of flexibility from the client. You could create a unique tag per wafer if you wanted and/or create groups of wafers by giving them the same tag. Kinda similar to CSS classes from that perspective.

munteannatan: It's a great ideea, having the highlight clearly defined to tags makes it less confusing and we leve the whole highlight scheme to the client, so they can make it suit their performance needs

rajsite (commented):
Waiting for updated to discussion on representation change

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,56 @@
...
+@observable public highlightedTags: string[][]
+```
+
+Example:
+
+```
+[
+    [{ 'sb2', 'hb3' }],
+    [{ 'sb1' }]
+]
```
rajsite, line 53: The spec should be precise and consider edge cases:
- This example is not valid syntax. It shows an array of arrays of objects. Provide realistic examples. `[['sb2', 'hb3'], ['sb1']]`
- Is this level of complexity actually needed? Can it just be a list of tags where if you match any tag in the list we highlight? i.e. ['a', 'b'] matches 'a' or 'b'. If this is not enough can you capture it in an alternatives section and describe why it is insufficient?
- If the complexity shown is needed, can you provide some clear use cases of what you are considering?
  - The underlying question is why are these rules enough? Do you need to be able to say something like "must have tag a and b but not c"? Should we have an entire Set Theory engine (intersection, union, complement, difference, etc)? Should we have an entire selection engine like CSS `.a.b:not(.c)`? Do we need to make it even more extensible, like you can plugin completely custom logic and if so what are the specific use cases?
- Clearly describe the rules and validation states. This is groups of tags. So describe precisely, examples:
    - If a wafer matches all the tags in a group then it is highlights. It must match all the tags in at least one group to highlight.
    - What happens if a group is an empty array? Is it ignored? Does it match everything?
    - What happens if a tag does not actually exist in the dies set? Is it ignored?
- Include a section about significant alternatives that we have discussed. Definitely why a single list of tags is insufficient if that's the case. But also some of the other patterns.

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,56 @@
+# Nimble Wafer Map Die Highlighting
+
+## Problem Statement
+
+The wafer map receives an array of strings as the values that will be highlighted.
+
+```
```
rajsite, line 7: make sure here and in the examples below to mark the code fence with `ts` so that it has syntax highlighting, i.e.
````
```ts
@observable public highlightedValues: string[] = [];
```
````

which renders as
```ts
@observable public highlightedValues: string[] = [];
```

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
Razvan1928: Hi Milan, I tried to address all your topics inside the spec, you can take a look

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
Razvan1928: Good point, I left a comment inside the spec specifying that we need to changes the tests and also the storybook.

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,230 @@
...
+dies[0][0] will not be highlighted because it does not contain 'b' neither 'ab'
+
+dies[1][0] will be highlighted because it contains 'b'
+
+dies[0][1] will not be be highlighted because it does not have any tag
+
+If the highlightedTags equals
+
+```ts
+[''];
```
munteannatan, line 141: i think not using the highliting feature should be signaled by an empty array ```[]``` and not an empty string

munteannatan (approved):

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,230 @@
...
+
+If the highlightedTags equals
+
+```ts
+['efg'];
+```
+
+All dies will be faded
+
+Note: Besides the code changes we will also need to update the tests and nimble story book.
```
rajsite, line 154: Is the tag empty string a valid tag for a wafer to define?
```
{
        x: 0,
        y: 0,
        value: '14',
        tags: ['']
    },
```

Is the tag empty string a valid tag for highlightedTags to define?
```
['']
```

I think it would be fine to do so. No reason to add complexity to block it. We should just have tests for that case. Can you make a section of expected test cases to cover?

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Alternative Designs
+
+1. Same as the main one but instead of using a strings array for highlighted tags use an matrix of strings. This will give us more flexibility and more highlight capabilities but it will also require more time and will increase the complexity of complexity of the highlighting. In this case:
```
rajsite, line 158: The numbered list does not render in markdown well. You would need to indent all the paragraphs for them to show up in the list. CCan youu instead make these different sub headings?

```
## Alternative Designs

### Highlighted tags with a matrix of strings

Same as the main one but instead of using a strings...

### Add an isHighlighted field to dies

Introduce a new logic of highlighting...
```

packages/nimble-components/src/wafer-map/specs/features/change-highlight.md
rajsite: Once you and @munteannatan agree the expected test cases section covers the edge cases needed feel free to merge. We'll reference those cases when reviewing tests in future submissions.

rajsite:
You don't need to wait for @mollykreis for this PR

PR #1679, opened 2023-11-20
Add expand/collapse button to TableRow

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #861

This PR puts the visual and reporting mechanisms in place for the `TableRow` to support hierarchical data in the `Table`.

## 👩‍💻 Implementation

The main hurdle in the implementation was providing the right margins/paddings for various visual elements for when the row was either at the very top level visually, where the button is visually outside of the column cells, _and_ when the button is visually within the area of the column content (any parent row that is not at the top level visually).

Created a new shared style for the expand-collapse button that we could possibly share amongst other components like `TreeItem`.

Created new localizable label APIs for the expand-collapse button, as we should show a different label than what is shown on the `GroupRow`.

Adding `aria-hidden=true` to the expand-collapse button in both the `Row` and `GroupRow` templates. This change will likely be temporary for the `Row` until we  introduce the [table keyboard interactions](https://github.com/ni/nimble/issues/1137).

## 🧪 Testing

Added unit tests to verify the visibility of the expand-collapse button in various component states. Added unit tests for clicking expand-collapse button raising expected event.

NOTE: Styling changes will not be captured by any Chromatic tests for this PR. These changes will be best captured by tests for the `Table` where it can represent multiple rows at once to demonstrate the various expected indentation levels.

## ✅ Checklist


packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -39,10 +70,13 @@ export const template = html<TableRow>`
...
                         :recordId="${(_, c) => c.parent.recordId}"
                         ?has-action-menu="${x => !!x.actionMenuSlot}"
+                        ?expanded="${(_, c) => c.parent.expanded}"
+                        row-hierarchy-level="${(_, c) => c.parent.rowHierarchyLevel}"
+                        ?is-first-cell="${(_, c) => c.index === 0}"
                         action-menu-label="${x => x.actionMenuLabel}"
                         @cell-action-menu-beforetoggle="${(x, c) => c.parent.onCellActionMenuBeforeToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>, x)}"
                         @cell-action-menu-toggle="${(x, c) => c.parent.onCellActionMenuToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>, x)}"
-                        :nestingLevel="${(_, c) => c.parent.cellIndentLevels[c.index]};"
+                        :nestingLevel="${(_, c) => c.parent.cellIndentLevels[c.index]}"
```
m-akinc, line 79: Can we use consistent terminology for "nesting level" vs "indent level"? Is there a meaningful distinction?

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -221,6 +223,46 @@ describe('TableRow', () => {
...
+        it('shows expand-collapse button when rowHierarchyLevel is "topLevelParent"', async () => {
+            const pageObject = new TableRowPageObject(element);
+            await connect();
+            element.rowHierarchyLevel = TableRowHierarchyLevel.topLevelParent;
+            await waitForUpdatesAsync();
+
+            expect(pageObject.getExpandCollapseButton()).toBeDefined();
+        });
+
+        it('hides expand-collapse button when rowHierarchyLevel is "parent"', async () => {
```
m-akinc, line 236: Should add test cases for `rowHierarchyLevel`s `undefined` and `leaf`.

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -24,11 +43,23 @@ export const template = html<TableRow>`
                 `)}
             </span>
         `)}
-        ${'' /* This is needed to help align the cell widths exactly with the column headers, due to the space reserved for
-                the collapse-all button in the header. */}
         <span class="row-front-spacer"></span>
+        ${when(x => x.rowHierarchyLevel === 'topLevelParent', html<TableRow>`
+            <${buttonTag}
+                    appearance="${ButtonAppearance.ghost}"
```
m-akinc, line 49: nit: reduce indentation

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -1,14 +1,33 @@
...
+    mediumPadding
+} from '../../../theme-provider/design-tokens';
 
 // prettier-ignore
 export const template = html<TableRow>`
-    <template role="row" aria-selected=${x => x.ariaSelected}>
+    <template 
+        role="row"
+        aria-selected=${x => x.ariaSelected}
+        style="--ni-private-table-row-spacer-width: ${x => (x.nestingLevel > 0 || (x.nestingLevel === 0 && !(x.rowHierarchyLevel === 'parent' || x.rowHierarchyLevel === 'topLevelParent'))
```
m-akinc, line 27: This is very confusing to me. Ideally (i.e. if we can stop conditionally puting the expand/collapse button in the cell), I would expect this to be something like `x => x.nestingLevel * controlHeight.getValueFor(x)`. But at the very least, I think this would be a slight improvement: `x => (x.nestingLevel > 0 || x.rowHierarchyLevel === undefined || x.rowHierarchyLevel === 'leaf') ? ...`.

m-akinc (commented):
My primary feedback is that I really hope we don't have to support the expand/collapse button in both the row and the first cell.

packages/nimble-components/src/table/types.ts
```diff
@@ -148,6 +157,17 @@ export interface TableColumnConfiguration {
     pixelWidth?: number;
 }
 
+/**
+ * @internal
+ */
+export const TableRowHierarchyLevel = {
+    topLevelParent: 'topLevelParent',
+    parent: 'parent',
+    leaf: 'leaf'
```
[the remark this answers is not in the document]
atmgrifter00: I believe the way I implemented that case (not in this PR) was that it was `undefined` (which ultimately is treated in the same way as `leaf` I believe).

packages/nimble-components/src/table/types.ts
```diff
@@ -123,6 +123,15 @@ export interface TableRowSelectionEventDetail {
     selectedRecordIds: string[];
 }
 
+/**
+ * Event detail type for row toggle events in the table.
+ */
+export interface TableRowExpandToggleEventDetail {
+    oldState: boolean;
```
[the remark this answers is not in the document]
atmgrifter00: I believe `oldState` and `newState` are basically naming conventions for these fields in the details of similar events. At least I don't see enough of a context difference between these events and other toggle events.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -64,6 +88,35 @@ export class TableCell<
...
+        this.$emit('row-expand-toggle', expandEventDetail);
+        event.stopImmediatePropagation();
+        // To avoid a visual glitch with improper expand/collapse icons performing an
+        // animation, we apply a class to the appropriate group row such that we can have
+        // a more targeted CSS animation. We use the 'transitionend' event to remove the
+        // temporary class and register a function reference as the handler to avoid issues
+        // that may result from the 'transitionend' event not firing, as it will never result
+        // in multiple event listeners being registered.
+        this.animationClass = 'animating';
+        this.expandIcon?.addEventListener(
```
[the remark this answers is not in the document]
atmgrifter00: While I don't mind making your suggested change, processing the DOM update needed to apply the class is not a synchronous operation with the actual setting of that attribute, so I'm pretty sure this is safe.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -64,6 +88,35 @@ export class TableCell<
...
+            'transitionend',
+            this.removeAnimatingClass
+        );
+    }
+
+    private readonly removeAnimatingClass = (): void => {
```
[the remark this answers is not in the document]
atmgrifter00: No, it's because we need to access `this` in the implementation of the handler. If you tried to pass a member function as the handler for an event listener, you would get a linter error.

packages/nimble-components/src/table/types.ts
```diff
@@ -148,6 +157,17 @@ export interface TableColumnConfiguration {
     pixelWidth?: number;
 }
 
+/**
+ * @internal
+ */
+export const TableRowHierarchyLevel = {
+    topLevelParent: 'topLevelParent',
```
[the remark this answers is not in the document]
atmgrifter00: Refactoring got rid of this completely so hopefully there is no more confusion here.

packages/nimble-components/src/table/types.ts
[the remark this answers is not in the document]
atmgrifter00: Refactoring removed this.

packages/nimble-components/src/table/types.ts
```diff
@@ -148,6 +157,17 @@ export interface TableColumnConfiguration {
     pixelWidth?: number;
 }
 
+/**
+ * @internal
+ */
+export const TableRowHierarchyLevel = {
```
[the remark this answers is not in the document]
atmgrifter00: Removed.

packages/nimble-components/src/table/types.ts
```diff
@@ -163,4 +183,5 @@ export interface TableRowState<TData extends TableRecord = TableRecord> {
     nestingLevel?: number;
     leafItemCount?: number;
     groupColumn?: TableColumn;
+    rowType?: TableRowHierarchyLevel;
```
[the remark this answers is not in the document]
atmgrifter00: Removed.

packages/nimble-components/src/table/components/cell/template.ts
```diff
@@ -6,11 +6,32 @@ import {
...
 
 // prettier-ignore
 export const template = html<TableCell>`
-    <template role="cell" style="--ni-private-table-cell-nesting-level: ${x => x.nestingLevel}">
+    <template role="cell"
+        class="${x => (x.rowHierarchyLevel !== 'parent' ? 'no-padding' : '')}"
+        style="--ni-private-table-cell-nesting-level: ${x => x.nestingLevel};"
+    >
+        ${when(x => x.rowHierarchyLevel === 'parent' && x.isFirstCell, html<TableCell>`
+            <${buttonTag}
```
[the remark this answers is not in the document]
atmgrifter00: It is indeed possible! I just had to manage padding/margins of various elements in different ways, but I think it's resulted in obvious reduced complexity.

packages/nimble-components/src/table/components/cell/tests/table-cell.spec.ts
```diff
@@ -79,4 +85,72 @@ describe('TableCell', () => {
...
+            expectedVisibleState: true
+        },
+        {
+            name: 'expand-collapse button is hidden when rowHierarchyLevel is undefined and isFirstCell is true',
+            rowHierarchyLevel: undefined,
+            isFirstCell: true,
+            expectedVisibleState: false
+        },
+        {
+            name: 'expand-collapse button is hidden when rowHierarchyLevel is "child" and isFirstCell is true',
```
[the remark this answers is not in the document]
atmgrifter00: Removed.

packages/nimble-components/src/table/components/cell/tests/table-cell.spec.ts
```diff
@@ -79,4 +85,72 @@ describe('TableCell', () => {
...
+            rowHierarchyLevel: TableRowHierarchyLevel.leaf,
+            isFirstCell: true,
+            expectedVisibleState: false
+        },
+        {
+            name: 'expand-collapse button is hidden when isFirstCell is false',
+            rowHierarchyLevel: undefined,
```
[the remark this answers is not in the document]
atmgrifter00: This is all gone.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -205,16 +253,27 @@ export class TableRow<
...
         this.updateCellIndentLevels();
     }
 
+    private isParentRowChanged(): void {
+        this.updateCellIndentLevels();
+    }
+
+    private isTopLevelRowChanged(): void {
+        this.updateCellIndentLevels();
+    }
```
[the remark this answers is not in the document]
atmgrifter00: Removed.

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
atmgrifter00: Not after refactor.

packages/nimble-components/src/table/components/row/template.ts
atmgrifter00: Didn't need this after refactor.

packages/nimble-components/src/table/components/cell/tests/table-cell.spec.ts
```diff
@@ -79,4 +85,72 @@ describe('TableCell', () => {
...
+            isFirstCell: false,
+            expectedVisibleState: false
+        }
+    ];
```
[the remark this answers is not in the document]
atmgrifter00: Removed.

packages/nimble-components/src/table/components/row/template.ts
atmgrifter00: Will revisit in follow-up work.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -64,6 +88,35 @@ export class TableCell<
...
+        const expandEventDetail: TableRowExpandToggleEventDetail = {
+            oldState: this.expanded,
+            newState: !this.expanded,
+            recordId: this.recordId!
+        };
+        this.$emit('row-expand-toggle', expandEventDetail);
+        event.stopImmediatePropagation();
+        // To avoid a visual glitch with improper expand/collapse icons performing an
+        // animation, we apply a class to the appropriate group row such that we can have
+        // a more targeted CSS animation. We use the 'transitionend' event to remove the
```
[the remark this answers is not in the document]
atmgrifter00: Updated comment.

packages/nimble-components/src/table/components/row/template.ts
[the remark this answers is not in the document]
atmgrifter00: Will revisit in follow-up work.

packages/nimble-components/src/table/components/cell/styles.ts
```diff
@@ -28,6 +28,13 @@ export const styles = css`
         align-items: center;
     }
 
+    :host(.nested-parent-row-first-cell) .cell-view {
+        padding-left: calc(
+            ${mediumPadding} + ${controlHeight} + ${controlHeight} *
+                (var(--ni-private-table-cell-nesting-level) - 1)
```
atmgrifter00, line 34: The subtraction of 1 here is essentially where the logic moved from the `updateCellIndentLevels` method in the `Row` class. It's possible that the `updateCellIndentLevels` method can now be removed entirely, but that is something I'd rather explore outside of this PR.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -56,6 +57,9 @@ export class TableRow<
     @attr({ attribute: 'hide-selection', mode: 'boolean' })
     public hideSelection = false;
 
+    @attr({ mode: 'boolean' })
+    public expanded = false;
```
m-akinc, line 61: Where does this get set to `true`?

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -205,18 +252,20 @@ export class TableRow<
...
+        this.updateCellIndentLevels();
     }
 
     private nestingLevelChanged(): void {
         this.updateCellIndentLevels();
     }
 
+    private isParentRowChanged(): void {
+        this.updateCellIndentLevels();
+    }
```
m-akinc, line 264: Is this needed? `updateCellIndentLevels` doesn't use `isParentRow`.

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -205,18 +252,20 @@ export class TableRow<
 
     private dataRecordChanged(): void {
         this.updateCellStates();
+        this.updateCellIndentLevels();
```
m-akinc, line 255: Is this needed? `updateCellIndentLevels` doesn't use `dataRecord`.

packages/nimble-components/src/table/components/row/index.ts
atmgrifter00: Ultimately, this will be set from the `Table` template, but I was going to leave that for the next PR I put together.

packages/nimble-components/src/table/components/row/index.ts
atmgrifter00: I think you are right. For some reason I thought I was running into issues with the indentations not updating as expected, but I can't seem to replicate that, so I've removed these changes.

packages/nimble-components/src/table/components/row/index.ts
atmgrifter00: Nope, I think you're right. Removed.

packages/nimble-components/src/table/components/cell/styles.ts
```diff
@@ -28,6 +28,13 @@ export const styles = css`
         align-items: center;
     }
 
+    :host(.nested-parent-row-first-cell) .cell-view {
+        padding-left: calc(
```
m-akinc, line 32: I'm not clear on how all these paddings work. 

1. Under various circumstances, there are left paddings on the cell, the cell view, the expand button, and there's a left margin on the cell container. Also, there's a front row spacer. I don't think any of these are overlaying each other, but it also doesn't seem like these would make sense all in series horizontally.
2. It seems like `controlHeight` is the width of the indentation (of a child row under its parent), but I don't understand why `1` is being subtracted from  `--ni-private-table-cell-nesting-level` here and in the row styles.

Naively, I would expect the left padding/margin to be on just one of the elements (or maybe no padding, but the the row front spacer's width would account for it). There's likely a reason for doing things this way, but it's not obvious to me. I just wanted to bring this up in case you can think of a way to simplify it or maybe just give some insight in a comment somewhere.

m-akinc (approved):

packages/nimble-components/src/patterns/expand-collapse/styles.ts
```diff
@@ -0,0 +1,34 @@
...
+    mediumDelay
+} from '../../theme-provider/design-tokens';
+
+export const styles = css`
+    :host {
+        --ni-private-expand-collapse-button-indent-level: 0px;
+    }
+
+    :host .animating {
+        transition: ${mediumDelay} ease-in-out;
```
jattasNI, line 13: How closely should we align this with [the tree item styles](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/tree-item/styles.ts)? Some differences I noticed 
1. tree uses `ease-in`, not `ease-in-out`
2. tree applies a `180deg` rotation while this uses `0deg` (haven't dug into how that manifests) 
3. tree supports rtl but this doesn't
4. similar but not identical names and purposes of private CSS variables

I don't think they necessarily have to align, especially if the DOM structure is not the same. But it'd be worth a pass to see if the tree is doing anything that this should copy. I doubt they'd share a pattern but maybe they could.

packages/nimble-components/src/table/components/cell/styles.ts
atmgrifter00: @m-akinc and I talked through this off-line. In summary, we think it works as needed and is clear enough particularly with the caveat that I was able to address his 2nd concern (I had dumb math in my CSS).

If others have similar concerns we can discuss the best way to resolve them, be it an alternative strategy or some form of inline docs.

packages/nimble-components/src/table/components/cell/index.ts
```diff
@@ -32,6 +32,12 @@ export class TableCell<
     @observable
     public recordId?: string;
 
+    @attr({ attribute: 'is-parent-row', mode: 'boolean' })
+    public isParentRow = false;
```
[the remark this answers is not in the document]
jattasNI: If we go that direction (I personally agree) then `isInParentRow` might be even better since it keeps the "is..." naming pattern for boolean properties.

packages/nimble-components/src/table/components/group-row/template.ts
```diff
@@ -39,9 +39,9 @@ export const template = html<TableGroupRow>`
                 class="expand-collapse-button"
                 tabindex="-1"
                 title="${x => (x.expanded ? tableGroupCollapseLabel.getValueFor(x) : tableGroupExpandLabel.getValueFor(x))}"
+                aria-hidden="true"
```
jattasNI, line 42: What's the rationale for `aria-hidden="true"`? Could you add a note to the PR description? Is there an accessible interaction or does that need to wait for #1137?

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
```diff
@@ -221,6 +222,48 @@ describe('TableRow', () => {
...
+            await waitForUpdatesAsync();
+            const expandCollapseButton = pageObject.getExpandCollapseButton();
+
+            const listener = createEventListener(element, 'row-expand-toggle');
+            expandCollapseButton!.click();
+            await listener.promise;
+
+            expect(listener.spy).toHaveBeenCalledTimes(1);
+            const expectedDetails: TableRowExpandToggleEventDetail = {
+                newState: true,
```
jattasNI, line 260: Can you assert the value of `element.expanded` before and after clicking the button too?

jattasNI (approved):

packages/nimble-components/src/table/components/row/tests/table-row.spec.ts
atmgrifter00: Unfortunately, no, at least as a unit test isolated to the `Row` component. This is admittedly a goofy concession with how the API ultimately works (which is also how the `GroupRow` works as well). The `expanded` API is controlled externally to the `Row` instance based on Tanstack-provided state. When we click the button, the toggle event is ultimately handled by the `Table` which then updates Tanstack, which, in turn, provides a new set of row state that is then set on the `Row` instances, which would include an updated value for the `expanded` state.

packages/nimble-components/src/table/components/group-row/template.ts
atmgrifter00: The rationale is simply that these buttons are not currently focusable and should thus be hidden from the accessibility tree. As I noted in the new description, this will be temporary for the expand-collapse button in the `Row` component due to the issue you linked.

I'll note that as a result of this change the Accessibility score in Lighthouse drops a couple of points (to 82), seemingly because the `button` element that is within the shadow root of our `Button` element (via the FAST template) still has a non-negative `tabindex`. Through experimentation, if you force that `button` element to have a `tabindex` of -1, the Lighthouse score improves. That said, it's not clear what the right approach to set the `tabindex` on that `button` element is (nor if that's actually what we should be doing).

packages/nimble-components/src/patterns/expand-collapse/styles.ts
atmgrifter00: Yes, as I mention in the description, sharing this style with the `TreeItem` would certainly be ideal, but there are some differences that prevent that outright. That said, we can certainly address some of the differences you outline (though I would prefer to do that as a separate PR).

rajsite (approved):
Did a fairly high-level review and didn't see anything concerning. Looks good!

packages/nimble-components/src/table/components/group-row/template.ts
atmgrifter00: I've update the [linked issue](https://github.com/ni/nimble/issues/1137) with this information.

packages/nimble-components/src/patterns/expand-collapse/styles.ts
atmgrifter00: I've updated to use `ease-in` for consistency with `Treee`. For now, I don't think there's any other sharing to be done, or consistency to enforce.

packages/nimble-components/src/patterns/expand-collapse/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+        transform: rotate(0deg);
+    }
+
+    :host([expanded]) .expander-icon {
+        transform: rotate(90deg);
+    }
+
+    @media (prefers-reduced-motion) {
+        :host .animating {
+            transition-duration: 0s;
```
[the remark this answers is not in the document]
atmgrifter00: Nope, we use `0s` everywhere in our styles for setting `transition-duration`.

PR #1683, opened 2023-11-22
Rich text | Support for @mention in Editor

aagash-ni:
# Pull Request

## 🤨 Rationale

[User Story 2536827](https://dev.azure.com/ni/DevCentral/_workitems/edit/2536827): FE | Add @ mention support in Nimble rich-text-editor and in markdown serializer

## 👩‍💻 Implementation

- Added `@tiptap/extension-mention` extension for mention support and configured for user mention support.
- Added `mention` event which will be emitted from the respective `RichTextMention` configuration element when the user is typing `@` in the editor.
- Added support for `RichTextMentionUsers` configuration element in the editor.
- Added `getMentionedHrefs()` method in the configuration element for the client to get the mention list id (`href`) which already exists in the editor/viewer
   
    -  In editor the `hrefs` will be generated from the `RichTextMarkdownSerializer` as it gets dynamically updated.
    -  In the viewer from `RichTextMarkdownParser`, 'hrefs' will be generated as it will be updated while parsing.

- Configured mention tag and decoration tag (Tag added when user typing `@` ) as `nimble-rich-text-mention-users-view` element for mention Node.
- Created abstract base class `RichText` for rich text components (editor and viewer). Common methods have been relocated to this base class, which is then extended by the editor and viewer classes. (As the viewer and editor have the common functionality for getting the configuration from `RichTextMentionUsers` configuration elements, moved it to a base class).
- Added `MentionNode` in `RichTextMarkdownSerializer` to serialize the mentioned node. (markdown format is same as of auto-link)
- Update markdown-parser to add support for mention node in the editor
- Handled copy-paste functionality for mention in the editor. Currently, when mentioned node is copied and pasted into the editor it will render it as plain text. For example, copying `@user` and pasting it into the editor will render it as `user`. The `@` character is stripped off as the text node gets generated only based on the `label` attribute of `rich-text-mention-users-view` while transform pasting. Since in the user view element, the `@` character is added on the template view, the label is missing the character.
- Modified the serializer from being static to having an instance. This change enables dynamic initialization of the serializer when new mentions are added, allowing for the serialization of the respective mention nodes.

## 🧪 Testing

- Added tests to check the dynamic loading of child elements into the editor component
- Added tests for parser and serializer `getMentionedHrefs` logic to check whether it is returning the existing mention `href` list.
- Added storybook doc (`rich-text-editor.mdx`) and linked mention configuration docs as well.
- Added tests for checking the copy/paste behavior of mention nodes.

## ✅ Checklist


packages/nimble-components/src/rich-text-mention/users/view/template.ts
```diff
@@ -5,6 +5,8 @@ import type { RichTextMentionUsersView } from '.';
 // that arises when the mention user view is activated at the end of a line in the rich text editor while in edit mode.
 // This can be removed when the below issue is resolved
 // See: https://github.com/ni/nimble/issues/1659
-export const template = html<RichTextMentionUsersView>`<span class="control"
+export const template = html<RichTextMentionUsersView>`<span
+        class="control"
+        part="control"
```
aagash-ni, line 10: For disabling view element when editor get disabled, query selecting all the mention view node and setting the property `disabled` to true actually made it to disabled state. Still, we tried finding some other ways to update the `disabled` dynamically but didn't get any. So, we just modified it in CSS by accessing it through part in editor styles. We also went through the [comments ](https://github.com/ni/nimble/pull/1653#discussion_r1386960217)that Milan mentioned not to expose the `part` unnecessarily.

Please let us know your thoughts on this, shall we go with the `css` approach or through `queryselecting` one?

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -213,6 +215,12 @@ export const styles = css`
         pointer-events: none;
     }
 
+    :host([disabled]) ${richTextMentionUsersViewTag}::part(control) {
```
aagash-ni, line 218: https://github.com/ni/nimble/pull/1683/files#r1402554313

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -389,4 +389,57 @@ describe('RichTextViewer', () => {
             ).toEqual(['John Doe', 'Mary Wilson']);
         });
     });
+
+    describe('getMentionedHref() for viewer mentions', () => {
```
m-akinc, line 393:
```suggestion
    describe('getMentionedHrefs() for viewer mentions', () => {
```

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -389,4 +389,57 @@ describe('RichTextViewer', () => {
...
+        beforeEach(async () => {
+            ({ element, connect, disconnect } = await setup());
+            pageObject = new RichTextViewerPageObject(element);
+            await connect();
+        });
+
+        afterEach(async () => {
+            await disconnect();
+        });
```
m-akinc, line 402: I just noticed that every `describe` block in this file has its own identical `beforeEach` and `afterEach`. Just add the call to `connect()` in the top-level `beforeEach`, then add the `afterEach` with `disconnect()` at the top level. Then you can delete all these duplicates in the `describe` blocks. As it is now, `setup()` is getting called and the page object is being created two times per test.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -389,4 +389,57 @@ describe('RichTextViewer', () => {
...
+
+        it('getMentionedHref() method should return the mentioned href when it valid mention configuration matching the pattern to mention node', async () => {
```
m-akinc, line 404: `it` refers to the thing specified in the `describe` string, i.e. "getMentionHref() for viewer mentions". So you don't need to put "getMentionedHref() method" in each of these names. The intent is for the test cases to read like, "`it` should return..."

m-akinc (changes requested):
I haven't finished reviewing. I still have to look at the test cases, but this should be the majority of my feedback.

packages/nimble-components/src/rich-text-mention/users/view/template.ts
aagash-ni: I have updated the suggested approach by keeping new CSS variable and toggling it on editor disabling. But anyway, keeping this open to know all your thoughts on this.

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -55,10 +57,38 @@ export abstract class RichTextMention<
...
+     * Implement a generator method responsible for generating hrefs for existing mentions in the editor/viewer.
+     * The corresponding function implementation should reside within the rich text components.
+     */
+    public getMentionedHrefGenerator: HrefGenerator = () => [];
+
+    /**
+     * @public
+     * Returns hrefs for existing mentions in the editor/viewer.
+     */
+    public getMentionedHref(): string[] {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -0,0 +1,102 @@
...
+                    return userHref;
+                };
+            }
+        });
+        this.updateView();
+    }
+
+    protected abstract updateView(): void;
+
+    protected abstract getMentionedUser(): string[];
```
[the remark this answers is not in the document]
aagash-ni: Updated it to `getMentionedHrefs`

packages/nimble-components/src/rich-text/base/index.ts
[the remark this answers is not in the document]
aagash-ni: I have updated few to protected.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -109,6 +114,9 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
         // Tiptap [isEmpty](https://tiptap.dev/api/editor#is-empty) returns false even if the editor has only whitespace.
         // However, the expectation is to return true if the editor is empty or contains only whitespace.
         // Hence, by retrieving the current text content using Tiptap state docs and then trimming the string to determine whether it is empty or not.
+        if (this.tiptapEditor.state.doc.toString().includes('mention')) {
+            return this.tiptapEditor.isEmpty;
+        }
         return this.tiptapEditor.state.doc.textContent.trim().length === 0;
```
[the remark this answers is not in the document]
aagash-ni: I think this will not work, as before `this.tiptapEditor.isEmpty;` will return true even if there is only white spaces, So to overcome that we went for `this.tiptapEditor.state.doc.textContent.trim().length === 0;`, which will trim white spaces. But in mention case, If only mention tags are available inside the editor `this.tiptapEditor.state.doc.textContent.trim().length === 0;` is returning empty as true by ignoring mention tags. So to achieve this case , added the if block which will simply return empty based on the existence of mention tag in it.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -9,12 +9,20 @@ import type { Node } from 'prosemirror-model';
...
+    private static mentionedUsers: string[];
+
     private static readonly markdownSerializer = this.initializeMarkdownSerializerForTipTap();
 
     public static serializeDOMToMarkdown(doc: Node): string {
+        RichTextMarkdownSerializer.mentionedUsers = [];
         return this.markdownSerializer.serialize(doc);
     }
 
+    public static getMentionedUser(doc: Node): string[] {
```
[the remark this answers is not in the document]
aagash-ni: Updated to `getMentionHrefs`

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -498,11 +518,91 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+            },
+            addAttributes() {
+                return {
+                    href: {
+                        default: null,
+                        parseHTML: element => element.getAttribute('mention-href'),
+                        renderHTML: attributes => {
+                            if (!attributes.href) {
+                                return {};
+                            }
```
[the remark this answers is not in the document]
aagash-ni: I have checked the behaviors after removing this check, working as expected. Removed the check.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -498,11 +518,91 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                    },
+
+                    label: {
+                        default: null,
+                        parseHTML: element => element.getAttribute('mention-label'),
+                        renderHTML: attributes => {
+                            if (!attributes.label) {
+                                return {};
+                            }
+
```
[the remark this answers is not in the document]
aagash-ni: Removed this as it is working as expected after removal.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -498,11 +518,91 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                    };
+                }
+            }
+        });
+    }
+
+    private updateUserLists(props: SuggestionProps): void {
+        const validUserMentionElement = this.mentionElements.find(
+            mention => mention.mentionInternals.validConfiguration
+                && mention.mentionInternals.character === '@'
```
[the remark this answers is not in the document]

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -498,11 +518,91 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                return [
+                    richTextMentionUsersViewTag,
+                    mergeAttributes(
+                        this.options.HTMLAttributes,
+                        HTMLAttributes,
+                        { 'disable-editing': true }
+                    )
+                ];
+            }
+        }).configure({
```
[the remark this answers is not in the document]
aagash-ni: Moved the link configuration inside `getCustomLinkExtension` method. As for mention extension, there might be different configuration for other mentions in future, so keeping it inside to avoid confusion while setting it to tip tap editor.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -498,11 +518,91 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+                        onUpdate: (props): void => {
+                            this.updateUserLists(props);
+                        }
+                    };
+                }
+            }
+        });
+    }
+
+    private updateUserLists(props: SuggestionProps): void {
```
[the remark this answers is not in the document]

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -25,6 +26,10 @@ export class RichTextMarkdownParser {
      */
     private static mentionConfigs?: MarkdownParserMentionConfiguration[];
 
+    public static getMentionedUser(): string[] {
```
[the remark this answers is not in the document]
aagash-ni: Updated it to `getMentionedHrefs`

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -55,10 +57,38 @@ export abstract class RichTextMention<
...
+    public onMention(filter: string): void {
+        this.$emit('mention', { filter });
+    }
+
+    /**
+     * @public
+     */
     public get validity(): RichTextMentionValidity {
         return this.validator.getValidity();
     }
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -191,6 +192,16 @@ export class RichTextEditorPageObject {
         );
     }
 
+    public getEditorMentionViewAttributeValues(attribute: string): string[] {
+        return Array.from(
+            this.getTiptapEditor()!.querySelectorAll(
+                richTextMentionUsersViewTag
+            )
+        )
+            .filter((el, _) => el.children.length === 0)
```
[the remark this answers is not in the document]
aagash-ni: Removed this filtering as this don't have any effect.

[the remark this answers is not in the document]
aagash-ni: Removed https://github.com/ni/nimble/pull/1683#discussion_r1402754290

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+<DocsStory of={editorStories.richTextEditor} expanded={false} />
+<Controls of={editorStories.richTextEditor} />
+
+# Usage Docs
+
+## User Mention Usage
+
+To enable support for `@mention` in the `nimble-rich-text-editor` component, include the specified configuration and mapping components as children. These components facilitate the conversion of input markdown into a mention node, displaying a user's name. Additionally, ensure that each user mentioned in the markdown input has a corresponding user mapping. For more details, see [Client Usage Guidance on Filtered Users](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text/specs/mention-hld.md#client-usage-guidance-on-filtered-users-1).
+
```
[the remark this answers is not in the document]
aagash-ni: `Client Usage Guidance on Filtered Users` cannot moved to the mention user's description as it is different for viewer and editor. In editor, client need to dynamically populate the mappings based on the `mention-update` event along with the already mentioned users. But in viewer just populating the already mentioned users alone is enough. So, we have separate client guidance for editor and viewer.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1585,6 +1636,28 @@ describe('RichTextEditor', () => {
                 '1. ***Numbered list with bold and italics***\n\n* ***Bulleted list with bold and italics***'
             );
         });
+
+        it('Should return same markdown when mention markdown string is assigned with valid comfiguration element', async () => {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1997,6 +2082,335 @@ describe('RichTextEditor', () => {
...
+        beforeEach(async () => {
+            ({ element, connect, disconnect } = await setup());
+            pageObject = new RichTextEditorPageObject(element);
+            await connect();
+        });
+
+        afterEach(async () => {
+            await disconnect();
+        });
+
```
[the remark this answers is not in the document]
aagash-ni: Removed

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
aagash-ni: Removed `beforeEach` and `afterEach` other than top level places.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -9,12 +9,20 @@ import type { Node } from 'prosemirror-model';
...
+        RichTextMarkdownSerializer.serializeDOMToMarkdown(doc);
```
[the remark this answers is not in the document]
aagash-ni: For viewer we will be utilizing getMentionedHrefs from parser as it has static markdown, and it won't get updated once parsed. So, the href will be taken from the parser itself.  But for editor, the state will be dynamically loaded, Consider this case where markdown we set can have one user and other users can be dynamically added to editor through UI, to get all the existing mentionedHrefs from the state, we need to get the list from the serializer. So whenever clients call `getMentionedhrefs()` method from configuration, In editor it will again serialize the current editor document and send us the mentioned hrefs. 

We can't move `mentionedUsers = []` into `getMentionedUser` method as we need to empty the array on each serialization, if not the array get populated on each `getMarkdown()` method call. Until the client didn't call `getMentionedUser` for the array to be updated. So to empty the array, we just kept under `serializeDOMToMarkdown()`.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -120,7 +169,9 @@ describe('RichTextEditor', () => {
         ).toBeTrue();
     });
 
-    it('clicking buttons in the slot element should call the click event once', () => {
+    it('clicking buttons in the slot element should call the click event once', async () => {
+        ({ element, connect } = await setupWithFooter());
+        await connect();
```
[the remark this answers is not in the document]
aagash-ni: Moved it to a separate describe block

packages/nimble-components/src/rich-text-mention/base/index.ts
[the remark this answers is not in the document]
aagash-ni: Updated

package-lock.json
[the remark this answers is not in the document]
aagash-ni: Added change file.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
m-akinc: > if not the array get populated on each `getMarkdown()` method call.

Items are only added to that array if they do not already exist in it. Plus, it will only grow excessively if old mentions keep being deleted (but the array would still contain them), while new, different mentions keep being added. So there isn't really a memory/performance issue with moving `mentionedUsers = []`. And since nothing other than `getMentionedUsers` actually cares about the array being up-to-date, there's no functional issue either. So for the sake of clarity, I recommend puting both the reset and the return of the array in `getMentionedUsers`.

m-akinc (approved):

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -11,6 +11,7 @@ import type { MarkdownParserMentionConfiguration } from './markdown-parser-menti
  * Provides markdown parser for rich text components
  */
 export class RichTextMarkdownParser {
+    private static mentionedHrefs: string[];
```
jattasNI, line 14: We should have test coverage for the changes to this file in the dedicated test spec for this file. (Note that the specific tests might change depending on what API we decide on for getting the mentioned hrefs; see my comment elsewhere).

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -25,6 +26,10 @@ export class RichTextMarkdownParser {
      */
     private static mentionConfigs?: MarkdownParserMentionConfiguration[];
 
+    public static getMentionedHrefs(): string[] {
```
jattasNI, line 29: I may be misreading something, but isn't this a case of a non-static method on the component calling a static method on the parser? How does this work? Doesn't the parser need context like the markdown and the mention config (similar to what's being passed to `parseMarkdownToDOM?` below)?

edit: I see that the parser is caching the mentioned hrefs from the most recent markdown that was parsed. And then I guess it's up to the caller component to retrieve those cached mentions before someone else calls parse. This seems really brittle. If there are multiple viewer components on the same page and the client waits until all of the viewers have been parsed before calling `getMentionedHrefs`, I think they'll all return the mentions from whichever viewer was parsed last.

Can we come up with a different pattern (e.g. returning the mentioned users from the parse method)?

jattasNI (commented):
Initial batch of comments, haven't reviewed serializer or editor yet

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
jattasNI: > In the parser, calling getMentionedUsers just returns whatever users were collected during the last parse. Should we follow that same pattern here?

FYI I think the parser implementation is wrong so we shouldn't ncessarily copy it here. See [my comment on the parser](https://github.com/ni/nimble/pull/1683#discussion_r1406924013).

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -9,12 +9,20 @@ import type { Node } from 'prosemirror-model';
  * Provides markdown serializer for rich text components
  */
 export class RichTextMarkdownSerializer {
+    private static mentionedHrefs: string[] = [];
```
jattasNI, line 12: Similar comment to the parser: we should have tests in the spec for this file to cover the changes to this file.

rajsite:
Unfortunately was not able to review today, expect to be able to review tomorrow

packages/nimble-components/src/rich-text/editor/index.ts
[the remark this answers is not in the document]
aagash-ni: > by adding `|| this.tiptapEditor.state.doc.textContent.trim().length === 0`.

Actually for the mention tags text content are considered as two white spaces, So the above text content will have only two white spaces in case of mention tags, If we are conditioning with `OR` it will return true as the `textcontent.trim().length` will return empty. 

> isEmpty actually returns true

`isEmpty` will return false if there are any mention nodes inside. So we are checking the condition of mention node existence in the editor document as `isEmpty` will return false for the mention nodes.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
```diff
@@ -44,6 +47,21 @@ const dataSets = {
     [exampleDataType.markdownString]: richTextMarkdownString
 } as const;
 
+type MentionDataType = (typeof mentionDataType)[keyof typeof mentionDataType];
```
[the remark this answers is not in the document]
aagash-ni: > 1. hooking up an action so you can see the event data in the actions tab

Updated it in the actions => handles in editor story.

> 2. listing the event name in the arguments table and describing when it fires

Added the description for `mention-update` event in configuration element args.

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
```diff
@@ -53,6 +53,11 @@ export const richTextMentionUsers: StoryObj = {
         },
         validity: {
             description: mappingUserValidityDescription
+        },
+        getMentionedHrefs: {
+            name: 'getMentionedHrefs()',
+            description:
+                'Returns mentioned users list in the current state of the rich text components.'
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -419,4 +419,47 @@ describe('RichTextViewer', () => {
             ).toEqual(['John Doe', 'Mary Wilson']);
         });
     });
+
+    describe('getMentionedHref() for viewer mentions', () => {
```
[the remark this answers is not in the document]
aagash-ni: Added cases to check the matching hrefs when having different types of mentions.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -419,4 +419,47 @@ describe('RichTextViewer', () => {
             ).toEqual(['John Doe', 'Mary Wilson']);
         });
     });
+
```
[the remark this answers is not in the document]
aagash-ni: Added cases to check the `getMentionedHrefs()` for pattern and dynamic configuration change.

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -209,6 +218,14 @@ export class RichTextEditorPageObject {
         return getTagsFromElement(this.getTiptapEditor() as HTMLElement);
     }
 
+    // Return list of tags, excluding those such as 'IMG' (prosemirror-separator) that do not affect the UI or markdown output.
+    // These tags are considered extraneous and are added by prosemirror.
+    public getMarkdownImpactTagNames(): string[] {
```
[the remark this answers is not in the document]
aagash-ni: Updated it to `getMarkdownRenderedTagNames()`

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -63,73 +43,18 @@ export class RichTextViewer extends FoundationElement {
...
-                )
-            );
-
-        this.updateView();
+    public getMentionedHrefs(): string[] {
+        return RichTextMarkdownParser.getMentionedHrefs();
     }
 
-    private updateView(): void {
+    protected updateView(): void {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
aagash-ni: > sake of clarity, I recommend puting both the reset and the return of the array in getMentionedUsers
 
 Moved it to `getMentionedUsers()`.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
aagash-ni: In serializer as all the tests are happening by interactions, Will be adding cases in popup PR as it will have logics for adding mention through interactions. Anyway, I have added cases in editor spec to check the behaviors.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
jattasNI: As we discussed on the call this is probably an acceptable approach for now but we should verify how often `getMentionedUsers()` will be called and how long it takes to return. And we should file a tech debt item to see if we can get the same info from tiptap's tree without reserializing to markdown.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
jattasNI: As we discussed in the call, let's update `parseMarkdownToDOM()` to return an object that contains both the document fragment and the list of mentioned users. You'd follow similar patterns for storing/clearing the static data as we use already in this class. Then the viewer can cache the result, return it from its `mentionedHrefs` accessor, and update the cached result every time the markdown is set.

jattasNI (commented):
As discussed, I'll be OOO so defer to others to approve the PR.

packages/nimble-components/src/rich-text/models/markdown-parser.ts
aagash-ni: Updated the parser logic. Now `parseMarkdownToDOM()` will return object having `fragment` and `mentionedHrefs`, Will be caching that in viewer and returning it on `getMentionedHrefs()` method.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
aagash-ni: Created tech debt https://github.com/ni/nimble/issues/1693 to track this approach.

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        // For example, having two `nimble-rich-text-mention-users` within the children of rich text viewer or editor is an invalid configuration
+        this.mentionConfig = [];
+        this.mentionElements.forEach(mention => {
+            if (mention.mentionInternals.validConfiguration) {
+                const markdownParserMentionConfiguration = new MarkdownParserMentionConfiguration(
+                    mention.mentionInternals
+                );
+                this.mentionConfig.push(markdownParserMentionConfiguration);
+            }
+        });
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            this.updateMentionConfig();
+        }
+    }
+
+    /**
+     * @internal
+     */
+    public abstract getMentionedHrefs(): string[];
+
+    protected abstract updateView(): void;
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -33,18 +39,25 @@ export class RichTextMarkdownParser {
...
-            );
+            return {
+                fragment: this.domSerializer.serializeFragment(
+                    parsedMarkdownContent.content
+                ),
+                mentionedHrefs: RichTextMarkdownParser.mentionedHrefs
+            };
         } finally {
             this.mentionConfigs = undefined;
+            RichTextMarkdownParser.mentionedHrefs = [];
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -495,15 +511,100 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+            autolink: true,
+            openOnClick: false,
+            // linkOnPaste can be enabled when hyperlink support added
+            // See: https://github.com/ni/nimble/issues/1527
+            linkOnPaste: false,
+            validate: href => this.validAbsoluteLinkRegex.test(href)
         });
     }
 
+    private getCustomUserMentionExtension(): Node<MentionOptions> {
```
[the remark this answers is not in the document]

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -209,6 +218,14 @@ export class RichTextEditorPageObject {
         return getTagsFromElement(this.getTiptapEditor() as HTMLElement);
     }
 
+    // Return list of tags, excluding those such as 'IMG' (prosemirror-separator) that do not affect the UI or markdown output.
+    // These tags are considered extraneous and are added by prosemirror.
+    public getMarkdownRenderedTagNames(): string[] {
+        return getTagsFromElement(this.getTiptapEditor() as HTMLElement).filter(
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/markdown-parser.ts
```diff
@@ -33,18 +39,25 @@ export class RichTextMarkdownParser {
     public static parseMarkdownToDOM(
         value: string,
         markdownParserMentionConfig?: MarkdownParserMentionConfiguration[]
-    ): HTMLElement | DocumentFragment {
+    ): ParseResult {
         try {
-            this.mentionConfigs = markdownParserMentionConfig;
+            RichTextMarkdownParser.mentionConfigs = markdownParserMentionConfig;
```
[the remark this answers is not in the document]
aagash-ni: Added `setup()` and `cleanup()` method.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -35,13 +50,21 @@ export class RichTextMarkdownSerializer {
             });
         };
 
+        const mentionNode = function mention(
+            state: MarkdownSerializerState,
+            node: Node
+        ): void {
+            const href = node.attrs.href as string;
+            state.write(`<${href}>`);
```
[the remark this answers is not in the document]
aagash-ni: While looking into the `prosemirror-markdown`, there is a method named `esc(str: string, startOfLine = false)` which will be escaping the special markdown stuffs. 

https://github.com/ProseMirror/prosemirror-markdown/blob/3e5a5d02f7c1b9336744740149767ee1dc4e954d/src/to_markdown.ts#L405C2-L416C4. 

Kindly share your thoughts on this.

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @aagash-ni There were several things I think we should update to make the update codepath clearer and some refactors. I captured them here: https://github.com/ni/nimble/pull/1699/files

I didn't figure out yet what is causing the test failures but I think it's a direction worth pursuing. Could you see if we could merge those changes in this PR and if you have ideas of why the tests are failing?

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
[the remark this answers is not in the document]
aagash-ni: The built-in auto-link feature itself does not escape the link. Also, we aren't able to use the auto-link serializer because the `link` is mark but in tiptap `mention` is a node. So we can't use the existing logic. Instead, we wrote our logic which is very similar to the absolute link serialization. 

Added the reference code as you suggested.

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
aagash-ni: Updated the changes.

packages/nimble-components/src/utilities/tests/rich-text-markdown-string.ts
```diff
@@ -1 +1,2 @@
-export const richTextMarkdownString = 'Supported rich text formatting options:\n1. **Bold**\n2. *Italics*\n3. Numbered lists\n   1. Option 1\n   2. Option 2\n4. Bulleted lists\n   * Option 1\n   * Option 2\n5. Absolute link: <https://nimble.ni.dev/>\n 6. @mention:\n    1. User pattern: <user:1>\n    2. HTTPS pattern: <https://user/2>';
+export const richTextMarkdownString = '<user:1>';
```
jattasNI, line 1: Looks like some temporary testing code snuck in here.

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
```diff
@@ -1243,14 +1331,73 @@ describe('Markdown parser', () => {
...
+                [
+                    new MarkdownParserMentionConfiguration(
+                        element.mentionInternals
+                    )
+                ]
+            );
+            expect(parserContent.mentionedHrefs).toEqual([]);
+        });
+
+        it('should return empty mentioned href list called without markdownParserMentionConfig', async () => {
```
jattasNI, line 1390: This test seems weird to me: it's not actually using the element that it creates. If we want to test the parser standalone then you could delete the element code. If we want it to be an integration test following the same pattern as the other tests in the file then we should instead configure the element to have no mention config.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -86,6 +86,7 @@ export class RichTextMentionValidator<
         );
     }
 
+    // TODO move this to child class like done for table
```
jattasNI, line 89: What's the plan for addressing this TODO? An in-work part of this PR or a follow up?

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -144,7 +145,8 @@ export class RichTextMentionValidator<
 
     private isInvalidRegex(pattern: string): boolean {
         try {
-            RegExp(pattern);
+            // eslint-disable-next-line no-new
```
jattasNI, line 148: ESLint disables should have an explanation justifying them (or even better, be fixed not to violate the rule). In this case if the goal is to see whether the constructor throws then you could add a comment saying that.

packages/nimble-components/src/rich-text-mention/users/index.ts
```diff
@@ -24,10 +21,19 @@ declare global {
 /**
  * Rich Text user mention configuration element which will have MappingMentionUser elements as children
  */
-export class RichTextMentionUsers extends RichTextMention<
-RichTextMentionConfig,
-RichTextMentionUsersValidator
-> {
+export class RichTextMentionUsers extends RichTextMention<RichTextMentionUsersValidator> {
+    // TODO this isn't a great way to do this. It's mixing configuration element state and parent state which are decoupled asynchronously from each other.
```
jattasNI, line 25: I'm guessing this was a TODO based on @rajsite's contribution this weekend. It sounds like there is some more work here. Is that happening in this PR or tracked elsewhere?

jattasNI (commented):
Started reviewing the latest updates but saw a few incomplete areas so I'm going to step back until they settle (I'm assuming there's still some back-and-forth with Milan going on). But I'm leaving some comments to track the issues I saw in my partial review.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
rajsite: That was a todo I added. I had updated the pattern in the table to make the child classes do the supportedMappingElements check for some typing reasons. 

It's a small change if we can go ahead and do it to have alignment in the pattern. See here: https://github.com/ni/nimble/pull/1656/files#diff-3566b942a10cfb5f154b500ba99f0cd6362cd60a5247840f499a37a8d12f3220R47

Where:
- we stop passing `supportedMappingElements` to the base class
- the child class just has it's own implementation `validateMappingTypes`

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -209,14 +211,35 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+        next: EditorConfiguration
+    ): void {
+        if (this.isOnlyMentionInternalsChanged(prev, next)) {
+            this.setMarkdown(this.getMarkdown());
+        } else {
+            const currentStateMarkdown = this.getMarkdown();
+            this.richTextMarkdownSerializer = new RichTextMarkdownSerializer(
+                this.configuration?.mentionExtensionConfig.map(
+                    config => config.name
+                ) ?? []
```
rajsite, line 236: All the generics added to `Configuration<>` can be avoided and we should remove.
You just need a check here like:
```
this.configuration instanceof EditorConfiguration ? this.configuration.mentionExtensionConfig.map(config => config.name) : []
```

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -335,175 +387,29 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
-            }
-        });
+    private initializeEditor(): void {
+        this.unbindEditorTransactionEvent();
+        this.unbindEditorUpdateEvent();
+        this.unbindNativeInputEvent();
+        this.tiptapEditor?.destroy();
+        this.tiptapEditor = createTiptapEditor(
+            this.editor,
+            this.configuration?.mentionExtensionConfig ?? []
```
rajsite, line 397: Similar to above, we can get rid of the generics complexity by just adding a check:

```
this.configuration instanceof EditorConfiguration ? this.configuration.mentionExtensionConfig : []
```

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,877 @@
...
+                'P',
+                `${richTextMentionUsersViewTag}`.toUpperCase(),
+                'BR'
+            ]);
+            expect(
+                pageObject.getEditorMentionViewAttributeValues('mention-label')
+            ).toEqual(['1']);
+        });
+
+        // TODO: Once the rich text validator added for duplicate configuration elements, below test case should be updated
```
rajsite, line 151: Can you link to a PR or issue associated with this TODO?

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -326,6 +349,35 @@ export class RichTextEditor extends FoundationElement implements ErrorPattern {
...
+            }
+        });
+        return Array.from(mentionedHrefs);
+    }
+
+    protected override createConfig(): EditorConfiguration {
+        return new EditorConfiguration(this.mentionElements);
+    }
+
+    private isOnlyMentionInternalsChanged(
```
rajsite, line 366: Isn't this more like `isMentionExtensionConfigUnchanged`

At this point it's not about mentionInternals only the configuration objects.

packages/nimble-components/src/rich-text/models/configuration.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Base class for Configuration, contains mention configuration required for RichTextMarkdownParser
+ */
+export class Configuration {
+    public parserMentionConfig: MarkdownParserMentionConfiguration[] = [];
+
+    protected isValid = true;
+
+    public constructor(mentionElements: RichTextMention[]) {
+        this.isValid = mentionElements.every(
+            mention => mention.mentionInternals.validConfiguration
```
rajsite, line 14: So many similar things called mention, lets be moree explicit

```suggestion
            mentionElement => mentionElement.mentionInternals.validConfiguration
```

packages/nimble-components/src/rich-text/models/configuration.ts
```diff
@@ -0,0 +1,25 @@
...
+        );
+
+        this.parserMentionConfig = this.isValid
+            ? mentionElements.map(
+                mention => new MarkdownParserMentionConfiguration(
+                    mention.mentionInternals
+                )
+            )
```
rajsite, line 22:
```suggestion
            ? mentionElements.map(
                mentionElement => new MarkdownParserMentionConfiguration(
                    mentionElement.mentionInternals
                )
            )
```

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -53,40 +52,64 @@ export abstract class RichTextMention<
...
-    ): TMentionConfig;
+    protected abstract createMentionConfig(): MentionConfig;
 
     protected abstract createMappingConfig(
         mapping: Mapping<unknown>
     ): MappingConfig;
 
+    protected get richTextParent(): RichText {
+        return this.parentElement as RichText;
+    }
```
rajsite, line 108: If we can get rid of getMentionedHrefs than we can remove this too

packages/nimble-components/src/utilities/tests/rich-text-markdown-string.ts
rajsite: woops, that was me too 😅 but yea can be cleaned up

rajsite (commented):
Implementation is looking pretty much done, after the final changes just need to make a pass on the tests

packages/nimble-components/src/rich-text/models/tests/markdown-parser.spec.ts
aagash-ni: Removed the element code.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: Moved `validateMappingTypes` to child class.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
aagash-ni: Added comment line

packages/nimble-components/src/rich-text-mention/users/index.ts
aagash-ni: Removed `getMentionedHrefs()` method from the configuration base.  The client will be getting all the mentionedHrefs from the editor/viewer.

packages/nimble-components/src/rich-text/editor/index.ts

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Added the PR link

packages/nimble-components/src/rich-text/models/configuration.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/configuration.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/index.ts
aagash-ni: Removed this as `getMentionedHrefs()` get removed from the base.

packages/nimble-components/src/rich-text/editor/index.ts
aagash-ni: Renamed

packages/nimble-components/src/rich-text/editor/index.ts

packages/nimble-components/src/utilities/tests/rich-text-markdown-string.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
```diff
@@ -25,6 +25,15 @@ The object's type is \`RichTextMentionValidity\`, and it contains the following
...
 
+const mentionUpdateEventDescription = `This event will be fired on following action:
+
+- Whenever the \`@\` character is entered into the editor
+- When the user types any character after \`@\` into the editor
+
+This fires with the \`eventData\` containing the current text that is added after the \`@\` character and before the current position of the
+    text cursor.
+`;
+
```
jattasNI, line 36: This description appears in the rich text viewer documentation too, but the actions described don't apply to the viewer. Could you update it to distinguish how the event behaves in both components? I'm not certain how that is but it could be something like:
```suggestion
const mentionUpdateEventDescription = `For the editor, this event will be fired on following action:

- Whenever the \`@\` character is entered into the editor
- When the user types any character after \`@\` into the editor

This fires with the \`eventData\` containing the current text that is added after the \`@\` character and before the current position of the
    text cursor.

For the viewer, this event will never be fired.
`;

```

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,831 @@
...
+        it('updating `display-name` in mapping mention should update the `mention-label` in view', async () => {
+            element.setMarkdown('<user:1>');
+            await appendUserMentionConfiguration(
+                element,
+                ['user:1'],
+                ['username1']
+            );
+
+            const renderedUserMention = element.lastElementChild as RichTextMentionUsers;
+            const renderedMappingUser = renderedUserMention.firstElementChild as MappingUser;
```
rajsite, line 267: We end up doing a lot of imperative dom lookups unnecessarily. Could you have appendUserMentionConfiguration return an object of what it created?
Something like:
```
{
    userMentionElement: RichTextMentionUsers,
    mappingElements: MappingUser[]
}
```
That way we can avoid all the `as` casting and the kind of confusing lastElementChild, firstElementChild lookups.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,831 @@
...
+              <${mappingUserTag} key="user:2" display-name="Mary Wilson"></${mappingUserTag}>
+          </${richTextMentionUsersTag}>
+      </${richTextEditorTag}>`
+    );
+}
+
+async function appendUserMentionConfiguration(
+    element: RichTextEditor,
+    userKeys?: string[],
+    displayNames?: string[]
```
rajsite, line 40: It's confusing reading the tests that these are sperate properties. Instead could it be:

interface MappingConfiguration {
    key?: string
    displayname?: string
}

appendUserMentionConfiguration(
    element: RichTextEditor,
    mappings? MappingConfiguration[]
)

That way it's clearer what the mapping objects will look like as the related properties are grouped together instead of in separate arrays:
```
appendUserMentionConfiguration(
                element,
                [{key: 'user:1', displayName: 'username1'},
                {key: 'user:2', displayName: 'username2'}]
            );
```

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,831 @@
...
+
+            expect(pageObject.getEditorTagNames()).toEqual(['P', 'A', 'A']);
+            expect(pageObject.getEditorLeafContents()).toEqual([
+                'user:1',
+                'test:2'
+            ]);
+        });
+    });
+
+    it('should fire "mention" event from configuration elment when there is @mention in editor', async () => {
```
jattasNI, line 646: We should have some more coverage to verify these cases:

1. `mention-update` is fired when appending to an existing mention
2. `mention-update` is fired when deleting from an existing mention
3. `mention-update` isn't fired when deleting the `@` of an existing mention
4. `mention-update` isn't fired when pasting a block of text that contains a mention in the middle
5. `mention-update` isn't fired for other interactions. For example, typing text near an existing mention (but not part of that mention)

Could you also update the names of these tests to use the actual name of the event (`mention-update`, not `mention`)?

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,831 @@
...
+            expect(
+                pageObject.getEditorMentionViewAttributeValues('mention-label')
+            ).toEqual(['username1', 'testname2']);
+            const renderedUserMention = element.lastElementChild as RichTextMentionTest;
+            const renderedMappingUser = renderedUserMention.firstElementChild as MappingUser;
+            renderedUserMention.removeChild(renderedMappingUser);
+            await waitForUpdatesAsync();
+            expect(pageObject.getMarkdownRenderedTagNames()).toEqual([
+                'P',
+                `${richTextMentionUsersViewTag}`.toUpperCase(),
```
rajsite, line 488: I'm not sure why the template insertion is needed, it could just be `richTextMentionUsersViewTag.toUpperCase()`

But even then can you just make a constant at the top of this file so this can be done one, like:
```
const RICH_TEXT_MENTION_USERS_VIEW_TAG = richTextMentionUsersViewTag.toUpperCase()
```

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,831 @@
...
+            await appendUserMentionConfiguration(
+                element,
+                ['user:1'],
+                ['username1']
+            );
+            await waitForUpdatesAsync();
+            const renderedUserMention = element.lastElementChild as RichTextMentionUsers;
+            expect(element.getMentionedHrefs()).toEqual(['user:1']);
+            element.removeChild(renderedUserMention);
+            expect(element.children.length).toBe(0);
```
jattasNI, line 723:
```suggestion
            expect(element.children.length).toBe(0);
            expect(element.getMentionedHrefs()).toEqual([]);
```

jattasNI (commented):
Getting close! If this round of feedback can be addressed smoothly then I think that's all I'll need to approve tomorrow.

packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -66,74 +48,28 @@ export class RichTextViewer extends FoundationElement {
...
-            .filter(mention => mention.mentionInternals.validConfiguration)
-            .map(
-                mention => new MarkdownParserMentionConfiguration(
-                    mention.mentionInternals
-                )
-            );
-
-        this.updateView();
+    public getMentionedHrefs(): string[] {
+        return this.mentionedHrefs;
```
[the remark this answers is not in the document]
rajsite: Can you add a test to viewer and editor to make sure getMentionedHrefs returns a new instance? (just to avoid a regression)

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -9,12 +11,18 @@ export interface MentionInternalsOptions {
...
  */
-export class MentionInternals<TMentionConfig> {
+export class MentionInternals {
     /**
-     * Configuration which will hold mentioning info, character, icon and pattern
+     * Mention-specific configuration. Can be used to, for example, pass mention-specific configuration to views.
      */
     @observable
-    public mentionConfig?: TMentionConfig;
+    public mentionConfig?: MentionConfig;
```
rajsite, line 19: The intention of these mentionConfig objects was for state to be passed all the way to the view elements. I.e. from here all the way to the rich text mention view.
We haven't actually implemented that plumbing or test these objects. It also has a bit of a confusing name since we have some many objects that sound similar to "mention config".

To simplify and avoid having dead code can you remove these MentionConfig and UserMentionConfig classes as well as its associated createMentionConfig objects? If we end up needing that kind of configurability in the future we can reintroduce it.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Updated as per your suggestion.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -419,4 +419,75 @@ describe('RichTextViewer', () => {
...
+            await appendUserMentionConfiguration(
+                element,
+                ['user:1'],
+                ['username1']
+            );
+            await waitForUpdatesAsync();
+            const renderedUserMention = element.lastElementChild as RichTextMentionUsers;
+            expect(element.getMentionedHrefs()).toEqual(['user:1']);
+            element.removeChild(renderedUserMention);
+            expect(element.children.length).toBe(0);
```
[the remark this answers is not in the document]
aagash-ni: Added

packages/nimble-components/src/rich-text-mention/users/tests/rich-text-mention-users.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/viewer/index.ts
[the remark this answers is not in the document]
aagash-ni: Added test case to check the instances. Without `Array.from` the instances are the same. Added it as per your suggestion.

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
aagash-ni: Removed `MentionConfig, UserMentionConfig and createMentionConfig()`

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Added

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Added test cases for these behaviours

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Updated

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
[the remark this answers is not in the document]
aagash-ni: Updated it to reload the placeholder on editor initialization.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+
+    it('should fire "mention-update" event from configuration elment when there is update in @mention in editor', async () => {
+        const { userMentionElement } = await appendUserMentionConfiguration(
+            element
+        );
+        const mentionUpdateSpy = jasmine.createSpy('mention-update');
+        userMentionElement.addEventListener('mention-update', mentionUpdateSpy);
+        await pageObject.setEditorTextContent('@test');
+        expect(mentionUpdateSpy).toHaveBeenCalledTimes(1);
+    });
```
jattasNI, line 611: I may be missing something, but these tests look identical other than their names. For the update test ideally you'd set the editor content to `@test`, then create the spy and add the event listener, then append/type text immediately after @test, and assert that the event fired

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+
+    it('should fire "mention-update" event when deleting an existing @mention in editor', async () => {
+        await pageObject.setEditorTextContent('@test');
+        const { userMentionElement } = await appendUserMentionConfiguration(
+            element
+        );
+        const mentionUpdateSpy = jasmine.createSpy('mention-update');
+        userMentionElement.addEventListener('mention-update', mentionUpdateSpy);
+        await pageObject.sliceEditorContent(2);
```
jattasNI, line 631: I don't think the implementation of `sliceEditorContent()` is correct for the intent of these tests. It basically behaves by manipulating the markdown in memory and then replacing the entire contents of the editor. This means these tests are more like paste tests. The test below which validates no event fired when `@` is removed is in fact validating that setting contents to a string with no `@` in it won't fire the event.

Instead I think `sliceEditorContent()` should set the focus at the end of the text and then press the Delete key a specified number of times. (The method might have a new name like `deleteEditorContent()`). Then these tests would be updated to assert that deleting the last character from `@test` does fire an event but deleting the last character from `@` doesn't fire an event.

rajsite (approved):

rajsite (approved):
Approved assuming Jesse's test pattern concerns are addressed

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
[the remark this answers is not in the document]
aagash-ni: Created a bug to track this behaviour https://github.com/ni/nimble/issues/1704.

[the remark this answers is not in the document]
aagash-ni: > it's showing a diff in the mention view story because the mention is now rendering as black instead of green.

Regarding this change, it is not because of the `disable-editing`, we have added a new CSS variable to control the color of mention view element when the editor gets disabled. Please take a look at the discussion [here ](https://github.com/ni/nimble/pull/1683#discussion_r1403301987). As of this change, the color will be set for the CSS variable only in the editor/viewer. So this is looking black in a standalone case but green when used along with the editor/viewer.

And for the change of mention view from bold to plain body font, We thought of updating it in the list box PR, But there is a decoration tag here as well right, So I have updated that font in this PR itself and updated the test.

> There are no diffs in the editor matrix tests which means we're not doing visual testing of a lot of this new work.

Updated this matrix test with the same markdown content as of viewer.

packages/nimble-components/src/rich-text/models/markdown-serializer.ts
```diff
@@ -5,17 +5,32 @@ import {
...
+        node: Node,
+        parent: Node,
+        index: number
+    ) => void;
+}
+
 /**
  * Provides markdown serializer for rich text components
  */
 export class RichTextMarkdownSerializer {
```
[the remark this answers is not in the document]
aagash-ni: Updated description

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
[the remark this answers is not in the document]
jattasNI: The matrix test still needs to be updated to increase the height of the editors. Right now you can't see all the text so the diff isn't exercising the mention view.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
jattasNI: Sounds good. I manually verified that the events fire correctly now, so we'll have to rely on that until you implement the integration test approach.

jattasNI (approved):

change/@ni-nimble-components-e1eca008-24b5-481f-a835-1e3157d4ac1b.json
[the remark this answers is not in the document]
jattasNI: Oops, I didn't see your latest comment when I edited the one above. Glad you were able to get it working.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+        expect(mentionUpdateSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not fire "mention-update" event when @ get removed in editor', async () => {
```
jattasNI, line 636:
```suggestion
    // TODO: Replace with integration test that uses keyboard events to trigger mention-update
    // https://github.com/ni/nimble/issues/1568
    it('should fire "mention-update" event when deleting an existing @mention in editor', async () => {
```

rajsite:
Restored the branch to make reviewing the mention-list PR against it easier

PR #1684, opened 2023-11-22
Make anchor behave like native anchor when contenteditable

### LINKED ISSUE #1502: Nimble anchor inside "contenteditable" div is clickable
# 🐛 Bug Report

When `nimble-anchor` is used inside a `contenteditable` div, the link is clickable, whereas the default behavior of the [native HTML anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) tag is not clickable.

_Note:_ Not sure if it is a bug or an expected behavior. Guide us in creating a correct issue for this behavior.

## 💻 Repro or Code Sample

A simple demo to reproduce: https://typescript-l63dbq.stackblitz.io

## 🤔 Expected Behavior

The default behavior of a `nimble-anchor` might be the same as that of a native HTML anchor inside a `contenteditable` div, meaning the link should not be clickable.
 
## 😯 Current Behavior

`nimble-anchor` is clickable and focusable inside `contenteditable` div.

## 🔦 Context

We are planning to render all links inside `nimble-rich-text-editor` as `nimble-anchor` as mentioned in this discussion: https://github.com/ni/nimble/issues/1372#issuecomment-1658622984. However, the rich text editor internally uses the `contenteditable` div to add rich text content to the editor. In this context, we should restrict the user from clicking any of the links that are being added to the editor.

## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser: Google Chrome
* Version: 116.0.5845.142

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1502

## 👩‍💻 Implementation

By design, the content-editable state is [not inherited](https://wicg.github.io/webcomponents/spec/shadow/#editing) across the shadow boundary, but that is really what we want. To mimic this, I introduced a new wrapper div* around the native anchor element and set its `contenteditable` attribute based on the value of `isContentEditable` of the host `nimble-anchor`. This approach works perfectly if the host's `isContentEditable` state never changes after being connected to the DOM (which is likely the vast majority of use cases). However, reacting to changes in the host's `isContentEditable` state is a problem. `isContentEditable` is a read-only property that is implemented by walking the ancestry chain until an elmement with `contenteditable` set is found. There is no practical way to detect changes to this value. Instead, I re-sync the content-editable state across the shadow boundary on `mouseenter` and `focus` events. This covers a couple important use cases:
- When interacting with the link via the mouse, it will update the content-editable state in time to allow/prevent clicks and show/hide the preview url at the bottom of the window.
- When the host becomes content-editable, tabbing to the link will immediately lose focus. Though the correct behavior would be to skip the link in the tab order entirely, dropping focus seems like the best we can do.

*Note that the new wrapper div is necessary, because if we instead set `contenteditable` directly on the native anchor, it behaves differently (than when it is contained by a content-editable element): it remains focusable. 

## 🧪 Testing

Manually tested in Storybook, and wrote new unit tests.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/anchor/index.ts
+++ b/packages/nimble-components/src/anchor/index.ts
@@ -34,6 +34,35 @@ export class Anchor extends AnchorBase {
      */
     @attr
     public appearance: AnchorAppearance;
+
+    /**
+     * @internal
+     */
+    public container?: HTMLElement;
+
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        this.updateContentEditable();
+    }
+
+    /**
+     * @internal
+     * We want our anchor to behave like a native anchor when in a content-editable
+     * region, i.e. not operable or focusable. The most reliable way to achieve this is
+     * to synchronize our shadow DOM's content-editable state with that of the host.
+     * We must look at the host's read-only isContentEditable property, which factors
+     * in any inherited value. Unfortunately, there is no way for us to detect when its
+     * value has changed, so the best we can do is to re-sync on specific events.
+     * This has shortcomings, e.g. if isContentEditable goes from true to false, our
+     * anchor will remain un-tabable until a mouseenter triggers a re-sync.
+     */
+    public updateContentEditable(): void {
+        if (this.isContentEditable) {
+            this.container!.setAttribute('contenteditable', '');
+        } else {
+            this.container!.removeAttribute('contenteditable');
+        }
+    }
 }
 
 // FoundationAnchor already applies the StartEnd mixin, so we don't need to do it here.
--- a/packages/nimble-components/src/anchor/template.ts
+++ b/packages/nimble-components/src/anchor/template.ts
@@ -9,7 +9,10 @@ import type { Anchor } from '.';
 export const template: FoundationElementTemplate<
 ViewTemplate<Anchor>,
 AnchorOptions
-> = (_context, definition) => html<Anchor>`<a
+> = (_context, definition) => html<Anchor>`<div
+        ${ref('container')}
+        class="top-container"
+    ><a
         class="control"
         part="control"
         download="${x => x.download}"
@@ -40,6 +43,8 @@ AnchorOptions
         aria-owns="${x => x.ariaOwns}"
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
+        @mouseenter="${x => x.updateContentEditable()}"
+        @focus="${x => x.updateContentEditable()}"
         ${ref('control')}
     >${
     /* Start and End slot templates inlined to avoid extra whitespace.
@@ -74,4 +79,4 @@ AnchorOptions
             @slotchange="${x => x.handleEndContentChange()}">
             ${definition.end || ''}
         </slot
-    ></span></a>`;
+    ></span></a></div>`;
--- a/change/@ni-nimble-components-be72d648-96c1-403d-80e8-2a68c3752207.json
+++ b/change/@ni-nimble-components-be72d648-96c1-403d-80e8-2a68c3752207.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Make anchor behave like native anchor when contenteditable",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor/styles.ts
+++ b/packages/nimble-components/src/anchor/styles.ts
@@ -20,6 +20,10 @@ export const styles = css`
         font: ${linkFont};
     }
 
+    .top-container {
+        display: contents;
+    }
+
     [part='start'] {
         display: none;
     }
--- a/packages/nimble-components/src/anchor/tests/anchor.spec.ts
+++ b/packages/nimble-components/src/anchor/tests/anchor.spec.ts
@@ -4,23 +4,7 @@ import { waitForUpdatesAsync } from '../../testing/async-helpers';
 import { fixture, Fixture } from '../../utilities/tests/fixture';
 import { getSpecTypeByNamedList } from '../../utilities/tests/parameterized';
 
-async function setup(): Promise<Fixture<Anchor>> {
-    return fixture<Anchor>(html`<nimble-anchor></nimble-anchor>`);
-}
-
 describe('Anchor', () => {
-    let element: Anchor;
-    let connect: () => Promise<void>;
-    let disconnect: () => Promise<void>;
-
-    beforeEach(async () => {
-        ({ element, connect, disconnect } = await setup());
-    });
-
-    afterEach(async () => {
-        await disconnect();
-    });
-
     it('should export its tag', () => {
         expect(anchorTag).toBe('nimble-anchor');
     });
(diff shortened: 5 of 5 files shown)


m-akinc:
@atmgrifter00 could I get a buddy on this?

packages/nimble-components/src/anchor/index.ts
```diff
@@ -34,6 +34,35 @@ export class Anchor extends AnchorBase {
      */
     @attr
     public appearance: AnchorAppearance;
+
+    /**
+     * @internal
+     */
+    public container?: HTMLElement;
```
atmgrifter00, line 41: You can just declare this as:
```
public container!: HTMLElement;
```
and thus avoid the unnecessary `!` checks later.

packages/nimble-components/src/anchor/index.ts
```diff
@@ -34,6 +34,35 @@ export class Anchor extends AnchorBase {
...
+     * We want our anchor to behave like a native anchor when in a content-editable
+     * region, i.e. not operable or focusable. The most reliable way to achieve this is
+     * to synchronize our shadow DOM's content-editable state with that of the host.
+     * We must look at the host's read-only isContentEditable property, which factors
+     * in any inherited value. Unfortunately, there is no way for us to detect when its
+     * value has changed, so the best we can do is to re-sync on specific events.
+     * This has shortcomings, e.g. if isContentEditable goes from true to false, our
+     * anchor will remain un-tabable until a mouseenter triggers a re-sync.
+     */
+    public updateContentEditable(): void {
```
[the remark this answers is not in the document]
m-akinc: Created https://github.com/microsoft/fast/issues/6870

packages/nimble-components/src/anchor/template.ts
```diff
@@ -40,6 +43,8 @@ AnchorOptions
         aria-owns="${x => x.ariaOwns}"
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
+        @mouseenter="${x => x.updateContentEditable()}"
+        @focus="${x => x.updateContentEditable()}"
```
rajsite, line 47: I dunno, this implementation feels like a whack-a-mole kind of edge case. An immediate concern is missing symmetric event handlers, mouseenter / mouseleave, focus/blur. But I don't have confidence that covers all interactions anyway.

I guess we should be considering how contenteditable applies to all of our components and this doesn't seem like a pattern that is easy to scale. At the same time this is really an edge case as far as I'm aware. I don't think 99% of usage of anchors cares about this but they all will have event listeners set-up to track this.

Can we just make this explicit instead? I'm pretty sure the rich-text-editor can set attributes it wants on the element it emits. Can it just emit the contenteditable attribute and we reflect it internally? Just make it a simple attribute reflection?

We can create usage notes saying due to browser limitations you have to explicitly set contenteditable on the nimble-anchor.

m-akinc: I like this very much. 👍

packages/nimble-components/src/anchor/index.ts
```diff
@@ -34,6 +34,18 @@ export class Anchor extends AnchorBase {
...
+     * @public
+     * @remarks
+     * HTML Attribute: contenteditable
+     *
+     * Ideally, proper support for contenteditable should come from FAST.
+     * I have filed bug https://github.com/microsoft/fast/issues/6870 to them.
+     * If/when it is fixed, we can remove this workaround.
+     */
+    @attr({ mode: 'boolean' })
+    public contenteditable = false;
```
rajsite, line 48: FYI contentEditable prop and contenteditable attribute are native built-in behaviors:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/contentEditable
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable
Doesn't seem like that has been taken into account

m-akinc: I'm not totally sure what the right approach is. I could rename the attribute, so that we don't interfere with the built-in attribute/property, but I feel like that would be confusing. A better approach might be to use a notifier to watch the value of the built-in `contentEditable` property and reflect it to the native anchor. Would you agree with that?

rajsite: The issue right now is that they are just weirdly inconsistent:
- there is the built in contentEditable property that maps to the attribute contenteditable with an enum strings of 'true', 'false', 'plaintext-only'
- This PR is adding a property that is different named contenteditable that maps to the exisitng contenteditable attribute but is a boolean attribute instead of an enum of strings supported by the native one

> If I just change the type and name of my declared attribute/property to match the built-in ones, would that avoid any potential issues?

I'm not sure did you try it and see if there are issues? Might need tests that show everything is sane on the host and child when manipulating either the property or the attribute. It's essentially what was done for `lang` right?

```
@attr('contenteditable')
contentEditable: AnEnumTypeForContentEditableValues!;
```

There might be some minor differences that would be caught in usage notes or something. I expect the main one would be that when unset the `contentEditable` property returns `undefined` instead of the browser default value. But I didn't try it to see what happens.

packages/nimble-components/src/anchor/index.ts
```diff
@@ -34,6 +34,19 @@ export class Anchor extends AnchorBase {
...
+     * @remarks
+     * HTML Attribute: contenteditable
+     * See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable
+     *
+     * Ideally, proper support for contenteditable should come from FAST.
+     * I have filed bug https://github.com/microsoft/fast/issues/6870 to them.
+     * If/when it is fixed, we can remove this workaround.
+     */
+    @attr({ attribute: 'contenteditable' })
+    public override contentEditable = 'inherit';
```
rajsite, line 49: This causes it to always be reflected to the DOM, which is unneeded since this is an edge case

```suggestion
    public override contentEditable!;
```

packages/nimble-blazor/NimbleBlazor/Components/NimbleAnchor.razor.cs
```diff
@@ -10,6 +10,12 @@ public partial class NimbleAnchor : NimbleAnchorBase
     [Parameter]
     public bool? UnderlineHidden { get; set; }
 
+    /// <summary>
+    /// Whether the anchor should behave like it is in an editable region.
+    /// </summary>
+    [Parameter]
+    public bool? ContentEditable { get; set; }
```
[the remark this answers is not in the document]
rajsite: > My impression is that we can take more liberties with the types in the Blazor wrappers 

I prefer to not have arbitrary inconsistencies between all the different framework wrappers. The goal someday would be to generate as much as possible and this would not be a high value exception to that case with custom logic.

Also this is a browser native API, not one we are inventing. You can use all the other HTML elements in a Blazor App. Ours should not have arbitrary differences. They are HTML Elements and this is a standard HTML Element API.

rajsite (approved):

packages/nimble-components/src/anchor/tests/anchor.stories.ts
```diff
@@ -56,12 +58,19 @@ const metadata: Meta<AnchorArgs> = {
             control: { type: 'radio' },
             description:
                 'Set to `prominent` to make the anchor appear in a different color than normal text.'
+        },
+        contenteditable: {
+            options: ['false', 'true'],
+            control: { type: 'radio' },
+            description:
+                'Set this to the string "true" (or set the attribute without any value)  when the anchor is within an editable region (i.e. element/hierarchy with [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable)). Whereas native elements inherit their `contenteditable` value by default, the `nimble-anchor` requires this attribute be explicitly set. When set, the anchor cannot be focused or operated.'
```
rajsite, line 66: So when I set this to true in storybook I expect to be able to click inside and edit the text but nothing seems to happen on chrome on windows: https://60e89457a987cf003efc0a5b-csyyuqlmje.chromatic.com/?path=/story/components-anchor--anchor&args=label:link;underlineHidden:!false;contenteditable:true;appearance:default

rajsite: Looks like contenteditable needs to be set on the span around the link. Then I can use arrow keys to navigate onto the link and edit its contents.

rajsite (approved):

msmithNI (approved):

rajsite:
Just a thought that occurred. In the future if we find ourselves having to do more with contenteditable we should treat it like lang and dir as a DesignToken that has to be configured on theme provider / DesignToken for nimble to be aware.

jattasNI (approved):

PR #1686, opened 2023-11-23
Angular support for @mention configuration and user mapping

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

[Task 2566806](https://dev.azure.com/ni/DevCentral/_workitems/edit/2566806): PR2 | Angular integration for configuration and mapping element

## 👩‍💻 Implementation

- Added directives for `nimble-rich-text-mention-users` and supporting element `nimble-mapping-user`. Both of these follow the same folder structure as in `nimble-components`
- Added tests for new directives
- Added rich text mention users to example client app

## 🧪 Testing

Written tests for both the directives

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1694, opened 2023-11-29
Rich text | Nimble mention list box (popup) and Integration in editor 

aagash-ni:
# Pull Request

## 🤨 Rationale

[User Story 2583344](https://dev.azure.com/ni/DevCentral/_workitems/edit/2583344): FE | Nimble mention listbox and integration in editor

## 👩‍💻 Implementation

- Generated configuration to list the items in the popup.
- Created `nimble-rich-text-mention-list-box` element which will be handling filtering logic and keydown handlers for the users list.
- Added Anchored region inside the editor template holding the `nimble-rich-text-mention-list-box` within it.
- Updated suggestion configuration in mention extension to process following things:
    - Toggling the popup (Anchored region) based on the interaction.
    - Updated the position of the anchored region relative to decoration tag.
    - Forwarding the keydown handlers to mention list box as the focus should retain on editor.
    - Updating the list items inside the mention-list-box when configuration is getting modified.

## 🧪 Testing
Yet to be added.

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/index.ts
m-akinc, line 1: My initial, high-level feedback is that the `RichTextEditor` feels too polluted with stuff related to mentioning. I suspect it might be cleaner and easier to understand if we pushed as much of that stuff as possible down into `RichTextMentionListBox`. I think you could go as far as moving the anchored region into the `RichTextMentionListBox`'s template. I'm imagining the interface between the `RichTextEditor` and the `RichTextMentionListBox` might result in something like this:
```TS
protected override updateMentionConfig(): void {
    super.updateMentionConfig();
    this.mentionListBox.updateMentionConfig(this.mentionElements);
}

...

private getCustomUserMentionExtension(): Node<MentionOptions> {
...
    suggestion: {
        ...
        render: () => {
            return {
                onStart: (props): void => {
                    this.mentionListBox.onMention(props, this.mentionElements);
                },
```
I haven't thought it all the way though, but I think that general direction would be an improvement.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private normalizeString(value: string): string {
+        // This implementation is based on the below reference.
+        // https://claritydev.net/blog/diacritic-insensitive-string-comparison-javascript
+        return value
+            .normalize('NFD')
+            .replace(/[\u0300-\u036f]/g, '')
+            .toLowerCase();
+    }
```
rajsite, line 174: I'm concerned about this getting out of sync with external usages and potentially even with server-side search in other features.

See discussion thread here: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/600443#1701308051

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -339,10 +386,44 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+            ) {
+                this.mentionInternalsMap.set(
+                    element.mentionInternals.character,
+                    element.mentionInternals
+                );
+            }
+            this.activeConfiguration = this.activeChar
+                ? this.mentionInternalsMap.get(this.activeChar)?.mentionConfig
+                    ?.mappingConfigs
+                : undefined;
```
rajsite, line 417: I think this mentionsInternalsMap stuff is a symptom of Mapping Extension not being configured correctly as discussed here: https://github.com/ni/nimble/pull/1683#discussion_r1410031640

As described in the comment, similar to the Markdown Parser, the TipTap Extension should not be dealing with mentionInternals, etc.

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +87,31 @@ export const template = html<RichTextEditor>`
...
+                        appearance="ghost"
+                        content-hidden
+                        ?disabled="${x => x.richTextEditor.disabled}"
+                        slot="start"
+                        title=${x => x.label}
+                        @click=${x => x.richTextEditor.mentionButtonClick(x.character)}
+                        @keydown=${(x, c) => x.richTextEditor.mentionButtonKeyDown(c.event as KeyboardEvent, x.character)}
+                    >
+                        ${x => x.label}
+                        ${x => x.getIconTemplate()}
```
suseendran-ni, line 102: Here we tried to render it directly in the template as below
`<${x => x.icon} slot="start"></${x => x.icon}>`.

But this does not work as expected. The element does not get rendered properly. 
image

Hence we fetched the template from the `getIconTemplate()` method. Please let me know if there is a better approach to handle this.

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +87,31 @@ export const template = html<RichTextEditor>`
...
                     </${toggleButtonTag}>
+                    ${repeat(
+        x => Array.from(x.mentionExtensionConfig ?? []),
+        html<MentionExtensionConfiguration>`<${buttonTag}
+                        appearance="ghost"
+                        content-hidden
+                        ?disabled="${x => x.richTextEditor.disabled}"
+                        slot="start"
+                        title=${x => x.label}
+                        @click=${x => x.richTextEditor.mentionButtonClick(x.character)}
```
suseendran-ni, line 98: Since the functionality of the `mention` buttons is to insert the respective characters into the editor, we kept the methods in the `RichTextEditor` class.

However, we are not sure of how to access the richTextEditor instance within this `repeat` block (as the instance is mapped to the `MentionExtensionConfiguration` type). Hence, we cached the `richTextEditor` instance within the `MentionExtensionConfiguration` object and bound the methods respectively.

Please suggest if there are better ways to handle this scenario.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    @observable
+    private anchorElement?: HTMLElement;
+
+    private hasAnySelectableOption = false;
+
+    private mentionExtensionConfig?: MentionExtensionConfiguration[];
+
+    private suggestionProps!: SuggestionProps;
+
+    private readonly intersectionObserver: IntersectionObserver = new IntersectionObserver(
```
suseendran-ni, line 84: We have introduced intersection observer here to close the mention popup, when the anchor element is hidden.

This is based on the issue that we faced with the `menu-button` in the comments functionality. The issue is, when we open the Menu and scroll through the comments list, the opened `menu button` appears on top of the entire page.
image

More details: https://teams.microsoft.com/l/message/19:0c59ac4a4cc1438894cb42f4b21703f9@thread.skype/1692631178768?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=41626d4a-3f1f-49e2-abdc-f590be4a329d&parentMessageId=1692631178768&teamName=ASW%20SystemLink&channelName=UI%20Working%20Group&createdTime=1692631178768

We suspect the same issue with the `mention-list` as well. Hence we handled it using the `intersection observer`.

Please let me know your thoughts.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+                        href: this.listBox.firstSelectedOption.value,
+                        displayName: this.listBox.firstSelectedOption.innerText
+                    } as MentionDetail);
+                    return true;
+                }
+                return false;
+            }
+            case keyArrowDown: {
+                if (!event.shiftKey && this.hasAnySelectableOption) {
+                    this.listBox.selectNextOption();
```
m-akinc, line 136: Both `selectNextOption()` and `selectPreviousOption()` are [documented as internal](https://github.com/microsoft/fast/blob/fd65254531ac3a593b32528d81b44ebf82c8e4ab/packages/web-components/fast-foundation/src/listbox/listbox.ts#L501). Maybe instead of adding all this logic ourselves, we can call `super.keydownHandler` and then handle just tab, enter, and escape?

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    private setActiveCharacter(activeCharacter: string): void {
+        this.activeCharacter = activeCharacter;
+    }
+
+    private setAnchorElement(anchorElement: HTMLElement): void {
+        this.anchorElement = anchorElement;
+    }
```
m-akinc, line 239: Seems like we should just inline these two functions, since they're each only called once.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    private filterOptions(): void {
+        if (!this.childItems || this.filter === undefined) {
+            return;
+        }
+        const normalizedFilter = normalizeString(this.filter);
+        this.childItems.forEach(element => {
+            const listOption = element as ListOption;
+            const normalizedText = normalizeString(listOption.text);
```
m-akinc, line 256: This seems like it could be slow. Every time the user types another character, the filter changes and we call `filterOptions`, which normalizes every list option. If there are dozens or hundreds of users, it might cause a noticeable delay. We should try this with lots of users in the list and see how it performs. If it's too slow, maybe we cache the list of normalized values, or we reconsider whether it is necessary to do a normalized match.

packages/nimble-components/src/rich-text/mention-list-box/styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from '@microsoft/fast-element';
+import { display } from '@microsoft/fast-foundation';
+
+export const styles = css`
+    ${display('inline-flex')}
+
+    nimble-listbox {
+        max-height: 168px;
```
m-akinc, line 8: This should probably be a token. Or maybe a better option is to use something like `calc(5.5 * ${controlHeight})` to indicate that we want to show at most 5.5 options. (There was a recommendation from UX that we use X.5 so that it is immediately obvious that there are more items scrolled out of view.)

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        const captured = (e.target as HTMLElement).closest(
+            'option,[role=option]'
+        ) as ListOption | null;
+
+        if (!captured || captured.disabled) {
+            return false;
+        }
+        this.activateMention({
+            href: captured.value,
+            displayName: captured.innerText
```
m-akinc, line 174: Use `text` instead of `innerText`. It has cleaned up whitespace and internally uses `textContent`, which is more efficient than `innerText`.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        });
+        this.setOpen(false);
+    }
+
+    private setOpen(value: boolean): void {
+        this.open = value;
+    }
+
+    private updateSelectableOption(): void {
+        this.hasAnySelectableOption = !this.childItems.map(element => element as ListOption).every(
```
m-akinc, line 277: It seems like you might not need this. The implementation of `selectFirstOption` already [excludes disabled options](https://github.com/microsoft/fast/blob/fd65254531ac3a593b32528d81b44ebf82c8e4ab/packages/web-components/fast-foundation/src/listbox/listbox.ts#L483), and `firstSelectedOption` [returns `null` if there is no selected option.](https://github.com/microsoft/fast/blob/fd65254531ac3a593b32528d81b44ebf82c8e4ab/packages/web-components/fast-foundation/src/listbox/listbox.ts#L40) Because of that, `scrollOptionIntoView` is safe to call even if there are no selectable options. If you can do without `hasAnySelectableOption`, that would be nice.

Unfortunately, I just noticed that `firstSelectedOption` is documented as internal. To be safe, I would probably just copy the implementation, since it is so simple:
```TS
    public get firstSelectedOption(): ListOption {
        return this.selectedOptions[0] ?? null;
    }
```

m-akinc (approved):

packages/nimble-components/src/rich-text/mention-list-box/index.ts
m-akinc: If that's the pattern we've used before, it makes sense to me.

packages/nimble-components/src/rich-text/editor/index.ts
suseendran-ni: Moved the mention related functionalities within the `mentionListBox` component

packages/nimble-components/src/rich-text/editor/index.ts
suseendran-ni: As we have refactored the entire logic, I am closing this thread

packages/nimble-components/src/rich-text/editor/template.ts
suseendran-ni: Updated the logic as per the suggestion.

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -42,9 +43,15 @@ export class MentionInternals<TMentionConfig> {
      */
     public readonly viewElement: string;
 
+    /**
+     * Mention label to render while hovering the rich text icon
```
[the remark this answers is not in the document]
suseendran-ni: Updated as per the suggestion

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -394,13 +421,20 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
                 )
             );
+            this.mentionListBox?.updateMentionExtensionConfig(this.mentionExtensionConfig);
 
             return;
         }
+        this.resetMentionExtensionConfig();
+    }
+
+    private resetMentionExtensionConfig(): void {
```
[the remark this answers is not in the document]
suseendran-ni: The logic was refactored in another PR and rebased on top. Hence resolving this thread.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Updated the logic

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Updated the logic

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -163,7 +170,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
     public editorContainer!: HTMLDivElement;
 
     @observable
```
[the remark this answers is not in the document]
suseendran-ni: This code is refactored in the other PR and merged here. Hence closing this thread

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -658,26 +692,53 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
                         }
                     };
                 }
             }
         });
     }
 
-    private triggerMentionEvent(filter: string): void {
+    private onMention(props: SuggestionProps): void {
+        this.triggerMentionEvent(props);
```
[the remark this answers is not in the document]
suseendran-ni: Updated this code in a different PR and merged the changes here.

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -10,10 +10,13 @@ import {
     richTextToggleBoldLabel,
     richTextToggleItalicsLabel,
     richTextToggleBulletedListLabel,
-    richTextToggleNumberedListLabel
+    richTextToggleNumberedListLabel,
```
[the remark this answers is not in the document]
suseendran-ni: Resolved

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
```diff
@@ -1,24 +1,39 @@
...
+        this.mappingConfigs = mentionInternals.mentionConfig?.mappingConfigs;
+        this.icon = mentionInternals.icon;
+        this.label = mentionInternals.label;
+        this.richTextEditor = richTextEditor;
         this.name = key;
         this.key = key;
     }
+
+    public getIconTemplate(): ViewTemplate {
+        return html`<${this.icon} slot="start"></${this.icon}>`;
```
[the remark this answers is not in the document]
suseendran-ni: Updated as per the suggestion

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
```diff
@@ -108,6 +108,9 @@ const metadata: Meta<RichTextEditorArgs> = {
             <${mappingUserTag} key="${x => mentionDataSets[x.mentionData].href}3" display-name="Sue Ann"></${mappingUserTag}>
```
jattasNI, line 108: When you open the listbox in storybook it is clipped by the bounds of its container. We have a pattern to fix this which you can see in [these other stories](https://github.com/search?q=repo%3Ani/nimble%20disableStorybookZoomTransform&type=code)

image

image

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,309 @@
+import { Notifier, Observable, observable } from '@microsoft/fast-element';
```
jattasNI, line 1: The biggest architectural concern we have is how much custom code exists here. Ideally we'd share a lot more of it with other components like the select and combobox. Is this handwritten? Copied from FAST? Copied from other Nimble components?

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Updated as per the suggestion

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Updated as per the suggestion

packages/nimble-components/src/rich-text/mention-list-box/index.ts
m-akinc: Sorry, we should also use the `super.keydownHandler` to navigate on Home/End keys.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,324 @@
...
+            : undefined;
+    }
+
+    private filterOptions(): void {
+        if (!this.childItems || this.filter === undefined) {
+            return;
+        }
+        const normalizedFilter = normalizeString(this.filter.trim());
+        this.childItems.forEach(listOption => {
+            const normalizedOptionText = normalizeString(listOption.text.trim());
```
m-akinc, line 294: Don't need to trim `listOption.text`, because [it already does it.](https://github.com/microsoft/fast/blob/2ea80697bc3a5193e6123fb08ac3be2a76571aeb/packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts#L204)

packages/nimble-components/src/utilities/models/string-normalizer.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Generic normalize method that removes the accents and special characters.
+ * This method can be used to normalize the string before performing diacritic-insensitive string comparisons
+ */
+export const normalizeString = (value: string): string => {
```
rajsite, line 5: lets call the file string-normalizers.ts and name this function diacriticInsensitiveStringNormalizer

packages/nimble-components/src/label-provider/rich-text/index.ts
```diff
@@ -39,6 +41,9 @@ export class LabelProviderRichText
     @attr({ attribute: 'toggle-numbered-list' })
     public toggleNumberedList: string | undefined;
 
+    @attr({ attribute: 'mention-user-label' })
+    public mentionUserLabel: string | undefined;
```
rajsite, line 45: User mentions are not applicable to all RichText elements. User mentions are like a plugin. Theoretically there could be many types of mentions. Some defined outside nimble. They shouldn't all be added to the label provider of rich text.

Lets instead make this an attribute on the rich-text-mention base class called `@observable buttonLabel?:string;`.

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -51,6 +52,11 @@ export class MentionInternals {
      */
     public readonly viewElement: string;
 
+    /**
+     * Label to use as accessible name and title of mention button
+     */
+    public readonly label: string;
```
rajsite, line 58: This label is being treated as if it is statically defined. But it can change dynamically. The lang of a page and the localized strings can be updated programmatically. Instead buttonLabel should be an observable property on mention internals updated from the buttonLabel attribute on the rich-text-mention-users element

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
```diff
@@ -63,6 +69,7 @@ export class MentionInternals {
         this.icon = options.icon;
```
rajsite, line 69: Instead of defining this upstream, we should change mentionInternals to be the iconTemplate. The iconTemplate instance can and should be shared instead of reconstructed everytime the configuration changes.

i.e.

```suggestion
        this.iconTemplate = html`<${options.icon} slot="start"></${options.icon}>`;
```

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
```diff
@@ -1,18 +1,24 @@
...
  */
 export class MentionExtensionConfiguration {
     private static instance = 0;
-
+    public readonly iconTemplate: ViewTemplate;
     public readonly viewElement: string;
     public readonly character: string;
     public readonly name: string;
     public readonly key: string;
+    public readonly label: string;
```
rajsite, line 20: We should propagate the anem buttonLabel from the other discussions. Also buttonLabel will be the first observable property that is part of MentionExtensionConfiguration so it needs `isObservedMentionInternalsProperty` for the buttonLabel and follow the handleChange patter from the editor integration PR.

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +87,31 @@ export const template = html<RichTextEditor>`
                         ${x => richTextToggleNumberedListLabel.getValueFor(x)}
                         <${iconNumberListTag} slot="start"></${iconNumberListTag}>
                     </${toggleButtonTag}>
+                    ${repeat(
+        x => Array.from(x.getMentionExtensionConfig()),
```
rajsite, line 91: Isn't mentionExtensionConfig already an array?
```suggestion
        x => x.getMentionExtensionConfig(),
```

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -320,6 +331,33 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+     * @internal
+     */
+    public mentionButtonKeyDown(
+        event: KeyboardEvent,
+        character: string
+    ): boolean {
+        if (this.keyActivatesButton(event)) {
+            this.tiptapEditor
+                .chain()
+                .insertContent(` ${character}`)
```
rajsite, line 353: Is this space before the character needed? If so ccan you put a comment explaining?
```
                .chain()
                // Need s space for some reason
                .insertContent(` ${character}`)
```

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -21,7 +21,7 @@ export abstract class RichText extends FoundationElement {
     public readonly childItems: Element[] = [];
 
     @observable
-    protected configuration?: Configuration;
+    public configuration?: Configuration;
```
rajsite, line 24: Mark internal

packages/nimble-components/src/rich-text-mention/users/view/styles.ts
```diff
@@ -32,4 +33,8 @@ export const styles = css`
     :host([disable-editing]) slot {
         display: none;
     }
+
+    :host([disable-editing]) {
+        font: ${mentionFont};
+    }
```
rajsite, line 39: This should be next to the other host selectors to be in DOM order, see: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

rajsite (commented):
Primarily reviewed API surface, haven't reviewed the mention-list-box implementation yet.

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -267,4 +267,8 @@ export const styles = css`
         top: ${mediumPadding};
         right: var(--ni-private-rich-text-editor-scrollbar-width);
     }
+
+    .anchored-region[hidden] {
```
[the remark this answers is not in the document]
suseendran-ni: Removed as per the suggestion.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,294 @@
...
+                return false;
+            }
+        }
+    }
+
+    /**
+     * @public
+     */
+    public clickHandler(e: MouseEvent): boolean {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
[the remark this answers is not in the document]
suseendran-ni: Removed the eslint comment

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -15,27 +15,31 @@ import HardBreak from '@tiptap/extension-hard-break';
...
 import type { MentionExtensionConfiguration } from '../../models/mention-extension-configuration';
+import type { RichTextMentionListBox } from '../../mention-list-box';
+import type { RichTextEditor } from '..';
 
 const validAbsoluteLinkRegex = /^https?:\/\//i;
 
 export function createTiptapEditor(
-    editor: HTMLDivElement,
-    mentionExtensionConfig: MentionExtensionConfiguration[]
+    richTextEditor: RichTextEditor,
```
rajsite, line 29: Think we should avoid coupling this to the entire editor element if possible

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: As we discussed, the clients are expected to filter and supply the mapping elements. Hence, this will not be a concern within the nimble layer.

packages/nimble-components/src/rich-text/mention-list-box/styles.ts
suseendran-ni: Updated as per the suggestion.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: As we updated the architecture to inherit from the list box, closing this thread. Please let us know if you have any concerns further.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Removed the trimming logic.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
m-akinc: Are we expecting the client to change the mapping elements on every change in the filter? If so, does that mean we don't need to filter in the editor at all?

rajsite: Client is expected to reduce the set of mappings to all the mappings already used + the mappings in the dropdown. So the expectations are pretty complex.

Slot only a few mentioned mappings (current mention mappings + a few more up to 10 ish)
On mention-update use the updated value to slot all currently mentioned mappings + whatever is in the list

So the filter needs to run so that the mappings that already mentioned don't show up in the list.

packages/nimble-components/src/label-provider/rich-text/index.ts
```diff
@@ -5,7 +5,8 @@ import {
     richTextToggleBoldLabel,
```
rajsite, line 5: FYI restored [users/aagash/mention-in-editor](https://github.com/ni/nimble/tree/users/aagash/mention-in-editor) and retargeted this branch to it to make reviewing a bit easier. But should merge from main for final reviews.

packages/nimble-components/src/rich-text/base/index.ts
rajsite: With the latest changes doesn't seem to need to be public anymore?

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
rajsite: We could do something similar to `mentionUpdateEmitter` i.e. pass in a function like `activeMentionCharacterEmitter` that is passed the active character.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -194,13 +200,31 @@ function createCustomMentionExtension(
...
+                    onKeyDown: (props): boolean => {
+                        if (props.event.key === keyEscape) {
+                            inSuggestionMode = false;
+                        }
+                        return (
+                            mentionListBox?.keydownHandler(props.event) ?? false
+                        );
+                    },
+                    onExit: (): void => {
+                        mentionListBox?.close();
```
rajsite, line 227: The state of `activeMentionCharacter` should always be accurate in the editor. Seems like we need to reset it onExit right?

packages/nimble-components/src/label-provider/rich-text/index.ts
```diff
@@ -5,7 +5,8 @@ import {
...
     richTextToggleItalicsLabel,
```
rajsite, line 6: It's a bit surprising to me how it looks with only a few matches:
image

And with no matches:
image

The combox behavior isn't perfect but is it possible to get closer to that?

Fits to the size when few items
image

Gets narrow for no items
image

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,285 @@
...
+
+        const capturedListOption = capturedElement as ListOption | null;
+
+        if (!capturedListOption || capturedListOption.disabled) {
+            return false;
+        }
+        this.activateMention({
+            href: capturedListOption.value,
+            displayName: capturedListOption.text
+        } as MentionDetail);
```
rajsite, line 193: avoid `as` casts
const mentionDetail: MentionDetail = {
    href: capturedListOption.value,
    displayName: capturedListOption.text
};
this.activateMention(mentionDetail);

packages/nimble-components/src/rich-text/mention-list-box/styles.ts
```diff
@@ -0,0 +1,18 @@
...
+    ${dropdownStyles}
+
+    :host(:hover)::after,
+    :host(${focusVisible})::after {
+        width: auto;
+    }
+    
+    .listbox {
+        min-width: ${menuMinWidth};
+        height: calc(5.5 * ${controlHeight})
```
rajsite, line 16: Guessing this should be max-height based on the other comment when few or no items exist

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: As we aligned on the above conversation, we will update the `user-service` library to reuse the nimble method for normalization.

Created a task to track this:
https://dev.azure.com/ni/DevCentral/_workitems/edit/2592172

packages/nimble-components/src/utilities/models/string-normalizer.ts
suseendran-ni: Updated the file name and method name as per the suggestion

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
suseendran-ni: Updated as per the suggestion.

packages/nimble-components/src/rich-text/editor/template.ts
suseendran-ni: Updated the logic as per the suggestion

packages/nimble-components/src/label-provider/rich-text/index.ts
suseendran-ni: We did the below steps:

1. Merged the angular changes [PR ](https://github.com/ni/nimble/pull/1686)to the `main` branch.
2. Merged the latest `main` branch to this PR branch. (This brings this PR on top of latest `main` branch)
3. Updated this PR's target branch to the `main` branch.

packages/nimble-components/src/label-provider/rich-text/index.ts
suseendran-ni: Updated the styling to match the combobox.

packages/nimble-components/src/rich-text/mention-list-box/styles.ts
suseendran-ni: Updated as per the suggestion.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: Removed unwanted casting.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -658,26 +692,53 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
                     return {
                         onStart: (props): void => {
-                            this.triggerMentionEvent(props.text);
+                            inSuggestionMode = true;
+                            this.onMention(props);
                         },
 
                         onUpdate: (props): void => {
-                            this.triggerMentionEvent(props.text);
+                            if (!inSuggestionMode) {
```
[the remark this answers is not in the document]
suseendran-ni: This handling is done to mitigate the existing [bug ](https://github.com/ueberdosis/tiptap/issues/214) in tiptap, 

Currently, the TipTap suggestions get completed only when we invoke `props.command();`. We do this operation when the user selects the dropdown option successfully (i.e., positive case). But there is no support to get out of the suggestion mode by pressing the `Escape` key. 

Hence we added this logic to not show the `mention-list` popup after the user presses the `Escape` key. With this, even if the suggestion mode exists in the background (i.e., the decoration tag is active), the user will not see any suggestions in the UI.

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -343,6 +350,26 @@ export class RichTextEditor extends RichText implements ErrorPattern {
         return true;
     }
 
+    /**
+     * Toggle the mention node and focus back to the editor
+     * @internal
+     */
+    public mentionButtonClick(character: string): void {
+        this.tiptapEditor.chain().insertContent(` ${character}`).focus().run();
```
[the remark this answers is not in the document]
vivinkrishna-ni: The prefixed space is required to be inserted before the character to open the mention popup but not in all cases. To open a mention popup, either it should be at the start of a line or it should at least have a single whitespace before the `@` char to open a popup.

We are trying to achieve a behavior like in AzDo or in the GitHub editor, i.e. when the text cursor is right next to a word and when the `@` button is clicked, add a space before `@` to open a popup.
@-space-before-char

However, if there is already a space or the text cursor is at the start of the line, we should not be inserting a space before the character. I have pushed the changes that add space before the char only when necessary.

packages/nimble-components/src/rich-text/base/index.ts
aagash-ni: Current test cases are using this configuration. We will revisit the test scripts and update this.

packages/nimble-components/src/label-provider/rich-text/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text-mention/base/models/mention-internals.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
aagash-ni: Updated the logic, now it will insert space based on `startOfLine` and whitespace on the previous node.

packages/nimble-components/src/rich-text-mention/users/view/styles.ts
aagash-ni: Updated.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * Triggers when the mention plugin is activated upon pressing the `key`
+     *
+     * @public
+     */
+    public onMention(props: SuggestionProps): void {
+        this.suggestionProps = props;
+        this.filter = props.query;
+        this.anchorElement = props.decorationNode as HTMLElement;
```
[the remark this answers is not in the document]
aagash-ni: Updated.

packages/nimble-components/src/label-provider/rich-text/index.ts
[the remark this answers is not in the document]
aagash-ni: Added a few emojis and hyphens.

packages/nimble-components/src/rich-text/mention-list-box/template.ts
```diff
@@ -0,0 +1,32 @@
...
+import { children, elements, html, ref, repeat, when } from '@microsoft/fast-element';
+import type { RichTextMentionListBox } from '.';
+import { Listbox, listboxTag } from '../../listbox';
+import { anchoredRegionTag } from '../../anchored-region';
+import type { MappingConfig } from '../../rich-text-mention/base/models/mapping-config';
+import { listOptionTag } from '../../list-option';
+
+// prettier-ignore
+export const template = html<RichTextMentionListBox>`
+  <template ?hidden="${x => !x.open}">
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -320,6 +331,33 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+    }
+
+    /**
+     * Toggle the mention node and focus back to the editor
+     * @internal
+     */
+    public mentionButtonKeyDown(
```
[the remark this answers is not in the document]
aagash-ni: Removed this method as the key down for `enter` and `space` has been called `@click event`.

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
```diff
@@ -23,6 +40,13 @@ export class MentionExtensionConfiguration {
 
         this.viewElement = mentionInternals.viewElement;
         this.character = mentionInternals.character;
+        this.mappingConfigs = mentionInternals.mappingConfigs;
+        this.iconTemplate = mentionInternals.iconTemplate;
+        this.label = mentionInternals.buttonLabel ?? '';
```
rajsite, line 45: Propagate the name change for consistency. It makes it hard to follow when the name changes unnecessarily multiple times.


```suggestion
        this.buttonLabel = mentionInternals.buttonLabel ?? '';
```

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -354,6 +406,40 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+                args
+            )
+        ) {
+            const mentionExtensionConfig = this.getMentionExtensionConfigFromCharacter(source.character);
+            if (mentionExtensionConfig) {
+                mentionExtensionConfig.label = source.buttonLabel ?? '';
+            }
+            return;
+        }
+        super.handleChange(source, args);
```
rajsite, line 440: I dunno if this is trying to be an optimization or something but don't be clever. Just recreate the config. This will write to the configuration property which is observable and shouuld refresh other bindings.

```suggestion
        if (
            source instanceof MentionInternals
            && MentionExtensionConfiguration.isObservedMentionInternalsProperty(
                args
            )
        ) {
            this.configuration = this.createConfig();
        } else {
            super.handleChange(source, args);
        }
        
```

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -475,6 +566,9 @@ export class RichTextEditor extends RichText implements ErrorPattern {
             'aria-disabled',
             this.disabled ? 'true' : 'false'
         );
+        if (this.disabled) {
+            this.mentionListBox?.close();
+        }
```
rajsite, line 571: This is surprising to me... What is the case where the editor is not disabled and we want to keep the mentionListBox open? MAke sure to comment abouut why this is significant.

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -45,6 +45,9 @@ export abstract class RichTextMention<
     @attr
     public pattern?: string;
 
+    @observable
+    public buttonLabel?: string;
+
```
rajsite, line 50: This is a user facing attribute. The idea is that when configuring the mention youu set this property


```suggestion
    @attr({ attribute: 'button-label' })
    public buttonLabel?: string;

```

This will need to be set in the anguular directive as well

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +90,37 @@ export const template = html<RichTextEditor>`
...
+                        content-hidden
+                        ?disabled="${(_x, c) => c.parent.disabled}"
+                        slot="start"
+                        title=${x => x.label}
+                        @click=${(x, c) => c.parent.mentionButtonClick(x.character)}
+                    >
+                        ${x => x.label}
+                        ${x => x.iconTemplate}
+                    </${buttonTag}>`
+    )}
```
rajsite, line 106: cleanup whitespace

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +90,37 @@ export const template = html<RichTextEditor>`
...
                 </${toolbarTag}>
                 <span class="footer-actions" part="footer-actions">
                     <slot name="footer-actions"></slot>
                 </span>
             </section>
             ${errorTextTemplate}
         </div>
+        <${richTextMentionListBoxTag}
```
rajsite, line 114: cleanup whitesapce

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +90,37 @@ export const template = html<RichTextEditor>`
...
+            ${ref('mentionListBox')}
+            @activate-mention=${(x, c) => x.activateMention(c.event as CustomEvent<MentionDetail>)}
```
rajsite, line 116: make sure to update the event name you bind to, i.e. mention-selected

packages/nimble-components/src/rich-text/mention-list-box/index.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,289 @@
...
+     * @internal
+     */
+    public override handleChange(source: unknown, args: string): void {
+        if (source instanceof AnchoredRegion) {
+            if (args === 'initialLayoutComplete') {
+                this.focusAndScrollOptionIntoView();
+            }
+        } else {
+            super.handleChange(source, args);
+        }
```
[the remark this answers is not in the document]
aagash-ni: Updated as per your suggestion

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,289 @@
...
+            requestAnimationFrame(() => {
+                this.firstSelectedOption?.scrollIntoView({ block: 'nearest' });
+            });
+        }
+    }
+
+    private activateMention(mentionDetail: MentionDetail): void {
+        this.$emit('activate-mention', mentionDetail);
+        this.setOpen(false);
+    }
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,289 @@
...
+     *
+     * @public
+     */
+    public onMention(mentionParam: MentionParam): void {
+        this.filter = mentionParam.filter;
+        this.anchorElement = mentionParam.anchorNode;
+        this.setOpen(true);
+        this.filterOptions();
+    }
+
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,289 @@
...
+            case keyTab:
+            case keyEnter: {
+                if (!this.hasSelectableOptions) {
+                    return false;
+                }
+                this.activateMention({
+                    href: this.firstSelectedOption.value,
+                    displayName: this.firstSelectedOption.text
+                } as MentionDetail);
+                return true;
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -354,6 +406,40 @@ export class RichTextEditor extends RichText implements ErrorPattern {
         return Array.from(mentionedHrefs);
     }
 
+    public getMentionExtensionConfig(): MentionExtensionConfiguration[] {
```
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -354,6 +406,40 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+        return this.configuration instanceof EditorConfiguration
+            ? this.configuration.mentionExtensionConfig
+            : [];
+    }
+
+    public activateMention(event: CustomEvent<MentionDetail>): void {
```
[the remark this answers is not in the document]

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -146,6 +162,23 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+    /**
+     * @internal
+     */
+    @observable
+    public activeMappingConfigs?: MappingConfigs;
+
+    /**
+     * @internal
+     */
+    public activeMentionCommand?: (props: unknown) => void;
```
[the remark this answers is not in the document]

packages/nimble-components/src/rich-text/mention-list-box/template.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
aagash-ni: Removed disable eslint for whole file.

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
```diff
@@ -1,20 +1,37 @@
...
+    private _label = '';
+
+    @observable
+    public get label(): string {
+        return this._label;
+    }
+
+    public set label(value: string) {
+        this._label = value;
+    }
```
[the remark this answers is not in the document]
aagash-ni: Updated, now the label change will create config again.

packages/nimble-components/src/rich-text/editor/index.ts

packages/nimble-components/src/rich-text/editor/template.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/mention-extension-configuration.ts
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Removed the condition check and just closed the popup when disabled

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,288 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-rich-text-mention-list-box': RichTextMentionListBox;
+    }
+}
+
+/**
+ * A rich text mention list box which acts as a popup for "@mention" support in editor
+ */
+export class RichTextMentionListBox extends FoundationListbox {
```
jattasNI, line 25: The approach to derive from `FoundationListbox` here is nice because it is more consistent with how the select and combobox work. But it leaves us with a testing gap since we're no longer using the Nimble listbox directly. That means nothing is testing the visual appearance created by this component's styles and template. I think we need a storybook matrix test for this component. You should be able to copy the cases from the listbox test: cover all themes plus ensuring that long text renders with an ellipsis. If you run into problems with having multiple mention-list-box components open on the same page, you can copy patterns from menu button/select/combobox where we create separate stories for each theme. (You don't need separate stories for open above/below like those components have)

I was also expecting to see unit tests for this component. We at least need the basic test that verifies its tag is registered and that the component can be created. I'm still looking through the coverage in other files: is the remaining behavior (key handling, open/close) covered by integration tests in those components?

packages/nimble-components/src/rich-text/mention-list-box/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface MentionListBoxShowOptions {
```
jattasNI, line 1:
```suggestion
/**
 * Options passed to the `show()` method of the mention list box
 */
export interface MentionListBoxShowOptions {
```

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
```diff
@@ -186,20 +230,27 @@ export class RichTextEditorPageObject {
...
     }
 
-    public async sliceEditorContent(number: number): Promise<void> {
-        const lastElement = this.getEditorLastChildElement();
-        const text = lastElement.parentElement!.textContent!;
-        lastElement.parentElement!.textContent = text.substring(
-            0,
-            text.length - number
-        );
+    public async sliceEditorContent(from: number, to: number): Promise<void> {
```
jattasNI, line 233: Does this implementation change address the TODOs in `rich-text-editor-mention.spec.ts` where this method is being called?

packages/nimble-components/src/rich-text/mention-list-box/index.ts
jattasNI, line 25: We should match FAST's casing for "Listbox": all one word, not "list box" or "ListBox" or "list-box". This affects a few class and file names in the PR.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -195,13 +212,41 @@ function createCustomMentionExtension(
...
+                            return;
+                        }
                         config.mentionUpdateEmitter(props.query);
+                        activeMentionCommandEmitter(props.command);
+                        mentionListBox?.show({
+                            filter: props.query,
+                            anchorNode: props.decorationNode as HTMLElement
+                        });
+                    },
+                    onKeyDown: (props): boolean => {
```
jattasNI, line 238: Should this handler also be returning early if `!inSuggestionMode`? Why do we need to call the mention listbox key handling code in that case?

jattasNI (commented):
I didn't get a chance to review test coverage of the editor yet, but here's my feedback of the implementation

packages/nimble-components/src/rich-text/mention-list-box/index.ts
```diff
@@ -0,0 +1,288 @@
...
+
+        if (this.filteredOptions.length) {
+            this.selectedOptions = [this.filteredOptions[0]!];
+            this.selectedIndex = this.options.indexOf(this.firstSelectedOption);
+        } else {
+            this.selectedOptions = [];
+            this.selectedIndex = -1;
+        }
+
+        this.selectFirstOption();
```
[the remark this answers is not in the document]
suseendran-ni: Removed the `selectFirstOption()` call as it was redundant

packages/nimble-components/src/rich-text/editor/template.ts
```diff
@@ -84,12 +93,33 @@ export const template = html<RichTextEditor>`
                         ${x => richTextToggleNumberedListLabel.getValueFor(x)}
                         <${iconNumberListTag} slot="start"></${iconNumberListTag}>
                     </${toggleButtonTag}>
+                    ${repeat(x => x.getMentionExtensionConfig(), html<MentionExtensionConfiguration, RichTextEditor>`
+                        <${buttonTag}
+                            appearance="ghost"
+                            content-hidden
+                            ?disabled="${(_x, c) => c.parent.disabled}"
+                            slot="start"
+                            title=${x => x.buttonLabel}
```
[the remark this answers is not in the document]
vivinkrishna-ni: This is because we haven't set the `button-label` attribute value to the `nimble-rich-text-mention-users` element in the storybook. As mentioned in [the other comment](https://github.com/ni/nimble/pull/1694#discussion_r1421101129) setting the title and label to this button is a user configuration, so, we cannot see a tooltip or the text in the DOM unless we set it in the element.

Therefore, I added value for the `button-label` on `nimble-rich-text-mention-users` in the storybook and updated the docs for the attribute addition as well.

packages/nimble-components/src/utilities/models/string-normalizers.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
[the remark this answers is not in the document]
vivinkrishna-ni: Added the wacky string test case and set the expected values as `output`

packages/nimble-components/src/rich-text/mention-list-box/template.ts
```diff
@@ -0,0 +1,34 @@
...
+// prettier-ignore
+export const template = html<RichTextMentionListBox>`
+    <template>
+        <${anchoredRegionTag}
+            ${ref('region')}
+            class="anchored-region"
+            fixed-placement
+            auto-update-mode="auto"
+            vertical-default-position="'bottom'"
+            vertical-positioning-mode="'locktodefault'"
```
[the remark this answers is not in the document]
vivinkrishna-ni: Yes, removed

packages/nimble-components/src/rich-text/mention-list-box/index.ts
vivinkrishna-ni: Updated

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
vivinkrishna-ni: Yes, this change addresses a couple of TODOs where we used the `sliceEditorContent` methods in the spec file. We are utilizing Tiptap's utilities, which perform the actual deletion of texts in the editor. So I have removed those TODOs from the spec file. In the same way, I have modified for text addition pageobject method too.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
aagash-ni: Added matrix test and unit test for mention listbox. Key handling and open/close logic are tested under editor cases.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
aagash-ni: Added `!inSuggestionMode` case to return early in `onKeyDown` when `inSuggestionMode` is in false.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
suseendran-ni: To capture the performance of the `diacriticInsensitiveStringNormalizer` method, I have created a code snippet that attempts to normalize 1000 user names.

The operation was completed within 0.5 to 1.5 ms on average.

Attaching the code snippet and the gif for reference.

```
let diacriticInsensitiveStringNormalizer = function(value) {
    return value
        .normalize('NFD')
        .replace(/[\u0300-\u036f]/g, '')
        .toLowerCase();
};

// holds 50 sample users
let sampleUsers = [
  "Alice Bérnard", "Böb Müller", "Chàrlie Dupont", "Dávid Sánchez", "Ëmma Jøhnsøn", "Fränk Dávïs", "Grâce Míller", 
  "Hénry Wilsön", "Ïvy Lée", "Jâck Thompšon", "Kátie Táylor", "Léo Märtínez", "Nöah Gómez", "Olívia Léblanc", 
  "Pénélope Dubois", "Quèntin Roüx", "Róse Mórgan", "Sàrah López", "Théodore Nguyên", "Ürsula Wagner", 
  "Víctor Íbrahim", "Wén Huáng", "Xàvier Delacroix", "Yüko Yamamoto", "Zoé Carré", "Andréa Fischer", 
  "Bèatrice Dupuis", "Cédric Lefèvre", "Délphine Martin", "Étienne Moreau", "Flörence Lambert", "Gabriel Lévesque", 
  "Hélène Dubois", "Isaïah André", "Jülia Costa", "Kévin Lemoine", "Lïna Deschamps", "Màrc Dupont", "Nàdia Rocha", 
  "Oscar Léger", "Päuline Girard", "Quincy Adams", "Rëbecca Märcel", "Sébastien Gagnon", "Tátiana Ruiz", "Ünique Léon",
  "Véra Leblanc", "William Martin", "Xénia Rodriguez", "Yánnis Papadopoulos"
]

let inputUsers = [];
let outputUsers = [];

// Generates 1000 users
for (let i = 0; i < 20; i++) {
  inputUsers = [...inputUsers, ...sampleUsers];
}

console.log("Sample count", inputUsers.length);

console.time("Measure");
for(let user of inputUsers) {
    outputUsers.push(diacriticInsensitiveStringNormalizer(user));
}
console.timeEnd("Measure");

console.log(outputUsers);

```


Normalization Metrics

packages/nimble-components/src/rich-text-mention/base/index.ts
vivinkrishna-ni: Updated as per your suggestion in components and created a PR for Angular directive: https://github.com/ni/nimble/pull/1706

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -588,9 +566,17 @@ describe('RichTextEditorMention', () => {
                 'test:2'
             ]);
         });
+
+        it('user mention button should have `at icon` buttons when user configuration and test elements are given', async () => {
+            await appendUserMentionConfiguration(element);
+            await appendTestMentionConfiguration(element);
+
+            expect(pageObject.getMentionButtonIcon(0)).toBe(ICON_AT_TAG);
+            expect(pageObject.getMentionButtonIcon(1)).toBe(ICON_AT_TAG);
```
jattasNI, line 575: Let's use a different icon for the test mention configuration in the fixtures file so that these can be distinguished from each other.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -849,4 +831,550 @@ describe('RichTextEditor user mention via template', () => {
             pageObject.getEditorMentionViewAttributeValues('mention-label')
         ).toEqual(['John Doe', 'Mary Wilson']);
     });
+
+    describe('user mention button', () => {
+        beforeEach(async () => {
+            await waitForUpdatesAsync();
+        });
+
+        it('should have `at icon` button', () => {
```
jattasNI, line 840: This test is essentially duplicated above in the 'user mention button' describe block. I think you can get rid of one of the copies.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -849,4 +831,550 @@ describe('RichTextEditor user mention via template', () => {
...
+                'P',
+                RICH_TEXT_MENTION_USERS_VIEW_TAG
+            ]);
+            expect(
+                pageObject.getEditorMentionViewAttributeValues('mention-label')
+            ).toEqual(['username2']);
+            expect(pageObject.isMentionListboxOpened()).toBeFalse();
+        });
+
+        it('focus out from the editor should close the mention popup', async () => {
```
jattasNI, line 1162: I think we're missing a case to verify that pressing ESC closes the popup

jattasNI (approved):

jattasNI:
Once you address my feedback you are welcome to complete the PR!

packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts
vivinkrishna-ni: Added a comment in the issue: https://github.com/ni/nimble/issues/1568#issuecomment-1853282332

packages/nimble-components/src/rich-text/mention-listbox/index.ts
```diff
@@ -0,0 +1,286 @@
...
+import type { MentionListboxShowOptions } from './types';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-rich-text-mention-listbox': RichTextMentionListbox;
+    }
+}
+
+/**
+ * A rich text mention list box which acts as a popup for "@mention" support in editor
```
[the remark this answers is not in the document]
AagashRaaj: Updated

packages/nimble-components/src/utilities/models/tests/string-normalizers.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { parameterizeNamedList } from '../../tests/parameterized';
+import { diacriticInsensitiveStringNormalizer } from '../string-normalizers';
+
+// Wacky strings copied from '../../tests/wacky-strings.ts' and updated their corresponding normalized string output
+export const wackyStrings = [
```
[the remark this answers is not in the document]
AagashRaaj: Updated

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
```diff
@@ -305,6 +318,95 @@ Plain text 3`);
...
+                { key: 'user:2', displayName: 'username2' }
+            ]);
+            await commitFirstMentionBoxOptionIntoEditor('@');
+            await pageObject.setEditorTextContent('@');
+            await waitForUpdatesAsync();
+            await pageObject.clickMentionListboxOption(1);
+            expect(element.getMarkdown()).toEqual('<user:1> <user:2> ');
+        });
+
+        it('Multiple Mention node of differnt type', async () => {
```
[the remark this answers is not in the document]
AagashRaaj: Updated

packages/nimble-components/src/rich-text/mention-listbox/index.ts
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/utilities/models/tests/string-normalizers.spec.ts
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/models/tests/markdown-serializer.spec.ts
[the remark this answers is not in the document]
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Removed.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Replaced with exclamatory icon

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Added a test case to check this esc behavior.

vivinkrishna-ni:
@jattasNI We have one medium-severity bug that needs a fix within this week. This bug is due to an issue in the Tiptap where the two anchor elements set via `setContent` render as a single link which causes the mentions to render weirdly. However, we found a quick workaround to resolve the issue for now and created [an issue in Tiptap](https://github.com/ueberdosis/tiptap/issues/4686). This change involves adding a `class` attribute to all the anchor elements renders in the rich text components and here is the [commit link](https://github.com/ni/nimble/pull/1694/commits/e4108306b52b33327e34d0b0b760856370429bc1) with those changes for your reference.

I have created an issue in Nimble https://github.com/ni/nimble/issues/1707 and explained in detail about the issue and workaround done. Also, added code comments in the necessary places by linking the issue to remove the workaround once the issue is resolved both in Tiptap and in Nimble. Written tests for the current bug fix in the same commit. Therefore, we are merging this PR now, and if you have any comments on these changes we will address them in the follow-up PR.

packages/nimble-components/src/rich-text/mention-list-box/index.ts
vivinkrishna-ni: Updated the usage docs on restricting the number of users at a time to improve the performance.

packages/nimble-components/src/rich-text/base/index.ts
aagash-ni: Updated it to `protected`. Generated the required configuration dynamically based on mappings just to parse and get the HTML in page object `getParserMentionConfigForUser()`, replaced the actual element configuration usage.

PR #1701, opened 2023-12-05
Improve wafer map die highlighting API

Razvan1928:
# Pull Request

## 🤨 Rationale
The current implementation of wafer map die highlighting has limitations, particularly when dealing with metrics and attribute data. This pull request aims to address these limitations and provide a more flexible solution for highlighting dies based on user-defined tags.

more details can be found in the [spec file](ni-nimble\nimble\packages\nimble-components\src\wafer-map\specs\features\change-highlight.md)

## 👩‍💻 Implementation

The proposed solution involves replacing the existing logic with a new one that introduces an optional field called "tags" for each die in the wafer map. The "tags" can be user-defined and may include simplified forms or attribute IDs. Additionally, a new property highlightedTags is introduced in the wafer map to specify which tags should trigger the highlighting.

## ✅ Checklist

----

## 🧪 Testing

Tested with the modified existing unit tests 

and manually using the nimble storybook, which was updated also


packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -131,13 +132,27 @@ export class Prerendering {
...
+        dieTags?: string[],
+        highlightedTags?: string[]
     ): number {
-        return highlightedValues.length > 0
-            && !highlightedValues.some(dieValue => dieValue === selectedValue)
-            ? this.nonHighlightedOpacity
-            : 1;
+        if (
+            highlightedTags?.length === 1
+            && highlightedTags?.every(tag => tag === '')
```
munteannatan, line 140: i think you do not have to check for '' tags as that is a valid value in the dieTags list

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -131,13 +132,27 @@ export class Prerendering {
...
+        ) {
+            return this.nonHighlightedOpacity;
+        }
+        const tagsMatch = this.isAnyDieTagPresent(dieTags, highlightedTags);
+        return tagsMatch ? this.nonHighlightedOpacity : 1;
+    }
+
+    private isAnyDieTagPresent(
```
munteannatan, line 148: I think the logic from here can be combined with the method above without issue, as that method just does a empty check before calling this

packages/nimble-components/src/wafer-map/modules/prerendering.ts
Razvan1928: you are right

change/@ni-nimble-components-f53bdc20-40c9-40ed-9375-2b99fccc4bfd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "changes",
```
atmgrifter00, line 3: Typically, we'd like the `comment` field to have a concise, but explanatory description of what these changes are providing.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -137,7 +137,7 @@ export class WaferMap extends FoundationElement {
      */
     @observable public hoverDie: WaferMapDie | undefined;
 
-    @observable public highlightedValues: string[] = [];
+    @observable public highlightedTags: string[] = [];
```
atmgrifter00, line 140: While it looks like this is the agreed upon terminology for this API in the discussion on the PR for the HLD, the spec document itself was not updated to reflect this. Can you correct that?

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -107,13 +107,13 @@ _The key elements of the component's public API surface:_
     -   `dieLabelsHidden` - a boolean value that determines if the die labels in the wafer map view are shown or not. Default value is false.
     -   `dieLabelsSuffix` - represent a string that can be added as a label in the end of the each data information in the wafer map dies value.
     -   `colorScaleMode` - represent an Enum value that determent if the colorScale is represent a continues gradient values (linear), or is set categorically (ordinal).
-    -   `highlightedValues` - represent a list of strings of dies values that will be highlighted in the wafer map view
+    -   `highlightedTags` - represent a list of strings based on which the highlight will be done
```
atmgrifter00, line 110:
```suggestion
    -   `highlightedTags` - a list of strings that will denote which dies to highlight
```

packages/nimble-components/src/wafer-map/tests/prerendering.coloring.spec.ts
```diff
@@ -363,7 +363,13 @@ describe('Wafermap Prerendering module', () => {
         it('should have highlighted value with full opacity and the rest with expected opacity', () => {
             const waferMapDies = getWaferMapDies();
             const expectedValues = waferMapDies.map(x => {
```
atmgrifter00, line 365: Can we change `x` to `die` or even `waferMapDie`?

packages/nimble-components/src/wafer-map/tests/utilities.ts
```diff
@@ -52,7 +52,7 @@ export function getColorScale(): WaferMapColorScale {
     return { colors: ['red', 'blue', 'green'], values: ['1', '2', '3'] };
 }
 
-export function getHighlightedValues(): string[] {
+export function getHighlightedTags(): string[] {
```
atmgrifter00, line 55: So, I may be misunderstanding something, but this looks like we are no longer returning tags that will ultimately be highlighted, as none of the dies defined in `getWaferMapDies()` are returning tags with these values.

Razvan1928: You are right, I added the values from getHighlightedTags() to be among the dies returned by getWaferMapDies(), tests are still passing

packages/nimble-components/src/wafer-map/specs/README.md
```diff
@@ -107,13 +107,13 @@ _The key elements of the component's public API surface:_
...
     -   `disabled` - it's represented by a boolean value and refers to the state of the `nimble-wafer-map` component. If true, the component should be rendered dimmed out and no user interaction should be allowed.
     -   `validity` - readonly object of boolean values that represents the validity states that the wafer map's configuration can be in. The object's type is `WaferMapValidity`, and it contains the following boolean properties:
         -   `invalidGridDimensions` : true when any of the `gridMinX`, `gridMinY`, `gridMaxX` or `gridMaxY` is `undefined`, but false when all of them are `undefined`.
 
 The `originLocation`, `orientation`, `gridMinX`, `gridMinY`, `gridMaxX`, `gridMaxY`, `dieCharacterCount`, `disabled`, `waferDataType` and `colorBy` properties will be configurable via properties and attributes.
-The `dies`, `colorScale` and `highlightedValues` properties will be configurable only via properties and will not have attributes.
+The `dies`, `colorScale` and `highlightedTags` properties will be configurable only via properties and will not have attributes.
```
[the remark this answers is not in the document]
Razvan1928: You are right, I added "tags?: string[]" at line 92 (where the WaferMapDie is displayed), and I expended the documentation to be more clear about how tags should be used, at 110

DStavilaNI:
Please add the required documentation for this PR:
image
I know that you mentioned the spec here, but I would like to see some PR description here.... also, I am not sure that the images that you have presented in the description are actually adding value ... at least not all of them. I am sure 1 or 2 images, lower size, would be enough to illustrate the concept.

DStavilaNI (commented):
I think that this could be packed into it's own function that would return a RenderInfo object... it would make the code a bit easier to understand/modular
image

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -131,13 +132,14 @@ export class Prerendering {
...
+        highlightedTags?: string[]
     ): number {
-        return highlightedValues.length > 0
-            && !highlightedValues.some(dieValue => dieValue === selectedValue)
-            ? this.nonHighlightedOpacity
-            : 1;
+        if (!highlightedTags || highlightedTags.length === 0) {
+            return 1;
+        }
+        const tagsMatch = dieTags?.some(dieTag => highlightedTags.some(highlightedTag => dieTag === highlightedTag));
```
DStavilaNI, line 141: This does not look very efficient ... I do not know where the `dieTags `or the `highlightedTags ` come from, but did you consider packing them into a set or a map instead of arrays ? It would make this kind of searches much easier and/or faster.

DStavilaNI:
Probably this should be a link

image

DStavilaNI:
For the overall understanding, and since this is not mentioned in the spec either, I think that it would be great to explain where do the `tags `come from. The idea of the `highlightedTags` and their relationship with the `tags` is quite easy to understand, but saying that the user will supply originally the list of `tags` is not very explanatory - it does lack some context. Might be worth saying something like (and if I understood correctly) the tags will be given/supplied from the list of visualizations, displayed in the same information context as the wafermap, and which display the same data (?), but from a different perspective.... if you agree with my comment, then I believe that the spec itself should also be updated.

packages/nimble-components/src/wafer-map/modules/prerendering.ts
rajsite: I had a similar thought but one discussion point was the ability to for example specify a unique tag for each die to be able to address them: https://github.com/ni/nimble/pull/1678#discussion_r1408282049

It's a contrived example but in that scenario caching dies by tag where the cache is larger than the full set of dies (1 per die and then additional tags) might be costly as well. 

Agree that the assumption that highlightedTags and dieTags is small for now is probably a good starting point and we can consider additional optimizations as the use-case comes up (unless we know of concerns now).

rajsite (commented):
My primary concern is keeping test / storybook data deterministic and avoiding random data.

change/@ni-nimble-components-f53bdc20-40c9-40ed-9375-2b99fccc4bfd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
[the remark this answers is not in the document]
Razvan1928: @jattasNI any opinion?

packages/nimble-components/src/wafer-map/tests/data-generator.ts
```diff
@@ -56,7 +69,11 @@ export const generateWaferData = (
                 j--
             ) {
                 stringValue = generateStringValue(i, j, valueGenerator);
-                diesSet.push(generateDieContent(i, j, stringValue));
+                // Math.random() * (max-min) + min Generate a random number between two numbers
```
jattasNI, line 72: What is the code in this comment referring to? If it's about the implementation of the generation function then it should go inside that function. But I think it's simple enough that it doesn't need to have a comment.

Razvan1928: My bad, it was there from the one previous implementation, I deleted it

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -225,12 +236,12 @@ const metadata: Meta<WaferMapArgs> = {
                 'String that can be added as a label at the end of each wafer map die value',
             control: { type: 'text' }
         },
-        highlightedValues: {
+        highlightedTagsSet: {
```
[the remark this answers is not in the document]
Razvan1928: Good suggestions!

munteannatan (approved):

jattasNI (approved):
The code and Chromatic diffs look good to me. You can approve the UI status checks if they look good to you. Then feel free to complete the PR; don't need to wait on remaining reviewers.

jattasNI:
Note: I changed your PR title so that it gave a bit more context about what component is being changed.

atmgrifter00 (approved):

PR #1705, opened 2023-12-07
Ensure all parameterized tests get run

m-akinc:
# Pull Request

## 🤨 Rationale

If test cases passed to `parameterizeNamedList` have multiple cases with the same name, only the last one gets run. This was the case for some rich-text editor tests. We want to ensure that all test cases get run.

## 👩‍💻 Implementation

- Now throwing error from `parameterizeNamedList` if multiple test cases with the same name are encountered. This should prevent developers from accidentally creating tests that don't get run.
- Uniquified some rich-text editor test case names so that they all get run.

## 🧪 Testing

Tests pass.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/utilities/tests/parameterized.ts
+++ b/packages/nimble-components/src/utilities/tests/parameterized.ts
@@ -144,6 +144,11 @@ export const parameterizeNamedList = <T extends readonly { name: string }[]>(
 ): void => {
     const testCases = list.reduce<{ [key: string]: { name: string } }>(
         (result, entry) => {
+            if (result[entry.name]) {
+                throw new Error(
+                    `Duplicate name found in test case list: ${entry.name}`
+                );
+            }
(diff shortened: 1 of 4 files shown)


packages/nimble-components/src/utilities/tests/parameterized.ts
```diff
@@ -144,6 +144,11 @@ export const parameterizeNamedList = <T extends readonly { name: string }[]>(
 ): void => {
     const testCases = list.reduce<{ [key: string]: { name: string } }>(
         (result, entry) => {
+            if (result[entry.name]) {
+                throw new Error(
+                    `Duplicate name found in test case list: ${entry.name}`
```
rajsite, line 149:
```suggestion
                    `Duplicate name found in test case list: ${entry.name}. Make sure all test names are unique.`
```

PR #1709, opened 2023-12-14
Rich text editor | Fix mention listbox opening at the wrong position when mention dynamically loads

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Whenever the `@mention` is triggered, moving the cursor away from the decoration element and updating the configuration element dynamically opens the popup at the top right of the page. 

In the GIF below, we are updating the configuration element by updating the `Load Config` button to demonstrate the issue
popup-at-the-top-issue-nimble

## 👩‍💻 Implementation

The issue is because the `onStart()` event of the suggestion plugin of Tiptap is triggered whenever we perform `setMarkdown()` when the configuration element updates dynamically. However, if we move the cursor to start as mentioned in the above GIF which triggers the `onExit()` event and then `onStart()` the props in `onStart()` called without a decoration node to load the popup right below `@` character.

So in `onstart()`, if the decoration node is null, we restrict opening the popup.

## 🧪 Testing

- Written a test for the popup issue with the same workflow.
- Manually verified in SLE for the same popup issue

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-77825b10-d033-43ac-9a4c-aa28738727ea.json
+++ b/change/@ni-nimble-components-77825b10-d033-43ac-9a4c-aa28738727ea.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix mention listbox when configuration dynamically changes and when cursor position change",
+  "packageName": "@ni/nimble-components",
+  "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
+++ b/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
@@ -1,4 +1,5 @@
 import { css } from '@microsoft/fast-element';
+import { display } from '@microsoft/fast-foundation';
 import {
     mentionFont,
     mentionDisabledFontColor,
@@ -8,6 +9,8 @@ import {
 } from '../../../theme-provider/design-tokens';
 
 export const styles = css`
+    ${display('inline')}
+
     :host {
         box-sizing: border-box;
         font: ${bodyFont};
--- a/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
+++ b/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
@@ -1,5 +1,4 @@
 import { css } from '@microsoft/fast-element';
-import { display } from '@microsoft/fast-foundation';
 import {
     mentionFont,
     mentionDisabledFontColor,
@@ -9,8 +8,6 @@ import {
 } from '../../../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('inline-block')}
-
     :host {
         box-sizing: border-box;
         font: ${bodyFont};
--- a/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
+++ b/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
@@ -9,7 +9,7 @@ import {
 } from '../../../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('inline')}
+    ${display('inline-block')}
 
     :host {
         box-sizing: border-box;
--- a/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
+++ b/packages/nimble-components/src/rich-text-mention/users/view/styles.ts
@@ -15,7 +15,7 @@ export const styles = css`
         box-sizing: border-box;
         font: ${bodyFont};
         color: ${bodyFontColor};
-        white-space: normal;
+        white-space: pre-wrap;
     }
 
     .control {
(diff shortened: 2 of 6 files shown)


packages/nimble-components/src/rich-text-mention/users/view/styles.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '../../../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('inline-block')}
```
vivinkrishna-ni, line 12: We also found another bug related to the view element where the cursor disappears when moved to the start of the line with mention. See below the GIF for the issue,
cursor-at-start-issue

When we updated the display from `inline-block` to `inline` the issue was resolved as we were not styling any block-level properties to the element.
cursor-at-start-fix

We did high-level testing with this change in the storybook and it is working as expected but we will test in SLE tomorrow and once confirmed we will keep the change. If not, we will remove this change from this PR and create a separate PR to fix this bug later.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -1407,5 +1407,19 @@ describe('RichTextEditorMentionListbox', () => {
                 'testname2'
             ]);
         });
+
+        it('should close mention listbox when cursor position is moved to start and configuration dynamically changes', async () => {
```
jattasNI, line 1411: nit: minor phrasing change
```suggestion
        it('mention listbox should be closed when cursor position is moved to start and configuration dynamically changes', async () => {
```

change/@ni-nimble-components-77825b10-d033-43ac-9a4c-aa28738727ea.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: There was a slight change to the Chromatic snapshots. I'm guessing this is because of the `inline` change? It seems acceptable to approve but you should probably wait to do so until you decide whether to keep that as part of this PR; if you remove that change then maybe the diff will go away.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -223,6 +223,14 @@ function createCustomMentionExtension(
                 let inSuggestionMode = false;
                 return {
                     onStart: (props): void => {
+                        /**
+                         * If the cursor position moves to the start of the mention and configuration element changes,
+                         * the setMarkdown() will trigger this `onStart` without a decoration node because the cursor
+                         * position is moved out of the suggestion decoration
```
jattasNI, line 229:
```suggestion
                         * If the cursor position moves to outside of the mention and configuration element changes,
                         * the setMarkdown() will trigger this `onStart` without a decoration node because the cursor
                         * position is temporarily moved out of the suggestion decoration. Ignore `onStart` in that case
                         * and don't show the mention list box since it doesn't have anything to anchor to.
```

jattasNI (approved):
You can submit without Milan's approval since he's OOO

packages/nimble-components/src/rich-text-mention/users/view/styles.ts
vivinkrishna-ni: Today we validated this change in SLE and found an issue when editing a comment where the cursor disappears after the editor transitioned from viewer. So I removed the change from this PR and created a bug to track and resolve in other follow-up PR. Bug link: https://github.com/ni/nimble/issues/1710

change/@ni-nimble-components-77825b10-d033-43ac-9a4c-aa28738727ea.json
vivinkrishna-ni: Removed the `inline` as we found an issue in the edit comment in SLE. I created a bug and will resolve it in a follow-up PR. https://github.com/ni/nimble/pull/1709#discussion_r1427980861

packages/nimble-components/src/rich-text-mention/users/view/styles.ts
```diff
@@ -15,7 +15,7 @@ export const styles = css`
         box-sizing: border-box;
         font: ${bodyFont};
         color: ${bodyFontColor};
-        white-space: normal;
+        white-space: pre-wrap;
```
vivinkrishna-ni, line 18: We had an issue when setting `white-space: normal` as there are `&nbsp;` added for every whitespace being entered by the user which creates a couple of issues: 
1. When a user types `@username` and adds more than three whitespaces and adds `@` again to the editor it will not trigger the popup as the last character is not a whitespace.
2. To filter with a space, the popup closes as the unicode value is different from the normalized string output. Now when the space is added when showing the suggestion popup, the space will also be considered for filtering the list.

image

By setting the `white-space: pre-wrap;`, it considers only the whitespace and when next time `@` is added it converts to a proper decoration node and shows the popup and filtering as expected with space.

Issue #1716, opened 2023-12-19
Render "nimble-rich-text-mention-users-view" as decoration tag in rich text editor


aagash-ni:
## 🧹 Tech Debt

Currently `nimble-rich-text-editor` uses tip-tap native configuration `span` as a decoration node in `mention-extension`. But the actual design is to use the `nimble-rich-text-mention-users-view` as a decoration tag in the suggestion mode and with the `disabled-editing` attribute in committed mode (After the user got selected).

Here are few difficulties while rendering `nimble-rich-text-mention-users-view` as a decoration tag in suggestions:

1. In Android ( Chrome ), selecting the user from the popup by clicking `Return` on the keyboard is not happening properly.

![Android](https://github.com/ni/nimble/assets/123377167/6bf624bb-60cd-4844-811b-05c28dfb263a)

2. In macOS (Safari) and IOS (chrome), removing the contents in the decoration tag by giving `backspace` is removing the whole decoration tag instead of the last character in it. 

![safari](https://github.com/ni/nimble/assets/123377167/7e0f81ae-79ad-4e0b-aaa1-dda5cd37119c)

vivinkrishna-ni:
For point 2, in macOS, it is an issue with the Safari browser. To verify that I created a contenteditable div and render a shadow DOM element within it. In which pressing backspace removes the entire content instead of the last character in the node in Safari whereas in other browsers like Chrome and Firefox on macOS, it removes only the last character. Here is the example to test it: https://stackblitz.com/edit/typescript-avgrtp?file=package.json,index.html

So I created a bug in WebKit: https://bugs.webkit.org/show_bug.cgi?id=266700. Once it is fixed in WebKit, we can consider rendering the decoration tag as a `nimble-rich-text-mention-users-view` in edit mode.

PR #1721, opened 2024-01-02
Fix for Safari (WebKit) browser issue in removing the suggestion text when mentioning a user

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Fixes: [Bug 2606385](https://dev.azure.com/ni/DevCentral/_workitems/edit/2606385): Mention | Removing the suggestion character removes the whole word in iOS and Mac

## 👩‍💻 Implementation

This PR includes rendering the default HTML span element for the suggestion node by Tiptap, instead of a view element during editing `@mention`. This is achieved by removing the decoration tag configuration in the Mention extension.

Our idea is to use a view element in two states: edit mode (during the mention popup is opened) and view mode (after committing from mention popup). See here for more details on the design discussion: https://github.com/ni/nimble/pull/1653#discussion_r1385849063. However, we have an issue in Safari browsers on Mac and all browsers on iOS & iPadOS where if the user presses backspace, it removes the whole suggestion text. For example, if I type `@john` and press backspace should delete a single character (`@joh`), but it deletes the whole node.

This is because whenever we render a shadow root element (View element) within a contenteditable div like editor, pressing backspace removes the whole content in Safari whereas, it removes a single character in other browsers. Here is the stackblitz example to reproduce: https://stackblitz.com/edit/typescript-avgrtp?file=package.json%2cindex.html

So I have created an issue to WebKit and here is the link to it: https://bugs.webkit.org/show_bug.cgi?id=266700 
In review we decided that we will keep the change of using a single span style for all mention editing and only use custom mention views for committed mentions. That change will happen in follow-up PR: https://github.com/ni/nimble/pull/1731

## 🧪 Testing

Updated test scripts to check if `span` element is rendered when `@mention` is triggered.
New tests were not added to validate backspace behavior. This should be revisited as an integration test, captured here: https://github.com/ni/nimble/issues/1568#issuecomment-1877589782

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-879a3c50-e10c-47fc-82c4-1e6dd51006d8.json
+++ b/change/@ni-nimble-components-879a3c50-e10c-47fc-82c4-1e6dd51006d8.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix for Safari browser issue in removing the suggestion text when pressing backspace",
+  "comment": "Rich text editor fix for Safari browser issue in removing the suggestion text when pressing backspace",
   "packageName": "@ni/nimble-components",
   "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@ni-nimble-components-879a3c50-e10c-47fc-82c4-1e6dd51006d8.json
+++ b/change/@ni-nimble-components-879a3c50-e10c-47fc-82c4-1e6dd51006d8.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for Safari browser issue in removing the suggestion text when pressing backspace",
+  "packageName": "@ni/nimble-components",
+  "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/rich-text/editor/index.ts
+++ b/packages/nimble-components/src/rich-text/editor/index.ts
@@ -593,8 +593,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
         this.tiptapEditor.state.doc.descendants((node, pos) => {
             if (node.type.name.startsWith(mentionPluginPrefix)) {
                 const updatedAttrs = {
-                    href: node.attrs.href as string,
-                    label: node.attrs.label as string,
+                    ...node.attrs,
                     disabled: this.disabled ? '' : null
                 };
                 const updatedNode = this.tiptapEditor.schema.node(
--- a/packages/nimble-components/src/rich-text/editor/index.ts
+++ b/packages/nimble-components/src/rich-text/editor/index.ts
@@ -220,6 +220,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
      */
     public disabledChanged(): void {
         this.disableEditor();
+        this.disableMentionViewElement();
     }
 
     /**
@@ -376,6 +377,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
     public setMarkdown(markdown: string): void {
         const html = this.getHtmlContent(markdown);
         this.tiptapEditor.commands.setContent(html);
+        this.disableMentionViewElement();
     }
 
     /**
@@ -503,6 +505,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
             this.placeholder
         );
         this.disableEditor();
+        this.disableMentionViewElement();
         this.bindEditorTransactionEvent();
         this.bindEditorUpdateEvent();
         this.stopNativeInputEventPropagation();
@@ -584,6 +587,30 @@ export class RichTextEditor extends RichText implements ErrorPattern {
             this.disabled ? 'true' : 'false'
         );
         this.mentionListbox?.close();
+    }
+
+    private disableMentionViewElement(): void {
+        this.tiptapEditor.state.doc.descendants((node, pos) => {
+            if (node.type.name.startsWith(mentionPluginPrefix)) {
+                const updatedAttrs = {
+                    href: node.attrs.href as string,
+                    label: node.attrs.label as string,
+                    disabled: this.disabled ? '' : null
+                };
+                const updatedNode = this.tiptapEditor.schema.node(
+                    node.type.name,
+                    updatedAttrs,
+                    node.content
+                );
+                const updatedTransaction = this.tiptapEditor.state.tr.replaceWith(
+                    pos,
+                    pos + node.nodeSize,
+                    updatedNode
+                );
+                this.tiptapEditor.view.dispatch(updatedTransaction);
+            }
+            return true;
+        });
     }
 
     /**
--- a/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
+++ b/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
@@ -185,13 +185,21 @@ function createCustomMentionExtension(
                         };
                     }
                 },
-
                 label: {
                     default: null,
                     parseHTML: element => element.getAttribute('mention-label'),
                     renderHTML: attributes => {
                         return {
                             'mention-label': attributes.label as string
+                        };
+                    }
+                },
+                disabled: {
+                    default: null,
+                    parseHTML: element => element.getAttribute('disabled'),
+                    renderHTML: attributes => {
+                        return {
+                            disabled: attributes.disabled as string
                         };
                     }
                 }
--- a/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
+++ b/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
@@ -216,7 +216,10 @@ function createCustomMentionExtension(
     }).configure({
         suggestion: {
             char: config.character,
-            decorationTag: config.viewElement,
+            // When rendering the view element as a decoration tag for suggestions, it leads to the deletion of the entire suggested word in Safari.
+            // Refer to the following issue for commenting out the specified line of code as a workaround
+            // https://github.com/ni/nimble/issues/1716
+            // decorationTag: config.viewElement,
             pluginKey: new PluginKey(config.key),
             allowSpaces: true,
             render: () => {
--- a/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
+++ b/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
@@ -219,11 +219,10 @@ function createCustomMentionExtension(
             /**
              * When rendering the view element as a decoration tag for suggestions,
              * it leads to the deletion of the entire suggested word in Safari when pressing backspace.
-             * Refer to the following issue for commenting the line of code as a workaround.
-             * https://github.com/ni/nimble/issues/1716
+             * See: https://github.com/ni/nimble/issues/1716
+             * When addressed, re-enable the view element as follows:
+             * decorationTag: config.viewElement,
              */
-
-            // decorationTag: config.viewElement,
             pluginKey: new PluginKey(config.key),
             allowSpaces: true,
             render: () => {
--- a/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
+++ b/packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
@@ -216,9 +216,13 @@ function createCustomMentionExtension(
     }).configure({
         suggestion: {
             char: config.character,
-            // When rendering the view element as a decoration tag for suggestions, it leads to the deletion of the entire suggested word in Safari.
-            // Refer to the following issue for commenting out the specified line of code as a workaround
-            // https://github.com/ni/nimble/issues/1716
+            /**
+             * When rendering the view element as a decoration tag for suggestions,
+             * it leads to the deletion of the entire suggested word in Safari when pressing backspace.
+             * Refer to the following issue for commenting the line of code as a workaround.
+             * https://github.com/ni/nimble/issues/1716
+             */
+
             // decorationTag: config.viewElement,
             pluginKey: new PluginKey(config.key),
             allowSpaces: true,
--- a/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
+++ b/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
@@ -865,7 +865,7 @@ describe('RichTextEditor user mention via template', () => {
 
             expect(pageObject.getMarkdownRenderedTagNames()).toEqual([
                 'P',
-                RICH_TEXT_MENTION_USERS_VIEW_TAG
+                'SPAN'
             ]);
             expect(pageObject.getEditorFirstChildTextContent()).toBe('@');
         });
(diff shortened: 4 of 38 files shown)


vivinkrishna-ni:
@m-akinc @rajsite @jattasNI, This PR includes the bug fix for a high-severity bug in the SLE on `@mention` feature. I have mentioned the rationale for this code change in the PR description but I'm yet to write the test scripts on the actual bug reproduction. Kindly let me know your initial thoughts/feedback on this workaround and we will add the test scripts ASAP when we get the way to emit the backspace keyboard event programmatically to test in WebKit.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -216,7 +216,14 @@ function createCustomMentionExtension(
...
             char: config.character,
-            decorationTag: config.viewElement,
+            /**
+             * When rendering the view element as a decoration tag for suggestions,
+             * it leads to the deletion of the entire suggested word in Safari when pressing backspace.
+             * Refer to the following issue for commenting the line of code as a workaround.
+             * https://github.com/ni/nimble/issues/1716
+             */
+
+            // decorationTag: config.viewElement,
```
rajsite, line 226: In trying to understand the expected styling. With this change I would expect that there are styles specific to the `nimble-rich-text-mention-users-view` that would temporarily have to move into the rich-text-editor element. 

However I don't see any style related changes in the PR.

Scanning the editor and viewer styles one concern I have is that viewer and the editor seem to have specific knowledge of how the mention view should be styled. I would not expect the following styles in the viewer and editor today:

editor:
https://github.com/ni/nimble/blob/6d862ff460fd93118459ce43fb9179dee2dd584e/packages/nimble-components/src/rich-text/editor/styles.ts#L36

https://github.com/ni/nimble/blob/6d862ff460fd93118459ce43fb9179dee2dd584e/packages/nimble-components/src/rich-text/editor/styles.ts#L85

https://github.com/ni/nimble/blob/6d862ff460fd93118459ce43fb9179dee2dd584e/packages/nimble-components/src/rich-text/editor/styles.ts#L88-L90

viewer:
https://github.com/ni/nimble/blob/3506db58b98fc38b29e5372a77843bdc3cc003a8/packages/nimble-components/src/rich-text/viewer/styles.ts#L20

The font, font color, should be completely determined by the view component and it's configuration. Certainly the rich-text-editor itself should not have any direct import to the `richTextMentionUsersViewTag` that it does now.

Conceptually what I expect to see before this change:

- rich-text-viewer and rich-text-editor styles do not have specific knowledge of any mentions styling
- If state needs to be communicated to children it is done so via properties. I'm not sure what features you have at your disposal in the tiptap apis, but can the disabled attribute be set on the view elements instead of relying on the CSS property like now? It looks like what information is missing on the view is the disabled state.

What I would expect to see as a result of this change:

There is a very temporary section added to the viewer styles for styling the span of the user mention span. The span we emit should also have a way to scope its styles to that specific mention view. For example we add a class based on the view element string, `i.e. <span class="nimble-rich-text-mention-users-view">`

And then in the rich text viewer stylesheet that is what we target temporarily:
```
span.nimble-rich-text-mention-users-view {
 // expected styles
}
```

And the expected styles come at least partially from:
https://github.com/ni/nimble/blob/main/packages/nimble-components/src/rich-text-mention/users/view/styles.ts#L15-L18

One question is are some of those styles needed, i.e. the display inline-block configuration and the no-wrap configuration.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -216,7 +216,14 @@ function createCustomMentionExtension(
     }).configure({
         suggestion: {
             char: config.character,
-            decorationTag: config.viewElement,
+            /**
```
rajsite, line 219: Based on the other comment I have a general concern that the rich-text-editor and the rich-text-viewer have too much knowledge of the user mention implementation.

for example:
rich-text\editor\testing\rich-text-editor.pageobject.ts
rich-text\editor\testing\rich-text-editor-utils.ts
rich-text\editor\testing\types.ts

It should be organized so that anything specific to user mentions are handled from those folders instead (rich-text-mention\users\testing\*), the rich-text\ directory should not have any dependencies on specific mention implementations. **We should be able to completely delete user mentions without it impacting tests \ implementation for rich-text-viewer and rich-text-editor**.

The parts related to this PR are specifically styling changes. So this concern could be captured as a separate tech debt issue and addressed separately.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
vivinkrishna-ni: > However I don't see any style related changes in the PR.

The decoration tag's (view element with no `disable-editing` attribute) font styles during editing are the same as the other text which is `bodyFont` styles. So I didn't add any specific styles to it. 

> If state needs to be communicated to children it is done so via properties

Our initial idea is also the same to use a property `disabled` for the view element whenever the editor is disabled. However, since the view element is not bound to the template of the editor we couldn't find a better way for that query to select from the content of the editor and change the disabled state of the view element when the editor is disabled. Link to the discussion where we decided to go with the current approach: https://github.com/ni/nimble/pull/1683#discussion_r1402554313

> are some of those styles needed, i.e. the display inline-block configuration and the no-wrap configuration.

display: inline-block - if it is removed for the view element (`disable-editing` attribute added) from the `:host`, we are facing an issue in the Chrome browser ie. When `@mention` is added at the end of the markdown string via `setMarkdown()`, we can't keep the cursor at the end of the content. 

white-space: pre-wrap - it fixes the following issues as the white space is preserved:
1. Filtering with space [BUG 2602624](https://dev.azure.com/ni/DevCentral/_workitems/edit/2602624)
2. Continuous @ with space in between is not triggering the mention list box. [Bug 2603492](https://dev.azure.com/ni/DevCentral/_workitems/edit/2603492)
3. Popup not opening for odd spaces

rajsite: I got this notification so some quick thoughts I can elaborate on tomorrow:

An assumption that keeps happening is that the user mention behavior applies to all mentions. Stop assuming that is true. Maybe system / bug / emoji / test mentions won't use the body font color but will use red text. That's the point of the encapsulation of the view in a separate web component. Each mention gets to define it's own style and some interactions.

So hopefully it's clearer why your response that user mention views render with the body font, so all mentions will now too is insufficient. And also why I brought up the concern and saying "well we said it was okay in the other PR" was insufficient. That PR was huge and imperfect, that's why it had 10 follow up bugs right? 😋

vivinkrishna-ni: > An assumption that keeps happening is that the user mention behavior applies to all mentions.

I definitely agree with this and that is the reason why we are still using the view element after committing the mention. So for user mentions, it is now a green text that is encapsulated within the shadow DOM, and for other mentions, we can create separate elements and use them.

The current proposal is only removing the view element during the editing and not after committing in the editor or rendering in the viewer. For example, if the user starts typing `@john`, the element will be span and after clicking from the suggestion popup, it becomes `nimble-rich-text-mention-users-view`. So I couldn't get a case where the styles and interactions will be different for different mentions during editing. I quickly checked in the AzDo and GitHub that during editing, the style remains the same for all mentions but it is different only after committing. 

I can understand the goal of our current design of encapsulating the styles for different view mention elements in both the state: during editing and after committing the mentions. However, due to the browser-specific issues and the limitation of dynamic updation of properties (like `disabled`) in the view element in Tiptap, we couldn't make it possible. I might be misunderstanding your concern so we are planning to schedule a call tomorrow to discuss it in detail along with other few bugs.

vivinkrishna-ni: We found a way to utilize the `disabled` attribute instead of setting the private CSS variable for each mention configuration. I have added a `disabled` attribute to each mention view node whenever the `disabled` is set in the editor. In this way, each mention can have its own font color so that the editor need not be aware of all mention styling. The UI tests are already converted for mention view element in a disabled state. Kindly review the newly added changes and let me know your comments.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
vivinkrishna-ni: Created a tech debt and will be taken up in the follow-up PR: https://github.com/ni/nimble/issues/1724

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
rajsite: To round out the current PR:

There was one spot in the explicit CSS clean-ups I pointed out that was missed, specifically where richTextMentionUsersViewTag styles are targeted. That should be removed:

https://github.com/ni/nimble/blob/fb5646e5f3b548244fdca1ee4040d97e83712152/packages/nimble-components/src/rich-text/editor/styles.ts#L84-L86

My expectation with this change which switches from per mention type decorationTags to using a common decorationTag for all mentions (i.e. the default span) is that the rich text editor styles that shared tag explicitly. Right now it's relying on font inheritance and being a child span. I'd prefer it be more explicit.

So with this workaround we should set a `decorationClass: 'nimble-mention-view-edit'` so that we can style it explicitly instead of implicitly and in the rich text editor styles I'd expect something like:

```
   ${'' /* Shared styles for all mention views at edit time.
             Can be removed when edit time mention views can be used cross-browser.
             See: https://github.com/ni/nimble/issues/1716 */ }
   .nimble-mention-view-edit {
      color: ${bodyFontColor};
      display: inline-block; /* Do we need it? Your response seems to indicate we do but the span in your PR that is emitted is configured for inline*/
      white-space: /* Similarly your comment seems to suggest that white-space pre-wrap fixes a bunch of issues but your span in this PR is not configured for white-space pre-wrap? Is that an oversight or is that not needed for the span?*/
   }
```

See how the current PR is not configuring the styles your response seems to state is needed:

> display: inline-block: if it is removed for the view element [...] we are facing an issue in the Chrome browser

Your change in this PR is setting the view element (span) to inline (the default) does that mean there is a problem in the chrome browser?

image

> white-space: pre-wrap - it fixes the following issues as the white space is preserved

Your change in this PR is setting the view element (span) to a different white-space mode (the default, preserve) does that mean the three issues are being re-introduced?

image


---

Separate discussion:

> I quickly checked in the AzDo and GitHub that during editing, the style remains the same for all mentions but it is different only after committing.
> 
> I can understand the goal of our current design of encapsulating the styles for different view mention elements in both the state: during editing and after committing the mentions. However, due to the browser-specific issues in the view element in Tiptap, we couldn't make it possible. I might be misunderstanding your concern so we are planning to schedule a call tomorrow to discuss it in detail along with other few bugs.

To be clear, my concern in this PR is that the code aligns with the behavior. Right now custom user mention views are used for both edit and after committing. So my feedback has been to make that function correctly. I'm not opposed to the idea that during mention editing we want all mentions to look the same and it is only after committing that they look different. In that scenario the changes I expect are:

- We remove the `disable-editing` attribute from the `RichTextMentionView` base class and any logic around configuring that attribute
- We add a `decorationClass` to the suggestion config so that we can target the span directly, something like `nimble-mention-view-edit`
- We update the styles for rich-text-editor to target that class and explicit configure the styling as needed:
   ```
   .nimble-mention-view-edit {
       color: ${bodyFontColor};
       /// etc
   } 
   ```
- We remove the todo and close associated issues like #1716 

If that is a direction we want to go that should be done in a follow-up PR as that is a bigger change in direction (intentionally not using the custom view for edit time but only after commit).

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -586,6 +589,30 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
     }
 
+    private disableMentionViewElement(): void {
+        this.tiptapEditor.state.doc.descendants((node, pos) => {
+            if (node.type.name.startsWith(mentionPluginPrefix)) {
+                const updatedAttrs = {
+                    href: node.attrs.href as string,
+                    label: node.attrs.label as string,
+                    disabled: this.disabled ? '' : null
+                };
```
rajsite, line 599: Can you just override the disabled property instead of manually picking them?

```suggestion
                const updatedAttrs = {
                    ...node.attrs,
                    disabled: this.disabled ? '' : null
                };
```

My concern is that it is not clear which attributes need to be kept in sync.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
rajsite: > We found a way to utilize the disabled attribute instead of setting the private CSS variable for each mention configuration.

Good change! It will be needed regardless of which direction we go for edit time mention views. It's needed for committed mention views either way.

change/@ni-nimble-components-879a3c50-e10c-47fc-82c4-1e6dd51006d8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for Safari browser issue in removing the suggestion text when pressing backspace",
```
jattasNI, line 3:
```suggestion
  "comment": "Rich text editor fix for Safari browser issue in removing the suggestion text when pressing backspace",
```

jattasNI (commented):
It looks like Milan has done a thorough job of understanding the approach and providing feedback and I don't think I'll have anything to add, so I'll defer to his judgment on this PR. No need to wait for my approval.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
vivinkrishna-ni: I have created a separate PR in which I have updated the `decorationClass` and removed `disable-editing`. Please review and let me know your comments. https://github.com/ni/nimble/pull/1731

> Your change in this PR is setting the view element (span) to inline (the default) does that mean there is a problem in the chrome browser?

Actually, the reason for setting inline to view element (span) is that we are toggling the view based on the `disable-editing` setting in the CSS styles. The span is set to display: none in edit mode and if the `disable-editing` is enabled, we set the display back to inline so that it is visible. Since we are removing the `disable-editing` attribute, we need not set inline to span(as it is the default). I have updated the styles in the follow-up PR linked above.

But in the `:host` of the view element, we need to set it to inline-block so that the problem in the Chrome browser is resolved. 

> Your change in this PR is setting the view element (span) to a different white-space mode (the default, preserve) does that mean the three issues are being re-introduced?

The white-space setting is required only if we are using the view element at the time of editing which causes the three issues I mentioned above. But it is not required as we are not using the view element for editing. I have updated in the above-linked follow-up PR and verified that the mentioned bugs are not reproducible.

packages/nimble-components/src/rich-text/editor/index.ts
vivinkrishna-ni: Yes, we can do that. I have updated it as suggested.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
vivinkrishna-ni, line 1: We couldn't find a better way to write the test for the bug linked in the PR description as the backspace keyboard event is not emitting. We can delete the last character by utilizing the `sliceEditorContent()` page object method but that will remove the last character even if we use our custom decoration tag. So we cannot reproduce the bug programmatically in WebKit test cases. We are checking the tag anyway in other tests and verifying the class in the follow-up PR which ensures the given class is updated as expected. So I am marking the PR ready for review as owners already started reviewing.

packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
rajsite: looks like a good split! thanks for all the follow-up

rajsite (approved):

PR #1722, opened 2024-01-02
Create enum for z-index values

### LINKED ISSUE #539: Create global enum for managing z-indexes
## 🧹 Tech Debt
We should create global enum for managing z-indexes rather than arbitrarily setting z-index values in our components.

This idea was brought up here: https://github.com/ni/nimble/pull/530#discussion_r865118893

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #539 

## 👩‍💻 Implementation

- Created a new enum in `utilities` with a few preset z-index values.
- Updated stylesheets for anchored-region, dropdown, and table to use enum values rather than hardcoded values.

## 🧪 Testing

None

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-bb93d4a0-4af8-4624-9783-709125113d4f.json
+++ b/change/@ni-nimble-components-bb93d4a0-4af8-4624-9783-709125113d4f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Create enum for z-index values",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/docs/css-guidelines.md
+++ b/packages/nimble-components/docs/css-guidelines.md
@@ -217,4 +217,4 @@ This utility will generate the appropriate display style, as well as a style rul
 
 ## Stick to enumerated values for `z-index`
 
-To avoid an unnecessary proliferation of `z-index` values (which make the code more difficult to reason about), we define a `ZIndexLevels` enum with a fixed set of values to choose from. If possible, use one of the existing values. If you instead need to establish a new stacking position relative to the existing values, create a new enum value to use. When possible, pick a value that is a power of 10.
+To avoid an unnecessary proliferation of `z-index` values (which make the code more difficult to reason about), we define a `ZIndexLevels` enum with a fixed set of values to choose from. If possible, use one of the existing values. If you instead need to establish a new stacking position relative to the existing values, create a new enum value to use.
--- a/packages/nimble-components/docs/css-guidelines.md
+++ b/packages/nimble-components/docs/css-guidelines.md
@@ -214,3 +214,7 @@ export const styles = css`
 ```
 
 This utility will generate the appropriate display style, as well as a style rule to hide the host element when its `hidden` attribute is set.
+
+## Stick to enumerated values for `z-index`
+
+To avoid an unnecessary proliferation of `z-index` values (which make the code more difficult to reason about), we define a `ZIndexLevels` enum with a fixed set of values to choose from. If possible, use one of the existing values. If you instead need to establish a new stacking position relative to the existing values, create a new enum value to use. When possible, pick a value that is a power of 10.
--- a/packages/nimble-components/src/patterns/dropdown/styles.ts
+++ b/packages/nimble-components/src/patterns/dropdown/styles.ts
@@ -25,7 +25,6 @@ import { focusVisible } from '../../utilities/style/focus';
 import { themeBehavior } from '../../utilities/style/theme';
 import { DropdownAppearance } from './types';
 import { userSelectNone } from '../../utilities/style/user-select';
-import { ZIndexLevels } from '../../utilities/style/types';
 
 export const styles = css`
     ${display('inline-flex')}
@@ -143,12 +142,10 @@ export const styles = css`
         box-sizing: border-box;
         display: inline-flex;
         flex-direction: column;
-        left: 0;
         overflow-y: auto;
         width: 100%;
         --ni-private-listbox-padding: ${smallPadding};
         max-height: calc(var(--ni-private-select-max-height) - ${smallPadding});
-        z-index: ${ZIndexLevels.justAboveDefault};
         box-shadow: ${elevation2BoxShadow};
         border: 1px solid ${popupBorderColor};
         background-color: ${applicationBackgroundColor};
--- a/packages/nimble-components/src/patterns/dropdown/styles.ts
+++ b/packages/nimble-components/src/patterns/dropdown/styles.ts
@@ -25,6 +25,7 @@ import { focusVisible } from '../../utilities/style/focus';
 import { themeBehavior } from '../../utilities/style/theme';
 import { DropdownAppearance } from './types';
 import { userSelectNone } from '../../utilities/style/user-select';
+import { ZIndexLevels } from '../../utilities/style/types';
 
 export const styles = css`
     ${display('inline-flex')}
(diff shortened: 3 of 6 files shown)


packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -142,12 +142,10 @@ export const styles = css`
         box-sizing: border-box;
         display: inline-flex;
         flex-direction: column;
-        left: 0;
         overflow-y: auto;
         width: 100%;
         --ni-private-listbox-padding: ${smallPadding};
         max-height: calc(var(--ni-private-select-max-height) - ${smallPadding});
-        z-index: 1;
```
m-akinc, line 150: This has no effect because the `position` of the element is `static` (the default).

m-akinc:
@atmgrifter00 could you please buddy this?

packages/nimble-components/src/utilities/style/types.ts
```diff
@@ -0,0 +1,8 @@
+export const ZIndexLevels = {
+    justBelowDefault: '-1',
+    justAboveDefault: '1',
+    high: '100',
+    higher: '1000',
+    veryHigh: '10000'
```
atmgrifter00, line 6: I'm struggling with the proposed names, because they don't quite convey an appropriate relative meaning to each other. For instance, `higher` could be thought to be "higher" than `veryHigh`. Ultimately, I think the goal of the enum is just to provide a mechanism so that we don't set a `z-index` to some arbitrary value. This means we could get _real_ specific with naming like `defaultPlus1`, `defaultPlus10`, etc... (maybe). I'm not sure if I have any better suggestions, but if we went with the current proposed scheme, I would think we want something where the largest value is called `highest`.

atmgrifter00 (approved):

packages/nimble-components/src/utilities/style/types.ts
```diff
@@ -0,0 +1,5 @@
+export const ZIndexLevels = {
```
jattasNI, line 1: Part of [the discussion](https://github.com/ni/nimble/pull/530#discussion_r865118893) linked from the issue mentions goals of either naming or documenting the rationale behind the values chosen so that users of the enum can have a better idea of how to choose between the values. Can you take a stab at that or make a case for not doing it?

change/@ni-nimble-components-bb93d4a0-4af8-4624-9783-709125113d4f.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Can you update [CSS Guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md) to establish a policy that components should always use the new enum rather than a literal value when setting `z-index`?

packages/nimble-components/src/utilities/style/types.ts
m-akinc: Sorry, I should have given a rationale for not adding documentation. I could not find any z-index value of 200 used by a grid (or anything else) as referenced in that discussion thread. I did not find any other meaningful z-index relationships that we might codify into enum names or document. There isn't any guidance to give users of the enum. They just have to pick a numeric value (which is part of the enum value's name) based on the z-index of whatever other element(s) they're concerned with. However, as your other comment pointed out, it would be worth adding something to our CSS Guidelines about needing to use an enum value, and how to create a new one if necessary.

rajsite: > There isn't any guidance to give users of the enum

So my question is are there z-index values that our apps are using to fight with parts of nimble-components. And that can result in two follow-up questions:
1. What will break with this PRs changes to enum values?
2. Do we need to export enum values that can be used externally? In which case they would be exported as CSS design tokens so users can reliably position things in stacks against our components. If that's the case I don't really know what it would look like, but might need things like tableZIndex and tableZIndexPlus10 or something.

I didn't claim to so a thorough search of how users may be impacted but here is at least one thing that looks fishy:

This is a search for [z-index in Skyline /Web](https://dev.azure.com/ni/_search?action=contents&text=z-index&type=code&lp=custom-Collection&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7DBranchFilters%7Bmaster%7DPathFilters%7B/Web%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/SystemLinkShared/projects/systemlink-repository-lib/src/repository/components/store-downloads-table/store-downloads-table.component.scss) and that linked example is for a component using a nimble table that has some kind of wrapper and z-indexes set.

m-akinc: > are there z-index values that our apps are using to fight with parts of nimble-components?

It's hard to understand why `z-index` is being set in many of those use cases. I don't see any evidence of them being used to "one-up" any Nimble components with explicit `z-index` values.

> 1. What will break with this PR's changes to enum values?

Not sure I understand. This PR does not change any `z-index` values of any elements. And there are no pre-existing enum values being changed.

> 2. Do we need to export enum values that can be used externally?

I don't think so, given that I don't see (or anticipate) a client need to stack elements in relation to anything in a Nimble component's shadow DOM. If there were such a need, then we would want semantically meaningful tokens, like `popUpMenuZIndex` or `tooltipZIndex` rather than the generic, number-based names (`zIndex1000`) of the enum.

rajsite: > This PR does not change any z-index values of any elements. And there are no pre-existing enum values being changed.

True, doesn't change. Think eventually the dropdown and and anchored region will be replaced by popover api usage so most of z-index usage will be gone at that point.

> I don't see any evidence of them being used to "one-up" any Nimble components with explicit z-index values.

I'll take your evaluation of it. May consider some validation after merging.

packages/nimble-components/docs/css-guidelines.md
```diff
@@ -213,4 +213,8 @@ export const styles = css`
 `;
 ```
 
-This utility will generate the appropriate display style, as well as a style rule to hide the host element when its `hidden` attribute is set.
\ No newline at end of file
+This utility will generate the appropriate display style, as well as a style rule to hide the host element when its `hidden` attribute is set.
+
+## Stick to enumerated values for `z-index`
+
+To avoid an unnecessary proliferation of `z-index` values (which make the code more difficult to reason about), we define a `ZIndexLevels` enum with a fixed set of values to choose from. If possible, use one of the existing values. If you instead need to establish a new stacking position relative to the existing values, create a new enum value to use. When possible, pick a value that is a power of 10.
```
jattasNI, line 220: I'd vote to leave the values to dev / code owner discretion. Picking powers of 10 could result in unneccessarily big gaps and hard-to-read counting of zeros (is that 1000000 or 10000000?)

```suggestion
To avoid an unnecessary proliferation of `z-index` values (which make the code more difficult to reason about), we define a `ZIndexLevels` enum with a fixed set of values to choose from. If possible, use one of the existing values. If you instead need to establish a new stacking position relative to the existing values, create a new enum value to use.
```

jattasNI (approved):

packages/nimble-components/src/utilities/style/types.ts
```diff
@@ -0,0 +1,5 @@
...
+    zIndex1: '1',
+    zIndex1000: '1000'
```
rajsite, line 3: Looking back at the original comment in https://github.com/ni/nimble/pull/530#discussion_r863076750, the reason this is 1000 is because it is a higher value than the [jqx grid](https://github.com/ni-kismet/jqx/blob/4b72d63749befd559576ecb258fbc6ad9a414b18/source/styles/default/jqx.table.css#L1033)

So we should include that context here

```suggestion
    // Original usages of 1000 were to compete with jqx-grid usage of z-index 200 for headers
    // See: https://github.com/ni/nimble/pull/530#discussion_r863076750
    zIndex1000: '1000'
```

m-akinc: Ah, so that's where it was. I had searched for `z-index` being set to 200 somewhere and assumed it had been removed/changed since the comment was made. Didn't think to look in jqx code.

Issue #503, opened 2022-04-18
Clean-up Flex layout align configuration


rajsite:
## 🧹 Tech Debt

We had found icons and controls were not aligning property. One workaround was to add the vertical-align: middle property on the button as show here: https://github.com/ni/nimble/blob/47839b3c67586cb98051a2639d9ba2cb6d21b5a4/packages/nimble-components/src/patterns/button/styles.ts#L37

In the FAST issue they found an alternate pattern that just changes flex settings: https://github.com/microsoft/fast/issues/5414#issuecomment-1100791752

We should verify that configuration works and propagate it to our controls where it makes sense. We should also add tests focused on validating inline behavior of the controls (ie inline with text and inline-block images, etc).

m-akinc:
FAST's approach results in alignment that is inconsistent with native `button` elements.

FAST's approach (updated stackblitz to latest FAST packages):
![image](https://github.com/ni/nimble/assets/7282195/6fe399d7-da13-495b-b829-2c9f408ec257)

Nimble's current approach (setting `vertical-align: center` on host):
![image](https://github.com/ni/nimble/assets/7282195/e9b5ea63-8e57-40d1-8f5c-d61a3e1d1c2c)

Native button element:
![image](https://github.com/ni/nimble/assets/7282195/58aa5a66-bb29-47db-abb3-aad14ec0081e)

I tried modifying FAST's approach by setting the inner, native button's `align-items` to something other than `baseline`, but that results in misaligned buttons (i.e. the original problem):
![image](https://github.com/ni/nimble/assets/7282195/7bc595c0-740a-4896-8331-3c8e9682473b)

My conclusion is that we should keep our current approach. I will still create a new Chromatic test for button alignment, though.

PR #1725, opened 2024-01-03
Add button alignment visual test

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes: #503 

## 👩‍💻 Implementation

Added a button Chromatic test for how buttons align with inline text and images.

## 🧪 Testing

New test added.

I noticed that the Safari snapshot of the new Chromatic test has unexpected alignment (top of text is aligned with top of buttons). From what I can tell, Chromatic is using version 16 of Safari. I checked the new story in versions 15.x and 17.1, and both rendered with the alignment I would expect (same as other browsers). I also tried it in Playwright's WebKit browser version 16.0, and I still got the expected alignment. I do not know how Chromatic ended up with weird alignment, but I'm satisfied enough that it looks right in all versions of Safari/WebKit I checked.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f4776f02-4ee0-4534-8a00-bbc51b09d9c3.json
+++ b/change/@ni-nimble-components-f4776f02-4ee0-4534-8a00-bbc51b09d9c3.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add button alignment visual test",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/packages/nimble-components/src/button/tests/button-matrix.stories.ts
+++ b/packages/nimble-components/src/button/tests/button-matrix.stories.ts
@@ -82,10 +82,22 @@ export const textCustomized: StoryFn = createMatrixThemeStory(
     textCustomizationWrapper(html`<${buttonTag}>Button</${buttonTag}>`)
 );
 
+// The baseline of the elements should be aligned when positioned inline.
 export const inlineAlignment: StoryFn = createStory(
-    html`Text
-        <${buttonTag}>Button</${buttonTag}>
-        Text
-        <${buttonTag}><${iconKeyTag} slot="start"></${iconKeyTag}>Button</${buttonTag}>
-        <div style="display:inline-block; width:50px; height:50px; border: 1px black solid;"></div>`
+    html`<span style="text-decoration: underline;">Text</span>
+        <nimble-button
+            ><span style="text-decoration: underline;"
+                >Button</span
+            ></nimble-button
+        >
+        <span style="text-decoration: underline;">Text</span>
+        <nimble-button
+            ><nimble-icon-key slot="start"></nimble-icon-key
+            ><span style="text-decoration: underline;"
+                >Button</span
+            ></nimble-button
+        >
+        <div
+            style="display:inline-block; width:50px; height:50px; border: 1px black solid;"
+        ></div>`
 );
--- a/packages/nimble-components/src/button/tests/button-matrix.stories.ts
+++ b/packages/nimble-components/src/button/tests/button-matrix.stories.ts
@@ -16,6 +16,7 @@ import { textCustomizationWrapper } from '../../utilities/tests/text-customizati
 import { buttonTag } from '..';
 import { iconKeyTag } from '../../icons/key';
 import { iconArrowExpanderDownTag } from '../../icons/arrow-expander-down';
+import { bodyFont } from '../../theme-provider/design-tokens';
 
 const metadata: Meta = {
     title: 'Tests/Button',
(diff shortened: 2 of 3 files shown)


packages/nimble-components/src/button/tests/button-matrix.stories.ts
```diff
@@ -81,3 +81,11 @@ export const hiddenButton: StoryFn = createStory(
 export const textCustomized: StoryFn = createMatrixThemeStory(
     textCustomizationWrapper(html`<${buttonTag}>Button</${buttonTag}>`)
 );
+
+export const inlineAlignment: StoryFn = createStory(
+    html`Text
+        <${buttonTag}>Button</${buttonTag}>
+        Text
+        <${buttonTag}><${iconKeyTag} slot="start"></${iconKeyTag}>Button</${buttonTag}>
+        <img src="https://github.com/ni/nimble/raw/main/docs/nimble-logo-icon.svg" width="50px" style="border: 1px black solid">`
```
rajsite, line 90: Don't couple to other arbitrary resources not owned by the test for no reason. If we move or rename that logo that would change the behavior of the tests.

Can you just do something like:

```suggestion
        <div style="display:inline-block; width:50px; height:50px; border: 1px black solid;"></div>`
```

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -47,6 +47,7 @@ export const styles = css`
                     Not sure why but this is needed to get buttons with icons and buttons
                     without icons to align on the same line when the button is inline-flex
                     See: https://github.com/microsoft/fast/issues/5414
+                    For additional discussion, see: https://github.com/ni/nimble/issues/503
```
rajsite, line 50: Don't need to link to both, just add useful info to our issue

```suggestion
                    See: https://github.com/ni/nimble/issues/503
```

rajsite (changes requested):

change/@ni-nimble-components-f4776f02-4ee0-4534-8a00-bbc51b09d9c3.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: As you pointed out offline, the Safari test is rendering incorrectly in Chromatic but correctly in local tests with both newer and older versions of Safari. It'd be good to update the PR description to mention that we noticed this and decided to accept it (or do something else, if you come up with an alternate resolution).

jattasNI, line 1: I haven't taken the time to fully understand the issue and its history, but why are we only focused on buttons? Would the same issue affect other components? If so, is adding that test coverage considered part of #503?

m-akinc: It was only with buttons that an alignment issue came up. [When adding icon support to the button](https://github.com/ni/nimble/pull/124), an alignment issue was discovered and worked around. It would make sense to add tests for other controls if we had some suspicion/history of them having a similar alignment issue. The only other control that supports icons and can be laid out in a row is the text field, and I verified that it doesn't have an alignment issue.

packages/nimble-components/src/button/tests/button-matrix.stories.ts
```diff
@@ -81,3 +81,11 @@ export const hiddenButton: StoryFn = createStory(
 export const textCustomized: StoryFn = createMatrixThemeStory(
     textCustomizationWrapper(html`<${buttonTag}>Button</${buttonTag}>`)
 );
+
+export const inlineAlignment: StoryFn = createStory(
```
rajsite, line 85: I think the test will make more sense to future visitors if there was a comment along the lines of "The baseline of the elements should be aligned when positioned inline" and then if we visualized the underline:

```ts
<span style="text-decoration: underline;">Text</span>
<nimble-button current-value="" appearance="outline"><span style="text-decoration: underline;">Button</span></nimble-button>
<span style="text-decoration: underline;">Text</span>
<nimble-button appearance="outline"><nimble-icon-key slot="start"></nimble-icon-key><span style="text-decoration: underline;">Button</span></nimble-button>
<div style="display:inline-block; width:50px; height:50px; border: 1px black solid;"></div>
```

image

rajsite: Although that doesn't really align that great. I wonder if it is better if the span text font between the buttons is the same as the buttonLabelFont

rajsite (approved):

change/@ni-nimble-components-f4776f02-4ee0-4534-8a00-bbc51b09d9c3.json
rajsite: The fonts in the plain text spans around the buttons are also different across browsers in chromatic. Wonder if using bodyFont on them will make them more consistent in all the screenshots.

Can wrap the whole thing in a span and set the font on the wrapper span since font is inherited.

packages/nimble-components/src/button/tests/button-matrix.stories.ts
```diff
@@ -81,3 +81,23 @@ export const hiddenButton: StoryFn = createStory(
 export const textCustomized: StoryFn = createMatrixThemeStory(
     textCustomizationWrapper(html`<${buttonTag}>Button</${buttonTag}>`)
 );
+
+// The baseline of the elements should be aligned when positioned inline.
+export const inlineAlignment: StoryFn = createStory(
+    html`<span style="text-decoration: underline;">Text</span>
+        <nimble-button
```
jattasNI, line 88: we should use `buttonTag` rather than `nimble-button` to get automatic updates in the unlikely event of a name change and to get a proper import tree if we ever move this story somewhere else.

jattasNI: Same for the icon element.

jattasNI (approved):

PR #1726, opened 2024-01-03
In anchor story, only emit contenteditable when set

### LINKED ISSUE #1723: Nimble anchor storybook page over-emphasizes contenteditable configuration
## 🧹 Tech Debt

The nimble-anchor story always emits the contenteditable configuation in the example snippet:
image

While not wrong, it adds extra noise to the snippet and makes it easy to copy and paste unnecessarily, for example see this stack enterprise post: https://ni.stackenterprise.co/q/15655/11

image

The story should be modified so that the attribute is only emitted when set to true and is absent when false.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1723

## 👩‍💻 Implementation

Treat `contenteditable` setting as a boolean rather than a string, and omit the attribute when the setting is false.

## 🧪 Testing

Storybook

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/anchor/tests/anchor.stories.ts
+++ b/packages/nimble-components/src/anchor/tests/anchor.stories.ts
@@ -15,7 +15,7 @@ interface AnchorArgs {
     label: string;
     href: string;
     underlineHidden: boolean;
-    contenteditable: string;
+    contenteditable: boolean;
     appearance: keyof typeof AnchorAppearance;
 }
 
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/anchor/tests/anchor.stories.ts
```diff
@@ -61,8 +61,6 @@ const metadata: Meta<AnchorArgs> = {
                 'Set to `prominent` to make the anchor appear in a different color than normal text.'
         },
         contenteditable: {
-            options: ['false', 'true'],
-            control: { type: 'radio' },
             description:
                 'Set this to the string "true" (or set the attribute without any value)  when the anchor is within an editable region (i.e. element/hierarchy with [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable)). Whereas native elements inherit their `contenteditable` value by default, the `nimble-anchor` requires this attribute be explicitly set.'
```
rajsite, line 65:
```suggestion
                'Set this to the string "true" (or an empty string)  when the anchor is within an editable region (i.e. element/hierarchy with [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable)). Whereas native elements inherit their `contenteditable` value by default, the `nimble-anchor` requires this attribute be explicitly set.'
```

rajsite (approved):

m-akinc:
Oops, @jattasNI I absent-mindedly set auto-merge before you reviewed. If you see anything you want changed, though, I'll open a new PR.

PR #1728, opened 2024-01-03
Update Playwright version

### LINKED ISSUE #1544: Uptake newer Playwright version and fix breaking change
## 🧹 Tech Debt

Milan pointed out in [this Renovate PR](https://github.com/ni/nimble/pull/1542#discussion_r1327685887) that our version of Playwright cannot currently be updated automatically.
1. The versions published for npm and Nuget don't align
2. There was a breaking change between 1.37 and 1.38 whereby browsers are no longer included in the pacakge, but we need them.

We should:
1. manually update Playwright to the latest version that makes sense for us
2. fix how we install browsers (see [playwright release notes](https://github.com/microsoft/playwright/releases/tag/v1.38.0) for details)
3. consider letting Renovate manage the Playwright dependency again

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1544 

## 👩‍💻 Implementation

- Added `postinstall` script to root package file that runs `playwright install --with-deps` to ensure browsers get installed.
- Bumped dependency versions of Playwright to 1.40.0 in nimble-components, nimble-blazor, and Angular workspace. They are now pinned to one version, because the root workspace package file now installs the browsers.
- Fixed two table page object functions that were incorrect.
- Added workaround (rounding cell width) for tests failing due to pulling in new version of Chrome browser (from 115.0.5790.75 to 120.0.6099.28). Six table interactive column resize tests began failing with errors like `Expected 100.984375 to be 101.` and `Expected 74.984375 to be 75.` I confirmed that this numerical imprecision is related to using fractional width units (e.g. `1fr`).

We still cannot have Renovate update the Playwright dependency because of the inconsistency of versions between the npm package and the nuget. E.g. currently the latest npm package is 1.40.1, but the latest nuget is only 1.40.0, so our build will fail as it tries to find a nuget with the same version as the npm package.

## 🧪 Testing

Pipeline build succeeds

## ✅ Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,6 @@ jobs:
       # See: https://github.com/actions/setup-node/issues/529
       - run: npm i -g npm@9
       - run: npm ci
-      - run: npx playwright install --with-deps
 
       # print a log of outdated npm dependencies
       # only informational so swallow error codes
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,7 @@ jobs:
       # See: https://github.com/actions/setup-node/issues/529
       - run: npm i -g npm@9
       - run: npm ci
+      - run: npx playwright install
 
       # print a log of outdated npm dependencies
       # only informational so swallow error codes
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
       # See: https://github.com/actions/setup-node/issues/529
       - run: npm i -g npm@9
       - run: npm ci
-      - run: npx playwright install
+      - run: npx playwright install --with-deps
 
       # print a log of outdated npm dependencies
       # only informational so swallow error codes
--- a/angular-workspace/package.json
+++ b/angular-workspace/package.json
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.2.2",
-    "playwright": "1.40.1",
+    "playwright": "1.40.0",
     "rollup": "^3.10.1",
     "typescript": "~4.8.2"
   }
--- a/angular-workspace/package.json
+++ b/angular-workspace/package.json
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.2.2",
-    "playwright": "^1.40.0",
+    "playwright": "1.40.1",
     "rollup": "^3.10.1",
     "typescript": "~4.8.2"
   }
--- a/angular-workspace/package.json
+++ b/angular-workspace/package.json
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.2.2",
-    "playwright": "^1.30.0",
+    "playwright": "^1.40.0",
     "rollup": "^3.10.1",
     "typescript": "~4.8.2"
   }
--- a/change/@ni-nimble-components-ee39f42f-cd28-42b3-837e-0df5b4842e70.json
+++ b/change/@ni-nimble-components-ee39f42f-cd28-42b3-837e-0df5b4842e70.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Update Playwright version",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "karma-jasmine": "^5.1.0",
         "karma-jasmine-html-reporter": "^2.0.0",
         "ng-packagr": "^15.2.2",
-        "playwright": "^1.30.0",
+        "playwright": "^1.40.0",
         "rollup": "^3.10.1",
         "typescript": "~4.8.2"
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -36268,7 +36268,7 @@
         "@rollup/plugin-node-resolve": "^15.0.1",
         "cross-env": "^7.0.3",
         "glob": "^8.1.0",
-        "playwright": "1.36.0",
+        "playwright": "^1.40.0",
         "rollup": "^3.10.1"
       }
     },
@@ -36307,34 +36307,6 @@
       },
       "engines": {
         "node": ">=10"
-      }
-    },
-    "packages/nimble-blazor/node_modules/playwright": {
-      "version": "1.36.0",
-      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.36.0.tgz",
-      "integrity": "sha512-ODVOTp5WoRMxDJY3liMmC+wVNicc0cQB17gASvQ+zLBggVK9a2x3gdT5mbl7av+zomvtdirWOqqfD+O6qIrFgw==",
-      "dev": true,
-      "hasInstallScript": true,
-      "dependencies": {
-        "playwright-core": "1.36.0"
-      },
-      "bin": {
-        "playwright": "cli.js"
-      },
-      "engines": {
-        "node": ">=16"
-      }
-    },
-    "packages/nimble-blazor/node_modules/playwright-core": {
-      "version": "1.36.0",
-      "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.36.0.tgz",
-      "integrity": "sha512-7RTr8P6YJPAqB+8j5ATGHqD6LvLLM39sYVNsslh78g8QeLcBs5750c6+msjrHUwwGt+kEbczBj1XB22WMwn+WA==",
-      "dev": true,
-      "bin": {
-        "playwright-core": "cli.js"
-      },
-      "engines": {
-        "node": ">=16"
       }
     },
     "packages/nimble-components": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,8 @@
         "@lhci/cli": "^0.12.0",
         "beachball": "^2.31.0",
         "cross-env": "^7.0.3",
-        "patch-package": "^6.4.7"
+        "patch-package": "^6.4.7",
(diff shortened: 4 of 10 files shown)


m-akinc:
@msmithNI could you please buddy this change?

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -564,10 +564,10 @@ describe('Table Interactive Column Sizing', () => {
                 0,
                 1
             );
-            pageObject.dragSizeColumnByRightDivider(0, [50]);
+            pageObject.dragSizeColumnByRightDivider(0, [5]);
```
msmithNI, line 567: Is this change related to the rounding issues too? I think I'm missing why you had to change the values for these tests.

m-akinc: Kind of. This test started failing due to the precision issue, but my workaround actually caused it (and the next test) to fail for a totally different issue. In these tests, `secondVisibleCellWidth` ends up being 133.3. It turns out that in js, 133.3 - 50 gives you 83.30000000000001. I guess previously `getCellRenderedWidth` was also returning 83.30000000000001, so the values were equal and the test passed. But now with my rounding workaround, `getCellRenderedWidth` returns 83.3, which != 83.30000000000001. By changing the delta value from 50 to 5, I avoid the numerical precision issue.

.github/workflows/main.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       # See: https://github.com/actions/setup-node/issues/529
       - run: npm i -g npm@9
       - run: npm ci
+      - run: npx playwright install --with-deps
```
msmithNI, line 44: I think I've gotten guidance to avoid using `npx` (in favor of creating new commands to invoke in package.json, like a `playwright-install-deps` command), but I don't remember the full justification. Can wait and see if @rajsite responds when he reviews.

msmithNI:
What's the behavior for devs locally once this goes in? Will the first `npm run tdd` command they run in nimble-components fail, until they do the `npx playwright install`? Assuming that message gets printed to the console that's probably fine, otherwise we may need to update the Contributing doc(s) to document that step.

msmithNI (approved):

change/@ni-nimble-components-ee39f42f-cd28-42b3-837e-0df5b4842e70.json
m-akinc, line 1: > What's the behavior for devs locally once this goes in? Will the first npm run tdd command they run in nimble-components fail, until they do the npx playwright install? Assuming that message gets printed to the console that's probably fine, otherwise we may need to update the Contributing doc(s) to document that step.

@msmithNI I tested this out by:
1. checked out `main`,
2. deleted the browsers from AppData/Local/ms-playwright,
3. ran `npm install`,
4. checked out my branch with updated Playwright version,
5. ran `npm install`,
6. ran `npm run tdd`

Surprisingly, step 6 ran fine, because step 5 actually updated/replaced the browsers in AppData/Local/ms-playwright. I was looking at the folder while `npm install` ran, and I saw the browsers being replaced. I can't explain why. Version 1.38+ of the `playwright` package [isn't supposed to download the browsers](https://playwright.dev/docs/release-notes#breaking-changes-playwright-no-longer-downloads-browsers-automatically) during `npm install`.

jattasNI: In [a Skyline PR setting up Playwright](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/617183#1704490486) we opted to install browsers in a `postinstall` script. We could consider doing that from Nimble's root `package.json` too, which would allow us to remove it from `main.yml` and the Blazor package.json.

I think that's be my default recommendation, though some counterarguments:
1. it moves us away from packages having independent builds (though `main.yml` changes do too)
2. it differs from [what other Skyline projects do](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=playwright%3Asetup&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//BuildTools/Pipeline/Templates/WebAppBuildLintTest.yml)

m-akinc: Moved browser installation to a `postinstall` script in the root package.json.

package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "description": "The NI Nimble Design System Monorepo",
   "scripts": {
-    "postinstall": "patch-package",
+    "postinstall": "patch-package && npm run playwright:setup",
+    "playwright:setup": "playwright install --with-deps",
```
rajsite, line 8: The playwright version should be captured in the devDependencies (this package.json is relying on that package to be installed but didn't declare a dependency on it, very fishy)

angular-workspace/package.json
```diff
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.2.2",
-    "playwright": "^1.30.0",
+    "playwright": "^1.40.0",
```
rajsite, line 58: playwright is a weird dep, unlike other packages where it is okay for the version to be different between packages, we need this to be the same in all packages in the monorepo (since the root package does extra work to fetch the package deps). Lets pin the version in all these packages to prevent a potential situation where the version is misaligned between packages.

package-lock.json
```diff
@@ -29156,25 +29156,27 @@
       }
     },
     "node_modules/playwright": {
-      "version": "1.36.1",
-      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.36.1.tgz",
-      "integrity": "sha512-2ZqHpD0U0COKR8bqR3W5IkyIAAM0mT9FgGJB9xWCI1qAUkqLxJskA1ueeQOTH2Qfz3+oxdwwf2EzdOX+RkZmmQ==",
+      "version": "1.40.1",
```
rajsite, line 29159: Yea you can see here that below under blazor that mutliple different versions of playwright are being resolved between packages. Should be fixed by pinning in all packages as described in the other comment

angular-workspace/package.json
```diff
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.2.2",
-    "playwright": "^1.30.0",
+    "playwright": "1.40.1",
```
rajsite, line 58: > We still cannot have Renovate update the Playwright dependency because of the inconsistency of versions between the npm package and the nuget. E.g. currently the latest npm package is 1.40.1, but the latest nuget is only 1.40.0, so our build will fail as it tries to find a nuget with the same version as the npm package.

https://www.nuget.org/packages/Microsoft.Playwright#versions-body-tab

rajsite (changes requested):

angular-workspace/package.json
```diff
@@ -55,7 +55,7 @@
     "karma-jasmine": "^5.1.0",
```
rajsite, line 55: make sure the pr description is updated

package.json
```diff
@@ -44,6 +45,7 @@
     "@lhci/cli": "^0.12.0",
     "beachball": "^2.31.0",
     "cross-env": "^7.0.3",
-    "patch-package": "^6.4.7"
+    "patch-package": "^6.4.7",
+    "playwright": "1.40.0"
```
jattasNI, line 49: Any thoughts on a policy for keeping this updated given that we can't rely on Renovate to do so?

1. Manually update it whenever we notice it's drastically out of date? 
2. Some kind of ongoing tech debt issue that reminds us to do it?
3. Policy to update it manually each month when completing Renovate PRs?

And any docs to write to help make the upgrade easier? Is it basically just "search package.json for playwright and manually change the version, making sure to keep them all aligned"?

jattasNI (approved):

package.json
m-akinc: How about I create a recurring reminder on my calendar (every 3 months?) to create a tech debt issue for updating Playwright?

I don't think docs are necessary. It's exactly what you said, and if someone does something different, it should be easy to notice in the PR.

jattasNI: If you're up for owning it, that sounds good to me! If it becomes too annoying we can always fall back to option 1.

rajsite: I added it to the Manual Regression Validations list: https://github.com/ni/nimble/issues/1747

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
rajsite: Could we instead keep the original setup and use the jasmine toBeCloseTo: https://jasmine.github.io/api/5.1/matchers#toBeCloseTo
Sounds like it may be less finicky if there are test changes. The floating point math scenario seems like a reasonable place for it.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -335,7 +335,11 @@ export class TablePageObject<T extends TableRecord> {
 
     public getCellRenderedWidth(rowIndex: number, columnIndex: number): number {
         const cell = this.getCell(rowIndex, columnIndex);
-        return cell.getBoundingClientRect().width;
+        const actualWidth = cell.getBoundingClientRect().width;
+        // Round to one decimal place. This is to work around a bug in Chrome related to
+        // fractional widths (e.g. '1fr') in grid layouts that results in some numerical
+        // precision issues. See: https://bugs.chromium.org/p/chromium/issues/detail?id=1515685
+        return Math.round(actualWidth * 10) / 10;
```
rajsite, line 342: We may be able to avoid this as well with toBeCloseTo

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
rajsite: eh, nvm, might require refactoring of more of the expects, can do it if it becomes flaky later

rajsite (approved):

rajsite:
Ah dang, we all missed it but testing/ is public apis so this should have been a patch change, not a none change. @jattasNI @msmithNI @m-akinc Hopefully the change in behavior doesn't impact someone else' next release.

PR #1729, opened 2024-01-03
Rich Text Editor | Fix for losing the state of format buttons when the configuration is updated dynamically

aagash-ni:
# Pull Request

## 🤨 Rationale

- Fix for  [Bug 2614292](https://dev.azure.com/ni/DevCentral/_workitems/edit/2614292): Editor format buttons active state (Bold and Italics) are not retained when user configuration gets loaded dynamically

- Currently, when there are changes in the configuration, the states of the bold and italics buttons are reverted to an inactive state. This is happening because of the tiptap API [setContent()](https://tiptap.dev/docs/editor/api/commands/set-content) which will reset the formatting button states when setting content.  This PR will address this issue by resetting its states. 

## 👩‍💻 Implementation

- Added the `resetEditorButtonsState()` method after  `setMarkdown()` method which will take previous button states as arguments and retain the formatting button states.

## 🧪 Testing

- Written test cases to check the formatting button states on configuration change.
- Manually tested and verified the functionality in Storybook.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
+++ b/change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add fix to retain the format button state when user configuration gets dynamically updated",
+  "packageName": "@ni/nimble-components",
+  "email": "123377167+aagash-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
+++ b/change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Add fix to retain the format button state when user configuration gets dynamically updated",
+  "comment": "Add fix to retain rich text editor format button state when user configuration gets dynamically updated",
   "packageName": "@ni/nimble-components",
   "email": "123377167+aagash-ni@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/nimble-components/src/rich-text/editor/index.ts
+++ b/packages/nimble-components/src/rich-text/editor/index.ts
@@ -265,19 +265,21 @@ export class RichTextEditor extends RichText implements ErrorPattern {
         prev: EditorConfiguration | undefined,
         next: EditorConfiguration
     ): void {
+        const formatButtonsState = this.getButtonsState(this.tiptapEditor);
+        const { from, to } = this.tiptapEditor.view.state.selection;
         if (this.isMentionExtensionConfigUnchanged(prev, next)) {
-            this.refreshMarkdownContent();
+            this.setMarkdown(this.getMarkdown());
         } else {
             const mentionExtensionConfig = this.getMentionExtensionConfig();
             const currentStateMarkdown = this.getMarkdown();
             this.richTextMarkdownSerializer = new RichTextMarkdownSerializer(
                 mentionExtensionConfig.map(config => config.name)
             );
-            const formatButtonsState = this.getButtonsState(this.tiptapEditor);
             this.initializeEditor();
             this.setMarkdown(currentStateMarkdown);
-            this.resetEditorButtonsState(formatButtonsState);
         }
+        this.tiptapEditor.commands.setTextSelection({ from, to });
+        this.resetEditorButtonsState(formatButtonsState);
         this.setActiveMappingConfigs();
     }
 
@@ -688,7 +690,7 @@ export class RichTextEditor extends RichText implements ErrorPattern {
     }
 
     private resetEditorButtonsState(
-        buttonsState: FormatButtonsState | null
+        buttonsState: FormatButtonsState | undefined
     ): void {
         if (buttonsState?.bold && !this.tiptapEditor.isActive('bold')) {
             this.tiptapEditor.chain().focus().toggleBold().run();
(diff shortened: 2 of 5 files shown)


packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -674,11 +687,34 @@ export class RichTextEditor extends RichText implements ErrorPattern {
         );
     }
 
-    // This method restore the cursor selection after setting the editor content when the editor is focused
+    private resetEditorButtonsState(
```
jattasNI, line 690: Instead of telling tip tap what we think the button state should be based on its previous state, is there a way to ask tip tap to recompute the state? That seems like it could be more reliable to always leave tip tap in control rather than risking that our guess about the previous state and our guess about tip tap's current state are correct.

I don't know what's possible so I'm approving the current approach since you have tests that would fail if those guesses are wrong, but it would be good to see if it can be simplified.

change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add fix to retain the format button state when user configuration gets dynamically updated",
```
jattasNI, line 3: Reminder that these comments appear in a changelog for the entire package so it's good to include the component name to help people understand the change scope.
```suggestion
  "comment": "Add fix to retain the rich text editor format button state when user configuration gets dynamically updated",
```

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -262,8 +273,10 @@ export class RichTextEditor extends RichText implements ErrorPattern {
             this.richTextMarkdownSerializer = new RichTextMarkdownSerializer(
                 mentionExtensionConfig.map(config => config.name)
             );
+            const formatButtonsState = this.getButtonsState(this.tiptapEditor);
             this.initializeEditor();
             this.setMarkdown(currentStateMarkdown);
+            this.resetEditorButtonsState(formatButtonsState);
         }
         this.setActiveMappingConfigs();
```
rajsite, line 281: It looks like every code path through configurationChanged ends up saving and restoring the button state.

Instead of duplicating the logic in two places can we write it that way? Also I don't think having the separate `refreshMarkdownContent` function is making this code clearer. They duplicate so much functionality it does not seem helpful.

```ts
configurationChanged () {
    const formatButtonsState = this.getButtonsState(this.tiptapEditor);
    if (this.isMentionExtensionConfigUnchanged(prev, next)) {
          const { from, to } = this.tiptapEditor.view.state.selection;
          this.setMarkdown(this.getMarkdown());
          this.tiptapEditor.commands.setTextSelection({ from, to });
    } else {
            const mentionExtensionConfig = this.getMentionExtensionConfig();
            const currentStateMarkdown = this.getMarkdown();
            this.richTextMarkdownSerializer = new RichTextMarkdownSerializer(
                mentionExtensionConfig.map(config => config.name)
            );
            this.initializeEditor();
            this.setMarkdown(currentStateMarkdown);
    }
    this.resetEditorButtonsState(formatButtonsState);
    this.setActiveMappingConfigs();
}

```

Also the above makes me wonder, how come in the else case we are not saving the selection state? Should we do it in both cases like the following?

```ts
configurationChanged () {
    const formatButtonsState = this.getButtonsState(this.tiptapEditor);
    const { from, to } = this.tiptapEditor.view.state.selection;
    if (this.isMentionExtensionConfigUnchanged(prev, next)) {
          this.setMarkdown(this.getMarkdown());
    } else {
            const mentionExtensionConfig = this.getMentionExtensionConfig();
            const currentStateMarkdown = this.getMarkdown();
            this.richTextMarkdownSerializer = new RichTextMarkdownSerializer(
                mentionExtensionConfig.map(config => config.name)
            );
            this.initializeEditor();
            this.setMarkdown(currentStateMarkdown);
    }
    this.tiptapEditor.commands.setTextSelection({ from, to });
    this.resetEditorButtonsState(formatButtonsState);
    this.setActiveMappingConfigs();
}

```

If not can we add a comment about why selection state needs to be saved in one and not the other?

packages/nimble-components/src/rich-text/editor/index.ts
rajsite: I don't think so, in one code path we are completely destroying and recreating the editor state. So anything we want to preserver I think we need to hold on to. Even the less destructive code path "refreshMarkdownContent" is saving and restoring the selection state manually..

packages/nimble-components/src/rich-text/editor/index.ts
```diff
@@ -674,11 +687,34 @@ export class RichTextEditor extends RichText implements ErrorPattern {
...
+            this.tiptapEditor.chain().focus().toggleBold().run();
+        }
+        if (buttonsState?.italics && !this.tiptapEditor.isActive('italic')) {
+            this.tiptapEditor.chain().focus().toggleItalic().run();
+        }
+    }
+
+    private getButtonsState(tiptapEditor: Editor): FormatButtonsState | null {
+        if (!this.$fastController.isConnected) {
+            return null;
```
rajsite, line 703: prefer `undefined` over `null`

change/@ni-nimble-components-456f6771-229b-4052-ada2-7f642545afdf.json
aagash-ni: Updated

packages/nimble-components/src/rich-text/editor/index.ts
aagash-ni: Added `resetEditorButtonsState()` in both the states.

packages/nimble-components/src/rich-text/editor/index.ts

packages/nimble-components/src/rich-text/editor/index.ts
aagash-ni: Similar to the selection state, tip tap doesn't know the previous states, So explicitly setting the cursor and button states whenever the `setMarkdown()` gets called.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -90,6 +90,47 @@ describe('RichTextEditorMention', () => {
     });
 
     describe('user mention dynamic loading', () => {
```
jattasNI, line 92: I did a quick search through this file for existing tests that cover the cursor position / text selection code but didn't see any. Could you ensure those changes have coverage too (either by identifying existing tests or writing new ones)?

rajsite (approved):
Approved pending test cases requested by Jesse

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
aagash-ni: Added test cases to check the cursor position.

PR #1730, opened 2024-01-03
User/fvisser/nimble-reboot

fredvisser:
# Pull Request

## ✅ Checklist


change/@ni-nimble-components-97ea2810-9e34-4df8-9c4b-86a1be02703b.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @fredvisser I'm going to close this draft since I don't see any open discussion. I have a lot of questions about the approach but I don't know the priority of this request / use-cases we intend to address so I won't enumerate them yet.
I think the quickest summarization of the concern is that I feel like the approach taken in the PR is too broad to be generally useful and we should focus on specific integrations / use-cases (i.e. completely static content, styles for frameworks like Angular / Blazor, evaluating as a CSS framework like bootstrap or tailwind vs a reset / strict component library).

rajsite: If we revisit this the web awesome native elements stylesheet may be worth being inspired by:
 https://webawesome.com/docs/utilities/native/
https://early.webawesome.com/webawesome@3.0.0-beta.1/dist/styles/native.css

Issue #1659, opened 2023-11-10
Mention users view node is not editable when there is no content next to it in rich text editor


vivinkrishna-ni:
### Context:

The `nimble-rich-text-mention-users-view` is an internal component created to render the `@mention` tag names in rich text editor and viewer components. This view component has two states: one displays in view mode once the user selects the mentioned users, and the other is used as a [decorative tag in tiptap](https://tiptap.dev/api/utilities/suggestion#decoration-tag) to mention a user.

### Problem:

The template of the mention users view looks like:

```HTML
<span>@${mention username}</span><slot></slot>
```
In the above template, the `span` element is used for view mode, and the `slot` is utilized for editing mode, i.e., when a user enters @ into the editor and types text to search for the desired name to mention. Since the element is within a shadow root, the currently typed text that goes into the slot is the expectation. However, in the Chrome browser, if we add the @ at the end of a line, it behaves as follows: the text cursor goes before the @ symbol, and it can't be moved next to @.
![contenteditable-slot-chrome](https://github.com/ni/nimble/assets/123377523/446762c7-e626-4f68-93c0-33e44324080a)

This issue is reproducible only in Chrome and Edge browsers.

For more details about the context and problem, see here: https://github.com/ni/nimble/pull/1653#discussion_r1388077218

### Workaround:

Currently, if we use `<slot contenteditable="true"></slot>` within the mentioned user view template works without the above issue in Chrome and other browsers.

### Action Items:

- [ ] Find a root cause of the problem and if it is related to Chrome browser report/file a bug.
- [ ] If the issue is fixed, remove the `contenteditable="true"` from the `<slot>` element

PR #1731, opened 2024-01-04
Remove `disable-editing` attribute from the mention view element

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Previously, we used `nimble-rich-text-mention-users-view` for two states one is at the time of triggering the mention (Edit mode) and the other one is after committing the mention (View mode) in the editor. However, due to https://github.com/ni/nimble/issues/1716, we wanted to remove the decoration tag (which is configured in Tiptap to render the same element in edit mode) here: https://github.com/ni/nimble/pull/1721. Also, during editing for all mentions we can have the same styling so we decided not to use this element for both edit and view mode. Only after committing in the editor and for viewing in the viewer, this view element can be used. During editing mention node, the styles can be explicitly mentioned in the editor styles.

Relevant discussion: https://github.com/ni/nimble/pull/1721#discussion_r1440837527

Will close the following issues upon merging this PR:
1. fixes https://github.com/ni/nimble/issues/1659
2. fixes https://github.com/ni/nimble/issues/1716

## 👩‍💻 Implementation

- Removed `disable-editing` attribute from the view element base class
- Removed styles related to `disable-editing` and other CSS cleanups in view element
- Added [`decorationClass`](https://tiptap.dev/docs/editor/api/utilities/suggestion#decoration-class) in the editor to style the default suggestion `span` in the editor and added necessary styles to it. 

## 🧪 Testing

- Removed the view element edit mode test case from the Chromatic matrix test
- Manually tested the mention view element in editor and viewer

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -80,6 +80,17 @@ export const styles = css`
         border: ${borderWidth} solid rgba(${borderRgbPartialColor}, 0.1);
     }
 
+    ${
+        /* Shared styles for all mention views at edit time. */ ''
+    }
+    .nimble-mention-view-edit {
+        color: ${bodyFontColor};
+    }
+
+    :host([disabled]) .nimble-mention-view-edit {
+        color: ${bodyDisabledFontColor};
+    }
+
     :host([error-visible]) .container {
         border-bottom-color: ${failColor};
     }
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -80,9 +80,7 @@ export const styles = css`
         border: ${borderWidth} solid rgba(${borderRgbPartialColor}, 0.1);
     }
 
-    ${
-        /* Shared styles for all mention views at edit time. */ ''
-    }
+    ${/* Shared styles for all mention views at edit time. */ ''}
     .nimble-mention-view-edit {
         color: ${bodyFontColor};
     }
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -80,15 +80,6 @@ export const styles = css`
         border: ${borderWidth} solid rgba(${borderRgbPartialColor}, 0.1);
     }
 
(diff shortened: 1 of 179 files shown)


m-akinc (approved):

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -80,6 +80,15 @@ export const styles = css`
...
     }
 
+    ${/* Shared styles for all mention views at edit time. */ ''}
+    .nimble-mention-view-edit {
+        color: ${bodyFontColor};
+    }
+
+    :host([disabled]) .nimble-mention-view-edit {
+        color: ${bodyDisabledFontColor};
+    }
```
rajsite, line 90: These are styles associated with the ProseMirror inserted content so for the styles to be in document order it should be aligneed with those down below:


```suggestion
    .ProseMirror .nimble-mention-view-edit {
        color: ${bodyFontColor};
    }

    :host([disabled]) .ProseMirror .nimble-mention-view-edit {
        color: ${bodyDisabledFontColor};
    }
```

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -80,6 +80,15 @@ export const styles = css`
         border: ${borderWidth} solid rgba(${borderRgbPartialColor}, 0.1);
     }
 
+    ${/* Shared styles for all mention views at edit time. */ ''}
```
rajsite, line 83: There is one selector in the ProseMirror inserted content section below that looks liek:

```
li > p {
        margin-block: 0;
    }
```

And should be updated to:
```
.ProseMirror li > p {
        margin-block: 0;
    }
```

rajsite (commented):
Looks good just some minor style tweaks

packages/nimble-components/src/rich-text/editor/styles.ts
vivinkrishna-ni: Moved the styles below to anchor styles and updated as suggested.

packages/nimble-components/src/rich-text/editor/styles.ts
vivinkrishna-ni: Thanks for pointing it out. I have updated the same.

rajsite (approved):

PR #1733, opened 2024-01-04
Add tests for list-option, menu-item, tab, and tree-view Angular directives

### LINKED ISSUE #318: Add tests for stateless directives
## 📌 User Story
As part of #220 an [example test](https://github.com/ni/nimble/pull/220/files#diff-e7eb866212909a71b1174b7f3c84824a3bd2c12cccb6115ea033fce6500d2782) was added to cover [testing the different template binding mechanisms.](https://github.com/ni/nimble/blob/0d1599ccb1fb11032d2a48b44360a9a37e7140a9/angular-workspace/projects/ni/nimble-angular/CONTRIBUTING.md#testing) 

We should expand that test coverage to all properties for all controls.

One potential way to avoid writing explicit tests for the behavior of each property may be to write reusable decorators that can be well-tested once and give confidence that, for example, boolean, string, or number binding work as expected. There is an [Angular Issue](https://github.com/angular/angular/issues/8968) discussing such a decorator approach that raises some concerns with that approach that should be evaluated to see if they are still issues today.

Edit: This thread also gives some [example decorators](https://github.com/angular/angular/issues/14761). If we end up using any we should make sure to preserve strict type checking.

Edit: [Input setter coercion](https://v12.angular.io/guide/template-typecheck#input-setter-coercion) sounds like something we may need when switching bavk to ivy an using strict templates.

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #318 

This PR adds tests for the list-option, menu-item, tab, and tree-view Angular directives. These are the directives called out in #318 as needing tests. The theme-provider was also mentioned in the issue, but tests have already been added for that directive.

## 👩‍💻 Implementation

The bulk of the changes are adding nimble's standard Angular directive tests for each directive that was missing tests.

The more interesting change is a new list-option directive to contain the properties of the `nimble-list-option` component. It lives alongside the select-list-option and combobox-list-option directives. However, those directives previously each had their own duplicated implementation for the `nimble-list-option` properties (currently only `disabled`). This didn't seem ideal, and it also posed a problem with testing because there wasn't a great way to test each directive in isolation because they had the same selector, the same property, and were exported from the same Angular module. Now, the list-option directive contains everything that would be needed to use the list-option (exports types and contains `disabled` property), and the combobox & select specific directives for the list-option contain the form integration code that is necessary to use the combobox and select within an Angular form. All three directives are exported from the existing list-option module, so this change should be unnoticed by clients.

## 🧪 Testing

Ran auto tests

## ✅ Checklist

### DIFF
--- a/angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
+++ b/angular-workspace/projects/ni/nimble-angular/src/directives/list-option/nimble-combobox-list-option.directive.ts
@@ -1,6 +1,5 @@
-import { Directive, ElementRef, Host, Inject, Input, Optional, Renderer2, AfterViewInit, OnDestroy } from '@angular/core';
+import { Directive, ElementRef, Host, Inject, Input, Optional, AfterViewInit, OnDestroy } from '@angular/core';
 import type { ListOption } from '@ni/nimble-components/dist/esm/list-option';
-import { BooleanValueOrAttribute, toBooleanProperty } from '@ni/nimble-angular/internal-utilities';
 import { NimbleComboboxControlValueAccessorDirective } from '../combobox/nimble-combobox-control-value-accessor.directive';
 
 /**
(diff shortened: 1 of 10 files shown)


mollykreis:
@atmgrifter00, can you buddy this for me?

atmgrifter00 (approved):

rajsite (approved):

PR #1734, opened 2024-01-04
Filterable select hld

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1273


packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
+# Filterable Select
+
+## Problem Statement
+
+Our clients have a need of a filterable dropdown component that does not require arbitrary text as a value, but only values available in the dropdown. The filtering will have the following behaviors:
```
mollykreis, line 5: I find the wording of _that does not require arbitrary text as a value_ a little confusing/misleading. Really, it's not that they don't require arbitrary text, it's that they don't allow arbitrary text, right? What would you think about replacing the word _require_ with _allow_?

```suggestion
Our clients have a need of a filterable dropdown component that does not allow arbitrary text as a value, but only values available in the dropdown. The filtering will have the following behaviors:
```

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
...
+
+-   The filter text will match _any_ text within the `textContent` of each `ListOption`
```
mollykreis, line 7: Could you mention here whether or not it will be case sensitive or case insensitive? Are there any localization considerations we need to consider?

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
...
+-   Each time the dropdown is opened the filter text is cleared
+-   While the dropdown is opened all keystrokes except `<ArrowUp>` and `<ArrowDown>` will apply to the filter.
```
mollykreis, line 9: The current implementation of the select uses `Home` and `End` to jump to the first or last item in the select. Will those now apply to the filter, or should they be called out here as keystrokes that won't apply to the filter?

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
...
+
+#### Select
+
+We will provide a means for clients to enable this feature with a new attribute:
+
+```ts
+export class Select() {
+    ...
+    @attr({ attribute: 'enable-filter', mode: 'boolean' })
+    public enableFilter = false;
```
mollykreis, line 29: Do you think we'd ever have different filter modes? Some things I can think of configuring are case sensitivity and whether the value needs to start with what's typed in.

If we think there are valid future filter modes, we should consider making this an enum called something like `filterMode` that could have a default value of `none`.

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
...
+
+public override set options(value: ListboxOption[]) {
+    this._options = value;
+    Observable.notify(this, 'options');
+}
+```
+
+#### Filter Template
+
+Since the input for the filter is devoid of interaction semantics we find in the `NimbleTextField` (notably no green underline that appears on hover/focus), we will instead use a minimally styled native `input` element.
```
mollykreis, line 66: If you haven't already, can you confirm that there aren't any accessibility issues with having text field with no focus/hover styling and having it keep focus keyboard interactions change the selected value?

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,78 @@
...
+
+One feature that we intend to add to the `Select` is the ability to specify "groups" of options, where each group will have non-selectable header text followed by the options under that group. Ultimately, this feature will have to work nicely with filtering, but I don't believe there are aspects of this that would interfere with the current proposed API in this HLD of a single boolean attribute to enable filtering.
+
+## Alternative Implementations / Designs
+
+None
+
+## Open Issues
+
+-   When there are no matches to the filter I am proposing that we display localizable text below/above the filter text area, the English version of which would be "No items found".
```
mollykreis, line 78: This doesn't seem like an open issue. Should this statement be moved to a different section of the spec?

atmgrifter00: Just added a bit in the `LabelProviderCore` section.

packages/nimble-components/src/select/specs/filerable-select-hld.md
atmgrifter00: I've added an 'Accessibility' section that acknowledges the issue that will be present in the accessibility tree, and proposes that this is acceptable.

packages/nimble-components/src/select/specs/filerable-select-hld.md
atmgrifter00: Called those keys out as not applying to the filter input.

packages/nimble-components/src/select/specs/filerable-select-hld.md
atmgrifter00: I've mentioned that it will be case insensitive. I don't think there are any localization considerations beyond what is called out in the `LabelProviderCore` section, as Nimble won't be responsible for any localization needed for text within each option, and ultimately the filtering will work against whatever the rendered text content is (which I believe would be the expected behavior).

mollykreis (approved):

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,97 @@
...
+-   [#1273: Filterable NimbleSelect](https://github.com/ni/nimble/issues/1273)
+-   [Filterable Select IxD](./IxD.md#filter)
+
+## Implementation / Design
+
+### API
+
+#### Select
+
+We will provide a means for clients to enable this feature with a new attribute:
```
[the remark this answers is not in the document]
atmgrifter00: Correct. Should I explicitly state that as the intent?

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,97 @@
...
+
+#### Filter Template
+
+Since the input for the filter is devoid of interaction semantics we find in the `NimbleTextField` (notably no green underline that appears on hover/focus), we will instead use a minimally styled native `input` element.
+
+#### Accessibility
+
+The accessibility tree will report that the search `input` element should have its own `focus` styling, however this should not be necessary, as the existence of the popup should provide the necessary focus hint to the user. Hiding the `input` from the accessibility tree using `aria-hidden` is strongly discouraged, and setting its `role` to `presentation` is not allowed.
+
+### Future considerations
```
jattasNI, line 83: A couple other things to briefly think about and then probably just list here:
1. any need for debouncing the filter so it doesn't happen on each keystroke? I'm guessing not needed if there aren't too many items.
2. any considerations for list options being more dynamically populated?
   - do we need to fire any new events as filter text is entered? That might enable clients to do their own server-side filtering of items.
   - do we need to define behavior if the set of items changes while the popup is open?

packages/nimble-components/src/select/specs/filerable-select-hld.md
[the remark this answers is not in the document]
jattasNI: Yes please!

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,97 @@
+# Filterable Select
+
+## Problem Statement
+
+Our clients have a need of a filterable dropdown component that does not allow arbitrary text as a value, but only values available in the dropdown. The filtering will have the following behaviors (in the initial feature set):
+
+-   The filter text will match _any_ text within the `textContent` of each `ListOption` in a case insensitive way.
+-   Each time the dropdown is opened the filter text is cleared.
+-   While the dropdown is opened all keystrokes except `<ArrowUp>`. `<ArrowDown>`, `<Home>` and `<End>` will apply to the filter.
```
jattasNI: How does the user select an item using only the keyboard when filtering is enabled? In the [keyboard section of the interaction spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/select/specs/IxD.md#dropdown) it mentions that Enter or Space should choose the currently highlighted item. But it looks like we're currently proposing that the filter will receive those keys, not the items / list.

One idea would be to simply forward those keystrokes too. I think that could make sense for Enter but I could imagine some users being surprised if they typed Space in the filter and it selected an item.

Or am I missing something about the focus states? I'm assuming the filter still has keyboard focus after you press ArrowUp/ArrowDown but maybe that's not accurate.

packages/nimble-components/src/select/specs/filerable-select-hld.md
atmgrifter00: I've created a section in 'Future Considersations' related to dynamically fetched options. I felt there were enough unknowns coupled with the fact that I don't believe it interferes with or alters the current proposed API, that we don't need to design for this use-case at the moment.

Regarding debouncing, I don't believe there's anything that needs to be addressed specifically within this HLD.

packages/nimble-components/src/select/specs/filerable-select-hld.md
```diff
@@ -0,0 +1,97 @@
...
+```ts
+export class Select() {
+    ...
+    @attr({ attribute: 'filter-mode' })
+    public filterMode = FilterMode.none;
+    ...
+}
+
+// Provide enum for filterMode to allow for future modes
+export const FilterMode  = {
```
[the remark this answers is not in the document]
atmgrifter00: I've made several changes to the HLD to address the concerns you raise. The summary is:
- Proposing we use a `filterMode` called `standard` that performs a diacritic and case insensitive filter.
- The `filterMode` is intentionally different than the `Combobox` `autocomplete` API (but I've noted some concerns about some behavior differences).
- Added 'Future considerations' for grouping and metadata

packages/nimble-components/src/select/specs/filerable-select-hld.md
atmgrifter00: I've changed it to mention that `<Enter>` will also not apply to the filter. To be clear, when filtering is enabled, `<Enter>` will commit the value (i.e. raise the `change` event), while `<Space>` will be added as part of the filter text.

You are correct about the focus states.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -0,0 +1,116 @@
...
+export const FilterMode  = {
+    none: 'none';
+    standard: 'standard';
+} as const;
+```
+
+-   The `standard` filterMode will result in case-insensitive, diacritic filtering.
+-   `filterMode` will default to `none` so as not to affect existing clients.
+
+_Note: The `filterMode` isn't meant to mirror the `Combobox` `autocomplete` API, as they do serve slightly different purposes: The `autocomplete` for the `Combobox` ultimately helps set the actual value of the `Combobox` as the user types, and isn't necessarily performing any filtering (e.g. the `inline` mode). One possible concern, however, is that we are presenting an API that will allow different types of filter behaviors (i.e. case sensitive) that the `Combobox` does not support. Additionally, I am proposing diacritic insenitive filtering, which the `Combobox` also does not currently support, but I feel this is quite likely a better default experience._
```
fredvisser, line 43:
```suggestion
_Note: The `filterMode` isn't meant to mirror the `Combobox` `autocomplete` API, as they do serve slightly different purposes: The `autocomplete` for the `Combobox` ultimately helps set the actual value of the `Combobox` as the user types, and isn't necessarily performing any filtering (e.g. the `inline` mode). One possible concern, however, is that we are presenting an API that will allow different types of filter behaviors (i.e. case sensitive) that the `Combobox` does not support. Additionally, I am proposing diacritic insensitive filtering, which the `Combobox` also does not currently support, but I feel this is quite likely a better default experience._
```

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Provide enum for filterMode to allow for future modes
+export const FilterMode  = {
+    none: 'none';
+    standard: 'standard';
+} as const;
+```
+
+-   The `standard` filterMode will result in case-insensitive, diacritic filtering.
```
jattasNI, line 40:
```suggestion
-   The `standard` filterMode will result in case-insensitive, diacritic-insensitive filtering.
```

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -0,0 +1,116 @@
...
+    return this.filteredOptions?.length ? this.filteredOptions : this._options;
+}
+
+public override set options(value: ListboxOption[]) {
+    this._options = value;
+    Observable.notify(this, 'options');
+}
+```
+
+As the `Select` will closely mirror the `Combobox`
```
jattasNI, line 80: sentence fragment

jattasNI (approved):

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -0,0 +1,114 @@
...
+
+// Provide enum for filterMode to allow for future modes
+export const FilterMode  = {
+    none: 'none';
+    standard: 'standard';
+} as const;
+```
+
+-   The `standard` filterMode will result in case-insensitive, diacritic-insensitive filtering.
+-   `filterMode` will default to `none` so as not to affect existing clients.
```
rajsite, line 41: Does a filterMode of none mean that the filterbox is hidden?
Guess I'm wondering if we should make none the default.. I think by default it should be visible. Do we have usage guidance for when it is hidden? If we made it a breaking change to show by default do we know the risks?

jattasNI: Good point, I hadn't thought about usage guidance for when to allow filtering. Once we decide that it could inform our choice of default.

It seems so universally useful that I'm trying to brainstorm reasons that showing it would be a bad idea.
1. It's a bit more visual noise / cognitive load when interacting with a mouse (have to skip over the filter field, differs from the "typical" select they've used elsewhere on the web)
2. It's possibly more cognitive load when interacting with a keyboard (slightly unintuitive that the arrow keys would change selection when the filter has focus)

For long lists of options those theoretical cons are obviously outweighed by the utility of being able to filter. For short lists maybe it's debatable? It would feel kinda goofy to have a list of 3 items with a filter.

So my initial stab at usage guidance would be "show the filter if there are more than ~6 options". If that's the guidance I'm still leaning towards keeping filtering off by default to reduce churn on existing clients.

But those are all loosely held opinions, I'd be interested in @fredvisser and @atmgrifter00 takes too.

fredvisser: I'd agree that `filterMode should default to `none`, as the additional visual complexity is counterproductive for the simple usecases and client apps can easily enable it when appropriate to their data.

- The [USWDS proposes](https://designsystem.digital.gov/components/select/) switching to a filterable option at more than **15 options**
- [NN/g notes](https://www.nngroup.com/articles/listbox-dropdown/) that (non-filterable) select controls should be used when you have between 5-15 options

I'll do a bit more research later, but I can see leveraging that usage guidance for our select as well.

atmgrifter00: I've added a remark that the `none` mode will result in no search input being shown in the dropdown. Whatever usage guidance we come up with will make its way into the Storybook docs with the actual implementation.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -0,0 +1,115 @@
...
+export class Select() {
+    ...
+    @attr({ attribute: 'filter-mode' })
+    public filterMode = FilterMode.none;
+    ...
+}
+
+// Provide enum for filterMode to allow for future modes
+export const FilterMode  = {
+    none: 'none';
```
rajsite, line 35: nit: the value of the default enum case should be `undefined` and not a string. https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#attribute-common-value-patterns

rajsite (approved):

PR #1735, opened 2024-01-05
Use and export parameterizeNamedList in nimble-angular

### LINKED ISSUE #1551: Migrate to the parameterize and parameterizeNamedList test helpers

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1551 

## 👩‍💻 Implementation

There were a few changes made here:
- Move `parameterizeNamedList` in `nimble-components` from `utilities/tests` to `testing` to move it from a private directory to a directory that is intended to be public
- Create a new `testing` entry point in the `nimble-angular` package that re-exports the parameterize utilities
    - Note: #172 indicates that `processUpdates` and `waitForUpdatesAsync` should ultimately be exported from the same `testing` entry point, but that hasn't been done yet since it is a breaking change
- Update Angular tests that used `forEach` to loop through test cases to use `parameterizeNamedList` instead

## 🧪 Testing

Ran the tests and verified the same number of tests are still being run

## ✅ Checklist


mollykreis:
@m-akinc, will you buddy this for me?

m-akinc (approved):

atmgrifter00 (approved):

angular-workspace/projects/ni/nimble-angular/src/directives/anchor-button/tests/nimble-anchor-button-router-link-with-href.directive.spec.ts
```diff
@@ -3,6 +3,7 @@ import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testin
 import { Router } from '@angular/router';
```
rajsite, line 3: Discussed how the parameterize test helper could be useful in skyline as well and is maybe not good to expose via nimble-components or nimble-angular packages (it's a generic jasmine helper not tied to our components).

Directions discussed:
- Create a separate package in nimble, like @ni/jasmine-parameterized (which is similar to this [public package](https://github.com/paucls/jasmine-parameterized/blob/master/package.json) except ours would be in the ni scope)
  - Would need typescript compile and jasmine test set-up
  - Would actually be neat to run it in node and and in browser, should be generic enough to, the typescript build would need to emit dist/esm and dist/cjs
- Copy and paste the utility to the @ni/nimble-angular/internal-utilities entry point so it isn't part of the publically supported api and make the package later in nimble or the js style guide.

rajsite: Moving to draft temporarily until a direction is picked

mollykreis: In #1738, I am creating a new @ni/jasmine-parameterized package with these utilities. Once that PR is merged, I plan to update this PR to reference the new location of `parameterizeNamedList`.

mollykreis:
I'm closing this PR in favor of #1768

PR #1736, opened 2024-01-07
Export tag name as literals

rajsite:
# Pull Request

## 🤨 Rationale

We include type definitions such as the following so that TypeScript is aware of the tag names we register:
```ts
declare global {
    interface HTMLElementTagNameMap {
        'nimble-anchor': Anchor;
    }
}
```

With that kind of definition using an API like `document.createElement('nimble-anchor')` will let TypeScript infer the result will be an instance of `Anchor`. However, because we were exporting tag names as runtime calculated strings, TypeScript was not able to infer the type using `document.createElement(anchorTag);`.

## 👩‍💻 Implementation

This change modifies our tag definitions to export them as constant string literals to improve inferred typing. Notice how in some of the tests `as` casts were removed by the linter for having unnecessary casting.

## 🧪 Testing

Relied on exisiting tests.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated the element creation doc snippets.

### DIFF
--- a/packages/nimble-components/src/table-column/base/tests/table-column.fixtures.ts
+++ b/packages/nimble-components/src/table-column/base/tests/table-column.fixtures.ts
@@ -11,6 +11,11 @@ import type {
 import { ColumnValidator } from '../models/column-validator';
 
 export const tableColumnEmptyCellViewTag = 'nimble-test-table-column-empty-cell-view';
+declare global {
+    interface HTMLElementTagNameMap {
+        [tableColumnEmptyCellViewTag]: TableCellView;
+    }
+}
 /**
  * Simple empty cell view for testing
  */
@@ -20,6 +25,11 @@ export const tableColumnEmptyCellViewTag = 'nimble-test-table-column-empty-cell-
 class EmptyTableCellView extends TableCellView {}
 
 export const tableColumnEmptyGroupHeaderViewTag = 'nimble-test-table-column-empty-group-header-view';
+declare global {
+    interface HTMLElementTagNameMap {
+        [tableColumnEmptyGroupHeaderViewTag]: EmptyTableGroupHeaderView;
+    }
+}
 /**
  * Simple empty group header view for testing
  */
@@ -29,6 +39,11 @@ export const tableColumnEmptyGroupHeaderViewTag = 'nimble-test-table-column-empt
 class EmptyTableGroupHeaderView extends TableGroupHeaderView {}
 
 export const tableColumnEmptyTag = 'nimble-test-table-column-empty';
+declare global {
+    interface HTMLElementTagNameMap {
+        [tableColumnEmptyTag]: TableColumnEmpty;
+    }
+}
 /**
  * Simple empty table column for testing
  */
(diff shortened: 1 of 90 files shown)


mollykreis (approved):

atmgrifter00 (approved):

packages/nimble-components/src/table-column/base/tests/table-column.fixtures.ts
```diff
@@ -11,6 +11,11 @@ import type {
 import { ColumnValidator } from '../models/column-validator';
 
 export const tableColumnEmptyCellViewTag = 'nimble-test-table-column-empty-cell-view';
+declare global {
+    interface HTMLElementTagNameMap {
+        [tableColumnEmptyCellViewTag]: TableCellView;
```
jattasNI, line 16: Could you note why the syntax here is different, with the tag surrounded by `[]`?

jattasNI (approved):
I suspect you'll agree that it's not necessary to wait for approvals from the remaining reviewers. 🚀

packages/nimble-components/src/table-column/base/tests/table-column.fixtures.ts
rajsite: FAST has two patterns for component registration, `compose()` and the `@customElement` decorator. I'd like to use just the decorator but it doesn't support overriding foundation configuration like compose does. So we could just use compose (which is needlessly verbose for the test fixtures) or mix both.

Another difference is that the decorator takes the full tag name while compose takes the prefix (nimble-) and base tag separately.

So to remove the constants in a compose based file I'd like to actually do (maybe also importing the prefix):

```ts
const baseName = 'anchor-button';
export const anchorButtonTag = `nimble-${baseName}`;
declare global {
    interface HTMLElementTagNameMap {
        [anchorButtonTag]: AnchorButton;
    }
}

const nimbleAnchorButton = AnchorButton.compose<AnchorOptions>({
    baseName,
});

DesignSystem.getOrCreate().withPrefix('nimble').register(nimbleAnchorButton());
```

And in modern version of typescript that template expression used for a type string is allowed ([see playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBARgQwgUwHIILbJgXhgcgTGAAsQAnAWjgFcopx8BuAKGQA8AHC2USWIqQoAhOgzAAVBAHNcMAAZgAlhjgAbZJQAkAb0Qp0WAL7zWAE2TA1CctmlqQiNTB0sY7mErBRk5AGYIwNgAEhIAsgAyAKIaWN5S0obIYQicLm4emQDagmTkovTgCQC6AFwwAILEeQXirJlGLI1WSBCV1SJi4C6NQA)) and calculated at compile time. I tried it in [this branch](https://github.com/ni/nimble/compare/main...strict-type-names#diff-276a46796655c4f0352e76dc16ce29522889d60dd46a12ec73073be024be4218R13) but it failed in the current version of typescript.

So ideally in the compose based files we would do the snippet above and in decorator based files (which we already can do today since no expression needs to be evaluated):
```ts
export const anchorTag= 'nimble-anchor';
declare global {
    interface HTMLElementTagNameMap {
        [anchroTag]: Anchor;
    }
}
@customElement({
    name: anchorTag
})
class Anchor {}
```

Soo we can use the decorator pattern today as you noticed and we can switch to the slightly better compose pattern if we cared in the future. This PR shows the quickest change to get the type improvements today but doesn't reduce the string duplication in component definition yet (we could reduce it slightly but doesn't seem worth it, would rather get it to the single sourced state after TS upgrade).

Issue #1589, opened 2023-10-05
Visual design improvements for rich text editor


jattasNI:
## 😯 Problem to Solve

Visual design has provided an approved spec for the rich text components: https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=2482%3A82389&mode=dev
(ensure you're looking at the component labelled "nimble-text-area_WIP"

## 💁 Proposed Solution

The changes needed are:
- [ ] footer should be 40px tall
- [ ] all buttons (both built-in and client-slotted) should be `$ni-nimble-control-slim-height` tall (24px). We should document this guidance for slotted buttons in Storybook.
- [ ] the line between the toolbar and the content should be 2px tall

Ideally fixing this would also make a dent in #610 by adding chromatic testing and storybook documentation for the smaller button height independent of the work in the rich text editor.

PR #1737, opened 2024-01-08
Visual design sizing improvements for rich text editor

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

- Resolves https://github.com/ni/nimble/issues/1589
- Fixing the sizing problems for screens with mobile width.

## 👩‍💻 Implementation

1. Reduce the height of all footer buttons to `$ni-nimble-control-slim-height` (24px) and update the docs to recommend the same height for footer buttons
2. Set the footer height to 40px and the border-top of footer-section to 2px (a line between the toolbar and editor)
3. Adjust the width by removing the gap in the toolbar to be compatible with the mobile screens.
4. Set the default height of the provided `footer-actions` slot buttons to `$ni-nimble-control-slim-height` using `::slotted()` and update the doc to provide only buttons.

## 🧪 Testing

- Updated the storybook matrix tests
- Added a matrix test for different slot buttons to test the default height in the `footer-actions` slot.
- Manually verified in the mobile screen widths

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -256,11 +256,8 @@ export const styles = css`
         gap: 8px;
     }
 
+    ${buttonTag},
     ${toggleButtonTag} {
-        height: ${controlSlimHeight};
-    }
-
-    ${buttonTag} {
         height: ${controlSlimHeight};
     }
 
@@ -270,6 +267,10 @@ export const styles = css`
         margin-inline-end: ${standardPadding};
         gap: ${standardPadding};
         place-items: center;
+    }
+
+    slot[name='footer-actions']::slotted(*) {
+        height: ${controlSlimHeight};
     }
 
     :host([error-visible]) .error-icon {
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -18,6 +18,9 @@ import {
     controlSlimHeight
 } from '../../theme-provider/design-tokens';
 import { styles as errorStyles } from '../../patterns/error/styles';
+import { toolbarTag } from '../../toolbar';
+import { toggleButtonTag } from '../../toggle-button';
+import { buttonTag } from '../../button';
 
 export const styles = css`
     ${display('inline-flex')}
@@ -241,7 +244,7 @@ export const styles = css`
         display: none;
     }
 
-    nimble-toolbar::part(positioning-region) {
+    ${toolbarTag}::part(positioning-region) {
         background: transparent;
         padding-right: 8px;
         box-sizing: border-box;
@@ -249,15 +252,15 @@ export const styles = css`
         height: var(--ni-private-rich-text-editor-footer-section-height);
     }
 
-    nimble-toolbar::part(start) {
+    ${toolbarTag}::part(start) {
         gap: 8px;
     }
 
-    nimble-toggle-button {
+    ${toggleButtonTag} {
         height: ${controlSlimHeight};
     }
 
-    nimble-button {
+    ${buttonTag} {
         height: ${controlSlimHeight};
     }
 
--- a/packages/nimble-components/src/rich-text/editor/styles.ts
+++ b/packages/nimble-components/src/rich-text/editor/styles.ts
@@ -14,7 +14,8 @@ import {
     smallDelay,
     mediumPadding,
     standardPadding,
-    linkFontColor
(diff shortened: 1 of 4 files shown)


vivinkrishna-ni:

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-matrix.stories.ts
```diff
@@ -149,10 +160,21 @@ export const richTextEditorSizing: StoryFn = createStory(html`
...
 const mobileWidthComponent = html`
-    <${richTextEditorTag} style="padding: 20px; width: 300px; height: 250px;">
-        <${buttonTag} slot="footer-actions" appearance="ghost">Cancel</${buttonTag}>
-        <${buttonTag} slot="footer-actions" appearance="outline">Ok</${buttonTag}>
+    <${richTextEditorTag} style="padding: 20px; width: 360px; height: 250px;">
+        <${richTextMentionUsersTag} pattern="^user:(.*)">
+            <${mappingUserTag} key="user:1" display-name="John Doe"></${mappingUserTag}>
+        </${richTextMentionUsersTag}>
+        <${buttonTag}
+            style="height: var(${cssPropertyFromTokenName(tokenNames.controlSlimHeight)}); width: 72px;"
```
m-akinc, line 170: Instead of `cssPropertyFromTokenName(tokenNames.controlSlimHeight)`, use `controlSlimHeight.cssCustomProperty` (in this file and the other stories file).

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
```diff
@@ -69,6 +73,8 @@ const setMarkdownDescription = 'A function that sets content in the editor with
 const getMarkdownDescription = 'A function that serializes the current data in the editor and returns the markdown string.';
 const footerActionButtonDescription = `To place content such as a button at the far-right of the footer section, set \`slot="footer-actions"\`.
 
+It is recommended to set the height of the buttons to \`24px\` or use \`$ni-nimble-control-slim-height\` token.
```
m-akinc, line 76: 1. I wonder if it would be better to enforce this height ourselves, e.g. by using `::slotted(*)` or `::slotted(${buttonTag}), ::slotted(${anchorButtonTag}), ...`. I'm not sure whether we would want to apply the styling for any slotted element, or for a specific set of expected types (like all buttons). @rajsite @jattasNI 
2.  If we want to stick with just documenting the recommendation, then I would just document the token, not "24px".

packages/nimble-components/src/rich-text/editor/styles.ts
```diff
@@ -236,15 +244,26 @@ export const styles = css`
...
         gap: 8px;
     }
 
+    ${toggleButtonTag} {
+        height: ${controlSlimHeight};
+    }
+
+    ${buttonTag} {
+        height: ${controlSlimHeight};
+    }
```
m-akinc, line 265: nit
```suggestion
    ${buttonTag},
    ${toggleButtonTag} {
        height: ${controlSlimHeight};
    }
```

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
jattasNI: I like the idea of providing default styling for the content. I think we do this for other components where we provide guidance about what should be slotted.

I have a mild preference to style just button tags and to update the docs to explicitly say that only buttons are permitted as content. "To place content such as a button..." -> "You can place a button or anchor button...". And the line of docs you added can either be deleted or replaced with "Nimble will set the height of the buttons to ni-nimble-control-slim-height"

jattasNI (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
vivinkrishna-ni: I have updated the styles as per your preference of styling only to button tags instead of styling to all elements. I added a chromatic test for all the nimble buttons to verify the default height of the slotted buttons. Also, updated the storybook docs as suggested.

rajsite (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.stories.ts
rajsite: Looks good!

PR #1739, opened 2024-01-08
Remove workaround for tabindex bug on Edge

### LINKED ISSUE #1118: Remove anchor menu item workaround when fix makes it into Edge
## 🧹 Tech Debt

The anchor menu item has a bug workaround that should eventually be removed.

The FAST Menu manages the `tabindex` of its menu items. Because of a bug in Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1346606), setting the tabindex on an element with `delegatesFocus=true` causes the element to lose focus. This causes the menu to close, preventing arrowing through the items or navigating to the url. As a workaround, we intercept attempts to set the tabindex on the host and instead set it on the inner anchor. The referenced Chromium bug is actually fixed, but it hasn't been pulled into Edge yet (it is in Chrome).

When the fix has been in Edge for 6 months or more, we should remove the workaround from our code.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1118
The bug was fixed in Chromium 112, which corresponds to Edge v112 (released April 2023).

## 👩‍💻 Implementation

Removed workaround (`setAttribute` override).

## 🧪 Testing

Verified in Edge that arrowing through menu with anchor menu item (in Storybook) does not close menu.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-139a7cb5-055b-4ee8-8ac8-7ba105bf516a.json
+++ b/change/@ni-nimble-components-139a7cb5-055b-4ee8-8ac8-7ba105bf516a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Remove workaround for tabindex bug on Edge",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor-menu-item/index.ts
+++ b/packages/nimble-components/src/anchor-menu-item/index.ts
@@ -67,21 +67,6 @@ export class AnchorMenuItem extends AnchorBase {
         return true;
     }
 
-    // The FAST Menu manages the `tabindex` of its menu items. Because of a bug in Chromium
-    // (https://bugs.chromium.org/p/chromium/issues/detail?id=1346606), setting the tabindex on an element
-    // with `delegatesFocus=true` causes the element to lose focus. This causes the menu to close, preventing
-    // arrowing through the items or navigating to the url. As a workaround, we intercept attempts to
-    // set the tabindex on the host and instead set it on the inner anchor.
-    // The referenced Chromium bug is actually fixed, but it hasn't been pulled into Edge yet (it is in
-    // Chrome). Issue https://github.com/ni/nimble/issues/1118 tracks removal of this workaround.
-    public override setAttribute(qualifiedName: string, value: string): void {
-        if (qualifiedName === 'tabindex') {
-            this.anchor.setAttribute(qualifiedName, value);
-        } else {
-            super.setAttribute(qualifiedName, value);
-        }
-    }
-
     // This is part of the bug workaround described above (in setAttribute)
     public override get tabIndex(): number {
         return this.anchor.tabIndex;


packages/nimble-components/src/anchor-menu-item/index.ts
```diff
@@ -67,21 +67,6 @@ export class AnchorMenuItem extends AnchorBase {
...
-    // Chrome). Issue https://github.com/ni/nimble/issues/1118 tracks removal of this workaround.
-    public override setAttribute(qualifiedName: string, value: string): void {
-        if (qualifiedName === 'tabindex') {
-            this.anchor.setAttribute(qualifiedName, value);
-        } else {
-            super.setAttribute(qualifiedName, value);
-        }
-    }
-
     // This is part of the bug workaround described above (in setAttribute)
```
rajsite, line 70: Did you scan this file? There are two comments below saying the source is related to this workaround?

packages/nimble-components/src/anchor-menu-item/index.ts
```diff
@@ -67,21 +67,6 @@ export class AnchorMenuItem extends AnchorBase {
         return true;
     }
 
-    // The FAST Menu manages the `tabindex` of its menu items. Because of a bug in Chromium
```
rajsite, line 70: There is a workaround cited at the bottom of this file linking to a FAST PR that was closed: https://github.com/microsoft/fast/pull/6667

Can we remove that workaround?

packages/nimble-components/src/anchor-menu-item/index.ts
```diff
@@ -67,21 +67,6 @@ export class AnchorMenuItem extends AnchorBase {
         return true;
     }
```
rajsite, line 68: There is a workaround cited at the top of this file linking to an existing fast issue, can we push that one forward? i.e. make a legacy branch pr and master branch PR?
https://github.com/microsoft/fast/issues/5415

packages/nimble-components/src/anchor-menu-item/index.ts
m-akinc: That PR was closed without merging, I think, because I unintentionally deleted the source branch. I'll resurrect the PR and push that forward, but I don't think we should keep this PR lingering around open until that goes in.

packages/nimble-components/src/anchor-menu-item/index.ts
m-akinc: Good catch. I had just searched for the issue number, so I didn't notice the getter/setter.

packages/nimble-components/src/anchor-menu-item/index.ts
m-akinc: It seems like they want to address that by introducing a new anchor menu item element into FAST. Maybe that's a simple as moving our implementation into FAST, but there's some uncertainty there, and I'm not sure we want to prioritize it. We shouldn't make this PR wait on that.

rajsite: > Maybe that's a simple as moving our implementation into FAST, but there's some uncertainty there, and I'm not sure we want to prioritize it. We shouldn't make this PR wait on that.

This comment was resolved but it's not clear how are we capturing following up on that?

> We shouldn't make this PR wait on that.

If a comment is closed with that resolution make sure it's clear how we track following up on that.

m-akinc: Filed tech debt issue: https://github.com/ni/nimble/issues/1740

change/@ni-nimble-components-139a7cb5-055b-4ee8-8ac8-7ba105bf516a.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
m-akinc: Ugh, yes. I reproduced it in Chrome. It seems to only affect the top-level menu. You can arrow past anchor menu items in a submenu.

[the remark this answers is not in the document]
m-akinc: It turns out we still need the workaround logic for another reason. The `AnchorMenuItem` delegates focus to the shadow DOM, so as we try to focus it, we're really trying to focus the `<a>` element. If that element doesn't have an `href` value set, then it isn't focusable.... Unless we have explicitly made it focusable by setting `tabindex="0"` on it. The menu code sets that on each item as you arrow to it. By restoring the workaround code, we pass that `tabindex` value on to the inner anchor.

rajsite (approved):

packages/nimble-components/src/anchor-menu-item/index.ts
```diff
@@ -67,13 +67,10 @@ export class AnchorMenuItem extends AnchorBase {
...
 
-    // The FAST Menu manages the `tabindex` of its menu items. Because of a bug in Chromium
-    // (https://bugs.chromium.org/p/chromium/issues/detail?id=1346606), setting the tabindex on an element
-    // with `delegatesFocus=true` causes the element to lose focus. This causes the menu to close, preventing
-    // arrowing through the items or navigating to the url. As a workaround, we intercept attempts to
-    // set the tabindex on the host and instead set it on the inner anchor.
-    // The referenced Chromium bug is actually fixed, but it hasn't been pulled into Edge yet (it is in
-    // Chrome). Issue https://github.com/ni/nimble/issues/1118 tracks removal of this workaround.
+    // The FAST Menu manages the `tabindex` of its menu items.
```
rajsite, line 70: Actually wait, just to be clear, does this manifest in an issue across all browsers or is it browser specific? I thought the arrow traversal issue was only in Edge?

rajsite (changes requested):
Making request changes to see outcome of discussion

packages/nimble-components/src/anchor-menu-item/index.ts
rajsite: Oh sorry the other thread discussed that it is across browser.

rajsite (approved):

change/@ni-nimble-components-139a7cb5-055b-4ee8-8ac8-7ba105bf516a.json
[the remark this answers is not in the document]
jattasNI: Reminder to update the PR title and description and the changefile comment to reflect the new plan.

jattasNI (approved):

PR #1742, opened 2024-01-09
Move column validator into columnInternals

### LINKED ISSUE #1389: Table columns should create validator as part of columnInternals
## 🧹 Tech Debt

Right now table columns need to ensure that any validator property comes first in the class declaration, so that it is initialized before other properties' `*Changed()` handlers run (because those will try to invoke the validator). This is subtle and fragile. The same kind of problem applied to the `columnInternals`, and we've addressed that by forcing every column type to define `getColumnInternalsOptions()`. We should piggy-back on that existing solution by initializing the validator at the same time.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1389 

## 👩‍💻 Implementation

Previously, any column that had validation associated with it would:
1. Create a validator class that derives from `ColumnValidator`
1. Create a class member within the column that was the validator, making sure that it was defined in the appropriate order within the file so that it got instantiated before it was used in any change handlers
1. Override the base `TableColumn` implementation of the `validity` getter to return the `validity` object from the class member's validator

With these changes, a column that has validation associated with it will:
1. Create a validator class that derives from `ColumnValidator`
1. Create an instance of the validator as part of creating its `ColumnInternalsOptions`, and return it as the `validator` property in the `ColumnInternalsOptions`
1. Use the validator returned from `getTypedValidator` on the base `TableColumn` to perform validation
    1. Note, the validator is owned by the column's `columnInternals` instance, but the column is the one that knows the type of the validator instance. Making `TableColumn` generic on the type of the validator proved problematic because of the way the mixins work. Similar to the fact that a given table column instance isn't strongly typed for `TColumnConfig`, it also isn't strongly typed for the validator type. Therefore, calling the new `getTypedValidator` utility allows columns to be written with a strong assertion that the validator has the expected type.

## 🧪 Testing

- Mostly relying on unit tests still passing for table columns and the table
- Manually spot checked a few table columns to make sure validation still worked as expected

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table-column/base/index.ts
+++ b/packages/nimble-components/src/table-column/base/index.ts
@@ -53,11 +53,11 @@ export abstract class TableColumn<
     public contentSlot!: HTMLSlotElement;
 
     public checkValidity(): boolean {
-        return this.columnInternals.validConfiguration;
+        return this.columnInternals.validator.isValid();
     }
 
     public get validity(): TableColumnValidity {
-        return {};
+        return this.columnInternals.validator.getValidity();
     }
 
     /** @internal */
--- a/packages/nimble-components/src/table-column/base/index.ts
+++ b/packages/nimble-components/src/table-column/base/index.ts
@@ -10,6 +10,10 @@ import {
     ColumnInternalsOptions
 } from './models/column-internals';
 import type { TableColumnValidity } from './types';
+import type { ColumnValidator } from './models/column-validator';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type ColumnValidatorConstructor<T extends ColumnValidator<[]>> = abstract new (...args: any[]) => T;
 
 /**
  * The base class for table columns
@@ -97,5 +101,16 @@ export abstract class TableColumn<
             this.columnInternals.currentSortDirection = this.sortDirection;
             this.columnInternals.currentSortIndex = this.sortIndex;
         }
+    }
+
+    protected getTypedValidator<T extends ColumnValidator<[]>>(requiredType: ColumnValidatorConstructor<T>): T {
+        if (
+            this.columnInternals.validator
+            instanceof requiredType
+        ) {
+            return this.columnInternals.validator;
+        }
+
+        throw new Error('Unexpected column validator type found');
     }
 }
--- a/packages/nimble-components/src/table-column/base/index.ts
+++ b/packages/nimble-components/src/table-column/base/index.ts
@@ -13,7 +13,9 @@ import type { TableColumnValidity } from './types';
 import type { ColumnValidator } from './models/column-validator';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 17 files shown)


packages/nimble-components/src/table-column/base/models/column-validator.ts
```diff
@@ -1,17 +1,17 @@
...
  */
 export class ColumnValidator<
     ValidityFlagNames extends readonly string[]
 > extends Validator<ValidityFlagNames> {
-    public constructor(
-        private readonly columnInternals: ColumnInternals<unknown>,
-        configValidityKeys: ValidityFlagNames
-    ) {
+    @observable
+    public isColumnValid = true;
```
mollykreis, line 12: I needed to add an `@observable` property here to track whether or not the column is valid because the base `Validator` does not have an observable way to know when its validity has changed, which is required by the table. Therefore, I've added `isColumnValid` here, and I mirror that in the existing `validConfiguration` property on the `ColumnInternals` object. I did this because I didn't want the table to have to listen to changes to the column, column internals, and column validator. The table is already handles changes coming from `columnInternals.validConfiguration`.

mollykreis:

packages/nimble-components/src/table-column/base/models/column-validator.ts
```diff
@@ -34,10 +34,10 @@ export class ColumnValidator<
         } else {
             this.untrack(name);
         }
-        this.updateColumnInternalsFlag();
+        this.updateIsColumnValidFlag();
```
atmgrifter00, line 37: Is there still value in having this private method (it seems to me that the value even before your change was questionable)?

atmgrifter00 (approved):

packages/nimble-components/src/table-column/base/models/column-validator.ts
mollykreis: That's a good point. It doesn't seem very valuable, so I'll get rid of it.

jattasNI (approved):

packages/nimble-components/src/table-column/base/index.ts
```diff
@@ -98,4 +104,14 @@ export abstract class TableColumn<
             this.columnInternals.currentSortIndex = this.sortIndex;
         }
     }
+
+    protected getTypedValidator<T extends ColumnValidator<[]>>(
```
rajsite, line 108: As an approach to giving feedback sooner (will try and discuss next meeting), I'll give the high level feedback that I'd like to try and find an alternative to this pattern. It is described in the PR description why the direction was chosen but still feels weird. But review is stalled because haven't had the chance to play with alternatives yet / have ideas to propose yet.

rajsite: Since it's not a high priority tech debt issue and haven't had a chance to look closer, assigning to myself and moving to draft to take another look when there is more bandwidth

mollykreis:
Closing this PR in favor of #1999

PR #1099, opened 2023-03-06
Batch table updates

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1091 

## 👩‍💻 Implementation

Create `UpdateTracker` utility class that helps manage what updates are needed for the table and queuing those updates so they run in batches after an animation frame.

The one update to the table that does not go through the `UpdateTracker` is updates to the table's data. This is because we didn't want the nimble table to hold on to an extra copy of the data, which it would need to do in order to apply it after an animation frame. Additionally, the performance benefit of batching TanStack updates for a data update did not seem as significant as performance benefits from other batching where it is expected to receive multiple changes within a single animation frame. For example, if the sorting changes, then it is expected to receive multiple change notifications about the `sort-index` and `sort-direction` on multiple columns all within a very short period of time.

## 🧪 Testing

- Relied primarily on existing auto tests
- I noticed that we previously were not handling the case that `action-menu-slot` changed after a column was created, so I added a new unit test for that code path

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1744, opened 2024-01-10
Add validator, tracker and batch updates in rich text components for mention configuration

aagash-ni:
# Pull Request

## 🤨 Rationale

1. Add a validator for rich text components to handle the duplicate configuration elements in their child and expose the validation APIs `checkValidity()` and `validity`. Related conversation: https://github.com/ni/nimble/pull/1664#discussion_r1401133453
2. As the `handleChange` for any changes in the configuration or mapping elements is called multiple times which eventually calls the `createConfig()` to get and set the markdown in both the editor and viewer. Since the parser and serializer are expensive operations, we should reduce the number of calls to update the view.

## 👩‍💻 Implementation

- Add a helper class `RichTextValidator` in the rich text model folder to validate the inputs given by the user to configure the mentions. The validation includes:
    - `invalidMentionConfiguration`: which reflects the validation state of the configuration element
    - `duplicateMentionConfiguration`: validates if there is more than one identical configuration element within the children of the rich text editor/viewer. As the editor/viewer should not have multiple data for one mention type, we restrict through a validation.
- Expose `checkValidity()` and `validity` read-only APIs for the user to know what aspect of the mention configuration fails for the editor/viewer.
- Added `RichTextUpdateTracker` to track the properties that are changing dynamically and update the respective rich text components (editor/viewer) when required using `queueUpdate()` method which utilizes the `DOM.queueUpdate` from FAST.

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -1396,6 +1490,7 @@ describe('RichTextEditorMentionListbox', () => {
             expect(pageObject.getMentionListboxItemsName()).toEqual([]);
             mappingElements[0]!.key = 'user:1';
             await waitForUpdatesAsync();
+            await waitForUpdatesAsync();
```
vivinkrishna-ni, line 1493: This test case passes only if we add the additional `waitForUpdateAsync()`. We tried to debug the issue and here are our observations:

1. Replicating the same flow in the storybook by updating the DOM opens the mention popup whenever the invalid `key` is changed to a valid key(works as expected with the current implementation):
invalid-to-valid-state-popup-test-issue

2. When tracking the flow, the `key` of mapping elements changes, we verified that after the first `waitForUpdateAsync()`, the `activeMappingConfigs` updated as expected with the valid mention config data which is responsible for populating the corresponding `nimble-list-option` within mention listbox. However, when we inspect the DOM in the testing environment, there is no `nimble-list-option` within mention listbox. The `activeMappingConfigs` is directly binding to the editor's template but is somehow not reflected in the DOM. Code reference: https://github.com/ni/nimble/blob/8ff070c7058f1602444a010db1dcc324016ff37f/packages/nimble-components/src/rich-text/editor/template.ts#L120C13-L122C36
image

After the second `waitForUpdatesAsync()`, the DOM updates with the expected `nimble-list-option` and opens the mention popup.

vivinkrishna-ni:
Reviewer's info: Since the other PR https://github.com/ni/nimble/pull/1688 with the same changes is not up to date with the latest changes in the rich text components folder, we have created this new PR with the necessary changes. We have addressed all the open comments from that PR and addressed them here. Also, I have closed the above linked PR.

vivinkrishna-ni:

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
m-akinc: I think that's okay. Please just add a comment explaining that after the first wait, `activeMappingConfigs` is updated, and after the second wait, the `nimble-list-options` get populated.

packages/nimble-components/src/rich-text/models/rich-text-tracker.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    public get updateButtonLabel(): boolean {
+        return this.isTracked('buttonLabel');
+    }
+
+    public trackMentionInternalsPropertyChanged(
+        changedMentionInternalsProperty: string
+    ): void {
+        if (
+            isMentionInternalsProperty(
```
m-akinc, line 47: `isMentionInternalsProperty` is only ever called with a single arg, so why not just use a simple `===` test instead? In fact, this can just be a `switch` statement. We can remove `isMentionInternalsProperty`.

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -35,27 +54,46 @@ export abstract class RichText extends FoundationElement {
...
 
-    protected createConfig(): Configuration {
-        return new Configuration(this.mentionElements);
+    protected validate(): void {
+        this.richTextValidator.validateMentionConfigurations(
+            this.mentionElements
+        );
+        this.richTextValidator.validateDuplicateMentionConfigurations(
+            this.mentionElements
+        );
```
m-akinc, line 87:
```suggestion
        this.richTextValidator.validate(this.mentionElements);
```

packages/nimble-components/src/rich-text/models/rich-text-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+    public isValid(): boolean {
+        return Object.values(this.getValidity()).every(x => x === false);
+    }
+
+    public validate(mentions: RichTextMention[]): void {
+        this.validateDuplicateMentionConfigurations(mentions);
+        this.validateMentionConfigurations(mentions);
+    }
+
+    public validateMentionConfigurations(mentions: RichTextMention[]): boolean {
```
m-akinc, line 28: `validateMentionConfigurations` and `validateDuplicateMentionConfigurations` can be private, I think.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -154,30 +160,22 @@ describe('RichTextViewer', () => {
...
-        it('adding two mention configuration elements in the same viewer should render as mention view', async () => {
+        it('adding duplicate mention configuration elements in the same viewer should not render as mention view', async () => {
             element.markdown = '<user:1>';
             await appendUserMentionConfiguration(element);
             await appendUserMentionConfiguration(element);
 
             expect(pageObject.getRenderedMarkdownTagNames()).toEqual([
                 'P',
-                `${richTextMentionUsersViewTag}`.toUpperCase()
+                'NIMBLE-ANCHOR'
```
m-akinc, line 170: Is there a reason not to use `anchorTag`?

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -360,6 +345,106 @@ describe('RichTextViewer', () => {
                 'user:2'
             );
         });
+
+        it('should have valid states for valid configurations', () => {
```
m-akinc, line 349: This test is a duplicate of "should have valid states by default". Remove it.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -461,37 +545,31 @@ describe('RichTextViewer', () => {
...
             element.markdown = '<user:1>';
-            await appendUserMentionConfiguration(
-                element,
-                ['user:1'],
-                ['username1']
-            );
+            await appendUserMentionConfiguration(element, [
+                { key: 'user:1', displayName: 'username1' }
+            ]);
             await waitForUpdatesAsync();
```
m-akinc, line 563: Looks like `appendUserMentionConfiguration` calls `waitForUpdatesAsync` before returning. Review all the call sites and remove any redundant calls to `waitForUpdatesAsync` that immediately follow. (If for some reason a second `waitForUpdatesAsync` call is actually necessary, please add a comment explaining that.)

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
vivinkrishna-ni: Added a comment explaining the rationale of two awaits as suggested.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: No, it should be `anchorTag`. I have updated the same

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: Removed

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
vivinkrishna-ni: Thanks for pointing it out. I have removed `waitForUpdateAsync` after `appendUserMentionConfiguration` calls in every place we use in this file.

packages/nimble-components/src/rich-text/models/rich-text-tracker.ts
vivinkrishna-ni: Updated the changes

atmgrifter00 (approved):

change/@ni-nimble-angular-ae616687-e489-4588-ba8f-583778696c91.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Just a heads-up: for the table we decided to call `checkValidity()` in some places in our SLE `sl-table` unit tests to help catch occasions where the app has configured the Nimble component incorrectly. You could consider doing the same for the rich text components in a follow up PR.

[Example here](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=checkvalidity&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/src/sl-table/sl-table.component.spec.ts)

jattasNI (approved):

change/@ni-nimble-angular-ae616687-e489-4588-ba8f-583778696c91.json
vivinkrishna-ni: Thanks for the heads-up. I have created a TD to track the work of writing pageobjects in `sl-comments`: [Technical Debt 2628236](https://dev.azure.com/ni/DevCentral/_workitems/edit/2628236): FE | Add `checkValidity()` for sl-comments unit tests

rajsite:
@aagash-ni I'm sorry did not have a chance to review the PR yet but I expect to be able to early next week

packages/nimble-components/src/rich-text/base/index.ts
```diff
@@ -36,34 +55,50 @@ export abstract class RichText extends FoundationElement {
...
 
-    private childItemsChanged(_prev: unknown, _next: unknown): void {
-        void this.updateMentionElementsFromChildItems();
+    private childItemsChanged(
+        prev: Element[] | undefined,
+        next: Element[]
+    ): void {
+        if (prev?.length || next.length) {
+            void this.updateMentionElementsFromChildItems();
+        }
```
rajsite, line 92: This seems fishy to me, if the mention elements are cleared then the rest of the system will ignore that change?

If the mention elements are removed we should NOT hold onto the old mention config. That means the element configuration does not map to what the element renders.

rajsite: The following should behave the exact same:

```
<rich-text>
```


and 

```
<rich-text> --> <rich-text><mention></rich-text> --> <rich-text>
```

If we ignore the mention config when it drops to zero then the above to will behave differently. That is wrong

rajsite: Oh wait, maybe I was confused. What is the check guarding against exactly? is it just the transition from `prev = undefined` to `next = empty array`?

If it is just that transition we are ignoreing can you write it as `if (prev !== undefined)`? I think that better conveys we are just ignoring the initialization.

rajsite (commented):
Looks great just looking for one minor clarification

packages/nimble-components/src/rich-text/base/index.ts
rajsite: @aagash-ni @vivinkrishna-ni there was offline requests to have this PR reviewed, so FYI it was reviewed yesterday and is waiting for your follow up

vivinkrishna-ni: Sorry for the delayed response. 

The reason for the check is that some of the async tests fail with `if (prev !== undefined)` and here is the [pipeline failure build](https://github.com/ni/nimble/actions/runs/7472203971/job/20333903895). This is because in some of the transitions, the `updateMentionElementsFromChildItems()` is called which has the async operation of waiting until the child items are defined and resolve promises. So I came up with the below observations of when it is required to update the content in the rich text components:

image

1. In the above image, the red-marked transition is the one that makes the mentioned test fail as we are using `waitUntilCustomElementsDefinedAsync()` operation within the update method. Also, in this transition, we don't want to update the mention elements, as the `childItems` is empty, so `prev?.length || next.length` condition guards the `prev === []` and `next === []`.
2. Also, the yellow marked record is expected to call the `updateMentionElementsFromChildItems()` but this is not an issue for this use case as we are initializing the `childItems` with an empty array during the initialization.

rajsite: Your response describes the behavior being observed, i.e it causes a test failure, but that doesn't tell me if the behavior makes sense and there is a valid reason to skip the cases, if there is a problem with how the test is written, or if skipping the cases is masking an underlying problem. As an FYI that means I need to spend more time understanding the code and why the change makes sense and that will take longer to review. 

> Also, in this transition, we don't want to update the mention elements, as the childItems is transitioned from empty to empty

That doesn't tell me why it makes sense for that to cause some error or test failure. It seems like a strong indication that something else is incorrect because it is surprising that causes an issue

vivinkrishna-ni: In the test cases, after the `await connect()`, the `waitUntilCustomElementsDefinedAsync()` is called. During that time, the element is not initialized and it looks like the below:
image

The breakpoint is here after the connect call for the above case:
image

I am exactly not sure of the reason and as mentioned in [this thread](https://github.com/ni/nimble/pull/1099/files#r1130001024) for the table batch update changes, once adding the `await waitForUpdatesAsync()` after the `await connect()`, the **test passes**. So I have updated the same and set the condition to `prev !== undefined`. So I think this test failure might be related to the test environments because of the async update.

rajsite: That seems aligned with behaviors seen in the table. Thanks for the investigation.
Though like seen with the table and in this PR you may need to expect that some timings in tests will change, I assume that may impact the Skyline tests. I'm assuming the page objects are being re-used in Skyline. You may want to consider if the timing changes can be encapsulated in the page objects.

rajsite (approved):

packages/nimble-components/src/rich-text/base/index.ts
vivinkrishna-ni: I think currently the Skyline tests are not using the page objects and there are necessary awaits already in place in the component tests: [@mention test block in sl-comments ](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemLinkShared/projects/systemlink-comments-lib/src/components/sl-comment/sl-comment.component.spec.ts&version=GC6aa2ab9ea824901f40c045b2abb90f99f702b11d&line=930&lineEnd=931&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). For now, I think we may not want to add timing changes to the page objects for encapsulation, but if needed, we will update it in a separate PR.

I am completing this PR now and if you have any concerns or any other test failures we encounter in Skyline we will create a follow-up PR to address it.

Issue #1747, opened 2024-01-11
Manual Regression Validations


rajsite:
## 🧹 Tech Debt

There are certain tasks that are difficult to automatically handle in source and cause regressions that may not be spotted.
This issue tracks those tasks, gives instructions for auditing them, and captures the last time they were audited.

---

#### Procedure
- Verify that all overrides for connectedCallbacks and disconnectedCallbacks invoke super
#### Why is it not automated?
- Not enforceable via TypeScript yet: https://github.com/microsoft/TypeScript/issues/21388
#### Last checked
1/10/2024

---

#### Procedure
- See the built package contents by visiting: https://unpkg.com/browse/@ni/nimble-components/dist/esm/
  - The above url should automatically redirect to the latest published package
  - Alternative: npm install the package in a new folder locally and check the contents in node_modules
  - Alternative: Check the npm pack step in a recent build in nimble and check the package content list that is printed
- Verify that the dist/esm directory does not contain unexpected files
  - In particular verify that no additional /dist/esm/src directory is present or test/ directories
#### Why is it not automated?:
- See discussion in: https://github.com/ni/nimble/issues/568
#### Last checked
1/10/2024

jattasNI:
Manual tasks complete for this quarter 🎉

jattasNI:
I added a step to manually run `npm audit` as described in #2240.

Issue #1746, opened 2024-01-10
WaferMap test expectations are not executing


rajsite:
## 🧹 Tech Debt

Several wafer map tests are resulting in warns on nimble builds such as the following: https://github.com/ni/nimble/actions/runs/7475603903/job/20344086522#step:17:262

> 'Spec 'Wafermap Prerendering module with undefined values should have empty color fill style' has no expectations.'

The warning states that no `expect` statements are running for the test. However the source of the test shows expect statements:
https://github.com/ni/nimble/blob/e78cd3773cf97604aabb5b4750b349b1c34b39d2/packages/nimble-components/src/wafer-map/tests/prerendering.coloring.spec.ts#L327-L331

Seems like the tests are incorrectly written / not validating behaviors as expected.

The pattern that likely let this issue through is that the tests have conditionally executing expect statements which is an issue discussed here: https://github.com/ni/javascript-styleguide/issues/125#issuecomment-1631721812

To address this tech debt the wafer map tests should be hardened to avoid conditionally executing expect statements as discussed in the above style guide issue.

m-akinc:
@munteannatan Just a heads up.

PR #1748, opened 2024-01-11
Elevate console warn and error to test failure and component lifecycle clean-up

rajsite:
# Pull Request

## 🤨 Rationale

Fixes a couple of small issues noticed recently:

- Tests were printing out console warnings
- Some component callback lifecycles were not calling their super implementation
- When test failures happened the current spec would stop making it difficult to see all the failures and handle them

## 👩‍💻 Implementation

- Elevated console warn and console errors to test failures
- Updated lifecycle callbacks to call super (and renamed some methods that weren't actually lifecycle callbacks, which was my idea, I was wrong 😅). Created a Manual Regression Validation issue to hopefully track that a little better in the future: https://github.com/ni/nimble/issues/1747
- Updated the jasmine configuration to continue running a test suite even on failure. We can see if it is useful and easily switch back if it is annoying. Was useful locally.
- Filed an issue for some Wafer Map tests that are running incorrectly: https://github.com/ni/nimble/issues/1746

## 🧪 Testing

Rely on CI.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
+++ b/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
@@ -1788,7 +1788,7 @@ describe('RichTextEditor', () => {
 
         expect(inputEventListener.spy).toHaveBeenCalledTimes(1);
 
-        await pageObject.setEditorTextContent('');
+        await pageObject.replaceEditorContent('');
         await inputEventListener.promise;
 
         expect(inputEventListener.spy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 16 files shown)


packages/nimble-components/src/rich-text/editor/tests/rich-text-editor.spec.ts
```diff
@@ -1788,7 +1788,7 @@ describe('RichTextEditor', () => {
 
         expect(inputEventListener.spy).toHaveBeenCalledTimes(1);
 
-        await pageObject.setEditorTextContent('');
```
rajsite, line 1791: This was resulting in the warning:

> WARN: '[tiptap warn]: Invalid content.', 'Passed value:', Object{type: 'text', text: ''}, 'Error:', RangeError: Empty text nodes are not allowed
RangeError: Empty text nodes are not allowed

The test description states removved text should trigger the event so used the same remove text pattern as the following test

jattasNI (approved):

rajsite:
overriding owners @atmgrifter00 @DStavilaNI @zszilagy 
please review when you get a chance and can address in follow-up

PR #1751, opened 2024-01-11
Fix styling of disabled tree items

### LINKED ISSUE #1358: Ensure styling of disabled tree items is correct
# 🐛 Bug Report

The styling of a disabled tree item is not consistent with the styling of other disabled nimble components, such as menu items. In most cases, the text/icon color is updated to have a 30% opacity. In the case of the tree item, a 50% opacity is applied. Additionally, disabled tree items use the `not-allowed` cursor, which is inconsistent with our other disabled components.

We need to follow up with visual design to confirm what the styling of a disabled tree item should be because there aren't any existing visual design specs that show disabled items in the tree.

Note: We need to apply the updated styling to both the `nimble-tree-item` and the `nimble-anchor-tree-item`.

## 💻 Repro or Code Sample

Look at the [tree item storybook story](https://nimble.ni.dev/storybook/?path=/story/components-tree-view--tree-item) and make the tree item disabled. Notice that the styling doesn't quite match our other disabled components.

## 🤔 Expected Behavior

We should use consistent styling that is specified by visual designers.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #1358 

## 👩‍💻 Implementation

The tree-item and anchor-tree-item disabled styles now match other components (such as the list-option) and the Figma spec. The changes made include:
- Making the cursor `default` rather than `not-allowed` when hovering over a disabled tree item
- Making the hover color on a disabled tree item transparent
- Making the icon & text associated with a tree item `bodyDisabledFontColor` rather than applying a 50% opacity. (Note that `bodyDisabledFontColor` uses 30% opacity, so this did change the rendered color).

## 🧪 Testing

- Inspected the chromatic diffs

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-84b7a395-1b56-4f91-b4bc-7b277151ba58.json
+++ b/change/@ni-nimble-components-84b7a395-1b56-4f91-b4bc-7b277151ba58.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix disabled styling of tree items",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor-tree-item/styles.ts
+++ b/packages/nimble-components/src/anchor-tree-item/styles.ts
@@ -10,7 +10,9 @@ import {
     fillHoverColor,
     fillHoverSelectedColor,
     fillSelectedColor,
-    bodyFontSize
+    bodyFontSize,
+    bodyDisabledFontColor,
+    iconColor
 } from '../theme-provider/design-tokens';
 import { focusVisible } from '../utilities/style/focus';
 import { userSelectNone } from '../utilities/style/user-select';
@@ -32,20 +34,20 @@ export const styles = css`
         --ni-private-tree-item-nested-width: 0;
     }
 
+    :host([disabled]) {
+        color: ${bodyDisabledFontColor};
+        cursor: default;
+    }
+
     .control {
         display: flex;
         text-decoration: none;
-        color: ${bodyFontColor};
     }
 
     .control${focusVisible} {
         box-shadow: 0px 0px 0px ${borderWidth} ${borderHoverColor} inset;
         outline: ${borderWidth} solid ${borderHoverColor};
         outline-offset: -2px;
-    }
-
-    :host([disabled]) .control {
-        cursor: not-allowed;
     }
 
     .positioning-region {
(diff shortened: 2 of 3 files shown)


mollykreis:

packages/nimble-components/src/anchor-tree-item/styles.ts
```diff
@@ -32,10 +34,15 @@ export const styles = css`
...
+    :host([disabled]) {
+        color: ${bodyDisabledFontColor};
+        cursor: default;
+    }
+
     .control {
         display: flex;
         text-decoration: none;
-        color: ${bodyFontColor};
+        color: currentcolor;
```
atmgrifter00, line 45: Setting `color` to `currentcolor` seems like a misuse of that setting. Shouldn't it just be `inherit`?

packages/nimble-components/src/tree-item/styles.ts
```diff
@@ -139,12 +144,12 @@ export const styles = css`
     }
     .start {
         display: flex;
-        fill: currentcolor;
         margin-inline-start: ${iconSize};
         margin-inline-end: ${iconSize};
     }
 
     slot[name='start']::slotted(*) {
+        ${iconColor.cssCustomProperty}: currentcolor;
```
atmgrifter00, line 152: Sorry, should this be `inherit` as well?

mollykreis: No, `inherit` here wouldn't work since there isn't an `iconColor` to inherit from. We want to set the `iconColor` to the `color` used on the parent.

As a note, we do this in a few different places: menu-item, anchor-menu-item, and text-field

change/@ni-nimble-components-84b7a395-1b56-4f91-b4bc-7b277151ba58.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: It'd be good to summarize the discussion we had at standup about remaining known issues in a comment thread on this PR or the linked issue. (I think the main one was mouse and keyboard respecting the disabled state differently). I think Fred said it wasn't high enough priority to track with a new issue but having a bit of traceability might be helpful if it comes up again.

jattasNI (approved):

change/@ni-nimble-components-84b7a395-1b56-4f91-b4bc-7b277151ba58.json
mollykreis: I've updated the PR description to describe the functionality that I am not addressing in this PR (i.e. the keyboard/mouse interactions of disabled parent tree items not being the same).

rajsite (approved):

PR #1755, opened 2024-01-17
New Rendering HLD Draft

munteannatan:
# Pull Request

## 🤨 Rationale

We want to implement Worker based rendering for the wafer map.

Concerns raised and adressed in the HLD:
1. changing the input interface from list to typed array matrixes then moving the rendering to a worker thread is increasing the performace by ~10x,
2. using multiple workers is possible by splitting the dataset and layering multiple canvasses. The improvement is proportional to the number of workers with some data size limitations.
3. we are planning to evaluate performance with incremental developments, and we have alternative strategies to reach our goals for rendering time and user interaction responsiveness.
4. the limited size of the canvas for very large data sets is going to be signaled with an event or a validity warning for now, while some alternatives are documented

## 👩‍💻 Implementation

n/a

## 🧪 Testing

n/a

## ✅ Checklist


packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,168 @@
...
+
+The main thread will communicate with the worker by signaling any data changes or user events which will trigger full or partial renders.
+
+The strategy for improvement will be gradual. We will try to reuse the existing code as much as possible, but if the performance will not be met, we will improve the existing parsing methods and increase the number of workers.
+
+#### Size Limitations
+
+Because we are reaching out to support very large amounts of data, an issue that is popping out is the limits of the component's canvas size to the amount of displayed data. For example we could find ourselves in the situation to render more dies than there are pixels on the canvas, at least in the zoomed out state.
+
+The solution to this issue is creating a new validity warning or event which will signal this state to the user when the count of individual dies will overtake the size of the canvas. A heuristic which will calculate this based on the zoom level and the component height and width will be implemented and used during data validation.
```
atmgrifter00, line 112: I'm wary of describing what you outline here as "the solution". It suggests that this is how we intend to handle this scenario if not ideally, than indefinitely, which I'm not convinced is the best answer. We discussed some UX and technical possibilities for improving this experience, from decimation of the data (server-side and possibly even client-side in workers), to providing a more instantaneous response to zoom gestures using bitmap scaling. I think these ideas should be captured here, and either have reasons for why we feel they will never be pursued or highlighted as future considerations.

Has the validity warning you describe been designed? I'd be curious to see what was envisioned for this.

change/@ni-nimble-components-063e914a-8704-466c-b58f-6ad350207e57.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @atmgrifter00 @jattasNI talked to @munteannatan offline and was told that they believe all current issues expressed by Meyers should be covered in the HLD update. Said as y'all may have limited availability this week said to go ahead and move to owners review to get more feedback.

change/@ni-nimble-components-063e914a-8704-466c-b58f-6ad350207e57.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
jattasNI, line 2: Change type can be `none` since this doesn't impact the published package in any way.
```suggestion
  "type": "none",
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
+# Rendering++
+
+## Problem Statement
+
+The problem we are addressing is the need for faster rendering in our application. Currently, the rendering process is slow and affects the overall performance of the system. This issue is particularly noticeable when dealing with large or complex datasets. Therefore, we need to design a solution that optimizes the rendering process and improves the user experience.
+
+The proposed design should consider the following factors:
+
+-   Efficiently handle large and complex datasets
```
jattasNI, line 9: What makes a dataset "complex"? Lots of metadata? Non-uniform values? Unsorted values?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
...
+| FTL   | method      | memory space | attempt 1 | attempt 2 | attempt 3 | size      |
+| ----- | ----------- | ------------ | --------- | --------- | --------- | --------- |
+| after | queueUpdate | main         | 5.6ms     | 3.5ms     | 4.0ms     | 100k dies |
+|       | renderDies  | worker       | 52.1ms    | 58.2ms    | 55.2ms    | 100k dies |
+| after | queueUpdate | main         | 20.3ms    | 16.0ms    | 36.0ms    | 1M dies   |
+|       | renderDies  | worker       | 490ms     | 476ms     | 511ms     | 1M dies   |
+| after | queueUpdate | main         | 126ms     | 104ms     | 124ms     | 10M dies  |
+|       | renderDies  | worker       | 3919ms    | 4135ms    | 4071ms    | 10M dies  |
+
+We can see a drastic increase in performance, moreover these benchmarks were
```
jattasNI, line 45: One factor to stay aware of is whether these improvements are being traded off with regressions in other performance metrics. For example, more memory usage or higher first rendering time for small data sets or worse interactive responsiveness because of more worker thread overhead.

I don't think you need to be too rigorous about tracking those, but we want to make sure they don't regress unacceptably. One idea for doing this would be to think about what those other metrics are, take manual snapshots before and after with the prototype, and ensure the after results are still acceptable (even if they're worse).

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
...
+
+[Feature 2391160: Faster Rendering of Large Wafer Maps](https://dev.azure.com/ni/DevCentral/_workitems/edit/2391160)
+
+[Faster Rendering of Large Wafer Maps HLD](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=/docs/design-documents/Ozone/Requirements/Optimize-Wafer-Map-to-Handle-NXP-s-340k-die-per-wafer-data-set/Faster-rendering-of-Large-Wafer-Maps.md&version=GC00f8bb9e698a7310f68fa54395eee63a99cec368&_a=preview)
+
+[Worker Rendering POC](https://github.com/ni/nimble/compare/main...users/munteannatan/scale_and_test_updates)
+
+## Implementation / Design
+
+We will change the whole process of ingesting the data and rendering it in the canvas. The main changes are the data structures used and the process from single threaded to multi threaded. We expect that using better structured data we will gain access time which will improve the rendering and allow distributing the workload if necessary. Moreover, the adoption of thread based processing will free the main thread of the browser to handle user requests and other page events, while allowing us to scale in the future by parallel processing if the need arises.
```
jattasNI, line 59: > we will gain access time

I don't think I follow what you mean by this phrase.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
...
+
+The detailed plan can be found in the azure HLD [Faster Rendering of Large Wafer Maps HLD](https://ni.visualstudio.com/DevCentral/_git/Skyline?path=/docs/design-documents/Ozone/Requirements/Optimize-Wafer-Map-to-Handle-NXP-s-340k-die-per-wafer-data-set/Faster-rendering-of-Large-Wafer-Maps.md&version=GC00f8bb9e698a7310f68fa54395eee63a99cec368&_a=preview).
+
+The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nimble/compare/main...users/munteannatan/scale_and_test_updates).
+
+### Data Structure and Interface
```
jattasNI, line 65: It would be helpful to add some context about why you're proposing using typed arrays. Is this a requirement of worker threads? More memory efficient? Faster to iterate? Easier to serialize/deserialize?

I'm not opposed to them, just want to record the rationale.

(you might already be addressing this in the sentences below, it's just not clear whether typed arrays are the reason for those benefits or something else is)

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
...
+
+1. For wafers >1m and <10m die
+    1. Initial rendering: <10 seconds
+    1. Zoom/pan actions: <1 second
+    1. Interactions that don't modify the dataset (e.g. show labels): <1 second
+    1. Interactions that modify the dataset (e.g. replace layers): <10 seconds
+
+## Alternative Implementations / Designs
+
+### Alternative Data Structures and Interfaces
```
jattasNI, line 170: When listing alternatives it's helpful to include the rationale for not choosing that alternative.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,189 @@
...
+
+Alternatives to the described rendering are splitting the data and canvas and using multiple threads to enhance performance even more.
+
+### Alternative Size Limitations
+
+In order to provide a smooth user experience, if the existing event interaction rendering is not sufficiently fast we would implement a more instantaneous response to zoom gestures using bitmap scaling of the canvas while the renderer is updating the final display.
+
+Another alternative to remove the size limits is to create a point reduction algorithm which will create a reduced set of dies fit to the size limits creating an average, median or other approximation. This approach introduces additional complexity to the rendering and is a lossy compression algorithm, so it's implementation should be carefully considered.
+
+## Open Issues
```
jattasNI, line 186: I think I remember someone suggesting that we should figure out the Blazor and Angular APIs for the new data representation early. Do you know how those frameworks would represent typed arrays? Would there be any performance concerns with transforming data from the framework's representation to the typed array?

jattasNI (commented):
I'll be OOO for about a week but feel free to proceed without me if other reviewers approve.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,207 @@
...
+    // the x coordinates of each column of dies
+    dieColIndexArray: Int32Array;
+    // the lengths of each row of dies
+    rowLengthsArray: Int32Array;
+    // the y coordinates of each die as a matrix row by row
+    dieRowIndexLayer: Int32Array;
+    // the value of each die as a matrix row by row
+    dieValuesLayer: Int32Array;
+    // the highlight state of each die as a matrix row by row
+    dieHighlightsLayer: Int8Array;
```
rajsite, line 96: There are a couple potential approaches that have not been decided yet and are open issues if this data format is used:

1. highlights are treated as an array of booleans as UInt8Array. That would be a bigger change from the public api and require users to pre-calculate tags into highlighted conditions.
2. Treat the tags as a bitset as a Int32Array. We don't care about the tag strings we just keep a 32-bitset of tags associated with a wafer. This more closely aligns to the public api but limits you to 32 tags (or whatever size is chosen for the array)

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,207 @@
...
+    // extra values layers
+    extraLayers : {
+        name: string;
+        type: string;
+        buffer: TypedArray;
+    }[]
```
rajsite, line 102: We talked about instead having this be a JS array where the length of the array corresponds to the wafer coordinates / values. i.e.: `unknown[]`.

The values never have to be sent to the render process and it very closely aligns to the public api where any possible object is valid.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Using TypedArrays has the benefit of direct transfer to web workers without structured cloning of the object by transferring the arrayBuffers and reconstructing the object.
+
+Other benefits of typedArrays include the low access time when iterating over the values, more memory efficiency and faster direct access to metadata layers values.
+
+The previous inputs can be adapted to this new structure to maintain backwards compatibility.
+
+This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
+
```
rajsite, line 113: In this section we should list the alternative of Apache Arrow and the conditions for it:

Pros:

- A row based format that aligns well with the existing public api
- Well supported and tested format
- Nice public API to use, don't have to invent a new format, just doocument our schema for the arrow tables
- Designed for large dataset visualizations

What we need to prototype:

- Does it have comparable memory performance
- Does it perform well or have significant overhead
- What other criteria are y'all interested in in order to evaluate usage?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,207 @@
...
+## Alternative Implementations / Designs
+
+### Alternative Data Structures and Interfaces
+
+The above mentioned structure can also be implemented as an [apache arrow](https://arrow.apache.org/docs/js/index.html) table with columns and metadata. This may introduce an additional overhead that is not needed for the moment as the research towards database caching is still proceeding.
+
+Another option is to break each object property as a separate attribute for the wafer map component. This can also lead to increased complexity and confusion for the user which will need to pass several structured objects instead of a singular object.
+
+### Alternative Rendering
```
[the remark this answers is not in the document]
munteannatan: I think this could be a great ideea. The wafer maps may have some different interaction feature sets, like grids, tooltips, and others, but there is an overlap. It may be something worth looking into in the future. Right now, there are still developments for single image layer wafer galleries and support for large images is being implemented, but this is a direction that could be explored.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -0,0 +1,242 @@
...
+
+In order to provide a smooth user experience, if the existing event interaction rendering is not sufficiently fast we would implement a more instantaneous response to zoom gestures using bitmap scaling of the canvas while the renderer is updating the final display. This approach will introduce the overhead of switching between the two canvas scaling strategies and may increase the load time for smaller datasets. The process can be optimized by pre-caching fixed zoom levels and switching between them.
+
+Another alternative is to create a point reduction algorithm which will create a reduced set of dies fit to the size limits creating an average, median or other approximation. This approach introduces additional complexity to the rendering and is a lossy compression algorithm, so it's implementation should be carefully considered.
+
+We may also implement an external queue canceling functionality.
+
+## Open Issues
+
+User Indication for interactions in progress (>200ms) alternatives:
```
atmgrifter00, line 231: I think it's wise to actually link to the article Milan mentioned as the basis for the decisions we are making, as it demonstrates that we aren't just inventing this ourselves and describes the methodology and importance behind it:
https://web.dev/articles/inp

atmgrifter00, line 231:
```suggestion
User Indication for interactions in progress (>200ms) possibilities:
```
Many of these are not mutually exclusive, which the term "alternative" somewhat suggests.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -91,37 +93,51 @@ class WaferData {
...
+Using TypedArrays has the benefit of direct transfer to web workers without structured cloning of the object by transferring the arrayBuffers and reconstructing the object. Other benefits of typedArrays include the low access time when iterating over the values, more memory efficiency and faster direct access to metadata layers values. The previous inputs can be adapted to this new structure to maintain backwards compatibility.
+
+This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
 
-Other benefits of typedArrays include the low access time when iterating over the values, more memory efficiency and faster direct access to metadata layers values.
+The alternative to the above mentioned data structure is an [apache arrow](https://arrow.apache.org/docs/js/index.html) table with columns and metadata.
 
-The previous inputs can be adapted to this new structure to maintain backwards compatibility.
+Pros of using Apache:
```
rajsite, line 113:
```suggestion
Pros of using Apache Arrow:
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -91,37 +93,51 @@ class WaferData {
...
+-   will be responsible for parsing each layer of the data set and it will render the information contained with the specific color codes and opacity.
+-   will be using the [comlink](https://github.com/GoogleChromeLabs/comlink) library to communicate with the main thread.
+-   will be precompiled by rollup in a library and sent as a Blob to the worker initializer. Workers normally require a separate script file that is downloaded dynamically to run. To avoid the need for dynamically fetched resources and enable static compilation the worker code will be precompiled by rollup in a library and sent as a Blob to the worker.
+-   will function as a state automaton which will need some data after initialization, and it will render everything using an [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas). `SharedArrayBuffers` can be used to efficiently share memory buffers across multiple web workers but require strict server configuration to be enabled. As such they will not be leveraged and are not needed to meet the performance targets.
```
rajsite, line 138: Please include the link that was provided describing the needed server config

```suggestion
-   will function as a state automaton which will need some data after initialization, and it will render everything using an [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas). `SharedArrayBuffers` can be used to efficiently share memory buffers across multiple web workers but require [strict server configuration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements) to be enabled. As such they will not be leveraged and are not needed to meet the performance targets.
```

rajsite (approved):

rajsite:
@atmgrifter00 fyi you had a comment vs approve, so just need to confirm if you needed to take another look
@jattasNI stated we don't have to block on their review
@mollykreis probably not interested in this review

atmgrifter00 (approved):

PR #1756, opened 2024-01-17
Angular support for nimble-unit-byte and nimble-unit-volt

m-akinc:
# Pull Request

## 🤨 Rationale

Angular support for #1573 
(Pipe will be added in a separate PR)

## 👩‍💻 Implementation

- added directives for the two new elements: `nimble-unit-byte` and `nimble-unit-volt`

## 🧪 Testing

Added tests.

## ✅ Checklist


m-akinc:

angular-workspace/projects/ni/nimble-angular/pipes/format-bytes-size.pipe.ts
```diff
@@ -0,0 +1,24 @@
...
+import { DecimalUnitFormat } from '@ni/nimble-components/dist/esm/utilities/unit-format/decimal-unit-format';
+import { byteUnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/byte-unit-scale';
+
+/**
+ * A pipe that transforms a number of bytes into a string formatted like "1.2 kB"
+ */
+@Pipe({
+    name: 'formatBytesSize'
+})
+export class FormatBytesSizePipe implements PipeTransform {
```
mollykreis, line 11: Sorry if I missed/forgot some conversations about this, but what was the rationale for creating this pipe with all the configuration hardcoded within it?

Also, are we going to create pipes for each unit, or create those on an as-needed basis?

angular-workspace/projects/ni/nimble-angular/unit/byte/tests/nimble-unit-byte.directive.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+                imports: [NimbleUnitByteModule]
+            });
+        });
+
+        it('custom element is defined', () => {
+            expect(customElements.get('nimble-unit-byte')).not.toBeUndefined();
+        });
+    });
+
+    describe('with template values', () => {
```
mollykreis, line 19: We normally have a `describe` block called something like _with no values in template_, but I don't see that in this file. Was that intentional?

angular-workspace/projects/ni/nimble-angular/pipes/format-bytes-size.pipe.ts
m-akinc: That's a good point. I had tunnel vision on satisfying the SLE use case, but it would be better to expose a more flexible formatting pipe. I have implementation questions, though. The configuration should be passed at construction, but it seems like clients will want to just get it via DI. Do we just expect that if a client wants a non-default configuration, they have to directly call the constructor?

mollykreis: I'm not sure what the expectation here should be. Ideally, the client gets the pipe through DI and doesn't have to call `new` on it. To use DI and format each value, any configuration would need to be passed into the `format` function. In this case, it seems like there might be performance concerns with that approach.

Two possible options I can think of off the top of my head (not sure if either are what we want to do) are:
1. We have the constructor take the configuration, and we document that most apps should actually extend (or use) this pipe with the configuration they want to use. For example, SLE could keep a `FileSizePipe` that extends/uses the `FormatBytesSizePipe` and choses the configuration.
2. Create injection tokens that the pipe uses so that a client can set the value of the configuration they want. This seems like it might be problematic because then the pipe can't be used with two different configurations within the same component.

m-akinc: I like option 1.

angular-workspace/projects/ni/nimble-angular/unit/byte/tests/nimble-unit-byte.directive.spec.ts
m-akinc: Thanks, that was an oversight. Added.

angular-workspace/projects/ni/nimble-angular/pipes/format-bytes-size.pipe.ts
m-akinc: I actually had to provide an injection token for the configuration options, because I got a build error otherwise. I think I got it so that the pipe can be injected with a default configuration, or the client can define a provider for the injection token to provide specific options to DI, or the client can construct an instance directly. I wanted to make sure that in the third case, any combination of options could be provided/omitted and defaults would work properly.

mollykreis (approved):

rajsite (approved):

PR #1760, opened 2024-01-17
Unify unit scale patterns

rajsite:
# Pull Request

## 🤨 Rationale

Follow-up PR for the `UnitFormat` implementation that:

- Removes the `ManuallyTranslatedUnitScale` base class
- Makes the pattern across unit scales consistent (they each construct an array of `ScaledUnit`s)
- Unifies the split directory structure of `Intl.NumberFormat` vs `ManuallyTranslated` unit scales

This has the additional (currently unleveraged) side effect of making it possible to create a unit scale that leverages both `Intl.NumberFormat` and manually translated scaled units. For example to support exabyte `ScaledUnit`s as they are not supported with `Intl.NumberFormat` used by the rest of the `ScaledUnit`s of the `byteUnitScale`.

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Merged `ManuallyTranslatedUnitScale` tests with the rest. Otherwise relied on exisiting tests.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated README.


rajsite:
@jattasNI can you buddy this so we can hash out the layout before putting it in final review with Mert?

jattasNI (approved):
Looks great! I'm finding it easier to understand.

m-akinc (approved):

PR #1738, opened 2024-01-08
Create jasmine-parameterized package

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1551

Based on [this discussion](https://github.com/ni/nimble/pull/1735#discussion_r1443326772), we decided to create a new npm package for `jasmine-parameterized` rather than have it be exported from an existing nimble package.

## 👩‍💻 Implementation

- Create a new package called `jasmine-parameterized` that contains the existing `parameterized.ts` and `tests\parameterized.spec.ts`
- Add appropriate support files to build, lint, and test the package
- Delete `parameterized.ts` and `tests\parameterized.spec.ts` from `nimble-components`
- Update the tests in `nimble-components` to use `parameterizeSpec` from `jasmine-parameterized`

## 🧪 Testing

- Verified that the jasmine-parameterized tests are run in the pipeline
- Verified that the jasmine-parameterized package contains the `dist` directory (without tests), `package.json`, and `README.md`
- Verified all nimble-components tests still pass

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1764, opened 2024-01-19
Export Angular number formatting pipe

m-akinc:
# Pull Request

## 🤨 Rationale

Provides a way for clients to use the same formatting provided by the number text table column in other parts of their apps.

## 👩‍💻 Implementation

- Created new pipe: `FormatNumberTextPipe` (using a corresponding new `NumberTextUnitFormat` in `nimble-components`) with same configuration options as the number-text table column
- Added documentation for new pipe in Storybook (number text column story)
- Made the new pipe a standalone component and changed the existing `DurationPipe` to be standalone as well. Removed the now empty `NimblePipesModule`. This is a breaking change.

## 🧪 Testing

Tests added

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/pipes/duration.pipe.ts
```diff
@@ -5,7 +5,8 @@ import { DurationFormatter } from '@ni/nimble-components/dist/esm/table-column/d
  * A pipe that transforms a number of milliseconds (string/number) into a string formatted like "1 day, 2 hr, 3 min, 4.567 sec"
  */
 @Pipe({
-    name: 'duration'
+    name: 'duration',
+    standalone: true
```
m-akinc, line 9: This may be controversial since it is an unnecessary breaking change, but it seems preferable to wrapping all our pipes together in one module.

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
```diff
@@ -0,0 +1,65 @@
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
+import { DecimalUnitFormat } from '@ni/nimble-components/dist/esm/utilities/unit-format/decimal-unit-format';
+import { passthroughUnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/passthrough-unit-scale';
+import { byteUnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/byte-unit-scale';
+import { byte1024UnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/byte-1024-unit-scale';
+import { voltUnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/volt-unit-scale';
+import type { UnitScale } from '@ni/nimble-components/dist/esm/utilities/unit-format/unit-scale/unit-scale';
+
+export { byteUnitScale, byte1024UnitScale, voltUnitScale };
```
m-akinc, line 9: I don't like the idea of exporting our unit scales from both the number format pipes, but making a separate module just for them seems weird, too. Open to suggestions.

m-akinc:
@mollykreis could you please buddy this for me?

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
```diff
@@ -0,0 +1,57 @@
...
+     * @internal
+     */
+    public decimalUnitFormat?: DecimalUnitFormat;
+    private minimumFractionDigits?: number;
+    private maximumFractionDigits?: number;
+    private unitScale?: UnitScale;
+
+    public constructor(@Inject(LOCALE_ID) private readonly locale: string) {}
+
+    public transform(value: number, {
```
mollykreis, line 29: Most of Angular's pipes (e.g. number & date) have the options as separate parameters instead of one configuration object. Was it intentional to do something different here? Can you make sure that this works as expected (e.g. make sure that you can define the options only in the template, assuming you aren't setting a unit scale)?

m-akinc: I did intentionally use a different pattern, because I feel it provides a better experience. E.g. if you only want to provide the units, it would be annoying to have to also provide the `minimumFractionDigits` and `maximumFractionDigits` just because of parameter order. This is especially true if I change the API to match the table column, like you suggested, because in that case, a client would have to provide both `digits` and `maximumFractionDigits`, which are mutually exclusive. If there were only two optional parameters, I would probably favor the typical (non-object) way of passing them instead. If the team feels we should not use an object parameter, I can change this.

> Can you make sure that this works as expected (e.g. make sure that you can define the options only in the template, assuming you aren't setting a unit scale)?

I don't follow. Are you asking me to create a test case that uses the pipe in a template, like this?
```HTML
<div>{{ values | formatNumberDecimal:{ maximumFractionDigits: 4 } }}</div>
```
What do you mean by "assuming you aren't setting a unit scale?" The unit scale can (only) be set through the pipe parameters in the template, too.

mollykreis: > Are you asking me to create a test case that uses the pipe in a template, like this?

I'm not necessarily asking for a unit test case (though I'm not against it if you want to write one), but I just wanted to double check that we were certain that doing something like you showed of putting the object literal in the Angular template worked correctly. For example,
- Does the code snippet you showed compile?
- Does it correctly fail to compile if you set the wrong time? (e.g. `values | formatNumberDecimal:{ maximumFractionDigits: "not-a-numeric-value"}`

Part of the reason I asked is because [the only example I could find of something similar in Angular's docs](https://angular.io/api/common/I18nPluralPipe#example) creates the object in TS and then uses it in the template. I wanted to make sure that wasn't a requirement because it seems like that might be a pain for anyone to apply even the most basic configuration.

>What do you mean by "assuming you aren't setting a unit scale?" The unit scale can (only) be set through the pipe parameters in the template, too.

When you use a unit scale, that unit scale object will need to be created in the TS and then used in the template. I was trying to ask whether or not creating the configuration object without a unit scale could be done completely within the template (no code related to the pipe in the TS at all).

m-akinc: >  (though I'm not against it if you want to write one)

Wrote one.

> Does it correctly fail to compile if you set the wrong type?

It doesn't fail to compile when in a test component in my test file, but it does fail to compile when in our example app's component template:
image
I don't see a meaningful difference in the tsconfig files for the example and the tests, but I'm going to assume we're good as long as we're getting the error in the app.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.mdx
```diff
@@ -0,0 +1,37 @@
...
+Angular pipes are provided for formatting number values into strings outside of the Nimble table. The "default" and "decimal" number format pipes correspond to the two `format` modes of the number text table column, using the same formatting rules (and localization).
+
+To use them:
+
+-   Import `FormatNumberDefaultPipe` or `FormatNumberDecimalPipe` from `@ni/nimble-angular/pipes`
+-   Use the pipe in your templates (pipe names are `formatNumberDefault` and `formatNumberDecimal`), or instantiate and use directly in your code
+-   `formatNumberDefault` optionally takes a `UnitScale` object as a parameter, which may be used to format with unit labels
+-   `formatNumberDecimal` optionally takes a configuration object as a parameter. The object may include any of the following properties:
+    - `minimumFractionDigits` - the minimum number of decimal places to show (defaults to 0)
+    - `maximumFractionDigits` - the maximum number of decimal places to show (defaults to 1)
```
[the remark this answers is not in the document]
m-akinc: Good point. I've changed the API of the pipe to match the decimal options of the number-text column. I moved the logic for handling the input parameters into a static function in `DecimalUnitFormat` and call it from both the pipe and the table column.

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
mollykreis: I lean towards creating a separate module for them so that we only import them if they are being used, but I'd be interested in hearing if other code owners have opinions before moving things around too much.

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
```diff
@@ -0,0 +1,63 @@
...
+        maximumDecimalDigits,
+        unitScale = passthroughUnitScale
+    }: {
+        decimalDigits?: number,
+        maximumDecimalDigits?: number,
+        unitScale?: UnitScale
+    } = {
+        unitScale: passthroughUnitScale
+    }): string {
+        if (decimalDigits !== undefined && maximumDecimalDigits !== undefined) {
```
mollykreis, line 40: Could we do something to enforce this requirement at compile-time rather than throwing an exception? For example, how difficult would it be to make the `transform` function look like:

```ts
public transform(value: number, options?: {
        decimalDigits?: undefined,
        maximumDecimalDigits?: number,
        unitScale?: UnitScale
    } | {
        decimalDigits?: number,
        maximumDecimalDigits?: undefined,
        unitScale?: UnitScale
    }): string {
        const configuredUnitScale = options?.unitScale ?? passthroughUnitScale;
        const configuredDecimalDigits = options?.decimalDigits;
        const configuredMaximumDecimalDigits = options?.maximumDecimalDigits;

        /* ... */
}
```

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -53,6 +54,37 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
         this.maximumFractionDigits = maximumFractionDigits;
     }
 
+    public static normalizeAndDefaultFractionDigitOptions(
+        fractionDigits?: number,
+        minimumFractionDigits?: number,
```
mollykreis, line 59: I believe both callers pass `undefined` for `minimumFractionDigits`. Is that correct? Should it be removed from the function signature?

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -53,6 +54,37 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
...
+        maximumFractionDigits?: number
+    ): {
+            minimumFractionDigits?: number,
```
mollykreis, line 62: Based on the implementation, it doesn't look like the returned value for `minimumFractionDigits` would ever be `undefined`.

packages/nimble-components/src/utilities/unit-format/tests/decimal-unit-format.spec.ts
```diff
@@ -223,6 +223,88 @@ describe('DecimalUnitFormat', () => {
...
+                {
+                    name: 'returns min=minimumFractionDigits and max=maximumFractionDigits when both passed',
+                    fractionDigits: undefined,
+                    minimumFractionDigits: 3,
+                    maximumFractionDigits: 4,
+                    expectedMinimumFractionDigits: 3,
+                    expectedMaximumFractionDigits: 4
+                }
+            ] as const;
+            parameterizeNamedList(
```
mollykreis, line 269: Note: With #1738 completing, you'll need to update this to `parameterizeSpec` and update the import to be `import { parameterizeSpec } from '@ni/jasmine-parameterized';`

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -53,6 +54,37 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
         this.maximumFractionDigits = maximumFractionDigits;
     }
 
+    public static normalizeAndDefaultFractionDigitOptions(
```
mollykreis, line 57: Putting `normalizeAndDefaultFractionDigitOptions` in the `DecimalUnitFormat` doesn't seem quite right to me because these defaults are something specific to our choices for the number-text column and our pipe, but not something that is fundamental to the `DecimalUnitFormat` class itself.

I'm not quite sure what a better location is, other than creating a new file to encapsulate this logic. Did you consider any location alternatives, or have a strong rationale of why it makes sense here?

m-akinc: I didn't like the idea of creating a new file for this one function, especially because it is so closely related to our `decimal` number formatting mode. `DecimalUnitFormat` models that mode, so I put the function there.

One possibility is to change `DecimalUnitFormat` so that it takes `decimalDigits` instead of `minimumFractionDigits`, and move the normalization/default logic into the constructor. The only reason I hesitate to do that is that the current API is more flexible and closer to the `Intl.NumberFormat` API. Another option is just to _add_ the `decimalDigits` parameter rather than replacing `minimumFractionDigits`. Either of these options would result in a messier API (i.e. exposing incompatible parameters), but the latter would at least preserve flexibility and compatibility with the `Intl.NumberFormat` API. Do you like that better than what I have currently? Or would you argue for moving the function to a new file?

mollykreis: I agree that keeping the `DecimalUnitFormatter` accepting `minimumFractionDigits` and `maximumFractionDigits` seems the best.

Part of what I'm considering is that I thought one goal was for the code within `src/utilities/unit-format` to represent what could be a third-party library (or split into a different package in the future). So, I guess my questions are:
- Is that a goal?
- If it is a goal, do we think that the defaults we've chosen for the number column belong in that same directory?

m-akinc: That's fair. If we truly want to `src/utilities/unit-format` to be standalone, it doesn't make much sense to add a secondary API with different defaults, just because that's what we want for the pipe and number column. I'll separate it.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
m-akinc: Ultimately removed `minimumFractionDigits` parameter.

angular-workspace/projects/ni/nimble-angular/pipes/tests/format-number-decimal.pipe.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+        {
+            name: 'honors the locale value',
+            options: {
+                locale: 'de',
+                unitScale: byteUnitScale
+            },
+            value: 300,
+            expected: '300,00 Byte'
+        },
+    ];
```
mollykreis, line 39:
```suggestion
    ] as const;
```

angular-workspace/projects/ni/nimble-angular/pipes/tests/format-number-default.pipe.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+        {
+            name: 'honors the locale value',
+            options: {
+                locale: 'de',
+                unitScale: byteUnitScale
+            },
+            value: 300,
+            expected: '300 Byte'
+        },
+    ];
```
mollykreis, line 28:
```suggestion
    ] as const;
```

angular-workspace/projects/ni/nimble-angular/pipes/tests/format-number-default.pipe.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    testCases.forEach(test => {
```
mollykreis, line 30: This should also be updated to `parameterizeSpec`.

change/@ni-nimble-angular-b4cf7416-1794-4759-adc7-bc068045f79d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Provide two number formatting pipes, and change existing DurationPipe to be a standalone component (removing NimblePipesModule)",
```
mollykreis, line 3: nit: 
```suggestion
  "comment": "Provide two number formatting pipes, and change existing DurationPipe to be a standalone pipe (removing NimblePipesModule)",
```

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -85,6 +85,32 @@ export class TableColumnNumberText extends TableColumnTextBase {
         }
     };
 
+    public static convertDigitOptionsForDecimalUnitFormat(
+        decimalDigits?: number,
+        maximumDecimalDigits?: number
+    ): {
+            minimumFractionDigits?: number,
```
mollykreis, line 92: Was there a reason to leave `minimumFractionDigits` as possibly `undefined`, or did that previous review comment get lost when this function moved?

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -727,4 +727,55 @@ describe('TableColumnNumberText', () => {
             expect(cellView.alignment).toEqual(TextCellViewBaseAlignment.left);
         });
     });
+
+    describe('convertDigitOptionsForDecimalUnitFormat', () => {
+        const convertDigitOptionsTestCases = [
```
mollykreis, line 732: I believe there is a code path that returns `undefined` as the `maximumFractionDigits`, right? There should be a test case for that.

mollykreis (approved):

packages/nimble-components/src/table-column/number-text/index.ts
m-akinc: Whoops. Yes, got lost in the churn.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
m-akinc: Actually, there isn't anymore. That used to be possible if only a min digits value was provided, but now that the function doesn't take min digits as a parameter, it always returns a number for both properties. I've updated the type accordingly.

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
```diff
@@ -0,0 +1,64 @@
...
+        maximumDecimalDigits?: undefined,
+        unitScale?: UnitScale
+    } | {
+        decimalDigits?: undefined,
+        maximumDecimalDigits?: number,
+        unitScale?: UnitScale
+    } = {
+        unitScale: passthroughUnitScale
+    }): string {
+        const { minimumFractionDigits, maximumFractionDigits } = TableColumnNumberText.convertDigitOptionsForDecimalUnitFormat(
```
rajsite, line 45: @m-akinc @mollykreis we're now in a weird cross where the api in angular does not represent any api in components. 

We should pick either:
- Expose the UnitFormat api directly as pipes (and the future goal of both the vanilla js and angular pipe moving to separate libraries)
- Expose an API from components that encapsulates the details of the table api.

Seems like work was going in here to align closer to the existing table number column api vs the separate unit format api (by overriding the default value behavior of the unit format apis). If that's the intended direction lets go all the way.

Proposal:

Create: `nimble-components/src/table-column/number-text/models/number-text-unit-format.ts`

```ts
class NumberTextUnitFormat extends UnitFormat {
   constructor(locale: string, options: {
         numberTextFormat: NumberTextFormat,
         decimalDigits?: number, // comment that unused for certain formats
         maximumDecimalDigits?: undefined, // comment that unused for certain formats
         unitScale?: UnitScale;
   }
   ) { /* encapsulate all the table logic of default values */}
}
```

And then we can use that in the table number column (it won't need to reference the separate unitformat pipes directly anymore) and then make a single `NumberTextUnitFormatPipe`. Thoughts?

angular-workspace/projects/ni/nimble-angular/pipes/public-api.ts
```diff
@@ -1,2 +0,0 @@
-export * from './duration.pipe';
```
rajsite, line 1: We can make them standalone modules, that's fine, but we still need to export them from an entrypoint. I assume we still want to export them from `@ni/nimble-angular/pipes` as right now they seem like dead code not exported from anywhere

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
```diff
@@ -0,0 +1,64 @@
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
+import { DecimalUnitFormat } from '@ni/nimble-components/dist/esm/utilities/unit-format/decimal-unit-format';
+import { TableColumnNumberText } from '@ni/nimble-components/dist/esm/table-column/number-text';
```
rajsite, line 3: The alternate proposal will prevent this but FYI we should not depend directly on a component from a pipe. Any dependency on a component as a side-effect registers the component and in this case that means pulling in fast-foundation, etc all for a function that has no actual component dependency. That is a unique behavior of importing from a component and part of the reason we have separate `types` files to avoid accidentally depending on the component registration side-effect.

angular-workspace/projects/ni/nimble-angular/pipes/public-api.ts
rajsite: We also add all components to the example app which exercises the entrypoint but that doesn't make as much sense for the pipes. Maybe the example app should have smoke tests that import the pipes to exercise the entrypoint behavior.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.mdx
```diff
@@ -0,0 +1,40 @@
...
+    of={tableColumnNumberTextStories.numberTextColumn}
+    expanded={false}
+/>
+<Controls of={tableColumnNumberTextStories.numberTextColumn} />
+
+# Usage Docs
+
+## Angular Usage
+
+### Number Format Pipes
```
jattasNI, line 26: I don't think table column pages are the right long term home for documenting our formatting libraries and pipes. Architecturally, the table will be just one of multiple clients of those libraries and the pipes will be just one of multiple ways those libraries are exposed.

I'd propose a structure something like:
1. a new top level section (sibling of Components, Incubating, and Tokens) called something like "Text Formatting" or "Utilities" or ["Guidelines"](https://carbondesignsystem.com/guidelines/content/overview/) (@fredvisser, welcome your input here)
2. that section contains pages about the formatting functions in nimble-components and about the pipe APIs that wraps them. It could also contain other guidance about text like styling, typography, and writing style. Or other general guidelines like Accessibility and Internationalization.
3. the table columns docs are updated to say which formatter is used to format values and to link to the formatting docs. In the future other components that use the library (e.g. numeric text field) would be updated similarly.

Thoughts on that proposed structure?

Thoughts on the urgency of making that change? I think I'd be ok deferring it until we move the formatting code into a separate library but would be fine with it happening sooner if you are excited to make it happen. If it's deferred, could you make sure it's tracked in a tech debt issue?

jattasNI (commented):
I'm fine with the directions proposed by Milan but I'll wait until that settles before doing a more thorough review

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
m-akinc: Makes sense. Any objection to making a factory rather than an actual, new subclass of `UnitFormat`? (I hit a snag trying to pass-through a call to `tryFormat`, which is `protected`.)

rajsite: It's not clear to me what the problem is from your description. You would implement the tryFormat interface and that would involve calls to the instances of DecimalUnitFormat.format and DefauultUnitFormat.format. I would not expect you to need to access their private tryFormat methods like the Angular Pipe usage didn't need to.

change/@ni-nimble-angular-b4cf7416-1794-4759-adc7-bc068045f79d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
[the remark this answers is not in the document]
m-akinc: Yes, I plan to handle uptaking the change.

angular-workspace/projects/ni/nimble-angular/pipes/public-api.ts
m-akinc: Why doesn't it make sense to add the pipes to the example app? Are we unwilling to break the symmetry between the Angular and Blazor example apps?

angular-workspace/projects/ni/nimble-angular/pipes/format-number-text.pipe.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    public constructor(@Inject(LOCALE_ID) private readonly locale: string) {}
+
+    public transform(value: number, {
+        numberTextFormat = NumberTextFormat.default,
+        decimalDigits,
+        decimalMaximumDigits,
+        unitScale = passthroughUnitScale
+    }: {
+        numberTextFormat: 'decimal',
```
m-akinc, line 37: I get a compile error if I try to use `NumberTextFormat.decimal` here instead of the value literal. I think the benefit of having type errors for incompatible options outweighs the downside of using `'decimal'` and `undefined` here instead of `NumberTextFormat.default/decimal`.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.mdx
m-akinc: Since the time you wrote this, the direction has changed to export just a single pipe that wraps a number-text-table-column-specific `NumberTextUnitFormat` object. At this point, I think it still makes sense to document the pipe as a way to match the formatting of the number-text table column in other parts of an Angular UI. I think we can hold off on documenting general formatting utilities until we actually have a separate library, but at that point, I think your proposed structure sounds good. I created issue #1790 to track this.

packages/nimble-components/src/table-column/number-text/models/tests/number-text-unit-format.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    const testCases: readonly {
+        name: string,
+        locale?: string,
+        format: NumberTextFormat,
+        decimalDigits?: number,
+        decimalMaximumDigits?: number,
+        unitScale?: UnitScale,
+        number: number,
+        expected: string
+    }[] = [
```
rajsite, line 17: Please make these test cases more explicit like the following commit: https://github.com/ni/nimble/pull/1764/commits/ec62572289b7312557dbe31f33580bfff339daf3

- just define each property in each case and then you can avoid this typing. It improves test understanding, I don't have to jump back and forth from cases to test code to see what parameters are missing and what the default values are 
- notice in that commit how with the properties defined in each case you don't run into the type behavior of an absent property vs a property of value undefined.

packages/nimble-components/src/table-column/number-text/models/number-text-unit-format.ts
```diff
@@ -0,0 +1,119 @@
...
+        return (
+            this._resolvedOptions.decimalDigits
+                === targetResolvedOptions.decimalDigits
+            && this._resolvedOptions.decimalMaximumDigits
+                === targetResolvedOptions.decimalMaximumDigits
+            && this._resolvedOptions.numberTextFormat
+                === targetResolvedOptions.numberTextFormat
+            && this._resolvedOptions.unitScale === targetResolvedOptions.unitScale
+        );
+    }
```
rajsite, line 50: Can you add some test cases for this optionsMatch behavior. A unique bit is that it is comparing resolved options, so we should exercise some resolved option cases, particularly around defaults.

Some example option sets to compare in optionsMatch
- undefined optionsMatch {}
- {} optionsMatch {numberTextFormat: default, decimalDigits: 7}
- {numberTextFormat: decimal} optionsMatch {numberTextFormat: decimal, decimalDigits: 2}

change/@ni-nimble-components-95469b36-c67b-4902-86ec-dcd63694f710.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
rajsite, line 2: nimble is exposing a new api, should be minor

change/@ni-nimble-components-95469b36-c67b-4902-86ec-dcd63694f710.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Document new Angular number formatting pipe and update documentation of DurationPipe",
```
rajsite, line 3: should mention the new api being exported

angular-workspace/projects/example-client-app/src/tests/pipes.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { DurationPipe, FormatNumberTextPipe } from '@ni/nimble-angular/pipes';
+
+describe('Pipes', () => {
+    it('exports DurationPipe', () => {
+        expect(new DurationPipe('en')).toBeTruthy();
+    });
+
+    it('exports FormatNumberTextPipe', () => {
```
rajsite, line 8: San you add tests for the exported unit scales as well?

angular-workspace/projects/ni/nimble-angular/pipes/format-number-text.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+import { NumberTextUnitFormat, NumberTextUnitFormatOptions } from '@ni/nimble-components/dist/esm/table-column/number-text/models/number-text-unit-format';
+
+/**
+ * A pipe that transforms a number into a string with optional unit
+ */
+@Pipe({
+    name: 'formatNumberText',
+    standalone: true
+})
+export class FormatNumberTextPipe implements PipeTransform {
```
rajsite, line 11: Comparing to other angular pipe names: https://angular.io/guide/pipes
| date
| number
| percent

They don't seem to include an action (i.e. `format`, I guess because they all do the format action since that's what all pipes do), maybe omit the action from the name like:
- `numberText` for the name
- `NumberTextPipe` for the class

angular-workspace/projects/ni/nimble-angular/pipes/format-number-decimal.pipe.ts
rajsite: The comment was resolved but didn't go the direction with the equals statement. Think this would have been an instance that would have been helpful to discuss why that suggestion was not followed and resolved without discussion.

I pushed changes that go in that direction with a `optionsMatch` method. That with a type for the options moved effectively all logic out of the Pipe. Re-opened this thread to use for discussion of that implementation. Definitely lets discuss any typing or approach Qs potentially offline if that's more efficient.

m-akinc: Whoops, I think I initially didn't understand where I would need an `equals` implementation, and then I forgot about it when I switched from my factory to a new `UnitFormat` subclass. You're not saying we still need `equals`, though, right? You had just meant for me to encapsulate the configuration equality check like you ended up doing in `optionsMatch`?

rajsite: yep

m-akinc: Your `optionsMatch` approach and implementation look good to me. Resolving since I don't think there are any further questions to discuss.

jattasNI (commented):
I did a cursory review of docs and high-level architecture and both look good to me. I'll defer to others who have been more involved with this on implementation detail and test coverage. Feel free to submit without my approval.

rajsite (approved):

rajsite:
@mollykreis FYI you were rest on this PR, did you want to take another look?

mollykreis:
> @mollykreis FYI you were rest on this PR, did you want to take another look?

@rajsite, I believe I was reset on accident, but since it seems like the changes have stabilized, I can take another quick look.

mollykreis (approved):

rajsite:
@atmgrifter00 bypassing as an Angular owner, but FYI there are breaking changes if you're interested

PR #1768, opened 2024-01-23
Use parameterizeSpec in nimble-angular

### LINKED ISSUE #1551: Migrate to the parameterize and parameterizeNamedList test helpers

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1551 

The last step of #1551 is to use the new `@ni/jasmine-parameterized` package in the Angular tests.

## 👩‍💻 Implementation

Replace usages of `testCases.forEach` in tests to use `parameterizeSpec` from `@ni/jasmine-parameterize` instead.

## 🧪 Testing

- Verified the same number of tests still run & pass

## ✅ Checklist

### DIFF
--- a/angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
+++ b/angular-workspace/projects/ni/nimble-angular/pipes/tests/duration.pipe.spec.ts
@@ -1,3 +1,4 @@
+import { parameterizeSpec } from '@ni/jasmine-parameterized';
 import { DurationPipe } from '../duration.pipe';
 
 describe('DurationPipe', () => {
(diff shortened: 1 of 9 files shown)


mollykreis:
@m-akinc, can you buddy this for me?

m-akinc (approved):

rajsite (approved):

PR #1774, opened 2024-01-24
Update Data Hierarchy HLD with updated plan of dealing with delayed hierarchy

mollykreis:
# Pull Request

## 🤨 Rationale

We've decided on a different direction for implementing delayed hierarchy in the nimble-table, so I'm updating the HLD to reflect that decision.

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,53 @@ interface TableRowExpansionToggleEventDetail {
...
+
+    // Specifies whether or not the row should be in a state of loading its
+    // children.
+    loading: boolean;
+}
+
+public Table() {
+    ...
+    // Sets the options for the rows specified by the passed IDs.
+    public async setRowOptions(rowOptions: { id: string, options: TableRowOptions }[]): Promise<void>;
```
mollykreis, line 74: I struggled to come up with a name for the type of `{ id: string, options: TableRowOptions }`, so I've just inlined the type in the function signature.

I'm open to names for this type to create an exported interface for it, but I also think we could proceed with the type only specified here.

mollykreis:

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,53 @@ interface TableRowExpansionToggleEventDetail {
...
+```
+
+The loading state of a row will look as follows:
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+Some notes about the `setRowOptions` API:
+
+-   previously set row options will be cleared when `setData` is called
+-   previously set row options will be cleared with the table's `idFieldName` changes
```
atmgrifter00, line 85:
```suggestion
-   previously set row options will be cleared when the table's `idFieldName` changes
```

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,53 @@ interface TableRowExpansionToggleEventDetail {
 
 Note: This event will _not_ be emitted for group rows.
 
+### `Table` API for specifying lazy loaded hierarchical data
+
+In some cases, a client might know that a given row has children, but for performance reasons, that data hasn't been loaded from the backend and given to the table. To enable this use case, there will be an API on the table to specify that a given row should be expandable even if it has no children within the data and to specify that a row is loading its children. That API will look as follows:
+
+```ts
+interface TableRowOptions {
```
jattasNI, line 61: Are the states mutually exclusive? (i.e. is it invalid for a row be both forced expandable and loading?) If so, do we think that will likely remain true for future row options we add?

In the event that both of those are true, we could consider a couple possible alternative APIs:

1. provide separate methods for each state a row can be in:
`setRowsForceExpandable(id: string[])`
`setRowsLoading(id: string[])`
`clearRowsOptions(id: string[])` (don't love this name but we can workshop it if you like the direction)

2. change the state to be an enum:
`setRowOptions({id: string, option: 'expandable' | 'loading' | 'normal'}[])

Both of those would allow us not to expose the `TableRowOptions` type and would mitigate ambiguous cases like setting multiple incompatible states.

But option 1 would require multiple calls if you wanted to set disparate row states and it would break down if we want to have options that aren't mutually exclusive.

I'm pretty neutral on these ideas and am ok with the current proposal, especially if we have visions of growing the set of options, but wanted to suggest the ideas in case others can see a strong reason to change direction.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: Closing as this is now outdated.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,69 @@ interface TableRowExpansionToggleEventDetail {
...
+    canLoadChildren: 'canLoadChildren',
+    loadingChildren: 'loadingChildren'
+} as const;
+export type TableRowDelayedHierarchyState =
+    (typeof TableRowDelayedHierarchyState)[keyof typeof TableRowDelayedHierarchyState];
+
+public Table() {
+    ...
+    // Sets the hierarchy options for the rows specified by the passed IDs.
+    public async setRowHierarchyOptions(rowHierarchyOptions: { id: string, options: TableRowHierarchyOptions }[]): Promise<void>;
```
mollykreis, line 82: I struggled to come up with a name for the type of `{ id: string, options: TableRowHierarchyOptions }`, so I've just inlined the type in the function signature.

I'm open to names for this type to create an exported interface for it, but I also think we could proceed with the type only specified here.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,53 @@ interface TableRowExpansionToggleEventDetail {
 
 Note: This event will _not_ be emitted for group rows.
 
+### `Table` API for specifying lazy loaded hierarchical data
```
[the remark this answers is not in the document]
mollykreis: I've added 2 alternatives in the HLD.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,53 @@ interface TableRowExpansionToggleEventDetail {
...
+```
+
+The loading state of a row will look as follows:
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+Some notes about the `setRowOptions` API:
+
+-   previously set row options will be cleared when `setData` is called
+-   previously set row options will be cleared when the table's `idFieldName` changes
```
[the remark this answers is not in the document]
mollykreis: Good question, and the answer actually depends on what type of information we think will ultimately be in the options object for each row. Therefore, I've updated the direction so that `setRowOptions` is now `setHierarchyRowOptions`, with the intention that those options will only be related to hierarchy. In which case, yes, we should clear those options if the `parentIdFieldName` changes.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: To try to answer your questions (without being able to predict the future):
> Are the states mutually exclusive?
Yes, depending on how we implement things, the states can be mutually exclusive.

> If so, do we think that will likely remain true for future row options we add?
No, I don't think future states will be mutually exclusive. For example, we might have an error message we want to set on a row through the options while also having it be force expandable.

Thinking through the questions you've asked and the alternatives you proposed, @atmgrifter00 and I have come up with a different proposal that is now reflected in HLD. We think that will make things less ambiguous and help make things more straightforward for our clients.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,69 @@ interface TableRowExpansionToggleEventDetail {
...
+-   previously set options will be cleared when the table's `parentIdFieldName` changes
+-   an option associated with a row ID that does not match a record in the data will be ignored
+-   all options will be ignored if `parentIdFieldName` is not configured on the table
+-   calling `setRowHierarchyOptions` when options have already been set on the table will have the following behavior:
+    -   existing options not passed in the new `rowHierarchyOptions` array will be left unmodified
+    -   existing options passed in the new `rowHierarchyOptions` array will be overwritten
+
+The expected usage of the dynamically loaded hierarchy is as follows:
+
+1. Call `setData` on the table with the records that are known
```
jattasNI, line 103: nit: for completeness
```suggestion
1. Configure `idFieldName` and `parentIdFieldName` on the table
1. Call `setData` on the table with the records that are known
```

jattasNI (approved):
Love the new API! Thanks for thinking it through and capturing all the context.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,69 @@ interface TableRowExpansionToggleEventDetail {
...
+Some notes about the `setRowHierarchyOptions` API:
+
+-   previously set options will be cleared when `setData` is called
+-   previously set options will be cleared when the table's `idFieldName` changes
+-   previously set options will be cleared when the table's `parentIdFieldName` changes
+-   an option associated with a row ID that does not match a record in the data will be ignored
+-   all options will be ignored if `parentIdFieldName` is not configured on the table
+-   calling `setRowHierarchyOptions` when options have already been set on the table will have the following behavior:
+    -   existing options not passed in the new `rowHierarchyOptions` array will be left unmodified
+    -   existing options passed in the new `rowHierarchyOptions` array will be overwritten
```
rajsite, line 99: chatted with molly and meyers about aligning behavior closer to the setSelectedRecordIds api which will be updated here (not clearing on setData)

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,69 @@ interface TableRowExpansionToggleEventDetail {
...
+    // Sets the hierarchy options for the rows specified by the passed IDs.
+    public async setRowHierarchyOptions(rowHierarchyOptions: { id: string, options: TableRowHierarchyOptions }[]): Promise<void>;
+}
+```
+
+The loading state of a row will look as follows:
+
+![Lazy Loading Spinner](./spec-images/LazyLoadingSpinner.gif)
+
+Some notes about the `setRowHierarchyOptions` API:
```
rajsite, line 90: To align more with setSelectedRecordIds should we use a name like setRecordHierarchyOptions?
It does emphasize the association to a specific record that we are switching to.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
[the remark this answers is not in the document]
mollykreis: After a bit of discussion with @rajsite and @atmgrifter00, I've changed this to no longer clear the options when `parentIdFieldName` changes. I've made some other behavior changes around `setRowHierarchyOptions` (including renaming it to `setRecordHierarchyOptions`), and one of those changes now makes it easy for a client to clear themselves by passing an empty array of options. Therefore, if they are explicitly changing the hierarchy settings of a table in their application, they can also explicitly clear the hierarchy options if the previously set ones don't make sense for their new configuration.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -59,6 +53,68 @@ interface TableRowExpansionToggleEventDetail {
...
+1. Handle the `row-expansion-toggle` event on the table by doing the following for rows that need to dynamically load their data:
+    1. Start loading the data in the background.
+    1. Call `setRecordHierarchyOptions` on the table to update the state of the row whose data is being loaded to `{ delayedHierarchyState: TableRecordDelayedHierarchyState.loadingChildren }`. This call must include the current state of any records not in the default state.
+1. When the dynamically loaded data has finished loading, do the following based on the result of loading data:
+    - If loading the data succeeds:
+        1. Call `setData` on the table with the new data
+        1. Call `setRecordHierarchyOptions` with the current row options for each row that is either still loading or still has data that hasn't been loaded yet.
+    - If loading the data fails but can be tried again:
+        1. Call `setRecordHierarchyOptions` to change the row state back to `{ delayedHierarchyState: TableRecordDelayedHierarchyState.canLoadChildren }`. This call must include the current state of any records not in the default state.
```
rajsite, line 112: What happens to the expanded state of that row in this scenario? Does calling setData clear the current expanded states since those rows don't have children so why would they be expanded?

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
```diff
@@ -190,11 +199,22 @@ By making the `TableRecord` support hierarchy in its structure, it seemed possib
...
+
+### Create a generic `setRowOptions` API that includes the hierarchy state
+
+The table could have a generic `setRowOptions` function instead of `setRecordHierarchyOptions`. The options set through that function could include delayed hierarchy information but also be extended easily in the future to include additional options. `setRecordHierarchyOptions` was chosen over a more generic `setRowOptions` function for the following reasons:
+
+-   Hierarchy options should be cleared when `parentIdFieldName` changes, but it doesn't make sense to clear non-hierarchical options in that case.
```
rajsite, line 210:
```suggestion
-   Hierarchy options should be ignored when `parentIdFieldName` changes, but it doesn't make sense to clear non-hierarchical options in that case.
```

ignored right? we aren't clearing that cached state

rajsite (approved):

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: Good catch. This bullet point can be removed now since it is no longer a difference between how `setRecordHierarchyOptions` behaves and how we expect a future `setRowOptions` function to behave.

packages/nimble-components/src/table/specs/data-hierarchy-hld.md
mollykreis: I imagined that the code for this would live within the implementation of `setRecordHierarchyOptions`, and if a row is being set to `canLoadChildren` without actually having any children, we'd tell the `ExpansionManager` to ensure the row is collapsed. In this scenario, there aren't any calls being made to `setData` (other than setting the initial table data), so I don't think that logic would play into this behavior.

mollykreis: Although, I think you're right that there is another case to handle here, which is a `setData` call removes all children of an expanded row that as `canLoadChildren` set on it. In that case, we should also make sure the row ends up collapsed again.

In summary, a row with no children in the data and `canLoadChildren` set on it needs to be collapsed, no matter what code path gets into that state.

rajsite (approved):

PR #1775, opened 2024-01-24
Uptake fast-foundation fixes and remove workaround

### LINKED ISSUE #1661: Overriding theme with nested nimble-theme-provider doesn't work in all cases
# 🐛 Bug Report

In some cases, overriding the theme on a page via a nested `nimble-theme-provider` does not work.

Acceptance criteria to resolve this issue:
- [x] Remove workarounds added for that theme provider bug in SystemLinkShared. The workaround that needs to be removed is in the Security UI. It got added [here](https://dev.azure.com/ni/DevCentral/_git/Skyline/commit/d1c7257a67b593403506d14e7ced1f3174398e78?refName=refs/heads/master&path=/Web/Workspaces/Security/src/app/app.component.ts).
- [x] Remove the following workaround from nimble: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/anchor-menu-item/index.ts#L107-L122
- [x] Unpin fast-foundation and fast-react-wrapper versions, see: https://github.com/ni/nimble/pull/1775#discussion_r1488670552

## 💻 Repro or Code Sample

In the Nimble Angular example app main page (`angular-workspace\projects\example-client-app\src\app\app.component.html`), replace the outer `nimble-theme-provider` with:
```html
<nimble-theme-provider [theme]="themeDark">
    <nimble-theme-provider [theme]="themeLight">
<!-- ... --> 
    </nimble-theme-provider>
</nimble-theme-provider>
```
with these property declarations in `app.component.ts`:
```ts
public themeLight: Theme = Theme.light;
public themeDark: Theme = Theme.dark;
```

## 🤔 Expected Behavior

Example app content renders with the Light theme (as it's the innermost theme provider, and `customapp` content is rendered inside of that)

## 😯 Current Behavior

Example app content renders with the Dark theme.

## 💁 Possible Solution

In `nimble-components`, we think removing the property initializer for `theme` on `ThemeProvider` is part of the fix, but may not fix this issue in all cases.

## 🔦 Context

The original context is an SLE team running into issues overriding theme in this PR: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/592543

We've seen different behavior of this bug depending on how the theme is set, and what Angular components set the theme. A few examples:
- In the Nimble Angular example app, have a child/nested theme provider in `customapp.component.html` and override the theme there instead
- Instead of setting theme like `[theme]="themeDark"` in a template, set it directly without a binding: `theme="dark"`

Optimally a fix to this issue will result in the innermost theme taking effect regardless of how the theme is set (`[theme]` vs `theme`), and whether the nested theme provider is in the same Angular component or a child component.

## 🌍 Your Environment

Windows 10 / Chrome browser

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1661

## 👩‍💻 Implementation

- Updated `fast-foundation` to 2.49.5
- Removed bug workaround from anchor menu item

## 🧪 Testing

Verified issues fixed in Storybook.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-59f3a4bc-c922-4175-9583-605bfcb27f1e.json
+++ b/change/@ni-nimble-components-59f3a4bc-c922-4175-9583-605bfcb27f1e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Uptake fast-foundation fixes and remove workaround",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -7143,9 +7143,9 @@
       "integrity": "sha512-gQutuDHPKNxUEcQ4pypZT4Wmrbapus+P9s3bR/SEOLsMbNqNoXigGImITygI5zhb+aA5rzflM6O8YWkmRbGkPA=="
     },
     "node_modules/@microsoft/fast-foundation": {
-      "version": "2.49.4",
-      "resolved": "https://registry.npmjs.org/@microsoft/fast-foundation/-/fast-foundation-2.49.4.tgz",
-      "integrity": "sha512-5I2tSPo6bnOfVAIX7XzX+LhilahwvD7h+yzl3jW0t5IYmMX9Lci9VUVyx5f8hHdb1O9a8Y9Atb7Asw7yFO/u+w==",
+      "version": "2.49.5",
+      "resolved": "https://registry.npmjs.org/@microsoft/fast-foundation/-/fast-foundation-2.49.5.tgz",
+      "integrity": "sha512-3PpG1BNmZ5kUM1goYU3SsxjsM2H7Rk0ZmpDJ7mnRhWDgKiM5SzJ02KvALRUqDrJQoeDnkW0Q2Q+r9SkEd68Gpg==",
       "dependencies": {
         "@microsoft/fast-element": "^1.12.0",
         "@microsoft/fast-web-utilities": "^5.4.1",
@@ -36415,7 +36415,7 @@
       "dependencies": {
         "@microsoft/fast-colors": "^5.3.1",
         "@microsoft/fast-element": "^1.12.0",
-        "@microsoft/fast-foundation": "^2.49.3",
+        "@microsoft/fast-foundation": "^2.49.5",
         "@microsoft/fast-web-utilities": "^6.0.0",
         "@ni/nimble-tokens": "^6.9.1",
         "@tanstack/table-core": "^8.10.7",
--- a/packages/nimble-components/package.json
+++ b/packages/nimble-components/package.json
@@ -59,7 +59,7 @@
   "dependencies": {
     "@microsoft/fast-colors": "^5.3.1",
     "@microsoft/fast-element": "^1.12.0",
-    "@microsoft/fast-foundation": "^2.49.3",
+    "@microsoft/fast-foundation": "^2.49.5",
     "@microsoft/fast-web-utilities": "^6.0.0",
     "@ni/nimble-tokens": "^6.9.1",
     "@tanstack/table-core": "^8.10.7",
--- a/packages/nimble-components/src/anchor-menu-item/index.ts
+++ b/packages/nimble-components/src/anchor-menu-item/index.ts
@@ -2,7 +2,6 @@ import { attr, observable } from '@microsoft/fast-element';
 import {
     DesignSystem,
     AnchorOptions,
-    MenuItem as FoundationMenuItem,
     MenuItemColumnCount
 } from '@microsoft/fast-foundation';
 import { keyEnter } from '@microsoft/fast-web-utilities';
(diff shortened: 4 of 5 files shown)


rajsite:
The PR is still failing builds and looks like it was prematurely set to owners review, moving back to draft

change/@ni-nimble-components-59f3a4bc-c922-4175-9583-605bfcb27f1e.json
m-akinc, line 1: I discovered that this new release of `fast-foundation` introduces an issue on Chrome/Edge. If a component's styling uses a `themeBehavior`, switching to the theme for that `themeBehavior` crashes the window (i.e. by setting the `theme` value of the containing theme provider). There seem to be exceptions, such as the menu and the select (both of which's `themeBehavior`s style only `slot` elements).

We can't update to this release of `fast-foundation` with this issue. If possible, we should try to fix this in FAST, but reporting a Chromium bug may also be an option.

rajsite: There is only one other change in fast foundation that was not contributed by us: https://github.com/microsoft/fast/commits/archives/fast-element-1/

Are you saying we introduced the regression? Did we test the changes at all?

rajsite: There was one other release before the one with our changes: https://github.com/microsoft/fast/commit/71481316c92e3689a4a0c243831693d9938474f8

Do you see the same failures on just that change?

rajsite: > We can't update to this release of fast-foundation with this issue

Then we may need to quickly release a version of nimble that pins our FAST version. It was a patch release so any app that does a general dependency update will start using the new version of FAST. 

If we introduced the regression and it is observable in FAST we probably want to notify them and get them to deprecate the package and warn their users.

rajsite: Searching Skyline [most apps are on 2.49.4](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=%40microsoft/fast-foundation%22%3A%20%7B&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/TagManager/package-lock.json) and some are older. Seems unlikely that 2.49.4 introduced the regression.

I'm going to force through a PR that locks us on 2.49.4: https://github.com/ni/nimble/pull/1779

rajsite: Gave the fast folks a heads-up on discord: https://discord.com/channels/449251231706251264/1197942296931483669/1199912333787541575

Reproduced as follows:
> rajsite — Today at 9:03 PM
@Chris Holt heads-up that during integration of fast-foundation@2.49.5 we found a regression that causes a crash in chrome (121.0.6167.86) and edge (120.0.2210.144) but not firefox (122.0) or safari (ipados 16.7.5).
We haven't narrowed down the cause to see if its a chrome/edge bug, a problem in the implementation we contributed to FAST, or an issue manifesting in our token usage (which is quite complex as we leverage tokens for many things). While investigating we are pinning users on fast-foundation@2.49.4.
You can see an example of the crash in chrome or edge in this example with 2.49.5 by going to the top right menu > user settings > and changing the select to dark mode: https://60e89457a987cf003efc0a5b-gnaiujxbdz.chromatic.com/example-client-app/#/customapp
And that the crash is not reproducible in any of the above browser with the 2.49.4 build: https://60e89457a987cf003efc0a5b-affprjyukn.chromatic.com/example-client-app/#/customapp
Chris Holt — Today at 9:03 PM
This is the latest version, yes?
rajsite — Today at 9:04 PM
Yep
Chris Holt — Today at 9:05 PM
Ok, I likely can’t test tonight, but my first approach is probably to check against other versions to isolate whether it’s related to the change or related to the browser. Then I’d probably go from there. If folks on your end can isolate, we can go from there - or I’ll look as soon as I’m able
rajsite — Today at 9:07 PM
Sounds good, we likely won't be able to investigate further tonight but will keep looking tomorrow

rajsite: Pinned the fast-foundation version so unlikely users that update dependencies will pick-up the version with thee regression and notified the fast folks on discord. 

The next steps are to investigate the issue further and see if we need a patch / revert the change in fast and/or report an issue to the browser vendors

m-akinc: We're talking to Chris Holt in a [Discord thread](https://discord.com/channels/449251231706251264/1197942296931483669/1200233903294791690) about possible workarounds we could submit to FAST. I also [filed a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1522263) to Chromium.

rajsite: I also had to pin `@microsoft/fast-react-wrapper` as the latest version tried to bring in 2.49.5 as well in #1771 . This all may be less painful if fast switches to peerDependencies, I'm not actually sure (and it's a change in main they are proposing so who know when it will release): https://github.com/microsoft/fast/issues/6908

Eitherway as part of resolving the upgrade we should unpin fast-foundation and fast-react-wrapper

Issue #1761, opened 2024-01-18
Possibly flaky test: "should commit mention into the editor on Enter"


m-akinc:
## 🧹 Tech Debt
Saw this test fail on an unrelated PR.

![image](https://github.com/ni/nimble/assets/7282195/f7e1ae7b-20a6-4cc9-8777-ed0c6de5bb0d)

https://github.com/ni/nimble/actions/runs/7576426399/job/20635217943

m-akinc:
Created AzDO work item assigned to LIMS: https://ni.visualstudio.com/DevCentral/_workitems/edit/2632606

rajsite:
Intermittent test disabled in https://github.com/ni/nimble/pull/1770
As per nimble intermittent test policy: https://github.com/ni/nimble/blob/adde25d1cad4b59868590e81684a622bbd38b23e/CONTRIBUTING.md#handling-intermittent-test-failures

AzDo issue filed to address disabled test: https://ni.visualstudio.com/DevCentral/_workitems/edit/2632606

PR #1780, opened 2024-01-25
Fix for intermittent test failure in rich test editor mention support

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Fixes: #1761 

## 👩‍💻 Implementation

- Re-enabled the test as it is not reproducible. See the comment below: https://github.com/ni/nimble/pull/1780#discussion_r1472879913
- As the test is not reproducible upon testing, added a comment to the test to report the issue in any case if it fails to understand the frequency better.

## 🧪 Testing

- Manually triggered the [branch build](https://github.com/ni/nimble/actions/runs/7639721881/job/20846662309) and [PR build](https://github.com/ni/nimble/actions/runs/7651590015/job/21060922018) to find the frequency of its failure.
- Tested locally by running the test in a loop a thousand times.
- The error indicates the selection has no range because of the code in [Tiptap mention extension](https://github.com/ueberdosis/tiptap/blob/1c5c087641162dc9d82814aaa84fcdc267469545/packages/extension-mention/src/mention.ts#L61). Verified this by adding logs before pressing Enter in the test case to find out that the editor has content and throw this error on a rare occasion.

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -1205,14 +1205,22 @@ describe('RichTextEditorMentionListbox', () => {
             expect(pageObject.isMentionListboxOpened()).toBeFalse();
         });
 
-        // Disabled due to intermittency. See: https://ni.visualstudio.com/DevCentral/_workitems/edit/2632606
-        xit('should commit mention into the editor on Enter', async () => {
+        // Re-enabled the test to monitor the intermittency failure
+        it('should commit mention into the editor on Enter', async () => {
```
vivinkrishna-ni, line 1209: @m-akinc, Regarding the intermittent test failure issue and as per the test policy, I have performed the following:
1. Created a branch, enabled the test, and monitored for a few days but couldn't reproduce the test failure. https://github.com/ni/nimble/actions/runs/7639721881/job/20846662309
2. To run in the main / pull request, I created this PR and tried to reproduce but still it didn't fail a single time. https://github.com/ni/nimble/actions/runs/7651590015/job/21060922018

When trying to debug from the actual error message mentioned in [the issue](https://github.com/ni/nimble/issues/1761), it seems that the TipTap internally calls the `window.getSelection().collapseToEnd()` ([Tiptap code link](https://github.com/ueberdosis/tiptap/blob/1c5c087641162dc9d82814aaa84fcdc267469545/packages/extension-mention/src/mention.ts#L61)) method when committing a mention node by pressing Enter key after the dropdown opens. So I tried adding console logs and thought of monitoring it for another couple of days as per the test policy. However, this option would want us to manually trigger the build each time so we might miss capturing the failure.

Therefore, I would like to get your thoughts on how could we proceed further. Can we try adding a `try..catch` (to not block other builds) within the test case with some similar logs and merge it with the main branch? So that in any of the builds, if the test fails, from the logs, we might get some insights to debug further on the root cause of the problem and provide a fix.

Since this process is quite new for us and I couldn't find any other similar example in Nimble, it would be really helpful if you could share your insights on the same.

m-akinc: > 1. Created a branch, enabled the test, and monitored for a few days but couldn't reproduce the test failure.

A build only runs against that branch when there's a new commit, so I'm not sure what you were monitoring for a few days. It seems like by doing 1 and 2, you only had maybe 10 runs of the tests. That's not likely enough to get a failure, since we only ever saw it fail once. In a situation like this, I would put the test in a loop to run it many, many times locally. Since it's a quick thing to do, I tried it myself, and I could not get it to fail in several thousand runs. Next, I would normally look at which tests had run before it when it failed, to see if there might be some state that another test had not cleaned up properly that was causing the failure. In this case, it turns out the test was the first RichTextEditorMentionListbox test that ran:
image

It is interesting that it happened to be the first test run (from the file), but it might also be a coincidence. In any case, the tests that ran immediately before the failure were some WaferMap tests. It's unlikely those would interfere with this test, but I tried enabling them along with the test we're interested in, just to see. Again, I could not get a failure.

At this point, I would probably try to add some logging and re-enable the test. Unfortunately, there's not much useful to log, since the error is coming from TipTap, and the error already tells us that the `Selection` has no range. The question is _why/how_ does the `Selection` have no range? I don't have a sense for anything we can log that would give us clues about that, especially given that we can't log anything right at the point where `collapseToEnd()` is being called.

Let's try re-enabling the test, closing the bug as not reproducible, and seeing if we ever get another failure. If we do, we can re-evaluate our options based on a better understanding of the frequency of this issue.

vivinkrishna-ni: Thank you so much for the details and trying it out yourself. I think I missed to mention about testing it locally multiple times for the past few days. I just wanted to make sure that sometimes the test might fail only on the build so tried only limited times by manually triggering a build against a branch and a PR. However, I didn't try testing like enabling the test that ran previously in the failed build. Thanks for that, we will make a note of it and try following it in the future.

As you mentioned, this test is first in the RichTextEditorMentionListbox block, in addition to the above ways, I tried running the entire RichTextEditorMentionListbox in a loop so that in any case if this test runs for the first time and fails, but that doesn't help too.

> why/how does the Selection have no range?

If there is no content in the editor, `collapseToEnd()` throws this error. So I just tried calling the `collapseToEnd()` before setting the @ character into the editor in the test, this throws because there is no content. Once after setting some content, when I called the same function, it ran without any issues. So in the failed build, there might be a chance that there is no content being set to the editor before pressing the Enter key. But it seems weird because the collapseToEnd is called within a mention extension that is triggered once after the @ character is added, otherwise, pressing enter will just move the cursor to a new line and not throw this error.

For now, I am re-enabling the test and added a comment saying if any failure is seen, report to the same issue.

change/@ni-nimble-components-2991eee5-08de-45ec-8f8a-eba5b9bbf8bb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Fix for intermittent test failure in rich test component for mention support",
```
m-akinc, line 3: Update this description, since we are not really fixing anything.

m-akinc (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
jattasNI: Thanks for all the ideas and details of what you tried in this discussion.

It sounds like some of the debugging strategies could have been documented better. Could you update [the intermittent test section of CONTRIBUTING](https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#handling-new-intermittent-test-tech-debt-issues) with any details that would have helped you (e.g. the idea to run a test locally in a loop)? This could happen in this PR or a follow up.

jattasNI: I'm a bit wary of re-enabling a test with suspected intermittency since it could disrupt other devs if it is still intermittent. I brainstormed a couple things we might do:
1. put a tag in the name of the test so that devs can know it's intermittent just from a test result, not needing to find the comment in source. e.g. `it('should commit mention into the editor on Enter #intermittent #1761'`. 
1. use the [existing `skip-tags` infrastructure](https://github.com/search?q=repo%3Ani%2Fnimble%20SkipFirefox&type=code) to execute intermittent tests differently. This could mean moving them to a separate command `npm run test:intermittent` which doesn't fail the pipeline but still publishes results so we can get an idea of the failure frequency.

If we decide to allow these tests, we should document the tagging format in CONTRIBUTING and include a note about how long a test can be tagged this way and what process we use to remind us to remove the tag if it is no longer intermittent.

I think 1 is worth doing. I'm ok not to do 2 right now but if it fails enough to be disruptive (once per week?) then we should consider it. I'm open to other opinions or ideas.

m-akinc: > I'm a bit wary of re-enabling a test with suspected intermittency since it could disrupt other devs if it is still intermittent.

I would be, too, except I don't expect it will be disruptive. I think I was the only one to run into this failure, and I didn't disable it immediately. I think it will be useful to see how often this actually fails.

rajsite: Looks like the test was first added [Dec 13](https://github.com/ni/nimble/blame/03bb474ca02a7ac9a13c7178af61544e9d742996/packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts#L1114) and the first identified failure was Jan 17 and it wasn't disabled until Jan 23.

So doesn't seem to fail often. And if it does fail again the policy would be to immediately disable and address it (and at that time maybe we would go with option 2 if we still believe we can't reproduce it locally).

I'm not sure about option 1 and adding the tag to the test name unless everyone is clear on the action. If it's tagged and fails for me does that mean it's okay, ignore and re-run, create an issue and disable it, let someone else know?

rajsite: If we do re-enable it in this state, saying it was truly a one-off over a period of a month and hundreds of builds then II'd say don't even have the comment inline. We are essentially saying it's not intermittent (because we are not adding any useful logging or info to help identify the cause of the intermittency) and to treat it with full force like normal if it is intermittent again (immediately disable and open an issue).

jattasNI: I'm fine to reenable it without a comment given the confidence that it was a one-off. Please also close the issue tracking it.

rajsite (approved):
Approved assuming we go with removing the comment and just re-enable the test

jattasNI (approved):

PR #1784, opened 2024-01-25
Created Blazor Wrapper for the Wafer Map Component

munteannatan:
# Pull Request

## 🤨 Rationale

There is a need for wafer map rendering in Blazor from an internal client.

## 👩‍💻 Implementation

Created the Blazor wrapper and required bindings.

## 🧪 Testing

Created an acceptance test and updated the demo app.

## ✅ Checklist


packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+        }
+
+        [Fact]
+        public async Task WaferMap_RendersCircleAsync()
+        {
+            await using (var pageWrapper = await NewPageForRouteAsync("WaferMapRendersCircle"))
+            {
+                var page = pageWrapper.Page;
+                var waferMap = page.Locator("nimble-wafer-map");
```
atmgrifter00, line 19: Is there anything else that we can assert against that would indicate that all of the proper JS calls were executed? At the moment, it seems like the assertions being made here can be done in a unit test. The purpose of the acceptance tests is to verify that any code defined in `NimbleBlazor.lib.module.js` executed as expected.

munteannatan: I think only the canvas painting can tell if the calls went through. I tried matching some bytes, but it doesn't seem to work. Do you have other ideas?

munteannatan: It seems that [playwright js uses this library](https://playwright.dev/docs/test-snapshots#maxdiffpixels) for comparing screenshots. [This ](https://www.nuget.org/packages/StronglyTyped.PixelMatch.net)is supposed to be a c# port but it uses Bitmap which is only supported on windows

packages/nimble-blazor/NimbleBlazor/wwwroot/NimbleBlazor.lib.module.js
```diff
@@ -169,5 +169,22 @@ window.NimbleBlazor = window.NimbleBlazor ?? {
         getValidity: function (themeProviderReference) {
             return themeProviderReference.validity;
         }
+    },
+    WaferMap: {
+        getValidity: function (waferMapReference) {
+            return waferMapReference.validity;
```
rajsite, line 175: Have you manually verified this works? I'm kinda surprised you didn't need to pass it through as JSON like all the others below.

munteannatan: I did not check, but I expected it to be similar with the table validity and the theme provider
image
will check right now

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
rajsite: Oh I forgot we have acceptance test infra in blazor! We should definitely make sure the code paths going through lib.Module are exercised like the getValidity method

I think we should avoid screenshot related approaches. I know playwright has the ability to evaluate JS logic on the page that we could use to probe the wafermap: https://playwright.dev/docs/evaluating

We would want that to be encapsulated in a page object. @msmithNI @jattasNI do yall have any insight on playwright page object and test patterns?

packages/nimble-blazor/NimbleBlazor/wwwroot/NimbleBlazor.lib.module.js
rajsite: Oh I forgot we have acceptance test infra in blazor! We should definitely make sure the code paths going through lib.Module are exercised like the getValidity method. Could running that be done in an acceptance test?

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class WaferMapTests : AcceptanceTestsBase
+    {
+        public WaferMapTests(PlaywrightFixture playwrightFixture, BlazorServerWebHostFixture blazorServerClassFixture)
+            : base(playwrightFixture, blazorServerClassFixture)
+        {
+        }
+
+        [Fact]
+        public async Task WaferMap_RendersCircleAsync()
```
rajsite: @atmgrifter00 I might be missing it but doesn't look like we have acceptance tests that exercise all the JS in NimbleBlazor.lib.module.js. The ones that come to mind are the getValidity methods. Did I just miss it? Is it known we have that gap and is it tracked anywhere? If not wanna make an issue to track it :)

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -358,5 +358,21 @@
...
+                MaxCharacters=2
+                DieLabelsHidden="false"
+                ColorScaleMode="WaferMapColorScaleMode.Linear"
+                @bind-Dies="Dies"
+                @bind-ColorScale="ColorScale"
+                @bind-HighlightedTags="HighlightedTags"
+            >
+            </NimbleWaferMap>
+            <NimbleButton @onclick=@(_ => AddDiesToRadius(10))>Increase Radius</NimbleButton>
+            <NimbleButton @onclick=@(_ => RemoveDiesFromRadius(10))>Decrease Radius</NimbleButton>
```
rajsite, line 375: I don't think the labels are super clear in the example. Could we use the labels "Increase Die Count" and "Decrease Die Count"?

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
atmgrifter00: This is a known gap, and I'm not sure we can address it. I don't believe we can properly assert that we've executed JS code if we can't access actual DOM elements. All Playwright gives us (afaik) is their `ILocator` objects which are an abstraction around the elements, and thus don't give us access to the mechanisms we need to assert as desired. If what we were asserting against was DOM state (like attributes) the locator APIs would be sufficient. @msmithNI can correct me here if I'm mistaken.

atmgrifter00: This is definitely a gap, but I was unaware of how to bridge it, and I'll admit I'm still not sure how the `evalutate/evaluateHandle` APIs can accomplish it, but maybe they do? It's definitely worth tracking in an issue.

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
munteannatan: The evaluated code returns different results based on where it's run. in Visual Studio 22 all tests pass with rgba 85 85 0 255, in the cli thei fail with mismatch of rgba 0000 and in the ci here its rgba 255 255 255 255

jattasNI: >do yall have any insight on playwright page object and test patterns?

You can see the patterns we've used in SLE [in the Skyline End2EndTests](https://dev.azure.com/ni/DevCentral/_git/Skyline?_a=contents&version=GBmaster&path=/End2EndTests/tests/shared/page-objects)

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @munteannatan this PR was moved into owner's review but does not have a passing build so is incomplete, what is the intention here?

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
rajsite: @munteannatan I see that this was moved into owners review and that the build is failing with your comment described here:

> The evaluated code returns different results based on where it's run. in Visual Studio 22 all tests pass with rgba 85 85 0 255, in the cli thei fail with mismatch of rgba 0000 and in the ci here its rgba 255 255 255 255

What is the expectation? Are you still investigating the behavior?

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
munteannatan: Sorry, my mistake.

The build is failing because the test with JavaScript evaluation is behaving inconsistently in the command line, vs editor and GitHub CI. I was hoping you would have an idea regarding a fix that I mentioned in my other comment.

I put it out of the draft because it seemed that was the last signaled issue and I was thinking you will get notified, but I should have just mentioned you.

If you think there are more issues to consider I will put it back in draft.

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
munteannatan: I was not currently investigating it because I had a couple of days of medical leave, sorry. 

My idea is that the canvas may be rendered with different dimensions or that the headless rendering may affect the expected colors.

If you have an idea for a fix or something I could try to find out I will do it next week

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
rajsite: I see. I unfortunately don't have too many specific thoughts on what the issue may be.

It could be timing, maybe try different waits before reading the canvas.
It could be visibility, maybe the canvas behaves different if it is not visible in the browser windows viewport.
It could be chrome setup, in our karma tests we pass [lots of flags](https://github.com/ni/nimble/blob/b61e3164866c8703fad31b2673fc88ebb694c6fd/packages/nimble-components/karma.conf.js#L17) to control different settings for avoiding background timing, animations, etc. Maybe something impacts how canvas behaves?

munteannatan: I will try to find the issue then, thanks for the hint!

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
rajsite: Thanks for the update and hope your doing well!

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
rajsite: Another thought is that for timing related issues, particularly in the table which also has batch updates, we added a private / internal method called [`processPendingUpdates`](https://github.com/ni/nimble/blob/3d77b048ac2f4cf191eb8464f1c014882e1d7d73/packages/nimble-components/src/table/index.ts#L706) that can way for any currently batched operations to be resolved. This pattern may be particularly useful when threaded /worker support is added. I would probably not recommend that it be made a public api but could be useful as private / internal for tests.

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/WaferMapTests.cs
rajsite: The evaluate/evaluateHandle methods weren't what I had in mind for this case. We want to see that the actual method execution and Blazor <-> JS communication is working. I think the example added in this PR where a button is pressed programmatically in playwright and a result is displayed somewhere (like old school LV boolean assert tests) would do the trick. Captured notes in: https://github.com/ni/nimble/issues/1804

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
munteannatan: the problem was that the eval was called before the canvas was there. I added an assert to wait for the canvas to be visible

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: New feature should be a minor change

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Createt Blazor Wrapper for Wafer Map Component",
```
rajsite, line 3: typo
```suggestion
  "comment": "Create Wafer Map Blazor Component",
```

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@ni/nimble-blazor",
+  "email": "33986780+munteannatan@users.noreply.github.com",
+  "dependentChangeType": "patch"
```
rajsite, line 6: Make sure to update the blazor status to incubating: 

https://github.com/ni/nimble/blob/24bb081b2930a9720f8420b6a36aaecedbff75ba/packages/nimble-components/src/tests/component-status.stories.ts#L527-L535

rajsite (approved):

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Createt Blazor Wrapper for Wafer Map Component",
+  "packageName": "@ni/nimble-blazor",
+  "email": "33986780+munteannatan@users.noreply.github.com",
```
rajsite, line 5: @msmithNI as @jattasNI and @atmgrifter00 are OOO can you give this PR a final review as Blazor code owner?

In particular I'm interested to know if the `GetValidityAsync` is a reasonable approach or if you are aware of other patterns we use. i.e. it is clicking a button via playwright, exercising the api under test in blazor and writing to a text field, and reading the result in the text field via playwright.

packages/nimble-blazor/NimbleBlazor/Components/NimbleWaferMap.razor.cs
```diff
@@ -0,0 +1,179 @@
...
+    public WaferMapOriginLocation? OriginLocation { get; set; }
+
+    /// <summary>
+    /// Represents the X coordinate of the minimum corner of the the grid bounding box for rendering the wafer map.
+    /// </summary>
+    [Parameter]
+    public double? GridMinX { get; set; }
+
+    /// <summary>
+    /// Represents the Y coordinate of the minimum corner of the the grid bounding box for rendering the wafer map.
```
msmithNI, line 40: Are these docs accurate / are they reversed with those of `GridMinY`? (right now the docs for `GridMaxX` say `Y coordinate of the minimum corner`)

packages/nimble-blazor/NimbleBlazor/Components/NimbleWaferMap.razor.cs
```diff
@@ -0,0 +1,179 @@
...
+            _colorScale = value;
+            _colorScaleUpdated = true;
+        }
+    }
+
+    /// <summary>
+    /// Will be triggered to inform the user that the state of hovering over a die has changed.
+    /// </summary>
+    [Parameter]
+    public EventCallback<WaferMapHoverDieChangedEventArgs> HoverDieChanged { get; set; }
```
msmithNI, line 142: It doesn't look like this event is hooked up correctly right now. In order to use Blazor's custom event bindings, in addition to what you have, you also need:
- NimbleBlazor.lib.module.js needs a call to `Blazor.registerCustomEventType()` for your event. The 1st arg is `nimblewafermapdiehoverchange` (note: without the `on` at the beginning), the browserEventName is what the Nimble component event name is (looks like `die-hover` for the wafer map), and you need to pass along the relevant parts of the event args (see the other calls in that file for an example)
- `NimbleWaferMap.razor` needs something like
```html
    @onnimblewafermapdiehoverchange="(__value) => HandleDieHoverChange(__value)"
```
then `HandleDieHoverChange` needs to trigger your `HoverDieChanged` `EventCallback`.

Since there's NimbleBlazor-specific JavaScript executing for the custom event code path, optimally you could add an acceptance test to exercise that too. Playwright has `hover()` on elements as well as `page.mouse` APIs that you could try using to hover over the wafer map. Then the acceptance test Razor file could listen to `HoverDieChanged` and toggle a checkbox, or something like that.

change/@ni-nimble-blazor-eed57bcb-a0ee-4959-bdc1-c9e233274889.json
msmithNI: The structure of that test looks reasonable to me, it's pretty similar to what we have for the others like DrawerTests/ DialogTests.

I'd expect most of the Blazor acceptance tests to look pretty similar:
- Load page with component, do any initial asserts (assert component is visible, for example)
- Interact with the component via Playwright actions. Either direct interaction, or via other buttons that trigger component APIs (like what this one is doing)
- Assert based on new component state, and/or assert that other controls got updated based on component logic that executed (like this one)

For this test, my only concern would be that it might pass even if the wafer map hasn't rendered, from what I can tell.

msmithNI (changes requested):

atmgrifter00 (approved):

msmithNI (approved):

PR #1786, opened 2024-01-29
Performance tests for wafermap

Razvan1928:
# Pull Request

## 🤨 Rationale

We wish to create performance tests using lighthouse, for that we are going to create most of the infrustructure

## 👩‍💻 Implementation

The wafer performance tests will appear in a lighthouse report generated by the CI after every push, or can be trigger manually using npm run performance -w packages/performance.

At runtime, while we create/populate the wafer component we are measuring the performance using windows.performance. The result will be available at 'User Timing marks and measures'

To avoid confusion the existing lighthouse tests were moved in the directory angular-workspace

The performance tests are built using vite

Despite lighthouse the landing site of the performance tests can be find in the Chromatic build

The values on the wafer dies are computed simply (i + j) % 100, where i and j are the coordinates of the die.

Also some math was used to compute the radius/sideLegth of the wafer.  
sideLength = Math.sqrt(diesCount / Math.PI) * 2;

it is derived from the circle's aria formula
A = π r²
where A in our case is the number of dies.

## 🧪 Testing

The results can be found in the CI

## ✅ Checklist


packages/nimble-components/index.html
atmgrifter00, line 1: Any reason not to host this in a subfolder called 'performance'?

wafermaplighthouserc.js
```diff
@@ -0,0 +1,22 @@
...
+              'http://localhost:58452/index.html',
+          ],
+          numberOfRuns: 1,
+      },
+      assert: {
+          assertions: {
+              'categories:performance': ['warn', { minScore: 0.9 }],
+              // TODO: reset to 0.9 as part of https://github.com/ni/nimble/issues/1090
+              // TODO: reset to error as part of https://github.com/ni/nimble/issues/1650
+              'categories:accessibility': ['warn', { minScore: 0.88 }]
```
atmgrifter00, line 15: Will this ultimately be handled by the lighthouse scoring on the Angular app?

Razvan1928: As discussed in private, maybe.

packages/nimble-components/index.html
Razvan1928: I managed to move it, good suggestion

packages/performance/index.html
atmgrifter00, line 1: Sorry, I should have been more specific. I still think this should live in the nimble-components folder, but should be in a non-package root folder.

Razvan1928: I changed its location to packages\nimble-components\performance\index.html, is it fine?

atmgrifter00: Yep, unless others have a stronger opinion.

packages/nimble-components/performance/index.html
```diff
@@ -0,0 +1,52 @@
...
+        
+        const colorScales = {
+            colors: ['red', 'orange', 'yellow', 'green'],
+            values: ['1', '33', '66', '100']
+        };
+
+        function createWaferMap(functionName, dieCount) {
+            performance.mark(`${functionName}_start`);
+            const waferMap = document.createElement('nimble-wafer-map');
+            waferMap.dies = generateDies(dieCount).wafermapDieSet;
```
munteannatan, line 19: can you leave the die data generation outside of the performance measures?

Razvan1928: sure, now it is outside

munteannatan (approved):

.github/workflows/main.yml
```diff
@@ -112,6 +112,13 @@ jobs:
         env:
           LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
 
+      # Run Wafer Lighthouse audit (from any push)
+      - name: Wafermap Performance Audit with Lighthouse CI
+        if: env.HAS_LHCI_GITHUB_APP_TOKEN == 'true' && github.event_name == 'push'
+        run: npm run wafer-performance
+        env:
+         LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
```
atmgrifter00, line 120: I'll defer to @rajsite on whether or not this is appropriate.

change/@ni-nimble-components-f50c155f-c243-4397-937d-fc4fa99c2aec.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I'd like to suggest a little bit of work to make it easier to leverage this infrastructure for other components:
1. Deploy the `/nimble-components/performance` directory in the Chromatic build so a user can visit it for each build (I believe that's configured in https://github.com/ni/nimble/blob/main/packages/site/.eleventy.js)
2. Link to it from [Storybook Getting Started](https://nimble.ni.dev/storybook/?path=/docs/getting-started--docs) alongside the Angular and Blazor apps
3. Change `performance/index.html` to be a simple table of contents that links to individual performance pages and move the existing contents to `performance/wafer-map.html`
4. Add some docs to `nimble-components/CONTRIBUTING.md` describing how to add a new performance test and look at its results
5. Add some docs to https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#pull-request-policies mentioning what is enforced by the pipeline and what to do if it fails

I'm open to feedback from @rajsite and others if any of this seems like the wrong direction or from you if it seems too complicated.

package.json
```diff
@@ -15,7 +15,8 @@
     "check": "beachball check --changehint \"Run 'npm run change' to generate a change file\"",
     "invoke-publish": "cross-env-shell beachball publish --yes --access public --message \\\"applying package updates [skip ci]\\\" -n $NPM_SECRET_TOKEN",
     "validate": "npm run build && npm run lint && npm run test",
-    "performance": "lhci autorun"
+    "performance": "lhci autorun",
+    "wafer-performance": "lhci autorun --collect.staticDistDir='./packages/nimble-components' --collect.numberOfRuns=1 --collect.url='http://localhost:58452/performance/index.html'"
```
jattasNI, line 19: Can you configure these parameters in a `.lighthouserc.js` file like [we do here](https://github.com/ni/nimble/blob/main/lighthouserc.js)? One way to do this would be to put a new lighthouserc file in the `nimble-components/performance` directory and then change this command to set that as the working directory before running `lhci`.

Rationale: keeping configuration consistent makes it easier to understand and we generally prefer config files over command line flags since it's easier to read

change/@ni-nimble-components-f50c155f-c243-4397-937d-fc4fa99c2aec.json
jattasNI: Milan mentioned he had some more ideas around this (e.g. separating the performance tests into a separate package) but we didn't get a chance to discuss them yet today. If it makes sense for you, I'd suggest waiting to address this feedback until we have a chance to agree on a concrete proposal. We'll aim to share something back before end of day tomorrow.

package.json
Razvan1928: did it

packages/site/.eleventy.js
```diff
@@ -4,7 +4,8 @@ module.exports = eleventyConfig => {
         'landing/dist': './',
         '../../packages/nimble-components/dist/storybook': 'storybook',
         '../../angular-workspace/dist/example-client-app': 'storybook/example-client-app',
-        '../../packages/nimble-blazor/dist/blazor-client-app': 'storybook/blazor-client-app'
+        '../../packages/nimble-blazor/dist/blazor-client-app': 'storybook/blazor-client-app',
+        '../../packages/nimble-components/performance': 'storybook/wafer-performance'
```
rajsite, line 8:
```suggestion
        '../../packages/nimble-components/performance': 'storybook/performance/wafer-map'
```

And also update the main storybook page to have a link to it: https://github.com/ni/nimble/blob/b61e3164866c8703fad31b2673fc88ebb694c6fd/packages/nimble-components/src/tests/nimble-intro.stories.mdx#L21

.github/workflows/main.yml
Razvan1928: I deleted this part, now the wafer will be tested once with the other tests using npm run performance

change/@ni-nimble-components-f50c155f-c243-4397-937d-fc4fa99c2aec.json
Razvan1928: I made a refactoring of most of the code, that’s why @munteannatan , @atmgrifter00, I reset your votes. The wafer performance tests are now aligned with the existing ones of the other components. The only concern may be that the wafer takes a few seconds to load, we may consider making it smaller. Some of the above comments may no longer be necessary, I’m waiting for a second review.

rajsite: I have step by step recommendations to make it easy for you to implement and to show you're following existing patterns, v it should be discussed before you go a different direction

> The wafer performance tests are now aligned with the existing ones of the other components.

The other components are not trying to run performance tests, they are using lighthouses other capabilities on page size measurement, accessibility scores, etc. It is a poor performance test because Angular adds overhead and all the other controls lacking at the same time add overhead.

rajsite (changes requested):

.github/workflows/main.yml
```diff
@@ -108,7 +108,14 @@ jobs:
       # Run Lighthouse audit (from any push)
       - name: Performance Audit with Lighthouse CI
         if: env.HAS_LHCI_GITHUB_APP_TOKEN == 'true' && github.event_name == 'push'
-        run: npm run performance
+        run: npm run performance -w angular-workspace
```
rajsite, line 111: Thiis branch still may be a work in progress but make sure to pull in the changes discussed in the comment thread as shown here: https://github.com/ni/nimble/compare/main...move-lh-to-angular-workspace

The goal is to not have component specific changes in the workflow file and I showed you how to accomplish that.

Razvan1928: I pulled the changes, but it's good that you highlighted this topic, for some reason I though that "npm run performance" will run only the tests from angular-workspace, that's why for making it more clear I added the workspace for both commands
image
but, as I tried just now, it was not needed

change/@ni-nimble-components-f50c155f-c243-4397-937d-fc4fa99c2aec.json
Razvan1928: @rajsite @jattasNI I think I implemented most of your feedback, with 2 exceptions, Jesse's comments 5 and 3, 5: I discussed with @munteannatan and he wants, at least the wafer performance tests pipeline, to not fail anytime. 3: now index.html doesn't contain wafer-specific code, the wafers are appended from wafer-map.ts to index.html, would you want a different approach?

jattasNI (commented):
I didn't get a chance to review before going out of office. If other reviewers approve then you don't need to wait for me.

packages/performance/package.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "@ni-private/performance",
+    "version": "1.0.0",
+    "private": true,
+    "description": "A package to measure the performance of the Nimble Components",
+    "scripts": {
```
rajsite, line 6: For the build, performance, and start commands I made it so that each component that will be tested has their own commands. I don't think we want one enormous lighthouse test for every component. The chance that the components will interfere with each other would be higher. Instead each test can independently on their own page.

packages/performance/wafer-map/lighthouserc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+    ci: {
+        collect: {
+            staticDistDir: '../dist/wafer-map',
+            numberOfRuns: 1,
+            settings: {
+                onlyAudits: ['user-timings']
```
rajsite, line 7: There are a whole bunch of performance metrics that are unnecessary and clutter the output and slow testing time. I referenced the lighthouse docs to show only the user-timings we are capturing

packages/performance/wafer-map/src/index.html
```diff
@@ -0,0 +1,17 @@
...
+    <title>Wafer Map Performance Tests</title>
+    <meta name='viewport' content='width=device-width, initial-scale=1'>
+    <meta name="description" content="Wafer map performance tests">
+    <script type="module" src="./index.ts"></script>
+</head>
+<body>
+    <p>See devtools performance timing for results.</p>
+    <p>Loading...</p>
+    <p class="running" hidden>Running...</p>
+    <p class="finished" hidden>Finished!</p>
```
rajsite, line 15: As we are cleaning up after the tests I added these status messages so it is easier to know when the tests are running or finihsed.

packages/performance/wafer-map/src/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en-us">
+<head>
+    <meta charset='utf-8'>
+    <meta http-equiv='X-UA-Compatible' content='IE=edge'>
+    <title>Wafer Map Performance Tests</title>
+    <meta name='viewport' content='width=device-width, initial-scale=1'>
+    <meta name="description" content="Wafer map performance tests">
```
rajsite, line 8: This page will be published and linked from stroybook so iit should have the basics of a good html page.

packages/performance/wafer-map/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    const start = performance.now();
+    const waferMap = document.createElement(waferMapTag);
+    waferMap.dies = waferMapDies;
+    waferMap.colorScale = {
+        colors: ['red', 'orange', 'yellow', 'green'],
+        values: ['1', '33', '66', '100']
+    };
+    document.body.append(waferMap);
+    await waitAnimationFrame();
+    performance.measure(benchmark, { start });
```
rajsite, line 24: I did not see the value of including the marks as previously it was even trying to clear the marks. Only included measures.

packages/performance/wafer-map/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    document.body.removeChild(waferMap);
+}
+
+function main(): void {
+    void (async (): Promise<void> => {
+        document.querySelector<HTMLParagraphElement>('.running')!.hidden = false;
+        await runBenchmark('wafer_10K_dies', 10 ** 4);
+        await runBenchmark('wafer_100K_dies', 10 ** 5);
+        await runBenchmark('wafer_1M_dies', 10 ** 6);
```
rajsite, line 34: Made the names more descriptive / useful. When looking at the stats "small" does not tell you as much about the benchmark as "10k dies"

packages/performance/wafer-map/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    return new Promise(resolve => {
+        requestAnimationFrame(() => resolve());
+    });
+}
+
+async function runBenchmark(benchmark: string, dieCount: number): Promise<void> {
+    const waferMapDies = generateDies(dieCount);
+
+    const start = performance.now();
+    const waferMap = document.createElement(waferMapTag);
```
rajsite, line 16: Elements should be construucted with `document.createElement` instead of using the constructor function,

packages/performance/wafer-map/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    waferMap.dies = waferMapDies;
+    waferMap.colorScale = {
+        colors: ['red', 'orange', 'yellow', 'green'],
+        values: ['1', '33', '66', '100']
+    };
+    document.body.append(waferMap);
+    await waitAnimationFrame();
+    performance.measure(benchmark, { start });
+
+    document.body.removeChild(waferMap);
```
rajsite, line 26: I made sure to clean-up the previous wafer-map before running the next test to avoid the chance that the wafer-maps will interfere with each other during benchmarking

packages/site/.eleventy.js
```diff
@@ -4,7 +4,8 @@ module.exports = eleventyConfig => {
         'landing/dist': './',
         '../../packages/nimble-components/dist/storybook': 'storybook',
         '../../angular-workspace/dist/example-client-app': 'storybook/example-client-app',
-        '../../packages/nimble-blazor/dist/blazor-client-app': 'storybook/blazor-client-app'
+        '../../packages/nimble-blazor/dist/blazor-client-app': 'storybook/blazor-client-app',
+        '../../packages/performance/dist': 'storybook/performance'
```
rajsite, line 8: Should need to remap each component test that is added here, made the performance package build ouput like:

dist/component-name/*

so we can just copy all the contents in one copy

change/@ni-nimble-components-f50c155f-c243-4397-937d-fc4fa99c2aec.json
rajsite: I addressed 3 in the changes I pushed by creating an index page for the benchmarks that is linked from storybook. This avoids a storybook change every time a new benchmark is added.

Agree that no status checks from the performance tests should fail the pipeline for now. Currently it is just being used to track values.

rajsite: I think with the changes I've pushed everything in the thread is resolved.

angular-workspace/lighthouserc.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
```
rajsite, line 1: @Razvan1928 I pushed some changes to the branch and added comments in the top-level PR describing the changes. Let me know if yo have any questions about the changes otherwise I think it's ready to merge.

rajsite (approved):

PR #1788, opened 2024-01-30
POC for worker based rendering

munteannatan:
# Pull Request

## ✅ Checklist


packages/nimble-components/src/wafer-map/modules/hover-handler.ts
```diff
@@ -45,36 +45,53 @@ export class HoverHandler {
...
-        ).data;
-        let rgbaSum = 0;
-        for (const color of rgba) {
-            rgbaSum += color;
-        }
-        return rgbaSum > 0;
+    private getWaferMapDie(
+        _dieCoordinates: PointCoordinates
+    ): WaferMapDie | undefined {
+        // const colIndex = this.wafermap.dieMatrix.dieColIndexArray.indexOf(
```
munteannatan, line 67: @rajsite In this place the handler was searching for the specific x and y coordinate and returning the value and metadata. the arrow table does not support filtering or any kind of search without iterating over the full table [as mentioned here](https://github.com/apache/arrow/issues/13233). a solution can be this [library on top of arrow](https://uwdata.github.io/arquero/api/verbs#filter) like you mentioned. I will test these two options and compare them to see if they have different performances

packages/nimble-components/build/generate-workers/source/renderWorker.ts
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+    }
+
+    public renderDies(
+            tableBuffer: Uint8Array
+    ): void {
+        const start = this.performanceTest !== undefined ? self.performance.now() : undefined;
+
+       this.dieTable = tableFromIPC(tableBuffer);
```
munteannatan, line 157: @rajsite the metod of transferring the IPC buffers is [suggested here](https://uwdata.github.io/arquero/api/table#toArrowBuffer) too. this transfer seems to have an impact on the performance so I will try the [geoarrow method](https://github.com/geoarrow/geoarrow-js?tab=readme-ov-file#web-worker-utilities) and compare.

packages/nimble-components/build/generate-workers/source/renderWorker.ts
```diff
@@ -0,0 +1,167 @@
...
+    public emptyMatrix(): void {
+        this.dieTable = tableFromArrays({
+            colIndex: new Int32Array(),
+            rowIndex: new Int32Array(),
+            value: new Float32Array()
+        });
+    }
+
+    public rerenderDies(): void {
+        for (const tableRow of this.dieTable) {
```
munteannatan, line 100: @rajsite the biggest time consumer seems to be [this iterator](https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/js/src/table.ts#L220). There seems to be several parsing stages for this, so i think that may be an issue. I will look more into the visitor pattern and see if I can format the table to hold just the bare information for rendering and maybe  transform it into the underlying array buffers to access the values quicker

packages/nimble-components/src/wafer-map/modules/hover-handler.ts
rajsite: Yea the hope is that these higher-level apis give easier to maintain code / algorithms while having good performance. 🤞

packages/nimble-components/build/generate-workers/source/renderWorker.ts
rajsite: I think today we would want to have the user create an Arrow Table and transfer it to the worker like the geoarrow method.
I could see it being useful in the future to allow users to assign Arrow Table instances or ArrowIPC instances, for example if the backend started returning ArrowIPC we could transfer that directly to the worker. But I think today serializing as ArrowIPC first just adds overhead.

packages/nimble-components/build/generate-workers/source/renderWorker.ts
rajsite: Yea es6 iterators are pretty slow.. They don't offer another api to iterate quuicker?

munteannatan: I tried different options but all of them were pretty slow, so eventually I got out of the table just the relevant columns and sent them over as array buffers. I also cached the column index positions so now the performance is comparable to before

packages/nimble-components/build/generate-workers/source/renderWorker.ts
munteannatan: I got the ideea from geoarrow to split the table in buffers which can get transferred, but there is no need to re-create the table itself in the worker

rajsite:
@munteannatan is this draft PR still valuable to keep around? I think the behavior covered has been merged, can we close this?

Issue #1442, opened 2023-08-18
Leverage appropriate number format options in number-text table column once supported in all browsers


mollykreis:
## 🧹 Tech Debt
As part of #1414, we had to make some concessions and add some work-arounds due to certain number format options not being supported yet in Firefox.

Once Firefox has parity support with other browsers for these options for at least 6 months we should update our numeric formatters for the number-text table column to use these features.

Specifically the options we would like to use but cannot yet are:
- `useGrouping: auto`
- `signDisplay: negative`
- `roundingPriority: 'lessPrecision'`

m-akinc:
Should check on this again in early 2024

rajsite:
@m-akinc please wait to address this till after the unit support is merged

m-akinc:
Waiting for a Chromatic update to get new enough browsers to support these options. Could be any time now.

PR #1791, opened 2024-01-30
Update usage of Intl.NumberFormat options

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes: #1442 

## 👩‍💻 Implementation

The options cited in the issue are:

- `useGrouping: 'auto'` - We previously were setting this to `true`, which means "display grouping separators even if the locale prefers otherwise." Instead we want `"auto"`, which is "display grouping separators based on the locale preference." The default for this option is `"auto"` unless `notation` is `"compact"`. We never set `notation` to `"compact"`, so we can just leave `useGrouping` to use the default.
- `signDisplay: 'negative'` - Though this option value has been [supported in all browsers since 8/23](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_numberformat_options_signdisplay_parameter_negative), there does not seem to be a version of Typescript whose definition of `Intl.NumberFormatOptions` allows it. There is an [open issue for adding support](https://github.com/microsoft/TypeScript/issues/56269). As a workaround, I've just added type assertions where necessary.
- `roundingPriority: 'lessPrecision'` - Just like `signDisplay: 'negative'`, the `roundingPriority` option has been [supported in browsers since 8/23](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_numberformat_options_roundingpriority_parameter), but Typescript doesn't have support yet. It is tracked by the same issue linked above, and our type assertion serves as a workaround for this, too. By using this option, we can get rid of the separate `leadingZeroFormatter` in `DefaultUnitFormat` and simplify the logic.

## 🧪 Testing

Existing tests pass.

## ✅ Checklist


m-akinc:
@mollykreis since you're somewhat familiar with this code, would you mind buddying it?

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -36,18 +35,21 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
...
             this.scaledUnitFormatters.set(
                 scaledUnit.scaleFactor,
                 scaledUnit.scaledUnitFormatFactory({
                     locale,
-                    intlNumberFormatOptions
+                    // Hack to avoid a ts error about signDisplay not accepting the value 'negative'.
+                    // The value has been supported by browsers since 8/23, but TypeScript still hasn't
+                    // added it to the type definitions. See https://github.com/microsoft/TypeScript/issues/56269
+                    intlNumberFormatOptions:
+                        intlNumberFormatOptions as Intl.NumberFormatOptions
```
mollykreis, line 49: This cast is a little concerning to me because it would hide any incorrect configuration of `intlNumberFormatOptions`, not just the unsupported `signDisplay` option.

Did you consider any other ways of writing this that would scope the cast to just the type that needs it?

One idea would be to do something like:
```ts
type SignDisplay = Intl.NumberFormatOptions['signDisplay'];

export class DecimalUnitFormat {
    ...

    const intlNumberFormatOptions = {
        maximumFractionDigits,
        minimumFractionDigits,
        signDisplay: 'negative' as SignDisplay
    };
}
```

packages/nimble-components/src/utilities/unit-format/tests/default-unit-format.spec.ts
```diff
@@ -303,11 +303,6 @@ describe('DefaultUnitFormat', () => {
                 value: 2000000,
                 expectedFormattedValue: '2,000 x1000'
             },
-            {
-                name: 'uses unit-scaled value when deciding whether to format with leading-zero formatter',
```
mollykreis, line 307: Could you explain why this test isn't needed anymore?

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
m-akinc: Love it. Thanks! Also did similar things in `DefaultUnitFormat`.

packages/nimble-components/src/utilities/unit-format/tests/default-unit-format.spec.ts
m-akinc: This test ensured that we didn't pick the leading-zero formatter to use when the original (unscaled) number has a leading zero but the scaled number does not (e.g. 0.123.. scaled to 12.3..). We no longer have a separate formatter for numbers that have a leading zero, so this is no longer a meaningful test case.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -33,10 +32,14 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
...
     ) {
         super();
+        type SignDisplay = Intl.NumberFormatOptions['signDisplay'];
         const intlNumberFormatOptions = {
             maximumFractionDigits,
             minimumFractionDigits,
-            useGrouping: true
+            // Hack to avoid a ts error about signDisplay not accepting the value 'negative'.
+            // The value has been supported by browsers since 8/23, but TypeScript still hasn't
+            // added it to the type definitions. See https://github.com/microsoft/TypeScript/issues/56269
```
mollykreis, line 41: Maybe it's just me, but I'm not a huge fan of calling something in our code a "hack". What do you think about calling this a "workaround" instead?

Same in the other similar comments.

packages/nimble-components/src/utilities/unit-format/default-unit-format.ts
```diff
@@ -69,20 +61,19 @@ export class DefaultUnitFormat extends UnitFormat {
...
-                    locale,
-                    intlNumberFormatOptions:
-                        this.leadingZeroIntlNumberFormatOptions
+                    // Hack to avoid ts error about roundingPriority not being a known option.
+                    // It has been supported by browsers since 8/23, but TypeScript still hasn't
+                    // added it to the type definitions. See https://github.com/microsoft/TypeScript/issues/56269
+                    intlNumberFormatOptions: {
+                        ...this.defaultIntlNumberFormatOptions,
+                        roundingPriority: 'lessPrecision'
+                    } as Intl.NumberFormatOptions
```
mollykreis, line 70: Do we need this `as Intl.NumberFormatOptions` cast still? It will hide any real issues with the type.

packages/nimble-components/src/utilities/unit-format/default-unit-format.ts
```diff
@@ -69,20 +61,19 @@ export class DefaultUnitFormat extends UnitFormat {
...
                 })
             );
         }
         this.exponentialScaledUnitFormatter = unitScale.baseScaledUnit.scaledUnitFormatFactory({
             locale,
+            // Same as above comment
```
mollykreis, line 76: It isn't clear to me what comment this is referring to (and how that applies to this line). Can you clarify this comment?

mollykreis (approved):

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
m-akinc: Ha! Okay, okay... it's a [workaround](https://en.wikipedia.org/wiki/A_rose_by_any_other_name_would_smell_as_sweet).

packages/nimble-components/src/utilities/unit-format/default-unit-format.ts
m-akinc: Yes we do, but we will catch any issues with everything other than `roundingPriority` because I am now typing `defaultIntlNumberFormatOptions` as `Intl.NumberFormatOptions`.

packages/nimble-components/src/utilities/unit-format/default-unit-format.ts
m-akinc: Whoops, should have removed that.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
```diff
@@ -18,7 +18,6 @@ export class DecimalUnitFormat extends UnitFormat<DecimalUnitFormatOptions> {
     private readonly maximumFractionDigits: number;
```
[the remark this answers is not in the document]
rajsite: We had a similar kind of issue with the dialog api, so maybe we can have a shared file of DOM type definition updates we can burn down over time: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/dialog/index.ts#L25

packages/nimble-components/src/utilities/unit-format/tests/decimal-unit-format.spec.ts
```diff
@@ -144,15 +144,15 @@ describe('DecimalUnitFormat', () => {
     });
 
     describe('with unit', () => {
-        class TestScaledUnitFormat extends ScaledUnitFormat {
+        class TestScaledUnitFormat extends IntlNumberFormatScaledUnitFormat {
```
rajsite, line 147: Can you explain why this change was made for the test unit scale? Maybe consistency?
Seems pretty much identical to the one in default-unit-format.spec.ts now, should it be moved to a shared fixture?

change/@ni-nimble-components-9d33a263-abcb-4b8e-a5f1-1ab02c9b4974.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The [diffs in the current Chromatic build](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=65bab7b6817624700576dac2) are concerning, with several number-text columns no longer rendering data. Can you take a look?

jattasNI (changes requested):

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
m-akinc: I don't think it's possible to modify the type of an existing property on an interface (which is what I need to make `signDisplay` accept a value of `'negative'`). I was able to use declaration merging to add the missing `roundingPriority` property, though. Seems like I can put the .d.ts file anywhere under `/src` and it gets pulled in automatically. I created `type-extensions.d.ts` directly under `/utilities` and put both the extended `Intl.NumberFormatOptions` and `HTMLDialogElement` types in there. `Dialog` and `Drawer` each had a `dialog` property that I changed from being `ExtendedDialog` to `HTMLDialogElement`, but they should be internal, so no breaking change.

change/@ni-nimble-components-9d33a263-abcb-4b8e-a5f1-1ab02c9b4974.json
m-akinc: Well, this is super annoying. Chromatic's versions of Chrome and Firefox are too old (9/22 and 8/22 respectively) to support the options I'm trying to use. Historically, Chromatic has [updated their browsers](https://www.chromatic.com/docs/infrastructure-release-notes/) about once every 6 or 12 months, and with their last update in February 2023, we should be due for another any time now.

Though it looks wrong, I'm inclined to say we just live with this small blind spot in our visual testing until they publish another update. The Chromatic tests are most useful for catching minor visual changes. We're probably not going to be doing any new development on the number-text column that would result in those kinds of subtle visual changes. Problems resulting in more significant visual changes are likely to be caught during dev via Storybook or functional tests.

packages/nimble-components/src/utilities/unit-format/tests/decimal-unit-format.spec.ts
m-akinc: Yeah, it was to be consistent, but I didn't go the next step to share it. I've now moved it into a separate file that they both import from.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
rajsite: > Seems like I can put the .d.ts file anywhere under /src and it gets pulled in automatically.

Can you provide more evidence about why that is working. It's not clear to me what is making that association. I'd expect at a minimum an import or type reference to the file.

> I don't think it's possible to modify the type of an existing property on an interface

I'm pretty sure that is possible. You might need to make sure you are modifying the correct type through the global

change/@ni-nimble-components-9d33a263-abcb-4b8e-a5f1-1ab02c9b4974.json
jattasNI: My vote would be to put this on the backburner until Chromatic publishes a browser update. For me, fixing the tech debt isn't worth regressing our test suite, even if it is a temporary regression that might not allow any bugs to slip through in the near term.

Another idea would be to keep both the old and new code around for now, and fall back to the old code if we can detect that the new code isn't supported. But I imagine that would be messy and it would still result in us needing a tech debt issue to track removing the old code.

Happy to discuss with the team at our meeting today.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
m-akinc: > Can you provide more evidence about why that is working.

Our tsconfig sets `include` to `["src"]`. This will [make Typescript aware](https://www.typescriptlang.org/tsconfig#include) of any *.d.ts files under `src`. Because we are extending built-in types (which do not require explicit imports), our extensions are automatically used without explicit imports. I can't find direct documentation of this, but [this StackOverflow answer](https://stackoverflow.com/questions/58265985/extending-built-in-types-in-typescript) implies it, because the .d.ts file never has to be imported anywhere.

> I'm pretty sure that is possible.

I've spent more time than I'd like trying and failing. The [best answer StackOverflow has](https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file) is to create a new interface that modifies the property type of an existing interface. I can't find a way to change the existing interface (other than adding to it).

change/@ni-nimble-components-9d33a263-abcb-4b8e-a5f1-1ab02c9b4974.json
m-akinc: Chromatic updated, but I'm now seeing improper rendering in the Chromatic test in Safari. I can't reproduce with any version of WebKit I try. According to the documentation, the new version of Chromatic Capture Cloud uses Safari 17, but when I dumped the value of `navigator.userAgent` from the test, it indicates that it's version 17.4 (which is weird, because 17.4 is currently only in beta):
image

m-akinc: Jesse tested Safari Tech Preview 17.4, and it still did not reproduce the issue. It's unclear what version/build of Safari is being used that seems to uniquely have this problem. Jesse stated that he was fine living with the spurious diff, and I think I agree.

packages/nimble-components/src/utilities/unit-format/default-unit-format.ts
```diff
@@ -3,7 +3,7 @@ import type { ScaledUnitFormat } from './scaled-unit-format/scaled-unit-format';
 import type { UnitScale } from './unit-scale/unit-scale';
 import { passthroughUnitScale } from './unit-scale/passthrough-unit-scale';
 
-type NumberStyle = 'default' | 'leadingZero' | 'exponential';
+type SignDisplay = Intl.NumberFormatOptions['signDisplay'];
```
rajsite, line 6: can you do something like:

```ts
// workaround comment
const signDisplay = 'negative' as Intl.NumberFormatOptions['signDisplay'];
```

and use that constant instead in the spots below?

packages/nimble-components/src/utilities/type-extensions.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * This is a workaround for an incomplete definition of the native dialog element. Remove when using Typescript >=4.8.3.
+ * https://github.com/microsoft/TypeScript/issues/48267
+ * @internal
+ */
+interface HTMLDialogElement {
+    showModal(): void;
+    close(): void;
+}
```
rajsite, line 9: Can we make this align with the TS definitions (https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/27fa26c78ce2401310e26a13ea233323707d88c2/baselines/dom.generated.d.ts#L10012-L10037)

```
showModal(): void;
close(returnValue?: string): void;
```

packages/nimble-components/src/utilities/type-extensions.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare namespace Intl {
+    // roundingPriority has been supported by browsers since 8/23, but TypeScript still hasn't
+    // added it to the type definition. See https://github.com/microsoft/TypeScript/issues/56269
+    interface NumberFormatOptions {
+        roundingPriority?: 'auto' | 'morePrecision' | 'lessPrecision';
```
rajsite, line 15: Can we align this to the PR that looks like will add these types? https://github.com/microsoft/TypeScript/pull/56902/files#diff-8cd7de35e3e99beffd3c7f2e8e537e6073b41bc1c10caff2e41be7c97a3c3595R13

```
roundingPriority?: "auto" | "morePrecision" | "lessPrecision" | undefined;
```

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
rajsite: Could you test with a npm pack tgz build in an app that does not enable skipLibCheck (so we know it's type checking us). If that works then I'm sold. 💯

packages/nimble-components/src/utilities/type-extensions.d.ts
m-akinc: Sure, but I'm curious: why bother with the redundant `undefined`?

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
m-akinc: I updated TestInsights's dependency to use a tgz of `nimble-components`. Then I added some code to create and use a `NumberTextUnitFormat` (to make sure it uses the code where we needed the extended `Intl.NumberFormatOptions` definition). It built without issue. Is that sufficient? The app build won't be type-checking the js logic in `nimble-components`, and none of our public APIs include `Intl.NumberFormatOptions` or `HTMLDialogElement`, so I'm not really sure what I need to verify.

packages/nimble-components/src/utilities/type-extensions.d.ts
rajsite: You can configure TS to distinguish between optional and assigning undefined: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#exact-optional-property-types---exactoptionalpropertytypes

And based on the other comment doesn't seem worth making an arbitrary difference to avoid conflicts with that type we don't own.

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
rajsite: We are modifying a global type that's not scoped to our components or apis. I'd guess our types are included in the build. By importing our lib and having a dep on it I'd expect the global types in that app to be mutated to have the properties we define. So without components dialog, etc would not have the type and with it it should. So HTMLDialogElement etc should be changed for those apps after importing our lib. And my concern about aligning the types to prevent conflicts.

rajsite (approved):
approved assuming confident builds will be good

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
m-akinc: From testing, it seems our type extensions are not being included by the client app. I don't think that's necessarily a problem, though. As long as the extended types are not part of our public API (e.g. we don't expect clients to pass us `Intl.NumberFormatOptions`), isn't it actually better that we don't modify the client's type definitions? As long as our builds work and client builds work, haven't our type extensions served their purpose?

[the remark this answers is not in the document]
jattasNI: After a chat with @rajsite I was confused about how this was working so I did a bit of reading / thinking. I think what's happening is:
1. nimble-components consumes the custom .d.ts file during its build to do type checking on anything that uses those types
     - I didn't track down exactly how `tsc` finds those .d.ts files but I suspect it's because the tsconfig.json says `include: "src"`
2. the build output of nimble-components is just js files and .d.ts files for the public APIs of those JS files
     - ok, technically the build output also contains source maps which point to inlined ts in the .d.ts files, but that's just used for runtime debugging, not app compilation
3. the build output of nimble-components does NOT include the custom internal .d.ts file
4. when an app builds against nimble-components it doesn't try to do any type checking of the contents of nimble. skipLibCheck just controls whether it type checks the .d.ts files of libraries
5. As long as Nimble doesn't use any of the augmented types in its public API, they won't impact the app's build at all.

So based on my understanding, I think the current approach is sane.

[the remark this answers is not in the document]
m-akinc: That was my understanding, too. 👍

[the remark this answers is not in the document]
rajsite: Gotcha, it was not clear to me that the .d.ts files are not included in the build output at all.

>  use any of the augmented types in its public API 

> As long as the extended types are not part of our public API 

HTMLDialogElement is in the public api from TypeScript's perspective:
image

Same for Intl.NumberFormatOptions:
image

But I agree that `showModal`, `close` ,`signDisplay`, and `roundingPriority` are not.

Is that what y'all are saying? How did y'all validate that? The same way I did by text searching all the .d.ts files?

I think we should capture some docs on this for future editors of the .d.ts files. @m-akinc I'll resolve this comment after adding additional ones.

packages/nimble-components/src/utilities/type-extensions.d.ts
```diff
@@ -0,0 +1,21 @@
+/**
```
rajsite, line 1: Can you move this file to `src/global-type-extensions.d.ts`

This isn't really a scoped utility and has a global impact so lets not hide it here.

rajsite, line 1: Can you add a comment to this file describing some of the things discussed for future devs:

```
This file has modifications to global types used for nimble-components builds:
- The types file should not be in the build output
  - You should not see a file with these types created in the dist build
- Type additions added to this file should not appear in any public API definitions
  - The fields added to global types should not be observable in any built .d.ts file by TypeScript. To verify you can do a text search of generated .d.ts files in the dist folder for the fields being added and verify they are not associated with the augmented types.
-  This file should only impact the nimble-components library build and not users of nimble-components. Any type changes that impact clients should not be placed here.
```

packages/nimble-components/src/utilities/unit-format/decimal-unit-format.ts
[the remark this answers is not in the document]
rajsite: > But I agree that showModal, close ,signDisplay, and roundingPriority are not.
> 
> Is that what y'all are saying? How did y'all validate that? The same way I did by text searching all the .d.ts files?

I added a comment about how I confirmed that these types are not in the public api. If y'all have another approach that was used can mention there.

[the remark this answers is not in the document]
m-akinc: > Is that what y'all are saying?

I was saying that our APIs do not expect/require clients to use `showModal`, `close`, `signDisplay`, or `roundingPriority`, so we shouldn't care about whether the client's definitions of `HTMLDialogElement` or `Intl.NumberFormatOptions` contain them.

> How did y'all validate that?

To see if the client's definitions had those functions/properties, I added some code to TestInsights that tried to use `showModal()` on `HTMLDialogElement` and set `roundingPriority` on `Intl.NumberFormatOptions`. The build (and editor) gave me errors about `roundingPriority`. Interestingly, though, it did not complain about `showModal()`, because it seemed to be using a definition from a typescript installation in the VS Code app data directory.

rajsite (approved):

PR #1792, opened 2024-01-30
Include href in Angular's NavigationGuard

mollykreis:
# Pull Request

## 🤨 Rationale

Including the `href` in the `NavigationGuard` will greatly simplify some usages of the table in SLE. Specifically, it keeps an application from having to look up the value applied to a given cell, which isn't always simple (or synchronous) when the data is stored in an Angular `Observable`.

## 👩‍💻 Implementation

Pass the `href` to the `NavigationGuard`.

## 🧪 Testing

- Manually tested with a locally build package in SLE
- Updated unit test

## ✅ Checklist


mollykreis:

change/@ni-nimble-angular-b0792776-c1c4-4c48-8f2b-80b116f75192.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
m-akinc, line 2: If we typed the `href` argument as optional, we could make this a patch update, right? Major version bumps have the downside of cutting off clients from getting automatic updates, so I wonder if it's worth making a minor concession to avoid it. I could be convinced either way.

m-akinc (approved):

change/@ni-nimble-angular-b0792776-c1c4-4c48-8f2b-80b116f75192.json
rajsite: I don't think it's actually an API breaking change. You can ignore extra parameters. See [playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcByBDAbgSwOYYw4QB2A4gK4ZQAmcAvHABRQQDuASsAMbQ0CSNAFxwAzjCg4SeOAB84FEjWAAzKcBoAaOAAsoqkeMnSAlAwB8cAEYQIAG2AYSAbgCwAKA+8S463Yw6Ipi4BESklNR0jExm9JYSFMBu7t6+Vv46AExB2PiExORUtAzMrJw8fIKxlgCENWVcvLSCyUA).

So I think marking it optional is unnecessary for the goal of API breaking changes. It's also maybe not desirable since the type would be incorrect. It would say that href is sometimes not passed to your typeguard callback buut it is always passed.

m-akinc: If there was existing client code that called a `NavigationGuard` (e.g. `blah('')`), that would break when an extra, required parameter is added to the type. 

I'm not sure I would call the type "incorrect" if the href was optional... maybe just unecessarily permissive, or less precise. But I hope we can still be friends. 😅

rajsite (approved):

change/@ni-nimble-angular-b0792776-c1c4-4c48-8f2b-80b116f75192.json
rajsite: > If there was existing client code that called a NavigationGuard (e.g. blah('')), that would break when an extra, required parameter is added to the type.

Client code doesn't call the navigation guard. A navigation guard instance is given to the directive and the directive calls it. The directive will never call it without passing an href.

I don't think we are tracking all possible usages of any of our types. We are saying if you give us this callback, this is what we are invoking it with. So saying we will sometimes invoke it with href undefined is incorrect.

>  But I hope we can still be friends. 😅

I'm not sure (j/k)

mollykreis: I was on the fence about calling this a major vs minor change since I don't actually expect clients to need to update their code to handle the new function signature. I'll change it to minor.

rajsite:
@atmgrifter00 pretty minor change, I put it in auto-merge as a bypass

Issue #1758, opened 2024-01-17
Table support dynamic loading of child rows


m-akinc:
## 😯 Problem to Solve

Allow a row to act like it has children, but only fetch/display them upon expanding it.

## 💁 Proposed Solution

Details are found throughout the [HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/data-hierarchy-hld.md#lazy-loading)

```[tasklist]
### Tasks
- [ ] https://github.com/ni/nimble/pull/1794
- [ ] #1813
- [ ] https://github.com/ni/nimble/pull/1829
- [ ] #1832
- [ ] #1846
```

PR #1794, opened 2024-01-31
Add basic support for delayed data hierarchy

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1758

This PR adds support for claiming a row has children that are not present in the data.

The ability to mark a row as loading its children will be included in a follow up PR.

## 👩‍💻 Implementation

The majority of the implementation is adding the following based on [the data hierarchy HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/data-hierarchy-hld.md#lazy-loading):
- `TableRecordHierarchyOptions` interface
- `TableRecordDelayedHierarchyState` enum (no `loadingChildren` state in this PR)
- `setRecordHierarchyOptions` function on the table

I created a `RowOptionsManager` to track the options for the table. This new manager, along with the `ExpansionManager` (which has a few updates), have the majority of the logic for this feature.

## 🧪 Testing

- Updated the matrix tests for the table to make one row have a state of `canLoadChildren`
    - Note: This update to the matrix tests exposed an existing bug with the alignment of parent rows in Safari. In the version of Safari used in Chromatic, the cell contents are 1px pushed to the right, but I wasn't able to reproduce this locally in webkit. We can look into this as a separate task if we want.
- Wrote new unit tests

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -1376,6 +1380,464 @@ describe('Table', () => {
...
+                            delayedHierarchyState:
+                                TableRecordDelayedHierarchyState.canLoadChildren
+                        }
+                    }
+                ]);
+                await waitForUpdatesAsync();
+                expect(pageObject.isCollapseAllButtonVisible()).toBeFalse();
+            });
+
+            it('collapse all button remains visible when row options are updated but one row is still forceExpandable', async () => {
```
atmgrifter00, line 1480:
```suggestion
            it('collapse all button remains visible when row options are updated but one row still has canLoadChildren configured', async () => {
```

atmgrifter00 (commented):
Overall looks solid, just some probing questions to be answered.

packages/nimble-components/src/table/index.ts
```diff
@@ -300,6 +314,26 @@ export class Table<
...
+            id: string,
+            options: TableRecordHierarchyOptions
+        }[]
+    ): Promise<void> {
+        await this.processPendingUpdates();
+
+        if (!this.isHierarchyEnabled()) {
+            // Do not allow hierarchy options to be configured when hierarchy is not
+            // configured on the table.
+            return;
```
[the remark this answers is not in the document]
mollykreis: Perhaps my intention in the first bullet point you mentioned would have been better captured by saying _all options will be **discarded** if parentIdFieldName is not configured_. The goal was to keep a client from configuring hierarchy options if the table wasn't in hierarchy mode, but not to throw existing configuration out in case a binding did something odd like switch from a defined `parentIdFieldName` to `undefined` and back. That being said, if we think clients will run into binding-related timing issues then perhaps we should do what you're suggesting and hold on to their updated hierarchy options.

@rajsite, any opinion one way or another here?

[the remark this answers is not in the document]
rajsite: > That being said, if we think clients will run into binding-related timing issues then perhaps we should do what you're suggesting and hold on to their updated hierarchy options.

Yea the discussed goal was to minimize the coupling to binding order. Users do need to worry about it for idFieldName but they don't need to worry about it for parentIdFieldName. So we should save whatever they specify for hierarchy state regardless if hierarchy mode is on or off at the moment. It's state that applies to the records so it only should shift around if the records change.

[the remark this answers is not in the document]
mollykreis: I've updated the implementation, and I've also removed the bullet point that says _all options will be ignored if parentIdFieldName is not configured on the table_ since that behavior is already described in the other bullet point @atmgrifter00  mentioned (_will not render delayed hierarchy state (loading or expandable) if the table's parentIdFieldName is not configured; however, the options will remain cached within the table if the parentIdFieldName becomes undefined, and that cached configuration will render in the table if the table's parentIdFieldName is changed back to a non-undefined value_).

packages/nimble-components/src/table/models/row-hierarchy-options-manager.ts
```diff
@@ -0,0 +1,42 @@
+import {
+    TableRecordDelayedHierarchyState,
+    type TableRecordHierarchyOptions
+} from '../types';
+
+/**
+ * Manages the hierarchy options that have been configured on a row.
+ */
+export class RowHierarchyOptionsManager {
```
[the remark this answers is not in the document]
mollykreis: I don't think there is much value in creating this as a new class since right now everything it manages has some overlap with the existing ExpansionManager. I've removed the new class and moved the necessary logic into the ExpansionManager.

packages/nimble-components/src/table/index.ts
```diff
@@ -904,10 +923,9 @@ export class Table<
         this.tableData = rows.map(row => {
             const isGroupRow = row.getIsGrouped();
             const hasParentRow = isGroupRow ? false : row.getParentRow();
-            // we check row.original.subRows below because row.subRows is populated for group rows
-            // which we don't want to include
-            const isParent = row.original.subRows !== undefined
-                && row.original.subRows.length > 0;
+            const isParent = !isGroupRow
+                && this.isHierarchyEnabled()
```
atmgrifter00, line 927: Should this check just move into `getRowCanExpand`?

atmgrifter00 (approved):

packages/nimble-components/src/table/index.ts
mollykreis: That's an interesting question. I actually thought about that, and I was a bit undecided myself because it isn't as straight forward as putting the check in `getRowCanExpand` because we want group rows to be expandable even when hierarchy isn't enabled. Therefore, the options I can think of are:

_Option 1:_ Keep this code as-is where the check is within `refreshRows`

_Option 2:_ Update `getRowCanExpand` to have a check that looks something like this:

```ts
private readonly getRowCanExpand = (
        row: TanStackRow<TableNode<TData>>
    ): boolean => {
        if (row.getIsGrouped()) {
            return;
        }
        return this.isHierarchyEnabled() && this.expansionManager.isRowExpandable(row);
    };
```

_Option 3:_ Update the expansion manager to know about whether or not _parentIdFieldName_ is configured

- I went with Option 1 since it was the most straight forward.
- I'm not a huge fan of Option 2 since I like the current simplicity of `getRowCanExpand` in that it is just a wrapper around the ExpansionManager
- I was concerned about the complexity of Option 3. I didn't want the ExpansionManager to have to keep track of the parentIdFieldName, but I actually think the change there could be as simple as passing the current value of parentIdFieldName to the ExpansionManager's isRowExpandable function.

I'm going to try out Option 3, and I'll go that direction unless it starts feeling overly complicated.

rajsite (commented):
Initial feedback

packages/nimble-components/src/table/models/expansion-manager.ts
```diff
@@ -63,28 +73,65 @@ export class ExpansionManager<TData extends TableRecord> {
...
+
+    public setHierarchyOptions(
+        hierarchyOptions: {
+            recordId: string,
+            options: TableRecordHierarchyOptions
+        }[]
+    ): void {
+        this.hierarchyOptions.clear();
+        for (const { recordId, options } of hierarchyOptions) {
+            this.hierarchyOptions.set(recordId, options);
```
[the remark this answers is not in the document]
rajsite: Actually we also talked about an optimization path here of sharing object references (using the same options object instance in multiple rows).

There is a way to preserve that if we copy the whole structure at once with structuredClone: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone
image

packages/nimble-components/src/table/tests/table.stories.ts
```diff
@@ -363,6 +382,11 @@ const metadata: Meta<TableArgs> = {
                 'A function that returns `true` if the configuration of the table is valid and `false` if the configuration of the table is not valid.',
             control: false
         },
+        addDelayedChildrenIfNeeded: {
+            name: 'setRecordHierarchyOptions()',
+            description: setRecordHierarchyOptionsDescription,
+            control: false
```
jattasNI, line 388: nit: I have a mild preference to show some kind of control for any paramaters that are writable. It's less helpful here since the configuration won't impact what shows in the code preview, but it can still be nice to be able to play with a setting toggled on and off.

The simplest configuration I can think of would be two radio buttons with names something like "No delayed children" and "Last record has delayed children" but I'm open to other ideas.

I'm also open to deferring this until the progress indicator PR or scrapping the idea if it complicates the implementation too much.

packages/nimble-components/src/table/types.ts
```diff
@@ -76,6 +78,20 @@ export interface TableValidity extends ValidityObject {
     readonly invalidParentIdConfiguration: boolean;
 }
 
+/**
```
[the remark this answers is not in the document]
jattasNI: At first glance I prefer the existing names. I find it helpful to be able to be able to tell from their names that the state and options types are applied to a table record, not the entire table. 

If your goal is to have a name for the pair of recordId+options, one idea is to call that a "config". The idea being that "state" and "options" are the type of configuration being applied and "config" is the how that state is applied to a record. "Configuration" or "Setting" would be two other possible synonyms.

```ts
TableRecordHierarchyConfig
{
      recordId: string;
      options: TableHierarchyOptions;
}
```

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -1376,6 +1380,513 @@ describe('Table', () => {
                 );
             });
         });
+
+        describe('delay loaded hierarchical data', () => {
+            const hierarchicalData: SimpleTableRecord[] = [
```
jattasNI, line 1385: nit: Since this initial data is flat, not hierarchical, maybe call it `initialData` or `parentData` or something like that?

Could also consider marking it [readonly](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#a-new-syntax-for-readonlyarray) or `as const` if the intent is that no test should modify its contents.

jattasNI (approved):

packages/nimble-components/src/table/types.ts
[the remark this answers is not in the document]
mollykreis: How about this?

```ts
export interface TableRecordHierarchyConfiguration {
    recordId: string;
    options: TableRecordHierarchyOptions;
}

export interface TableRecordHierarchyOptions {
    delayedHierarchyState: TableRecordDelayedHierarchyState;
}

export const TableRecordDelayedHierarchyState = {
    none: undefined,
    canLoadChildren: 'canLoadChildren'
} as const;
```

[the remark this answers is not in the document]
rajsite: I think the type names `TableRecordHierarchyConfiguration`  vs `TableRecordHierarchyOptions` sound like the same thing / hard to distinguish 

What about naming the first one for the method name `TableSetRecordHierarchyOptions `?

```
export interface TableSetRecordHierarchyOptions {
    recordId: string;
    options: TableRecordHierarchyOptions;
}

export interface TableRecordHierarchyOptions {
    delayedHierarchyState: TableRecordDelayedHierarchyState;
}

export const TableRecordDelayedHierarchyState = {
    none: undefined,
    canLoadChildren: 'canLoadChildren'
} as const;
```

packages/nimble-components/src/table/tests/table.stories.ts
rajsite: What about making it a separate story? like how the table column configuration it broken into separate workflow:
image

Can have a "Table/Delayed Loading" story and it can have a control where you explicit set the state of a row with different options:
- default
- hasChildren
- loading

And can move the expected workflow discussion into Usage Docs for that story

packages/nimble-components/src/table/types.ts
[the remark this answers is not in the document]
jattasNI: That would work for me

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -156,6 +156,12 @@ export class TableValidator<TData extends TableRecord> {
         return requestedRecordIds.filter(id => this.recordIds.has(id));
     }
 
+    public getItemsWithPresentIds<T extends { recordId: string }>(
```
rajsite, line 159: Should be able to use the new type being defined here

rajsite (commented):
tests and impl look good! Just want to see storybook updates to finish

packages/nimble-components/src/table/tests/table.stories.ts
jattasNI: I think a separate story within the same MDX could work. Another way to split it would be to put all the flat data examples in one story and the hierarchical data examples (both pre-loaded and delay-loaded) in a second story. No strong preference from me, whatever works better as you start moving parameters around.

Brainstorming on that idea a little further, maybe we could split up the huge parameter table we currently have into a few smaller ones that each live in their own section within the MDX. e.g. Maybe Validity be its own section that's just a parameter table, no embedded nimble-table. Just brainstorming, that doesn't have to be in this PR.

mollykreis: I've updated the stories to split out delayed-hierarchy into its own story, with more of the documentation in the mdx file rather than within the controls table.

packages/nimble-components/src/table/tests/table.mdx
```diff
@@ -16,6 +16,42 @@ import * as tableStories from './table.stories';
...
+
+<Description of={tableStories.delayedHierarchy} />
+<DocsStory of={tableStories.delayedHierarchy} expanded={false} />
+<Controls of={tableStories.delayedHierarchy} />
+
+### Delay-loaded Hierarchy Usage
+
+The expected usage of the dynamically loaded hierarchy is as follows:
+
+1. Configure `idFieldName` and `parentIdFieldName` on the table
```
mollykreis, line 29: Note: This expected usage is copied directly from the HLD. I decided to leave in the documentation that included `loadingChildren` to avoid extra churn here and since I already added a note saying that the loading state wasn't implemented yet, but I could also see the argument that we should adjust these steps based on what is implemented today.

packages/nimble-components/src/table/tests/table.stories.ts
jattasNI: I like how it looks! But in [the current build](https://60e89457a987cf003efc0a5b-owxluyyebl.chromatic.com/?path=/docs/components-table--docs) clicking the expander for the delay loaded rows doesn't ever populate the rows with data. (macOS Firefox in case that matters)

It was also slightly unexpected to see the "Localizable labels" entry in the parameter table for delay-loaded hierarchy.

https://github.com/ni/nimble/assets/10500124/8a5df9a6-9458-49c1-a38a-17070f7114b5

mollykreis: > clicking the expander for the delay loaded rows doesn't ever populate the rows with data

That is correct. The row state is now completely driven by the story's control. It seemed odd to have the state coming from the control but then change the state out from under the user when they interact with the table, but maybe that's better than doing nothing when the expand button is clicked?

jattasNI: Ah, gotcha, I was still stuck on the previous behavior. I can see both sides. It'll be useful to have demos that clearly show the effect of putting a row into each state. And it'll be useful to have a demo that automatically transitions a row through the states to show how the feature behaves as a whole (and what a client has to do to achieve that).

I'm not sure of the best way to show both of those demos and I'm about to be OOO for a while so I'll have to leave you to figure it out. But I'm fine with keeping it as you have it for this PR and then adding the more comprehensive demo (either in storybook or the example app) in a follow up once the loading progress indicator is visible.

rajsite: Maybe instead of the exact row state the control can be phrased in the workflow steps and describe the expectations better. If we experiment with that lets do it in a separate PR, keeping the extra complexity of this niche / not common workflow out of the main table story was my goal.

packages/nimble-components/src/table/models/expansion-manager.ts
```diff
@@ -31,6 +39,9 @@ export class ExpansionManager<TData extends TableRecord> {
             return false;
         }
 
+        if (row.subRows.length === 0) {
+            return false;
+        }
```
rajsite, line 44: So this makes it so that after the user expands a row we don't expand immediately (i.e. how it doesn't expand in storybook)?

I think I find the behavior surprising. If you look at the row-expand-toggle event it shows that the row is expanded each time you click it:
image

From an event lifecycle that seems surprising, maybe unexpected for an app, how does it go from false -> true without going true -> false in between? And that doesn't match the visual. The newState isn't actually expanded true.

Lets say there is a slow event reaction time (like Blazor server has to respond to the event over the network so the transition to loading is delayed). I think for a user it would be surprising, it might feel broken, they may click it again firing off more events.

I wonder if instead it should be disabled? The representation of a row that is:
- expanded in the `canLoadChildren` state is a disabled expand button
- collapsed in the `canLoadChildren` state is an enabled expand button

Thoughts? If there is more to consider / it's not a straightforward tweak we could avoid blocking the PR and track it in an issue.

rajsite (approved):

packages/nimble-components/src/table/tests/table.stories.ts
```diff
@@ -16,17 +19,51 @@ import {
...
+    title: 'Components/Table',
+    decorators: [withActions],
+    parameters: {
+        actions: {
+            handles: [
+                'action-menu-beforetoggle',
+                'action-menu-toggle',
+                'selection-change',
+                'column-configuration-change',
+                'row-expand-toggle'
```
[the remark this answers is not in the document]
rajsite: Since the action pane works I'm guessing the mdx is outdated

packages/nimble-components/src/table/models/expansion-manager.ts
mollykreis: The current behavior comes from the line in the HLD that says _a row with no children and a `delayedHierarchyState` of `canLoadChildren` will always be collapsed_. The goal with that statement was to handle the cases that (1) moving from `loadingChildren` -> `canLoadChildren` ensures the row's state will be collapsed, and (2) a row with a state of `canLoadChildren` that had children & then the children are removed goes back to being collapsed.

I agree, though, that having the button not change is a bit inconsistent. I'm not sure that disabling the button is the right way to go. Perhaps I can handle the specific cases I mentioned more explicitly rather than always having the row be collapsed. It almost certainly won't be a straightforward tweak to the existing code, so I'll probably address that in a follow up PR.

packages/nimble-components/src/table/tests/table.stories.ts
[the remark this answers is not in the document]
mollykreis: The mdx was copied from the HLD, which was outdated. I've updated both.

packages/nimble-components/src/table/tests/table.stories.ts
mollykreis: I'm leaving this as-is for now, but I'm happy to iterate on how we present this workflow in storybook in a separate PR.

PR #1801, opened 2024-02-02
Refactor common dialog/drawer code into a base class

m-akinc:
# Pull Request

## 🤨 Rationale

Refactor to avoid duplication between dialog and drawer.

## 👩‍💻 Implementation

Created common base class `DialogBase`.

## 🧪 Testing

Existing tests pass.

## ✅ Checklist


m-akinc:
@atmgrifter00 would you mind buddying this?

change/@ni-nimble-components-dc071140-b97c-473f-b929-1268a3aa3a40.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
atmgrifter00, line 2: What makes this "minor" instead of "patch"?

atmgrifter00 (approved):

change/@ni-nimble-components-dc071140-b97c-473f-b929-1268a3aa3a40.json
m-akinc: I wasn't sure, but I thought since we were technically exporting a new type (`DialogBase`) it should be "minor". But it's intended as an internal type, so I'm not sure.

rajsite:
Adding @mollykreis to take a look since they are familiar with original code and the issue being addressed

packages/nimble-components/src/dialog-base/index.ts
```diff
@@ -0,0 +1,103 @@
...
+        } else {
+            this.notifyClosed(UserDismissed);
+        }
+        return true;
+    }
+
+    /**
+     * @internal
+     */
+    public closeHandler(): void {
```
m-akinc, line 85: This is the `CloseWatcher` mitigation. Basically everything else is just refactoring `Drawer` and `Dialog`.

packages/nimble-components/src/dialog/tests/dialog.spec.ts
```diff
@@ -1,14 +1,15 @@
...
 import { waitForUpdatesAsync } from '../../testing/async-helpers';
+import type { ExtendedDialog } from '../../dialog-base';
 
 // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
 async function setup<CloseReason = void>(
     preventDismiss = false
 ): Promise<Fixture<Dialog<CloseReason>>> {
     const viewTemplate = html`
-        <nimble-dialog ?prevent-dismiss="${() => preventDismiss}">
+        <nimble-dialog ?prevent-dismiss="${() => preventDismiss ?? null}">
```
mollykreis, line 12: Why do you need the `?? null` here? Isn't the type of `preventDismiss` a `boolean`, which would mean it is never not defined?

packages/nimble-components/src/drawer/index.ts
```diff
@@ -25,109 +24,83 @@ declare global {
...
+        this.dialogElement.classList.remove('animating');
         if (this.closing) {
-            this.dialog.classList.remove('closing');
-            this.dialog.close();
-            this.closing = false;
-            this.resolveShow!(this.closeReason);
-            this.resolveShow = undefined;
+            this.dialogElement.classList.remove('closing');
+            this.dialogElement.close();
+            this.doneClosing();
```
mollykreis, line 103: It doesn't look like `doneClosing` is called anywhere but here, so I'm curious why you decided to create a helper function to wrap two lines of this function.

mollykreis (approved):

packages/nimble-components/src/dialog/tests/dialog.spec.ts
m-akinc: You're right. I must have been confused.

packages/nimble-components/src/drawer/index.ts
m-akinc: It was for symmetry with the `startClosing` function that I added. One sets `closeReason` and sets `closing = true`, and the other uses `closeReason` and sets `closing = false`.

packages/nimble-components/src/dialog/styles.ts
```diff
@@ -44,6 +45,10 @@ export const styles = css`
         display: flex;
     }
 
+    dialog${focusVisible} {
```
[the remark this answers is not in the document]
m-akinc: Good point. I'm now adding a 1px green border. I'm waiting on Brandon's approval.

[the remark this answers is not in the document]
m-akinc: He wanted a 2px border instead, so I made that change.

jattasNI (commented):
I didn't get a chance to review before going out of office. If other reviewers approve then you don't need to wait for me.

packages/nimble-components/src/drawer/index.ts
```diff
@@ -25,109 +24,83 @@ declare global {
...
-    }
-
-    /**
-     * Closes the drawer
-     * @param reason An optional value indicating how/why the drawer was closed.
+     * Closes the Drawer
+     * @param reason An optional value indicating how/why the Drawer was closed.
      */
-    public close(reason: CloseReason): void {
+    public override close(reason: CloseReason): void {
```
rajsite, line 43: I am just starting to look at it. My first impression is that if around half the methods of the base class are overridden without calling super for the default behavior then I'm not sure how much value we are getting in the abstraction / maybe the abstraction is misaligned. I'll have to review it more closely to understand the goals of the split / responsibilities of the base class.

packages/nimble-components/src/dialog-base/index.ts
```diff
@@ -0,0 +1,108 @@
...
+export interface ExtendedDialog extends HTMLDialogElement {
+    showModal(): void;
+    close(): void;
+}
+
+/**
+ * A base class for components based on a native dialog.
+ */
+// eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+export abstract class DialogBase<CloseReason = void> extends FoundationElement {
```
rajsite, line 19: I'd probably frame this as "Modal" of which the Dialog and Drawer are implementations of. But like the other comment says haven't fully thought about what the specific responsibilities of the Modal base class should be

packages/nimble-components/src/drawer/index.ts
rajsite: @m-akinc to prevent blocking the functional aspect of the change, can we split the refactor and event handling updates into separate PRs?

m-akinc: Moved the functional aspects into PR #1848

jattasNI: I agree after scanning that the benefit is less compelling if only one of the derived classes uses many of the methods in the base class. Some general strategies around this:
1. don't put those methods in the base class. Each derived class has its own implementation. If, after doing this, the base class is mostly empty or isn't cohesive (e.g. has an `open` but not a `close`) then we could just get rid of it.
2. narrow the scope of those methods to only include the common functionality, then have the derived classes call super and then add their custom functionality
3. make the base class handle both cases and choose between them based on some configuration of the object (don't love this option, probably a mixin or larger class hierarchy would make more sense, but none of that feels right for this tiny class hierarchy)
4. find a way to make the behaviors identical so they can be handled entirely by the base class

@m-akinc wanna take a pass at applying one or more of these strategies?

rajsite: One thought I had is having Modal as the base class, the promise workflow is all handled there (we don't need implementation specific open and close value tracking and promise resolution). So things like resolveShow are private there and the dialog reference is protected (calling showModal on a dialog element is managed in the base class).

And then the base class has lifecycle methods that can be overridden / called and some state tracking. 

Something like states: {opening, open, closing, closed}
and methods (some protected) for the transitions: open, finishOpening, close, finishClosing.

So a dialog without animations in close will immediately call finishClosing while one with other work like animations can call finishClosing later. Think the logic will be clearer too tracking a state vs implied knowledge based on whether resolveShow is set or not.

m-akinc: > the dialog reference is protected

I guess it needs to be public for `ref` to work in the template.

jattasNI (approved):
@m-akinc FYI there are some lint errors but otherwise looks good!

packages/nimble-components/src/modal/index.ts
```diff
@@ -0,0 +1,123 @@
...
+ * This is a workaround for an incomplete definition of the native dialog element. Remove when using Typescript >=4.8.3.
+ * https://github.com/microsoft/TypeScript/issues/48267
+ * @internal
+ */
+export interface ExtendedDialog extends HTMLDialogElement {
+    showModal(): void;
+    close(): void;
+}
+
+export const ModalState = {
```
rajsite, line 15: types should be a types.ts file

packages/nimble-components/src/drawer/index.ts
rajsite: Looking a lot better. The thing I want to emphasize was the goal "So things like resolveShow are private there" and "And then the base class has lifecycle methods that can be overridden / called and some state tracking."

The intention is that the ModalState tracking is completely handled in the base class and private. All modals go through the state transitions (off the top, didn't double check what was implemented) closed -> opening -> open -> closing -> closed. And the transitions between states are handled by methods that do the transitions finishOpening, finishClosing, etc.

Previously `resolveShow` represented the state of the control and now `state` represents the state of control. Either way the goal of my statement is that the state management is private / encapsulated in Modal.

packages/nimble-components/src/modal/index.ts
```diff
@@ -0,0 +1,123 @@
...
+            this.resolveShow = resolve;
+        });
+    }
+
+    /**
+     * Closes the Modal
+     * @param reason An optional value indicating how/why the Modal was closed.
+     */
+    public close(reason: CloseReason): void {
+        if (!this.open || this.state === ModalState.closing) {
```
rajsite, line 75: nit/pref: instead of accessing the public accessor inside the methods, which could have different meanings over time, lets compare the states instead. A nice syntax shortcut is something like `[ModalState.open, ModalState.closing].includes(this.state)`

packages/nimble-components/src/modal/index.ts
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public readonly dialogElement!: ExtendedDialog;
+
+    protected state: ModalState = ModalState.closed;
+
+    /**
+     * True if the Modal is open/showing, false otherwise
+     */
+    public get open(): boolean {
+        return this.state === ModalState.open;
```
rajsite, line 51: Should the dialog be considered "open" if the state is opening or closing? Do we have test coverage here?

packages/nimble-components/src/modal/index.ts
m-akinc: Actually, we don't even have to check if closing. The open check is sufficient.

rajsite: The type pattern we use for enums is to treat ModalState as a string. That's how we get type checked attribute strings in the angular templates, so the type `("open" | "closing")[]` is expected.

Looks like the surprising bit is the type for `includes` where I'd be on team the TypeScript is being too strict: https://github.com/Microsoft/TypeScript/issues/26255

So shortcut not worth the effort, just some good ol fashion `this.state === foo || this.state === bar`

packages/nimble-components/src/modal/index.ts
m-akinc: It turns out that we do have a test that expects `open` to be true while closing. So I've changed `open` to mean `state !== closed`. Neither Dialog nor Drawer distinguish between `open` and `opening` states, though, so I don't think we need to test that the property `open` is true when `state === opening`.

packages/nimble-components/src/drawer/index.ts
m-akinc: > The intention is that the ModalState tracking is completely handled in the base class and private.

I had tried to do that, but it didn't seem like it would work. On second look, we can make it work with a couple assumptions:
1. Derived classes can override the `startOpening/Closing` methods, but should not call them (except from inside an override). I'm having `Modal` set the state immediately before calling those methods, so the state won't be set correctly if the derived class calls `startOpening` somewhere else. 
2. The base implementation for `finishOpening/Closing` sets the state, so if a derived class overrides one of those functions, it _must_ call the base implementation (or the state won't get updated properly).

I wish there were a nice way to enforce these assumptions, but instead I've just added comments to the base implementations.

packages/nimble-components/src/modal/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    protected startClosing(reason: CloseReason | UserDismissed): void {
+        this.finishClosing(reason);
+    }
+
+    // Overrides must call the base implementation, so that state gets updated.
+    protected finishClosing(reason: CloseReason | UserDismissed): void {
+        this.state = ModalState.closed;
+        this.dialogElement.close();
+        this.doResolveShow(reason);
+    }
```
rajsite, line 132: I think the issues here for needing a buunch of overrides are that the Modal class is trying to implement a specific workflow, where the transition is preset so i.e. startClosing immediately calls finishClosing. Instead there should be lifecycle hooks that each subclass is implementing.

startOpening is abstract and in the dialog implementation iit immediately calls finishOpeniing while in the drawer implementation it does its animations beforee calling finishClosing.

m-akinc: I made `startOpening/Closing` abstract. It doesn't really change much, other than moving the default implementations to `Dialog`. I guess it technically reduces the number of overrides by 2 (if you don't count the abstract method implementations). That's all you wanted me to change, right?

packages/nimble-components/src/modal/index.ts
```diff
@@ -0,0 +1,135 @@
...
+     */
+    public close(reason: CloseReason): void {
+        if (this.state !== ModalState.open) {
+            throw new Error('Not open');
+        }
+        this.state = ModalState.closing;
+        this.startClosing(reason);
+    }
+
+    /**
```
rajsite, line 85: The responsibilities of cancel and close handlers don't seem quite right. I think similar to some previous commits they seem over coupled to specific workflows. Modal instances like drawer calling "super.cancelHandler" to start the transition to Closing and check the preventDismiss state seems like unclear responsibilities for a method called "cancelHandler".

Looking at it again I also have a more general question of why are we not designing animations into the Modal base class. I'm actually wondering now if it is really intentional that the dialog has no animations. Seems like even short, fast animations would be useful for all modals.

Some directions:
- We can continue to discuss and iterate on the responsibilities of the base class, including seeing if animation responsibilities make more sense there, it probably makes sense to discuss it first to prevent review cycles or continue the approach of aync comment and feedback if you want to keep going that way
- We pause the refactor for now since we don't have a clear need for shared Modal base classes at the moment (I'm not aware of new modals we need to implement or if we want to share more behavior like animations across all the modals)

m-akinc: I'll de-prioritize and move to draft until I have new changes to review.

rajsite:
Closing the draft as unprioritized right now. Think it would be useful to reconsider if adding additional modal types / decide to share more functionality (i.e. animations for dialogs and drawers)

PR #1808, opened 2024-02-07
Update lighthouse url path and categories

rajsite:
# Pull Request

## 🤨 Rationale

In the GitHub status checks for lighthouse, the name of the status check is based on the URL of the page. This means two completely different pages that end up with the same url (for example served from `/`) would overwrite the same status check. 

Also, as-is the urls are not very descriptive:
image

## 👩‍💻 Implementation

This PR updates the URLs so that the lighthouse status check includes the package that was analyzed:
image

The PR also updates the Angular lighthouse status check to remove the PWA category since that is most useful for applications trying to implement PWA features.

## 🧪 Testing

Ran on CI.

## ✅ Checklist

### DIFF
--- a/angular-workspace/lighthouserc.js
+++ b/angular-workspace/lighthouserc.js
@@ -1,11 +1,17 @@
 module.exports = {
     ci: {
         collect: {
-            staticDistDir: './dist/example-client-app',
+            // Serve from dist folder so Angular project name is part of url path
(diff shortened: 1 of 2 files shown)


fredvisser (approved):
Looks great - 👍

rajsite:
Overriding owner review as Fred took a look @jattasNI

Issue #1650, opened 2023-11-06
Lighthouse failing with axe-core Error: Cannot read properties of undefined (reading 'actualNode')


rajsite:
## 🧹 Tech Debt

Looks like lighthouse is failing intermittently on main, since: https://github.com/ni/nimble/commit/090ed10ab16edf555d7ed2114eb8eb54130b2c64
See: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1699279168857-58396.report.html

And consistently with the changes proposed here: https://github.com/ni/nimble/pull/1646
See: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1699279168857-58396.report.html

With error message:
```
axe-core Error: Cannot read properties of undefined (reading 'actualNode')
```
See:
![image](https://github.com/ni/nimble/assets/1588923/fd50b4e0-2cc0-449e-9233-f1afa2c5a3dd)

Issue status: @kjohn1922 found a workaround in #1646, we should try and report it upstream to axe-core or lighthouse

rajsite:
@kjohn1922 thanks for digging in and finding a workaround for what was breaking axe core in lighthouse. If you're feeling ambitious can try to reproduce and report it upstream to https://github.com/dequelabs/axe-core

rajsite:
Found a way to reproduce the axe-core error in dev tools:
1. Navigate to a built page that causes the error in lighthouse, i.e. currently: 
2. Open dev tools, in the sources tab under breakpoints enable `Pause on caught exceptions`
3. In dev tools in the console run the latest axe-core on the page, i.e. 
   ```js
   document.body.append(Object.assign(document.createElement('script'), {src: 'https://unpkg.com/axe-core/axe.js', onload: () => {axe.run().then(results => {console.log(results)}).catch(err => {console.error('Something bad happened:', err.message)});}}));
   ```
   Note: the script is loading unminified code and targeting the latest version. As of writing the version resolved to `https://unpkg.com/axe-core@4.8.4/axe.js`
4. The script will download and run, may take a few seconds, but should break on the line resulting in an error

rajsite:
The lighthouse update merged as part of #1786 caused this card issue to come-up again so the card was disabled in the example app and the accessibility assertions re-enabled as part of #1809

rajsite:
Assigning to myself to file an issue upstream to axe-core

rajsite:
Created a minimal reproducing example that causes the lighthouse failure

[...]

Issue #280, opened 2022-01-28
Fix accessibility issues in Angular Example app


fredvisser:
# 🐛 Bug Report

We should fix all remaining accessibility issues in the Angular Example app, and change the Lighthouse budget to 100% required to pass

rajsite:
Removing the blocked tag as I think the current working decision is we own our select and combobox implementations and can improve accessibility ourselves. The score has also dropped lower and likely impacted by more than just those issues.

jattasNI:
Next step is to create issues for each accessibility problem and prioritize them individually. Then we should close this issue: we don't see much value in tracking all these issues together.

msmithNI:
Marked as team review since the select/combobox label change is now in, so we can discuss if there's further work to do here.

Lighthouse reports:
- [Before select/combobox built-in label change](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1722636864491-43521.report.html) : Accessibility = 85
- [After](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1722639228008-80394.report.html): Accessibility = 93

m-akinc:
Make sure remaining problems are tracked in separate issues, update the threshold back to 90 and then close this.

jattasNI:
Menu header issues are already tracked by #185

I added the table issues to #2285

I filed #2345 for the rich text issue

PR #1809, opened 2024-02-07
Workaround lighthouse build

rajsite:
# Pull Request

## 🤨 Rationale

Disables the Card in the Angular and Blazor example apps as it breaks axe-core tooling used by lighthouse, see: https://github.com/ni/nimble/issues/1650

## 👩‍💻 Implementation

Disabled the nimble-card in the Angular and Blazor apps

Re-enabled the accessibility lighthouse error. The thought is that with this change the accessibility score will be reliably calculated and not cause intermittent blocking of merges, only blocks on merges that further regress the score. If it is not the case and lighthouse becomes unreliable calculating the score we can disable it again.

Removed the exisiting comment for an issue to get us back to 0.9 score as there are multiple issues impacting our score now.

## 🧪 Testing

Relying on lighthouse CI execution.

## ✅ Checklist

### DIFF
--- a/angular-workspace/lighthouserc.js
+++ b/angular-workspace/lighthouserc.js
@@ -16,9 +16,7 @@ module.exports = {
         assert: {
             assertions: {
                 'categories:performance': ['warn', { minScore: 0.9 }],
-                // TODO: reset to 0.9 as part of https://github.com/ni/nimble/issues/1090
-                // TODO: reset to error as part of https://github.com/ni/nimble/issues/1650
-                'categories:accessibility': ['warn', { minScore: 0.88 }]
+                'categories:accessibility': ['error', { minScore: 0.8 }]
             }
         },
         upload: {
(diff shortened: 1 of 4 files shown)


rajsite:
@mollykreis Could you review this PR for re-enabling the lighthouse accessibility check (and if that seems like a good idea)

angular-workspace/lighthouserc.js
```diff
@@ -16,9 +16,7 @@ module.exports = {
         assert: {
             assertions: {
                 'categories:performance': ['warn', { minScore: 0.9 }],
-                // TODO: reset to 0.9 as part of https://github.com/ni/nimble/issues/1090
-                // TODO: reset to error as part of https://github.com/ni/nimble/issues/1650
-                'categories:accessibility': ['warn', { minScore: 0.88 }]
+                'categories:accessibility': ['error', { minScore: 0.8 }]
```
mollykreis, line 19: > Removed the exisiting comment for an issue to get us back to 0.9 score as there are multiple issues impacting our score now.

Do we want to bring this number back up eventually? If so, any ideas on how we should remember to do that?

mollykreis (approved):

angular-workspace/lighthouserc.js
rajsite: I guess that is covered in https://github.com/ni/nimble/issues/280 which states a more ambitious 1.0 goal

rajsite:
Bypassing owners (mostly OOO) and buddied with Molly

PR #1810, opened 2024-02-08
Lighthouse desktop profile

rajsite:
# Pull Request

## 🤨 Rationale

Switch lighthouse to using a desktop profile for all current pages.
The default low-powered mobile profile isn't the best for measuring the example angular app with all components listed (not a realistic use-case) or the benchmarking pages.
It would be nice to be performance optimized for mobile but the current performance scores aren't adding much value.

## 👩‍💻 Implementation

Switch lighthouse config to the desktop profile.
Address some of the easy lighthouse complaints on basic page structure.

## 🧪 Testing

Manually ran lighthouse on our published pages.

## ✅ Checklist

### DIFF
--- a/angular-workspace/lighthouserc.js
+++ b/angular-workspace/lighthouserc.js
@@ -16,7 +16,7 @@ module.exports = {
         },
         assert: {
             assertions: {
-                'categories:performance': ['warn', { minScore: 0.9 }],
+                'categories:performance': ['error', { minScore: 0.9 }],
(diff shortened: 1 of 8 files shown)


rajsite:
@fredvisser wanna take a look at this idea of switching to the desktop lighthouse profile?

msmithNI (approved):

rajsite:
@fredvisser gonna go with it, simple change if we don't like it

PR #1811, opened 2024-02-08
Allow a childless row to be expanded

mollykreis:
# Pull Request

## 🤨 Rationale

When a row has a state of `canLoadChildren`, clicking the expand button does not actually expand the row. The `row-expand-toggle` event is fired, but the row stays collapsed until children are provided in the data. This isn't ideal because the row state isn't consistent with what it should be given the user action and the details of the `row-expand-toggle` event. Therefore, this PR allows the childless row to be expanded. It also automatically collapses rows if they are still expandable when transitioning into a state where they have no children.

This is part of #1758, and this specific issue was discussed in [this PR thread](https://github.com/ni/nimble/pull/1794#discussion_r1481996663).

## 👩‍💻 Implementation

Refactor of the expansion manager to accomplish the following:
- Allow each row to be in an explicitly expanded, explicitly collapsed, or default state
    - Purpose: The previous code in the expansion manager only tracked two things: (1) whether every row was in a default state, and (2) which rows were collapsed. This made it difficult to manage the state when some rows defaulted to expanded and other to collapsed.
- Track which rows are expandable but have no children.
    - Purpose: We want to collapse rows that transition into have no children. Therefore, on a data update, we need to know whether or not a row previously had children.

## 🧪 Testing

- Wrote new unit tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-6653cda2-ff6f-489e-accf-68cec27fb146.json
+++ b/change/@ni-nimble-components-6653cda2-ff6f-489e-accf-68cec27fb146.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Allow a row with a state of `canLoadChildren` to be expanded when it does not have children",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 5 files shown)


mollykreis:
@atmgrifter00, will you buddy this for me?

mollykreis:
Closing this PR because I found an issue with it. I'll revisit later.

PR #1813, opened 2024-02-08
Add loadingChildren delayed hierarchy state

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1758. It adds support for showing a loading spinner on rows that are currently loading its children.

## 👩‍💻 Implementation

- Extend `TableRecordDelayedHierarchyState` to include `loadingChildren`
- Add `loading` attribute to the `nimble-table-row` that controls the visibility of the spinner
- Add a new label to the table-label-provider for the "Loading" string that is used to label the spinner

## 🧪 Testing

- Added new unit tests for the table and for the table row

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/components/row/styles.ts
+++ b/packages/nimble-components/src/table/components/row/styles.ts
@@ -5,6 +5,7 @@ import {
     applicationBackgroundColor,
     borderWidth,
     controlHeight,
+    controlSlimHeight,
     fillHoverColor,
     fillHoverSelectedColor,
     fillSelectedColor,
@@ -52,14 +53,26 @@ export const styles = css`
         background-color: ${fillHoverSelectedColor};
     }
 
+    .spinner-container,
     .expand-collapse-button {
+        flex: 0 0 auto;
         padding-left: calc(
             ${mediumPadding} + (var(--ni-private-table-row-indent-level) - 1) *
                 ${controlHeight}
         );
     }
 
+    .spinner-container {
+        width: ${controlSlimHeight};
+        height: ${controlSlimHeight};
+        align-self: center;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+    }
+
     .row-operations-container {
+        flex: 0 0 auto;
         display: flex;
     }
 
--- a/packages/nimble-components/src/table/components/row/template.ts
+++ b/packages/nimble-components/src/table/components/row/template.ts
@@ -9,11 +9,14 @@ import { checkboxTag } from '../../../checkbox';
 import {
     tableRowCollapseLabel,
     tableRowExpandLabel,
+    tableRowLoadingLabel,
     tableRowSelectLabel
 } from '../../../label-provider/table/label-tokens';
 import type { TableColumn } from '../../../table-column/base';
 import { buttonTag } from '../../../button';
 import { iconArrowExpanderRightTag } from '../../../icons/arrow-expander-right';
+import { spinnerTag } from '../../../spinner';
+import { SpinnerAppearance } from '../../../spinner/types';
 
 // prettier-ignore
 export const template = html<TableRow>`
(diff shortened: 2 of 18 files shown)


mollykreis:

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -40,17 +43,29 @@ export const template = html<TableRow>`
...
-                class="expand-collapse-button"
-                tabindex="-1"
-                @click="${(x, c) => x.onRowExpandToggle(c.event)}"
-                title="${x => (x.expanded ? tableRowCollapseLabel.getValueFor(x) : tableRowExpandLabel.getValueFor(x))}"
-                aria-hidden="true"
-            >
-                <${iconArrowExpanderRightTag} ${ref('expandIcon')} slot="start" class="expander-icon ${x => x.animationClass}"></${iconArrowExpanderRightTag}>
-            </${buttonTag}>
+            ${when(x => x.loading, html<TableRow>`
+                <span class="spinner-container">
```
atmgrifter00, line 47: What is the "container" element getting you?

atmgrifter00, line 47: I presume this container element is providing the same layout behaviors for the spinner element that the expander `button` element provides for its content? Basically I'm just fishing for why this container element is necessary.

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -40,17 +43,29 @@ export const template = html<TableRow>`
...
+                    <${spinnerTag}
+                        appearance="${SpinnerAppearance.accent}"
+                        aria-label="${x => tableRowLoadingLabel.getValueFor(x)}"
+                        title="${x => tableRowLoadingLabel.getValueFor(x)}"
```
atmgrifter00, line 51: I'm not seeing the tooltip on hover for the spinner. Is this expected?

mollykreis: No, that isn't expected. I see it on the published storybook for this PR. You don't?

packages/nimble-components/src/table/components/row/template.ts
mollykreis: Yes, your assumption is correct. I create the container so that I can size & position it exactly as the expander button. Then, I put the spinner centered in the container so that I know that the spinner will be positioned where the expand/collapse icon is within the button.

packages/nimble-components/src/table/components/row/template.ts
atmgrifter00: Well, in Firefox, if I go to the 'loadingChildren' option for the 'Delayed Hierarchy' storybook, and hover over the spinner, I get no tooltip. In Chrome it seems to work.

mollykreis: Interesting. From what I can tell, this is an existing issue in Firefox where the `title` on the spinner is very difficult to get to show up with the animation. If I force the animation to stop by applying the `--ni-private-spinner-animation-play-state:paused` CSS, then the tooltip shows up as expected.

Do you think we should do something to work around the issue?

atmgrifter00: I don't feel strongly about it.

atmgrifter00 (approved):

packages/nimble-components/src/table/components/row/template.ts
mollykreis: I'll leave the discussion open so other can weigh in.

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -52,14 +53,26 @@ export const styles = css`
...
     }
 
+    .spinner-container {
+        width: ${controlSlimHeight};
+        height: ${controlSlimHeight};
+        align-self: center;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+    }
```
rajsite, line 72: Can we avoid sharing styles in this way to [preserve dom order of selectors](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order)?
If we want to avoid duplication of that padding styling we can do so via private css properties.

packages/nimble-components/src/table/components/row/template.ts
rajsite: It's just the title breaking not the aria-label? If so I don't feel strongly about it.

mollykreis: From what I can tell, it is just the title. I think the mouse has to hover over the bits of the spinner in order for the title to be displayed, and it's pretty finicky to get that to work correctly when they keep moving.

rajsite: Oh, hmmm, that sounds less like a generic animations bug and more like something we should handle. Maybe we need an overlay div that is a pointer event target instead of the browser trying to target the moving divs. Still wouldn't hold up the PR but may be worth a tech debt issue. Definitely felt pretty clunky when I tried it

mollykreis: Created #1825

rajsite:
@jattasNI this is an incremental change on the current behavior. Think implementation is good, there were previous discussions on the best way to represent delayedLoading in storybook. If we want to change that can do in a follow-up PR but can discuss here in a thread after merge or in a tech debt issue.

PR #1817, opened 2024-02-12
Create a Web Worker Class 

Razvan1928:
# Pull Request

## ✅ Checklist


package.json
```diff
@@ -48,5 +48,8 @@
     "cross-env": "^7.0.3",
     "patch-package": "^6.4.7",
     "playwright": "1.40.0"
+  },
+  "dependencies": {
+    "comlink": "^4.4.1"
```
munteannatan, line 53: i think this can go into the nimble-components workspace

packages/nimble-components/.gitignore
```diff
@@ -1 +1,2 @@
 src/icons
+src/wafer-map/workers
```
munteannatan, line 2: you can also add this to prettierignore

packages/nimble-components/build/generate-workers/source/index.ts
```diff
@@ -0,0 +1,28 @@
...
+fs.mkdirSync(workersDirectory);
+console.log('Finished creating workers directory');
+
+const modulePath = require.resolve(moduleName);
+const sourceCode = fs.readFileSync(modulePath, 'utf-8');
+
+const componentFileContents =
+`/* eslint-disable no-useless-escape */
+export const workerCode = \`
+${sourceCode.replaceAll('`','\\`').replaceAll('$','\\$').replaceAll('exports.RenderWorker = RenderWorker;','')}
```
munteannatan, line 22: can you try to see if json parsing the string will make it exportable without the need to replace everything? see https://github.com/ni/nimble/pull/1788#discussion_r1471668308

packages/nimble-components/build/generate-workers/source/renderWorker.ts
```diff
@@ -0,0 +1,67 @@
...
+        // the highlight state of each die as a matrix row by row
+        dieHighlightsLayer: Int8Array;
+    } = {
+            dieColIndexArray: Int32Array.from([]),
+            rowLengthsArray: Int32Array.from([]),
+            dieRowIndexLayer: Int32Array.from([]),
+            dieValuesLayer: Int32Array.from([]),
+            dieHighlightsLayer: Int8Array.from([])
+        };
+    private verticalScale: { a: number, b: number } = { a: 0, b: 1 };
```
munteannatan, line 25: can you please remove the unused properties? we will add them as we develop the worker

packages/nimble-components/build/generate-workers/source/renderWorker.ts
```diff
@@ -0,0 +1,67 @@
+import * as Comlink from 'comlink';
+
+export class RenderWorker {
+    private canvas!: OffscreenCanvas;
+    private worker!: number;
+    private context!: OffscreenCanvasRenderingContext2D;
+    private dieMatrix: {
```
munteannatan, line 7: There is still some debate about the matrix API, can you please set it as a simple UInt8Array for testing purposes only? we will change it accordingly after the debate is settled

packages/nimble-components/src/wafer-map/tests/matrixRenderer.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { WaferMap } from '..';
+import { MatrixRenderer } from '../modules/matrixRenderer';
+
+describe('MatrixRenderer', () => {
+    it('worker should be healthy', () => {
+        const waferMap = new WaferMap();
```
munteannatan, line 6: I think you can use a WaferMapMock here, please see other tests

packages/nimble-components/src/wafer-map/modules/matrixRenderer.ts
```diff
@@ -0,0 +1,23 @@
...
+export class MatrixRenderer {
+    public readonly workerOne: Comlink.Remote<Worker>;
+
+    public constructor(private readonly wafermap: WaferMap) {
+        const blob = new Blob([workerCode], { type: 'text/javascript' });
+        const url = URL.createObjectURL(blob);
+        this.workerOne = Comlink.wrap(new Worker(url));
+    }
+
+    public isWorkerHealthy(): boolean {
```
munteannatan, line 17: maybe requesting a status from the worker using comlink is also a good idea

Razvan1928: I added a different health check logic, you can check it in render-worker

packages/nimble-components/src/wafer-map/tests/matrixRenderer.spec.ts
Razvan1928: I deleted this file as we discussed, I think this comment is not longer needed

packages/nimble-components/build/generate-workers/source/index.ts
Razvan1928: it is working using JSON.stringify

package.json
Razvan1928: indeed

packages/nimble-components/build/generate-workers/tests/render-worker.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { HealthStatus } from '../source/health-status';
+import { RenderWorker } from '../source/render-worker';
+
+describe('RenderWorker', () => {
+    it('worker should be healthy', () => {
+        const renderer = new RenderWorker();
+        expect(renderer.isWorkerHealthy()).toBe(HealthStatus.Healty);
```
munteannatan, line 7: can you check if the worker is healthy after wrapping it with comlink like you did in the renderer?

packages/nimble-components/build/generate-workers/tests/render-worker.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { HealthStatus } from '../source/health-status';
+import { RenderWorker } from '../source/render-worker';
+
+describe('RenderWorker', () => {
```
munteannatan, line 4: I am also not sure if these tests are run with the CI, can you check and see if you need to make a new test script?

Razvan1928: you are right, the test was not running, now it should be fine

packages/nimble-components/build/generate-workers/source/index.ts
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs');
+const path = require('path');
+
+const moduleName = '../bundle/render-worker.js';
+const healthStatusFileName = './health-status.js';
```
munteannatan, line 5: are you using this file somewhere?

packages/nimble-components/.gitignore
```diff
@@ -1 +1,3 @@
...
+src/build/generate-workers/build
```
munteannatan, line 3: is this folder generated?

Razvan1928: it wasn't, I deleted that line

packages/nimble-components/build/generate-workers/source/index.ts
Razvan1928: what do you mean? it is the file that moves the render-worker code into the wafer component

Razvan1928: oh, indeed healthStatusFileName is not used, I will delete it

packages/nimble-components/package.json
```diff
@@ -87,6 +91,7 @@
     "@types/d3-selection": "^3.0.0",
     "@types/d3-zoom": "^3.0.0",
     "@types/markdown-it": "^13.0.0",
+    "comlink": "^4.4.1",
```
munteannatan, line 94: Should comlink be both in here and devDependencies?

munteannatan (approved):

packages/nimble-components/package.json
Razvan1928: indeed, it is not needed

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
```diff
@@ -0,0 +1,17 @@
+import * as Comlink from 'comlink';
```
atmgrifter00, line 1: Is the folder meant to be called 'modules', or was it supposed to be 'models'? If you intended "modules" what's the reasoning for that name?

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { WaferMap } from '..';
+import { workerCode } from '../workers/render-worker';
+import type { RenderWorker } from '../../../build/generate-workers/dist/esm/render-worker';
```
atmgrifter00, line 4: This relative path seems suspicious. Why can't it just reference the type from its `source` folder location?

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,44 @@
...
+        data: Iterable<number>
+    ): void {
+        const start = this.performanceTest !== undefined ? self.performance.now() : undefined;
+        this.dieMatrix = Uint8Array.from(data);
+        if (this.performanceTest !== undefined) {
+            self.performance.measure(`${this.performanceTest} - worker:${this.worker} - renderDies`, { start });
+        }
+    }
+
+    public areMethodsCallable(): boolean {
```
atmgrifter00, line 27: I don't understand the purpose of this method. When would the tested condition ever be false?

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,44 @@
+import * as Comlink from 'comlink';
+import { HealthStatus, healthStatus } from './health-status';
+
+export class RenderWorker {
+    private worker!: number;
+    private dieMatrix: Uint8Array = Uint8Array.from([]);
+    private performanceTest: string | undefined;
```
atmgrifter00, line 7: When is this not `undefined`?

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
Razvan1928: When I try to use the path from source I get this error
import from source error
didn't get deep into that but my guess is that the source typescript file is outside of the nimble-components/src directory

packages/nimble-components/build/generate-workers/source/render-worker.ts
Razvan1928: The code from render-worker will be built, packed and later moved to a different location to be used. We wanted a way to check that the render code is reachable. (maybe it doesn't make sense indeed)
image
this is how the code looks in the end

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
Razvan1928: image
I don't have a strong point, it was there before, I assumed that files related with the wafermap should stay there. Should I put this file in a new directory src/wafer-map/models?

packages/nimble-components/build/generate-workers/source/render-worker.ts
Razvan1928: We wanted to provide some more accurate performance tests, but after talking with @munteannatan I removed this part

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
atmgrifter00: I think the folder should just be re-named 'models', but that could be done in a separate PR.

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
atmgrifter00: Oh I see. Sorry, I didn't recognize that at first.

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export class RenderWorker {
+    private worker!: number;
+    private dieMatrix: Uint8Array = Uint8Array.from([]);
+    private healthStatus: HealthStatus = healthStatus.unknown;
+
+    constructor() {
+    }
+
+    public emptyMatrix(): void {
```
atmgrifter00, line 12: We likely should have unit tests for each of these public methods.

packages/nimble-components/build/generate-workers/source/render-worker.ts
Razvan1928: I decided to delete all the healthcheck related code and rely on unit tests to ensure the code still works after bundling with rollup, now that I think of this it seems indeed a not so good idea

packages/nimble-components/build/generate-workers/source/tests/render-worker.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+        worker.emptyMatrix();
+        expect(worker.dieMatrix.length).toEqual(0);
+    });
+
+    it('updateMatrix should update the dieMatrix correctly', () => {
+        const testData: Iterable<number> = [4, 5, 6];
+        worker.updateMatrix(testData);
+        expect(worker.dieMatrix).toEqual(Uint8Array.from(testData));
+    });
+});
```
jattasNI, line 21: Are there any tests you can add to inspect `Comlink`'s state and verify these lines?

```ts
const worker = new RenderWorker();
Comlink.expose(worker);
```

packages/nimble-components/src/wafer-map/tests/matrix-renderer.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { WaferMap } from '..';
+import { MatrixRenderer } from '../modules/matrix-renderer';
+
+/*
+ * Sometimes methods from render worker become bugged after being built, bundled and copied to a different location
+ * with these tests we ensure they behave as expected after this process
+ */
```
jattasNI, line 7: I don't think this comment adds much value so I'd suggest deleting it. Tests generally serve to validate build and execution so describing ways they can fail seems redundant to me.

Also, it sounds like the build process is buggy? Can we get to the bottom of why that is? Is that going to cause problems on the pipeline or for developers running local builds? We want our infrastructure to be extremely reliable.

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
jattasNI, line 1: Instead of importing `*` is it possible to import only the methods we need?

jattasNI (changes requested):

packages/nimble-components/build/generate-workers/rollup.config.js
```diff
@@ -0,0 +1,13 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
```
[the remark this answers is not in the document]
jattasNI: @rajsite and @atmgrifter00 and I talked offline. We think the right fix for this confusion would be to move the web worker code and build to a separate package that nimble-components depends on. This would also help solve some other concerns like:
1. guaranteeing Nimble and clients are all using the same version of comlink
2. avoiding possible circular dependencies between the component and the web worker code
3. clearer dependencies between build and source which prevents issues like linting/testing/storybook not working because code is in unexpected locations

However, setting up a new package is a non-trivial amount of work and requires a little bit of Nimble repo expertise. @rajsite volunteered to do this but it would be done in a follow up PR. We'd let this PR go in to unblock follow up work but expect it to be refactored shortly after. In the meantime you should still address the feedback above and elsewhere in the PR to clean things up as best we can before adding a new package.

Seem like a good plan?

packages/nimble-components/package.json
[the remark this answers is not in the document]
Razvan1928: Indeed I had an older version of node, 16.x, updated it, deleted package-lock.json ran npm i, now it should be fine, I think? 
image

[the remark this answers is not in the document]
Razvan1928: added
image

change/@ni-nimble-components-46e491a1-b49a-4d5c-b64e-6eb6fae50e0b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Create web worker class for future faster rendering wafer map",
```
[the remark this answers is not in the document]
Razvan1928: indeed, thanks :)

packages/nimble-components/src/wafer-map/tests/matrix-renderer.spec.ts
Razvan1928: Indeed the comment was not very useful, I deleted it. 
By 'render worker become bugged after being built...' I wanted to say that making changes to render-worker.ts or rollup config can break the code without being noticed, @munteannatan had some incidents, but the potential of breaking a code portion while developing is expected and should be prevented with unit tests, now that I thought a second time my comment didn't make much sense. Beside that the pipeline is stable and doesn't fail without a reason intermittently, from that perspective we should not worry.

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,18 @@
+import * as Comlink from 'comlink';
+
+export class RenderWorker {
```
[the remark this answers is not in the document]
Razvan1928: sure 
image

packages/nimble-components/build/generate-workers/rollup.config.js
```diff
@@ -0,0 +1,13 @@
...
+import path from 'path';
```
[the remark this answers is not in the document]
Razvan1928: did it :)

packages/nimble-components/build/generate-workers/rollup.config.js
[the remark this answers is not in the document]
Razvan1928: I wrote a short README.md inside generate-workers, tried to make it similar to the other generate-* directories READMEs. Regarding the second comment, yeah, I understand and it's good to hear that our workers infrastructure will receive a more professional approach😁

packages/nimble-components/build/generate-workers/source/tests/render-worker.spec.ts
jattasNI: Discussed offline and there isn't an easy way to test via a unit test but this is covered by integration tests. Closing.

packages/nimble-components/build/generate-workers/source/render-worker.ts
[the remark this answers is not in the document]
jattasNI: Sorry, I meant adding a JSDoc comment on the class explaining its purpose. Like this but with real content 😃 

The markdown docs are good and worth keeping too.

```suggestion
/**
 * <A comment explaining where this code runs and what it does>
 */
export class RenderWorker {
```

packages/nimble-components/package.json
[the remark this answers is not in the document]
jattasNI: There were fewer diffs but still more than I expected. I pushed a change that fixes it now.

packages/nimble-components/build/generate-workers/README.md
```diff
@@ -0,0 +1,16 @@
+# Generate Workers
+
```
jattasNI, line 2:
```suggestion

This directory contains wafer map rendering code and build scripts which compile it to run in a web worker.

```

packages/nimble-components/build/generate-workers/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Behavior
+
+- the source directory and index.ts are built using tsc
+- source/render-worker.ts is bundled with rollup
+- index.ts has the purpose to prepare and move dist/bundle/render-worker to src/wafer-map/worker
+- src/wafer-map/worker/render-worker will be used to create a `Blob` object which is then used to create a `URL` for a web worker.
```
jattasNI, line 8: nit: more code formatting

```suggestion
- the `source` directory and `index.ts` are built using `tsc`
- `source/render-worker.ts` is bundled with rollup
- `index.ts` has the purpose to prepare and move `dist/bundle/render-worker` to `src/wafer-map/worker`
- `src/wafer-map/worker/render-worker` will be used to create a `Blob` object which is then used to create a `URL` for a web worker.
```

packages/nimble-components/build/generate-workers/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const renderModuleName: string = '../bundle/render-worker.js';
+const workersDirectory: string = path.resolve('./src/wafer-map/workers');
+
+prepareDirectory(workersDirectory);
+
+const modulePath: string = resolveModulePath(renderModuleName);
+const sourceCode: string = fs.readFileSync(modulePath, 'utf-8');
+
+const fileContent: string = `// eslint-disable-next-line no-template-curly-in-string\nexport const workerCode = ${JSON.stringify(sourceCode)};`;
```
jattasNI, line 33: nit: rather than having to read escape codes like `\n` it's more readable just to use newlines, which the template literals should support.
```suggestion
const fileContent: string = `// eslint-disable-next-line no-template-curly-in-string
export const workerCode = ${JSON.stringify(sourceCode)};`;
```

jattasNI (approved):

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,25 @@
...
+        this.dieMatrix = Uint8Array.from([]);;
+    }
+
+    public updateMatrix(
+        data: Iterable<number>
+    ): void {
+        this.dieMatrix = Uint8Array.from(data);
+    }
+}
+const worker = new RenderWorker();
```
rajsite, line 24: Instead of exposing an instance of the class, lets expose the class itself.
See: https://github.com/GoogleChromeLabs/comlink/blob/dffe9050f63b1b39f30213adeb1dd4b9ed7d2594/tests/same_window.comlink.test.js#L195

packages/nimble-components/build/generate-workers/source/tests/render-worker.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { RenderWorker } from '../render-worker';
+
+describe('RenderWorker', () => {
+    let worker: RenderWorker;
+
+    beforeEach(() => {
+        worker = new RenderWorker();
```
rajsite, line 7: Instead of testing this directly, I think a more representative test would be to test this in process via MessageChannel.

See these comlink tests:

https://github.com/GoogleChromeLabs/comlink/blob/dffe9050f63b1b39f30213adeb1dd4b9ed7d2594/tests/same_window.comlink.test.js#L69-L75

https://github.com/GoogleChromeLabs/comlink/blob/dffe9050f63b1b39f30213adeb1dd4b9ed7d2594/tests/same_window.comlink.test.js#L195-L200

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Responsible for drawing the dies inside the wafer map, adding dieText and scaling the canvas
+ */
+export class MatrixRenderer {
+    public readonly workerOne: Remote<RenderWorker>;
+
+    public constructor(private readonly wafermap: WaferMap) {
+        const blob = new Blob([workerCode], { type: 'text/javascript' });
+        const url = URL.createObjectURL(blob);
```
rajsite, line 14: Need to make sure to clean-up the objectUrl after the worker is made.
Either that that or lets just make it once lazily instead of every construction. (that might be better)

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
```diff
@@ -0,0 +1,17 @@
...
+        this.workerOne = wrap<RenderWorker>(new Worker(url));
+    }
+}
```
rajsite, line 17: Instead of having this wrapper around the RenderWorker instance, I think with the above changes we can expose the RenderWorker class more directly instead of having to access the "workerOne" property.

```ts
export createMatrixRenderer = async (): Remote<RenderWorker> {
    const blob = new Blob([workerCode], { type: 'text/javascript' });
    const url = URL.createObjectURL(blob);
    const RemoteMatrixRenderer = wrap<RenderWorker>(new Worker(url));
    const instance = await new RemoteMatrixRenderer();
    return instance;
}

```

packages/nimble-components/build/generate-workers/source/render-worker.ts
```diff
@@ -0,0 +1,25 @@
+import { expose } from 'comlink';
+
+/**
+ * RenderWorker class is meant to be used within a Web Worker context, 
+ * using Comlink to facilitate communication between the main thread and the worker. 
+ * The RenderWorker class manages a matrix of dies, once an instance of RenderWorker is created, 
+ * it is exposed to the main thread using Comlink's `expose` method.
+ * This setup is used in the wafer-map component to perform heavy computational duties
+ */
+export class RenderWorker {
```
rajsite, line 10: Based on the comment below, the MatrixRenderer abstraction can be removed, so we can instead just call this MatrixRenderer

packages/nimble-components/package.json
rajsite, line 94: Let's pin the version to prevent the chance that the main thread and the pre-built library get out of sync.

And don't forget to run npm install to get the package-lock update after

```suggestion
    "comlink": "4.4.1",
```

packages/nimble-components/src/utilities/tests/setup.ts
```diff
@@ -10,3 +10,4 @@ require('./setup-configuration.js');
 
 // all browser test scripts
 importAll(require.context('../../', true, /\.spec\.js$/));
+importAll(require.context('../../../../build', true, /\.spec\.js$/));
```
rajsite, line 13: Instead of the whole build dir can you point to this specific context:

```suggestion
importAll(require.context('../../../../build/generate-workers/dist/esm', true, /\.spec\.js$/));
```

That way both these paths are search dist/esm directories.

packages/nimble-components/build/generate-workers/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+        "importHelpers": true,
+        "resolveJsonModule": true,
+        "sourceMap": true,
+        "lib": ["DOM", "ES2015", "ES2016.Array.Include", "ES2017.Object"],
+        "types": ["jasmine", "webpack-env"],
+        "importsNotUsedAsValues": "error",
+        "noImplicitOverride": true,
+        "noUncheckedIndexedAccess": true,
+        "noPropertyAccessFromIndexSignature": false
+    },
```
rajsite, line 21: can you instead extend the nimble-components tsconfig?
```suggestion
    "extends": "../../tsconfig.json"
```

packages/nimble-components/build/generate-workers/source/tests/render-worker.spec.ts
rajsite: You need to give more feedback if not following the comment when resolving if you don't want to get stuck in review forever.

I think your unit tests here should be with MessageChannel (that's why I commented here) and your integration tests with WebWorker (now there are none). 

Why was that suggestion ignored?

packages/nimble-components/build/generate-workers/source/render-worker.ts
Razvan1928: did it 
image

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
Razvan1928: good point, did it
image

packages/nimble-components/build/generate-workers/source/tests/render-worker.spec.ts
Razvan1928: Now they should be fine (I hope) 
image

Razvan1928: Hope it's fine now
image

packages/nimble-components/build/generate-workers/source/render-worker.ts
Razvan1928: I removed the MatrixRenderer class and renamed RenderWorker into MatrixRenderer

change/@ni-nimble-components-46e491a1-b49a-4d5c-b64e-6eb6fae50e0b.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @Razvan1928 one thing I want to mention is that it does seem like your current workflow is to push each individual commit remote at a time. I just want to point out that each commit pushed remote does a PR build so this PR as an example has run over 100 builds.

It is different from Skyline where you have to explicit queue a build. It is not necessarily a problem and if it useful it is fine. But if you felt that you could create multiple commits locally and push them together so fewer builds are created and that does not impact your workflow that could be a small tweak to reduce the number of builds

image

packages/nimble-components/build/generate-workers/source/tests/matrix-renderer.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { Remote, expose, wrap } from 'comlink';
+import { MatrixRenderer } from '../matrix-renderer';
+
+describe('MatrixRenderer with MessageChannel', () => {
+    let wrappedWorker: Remote<MatrixRenderer>;
```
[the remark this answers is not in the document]
Razvan1928: renamed
image

packages/nimble-components/build/generate-workers/source/tests/matrix-renderer.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    beforeEach(async () => {
+        const { port1, port2 } = new MessageChannel();
+        const worker = new MatrixRenderer();
+        expose(worker, port1);
+        wrappedWorker = await wrap<MatrixRenderer>(port2);
+    });
+
+    it('updateMatrix should update the dieMatrix', async () => {
+        const testData: Iterable<number> = [4, 5, 6];
```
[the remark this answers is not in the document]
Razvan1928: Replaced all 3 instances, it wasn't needed
image
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,28 @@
+import type { Remote } from 'comlink';
+import { createMatrixRenderer } from '../modules/worker-renderer';
+import type { MatrixRenderer } from '../../../build/generate-workers/dist/esm/source/matrix-renderer';
+
+describe('MatrixRenderer worker:', () => {
+    let remoteWorker: Remote<MatrixRenderer>;
```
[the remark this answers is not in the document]
Razvan1928: renamed
image

packages/nimble-components/src/wafer-map/modules/worker-renderer.ts
```diff
@@ -0,0 +1,19 @@
...
+Remote<MatrixRenderer>
+> => {
+    const blob = new Blob([workerCode], { type: 'text/javascript' });
+    const url = URL.createObjectURL(blob);
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const RemoteMatrixRenderer = wrap<new() => MatrixRenderer>(
+        new Worker(url));
+    const instance = await new RemoteMatrixRenderer();
+    return instance;
+};
```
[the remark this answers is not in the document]
Razvan1928: implemented the feedback
image

packages/nimble-components/src/wafer-map/modules/worker-renderer.ts
```diff
@@ -0,0 +1,19 @@
...
+export const createMatrixRenderer = async (): Promise<
+Remote<MatrixRenderer>
+> => {
+    const blob = new Blob([workerCode], { type: 'text/javascript' });
+    const url = URL.createObjectURL(blob);
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const RemoteMatrixRenderer = wrap<new() => MatrixRenderer>(
+        new Worker(url));
+    const instance = await new RemoteMatrixRenderer();
+    return instance;
```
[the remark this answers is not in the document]
Razvan1928: implemented the feedback
image
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The MatrixRenderer class manages a matrix of dies, once an instance of MatrixRenderer is created, 
+ * it is exposed to the main thread using Comlink's `expose` method.
+ * This setup is used in the wafer-map component to perform heavy computational duties
+ */
+export class MatrixRenderer {
+    public worker!: number;
+    public dieMatrix: Uint8Array = Uint8Array.from([]);
+
+    public constructor() {
+    }
```
rajsite, line 15: Shouldn't need an empty constructor

```suggestion
```

packages/nimble-components/src/wafer-map/modules/worker-renderer.ts
```diff
@@ -0,0 +1,28 @@
+import { wrap, Remote } from 'comlink';
```
rajsite, line 1: Lets have this file name align with the api it is creating.

`create-matrix-renderer.ts`

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
+import type { Remote } from 'comlink';
```
rajsite, line 1: Lets update the file `create-matrix-renderer.spec.ts`

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    beforeEach(async () => {
+        const result = await createMatrixRenderer();
+        matrixRenderer = result.matrixRenderer;
+        terminate = result.terminate;
+    });
+
+    afterEach(() => {
+        terminate();
+        matrixRenderer = undefined;
```
rajsite, line 17: Lie to the compiler in one spot instead of checking in every spot.

```suggestion
        matrixRenderer = undefined!;
```

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
+import type { Remote } from 'comlink';
+import { createMatrixRenderer } from '../modules/worker-renderer';
+import type { MatrixRenderer } from '../../../build/generate-workers/dist/esm/source/matrix-renderer';
+
+describe('MatrixRenderer worker:', () => {
+    let matrixRenderer: Remote<MatrixRenderer> | undefined;
```
rajsite, line 6:
```suggestion
    let matrixRenderer: Remote<MatrixRenderer>;
```

rajsite (commented):
Very close, just some minor consistency and naming things left

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
+import type { Remote } from 'comlink';
+import { createMatrixRenderer } from '../modules/worker-renderer';
+import type { MatrixRenderer } from '../../../build/generate-workers/dist/esm/source/matrix-renderer';
+
+describe('MatrixRenderer worker:', () => {
```
rajsite, line 5:
```suggestion
describe('MatrixRenderer worker', () => {
```

change/@ni-nimble-components-46e491a1-b49a-4d5c-b64e-6eb6fae50e0b.json
rajsite, line 1: fyi @Razvan1928 I was just reset on the PR but I have already put up new comments, I'm not sure if that was intentional

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
Razvan1928: deleted
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
Razvan1928: removed it
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
Razvan1928: so evil, did it without hesitation 
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
Razvan1928: updated!
image

packages/nimble-components/src/wafer-map/modules/worker-renderer.ts
Razvan1928: good point
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: I think I solved some error with it in the past, indeed now it's useless, deleted it

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * MatrixRenderer class is meant to be used within a Web Worker context, 
+ * using Comlink to facilitate communication between the main thread and the worker. 
+ * The MatrixRenderer class manages a matrix of dies, once an instance of MatrixRenderer is created, 
+ * it is exposed to the main thread using Comlink's `expose` method.
+ * This setup is used in the wafer-map component to perform heavy computational duties
+ */
+export class MatrixRenderer {
+    public worker!: number;
```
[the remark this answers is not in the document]
Razvan1928: removed it!
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    afterEach(() => {
+        terminate();
+        matrixRenderer = undefined;
+    });
+
+    it('updateMatrix should update the dieMatrix', async () => {
+        const testData = [4, 5, 6];
+        if (matrixRenderer === undefined) {
+            throw new Error('matrixRenderer is undefined');
+        }
```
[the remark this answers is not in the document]
Razvan1928: removed it
image

packages/nimble-components/src/wafer-map/tests/worker-renderer.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+        expect(Array.from(resolvedDieMatrix)).toEqual(
+            Array.from(Uint8Array.from(testData))
+        );
+    });
+
+    it('emptyMatrix should empty the dieMatrix', async () => {
+        const testData = [4, 5, 6];
+        if (matrixRenderer === undefined) {
+            throw new Error('matrixRenderer is undefined');
+        }
```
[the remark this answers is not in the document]
Razvan1928: removed it
image

rajsite (approved):

rajsite:
@DStavilaNI and @zszilagy are still codeowners. Make sure to work with them for what needs to be reviewed

PR #1818, opened 2024-02-13
Rich text editor | Update dependency on Tiptap

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

1. Update all the Tiptap extensions to the latest version and fixes the build error from #1771
2. [Bug 2651659](https://dev.azure.com/ni/DevCentral/_workitems/edit/2651659): Comments acceptance tests failing in latest nimble/sl-comments-lib update
The linked tests might fail because of the package update with the [deprecated method](https://github.com/ueberdosis/tiptap/blob/3e440c2f96cd56683d3d7eb3c0e6cea83f7a31b7/packages/extension-mention/src/mention.ts#L8C3-L9C86) used here for the mention-extesion. Here is the error from the console log: `ERROR TypeError: this.options.renderLabel is not a function
    at Object.renderHTML (create-tiptap-editor.ts:213:30)`. 

## 👩‍💻 Implementation

1. Bump all the Tiptap versions to the latest `2.2.2`
2. Replaced the deprecated `renderLabel` method with the recommended `renderText` method for mention support.

## 🧪 Testing

Manually verified the mention node rendering and other basic flows like adding an @mention and setting it via setMarkdown etc.,

## ✅ Checklist


packages/nimble-components/src/rich-text/editor/models/create-tiptap-editor.ts
```diff
@@ -210,7 +210,8 @@ function createCustomMentionExtension(
             return [
                 config.viewElement,
                 mergeAttributes(this.options.HTMLAttributes, HTMLAttributes),
-                this.options.renderLabel({
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
m-akinc, line 213: Why is there a suppression here?

vivinkrishna-ni: It was a miss. My VS code shows the type error but once reloaded, it doesn't. I have removed the comment.

m-akinc (approved):

rajsite (approved):

PR #1826, opened 2024-02-14
Add publishConfig key to package files

m-akinc:
# Pull Request

## 🤨 Rationale

For consistency, setting each package's npm `publishConfig` as public access. This is necessary for the initial creation of the package, so if we use any of these as a starting point for future packages, it's useful to include. It also makes the intent explicit.

## 👩‍💻 Implementation

Added `publishConfig` setting to every package in Nimble repo, except for `jasmine-parameterized`, which already had it.

## 🧪 Testing
N/A

## ✅ Checklist


packages/nimble-blazor/package.json
```diff
@@ -28,6 +28,9 @@
     "type": "git",
     "url": "git+https://github.com/ni/nimble.git"
   },
+  "publishConfig": {
+    "access": "public"
+  },
```
rajsite, line 33: this package is weird, we don't publish it (it's not a published npm package, it's a published nuget package) but we don't mark it private so beachball will rev it. notice how the invoke-publish:npm ignores the command above.
So we shouldn't mark this as publicjConfig access public

PR #1834, opened 2024-02-16
Update spec with wafer map API decision 

munteannatan:
# Pull Request

## 🤨 Rationale

Prototyped Apache Arrow API and recorded performance.

## 👩‍💻 Implementation

Updated the spec file. 
Prototype can be found here https://github.com/ni/nimble/pull/1788

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/wafer-map/specs/features/rendering.md
+++ b/packages/nimble-components/src/wafer-map/specs/features/rendering.md
@@ -82,6 +82,46 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
 
 ### Data Structure and Interface
 
+The best solution to solve teh API of the wafermap is to use both of the proposed methods.
+
+The Public API will be the following:
+
+```TS
+public diesTable: Table<{
+        colIndex: Int32,
+        rowIndex: Int32,
+        value: Float32,
+        tags: Uint32;
+        metadata: never;
+    }>
+```
+
+This will be the [Apache Arrow](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html) table schema.
+The row and column indices will be `Int32` columns, the values will be `Float32` columns.
+The tags for each die will be represented as a 32 bit mask stored in a `Uint32` column.
+The metadata column will be stored in an wildcard typed column.
+
+This approach has the benefits of a row based format that aligns well with the existing public api, a nice public API to use and the ease of future improvements.
+
+The limits for this approach are the following:
+
+1. There seems to be no support for columns of lists of strings. We decided to overcome this using a bit mask of tags. Another possible solution can be a dynamic number of rows for storing tags, but the performance may suffer.
+1. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). The possible solutions for this are searching by iterating over the whole table, which is not feasible (see 3.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/).The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies.
+1. The transfer method for arrow tables is cumbersome, we would have to use another higher level library [geoarrow](https://github.com/geoarrow/geoarrow-js/blob/main/src/worker/transferable.ts). Fortunately we can skip over this problem by not transferring the tables.
+1. The iteration over stored rows is very bad compared to typed arrays as seen in the table below. This impacts the goals we set for this rendering improvement. The solution to this issue and the transferring issue is splitting the relevant columns from the table (rows, columns, values, tags mask) and messaging them to the worker separately. This can be done with a very small overhead using the [getChild](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html#getChild) method and calling [toArray](https://arrow.apache.org/docs/js/classes/Arrow_dom.Vector.html#toArray) on the resulting vector. After being transferred, The buffers can be cached to speed up value access and filtering.
+
+| name                      | duration (ms) [1]  | duration (ms) [2]  | detail                                                                                                    |
+| ------------------------- | ------------------ | ------------------ | --------------------------------------------------------------------------------------------------------- |
+| typed iterate             | 7.551699995994568  | 6.052600026130676  | iterating over two 1M typed arrays and calculating the sums                                               |
+| typed from table iterate  | 6.4953999519348145 | 5.136900067329407  | iterating over two 1M typed arrays from Table columns and calculating the sums (time includes conversion) |
+| vector iterate            | 76.4708000421524   | 66.58230006694794  | iterating over two 1M Vectors and calculating the sums                                                    |
+| table get() iterate       | 1350.0404000282288 | 1030.582899928093  | iterating over the 1M Table using `table.get(rowIndex)` and calculating the sums                          |
+| table [iterator] iterate  | 1091.6706000566483 | 1011.069100022316  | iterating over the 1M Table using the [iterator] and calculating the sums                                 |
+| array from table iterate  | 943.0076999664307  | 980.0875999927521  | iterating over the 1M Table after converting `toArray()` and calculating the sums                         |
+| vector from table iterate | 965.2465000152588  | 1012.9023000001907 | iterating over the 1M Vector after converting the Table with `makeVector()` and calculating the sums      |
+
+#### Previously proposed solutions
+
 We have two possible solutions for representing the data in the memory. They will be decided with a spec update. The fist one is an in-house solution:
 
 ```TS


packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,6 +82,48 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
 
 ### Data Structure and Interface
 
+The best solution to solve the API of the wafermap is to use both of the proposed methods.
```
atmgrifter00, line 85: Can you provide a brief summary of what "both of the proposed methods" is referring to? Being in a new section it's unclear at a glance what the referenced methods are.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,6 +82,48 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+        metadata: never;
+    }>
+```
+
+This will be the [Apache Arrow](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html) table schema.
+The row and column indices will be `Int32` columns, the values will be `Float32` columns.
+The tags for each die will be represented as a 32 bit mask stored in a `Uint32` column.
+The metadata column will be stored in an wildcard typed column.
+
+This approach has the benefits of a row based format that aligns well with the existing public api, a nice public API to use and the ease of future improvements.
```
atmgrifter00, line 104:
```suggestion
This approach has the benefits of a row based format that aligns well with the existing public API, as well as a nice public API that easily allows future improvements.
```
I'm not exactly sure if this rewording is what you were trying to get across, but I think it reads better.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,6 +82,48 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+
+The limits for this approach are the following:
+
+1. There seems to be no support for columns of lists of strings. We decided to overcome this using a bit mask of tags. Another possible solution can be a dynamic number of columns for storing tags, but the performance may suffer.
+1. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). The possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/).The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies.
```
atmgrifter00, line 109: While the end-result renders as expected, can you update the raw bulleted list to be "1.", "2.", "3.", etc...?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,6 +82,48 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+| table get() iterate       | 1350.0404000282288 | 1030.582899928093  | iterating over the 1M Table using `table.get(rowIndex)` and calculating the sums                          |
+| table [iterator] iterate  | 1091.6706000566483 | 1011.069100022316  | iterating over the 1M Table using the [iterator] and calculating the sums                                 |
+| array from table iterate  | 943.0076999664307  | 980.0875999927521  | iterating over the 1M Table after converting `toArray()` and calculating the sums                         |
+| vector from table iterate | 965.2465000152588  | 1012.9023000001907 | iterating over the 1M Vector after converting the Table with `makeVector()` and calculating the sums      |
+
+The memory impact is not very significant, amounting to 74.01MB for 1M dies compared with 44.65MB for the previously prototyped API.
+
+#### Previously proposed solutions
+
 We have two possible solutions for representing the data in the memory. They will be decided with a spec update. The fist one is an in-house solution:
```
atmgrifter00, line 127: Does this section of the spec need to be updated? The wording in the new content added suggests that you have settled on the Apache Arrow API over the TypedArray API, so it would seem that _this_ is the spec update that settles this decision.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,7 +82,49 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+| table get() iterate       | 1350.0404000282288 | 1030.582899928093  | iterating over the 1M Table using `table.get(rowIndex)` and calculating the sums                          |
+| table [iterator] iterate  | 1091.6706000566483 | 1011.069100022316  | iterating over the 1M Table using the [iterator] and calculating the sums                                 |
+| array from table iterate  | 943.0076999664307  | 980.0875999927521  | iterating over the 1M Table after converting `toArray()` and calculating the sums                         |
+| vector from table iterate | 965.2465000152588  | 1012.9023000001907 | iterating over the 1M Vector after converting the Table with `makeVector()` and calculating the sums      |
+
+The memory impact is not very significant, amounting to 74.01MB for 1M dies compared with 44.65MB for the previously prototyped API.
+
+#### Previously proposed solutions
+
+These are the two possible solutions identified for representing the data in the memory. The fist one is an in-house solution:
```
atmgrifter00, line 128: Unless I'm mistaken we can heavily edit the below section. First, we can rename 'Previously proposed solutions' to 'Alternative solutions'. Then, I would think the only solution we would need to call out is the in-house TypedArray API. The Apache Arrow API _is_ the API of choice, and thus doesn't need to be mentioned here. Finally you can remove the blurb at the bottom of the section that described how you will ultimately decide on the solution.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,7 +82,49 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
 
 ### Data Structure and Interface
 
-We have two possible solutions for representing the data in the memory. They will be decided with a spec update. The fist one is an in-house solution:
+The best solution to solve the API of the wafermap is to use parts of both of the proposed methods mentioned in the next section. This means using Apache Arrow as the wafer component API and Typed Arrays for their iterating performance and transferability to worker threads.
+
+The Public API will be the following:
+
+```TS
+public diesTable: Table<{
```
jattasNI, line 90: To make it even more clear where the `Table` type is coming from you could add something like this (feel free to change the import path if it's not right):
```suggestion
import { Table } from 'apache-arrow';

public diesTable: Table<{
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,7 +82,49 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+The tags for each die will be represented as a 32 bit mask stored in a `Uint32` column.
+The metadata column will be stored in an wildcard typed column.
+
+This approach has the benefits of a row based format that aligns well with the existing public API, as well as a nice public API that easily allows future improvements.
+
+The limits for this approach are the following:
+
+1. There seems to be no support for columns of lists of strings. We decided to overcome this using a bit mask of tags. Another possible solution can be a dynamic number of columns for storing tags, but the performance may suffer.
+2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). The possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/).The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies.
```
jattasNI, line 109: nit: fomratting
```suggestion
2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). The possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/). The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,7 +82,49 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+3. The transfer method for arrow tables is cumbersome, we would have to use another higher level library [geoarrow](https://github.com/geoarrow/geoarrow-js/blob/main/src/worker/transferable.ts). Fortunately we can skip over this problem by not transferring the tables.
```
jattasNI, line 110: Could you clarify where the "transfer" is from and to in this sentence? From the main thread to the web worker?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,7 +82,49 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+4. The iteration over stored rows is very bad compared to typed arrays as seen in the table below. This impacts the goals we set for this rendering improvement. The solution to this issue and the transferring issue is splitting the relevant columns from the table (rows, columns, values, tags mask) and messaging them to the worker separately. This can be done with a very small overhead using the [getChild](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html#getChild) method and calling [toArray](https://arrow.apache.org/docs/js/classes/Arrow_dom.Vector.html#toArray) on the resulting vector. After being transferred, The buffers can be cached to speed up value access and filtering.
```
jattasNI, line 111: nit: "bad" is vague and could refer to performance / API / correctness / anything else :-). "Slow" (or some other word) would be more precise.

```suggestion
4. The iteration over stored rows is very slow compared to typed arrays as seen in the table below. This impacts the goals we set for this rendering improvement. The solution to this issue and the transferring issue is splitting the relevant columns from the table (rows, columns, values, tags mask) and messaging them to the worker separately. This can be done with a very small overhead using the [getChild](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html#getChild) method and calling [toArray](https://arrow.apache.org/docs/js/classes/Arrow_dom.Vector.html#toArray) on the resulting vector. After being transferred, The buffers can be cached to speed up value access and filtering.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
jattasNI, line 109: Can you clarify what wafer map features require searching or filtering the table? Is it something that happens on every data update or just when certain attributes change?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +82,50 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
 
 ### Data Structure and Interface
 
-We have two possible solutions for representing the data in the memory. They will be decided with a spec update. The fist one is an in-house solution:
+The best solution to solve the API of the wafermap is to use Apache Arrow as the wafer component API, and Typed Arrays as teh worker API for their iterating performance and transferability to worker threads.
```
jattasNI, line 85: typo 
```suggestion
The best solution to solve the API of the wafermap is to use Apache Arrow as the wafer component API, and Typed Arrays as the worker API for their iterating performance and transferability to worker threads.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +82,50 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
-
--   A row based format that aligns well with the existing public api
--   Well supported and tested format
--   Nice public API to use, we don't have to invent a new format, just document our schema for the arrow tables
--   Designed for large dataset visualizations
+1. There seems to be no support for columns of lists of strings. We decided to overcome this using a bit mask of tags. Another possible solution can be a dynamic number of columns for storing tags, but the performance may suffer.
+2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). The possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/). Searching for dies based on their position is crucial for highlighting and sending the highlighted die metadata with the `die-hover` event. The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies.
```
jattasNI, line 114: nit: reordered to flow better by presenting the use case before the decision 

```suggestion
2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). Searching for dies based on their position is crucial for highlighting and sending the highlighted die metadata with the `die-hover` event. The solution we chose is using a custom method for finding rows based on column and row indexes cached as typed arrays. This method provides faster access to row values and metadata and does not induce additional dependencies. Other possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using a higher level library such as [aquero](https://uwdata.github.io/arquero/). 
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +82,50 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
 
-In order to choose from these alternatives we will prototype and check:
+| name                      | duration (ms) [1]  | duration (ms) [2]  | detail                                                                                                    |
+| ------------------------- | ------------------ | ------------------ | --------------------------------------------------------------------------------------------------------- |
+| typed iterate             | 7.551699995994568  | 6.052600026130676  | iterating over two 1M typed arrays and calculating the sums                                               |
```
jattasNI, line 120: We don't need to show these values with so many digits of precision. It makes it hard to scan them and provides a false sense of how accurate they are. Let's just show whole ms, not fractions of ms.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +82,50 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+3. The transfer method between the main an worker thread for arrow tables is cumbersome, we would have to use another higher level library [geoarrow](https://github.com/geoarrow/geoarrow-js/blob/main/src/worker/transferable.ts). Fortunately we can skip over this problem by not transferring the tables to the worker.
+4. The iteration over stored rows is very slow compared to typed arrays as seen in the table below. This impacts the goals we set for this rendering improvement. The solution to this issue and the transferring issue is splitting the relevant columns from the table (rows, columns, values, tags mask) and messaging them to the worker separately. This can be done with a very small overhead using the [getChild](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html#getChild) method and calling [toArray](https://arrow.apache.org/docs/js/classes/Arrow_dom.Vector.html#toArray) on the resulting vector. After being transferred, The buffers can be cached to speed up value access and filtering.
```
jattasNI, line 116: I'm having trouble separating the proposed solution from the alternatives. I'd suggest:
1. Update this section to describe the proposal and remove any discussion of pros/cons, alternatives, or timing. 
2. Move all of that discussion to the Alternatives section. Put each alternative in a separate sub-section with a clear title like "Use typed arrays instead of Apache Arrow for public API" or "Use untyped arrays instead of typed arrays for transfer to worker". Then describe the pros and cons in each of those subsections.
3. Update the timing table to use the same naming as these subsections (or give them numbers / abbreviations if the names are too long to fit in the table)

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +82,50 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
-The alternative to the above mentioned data structure is an [apache arrow](https://arrow.apache.org/docs/js/index.html) table with columns and metadata.
+The limits for this approach are the following:
 
-Pros of using Apache Arrow:
-
--   A row based format that aligns well with the existing public api
--   Well supported and tested format
--   Nice public API to use, we don't have to invent a new format, just document our schema for the arrow tables
--   Designed for large dataset visualizations
+1. There seems to be no support for columns of lists of strings. We decided to overcome this using a bit mask of tags. Another possible solution can be a dynamic number of columns for storing tags, but the performance may suffer.
```
jattasNI, line 113: Could you go into more detail on how the bit mask of tags will impact clients? Will they be disappointed to lose the ability to specify arbitrary strings for tags? Or to specify more than 32 tags? Do we need to provide a library of functions to translate between an array of tag strings and a corresponding bit mask?

change/@ni-nimble-components-c0f72f9f-b879-43ad-bc5b-ee8b74e2700f.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Have some questions @atmgrifter00 will bring up in the morning but I can join later to discuss if needed:
- In the Apache Arrow format does the metadata need to be included?
  
  It does not seem relevant to the render process based on discussions of the other table format. Also the apache arrow format cannot represent any JS value, i.e. `metadata: unknown[]` could be a class, function, dom element, etc.
  
  Should we use a format like `{table: Table<>, metadata?: unknown[]}` so any JS value is allowed?
  Should we only allow metadata to be possible arrow values? (breaking change)
  Should we force a type on metadata? (i.e. string, braking change)
- In the section on geoarrow and arquero my takeaway from the description was that the usage was "cumbersome" and we want to "avoid dependencies". What I want to understand is if those tools help make more maintainable code over time with good performance. It is not clear to me that those are behaviors of what was evaluated in the spec.
  - What I want to know if we have a clear evaluation of is if the high-level filter and query features of arquero can be used to manipulate data efficiently while also letting us use higher-level patterns for writing algorithms cleanly.
  - I created a stackblitz that sets up both libraries together that can be used to experiment with more easily: https://stackblitz.com/edit/geoarrow-worker-arquero-demo?file=src%2Fmain.ts&terminal=dev

packages/nimble-components/src/wafer-map/specs/features/rendering.md
munteannatan: After discussing with @rajsite and @atmgrifter00 we decided to try using dynamic columns for the table and highlights which can be joined and filtered using [arquero](https://uwdata.github.io/arquero/) in place of the bitmasks

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -210,8 +211,69 @@ The current expectation is for a singular wafer component to be displayed on the
...
+    // a 32 bitset array of tags for each die; aligns more closely to the existing public api but limits users to 32 tags.
+    dieHighlightsLayer: Int32Array;
+    // metadata array for each die; it will not be sent to the worker
+    metadata : unknown[]
+}
+```
+
+Using TypedArrays has the benefit of direct transfer to web workers without structured cloning of the object by transferring the arrayBuffers and reconstructing the object. Other benefits of typedArrays include the low access time when iterating over the values, more memory efficiency and faster direct access to metadata layers values. The previous inputs can be adapted to this new structure to maintain backwards compatibility.
+
+This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
```
jattasNI, line 238: The document describes how we'd handle Angular/Blazor integration for this TypedArrays alternative but not for the recommended Apache Arrow approach. Could you add an equivalent paragraph to the section above?

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,48 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
 
-This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
+-   The row and column indices will be `Int32` columns
+-   The values will be a `Float32` column.
 
-The alternative to the above mentioned data structure is an [apache arrow](https://arrow.apache.org/docs/js/index.html) table with columns and metadata.
+If there are more columns needed to store metadata or other values the schema will be extensible.
 
-Pros of using Apache Arrow:
+The `highlightedTable` will contain rows partially filled with values which will be used to filter the `diesTable` and enable highlighting.
```
jattasNI, line 113: What would the `value` field mean in the `highlightedTable` data structure? 
1. Is it ignored, with any die that appears in `highlightedTable` being highlighted?
2. Is it treated as a boolean, where dies with value 0 are unhighighlighted and value >0 are highlighted?
3. Does it have some other meaning, like each value getting a different highlight color or being included in metadata?

If it's 1 or 2 we could consider changing its type to capture this more accurately. That might mean making the class generic on two types, `TDieValues` and `THighlightValues` or something like that.

We should document this in more detail here.

jattasNI (commented):
Thanks for the document updates, the proposal is much more clear now. Just a couple more details to nail down but I'm ok with the general direction.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,47 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+They will be checked at runtime and a `WaferMapValidity` flag will be raised signaling an `invalidTableInput`.
 
-Pros of using Apache Arrow:
+If there are more columns needed to store metadata or other values the schema will be extensible. This will induce a breaking change in the API, as the metadata which was previously `unknown` will have to be recorded in table using the supported column types.
 
--   A row based format that aligns well with the existing public api
--   Well supported and tested format
--   Nice public API to use, we don't have to invent a new format, just document our schema for the arrow tables
--   Designed for large dataset visualizations
+The `highlightedTable` will have the same columns, but they will contain rows only partially filled with values, which will be used to filter the `diesTable` and enable highlighting. The values which are not empty on each individual row, including `colIndex`, `rowIndex`, `value` and others will be used to filter the table as an `AND` operation. Multiple rows will be used as filters with the `OR` operation.
```
atmgrifter00, line 109: There are a few concerns that @jattasNI, @rajsite and myself worked through that we feel the HLD should address:
1. While this may seem extreme, it seems beneficial to remove the metadata concept altogether from the proposed APIs. If we understand things correctly, the metadata is simply used to attach to the hover events that the client can subscribe to, with its original design allowing clients to attach arbitrary data (which they could then consume in an informed way...which the wafer-map is incapable of). Why can't the hover event simply return index information which the client can then use to associate against some map they track on the client side? This removes the odd typing concerns from the Nimble API completely, and the seemingly arbitrary added complexity that Nimble ultimately has to maintain.
2. The proposed API for the `highlightedTable` seems like it might be overly flexible, that while making some use-cases easier to accomplish (such as highlighting entire rows/columns), accomplishing things like highlighting ranges of values doesn't seem as straightforward (but still highly desirable). In order to unblock this HLD, it seems prudent to move this API/feature to an open question that will be addressed later.

@rajsite and @jattasNI, please add additional thoughts as needed.

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,47 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+They will be checked at runtime and a `WaferMapValidity` flag will be raised signaling an `invalidTableInput`.
 
-Pros of using Apache Arrow:
+The schema will be extensible. This will induce a breaking change in the API, as the metadata which was previously `unknown` will not recorded in table, but the hover event will reference an index which can e used by the client to select the metadata outside the component.
 
--   A row based format that aligns well with the existing public api
--   Well supported and tested format
--   Nice public API to use, we don't have to invent a new format, just document our schema for the arrow tables
--   Designed for large dataset visualizations
+The `highlightedTable` will have the same columns, but they will contain rows only partially filled with values, which will be used to filter the `diesTable` and enable highlighting. The values which are not empty on each individual row, including `colIndex`, `rowIndex`, `value` and others will be used to filter the table as an `AND` operation. Multiple rows will be used as filters with the `OR` operation. More details regarding highlights will be discussed in an open issue.
```
atmgrifter00, line 109: As I mentioned in the [Teams chat](https://teams.microsoft.com/l/message/19:meeting_OTJkYmU1M2QtN2Q2YS00M2E5LTgzODctMzEyZTkxNTA0ZDFi@thread.v2/1709071000119?context=%7B%22contextType%22%3A%22chat%22%7D), it appears that the Apache Arrow Table structure supports list (array) types. This is suggestive that we could support the current API (`string[]`) for `highlightedTags`, which we feel would be preferable, barring some other noted reason why the API change is beneficial.

With that said, this determination can happen at a later time, if it is worthwhile to put this decision on the backburner, to unblock the remainder of the work outlined here. If you do decide to make this an open issue, then I believe what is proper is to remove all mention of the `highlightedTable` from this section and the above "Public API" description.

@jattasNI @rajsite

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,44 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
 }
 ```
 
-Using TypedArrays has the benefit of direct transfer to web workers without structured cloning of the object by transferring the arrayBuffers and reconstructing the object. Other benefits of typedArrays include the low access time when iterating over the values, more memory efficiency and faster direct access to metadata layers values. The previous inputs can be adapted to this new structure to maintain backwards compatibility.
+It will be using the [Apache Arrow Table](https://arrow.apache.org/docs/js/classes/Arrow_dom.Table.html).
+It will require at least three columns for the `diesTable`:
 
-This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
+-   The row and column indices will be `Int32` columns
+-   The values will be a `Float32` column.
```
rajsite, line 102: Are these the names being proposed? If so lets be specific.

```suggestion
-   The `rowIndex` and `colIndex` will be `Int32` columns
-   The `value` will be a `Float32` column.
```

rajsite, line 102: [JavaScript numbers are Float64](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding) (i.e doubles).

If there is not a specific reason for the discrepancy we should align it with the JS equivalent api and make it Float64

If there is a specific reason it should be documented

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,44 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+```
 
-In order to choose from these alternatives we will prototype and check:
+When filtering the highlighted dies and searching for their index we will use [arquero](https://uwdata.github.io/arquero/) to perform joins and other operations involving the tables.
 
--   Does it have comparable memory performance
--   Does it perform well or have significant overhead
--   Is it easy to divide and use in parallel
+The [JavaScript implementation of Apache Arrow](https://arrow.apache.org/docs/js/index.html) provides TypeScript Types which will work in Angular applications.
+The [C# implementation of Apache Arrow](https://github.com/apache/arrow/blob/main/csharp/README.md) is also providing support for reading Arrow IPC streams which can be used to convert inputs from Balzor.
```
rajsite, line 121:
```suggestion
The [C# implementation of Apache Arrow](https://github.com/apache/arrow/blob/main/csharp/README.md) is also providing support for reading Arrow IPC streams which can be used to convert inputs from Blazor.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -210,8 +207,69 @@ The current expectation is for a singular wafer component to be displayed on the
...
+-   Nice public API to use, we don't have to invent a new format, just document our schema for the arrow tables
+-   Designed for large dataset visualizations
+
 Another option is to break each object property as a separate attribute for the wafer map component. This can also lead to increased complexity and confusion for the user which will need to pass several structured objects instead of a singular object.
 
+#### Alternative Iteration and Filtering with Apache Arrow Table
+
+The limits for the apache arrow table approach are the following:
+
+1. There seems to be no support for columns of lists of strings.
```
rajsite, line 249:
```suggestion
1. Apache arrow documentation shows [list types are supported](https://arrow.apache.org/docs/status.html) but more research is needed to understand their usage, if they are useful in Tables, and their performance if used.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -210,8 +207,69 @@ The current expectation is for a singular wafer component to be displayed on the
...
+2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). Searching for dies based on their position is crucial for highlighting and sending the highlighted die metadata with the `die-hover` event.
```
rajsite, line 250:
```suggestion
2. There is no support currently for [searching or filtering the table](https://github.com/apache/arrow/issues/13233). Searching for dies based on their position is crucial for highlighting and sending the highlighted die metadata with the `die-hover` event. More research is needed to see if alternative libraries can be used for filtering / data analysis. ([POC with arquero](https://stackblitz.com/edit/geoarrow-worker-arquero-demo?file=src%2Fmain.ts)).
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -210,8 +207,69 @@ The current expectation is for a singular wafer component to be displayed on the
...
+3. The transfer method between the main an worker thread for arrow tables is cumbersome.
```
rajsite, line 251:
```suggestion
3. Apache arrow does not yet have first class support for efficiently transferring arrow data structures in workers. When asked they said they are [supportive of adding the APIs](https://github.com/apache/arrow/issues/39017#issuecomment-1955653556).
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -210,8 +207,69 @@ The current expectation is for a singular wafer component to be displayed on the
...
+4. The iteration over stored rows is very slow compared to typed arrays as seen in the table below. This impacts the goals we set for this rendering improvement.
+
+Alternatives for solving these problems are the following:
+
+1. A dynamic number of columns for storing tags, but the performance may suffer.
+2. Possible solutions for this are searching by iterating over the whole table, which is not feasible (see 4.) or using typed arrays and caching to speed up the search for the relevant columns.
+3. The use of a higher level library [geoarrow](https://github.com/geoarrow/geoarrow-js/blob/main/src/worker/transferable.ts)
```
rajsite, line 258:
```suggestion
3. The use of a higher level library [geoarrow](https://github.com/geoarrow/geoarrow-js/blob/main/src/worker/transferable.ts). ([POC](https://stackblitz.com/edit/geoarrow-worker-arquero-demo?file=src%2Fmain.ts)).
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -228,13 +286,86 @@ We may also implement an external queue canceling functionality.
...
+| 2       | 1        | 3        | 67.93000030517578  | a        | null      |
+| 3       | 2        | 1        | 79.04000091552734  | b        | null      |
+| 4       | 2        | 0        | 26.489999771118164 | c        | null      |
+| 5       | 3        | 3        | 20.829999923706055 | c        | null      |
+
+The filter matched the rows with the same values from the highlight table. This can be used for tags when filtering value ranges, values themselves, column and row indexes or other types of supported data types.
+
+The details of the implementation and more refined filtering will be discussed.
+
+Anther option is using the existing `highlightedTags` API with a `List` column in the table [(listed as supported)](https://arrow.apache.org/docs/status.html).
```
rajsite, line 346:
```suggestion
Another option is using the existing `highlightedTags` API with a `List` column in the table [(listed as supported)](https://arrow.apache.org/docs/status.html).
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -228,13 +286,86 @@ We may also implement an external queue canceling functionality.
...
+
```
rajsite, line 347:
```suggestion
Specific open questions:
- For a highlightedTags table api:
  - Should the API be constrained in the [supported arrow types](https://arrow.apache.org/docs/status.html)?
    - Should just primitive types be supported?
    - Should just types supported in the JavaScript and C# languages be supported?
    - Should just types well-supported in all exisiting implementation be supported?
    - Would certain types, ex. like strings, lead to poor performance and be discouraged?
    - Are there implementation challenges transferring buffers of arbitrary types across the Web Worker boundary?
  - Should the API be constrained in the set of columns that participate in highlighting instead of all columns? Maybe columns with a specific name prefix like `highlighted_`?
  - Is there real known benefits for specifying per floating point `value`? Or specific `rowIndex` / `columnIndex` independently?
  - Does the highlightTable need to contain all columns used in diesTable? Can it just be a subset of columns?
- For a tags columns api:
  - Do columns of List<string> work in tables?
  - Do dictionary columns work in tables to improve efficiency compared to Liist<string>?
  - What is the performance of a List<string> / Dictionary column api compared to the alternatives?
```

rajsite (approved):

rajsite:
Don't need to wait for @mollykreis, but FYI @jattasNI this is ready for a final review

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,44 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
 
-This API will have [optimized byte-array interop from Blazor](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/byte-array-interop) and should be supported by Angular as a [vanilla javascript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer).
+-   The `rowIndex` and `colIndex` will be `Int32` columns
+-   The `value` will be a `Float64` column.
 
-The alternative to the above mentioned data structure is an [apache arrow](https://arrow.apache.org/docs/js/index.html) table with columns and metadata.
+They will be checked at runtime and a `WaferMapValidity` flag will be raised signaling an `invalidTableInput`.
 
-Pros of using Apache Arrow:
+The schema will be extensible. This will induce a breaking change in the API, as the metadata which was previously `unknown` will not recorded in table, but the hover event will reference an index which can e used by the client to select the metadata outside the component.
```
jattasNI, line 106:
```suggestion
The schema will be extensible. This will induce a breaking change in the API, as the metadata which was previously `unknown` will not recorded in table, but the hover event will reference an index which can be used by the client to select the metadata outside the component.
```

packages/nimble-components/src/wafer-map/specs/features/rendering.md
```diff
@@ -82,46 +81,44 @@ The POC is found in this branch [Worker Rendering POC](https://github.com/ni/nim
...
+We are going to split the columns relevant to rendering from the table (rows, columns, values) and transfer them to the worker separately. This can be done with a very small overhead using the method below on the resulting vector. After being transferred, the buffers can be cached to speed up value access and filtering.
+
+```TS
+    const colIndex: Int32Array = diesTable.getChild('colIndex').toArray();
+    const rowIndex: Int32Array = diesTable.getChild('rowIndex').toArray();
+    ...
+```
 
-In order to choose from these alternatives we will prototype and check:
+When filtering the highlighted dies and searching for their index we will use [arquero](https://uwdata.github.io/arquero/) to perform joins and other operations involving the tables.
```
jattasNI, line 118: Should this sentence move to the Alternatives section since it's talking about highlighting? (Or be deleted since you already mention arquero there?) (Or updated to mention non-highlighting use cases?)

jattasNI (approved):

PR #1837, opened 2024-02-19
Fix issue with underline-hidden on the anchor column

mollykreis:
# Pull Request

## 🤨 Rationale

This fixes an issue I noticed with the anchor column where the `underline-hidden` configuration on the anchor column didn't always work as expected. This is because the `underline-hidden` styling is based on the presence of the `underline-hidden` attribute, but the cell-view didn't use `?` when configuring `underline-hidden` on the underlying `nimble-anchor` element.

## 👩‍💻 Implementation

Add `?` in the anchor column's cell-view template when setting `underline-hidden` on the underlying `nimble-anchor`.

## 🧪 Testing

- Wrote new unit test that fails without this change and passes with it

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table-column/anchor/cell-view/template.ts
+++ b/packages/nimble-components/src/table-column/anchor/cell-view/template.ts
@@ -26,7 +26,7 @@ export const template = html<TableColumnAnchorCellView>`
                 target="${x => x.columnConfig?.target}"
                 type="${x => x.columnConfig?.type}"
                 download="${x => x.columnConfig?.download}"
-                underline-hidden="${x => x.columnConfig?.underlineHidden}"
+                ?underline-hidden="${x => x.columnConfig?.underlineHidden}"
                 appearance="${x => x.columnConfig?.appearance}"
                 title=${x => (x.hasOverflow ? x.text : null)}
             >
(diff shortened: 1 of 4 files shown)


packages/nimble-components/src/table-column/anchor/cell-view/template.ts
```diff
@@ -26,7 +26,7 @@ export const template = html<TableColumnAnchorCellView>`
                 target="${x => x.columnConfig?.target}"
                 type="${x => x.columnConfig?.type}"
                 download="${x => x.columnConfig?.download}"
-                underline-hidden="${x => x.columnConfig?.underlineHidden}"
+                ?underline-hidden="${x => x.columnConfig?.underlineHidden}"
```
rajsite, line 29: Looks like the matrix tests did the same thing, 

https://github.com/ni/nimble/blob/9f63cc8bb43b3c28fe1ccdec774e05693de1e885/packages/nimble-components/src/table-column/anchor/tests/table-column-anchor-matrix.stories.ts#L66

Could you update the binding and make the matrix test use true|false instead of true|undefined?

rajsite (approved):

rajsite:
Bypassing @jattasNI storybook looks good, status check just didn't make it through to the PR, but is unchanged.

PR #1841, opened 2024-02-20
New Wafer Component API using Apache Arrow

munteannatan:
# Pull Request

## 🤨 Rationale

Details about the API, implementation and other changes can be found in the spec file https://github.com/ni/nimble/blob/main/packages/nimble-components/src/wafer-map/specs/features/rendering.md.

Other requirements are under review https://github.com/ni/nimble/pull/1834.

## 👩‍💻 Implementation

Created the new API property and changed the update cycle to use two rendering strategies.
Created storybook controls and data.

## 🧪 Testing

unit testing created for the API access and update.

## ✅ Checklist


packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -85,6 +87,41 @@ describe('WaferMap', () => {
...
+            expect(spy).toHaveBeenCalledTimes(1);
+        });
+
+        it('will have matrix render strategy after diesTable change', () => {
+            element.diesTable = new Table();
+            processUpdates();
+            expect(element.renderStrategy).toEqual('matrix');
+        });
+
+        xit('will trigger render-complete after diesTable change', async () => {
```
munteannatan, line 108: @atmgrifter00 this test is failing waiting on the promise. I implemented this the same way that the toggle event was tested. Is there something else I should do?

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -125,6 +184,7 @@ const metadata: Meta<WaferMapArgs> = {
             grid-max-y=${x => x.gridMaxY}
             :colorScale="${x => x.colorScale}"
             :dies="${x => getDiesSet(x.dies, wafermapDieSets)}"
+            :diesTable="${x => getDiesTableSet(x.diesTable, wafermapDiesTableSets)}"
```
munteannatan, line 187: @atmgrifter00 setting the `diesTable` control does not change the `dies` control. How should I treat this use case? Should I hide the storybook controls until the feature is finished?

atmgrifter00: I would presume these should be mutually exclusive, yes? If so, I think you should control which API you use through a Storybook control (like a switch). You can call the storybook option something like "die API", but it's just a boolean property that determines which property you bind to in your template code. You would then have a separate storybook property that represents the data set. This may be the option you call "dies", but it just presents the choice of data sets to use.

packages/nimble-components/src/wafer-map/modules/matrix-renderer.ts
```diff
@@ -0,0 +1,55 @@
+import type { WaferMap } from '..';
+import { HoverDieOpacity } from '../types';
+
+/**
+ * Responsible for drawing the dies inside the wafer map, adding dieText and scaling the canvas
+ */
+export class MatrixRenderer {
```
atmgrifter00, line 7: What's the idea behind the term "Matrix" here? Nothing on the API surface seems directly related to a matrix.

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
atmgrifter00: I'll have to pull down your branch and play with it (which I haven't been able to do yet).

munteannatan: I removed the event and made a spy instead

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
munteannatan: I made a radio control for api versions

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -36,7 +38,7 @@ export class WaferMap extends FoundationElement {
      * @internal
      * needs to be initialized before the properties trigger changes
      */
-    public readonly waferMapUpdateTracker = new WaferMapUpdateTracker(this);
+    public readonly waferMapUpdateTracker: WaferMapUpdateTracker = new WaferMapUpdateTracker(this);
```
atmgrifter00, line 41: This change isn't needed right? You should only have to supply a type if it isn't set to a default.

packages/nimble-components/src/wafer-map/tests/wafer-map-validator.spec.ts
```diff
@@ -78,6 +79,89 @@ describe('Wafermap Validator module', () => {
         });
 
         it('should not be valid', () => {
+            expect(waferMapValidator.getValidity()).toEqual({
```
atmgrifter00, line 82: I'll just note that it seems a little strange to have so many describe blocks that just have a single test in them. I see that the whole test file is written this way, so I'm fine with being consistent, I just don't understand why it's in its current form.

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -33,8 +34,10 @@ describe('WaferMap', () => {
 
     describe('update flow', () => {
         let spy: jasmine.Spy;
+        let workerRendererSpy: jasmine.Spy;
         beforeEach(() => {
             spy = spyOn(element, 'update');
```
atmgrifter00, line 39: Could you rename this to `updateSpy`?

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -85,6 +88,30 @@ describe('WaferMap', () => {
...
+            expect(spy).toHaveBeenCalledTimes(1);
+        });
+
+        it('will use `worker` render strategy after diesTable change', () => {
+            element.diesTable = new Table();
+            processUpdates();
+            expect(element.renderStrategy).toEqual('worker');
+        });
+
+        xit('will call drawWafer after diesTable change', () => {
```
atmgrifter00, line 109: Does this still need a resolution?

packages/nimble-components/src/wafer-map/tests/wafer-map-validator.spec.ts
munteannatan: I think there were many types of inputs to be validated, but the outputs are just is valid or not for this class

atmgrifter00: So, just to make sure there is no confusion with my first remark: I would think most of these tests in this file could be of the form (to use the test this comment is attached to as a specific example):

```
it('with impossible grid dimension wafer map should not be valid', () => {
   const waferMock = getWaferMapMockValidator(1, -1, 1, -1);
   waferMapValidator = new WaferMapValidator(waferMock as WaferMap);
   waferMapValidator.validateGridDimensions();
   
   expect(waferMapValidator.getValidity()).toEqual({
        invalidGridDimensions: true,
        invalidDiesTableSchema: false
    });
    expect(waferMapValidator.isValid()).toBeFalse();
}
````
Now, if there were another test to put under the `with impossible grid dimension` umbrella, this idea would obviously not be good, but the pattern throughout this file suggests that it's intentional to have a single test per `describe` block, which is unconventional.

File: packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts [hunk 8]
```diff
@@ -106,6 +125,56 @@ describe('WaferMap', () => {
         });
     });
 
+    describe('worker renderer draw flow', () => {
+        let spy: jasmine.Spy;
+        beforeEach(() => {
+            spy = spyOn(element.workerRenderer, 'drawWafer');
```
atmgrifter00, line 131: Conventionally, we will name the `spy` local to be related to the actual member being spied on. So, here, it would be best named `drawWaferSpy`. This makes its usage in the test more readable.

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -106,6 +125,56 @@ describe('WaferMap', () => {
...
+            processUpdates();
+            expect(element.validity.invalidDiesTableSchema).toBeTrue();
+            expect(spy).toHaveBeenCalledTimes(0);
+        });
+    });
+
+    describe('worker renderer flow', () => {
+        let spy: jasmine.Spy;
+        beforeEach(() => {
+            spy = spyOn(element.workerRenderer, 'renderHover');
```
atmgrifter00, line 156: Same remark as above.

atmgrifter00 (approved):

change/@ni-nimble-components-9a663b6a-ecb5-4698-bd2c-77cb3281255a.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: The PR description can include a link to the spec but should also be self contained.
Please update the PR description to describe what is implemented in this specific PR. (i.e. added the diesTable, added the WorkerRenderer)

packages/nimble-components/package.json
```diff
@@ -84,13 +85,13 @@
...
     "@tiptap/extension-placeholder": "^2.2.2",
     "@tiptap/extension-text": "^2.2.2",
-    "@tiptap/extension-hard-break": "^2.2.2",
     "@types/d3-array": "^3.0.4",
     "@types/d3-random": "^3.0.1",
     "@types/d3-scale": "^4.0.2",
     "@types/d3-selection": "^3.0.0",
     "@types/d3-zoom": "^3.0.0",
     "@types/markdown-it": "^13.0.0",
+    "apache-arrow": "^15.0.0",
```
rajsite, line 94: This should be in peerDependencies since users will be expected to use the library directly and construct objects based on it to pass to us. This will be our first peer dependency.

packages/nimble-components/src/wafer-map/tests/wafer-map-validator.spec.ts
munteannatan: ok, I fixed them

change/@ni-nimble-components-9a663b6a-ecb5-4698-bd2c-77cb3281255a.json
rajsite, line 1: FYI I started review with individual comments accidentally but will bundle the rest of the comments in a review.

packages/nimble-components/package.json
jattasNI: We should probably change [renovate.json](https://github.com/ni/nimble/blob/136a79bc77cf5722f1d06201bd5fa7afe9164bf0/.github/renovate.json#L20) to automatically update `peerDependencies` in addition to the types its already updating.

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -42,7 +47,6 @@ interface WaferMapArgs {
     gridMaxY: number | undefined;
     dieHover: unknown;
     validity: WaferMapValidity;
```
jattasNI, line 49: Reminder to document the new validity flag in storybook alongside the existing one.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -272,6 +298,15 @@ export class WaferMap extends FoundationElement {
 
     private diesChanged(): void {
         this.waferMapUpdateTracker.track('dies');
+        this.renderStrategy = 'main';
+        this.waferMapUpdateTracker.queueUpdate();
+    }
+
+    private diesTableChanged(): void {
+        this.waferMapUpdateTracker.track('dies');
+        if (this.diesTable !== undefined) {
```
jattasNI, line 307: Why is the `worker` strategy set conditionally but the `main` strategy set unconditionally? Will the component be in a bad state if it tries to queue an update where someone set `diesTable = undefined` but never set `dies`? Or if they set them both to valid values?

I'm not saying we need to gracefully handle switching back and forth between renderers but I do want to avoid an in-between state where some data is set for each renderer and it's unclear which one should be used.

packages/nimble-components/src/wafer-map/tests/data-generator.ts
```diff
@@ -18,6 +19,19 @@ const generateStringValue = (
     }
     return valueToString(value);
 };
+const generateFloatValue = (
```
jattasNI, line 22:
```suggestion

const generateFloatValue = (
```

packages/nimble-components/src/wafer-map/tests/wafer-map.stories.ts
```diff
@@ -239,7 +280,46 @@ const metadata: Meta<WaferMapArgs> = {
                     set4: 'Many dies are highlighted'
                 }
             },
-            defaultValue: 'set1'
+            defaultValue: 'set1',
+            if: { arg: 'apiVersion', eq: 'stable' }
```
jattasNI, line 284: The ability to hide parameters based on the value of another one is neat. But I think it's a bit subtle. I'd suggest:
1. adding a note to both `dies` and `diesTable` saying which API they're part of
2. moving the apiVersion parameter to be first in the list so it's obvious it controls everything about the component
3. giving the apiVersion parameter a display name like "API Version" so it's obvious it isn't a property of the component
4. updating the description of apiVersion to mention what controls it in code. Something like: "the component will use the stable API if data is provided via `dies` and the experimental API if data is provided via `diesTable`"

change/@ni-nimble-components-9a663b6a-ecb5-4698-bd2c-77cb3281255a.json
rajsite: Sorry I did not finish the review today but expect to tomorrow

packages/nimble-components/src/wafer-map/index.ts
munteannatan: I made both conditional. I think when both are set the latest one will be evaluated when the update queue gets popped. Right now the worker render strategy will try and calculate based on the empty dies api, but that will cahnge in the future updates. If the dies are going to be populated afterwards, the calculations will be redone based on the changes. Is there a better strategy to follow? I think we will adapt this along as we keep developing the new rendering

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -272,6 +298,13 @@ export class WaferMap extends FoundationElement {
...
         this.waferMapUpdateTracker.track('dies');
+        this.renderStrategy = this.dies === undefined ? 'worker' : 'main';
+        this.waferMapUpdateTracker.queueUpdate();
+    }
+
+    private diesTableChanged(): void {
+        this.waferMapUpdateTracker.track('dies');
+        this.renderStrategy = this.diesTable === undefined ? 'main' : 'worker';
         this.waferMapUpdateTracker.queueUpdate();
     }
```
rajsite, line 309: This seems kind of racy. Property binding should behave as declaratively as possible. It's weird that assigning these in different orders will result in different behavior. Can we instead say that if diesTable is defined use that for both of these?

```ts
if (this.diesTable) {
    return 'worker';
}
return 'main';
```

Also is having the string representation actually useful?

Seems like an unnecessary extra abstraction. Can we just do

```ts
    public readonly mainRenderer = new RenderingModule(this);
    public readonly workerRenderer = new WorkerRenderer(this);

    @observable
    public renderer = this.mainRenderer;
```

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
```diff
@@ -40,4 +42,25 @@ export class WaferMapValidator {
...
         return !this.invalidGridDimensions;
     }
+
+    public validateDiesTableSchema(): boolean {
+        this.invalidDiesTableSchema = false;
+        if (this.wafermap.diesTable === undefined) {
+            this.invalidDiesTableSchema = false;
+        } else if (
+            this.wafermap.diesTable.numCols < 3
+            || this.wafermap.diesTable.schema.fields.findIndex(
```
rajsite, line 52: We should be checking the Type of the fields as well to align with what was in spec

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -272,6 +298,13 @@ export class WaferMap extends FoundationElement {
 
     private diesChanged(): void {
         this.waferMapUpdateTracker.track('dies');
+        this.renderStrategy = this.dies === undefined ? 'worker' : 'main';
+        this.waferMapUpdateTracker.queueUpdate();
+    }
+
+    private diesTableChanged(): void {
+        this.waferMapUpdateTracker.track('dies');
```
jattasNI, line 306: Should this be `this.waferMapUpdateTracker.track('diesTable');`?

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -53,7 +53,7 @@ export class Prerendering {
             this.wafermap.colorScaleMode
         );
         this._diesRenderInfo = this.wafermap.dies
-            .map(die => this.computeDieRenderInfo(die))
+            ?.map(die => this.computeDieRenderInfo(die))
             .filter(info => info !== null) as DieRenderInfo[];
     }
```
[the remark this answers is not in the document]
munteannatan: the issue seems to be that the type after the map resolves to (DieRenderInfo | null)[], but this does not disappear after the filter.

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -106,6 +125,56 @@ describe('WaferMap', () => {
...
+
+        it('will not call drawWafer after unsupported diesTable change', () => {
+            element.diesTable = new Table();
+            processUpdates();
+            expect(element.validity.invalidDiesTableSchema).toBeTrue();
+            expect(drawWaferSpy).toHaveBeenCalledTimes(0);
+        });
+    });
+
+    describe('worker renderer flow', () => {
```
jattasNI, line 153: Looks like there's a test failure to fix

packages/nimble-components/src/wafer-map/index.ts
munteannatan: I was aiming to replace the dies step in the update cycle, so I thought that duplicating it will not be necessary

jattasNI (approved):

packages/nimble-components/src/wafer-map/modules/prerendering.ts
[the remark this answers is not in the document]
rajsite: > but this does not disappear after the filter.

You need to use a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) as the filter function for TS to see it. I'll pushed a change showing that.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -138,7 +147,9 @@ export class WaferMap extends FoundationElement {
     @observable public hoverDie: WaferMapDie | undefined;
 
     @observable public highlightedTags: string[] = [];
-    @observable public dies: WaferMapDie[] = [];
+    @observable public dies: WaferMapDie[] | undefined;
```
rajsite, line 150: We can leave this initialized to empty array now right? the condition is that diesTable is assigned to switch the renderer. Then we can avoid almost all the null checks being added.

```suggestion
    @observable public dies: WaferMapDie[] = [];
```

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
```diff
@@ -40,4 +43,48 @@ export class WaferMapValidator {
...
+            if (
+                !DataType.isInt(
+                    this.wafermap.diesTable.schema.fields[colIndex]!.type
+                )
+                || !DataType.isInt(
+                    this.wafermap.diesTable.schema.fields[rowIndex]!.type
+                )
+                || !DataType.isFloat(
+                    this.wafermap.diesTable.schema.fields[value]!.type
+                )
```
rajsite, line 83: I don't think these checks include the size checks as specified in the spec, i.e. Int32 and Float64. Can you make sure they are checking size as well and include tests?

rajsite (commented):
Almost ready! Just a quick check after comments addressed and It'll be good 👍

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -52,9 +52,12 @@ export class Prerendering {
...
             this.wafermap.colorScaleMode
         );
+        const isDieRenderInfo = (
+            info: DieRenderInfo | null
+        ): info is DieRenderInfo => info !== null;
         this._diesRenderInfo = this.wafermap.dies
-            .map(die => this.computeDieRenderInfo(die))
-            .filter(info => info !== null) as DieRenderInfo[];
+            ?.map(die => this.computeDieRenderInfo(die))
+            .filter(isDieRenderInfo) ?? [];
```
rajsite, line 60: Hopefully after dies is restored to not being nullable this can be simplified to

```suggestion
            .map(die => this.computeDieRenderInfo(die))
            .filter(isDieRenderInfo);
```

packages/nimble-components/src/wafer-map/modules/data-manager.ts
```diff
@@ -93,7 +93,7 @@ export class DataManager {
 
     private updateDataMap(): void {
         this.dataMap = new Map(
-            this.wafermap.dies.map(die => [`${die.x}_${die.y}`, die])
+            this.wafermap.dies?.map(die => [`${die.x}_${die.y}`, die])
```
rajsite, line 96: think with the above discussion this change can be avoided

packages/nimble-components/src/wafer-map/modules/computations.ts
```diff
@@ -160,9 +160,9 @@ export class Computations {
     }
 
     private calculateGridDimensionsFromDies(
-        dies: Readonly<Readonly<WaferMapDie>[]>
+        dies: Readonly<Readonly<WaferMapDie>[]> | undefined
     ): GridDimensions {
-        if (dies.length === 0 || dies[0] === undefined) {
+        if (dies === undefined || dies.length === 0 || dies[0] === undefined) {
```
rajsite, line 165: think with the above discussion this change can be avoided

packages/nimble-components/src/wafer-map/modules/wafer-map-validator.ts
```diff
@@ -40,4 +43,48 @@ export class WaferMapValidator {
         }
         return !this.invalidGridDimensions;
     }
+
+    public validateDiesTableSchema(): boolean {
```
rajsite, line 47: Can probably save off the indexes to reduce the verbosity in this method to `colIndexField`, `rowIndexField`, `valueField`, etc. Seems unnecessary to have to redo the linear searches to find the indexes again.

rajsite (approved):

Issue #1838, opened 2024-02-19
Nimble Select needs placeholder support


atmgrifter00:
## 😯 Problem to Solve

In many valid usages of a `NimbleSelect` an initial value isn't provided, but it is still a required input for form submission. In such cases it's ideal to have placeholder text in the display area of the select informing the user that they must select a value. The [HTML select spec](https://html.spec.whatwg.org/multipage/form-elements.html#attr-select-required) actually outlines a convention to use for such a situation.

## 💁 Proposed Solution

The HTML spec essentially calls for using an option to be used for the placeholder, particularly the first option with it's value set to empty string, in addition to having the `required` attribute set on the `select` itself. We could either adopt a similar conventional strategy (not taking some elements of the convention like the `required` attribute as that isn't aligned with the Angular forms use-case), or specify some other convention criteria that we document like if the first option is `disabled` and `selected` we will use its text as placeholder text, but not show it in the dropdown.

atmgrifter00:
We were already tracking this issue with #350.

PR #1842, opened 2024-02-20
Iniital implementation of Select placeholder

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1838 

I wanted to get some early feedback on the API design for this feature before writing the tests.

## 👩‍💻 Implementation

The API design present in this PR requires that for a `Select` to show placeholder text, the first option in the set of options in the DOM _must_ set both its `selected` and `disabled` attributes, and that's it. This is distinct from the conventional API that the [HTML `select` spec](https://html.spec.whatwg.org/multipage/form-elements.html#attr-select-required) suggests, where, in addition to the `select` needing to set its `required` attribute, that the first option must have its `value` set to an empty string.

The proposed API seems possibly preferable in that the initial state for the first option (to be a placeholder) is intended to be a "non-possible" state for other options. Whereas the empty string value is a supported state for an option.

The implementation present is essentially doing the following: 
- Determine if the first option is configured to be a placeholder option. If so:
    - Remove that option from the `Select'`s internal notion of its `_options` (this makes it invisible to other things like filtering).
    - Cache the display text for the placeholder option.
- Provide observable state that determines if the placeholder is currently visible, and then show the placeholder text if it is with expected placeholder styling.

## 🧪 Testing

TODO.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/select/index.ts
+++ b/packages/nimble-components/src/select/index.ts
@@ -82,6 +82,9 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     @attr({ attribute: 'filter-mode' })
     public filterMode: FilterMode = FilterMode.none;
 
+    @attr({ attribute: 'placeholder-visible', mode: 'boolean' })
+    public placeholderVisible = false;
+
     /**
      * @internal
      */
@@ -156,7 +159,13 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
      * @internal
      */
     @observable
-    public committedSelectedOption: ListboxOption | undefined = undefined;
+    public committedSelectedOption?: ListboxOption;
+
+    /**
+     * @internal
+     */
+    @observable
+    public placeholderOption?: ListboxOption;
 
     /**
      * The max height for the listbox when opened.
@@ -205,6 +214,11 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     }
 
     public override set options(value: ListboxOption[]) {
+        const firstOptionIsPlaceholder = value.length > 0 && value[0]?.selected && value[0].disabled;
+        if (firstOptionIsPlaceholder) {
+            this.placeholderOption = value.splice(0, 1)[0];
+            this.placeholderOption!.hidden = true;
+        }
         this._options = value;
         Observable.notify(this, 'options');
     }
@@ -256,9 +270,12 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     /**
      * @internal
      */
+    @volatile
     public get displayValue(): string {
         Observable.track(this, 'displayValue');
-        return this.committedSelectedOption?.text ?? '';
+        return this.placeholderVisible
+            ? this.placeholderOption!.text
+            : this.committedSelectedOption?.text ?? '';
     }
 
     /**
@@ -438,7 +455,7 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
         this.filter = this.filterInput?.value ?? '';
         if (!this.committedSelectedOption) {
             this.committedSelectedOption = this._options.find(
-                option => option.selected
+                option => option.selected && !option.disabled
             );
         }
         this.clearSelection();
@@ -740,6 +757,12 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
                 || el.value === this.value
         );
 
+        if (selectedIndex === -1 && this.placeholderOption) {
+            this.selectedIndex = selectedIndex;
+            this.placeholderVisible = true;
+            return;
+        }
+
         if (selectedIndex !== -1) {
             this.selectedIndex = selectedIndex;
             return;
@@ -786,12 +809,17 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
 
         if (filter) {
             this.filteredOptions = this._options.filter(option => {
-                return diacriticInsensitiveStringNormalizer(
-                    option.text
-                ).includes(diacriticInsensitiveStringNormalizer(filter));
+                return (
+                    !option.disabled
+                    && diacriticInsensitiveStringNormalizer(option.text).includes(
+                        diacriticInsensitiveStringNormalizer(filter)
+                    )
+                );
             });
         } else {
-            this.filteredOptions = this._options;
+            this.filteredOptions = this._options.filter(
+                o => !(o.selected && o.disabled)
+            );
         }
 
         this._options.forEach(o => {
@@ -812,6 +840,12 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
         }
 
         if (shouldEmit) {
+            if (
+                this.placeholderVisible
+                && this.value !== this.placeholderOption!.value
+            ) {
+                this.placeholderVisible = false;
+            }
             this.$emit('input');
             this.$emit('change', this, {
                 bubbles: true,
--- a/packages/nimble-components/src/select/index.ts
+++ b/packages/nimble-components/src/select/index.ts
@@ -162,12 +162,6 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     public committedSelectedOption?: ListboxOption;
 
     /**
-     * @internal
-     */
-    @observable
-    public placeholderOption?: ListboxOption;
-
-    /**
      * The max height for the listbox when opened.
      *
      * @internal
@@ -188,6 +182,7 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     private _value = '';
     private forcedPosition = false;
     private indexWhenOpened?: number;
+    private placeholderText?: string;
 
     /**
      * @internal
@@ -216,8 +211,9 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     public override set options(value: ListboxOption[]) {
         const firstOptionIsPlaceholder = value.length > 0 && value[0]?.selected && value[0].disabled;
         if (firstOptionIsPlaceholder) {
-            this.placeholderOption = value.splice(0, 1)[0];
-            this.placeholderOption!.hidden = true;
+            const placeholderOption = value.splice(0, 1)[0];
+            this.placeholderText = placeholderOption!.text;
+            placeholderOption!.hidden = true;
         }
         this._options = value;
         Observable.notify(this, 'options');
@@ -274,7 +270,7 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     public get displayValue(): string {
         Observable.track(this, 'displayValue');
         return this.placeholderVisible
-            ? this.placeholderOption!.text
+            ? this.placeholderText!
             : this.committedSelectedOption?.text ?? '';
     }
 
@@ -757,7 +753,7 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
                 || el.value === this.value
         );
 
-        if (selectedIndex === -1 && this.placeholderOption) {
+        if (selectedIndex === -1 && (typeof this.placeholderText === 'string')) {
             this.selectedIndex = selectedIndex;
             this.placeholderVisible = true;
             return;
@@ -842,7 +838,7 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
         if (shouldEmit) {
             if (
                 this.placeholderVisible
-                && this.value !== this.placeholderOption!.value
+                && this.value !== this.placeholderText
             ) {
                 this.placeholderVisible = false;
             }
--- a/packages/nimble-components/src/select/styles.ts
+++ b/packages/nimble-components/src/select/styles.ts
@@ -25,6 +25,10 @@ export const styles = css`
         /* We are using flex `order` to define the visual ordering of the selected value,
          error icon, and dropdown arrow because they are not "interactive" i.e. part of the tab order */ ''
     }
+    :host([placeholder-visible]) [part='selected-value'] {
+        color: ${placeholderFontColor};
+    }
+
     [part='selected-value'] {
         order: 1;
     }
--- a/packages/nimble-components/src/select/tests/select.mdx
+++ b/packages/nimble-components/src/select/tests/select.mdx
@@ -6,7 +6,7 @@ import { listOptionTag } from '../../list-option';
 <Meta of={selectStories} />
 <Title of={selectStories} />
 
-Select is a control for selecting amongst a set of options. Its value comes from the `value` of the currently selected <Tag name={listOptionTag}/>, or, if no value exists for that option, the option's content. Upon clicking on the element, the other options are visible. The user cannot manually enter values, and thus the list cannot be filtered.
+Select is a control for selecting amongst a set of options. Its value comes from the `value` of the currently selected <Tag name={listOptionTag}/>, or, if no value exists for that option, the option's content. Upon clicking on the element, the other options are visible.
 
 <Canvas of={selectStories.underlineSelect} />
 <Controls of={selectStories.underlineSelect} />
--- a/packages/nimble-components/src/select/tests/select.stories.ts
+++ b/packages/nimble-components/src/select/tests/select.stories.ts
@@ -1,4 +1,4 @@
-import { html, repeat } from '@microsoft/fast-element';
+import { html, repeat, when } from '@microsoft/fast-element';
 import { withActions } from '@storybook/addon-actions/decorator';
 import type { HtmlRenderer, Meta, StoryObj } from '@storybook/html';
 import {
@@ -20,12 +20,14 @@ interface SelectArgs {
     optionsType: ExampleOptionsType;
     appearance: string;
     filterMode: keyof typeof FilterMode;
+    placeholder: boolean;
 }
 
 interface OptionArgs {
     label: string;
     value: string;
     disabled: boolean;
+    selected?: boolean;
 }
 
 const simpleOptions: readonly OptionArgs[] = [
@@ -73,6 +75,10 @@ The act of filtering will use the \`hidden\` attribute on the options to remove 
 It is recommended that if the \`Select\` has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`.
 `;
 
+const placeholderDescription = `
+To display placeholder text within the \`Select\` you must provide an option before all other options that has both the \`disabled\` and \`selected\` attributes initially set. This option will not be available in the dropdown, and its contents will be used as the placeholder text.
+`;
+
 const metadata: Meta<SelectArgs> = {
     title: 'Components/Select',
     decorators: [withActions<HtmlRenderer>],
@@ -96,10 +102,18 @@ const metadata: Meta<SelectArgs> = {
             filter-mode="${x => (x.filterMode === 'none' ? undefined : x.filterMode)}"
             style="width: var(${menuMinWidth.cssCustomProperty});"
         >
+            ${when(x => x.placeholder, html`
+                <${listOptionTag}
+                    disabled
+                    selected>
+                    Select an option:
+                </${listOptionTag}?
+            `)}
             ${repeat(x => optionSets[x.optionsType], html<OptionArgs>`
                 <${listOptionTag}
                     value="${x => x.value}"
                     ?disabled="${x => x.disabled}"
+                    ?selected="${x => x.selected}"
                 >
                     ${x => x.label}
                 </${listOptionTag}>
@@ -126,6 +140,10 @@ const metadata: Meta<SelectArgs> = {
         errorVisible: {
             name: 'error-visible'
         },
+        placeholder: {
+            name: 'placeholder',
+            description: placeholderDescription
+        },
         optionsType: {
             name: 'options',
             options: Object.values(ExampleOptionsType),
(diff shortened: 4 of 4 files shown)


atmgrifter00:
@rajsite and @jattasNI would you mind taking a look at this draft PR to review the proposed API for the `Select` placeholder text? As I discussed with @jattasNI, this API is orthogonal to any work we define for the interaction to remove a selected value (which would ultimately display the placeholder text if configured that way).

packages/nimble-components/src/select/index.ts
```diff
@@ -82,6 +82,12 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     @attr({ attribute: 'filter-mode' })
     public filterMode: FilterMode = FilterMode.none;
 
+    /**
```
rajsite, line 85: @atmgrifter00 moved to comment as feedback was requested

@rajsite and @jattasNI would you mind taking a look at this draft PR to review the proposed API for the `Select` placeholder text? As I discussed with @jattasNI, this API is orthogonal to any work we define for the interaction to remove a selected value (which would ultimately display the placeholder text if configured that way).

packages/nimble-components/src/select/tests/select.stories.ts
```diff
@@ -73,6 +75,10 @@ The act of filtering will use the \`hidden\` attribute on the options to remove
 It is recommended that if the \`Select\` has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`.
 `;
 
+const placeholderDescription = `
+To display placeholder text within the \`Select\` you must provide an option before all other options that has both the \`disabled\` and \`selected\` attributes initially set. This option will not be available in the dropdown, and its contents will be used as the placeholder text.
```
jattasNI, line 79: What benefit do we get from specifying a placeholder via an impossible combination of existing attributes (`disabled` and `selected`) vs adding a new attribute? To me it seems like a somewhat convoluted way to achieve this outcome.

When we were talking earlier I thought the benefit was that it matched the native select API. But from the spec link you added, the only part that I see being aligned is that the placeholder is specified by configuring the first option. So could we achieve the same thing more explicitly with something like this?

```html
<nimble-select>
  <nimble-list-option placeholder>Select an item</nimble-list-option>
  <nimble-list-option>First</nimble-list-option>
  <nimble-list-option>Second</nimble-list-option>
</nimble-select>
```

We could enforce (or document but not enforce) whatever rules we wanted like:
1. `placeholder` can only be added to one option
2. `placeholder` must be added to the first option
3. `placeholder` shouldn't be used with `disabled` or `selected`

atmgrifter00: At the moment I have nothing against this approach.

jattasNI: Since [`placeholder` is a built-in HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder) we might decide to use something custom like `is-placeholder` to follow our [attribute naming convention](https://github.com/ni/nimble/blob/877f01c4512dccfe59767c82806467861460b227/packages/nimble-components/CONTRIBUTING.md#attribute-naming-convention).

rajsite: > What benefit do we get from specifying a placeholder via an impossible combination of existing attributes (disabled and selected) vs adding a new attribute?

Because it aligns with conventions in HTML and we should follow that over inventing new apis.

rajsite: Openui is making the new selectmenu element and this is their current idea: https://github.com/openui/open-ui/issues/637#issuecomment-1682780451

> RESOLVED: Support the value="" attribute, selected attribute, disabled attribute, and hidden attribute on option elements in selectlist. For v2, consider a placeholder attribute or element that implies <option value="" disabled hidden>.

jattasNI: I didn't see anything about the `disabled` and `selected` parts of that API in the spec link in the PR description:
>If a select element has a required attribute specified, does not have a multiple attribute specified, and has a display size of 1; and if the value of the first option element in the select element's list of options (if any) is the empty string, and that option element's parent node is the select element (and not an optgroup element), then that option is the select element's placeholder label option.

Open UI is a good guide to follow. If we can include links to that and to the place the spec specifies `disabled` and `selected`, then I'm fine to follow that convention.

rajsite: > I didn't see anything about the disabled and selected parts of that API in [the spec link in the PR description](https://html.spec.whatwg.org/multipage/form-elements.html#attr-select-required)

Sorry, you're right spec didn't show it because that convention for placeholder described in the spec allows the placeholder to be selected.

Our usage guidance would be more like the openui resolution to configure:
- required + value="": causes it to fail HTML 5 form validation
- disabled: prevents it from being selectable but also prevents it from being the initial value
- selected: forces it to be selected even though it is disabled
- hidden: hides it in the option list

jattasNI: The API that started out seeming convoluted makes more sense when broken down into those pieces. Thanks!

This brings up another question: right now the implementation is has special cases to detect this combination of attributes and behave differently. Should/could we instead implement each of the behaviors independently and have the placeholder behavior fall out naturally?

atmgrifter00: I think your suggestion is aimed at responding to changes to the options themselves, and if one ends up in the state that is indicative of the placeholder, that it _becomes_ the placeholder (presuming there is no option before it that currently represents the placeholder). I have nothing against this. I think this is an additive change to the existing implementation.

rajsite: Very apropos 
Screenshot_20240221-185226

atmgrifter00: I have implemented the API in such a way that the "placeholder" option more-or-less just falls out naturally from the API configuration. The `Select` still must know that its current displayed option is the "placeholder" option so that it can be styled accordingly, and thus there is some implementation in place to make that possible.

packages/nimble-components/src/list-option/index.ts
```diff
@@ -30,6 +42,42 @@ export class ListOption extends FoundationListboxOption {
...
+            this.parentElement.updateDisplayValue();
+        }
+    }
+
+    private hiddenChanged(): void {
+        if (this.hidden) {
+            this.classList.add('hidden-option');
+        } else {
+            this.classList.remove('hidden-option');
+        }
```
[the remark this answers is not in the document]
atmgrifter00: I chatted with @jattasNI off-line about this, and he had the good idea that we should be able to just subscribe to `disabled` and `hidden` changes on the currently selected option. I'm going to run with this as an idea.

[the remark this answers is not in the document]
rajsite: oops, missed the notif, but yea looks like the [handleChanged listening on options](https://github.com/microsoft/fast/blob/c64110d2f1dfbc24a0978c362bcf72b8958e6d73/packages/web-components/fast-foundation/src/listbox/listbox.ts#L289) is public so would make sense to override to listen for our extra props and call super 👍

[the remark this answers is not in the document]
atmgrifter00: @jattasNI and @rajsite, well it turns out I needed to add notifier subscriptions to `hidden` and `disabled` for _all_ options. It was either that or add some side-effecty logic when the selection state changes. Ultimately, I think it's just more clear in its current form, and, again, we're not adding any real overhead to what was there before, as we were already subscribing to the `selected` property change.

packages/nimble-components/src/select/tests/select.stories.ts
```diff
@@ -1,4 +1,4 @@
-import { html, repeat } from '@microsoft/fast-element';
+import { html, repeat, when } from '@microsoft/fast-element';
```
jattasNI, line 1: Is this text from the filterMode attribute still accurate?
>The act of filtering will use the hidden attribute on the options to remove and re-add them to the visible set. Thus, any client-provided hidden settings of the options will be overridden.

Also, not part of this PR but worth quickly fixing: I noticed we called it `filterMode` with the property naming convention but for storybook we generally prefer the attribute naming convention `filter-mode`.

packages/nimble-components/src/select/tests/select.stories.ts
```diff
@@ -73,6 +74,10 @@ The act of filtering will use the \`hidden\` attribute on the options to remove
 It is recommended that if the \`Select\` has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`.
 `;
 
+const placeholderDescription = `
+To display placeholder text within the \`Select\` you must provide an option that has the \`disabled\`, \`selected\` and \`hidden\` attributes set. This option will not be available in the dropdown, and its contents will be used as the placeholder text.
```
jattasNI, line 78: There's a bit of usage guidance about placeholder text in the [select IxD spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/select/specs/IxD.md#field-button). Assuming we still agree with it, we should replicate it here.

**Original text**
>Selects can be empty by default or have a default selection from the menu. Empty selects should include placeholder text that prompts the end-user to select an option from the menu. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Ellipses are not needed. Use sentence casing.

**My editorialized suggestion, feel free to tweak**
```suggestion
To display placeholder text within the \`Select\` you must provide an option that has the \`disabled\`, \`selected\` and \`hidden\` attributes set. This option will not be available in the dropdown, and its contents will be used as the placeholder text.

Any `Select` without a default selected option should provide placeholder text. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Use sentence casing and don't include punctuation at the end of the prompt.
```

packages/nimble-components/src/select/tests/select.stories.ts
```diff
@@ -96,6 +101,14 @@ const metadata: Meta<SelectArgs> = {
             filter-mode="${x => (x.filterMode === 'none' ? undefined : x.filterMode)}"
             style="width: var(${menuMinWidth.cssCustomProperty});"
         >
+            ${when(x => x.placeholder, html`
+                <${listOptionTag}
+                    disabled
+                    selected
+                    hidden>
+                    Select an option:
```
jattasNI, line 109: Assuming we agree on the guidance in the other thread, we should remove the colon
```suggestion
                    Select an option
```

packages/nimble-components/src/select/index.ts
```diff
@@ -293,16 +278,20 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
             const notifier = Observable.getNotifier(o);
             notifier.unsubscribe(this, 'value');
+            notifier.unsubscribe(this, 'hidden');
+            notifier.unsubscribe(this, 'disabled');
         });
 
         super.slottedOptionsChanged(prev, next);
 
-        this._options.forEach(o => {
+        this.options.forEach(o => {
```
jattasNI, line 290: While playing around in dev tools in storybook, I noticed that changing the content of the selected list option (or the placeholder when nothing is selected) does NOT reflect in the select. Can it / should it?

Since this is pre-existing behavior I don't think it needs to be changed as part of this PR but if we think it matters we could fix in a follow up.

jattasNI (changes requested):
Initial feedback from playing with storybook. Haven't looked at component implementation yet.

packages/nimble-components/src/select/tests/select.foundation.spec.ts
```diff
@@ -245,24 +245,6 @@ describe('Select', () => {
         await disconnect();
     });
```
jattasNI, line 247: It looks like you can now select a disabled option using the keyboard. This seems wrong.

jattasNI: https://github.com/ni/nimble/assets/10500124/0b097f5d-dde2-4d94-b45b-2dfae04ee550

packages/nimble-components/src/select/index.ts
```diff
@@ -357,9 +342,38 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
+                    this.selectedIndex = this.options.indexOf(sourceElement);
+                }
+                this.setSelectedOptions();
+                this.updateDisplayValue();
+                break;
+            }
+            case 'hidden': {
+                if (isListboxOption(sourceElement)) {
+                    if (sourceElement.hidden) {
+                        sourceElement.classList.add('hidden-option');
```
rajsite, line 363: [we don't own the class attribute / classList configuration of an element](https://github.com/ni/nimble/blob/33354140e84c85eff7fe47c2452f559a29a52ac0/packages/nimble-components/CONTRIBUTING.md#why-attributes-over-classes) but we do own attribute definition. To avoid potentially fighting with the user manipulating classes on the element can define an internal property or attribute?
And like we've chatted about before, all else equal when dealing with internal state a property is preferred but in this case since we want to style against it an attribute is perfectly reasonable.
Maybe `visually-hidden` (to distinguish from logically hidden)?

packages/nimble-components/src/select/tests/select.foundation.spec.ts
atmgrifter00: Whoops! Fixed and added test.

packages/nimble-components/src/select/index.ts
atmgrifter00: Going by how the native `select` works, a change to the content of a selected option _should_ update the `Select` display. I'll look into this for a follow up PR.

atmgrifter00: @jattasNI, I'll note that if you update the content of an option by assigning its `textContent` a new value in the console, the `Select` display _does_ update. I imagine the only way to get it to work in a more complete way is to use a `MutationObserver`, which FAST does not do, and I'm unsure there's enough value there to implement that.

jattasNI: OK, cool. I was updating it by editing the DOM in dev tools, which is not nearly as important a use case. I'm fine to leave the behavior as-is since the programmatic API works.

Though we should probably have a test to ensure the `textContent` case continues to work.

packages/nimble-components/src/select/index.ts
```diff
@@ -785,17 +848,28 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
 
-        this._options.forEach(o => {
-            o.hidden = !this.filteredOptions.includes(o);
+        this.options.forEach(o => {
+            if (isListOption(o)) {
+                if (!this.filteredOptions.includes(o)) {
+                    o.visuallyHidden = true;
+                } else {
+                    o.visuallyHidden = false;
+                }
```
jattasNI, line 871:
```suggestion
                o.visuallyHidden = !this.filteredOptions.includes(o);
```

jattasNI (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: 🤦‍♂️

packages/nimble-components/src/select/styles.ts
```diff
@@ -25,6 +25,11 @@ export const styles = css`
...
          error icon, and dropdown arrow because they are not "interactive" i.e. part of the tab order */ ''
     }
+
+    [part='selected-value'].placeholder {
+        color: ${placeholderFontColor};
+    }
+
     [part='selected-value'] {
         order: 1;
     }
```
rajsite, line 35: nit / pref. We already have a class on that item in the template. Using that is a bit cleaner
```suggestion
    .selected-value.placeholder {
        color: ${placeholderFontColor};
    }

    .selected-value {
        order: 1;
    }
```

packages/nimble-components/src/select/template.ts
```diff
@@ -119,7 +119,7 @@ SelectOptions
                         class="scrollable-region">
                         <slot
                             ${slotted({
-                                filter: (n: Node) => n instanceof HTMLElement && Listbox.slottedOptionFilter(n),
+                                filter: (n: Node) => n instanceof HTMLElement && isListboxOption(n),
```
rajsite, line 122: Need both?

```suggestion
                                filter: (n: Node) => isListboxOption(n),
```

[isListboxOption](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts#L22) [isHTMLElement](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/utilities/fast-web-utilities/src/dom.ts#L6)

packages/nimble-components/src/select/index.ts
```diff
@@ -51,6 +50,10 @@ declare global {
 // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
 type BooleanOrVoid = boolean | void;
 
+const isListOption = (el: Element): el is ListOption => {
+    return el instanceof ListOption;
+};
```
rajsite, line 55: Can we use that `isListboxOption` from foundation like we are in the template?

If we are looking for s stricter check with our nimble specific apis maybe call it `isNimbleListOption`?

packages/nimble-components/src/select/index.ts
rajsite: It does seem kinda racy. Should maybe be a usage note in the select docs that the content of the placeholder option needs to be set before setting the attributes for the display text to be correct.

We could consider in the future just setting the MutationObserver up on the placeholder option instead of all options.

packages/nimble-components/src/select/index.ts
```diff
@@ -785,17 +844,28 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
         if (filter) {
-            this.filteredOptions = this._options.filter(option => {
-                return diacriticInsensitiveStringNormalizer(
-                    option.text
-                ).includes(diacriticInsensitiveStringNormalizer(filter));
+            this.filteredOptions = this.options.filter(option => {
+                return (
+                    !option.hidden
+                    && diacriticInsensitiveStringNormalizer(option.text).includes(
+                        diacriticInsensitiveStringNormalizer(filter)
```
rajsite, line 851: `diacriticInsensitiveStringNormalizer(filter)` can we run that once outside the `filter` loop?

rajsite (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: I'm not sure I follow your suggestion for the usage docs. Do you simply mean that we should stress that when setting up a placeholder option that they should follow the convention of putting the contents in the markup, and not assigning it later? I'm not sure what the race condition is, and this isn't a placeholder issue, but a general one.

packages/nimble-components/src/list-option/index.ts
```diff
@@ -19,6 +19,23 @@ export class ListOption extends FoundationListboxOption {
...
+     * @remarks
+     * HTML Attribute: hidden
+     */
+    @attr({ mode: 'boolean' })
+    public override hidden = false;
+
+    /**
+     * @internal
+     */
+    @attr({ attribute: 'visually-hidden', mode: 'boolean' })
```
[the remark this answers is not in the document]
atmgrifter00: Created #1902.

packages/nimble-components/src/select/template.ts
atmgrifter00: Sorry...this doesn't work. The `instanceof` check allows for it to be passed to the `isListboxOption` method which expects an `Element` type.

packages/nimble-components/src/select/tests/select.mdx
```diff
@@ -6,7 +6,7 @@ import { listOptionTag } from '../../list-option';
 <Meta of={selectStories} />
 <Title of={selectStories} />
 
-Select is a control for selecting amongst a set of options. Its value comes from the `value` of the currently selected <Tag name={listOptionTag}/>, or, if no value exists for that option, the option's content. Upon clicking on the element, the other options are visible. The user cannot manually enter values, and thus the list cannot be filtered.
+Select is a control for selecting amongst a set of options. Its value comes from the `value` of the currently selected <Tag name={listOptionTag}/>, or, if no value exists for that option, the option's content. Upon clicking on the element, the other options are visible.
```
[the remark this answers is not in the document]
atmgrifter00: @fredvisser do you have an opinion? My current thought is that might be overly prescriptive, or at the very least "overly detailed".

packages/nimble-components/src/select/index.ts
rajsite: I see, it's an issue for any selected option, placeholder or otherwise.

> I'm not sure what the race condition is

> they should follow the convention of putting the contents in the markup, and not assigning it later

Maybe race is the wrong word. But like the subject of this thread the order of operations is significant. Your display text won't be recognized in some cases. That should be documented.

atmgrifter00: I've added the test @jattasNI suggested. I'll update docs as needed in a follow-up PR if there is, in fact, some agreed upon doc changes.

packages/nimble-components/src/select/tests/select.mdx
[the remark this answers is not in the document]
atmgrifter00: I'll leave any agreed upon doc changes for a follow-on PR.

PR #1846, opened 2024-02-20
Add Blazor support for delayed hierarchy in the table

mollykreis:
# Pull Request

## 🤨 Rationale

This PR adds Blazor support for #1758 

## 👩‍💻 Implementation

This PR:
- Exposes `SetRecordHierarchyOptionsAsync` on the Blazor table to allow hierarchy options to be configured
- Exposes the `RowExpandToggle` event on the Blazor table
- Adds the `RowLoading` attribute to Blazor's table label provider
- Adds a new table to the Blazor example app that illustrates the usage of the hierarchy options in the exact same manner as was done in the Angular example app
- Modifies the strings associated with the `canLoadChildren` and `loadingChildren` enum values in nimble-components because they didn't follow nimble's kebab casing convention, which ended up causing problems during serialization in Blazor
- Fixes a bug in the table row's template where the closing tag for the nimble-spinner was missing its `/`

## 🧪 Testing

- Added new unit and acceptance tests following the existing patterns in the repo
- Manually tested the changes to the example app

## ✅ Checklist


mollykreis:

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/TableTests.cs
```diff
@@ -24,5 +24,45 @@ await using (var pageWrapper = await NewPageForRouteAsync("TableSetData"))
                 await Assertions.Expect(rows).ToContainTextAsync(new string[] { "A0", "A1", "A2", "A3", "A4" });
             }
         }
+
+        [Theory]
+        [InlineData(0, true, false)]
+        [InlineData(1, false, false)]
+        [InlineData(2, false, true)]
+        [InlineData(3, false, false)]
```
atmgrifter00, line 32: I presume the test is structured this way so we don't do assertions within a loop? Can't think of a real alternative to this (this initially felt odd to me as we are making a theory-based test that doesn't actually alter any state based on the theory data).

mollykreis: Yes, you are correct. I created it like this to have a test per row, which corresponds to a test per possible delayed hierarchy state (`canLoadChildren`, `loadingChildren`, `none`, and not set), rather than a single test that loops through checking each row.

packages/nimble-components/src/table/types.ts
```diff
@@ -95,8 +95,8 @@ export interface TableRecordHierarchyOptions {
 
 export const TableRecordDelayedHierarchyState = {
     none: undefined,
-    canLoadChildren: 'canLoadChildren',
-    loadingChildren: 'loadingChildren'
+    canLoadChildren: 'can-load-children',
```
rajsite, line 98: Looks like TableRowSelectionState has the same inconsistency and appears to be all internal, Could we clean that one up as well?

packages/nimble-blazor/Tests/NimbleBlazor.Tests.Acceptance/Tests/TableTests.cs
rajsite, line 32: nit / pref: Could we have a single enum for the states instead of two booleans?

packages/nimble-blazor/NimbleBlazor/wwwroot/NimbleBlazor.lib.module.js
```diff
@@ -118,6 +118,17 @@ export function afterStarted(Blazor) {
             };
         }
     });
+    // Used by NimbleTable.razor
+    Blazor.registerCustomEventType('nimbletablerowexpandtoggle', {
+        browserEventName: 'row-expand-toggle',
```
rajsite, line 123: I certainly may have missed it but is there an acceptance / playwright test for this event?

rajsite (approved):

packages/nimble-components/src/table/types.ts
mollykreis: Yes, I can update that. It's an internal enum, so I wouldn't expect it to ever be used in or cause problems in Blazor. However, it seems like a good idea to go ahead and make it align with out conventions.

jattasNI (approved):

packages/nimble-blazor/NimbleBlazor/wwwroot/NimbleBlazor.lib.module.js
mollykreis: There wasn't one, but now I've added it.

rajsite: love it 👍

PR #1847, opened 2024-02-20
Preserve spaces between inline elements

### LINKED ISSUE #1840: Preserve spaces between inline elements, "captureSpacesBetweenElements"
## 😯 Problem to Solve

Currently xliff contents don't keep the spaces between inline elements.

```
<?xml version="1.0" encoding="UTF-8"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
    <file source-language="en-US" datatype="plaintext" original="ng2.template" target-language="de-DE">
        <body>
            <trans-unit id="5478462691360784268" datatype="html">
                <source><x id="PH" equiv-text="fileName1"/> <x id="PH_1" equiv-text="fileName2"/> + 1 other</source>
                <target state="final"><x id="PH" equiv-text="fileName1"/> <x id="PH_1" equiv-text="fileName2"/> + 1 weitere</target>
            </trans-unit>
        </body>
    </file>
</xliff>
```

Is converted to: `{$PH}{$PH_1} + 1 weitere`

While the expected behaviour is that the space between the first & second tag is preserved, resulting to: `{$PH} {$PH_1} + 1 weitere`

## 💁 Proposed Solution

As suggested in one of the github issues in xliff library, the method _xliff12ToJS_, now supports an option `captureSpacesBetweenElements` : https://github.com/locize/xliff/issues/55
since version xliff v6.1.0

My suggestion is to pass `captureSpacesBetweenElements: true` as default option to _xliff12ToJs_. 

I can create a PR.

## 📋 Tasks
- [ ] This is a subtask of the feature. (It can be converted to an issue.)
-->

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1840

## 👩‍💻 Implementation

- Bump the minimum version of the `xliff` dependency to one that supports the `captureSpacesBetweenElements` option
- Pass `{ captureSpacesBetweenElements: true }` as options to `xliff12ToJs`
- Update our type definition for the `xliff12ToJs` to claim it can be passed an option object containing a `captureSpacesBetweenElements` boolean
    - Note: It appears that the `xliff` library supports more option configurations than simply a `captureSpacesBetweenElements`, but I couldn't find a clear definition of what all the options were. Therefore, I decided to define it in our type with only the option we need.

## 🧪 Testing

- Created a new unit test that failed before the change and passes with it

## ✅ Checklist

### DIFF
--- a/change/@ni-xliff-to-json-converter-03082322-18a9-410a-9e3b-6625f96dc66e.json
+++ b/change/@ni-xliff-to-json-converter-03082322-18a9-410a-9e3b-6625f96dc66e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Preserve spaces between inline elements",
+  "packageName": "@ni/xliff-to-json-converter",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -34152,7 +34152,7 @@
       "version": "1.1.3",
       "license": "MIT",
       "dependencies": {
-        "xliff": "^6.0.3",
+        "xliff": "^6.1.0",
         "yargs": "^17.5.1"
       },
       "bin": {
--- a/packages/xliff-to-json-converter/package.json
+++ b/packages/xliff-to-json-converter/package.json
@@ -37,7 +37,7 @@
     "typescript": "~4.8.2"
   },
   "dependencies": {
-    "xliff": "^6.0.3",
+    "xliff": "^6.1.0",
     "yargs": "^17.5.1"
   }
 }
--- a/packages/xliff-to-json-converter/src/convert.spec.ts
+++ b/packages/xliff-to-json-converter/src/convert.spec.ts
@@ -229,6 +229,28 @@ describe('xliff2Json', () => {
         expect(json.translations['5478462691360784268']).toEqual('{$PH}, {$PH_1} + 1 weitere');
     });
 
+    it('preserves spaces between inline elements', async () => {
+        const xliffContents = `
+        <?xml version="1.0" encoding="UTF-8"?>
+        <xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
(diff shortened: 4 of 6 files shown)


mollykreis:
@jattasNI, can you buddy this PR for me?

change/@ni-xliff-to-json-converter-03082322-18a9-410a-9e3b-6625f96dc66e.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I [created a PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/657293) to configure Renovate to uptake this in Skyline.

I also did a search of the `Web` directory for instances of `/> <x id="PH` in xlf files and there are a lot of them (139 results in 20 files). There are a few hundred more if you omit the `id="PH` part. This makes me a bit nervous that this change will have a noticeable impact in SLE or cause some extra work for the Localization team. Some ideas to mitigate that before completing this PR:
1. spot check a few of those results and see if the generated JSON file changes as a result of this update
2. if it does, leave a comment on [the localization team](https://teams.microsoft.com/l/channel/19%3Acd7187eeafd24fc3a3d98256ba0c55c2%40thread.skype/Prj_SystemLink?groupId=0af69736-3f49-48ec-8c7e-d122f54130d5&tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e) to warn them that this change is coming and give them a chance to push back.

mollykreis: I ran our string extraction & conversion commands on one of our apps with these changes, and it appears that the update is fairly innocuous.

To give a bit more detail, the application has ~1220 strings that get extracted, and 17 of those were modified with this change. Most of those seem irrelevant because they are strings that look to be incorrectly marked for localization because they contain element tags in the string. And the ones that actually look relevant (3 of the 17 changes), that change looks beneficial because it is now maintaining spaces that were present in the original string in the source code.

jattasNI (approved):

rajsite (approved):

TrevorKarjanis (approved):

Issue #1445, opened 2023-08-18
nimble-dialog does not have a styled focus-visible state


mollykreis:
# 🐛 Bug Report

The `nimble-dialog` does not have a styled focus-visible state. As a result, it is using the browser's default focus-visible styles, which is likely not what we want for our nimble component.

In Edge, the default focus-visible state makes the dialog look like this:
![image](https://github.com/ni/nimble/assets/20542556/b2bfe74f-01ae-4c0b-8dca-a3980dee594e)

## 🤔 Expected Behavior

The focus-visible state of the nimble-dialog should be determined by nimble's styling, as guided by nimble designers.

## 😯 Current Behavior

The browser is using its default focus-visible styling.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

Issue #1713, opened 2023-12-18
Nimble Drawer is not correctly registering closing by pressing "esc" key.


SPathani-NI:
# 🐛 Bug Report

If "esc" is used to close nimble drawer, then then after closing the drawer will never reopen again with the following error in the console:
![image](https://github.com/ni/nimble/assets/147801131/5e763a9d-6ca6-4fac-b326-41b4feb51d24)

## 💻 Repro or Code Sample

can be replicated here : https://nimble.ni.dev/storybook/?path=/docs/components-drawer--docs

once open close the drawer by pressing esc then try opening it again.
## 🤔 Expected Behavior

Should properly register pressing esc as drawer closed.

## 😯 Current Behavior

https://github.com/ni/nimble/assets/147801131/87d97dc8-6ff8-4618-b69d-efcd3c225587

mollykreis:
This is a Chromium issue with the native `dialog` element that affects the latest versions of Edge and Chrome. The issue exists in the `nimble-drawer` and in the `nimble-dialog`.

From what I can tell, Chromium is working on getting the issue fixed soon. https://bugs.chromium.org/p/chromium/issues/detail?id=1512224

rajsite:
Discussion at standup:

- We will wait until next triage (1/9) to see if fixed in Chrome stable, if not we should consider a workaround
- The chromium issue suggests that user interaction requirement aligns with spec (spec requires user interaction to observe CloseWatcher events for security). So they may decide the behavior is not a bug in the future if the security requirement is necessary. We should make sure to have a regression test.

saikrishnan-ni:
Team,
We found a behavior with Chromium browsers for the drawer component and felt to be related to this issue, hence posting here.

When used in `Edge` browser, even with `preventDismiss` set pressing `Esc` registers and closes the drawer. 
This makes the drawer to not open again next time as it's state will mostly be dependent on any variable and whose state won't be updated.

i: This works as expected in chrome browser

m-akinc:
Issue is currently fixed in Chrome. Need to validate that the issue is fixed in the next Edge release (should be on Jan 25th). Releases: https://learn.microsoft.com/en-us/deployedge/microsoft-edge-release-schedule

m-akinc:
Validated in latest version of Edge and Chrome.

PR #1848, opened 2024-02-20
Make dialog and drawer more resilient to upcoming CloseWatcher API

m-akinc:
# Pull Request

## 🤨 Rationale

There is a change to the HTML spec introducing the `CloseWatcher` API, which affects the way the native `dialog` element behaves. Recently, Chrome had released their implementation, and it broke our drawer and dialog (see issue https://github.com/ni/nimble/issues/1713, as well as resulting [Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1512224)). Though they rolled back their release of that feature (by moving it back to experimental), it will eventually be implemented in Chrome and other browsers in some form. We want to at least guard against leaving the drawer/dialog in a bad state.

Also fixes https://github.com/ni/nimble/issues/1445

## 👩‍💻 Implementation

The issue is that the `cancel` event is intentionally not fired in a couple cases where the `dialog` is dismissed with ESC.
- One such case was when the there is no user interaction (e.g. clicking, scrolling) in the `dialog` before pressing ESC. Apparently, this is [viewed as a mistake and will likely be changed](https://github.com/whatwg/html/issues/10046).
- Another case that is [more fundamental to the `CloseWatcher` API](https://github.com/WICG/close-watcher#abuse-analysis) is when there are two ESC presses without an intervening user interaction.

I have added a `close` event handler in both the drawer and dialog that resolves and clears the promise, if there is one still set. This ensures we are not left in a bad state where the drawer/dialog can't be reopened. Note there are still problems:
- the drawer closes without animation (because we rely on the `cancel` event to trigger the animation)
- the drawer/dialog can be closed with ESC even if `prevent-dismiss` is set

Depending on the ultimate `CloseWatcher` specification, we may decide to deprecate the `prevent-dismiss` option.

OTHER CHANGES:
- Gave the dialog proper focus styling, rather than the default.

## 🧪 Testing

I added tests for the specific case now being handled, though these won't currently be running against any browser versions that would exhibit the problem behavior.

I did manual testing using a [version of Chromium](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/1241130/) with the `CloseWatcher` API enabled.

## ✅ Checklist


packages/nimble-components/src/dialog/styles.ts
```diff
@@ -15,7 +15,8 @@ import {
     elevation3BoxShadow,
     dialogSmallWidth,
     dialogSmallHeight,
```
jattasNI, line 17: I was surprised not to see Chromatic diffs for the styling change.

Is it because the focus visible state only happens if you open the dialog from a focus visible element (i.e. by focusing the button and hitting enter)? If so should we add a matrix test that does that in a play function?

Or is it a gap in how Chromatic takes snapshots? I noticed [Malcolm added a workaround to one of his PRs for something that seems related](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/652640?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/src/tests/angular-material/angular-material-host.component.scss) so you could talk to him about copying that pattern.

It's not critical to add test coverage since we have lots of existing gaps for focus visible chromatic testing, but if it's easy then let's do it.

jattasNI (commented):
Great job providing rationale and comments for all the subtle behaviors in this PR!

rajsite:
Sorry, but last time adding @mollykreis as they are very close to the implementation and aware of the bug. Should be a quick double check.

packages/nimble-components/src/dialog/index.ts
```diff
@@ -125,11 +124,33 @@ export class Dialog<CloseReason = void> extends FoundationElement {
...
+        if (this.resolveShow) {
+            // If
+            // - the browser implements dialogs with the CloseWatcher API, and
+            // - the user presses ESC without first interacting with the dialog (e.g. clicking, scrolling),
+            // the cancel event is not fired and the dialog just closes.
+            this.notifyClosed(UserDismissed);
+        }
+    }
+
+    private notifyClosed(reason: CloseReason | UserDismissed): void {
```
rajsite, line 145: I don't think notify is the right term. The convention is notify is used for event listeners or observation.

The action is to resolve the promise, maybe `completeResolveShow` or `finishResolveShow` or `doResolveShow`

packages/nimble-components/src/dialog/index.ts
```diff
@@ -125,11 +124,33 @@ export class Dialog<CloseReason = void> extends FoundationElement {
...
+
+    /**
+     * @internal
+     */
+    public closeHandler(): void {
+        if (this.resolveShow) {
+            // If
+            // - the browser implements dialogs with the CloseWatcher API, and
+            // - the user presses ESC without first interacting with the dialog (e.g. clicking, scrolling),
+            // the cancel event is not fired and the dialog just closes.
```
rajsite, line 140: the cancel event is not fired but the close event still is? If that's what's described can that be clarified?

```suggestion
            // the cancel event is not fired, but the close event still is, and the dialog just closes.
```

rajsite (approved):

packages/nimble-components/src/dialog/index.ts
```diff
@@ -125,11 +124,33 @@ export class Dialog<CloseReason = void> extends FoundationElement {
...
+            this.notifyClosed(UserDismissed);
+        }
+    }
+
+    private notifyClosed(reason: CloseReason | UserDismissed): void {
+        if (!this.resolveShow) {
+            throw new Error(
+                'Do not call notifyClosed unless there is a promise to resolve'
```
rajsite, line 148: Can a user end up with this error being called or is this an internal error / guard against implementation errors?

If it's user facing / possible for a user to trigger the code path with the public api can we have a test to cover it and frame the error for the user?

packages/nimble-components/src/drawer/index.ts
```diff
@@ -92,6 +92,29 @@ export class Drawer<CloseReason = void> extends FoundationElement {
...
+    /**
+     * @internal
+     */
+    public closeHandler(): void {
+        if (this.resolveShow) {
+            // If
+            // - the browser implements dialogs with the CloseWatcher API, and
+            // - the user presses ESC without first interacting with the drawer (e.g. clicking, scrolling),
+            // the cancel event is not fired and the drawer just closes.
+            this.notifyClosed(UserDismissed);
```
rajsite, line 104: Specifically for the drawer the assumption is that if the close handler is called with resolveShow still defined then the animation did not start right? I want to make sure that there are not animation event listeners, etc. that need to be cleaned up.

If that's true can you mention that assumption in the comment?

mollykreis (approved):

packages/nimble-components/src/dialog/styles.ts
m-akinc: The only way I know of to get the dialog in a focused state is to set `prevent-dismiss`, make sure nothing in the dialog has focus, then hit ESC.

We should add a focus-state visual test once I get my Innovation Day changes submitted to the Storybook pseudo-states addon. Until then, I think we're fine without it.

packages/nimble-components/src/dialog/index.ts
m-akinc: Went with `doResolveShow`

packages/nimble-components/src/dialog/index.ts
m-akinc: > or is this an internal error / guard against implementation errors?

Yes, that.

jattasNI (approved):

Issue #1479, opened 2023-09-01
Common issues caught in PR review


jattasNI:
## 🧹 Tech Debt

We'll use this issue to collect common issues found in PR review. One comment per issue, feel free to edit other people's comments with more info (does everyone have that permission?)

Once we've collected enough we can figure out what to do with the list. Maybe a code review checklist, maybe training material, maybe automated lint rules to catch this stuff for us.

jattasNI:
**Beachball change files**
1. wrong change type
2. description is vague or focuses on implementation rather than client impact

jattasNI:
**Testing**
1. Page objects should not expose `Element` types as they leak implementation detail and expose too much API surface area. Instead they should expose methods that access specific properties of the element and return primitive types.
2. All functionality should be covered by auto tests
3. Tests should be at the appropriate level (unit test or Chromatic test)
4. If something's not covered by auto tests, the PR description should include justification

jattasNI:
**Storybook**
1. Everything in the public API that we expect clients to use should be documented, including important events and methods. Documentation should follow the patterns of similar APIs on other components.

jattasNI:
**Styles**

1. Styles should be ordered according to our CSS conventions
2. Don't set size attributes to `inherit`
    See examples in [discussion 1](https://github.com/ni/nimble/pull/457#discussion_r834596232) and [discussion 2](https://github.com/ni/nimble/pull/1446#discussion_r1309430889)
3. For sizing, use appropriate size tokens for the right context (i.e. control height), but for sizes without a context aligned token, use a size from the 4px grid [size ramp](https://nimble.ni.dev/storybook/?path=/docs/tokens-size-ramp--docs).
   - Do not hard code specific pixel values, use the size ramp
   - If you need to add additional values to the size ramp, see the latest [naming guidance decision](https://github.com/ni/nimble/issues/1518#issuecomment-1834699333)
   - Yes the naming system for the fixed pixel size ramp is not great and uses the term "padding". It should still be used.
4. For configurable sizing, have the control respond to sizing based on normal CSS property configuration (i.e. control responds to setting height, width, etc) or expose a shadow part that can be sized.
   - Do not create new CSS custom properties to configure size, i.e. user has to set `--ni-nimble-my-custom-component-height`
   - Do not create custom attributes to control size, i.e. `<nimble-my-control size="small"></nimble-my-control>`
   - See following comment for more detail: https://github.com/ni/nimble/issues/610#issuecomment-4314554819
[...]

PR #1849, opened 2024-02-21
Fix bug with setting a title on spinner in Firefox

### LINKED ISSUE #1825: Setting title on nimble-spinner doesn't work correctly in FF
# 🐛 Bug Report

In Firefox, setting a `title` on the `nimble-spinner` doesn't work correctly. It is very difficult to get the title to appear by hovering over the spinner. It seems related to the animation because if you stop the animation, the title works correctly.

## 🤔 Expected Behavior

Hovering a `nimble-spinner` should show the `title`, if set.

## 😯 Current Behavior

Hovering a `nimble-spinner` with a `title` set doesn't show the title in Firefox.

## 💁 Possible Solution

Add an overlay element, such as a `div` over the spinner that the `title` can be set on that doesn't have animation.

## 🌍 Your Environment

Firefox 122.0.1

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1825 

## 👩‍💻 Implementation

I created a transparent overlay on top of the spinning bits in the spinner's template. This is enough to get the title to display correctly in Firefox and not flicker when the bits move.

## 🧪 Testing

Manually verified that the title now displays correctly in Firefox and still displays correctly in Edge.

## ✅ Checklist


mollykreis:
@msmithNI, can you buddy this PR for me?

msmithNI (approved):

packages/nimble-components/src/spinner/styles.ts
```diff
@@ -19,6 +19,13 @@ export const styles = css`
         aspect-ratio: 1 / 1;
     }
 
+    div.overlay {
+        z-index: 1;
+        position: absolute;
+        height: inherit;
```
rajsite, line 25: We shouldn't be configuring sizes as inherit, see: https://github.com/ni/nimble/issues/1479#issuecomment-1703313419

jattasNI (approved):
Validated titles work well in macOS FF (and Safari, though they worked well there already)

packages/nimble-components/src/spinner/styles.ts
mollykreis: I changed the display of the host to be `inline-grid` so that I could easily overlap the overlay and spinner container without having to use `position: absolute` and `height: inherit`.

packages/nimble-components/src/spinner/styles.ts
```diff
@@ -12,16 +12,22 @@ import { Theme } from '../theme-provider/types';
...
-    ${display('inline-flex')}
+    ${display('inline-grid')}
 
     :host {
         height: ${spinnerSmallHeight};
         aspect-ratio: 1 / 1;
     }
 
+    div.overlay {
+        z-index: 1;
```
m-akinc, line 23: Should use `ZIndexLevels.zIndex1` from `utilties/style/types.ts`.

rajsite (approved):

Issue #1850, opened 2024-02-21
Create ADR characterizing Nimble, Spright, and incubating components


m-akinc:
## 📌 User Story
Need to clarify the requirements for components in the Nimble repo, whether incubating or not, and for those in the Spright repo.

Issue #1682, opened 2023-11-21
Reduce duplication in Spright packages by single-sourcing and exporting from Nimble


jattasNI:
## 🧹 Tech Debt

This issue tracks parts of Nimble that would be useful in https://github.com/ni/spright but which we had to duplicate (or not use) since they weren't easily consumable.

1. `nimble-components/src/utilities/tests`. `matrix` and `storybook` and `hidden` are used by matrix tests

m-akinc:
Closing because Spright will now be hosted in the Nimble repo, and we will share code differently.

PR #1856, opened 2024-02-22
ADR for Nimble vs Spright

m-akinc:
# Pull Request

## 🤨 Rationale

Need to define the expectations for components that go in Nimble vs Spright vs ...
Fixes: #1850

## 👩‍💻 Implementation

Wrote ADR

## 🧪 Testing

N/A

## ✅ Checklist


docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Decision
+
+- Components with potential for reuse are strongly encouraged to go in Nimble.
+    - In some cases (mentioned in the preceeding section), the component may be added as "incubating". It should not be assumed that doing so will significantly shorten development time.
+- The [Spright repo](https://github.com/ni/spright) will be an alternative to Nimble. Ownership and review of Spright components falls to the contributing developers. It is up to those owners to define and enforce appropriate standards, but they should conform to a common set of Spright-wide minimum requirements (see table below).
```
jattasNI, line 31: We should add clear notes to the decision about what the infrastructure ownership will look like in Spright. I think the current proposal is something like this:
1. separate workflows (and npm workspaces?) per contributing team
2. Nimble team maintains repo-wide infrastructure like PR status requirements, repo docs, specs

And this adds some additional consequences
1. contributing teams responsible for uptaking new Nimble versions and other Renovate PRs
2. packages in Spright might depend on differing versions of Nimble

jattasNI (commented):
Some initial thoughts, will probably have more as I keep pondering

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,79 @@
...
+- Product- or service-specific components go in Spright, in a suitably scoped package named like `spright-components-stratus`.
+- TODO: When should (strictly Angular) components go in SystemLinkShared?
+
+The following table compares the requirements for developing a component in different libraries: 
+
+|                            | Approved Spec | Unit Tests | SB* Visual Tests | SB* API Docs | SB* Usage Docs | Approved** Visual Design | Approved** UX Design | Angular/Blazor Support | Proper a11y | Minimal Tech Debt | Mobile Support |
```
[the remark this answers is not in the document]
m-akinc: I'm going to assume that having a plan for a11y is part of the IxD, so marking IxD as required for incubating components should cover your concern.

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,79 @@
...
+-->
+- Spright's infrastructure increases the ease and consistency of UI component development (compared to creating them in an application).
+- Spright gives more visibility to components (than if within an application), increasing likelihood of reuse.
+- By using similar infrastructure, utilities, patterns, and rules as Nimble, Spright can make it easier to "graduate" a component to Nimble (than if written within an application).
+- Spright may provide a path to shorter development time (compared to Nimble) based on which requirements a team chooses to enforce, and the responsiveness/availability of their reviewers. This can give a team more control over the tradeoffs involved in hitting their deadlines.
+- Unless otherwise enforced by individual teams, products may ship with UI components that have:
+    - inconsistent, unapproved visual designs
+    - inconsistent, unapproved UX interactions
+    - missing or incorrect accessibility support
+
```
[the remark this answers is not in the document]
m-akinc: > Some of them might have different answers if contributed quickly by an internal team vs if on the Nimble team's roadmap.

I think that's always the case, except when a component is clearly product-specific. The sequence editor table is probably the only one that falls into that category. In all other cases, I think our default would be to build the thing in `nimble-components`, but depending on business/scheduling trade-offs, we might settle for incubating or Spright components.

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,79 @@
...
+## Consequences
+
+<!--
+This section describes the resulting context, after applying the decision. All
+consequences should be listed here, not just the "positive" ones. A particular
+decision may have positive, negative, and neutral consequences, but all of them
+affect the team and project in the future.
+-->
+- Spright's infrastructure increases the ease and consistency of UI component development (compared to creating them in an application).
```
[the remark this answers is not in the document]
m-akinc: I guess I was thinking it would be easier to create components in Spright, since it has everything already set up (including patterns, utilities, examples to follow, etc), but I wasn't considering that it's not really comparable to direct Angular/Blazor component development. I'm just going to remove this line.

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,79 @@
...
+## Conformance
+
+<!--
+This section describes the mechanisms that will be instituted to drive
+compliance with the decision. Mechanisms can be automated or manual. Part of the
+consideration for conformance should be the need for exceptions, the benefit of
+prevention versus detection, and the efficacy of automated versus manual
+processes compared to the cost of automation.
+-->
+- It will be up to UI component developers to decide where to put their components. Nimble code owners may redirect developers to Spright if a component does not belong in Nimble. There will be no formal oversight to make sure components that belong in Nimble go there rather than Spright, but guidance may be given informally at Stratus UI Working Group, etc.
```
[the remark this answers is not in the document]
m-akinc: But how do we enforce that specs get written, or that they get written with enough rigor? I don't think we'd want to make ourselves reviewers, so we can't really prevent people from copy/pasting some perfunctory spec. We'd have to trust developers to create sufficient specs, at which point it's more honest just to say specs are optional.

docs/nimble-spright-requirements-adr.md
m-akinc: I wonder if we should stick with one npm workspace so that every package keeps dependency versions in sync. But if we do that, I suppose it wouldn't make sense to have separate builds. If a team wants to upgrade a dependency, it could potentially break some other packages' builds. I'm not sure whether it's more important to keep dependencies synced or have independent builds.

jattasNI: In discussions with @rajsite we were really wary of the case where Nimble makes a breaking change (to a public API or an internal API that is being leveraged by a spright package). We didn't want to be responsible for uptaking those in another new repo, which means we would ask other teams to be responsible for it. But we don't trust that they would do so on a timely basis, which would mean the build would be broken for everyone if we have a monobuild. So our leaning was to move away from monobuild, which would be at the expense of keeping dependency versions in sync.

But we should think through the consequences of not keeping dependency versions in sync. Obviously we'd want versions in sync between an app->nimble and app->spright->nimble. But I think that comes for free from the way npm does dependency resolution in the app (assuming the deps are compatible). I haven't been able to think of consequences of spright libraries depending on different Nimble version. Do you have anything in mind?

docs/nimble-spright-requirements-adr.md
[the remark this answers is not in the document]
jattasNI: I'm more ok with Nimble owners reviewing spright contributors' specs than I am with us reviewing their implementations. It seems like a good time investment to make sure they get started with sane architecture / API design / code location.

But probably worth a team discussion about how much guidance we want to provide.

docs/nimble-spright-requirements-adr.md
rajsite: Jumping in the middle of the convo so may not be following context but for

> So our leaning was to move away from monobuild, which would be at the expense of keeping dependency versions in sync.

I don't know if I'm leaning one way or another. But I think ownership should be explicit and clear. If it's monolithic, frankly nimble would be the implicit owner so we make it explicit. If it's separate builds I think we risk lots of stagnation and inconsistencies building up and losing the ability to influence libraries moving forward (need to update Angular, blazor, foundation dependencies, etc). Also higher chance of more unconventional / wacky build configurations getting pushed through.

m-akinc: I've updated the ADR based on our conversation.

docs/nimble-spright-requirements-adr.md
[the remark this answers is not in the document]
m-akinc: Even if we are okay reviewing Spright specs, I would question whether we make it a requirement. I think a big selling point of Spright is that we can't/won't impede their progress, and requiring a spec review from us might significantly undercut that.

jattasNI (commented):
Just a couple minor suggestions I had from my initial scan of the document. The changes we discussed offline are more important.

docs/nimble-spright-requirements-adr.md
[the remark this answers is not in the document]
m-akinc: Decided in discussion that we do want specs.

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,88 @@
+# ADR: UI Component Contribution Requirements
```
jattasNI, line 1: What are the next steps once this is approved? We've talked about one or more PRs to create the spright packages in the monorepo. But I want to make sure we don't miss other steps. 

We could brainstorm it in an HLD but I think we might be able to get by with brainstorming a checklist, capturing that in one or more issues, and then working through it. If you agree, could you create the issue(s)?

I'll start with a brain dump here but if you want to separate this conversation from the ADR, we can:
1. Create packages in nimble repo (including integration with all the build/lint/test/publish steps). As Milan was hinting yesterday we might want to split this into a few PRs.
2. Configure CODEOWNERS for the new packages. I'd propose keeping the same owners as each corresponding Nimble package to start with.
3. Create storybook for spright components (I'm still not clear on the architecture here - does this require refactoring any Nimble storybook infrastructure?)
4. Link to spright storybook from relevant places like nimble.ni.dev and Nimble storybook
4. Do we want example apps too? I'd vote yes.
5. Do we want performance tests? I'd be fine to skip this for now.
6. Any changes to issue or PR templates? I can't think of any.
7. Any changes to Renovate? I think it'll work for spright automatically
10. Update repo-wide README / CONTRIBUTING to mention Spright
11. Update spec instructions and templates to consider Spright (e.g. annotate existing templates to say which sections are skippable for Spright contributions)
12. Is #1682 still needed or will those utilties now be accessible because they're in the monorepo?
13. create a demo component in spright and publicize in various forums
8. #1890 
9. Deprecate spright repo and turn off Renovate
15. ask Brandon to create a sweet new logo

Open to feedback on what's missing or not needed from this list!

jattasNI (approved):

docs/nimble-spright-requirements-adr.md
m-akinc: I have created the following issues which should track all this work:
- #1974 
- #1975 
- #1976 
- #1977 
- #1978 
- #1979 

The last is kind of a catch-all for small tasks that don't warrant a separate issue. I'm not quite sure what refactoring will be necessary to share code/configuration between Nimble and Spright. I don't think we need to work that out ahead of time, and we may end up creating tech debt items as needed.

docs/nimble-spright-requirements-adr.md
```diff
@@ -0,0 +1,95 @@
...
+      - Notably, Spright is intended for components leveraging the Nimble technology stack. Spright is *NOT* intended as a place to host arbitrary Angular or Blazor components that are not implemented as web components via FAST.
+- The Nimble and Spright packages will be part of the same build, use the same infrastructure, and share the same dependency version management. Spright packages will depend on the Nimble packages.
+- Code in Spright packages will be co-owned and reviewed by the Design System team and the contributing team.
+  - The goal of spright contributions is to be technically rigorous. To follow modern HTML, CSS, and TypeScript coding styles and follow Nimble's Contributing guidelines, CSS Best Practices, naming conventions, token usage, etc.
+  - Nimble developers remain as co-owners to enforce technical quality of implementation. Long-term components live in the Nimble monorepo and have to be patched to follow common patterns, updated with dependencies, supported on test infrastructure, and continually released. This is only mantainable when all code meets the same technical implementation bar.
```
rajsite, line 42: @m-akinc @jattasNI I pushed some changes to be crisper on what I understand as our alignment on spright ownership. Made changes to context and decision  but the consequences, risks, and conformance looked spot on. If we have concerns / misalignment on the updates lets chat Friday morning quickly to resolve.

m-akinc: LGTM

jattasNI (approved):

PR #1871, opened 2024-02-26
Fix drawer/select animation bug

### LINKED ISSUE #1097: Drawer for Grid Column Configuration "Jumps" in Safari
# 🐛 Bug Report

In Safari the drawer used to configure the visible grid columns and ordering _jumps_ when it is opened before settling in the correct position. This is not reproducible in Chrome and other drawers such as the user account drawer do not exhibit this behavior.

## 💻 Repro or Code Sample

See the video. 

https://user-images.githubusercontent.com/5823782/222743404-616aa413-c92b-48df-8046-3666832bd561.mov

## 🤔 Expected Behavior

Drawer opens smoothly.

## 😯 Current Behavior

As the drawer opens the entire components moves too far to center leaving a gap between the right edge and the browser where the content behind the drawer is visible.

## 💁 Possible Solution

🤷

## 🔦 Context

See something, say something.

## 🌍 Your Environment

macOS 12.6
Safari 16.0

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1097

## 👩‍💻 Implementation

It turns out that when the drawer is opening from the right, and the focused control in the drawer is a select, then the animation issue occurs. The root cause is that as part of the FAST Listbox's `focusAndScrollOptionIntoView` implementation, it queues `scrollIntoView` on the selected option in a `requestAnimationFrame` call. This interferes with the drawer's animation due to browser bugs ([Chromium](https://issues.chromium.org/issues/327554079), [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1882886)). 

We already override `focusAndScrollOptionIntoView` in our select, so I've just moved the call to `super.focusAndScrollOptionIntoView` into an `if (this.open)` block. This avoids (uneccessarily) calling `scrollIntoView` while the animation is in progress.

## 🧪 Testing

I have added a test to ensure that option scrolling still happens properly.

## ✅ Checklist


change/@ni-nimble-components-32486c60-9e10-4d2a-9c58-b437c26cafbb.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
m-akinc: It wasn't the appropriate fix. I put in some more debugging time, and I tracked it down to a `requestAnimationFrame` call in the FAST Listbox. I've started a [Discord post](https://discord.com/channels/449251231706251264/1212105703100121109) with FAST to see if they can give any details to help us find the right solution.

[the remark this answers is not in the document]
m-akinc: See updated description for new approach.

packages/nimble-components/src/select/index.ts
```diff
@@ -644,7 +644,11 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     }
 
     protected override focusAndScrollOptionIntoView(): void {
-        super.focusAndScrollOptionIntoView();
+        const optionToFocus = this.firstSelectedOption;
```
mollykreis, line 647: A few questions with this implementation:
- Any idea why calling `requestAnimationFrame` would break the drawer's animation? That seems like a browser bug unless there is something weird in the way FAST's DOM class is interacting with the browser's animations
- What about anything else that could cause `requestAnimationFrame`? Would anything calling `requestAnimationFrame` during the drawer's animation cause the animation to break? At the least, there is a call to `requestAnimationFrame` on line 653 below, but it seems unrealistic to avoid every possible call to `requestAnimationFrame` that could ever happen during a drawer's animation.

m-akinc: There probably is a very specific browser bug. I don't see this in the Playwright WebKit browser, and it only happens if `scrollIntoView` is called in `requestAnimationFrame` (though I've only tried a couple other functions), and it only happens if the drawer is sliding in from the right. I can't really guess at the reason/mechanism for what we see happening. I can try to make a simplified reproducing case to file a browser bug with, but I don't think I should spend too much more time on this.

mollykreis: I see. It wasn't clear to me from the description in the PR that the issue was the call to `scrollIntoView` within the `requestAnimationFrame` rather than the `requestAnimationFrame` call itself. Perhaps you could clarify that a little in the description.

This still seems like an odd fix for a drawer animation bug, but I agree that I'm not sure how much time we should spend on this since that combination of calls is probably unlikely to occur often.

packages/nimble-components/src/select/index.ts
```diff
@@ -644,7 +644,11 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
+        if (this.contains(document.activeElement) && optionToFocus !== null) {
```
mollykreis, line 648: Can all of this code get moved into the `this.open` if-statement? Is there a reason to try to focus and scroll an item into view when the element isn't even open?

mollykreis:
Seeing as @atmgrifter00 is the most familiar with this code, I'm going to add him to this PR as well.

packages/nimble-components/src/select/index.ts
m-akinc: Got a [reproducing Stackblitz](https://stackblitz.com/edit/stackblitz-starters-ydmp5r?file=index.html,styles.css,script.js). It is broken on both Chromium and Firefox (but in different ways). Not sure about Safari. I'm going to file bugs.

packages/nimble-components/src/select/index.ts
m-akinc: I'm not sure if it has an effect when `!this.open`. If I move it inside the `if`, the listbox seems to still be scrolled properly when I subsequently open the drop-down, so maybe it's okay. I'm a little hesitant to introduce another change to the pre-existing behavior without knowing that it doesn't have other side-effects, though.

m-akinc: After our conversation at stand-up, I'm convinced that it's better to just leave the original `super` call but move it inside the `this.open` if statement, like you suggested.

mollykreis:
Can you update the PR description again with the new implementation direction?

mollykreis (approved):

atmgrifter00 (approved):

packages/nimble-components/src/select/index.ts
```diff
@@ -644,8 +644,8 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
     }
 
     protected override focusAndScrollOptionIntoView(): void {
```
rajsite, line 646: Looks like mention listbox overrides the same method. Adding the same check seems reasonable there: https://github.com/ni/nimble/blame/2abc445190bb6b00dcb977b0f15a77e6d9ea3953/packages/nimble-components/src/rich-text/mention-listbox/index.ts#L257-L263

Looks like the foundation combobox overrides the same method. Seems like having the nimble combobox override the method and adding the same check before calling super would be reasonable there:
https://github.com/microsoft/fast/blob/97ef4d7a3880212bbb4f895459a047429a113472/packages/web-components/fast-foundation/src/combobox/combobox.ts#L340-L349

The check seems reasonable for all these places. Why would they try and focus something when not open

packages/nimble-components/src/combobox/index.ts
```diff
@@ -154,6 +154,19 @@ export class Combobox
         this.filteredOptions = enabledOptions;
     }
 
+    protected override focusAndScrollOptionIntoView(): void {
```
rajsite, line 157: > . Seems like having the nimble combobox override the method and adding the same check before calling super would be reasonable there:

```ts
protected override focusAndScrollOptionIntoView(): void {
  if(this.open) {
    super.focusAndScrollOptionIntoView()
  }
}
```
unless there is an important reason to duplicate the code slightly differently...

change/@ni-nimble-components-4411dde6-cf36-470b-9bfe-f527c5d48d1f.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @m-akinc  before resetting me on PRs
- please make sure the build passes
- please make sure new code follows normal expectations, ie. testing
it would really help reduce the number of re-reviews of PRs and save us all time

m-akinc: Ok, since build failures are usually something minor (e.g. lint issue, test needing update, or trivial logic error), I don't expect them to be relevant to reviewer feedback. But I'll try to remember to wait for a green check before resetting you.

> please make sure new code follows normal expectations, ie. testing

Do you want me to add tests for the combobox and mention listbox, or are you talking about something else? I didn't think new tests were necessary for those, but I can add some if you want.

rajsite: > I didn't think new tests were necessary for those

It's not clear to me why if it was valuable for select it was not valuable for other controls doing the same thing.

m-akinc: When I wrote the select test, my workaround had been to call `scrollIntoView` right after `focus` instead of delayed with `requestAnimationFrame`. The test was to ensure that scrolling still happened when calling those two functions back-to-back (because FAST's comment suggested they needed to be called in separate frames). Now that I've changed the approach to just short-circuting the focusing and scrolling when `!open`, my original motivation for the test is gone. The new approach should not affect the behavior when opening the listbox, so I didn't think I needed to test that. But there's no harm in having the tests, so I've added them.

packages/nimble-components/src/testing/async-helpers.ts
```diff
@@ -11,3 +11,10 @@ export const waitForUpdatesAsync = async (): Promise<void> => DOM.nextUpdate();
  * updates that resulted from DOM interactions like property or attribute changes.
  */
 export const processUpdates = (): void => DOM.processUpdates();
+
+/**
+ * Waits for a requested animation frame to occur.
+ */
+export const waitAnimationFrame = async (): Promise<void> => new Promise(resolve => {
+    requestAnimationFrame(() => resolve());
+});
```
rajsite, line 20: lets not put this in testing since that's a public api and this is for a very specific case. Instead could be in page objects or /patterns/tests or somewhere under the root /tests

change/@ni-nimble-components-4411dde6-cf36-470b-9bfe-f527c5d48d1f.json
rajsite: Gotcha, and what I'm looking for is consistency and avoiding one offs. So when I see a test only testing one code path where we have modified similar code paths (in this case three slightly different ways), I'm wondering why one is tested and the others aren't. So saying it was an unnecessary holdover might mean needing to pause and make another self check of the PR before resetting reviewers. In that scenario before resetting the test may have been removed.

If the test was removed then in that case if you showed and discussed in PR testing section that you found existing tests covering the open + focus behavior in fast foundation and we were calling the super behavior in all three cases and that are changes were minimal, it could make sense to skip adding extra tests. In this case we are only calling the base method in 2 of 3 places so I'd either ask to align (update the mention listbox to also call super) and then could agree that test coverage is probably good enough. Alternatively, in this case we are not aligned so having the test coverage is probably the right call. (And I also didn't check listbox test coverage in fast foundation so don't even know if we'd feel confident in the coverage).

Anyway the above reply isn't requiring an action or anything. Just trying to share the point of view I'm coming from to make our review cycles more efficient.

rajsite (approved):
Looks great!

rajsite (approved):

rajsite:
@jattasNI marking automerge because I think it has sufficient review

PR #1874, opened 2024-02-27
Empty icon mapping

mollykreis:
# Pull Request

## 🤨 Rationale

Grouping on an icon value isn't ideal if any of the cells don't have an icon. Specifically, the group row doesn't have a value associated with it and renders only a row count.

It would be nice if we could allow a client to leave a cell blank for certain field values but still have the group rows have a value associated with them. For example, have a column that shows either a spinner or nothing based on whether a task is in progress. When grouping, it would be nice to show a string, such as "Idle", associated with the records that have nothing in progress without cluttering the table with extra icons in cells.

This is the Future Work described in #1869

## 👩‍💻 Implementation

Updated the icon column to allow a `nimble-mapping-icon` to have an `undefined` icon. In that case, the cell renders nothing but the group row renders only the text associated with the mapping.

## 🧪 Testing

Manually tested in storybook.
Added new unit tests for having an `undefined` icon
Updated storybook/matrix tests

## ✅ Checklist


mollykreis:
@m-akinc, will you buddy these changes for me?

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
```diff
@@ -40,6 +40,12 @@ const simpleData = [
         lastName: 'Simpson',
         status: 'success',
         isChild: true
+    },
+    {
+        firstName: 'John',
+        lastName: 'Doe',
```
rajsite, line 46:
```suggestion
        firstName: 'Abbey',
        lastName: '?',
```

https://simpsonswiki.com/wiki/Abbey

packages/nimble-components/src/table-column/enum-base/models/mapping-icon-config.ts
```diff
@@ -1,28 +1,36 @@
...
-        aria-label="${x => x.text}"
-        severity="${x => x.severity}"
-        class="no-shrink"
-    >
-    </${icon}>`;
+const createIconTemplate = (
+    icon: string | undefined
+): ViewTemplate<IconView> => html`
+    ${when(
+        () => icon,
```
rajsite, line 13: In this case the evaluation of icon is static at the call of createIconTemplate, not dynamically determined (i.e. we don't check the context of what the template is bound to with `x => x.icon`, iit is always statically determined no matter what we are bound to `() => icon`).

Since it is static, can we instead avoid the extra runtime logic in the template?

```ts
const emptyTemplate = html<IconView>` `; // I think you at least need a space in there or you get a runtime error.

const createIconTemplate = (icon) => icon === undefined ? emptyTemplate : html<IconView>`
        <${icon}
            title="${x => x.text}"
            role="img"
            aria-label="${x => x.text}"
            severity="${x => x.severity}"
            class="no-shrink"
        >
        </${icon}>
```

Has the minor added benefit that all empty icons get to re-use the same template instance.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.spec.ts
```diff
@@ -25,7 +25,7 @@ interface SimpleTableRecord extends TableRecord {
 interface BasicIconMapping {
     key?: MappingKey;
     text?: string;
-    icon: string;
+    icon: string | undefined;
```
rajsite, line 28:
```suggestion
    icon?: string;
```

rajsite (approved):

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
```diff
@@ -40,6 +40,12 @@ const simpleData = [
         lastName: 'Simpson',
         status: 'success',
         isChild: true
+    },
+    {
+        firstName: 'Abbey',
```
jattasNI, line 45: I'd never heard of a Simpsons character named Abbey. In the process of searching for a [long-lost Simpsons relative](https://simpsons.fandom.com/wiki/Category:Long-lost_relatives) to replace this with, I learned that Homer [has a half-sister named Abbey](https://simpsons.fandom.com/wiki/Abbey). Well played! 👍

jattasNI (approved):

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
mollykreis: That was all @rajsite. He made that suggestion in a previous comment on this PR -- I personally know absolutely nothing about the Simpsons 😄

PR #1877, opened 2024-02-28
Add docs and test for table column header text inside a <span>

jattasNI:
# Pull Request

## 🤨 Rationale

[Some localization use cases](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/661277?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/asset-manager-lib/src/assets/asset-move-dialog/asset-move-dialog.component.html) require placing table header text inside a span rather than being text content of the header. This works fine today (text is rendered all caps, ellipsized when necessary, title set when necessary) but we don't document or test this case.

## 👩‍💻 Implementation

1. Add a note in storybook that this is allowed
2. Add a unit test to ensure title is set

## 🧪 Testing

1. New unit test for title.
2. Looked for unit / chromatic tests for ellipsis but didn't see any to copy and didn't think it was worth writing one for this case
3. Locally verified in Angular example app that header text in a span behaves identically to header text as text content

## ✅ Checklist


packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -215,6 +215,20 @@ describe('Table', () => {
...
 
+        it('sets title when header text is ellipsized in a span', async () => {
+            const headerContents = 'a very long value that should get ellipsized due to not fitting within the default header width';
+            await element.setData(simpleTableData);
+            await connect();
+            await waitForUpdatesAsync();
+            const span = document.createElement('span');
+            span.textContent = headerContents;
+            element.columns[0]!.textContent = '';
+            element.columns[0]!.appendChild(span);
```
rajsite, line 226: wanna show off to your friends? (almost definitely ignore)

```suggestion
            element.columns[0]!.replaceChildren(Object.assign(document.createElement('span'), {textContent: headerContents}))
```

change/@ni-nimble-components-4ec55901-eaa7-4bbb-a68e-0fdfd13c03a7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
rajsite, line 2:
```suggestion
  "type": "none",
```

packages/nimble-components/src/table/tests/table.spec.ts
rajsite: Ignoring :P

Issue #1873, opened 2024-02-27
Fixed width support for icon column


m-akinc:
## 😯 Problem to Solve

Need to support a fixed width for icon columns. There is a separate issue (#1332) about supporting fixed widths for columns more generally, but for the time being we are choosing to put that on hold and address this need for the icon column only.

Details from other issue:
- We will support fixed width only for icon columns with icon headers. 
- In this case space is not reserved for extra indicators / adorners (sort order, grouping, potential action menus)
- Only a header icon is shown, not header text
- Column maintains a fixed width the size of the icon
- The exact API is undecided, likely needs an HLD update

PR #1878, opened 2024-02-28
Fixed icon column width HLD update

mollykreis:
# Pull Request

## 🤨 Rationale

HLD update for #1873 

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/specs/table-column-width-hld.md
+++ b/packages/nimble-components/src/table/specs/table-column-width-hld.md
@@ -82,6 +82,12 @@ ColumnInternals<TColumnConfig> {
     @observable
     public minPixelWidth = defaultMinPixelWidth;
 
+    /**
+     * Whether or not this column can be interactively resized.
+     */
+    @observable
+    public resizingDisabled = false;
+
     ...
 }
 ```
@@ -150,10 +156,18 @@ At the moment there is no recognized use case for a pixel-width column that woul
 export class MyPixelWidthColumn : TableColumn<...> {
     public constructor() {
         super();
-        this.columnInternals.currentPixelWidth = 100;
+        this.columnInternals.resizingDisabled = true;
+        this.columnInternals.pixelWidth = 100;
+        this.columnInternals.minPixelWidth = 0;
     }
 }
 ```
+
+#### Hiding Header Indicators On Narrow Columns
+
+In some cases a column may not have space in its header for the sorting indicator or grouping indicator. For example, the icon column will be fixed width with enough space to render only a single icon. Therefore, there will not be space for a sorting indicator or grouping indicator next to the column's header icon.
+
+In this case, the sorting indicator and grouping indicator will be hidden in the column header. This scenario will be determined by comparing the column's `minPixelWidth` with `defaultMinPixelWidth`. If the `minPixelWidth` of the column is less than `defaultMinPixelWidth`, then the sorting and grouping indicator will automatically be hidden in the header.
 
 ### Implementation considerations
 
@@ -176,7 +190,9 @@ Because we will allow a horizontal scrollbar once the right-most column reaches 
 
 #### **Interactive visual states**
 
-Consult the [design document](https://xd.adobe.com/view/5b476816-dad1-4671-b20a-efe796631c72-0e14/) for details on the column divider appearance states, as well as the cursor appearance while hovering over a divider. If a user hovers over a column that is not resizable, no dividers will be shown.
+Consult the [design document](https://xd.adobe.com/view/5b476816-dad1-4671-b20a-efe796631c72-0e14/) for details on the column divider appearance states, as well as the cursor appearance while hovering over a divider.
+
+If a divider can never be dragged because only non-resizable columns exist to that column's left or right, the divider will be hidden. It is possible, however, for a column that is not resizable to have visible dividers if the column divider can be moved in a way such that other columns are resized.
 
 #### **Mobile considerations**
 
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -150,11 +156,19 @@ At the moment there is no recognized use case for a pixel-width column that woul
 export class MyPixelWidthColumn : TableColumn<...> {
     public constructor() {
         super();
-        this.columnInternals.currentPixelWidth = 100;
+        this.columnInternals.resizingDisabled = true;
+        this.columnInternals.pixelWidth = 100;
+        this.columnInternals.minPixelWidth = 0;
```
mollykreis, line 161: It seems a little verbose for a non-resizable column to have to set the `minPixelWidth` in addition to giving a `pixelWidth` and disabling resize, but I was hesitant to make too many assumptions in other parts of the code (particularly the `TableLayoutManager`) about what was true in a non-resizable column.

For example, should the `TableLayoutManager` assume that a non-resizable column has a `pixelWidth` set and is not a fractional width column? While that probably isn't an unreasonable assumption right now since we are only planning on having one non-resizable column and it will be a pixel-width column, I didn't like baking in assumptions that were not enforced anywhere.

Therefore, I decided that having a narrow fixed pixel column update the `minPixelWidth` was better than making those assumptions in the code. However, I'm open to any other suggestions anyone has.

mollykreis:

packages/nimble-components/src/table/specs/table-column-width-hld.md
atmgrifter00: Ultimately, I think all the `TableLayoutManager` should care about is whether it needs to resolve a column drag operation to a new `currentFractionalWidth` value or not. So, as I see it, the new `resizingDisabled` state is all that is needed to control appropriate divider visibility behavior for any given column. The state that controls how we [resolve `currentFractionalWidth` after a drag operation](https://github.com/ni/nimble/blob/937a555e4fd5d4928d64a5fb706d083acdc42aad/packages/nimble-components/src/table/models/table-layout-manager.ts#L245) is that column's `pixelWidth` state. If `pixelWidth` is not undefined, we do nothing, otherwise we do what is currently being done.

If that is an accurate understanding of things, then I would think we _don't_ need to require a `minPixelWidth` being set for a fixed width column. I easily could be missing something, and I'm happy to talk through it if needed.

mollykreis: After some discussion, we've landed on having the column update the `minPixelWidth` to an appropriate value if it supports being narrower than the default minimum width. Therefore, in this example, the `minPixelWidth` would be set to the same value as the actual width since the column isn't resizable.

Some of the rationale for this is:
- It keeps us from making assumptions in the `TableLayoutManager`, such as when we coerce `currentPixelWidth` to `minPixelWidth` if the `currentPixelWidth` is too small
- Some odd things could happen if the current width of the column is smaller than its minimum width
    - A resizable column in that scenario could get bigger but never be shrunk back to its original size
    - There's a number of places in the `TableLayoutManager` that assume today the `currentPixelWidth` will not be smaller than the column's `minPixelWidth`
- It doesn't seem unreasonable for a column that is explicitly trying to set its size or allow its size to be smaller than the default minimum width to have to set `minPixelWidth`.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/table-column-width-hld.md
```diff
@@ -150,11 +156,21 @@ At the moment there is no recognized use case for a pixel-width column that woul
 export class MyPixelWidthColumn : TableColumn<...> {
     public constructor() {
         super();
-        this.columnInternals.currentPixelWidth = 100;
+        this.columnInternals.resizingDisabled = true;
```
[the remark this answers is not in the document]
mollykreis: Those are good questions. I didn't really get into those details here since I didn't think they belonged in the Column Width HDL, but I should have included them somewhere. I'll update the Mapping Column HLD with that information.

rajsite (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -259,7 +258,22 @@ For icons, if multiple values map to the same icon, it is possible that sorting
 
 ### Sizing
 
-`nimble-table-column-icon` will support only a fixed width. We will introduce a new mixin for fixed-width support that exposes a `pixel-width` property. The default value will be the minimum supported by the table, which is still significantly larger than the width of an icon.
+`nimble-table-column-icon` will be a fixed pixel size (32px) and will not be resizable. The 32px fixed size allows room from a single icon along with left and right cell padding of 8px each.
+
+This has the following implications:
+
+-   The grouping indicator and sorting indicator will always be hidden on the icon column.
+-   A client is expected to only place an icon as the header content of an icon column.
```
jattasNI, line 266: Idle thoughts for implementation / future docs, no action needed now:
1. Some SLE grids like Systems currently include an icon column with no header content so we'll either want to support that (I assume it works already) or suggest they start using an icon header
2. We'll need to decide whether to enforce the icon header expectation. I'm inclined not to write any code for it, so if a client tries to put a long text header in an icon column they'll just get truncated text.

jattasNI (approved):

PR #1879, opened 2024-02-29
Create an adaptor for the old api dies inputs

Razvan1928:
# Pull Request

## 🤨 Rationale

We wanted to maintain the compatibility with the old api data
## 👩‍💻 Implementation
Created a class that gets the old api data as input and returns it as an apache arrow table.

## 🧪 Testing
With a few unit tests

## ✅ Checklist


packages/nimble-components/src/wafer-map/modules/wafer-map-convertor.ts
```diff
@@ -0,0 +1,48 @@
+import { Field, List, Table, Utf8, vectorFromArray } from 'apache-arrow';
+import type { WaferMapDie } from '../types';
+
+/**
+ * This class is used to convert old wafer map data to new wafer map data
+ */
+export class WaferMapConvertor {
```
atmgrifter00, line 7:
```suggestion
export class WaferMapConverter {
```

Razvan1928: renamed

packages/nimble-components/src/wafer-map/modules/wafer-map-convertor.ts
```diff
@@ -0,0 +1,48 @@
...
+    public colIndex: number[];
```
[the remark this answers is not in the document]
Razvan1928: I made the class stateless and the methods static, should be fine now. About the sentence 'Without seeing this being used as intended...' for now I don't think I get give you a quick demo, at best this is the table logged in the console after the conversion in the storybook
image
the new wafer map rendering capability is not done yet :(

packages/nimble-components/src/wafer-map/modules/wafer-map-converter.ts
```diff
@@ -0,0 +1,45 @@
+import { Field, List, Table, Utf8, vectorFromArray } from 'apache-arrow';
+import type { WaferMapDie, WaferMapLayerData } from '../types';
+
+/**
+ * This class is used to convert old wafer map data to new wafer map data
+ */
+export class WaferMapConverter {
```
atmgrifter00, line 7: So, again, I think the lack of context for how this class will be used is problematic. It's hard to understand the utility of this functionality even being encapsulated in a separate class. For instance, if there was a single place where we intended for either of these "conversions" to occur, then why can we not just make these private methods on the `WaferMap` component class?

Further, is there a use-case in mind for when `populateLayers` would be called outside of the context of the `toApacheTable` method? If not, then why is it public? These are the kinds of questions that are more easily answered when this implementation is accompanied with its usage in production.

Ultimately, while I think smaller PRs are a good goal, there are limits before they become problematic. At worst, you would be adding 4 files to a different PR, two of which are extremely light. My preference would be to bundle this PR with another.

Razvan1928: I discussed with @munteannatan and you are right, we will postpone the work on this task until the infrastructure will allow this API to be actually used. If we will continue on this pr or we will reuse the code in a different one will be decided later.

rajsite:
@munteannatan I'm going to close this draft PR for now since it has no activity for a while. Feel free to re-open when relevant.

PR #1885, opened 2024-03-01
Make renovate only update semver compatible versions

rajsite:
# Pull Request

### NPM

Switching the renovate behavior to just use semver compatible ranges for npm (in-range-only) and target all npm managed packages while doing so.

Benefits:
- All deps get managed instead of a select subset
- Single PR (for dev and devDeps) modifying the lockfile at a time
- Can likely be merged without any changes
- Future potential changes
  - We can potentially increase the rate for the PRs if they behave well

Cons:
- Not getting semver incompatible updates (i.e. updates to the package.json that require a beachball change)

Future potential changes:
  - Add back a select list of packages we are interested in getting latest changes beyond semver ranges
  - Run it less often, once a cycle or monthly
  - Make a manual task to manually go through and bump outdated semver ranges being careful to track typescript versions, etc., once a cycle?

### Nuget

- Think I followed the pattern npm was doing to disable generally and make a group for the nuget packages to reduce the number of separate PRs.

## 👩‍💻 Implementation

See above

## 🧪 Testing

Can try it out for a while

## ✅ Checklist

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -14,31 +14,10 @@
       "enabled": false
     },
     {
-      "groupName": "npm dependencies",
+      "groupName": "npm all dependencies",
       "matchManagers": ["npm"],
-      "rangeStrategy": "update-lockfile",
-      "matchDepTypes": ["dependencies"],
-      "matchPackagePatterns":[
-        "^@microsoft/fast",
-        "^@tanstack/",
-        "^@tiptap/",
-        "@types/markdown-it",
-        "comlink",
-        "d3-",
-        "prosemirror-"
-      ],
-      "enabled": true
-    },
-    {
-      "groupName": "npm dev dependencies",
-      "matchManagers": ["npm"],
-      "rangeStrategy": "update-lockfile",
-      "matchDepTypes": ["devDependencies"],
-      "matchPackagePatterns":[
-        "^@ni/eslint-config",
-        "beachball",
-        "storybook"
-      ],
+      "rangeStrategy": "in-range-only",
+      "matchDepTypes": ["dependencies", "devDependencies"],
       "enabled": true
     },
     {
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,22 +10,19 @@
   ],
   "packageRules": [
     {
-      "matchManagers": ["npm"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -10,43 +10,19 @@
   ],
   "packageRules": [
     {
-      "matchManagers": ["npm"],
+      "matchManagers": ["npm", "nuget"],
```
rajsite, line 13: I think the reason we end up with so many separate .net PRs is because they were not all disabled first?

rajsite, line 13: @jattasNI do youu think this missing is why renavate ended up making three separate .net related PRs? That was my theory

- https://github.com/ni/nimble/pull/1887
- https://github.com/ni/nimble/pull/1884
- https://github.com/ni/nimble/pull/1881

jattasNI: Yeah, partly. We enabled the `nuget` manager above and then left it with largely its default configuration. The default configuration includes [Renovate presets](https://docs.renovatebot.com/key-concepts/presets/) where they purposely batch [groups of related dependencies](https://docs.renovatebot.com/presets-monorepo/) together. The only change we made to that for nuget was to also group the bunit dependencies together since they didn't have a preset for that.

For the `npm` manager we took the opposite approach and disabled all the defaults here. That turns off all the presets and then we explicitly opted back in to our own configuration of groups.

I believe your current configuration will achieve the goal of one PR for all npm dependencies and devDependencies and another PR for all nuget dependencies. I'm less sure whether that's a good idea (Renovate's docs really encourage using their default settings) but I'm willing to try it out and see if we like it.

.github/renovate.json
```diff
@@ -10,43 +10,19 @@
   ],
```
jattasNI, line 10: Since it's currently scheduled to only run monthly, we should trigger a manual run after this PR completes (can do from #1537) so that we don't wait a month to discover any issues with this approach.

jattasNI (approved):

.github/renovate.json
rajsite: Makes sense, can close the existing PRs, clear the ignore lists (found the UI yesterday!), and run again

PR #1888, opened 2024-03-01
Green column divider

### LINKED ISSUE #1880: Header divider doesn't turn green on hover
# 🐛 Bug Report

According to the [visual design for the table](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=5661-6994&mode=design&t=APYAPbMUE13p5LMT-0), the divider between headers should turn green on hover rather than staying gray.

## 🤔 Expected Behavior

The column divider should change color when it is hovered.

## 😯 Current Behavior

The column divider does not change color when it is hovered.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #1880 

## 👩‍💻 Implementation

- A hovered column divider is now `borderHoverColor`
- The column divider that is being dragged is styled using the `active` class to be `borderHoverColor`
- The second column divider of the column being resized (assuming it exists) is styled using the `visible` class to be visible

## 🧪 Testing

- Update unit tests that check for the correct CSS classes to be applied to dividers during a drag operation

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/styles.ts
+++ b/packages/nimble-components/src/table/styles.ts
@@ -116,25 +116,18 @@ export const styles = css`
         height: ${controlSlimHeight};
         cursor: col-resize;
         position: absolute;
-    }
-
-    .column-divider:hover {
-        border-color: ${borderHoverColor};
-    }
-
-    .column-divider.visible {
-        display: block;
         z-index: ${ZIndexLevels.zIndex1};
     }
 
-    .column-divider.active {
+    .column-divider:hover,
+    .column-divider.divider-active {
         border-color: ${borderHoverColor};
     }
 
-    .header-container:hover .column-divider.left,
-    .header-container:hover .column-divider.right {
+    .column-divider.column-active,
+    .header-container:hover .column-divider,
+    .header-container:hover .column-divider {
         display: block;
-        z-index: ${ZIndexLevels.zIndex1};
     }
 
     .column-divider::before {
--- a/packages/nimble-components/src/table/styles.ts
+++ b/packages/nimble-components/src/table/styles.ts
@@ -125,7 +125,6 @@ export const styles = css`
     }
 
     .column-divider.column-active,
-    .header-container:hover .column-divider,
     .header-container:hover .column-divider {
         display: block;
     }
--- a/packages/nimble-components/src/table/styles.ts
+++ b/packages/nimble-components/src/table/styles.ts
@@ -9,7 +9,8 @@ import {
     controlSlimHeight,
     mediumPadding,
     standardPadding,
-    tableRowBorderColor
+    tableRowBorderColor,
+    borderHoverColor
 } from '../theme-provider/design-tokens';
 import { Theme } from '../theme-provider/types';
 import { hexToRgbaCssColor } from '../utilities/style/colors';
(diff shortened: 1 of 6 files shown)


mollykreis:

packages/nimble-components/src/table/template.ts
```diff
@@ -83,8 +83,14 @@ export const template = html<Table>`
                             ${repeat(x => x.visibleColumns, html<TableColumn, Table>`
                                 <div class="header-container">
                                     ${when((_, c) => c.index > 0, html<TableColumn, Table>`
-                                        <div class="column-divider left ${(_, c) => `${c.parent.layoutManager.activeColumnIndex === c.index ? 'active' : ''}`}" 
-                                             @mousedown="${(_, c) => c.parent.onLeftDividerMouseDown(c.event as MouseEvent, c.index)}">
+                                        <div
+                                            class="
+                                                column-divider
+                                                left
+                                                ${(_, c) => `${c.parent.layoutManager.activeColumnIndex === c.index ? 'visible' : ''}`}
```
m-akinc, line 90: So a divider can be visible without having the `visible` class, right? It looks like `visible` is only set when a divider is being dragged. That initially confused me. Could we change the class to something like `column-sizing` or `column-active`?

packages/nimble-components/src/table/styles.ts
```diff
@@ -117,6 +118,25 @@ export const styles = css`
...
+        display: block;
+        z-index: ${ZIndexLevels.zIndex1};
+    }
+
+    .column-divider.active {
+        border-color: ${borderHoverColor};
+    }
+
+    .header-container:hover .column-divider.left,
+    .header-container:hover .column-divider.right {
```
m-akinc, line 135: Can't `left` and `right` be omitted?:
```suggestion
    .header-container:hover .column-divider {
```

packages/nimble-components/src/table/styles.ts
```diff
@@ -117,6 +118,25 @@ export const styles = css`
         position: absolute;
     }
 
+    .column-divider:hover {
+        border-color: ${borderHoverColor};
+    }
+
+    .column-divider.visible {
+        display: block;
+        z-index: ${ZIndexLevels.zIndex1};
```
m-akinc, line 127: Maybe `z-index` should just go in the base `.column-divider` styling, so it doesn't have to be set in two places.

packages/nimble-components/src/table/styles.ts
```diff
@@ -117,6 +118,25 @@ export const styles = css`
         position: absolute;
     }
 
+    .column-divider:hover {
```
m-akinc, line 121: Any reason not to combine the selectors to avoid duplicating the styles? I.e.
```css
.column-divider:hover,
.column-divider.active { ... }

.column-divider.visible,
.header-container:hover .column-divider { ... }
```

m-akinc (approved):

packages/nimble-components/src/table/styles.ts
mollykreis: Yes, it seems like they can be omitted. I'm not sure why the styling wasn't originally written that way.

packages/nimble-components/src/table/template.ts
mollykreis: That's a good point. I've updated the two classes to be `column-active` and `divider-active`.

packages/nimble-components/src/table/styles.ts
```diff
@@ -115,6 +116,18 @@ export const styles = css`
...
+        z-index: ${ZIndexLevels.zIndex1};
+    }
+
+    .column-divider:hover,
+    .column-divider.divider-active {
+        border-color: ${borderHoverColor};
+    }
+
+    .column-divider.column-active,
+    .header-container:hover .column-divider,
```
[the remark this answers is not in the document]
mollykreis: ah...yes

rajsite (approved):

jattasNI (approved):

rajsite:
@mollykreis unmarking auutomerge to get the renoovate PRs through, will re-enable after

PR #1903, opened 2024-03-05
Workflow updates

rajsite:
# Pull Request

## 🤨 Rationale

- Updated `fetch-depth: 0` comment from being [needed by beachball](https://github.com/microsoft/beachball/pull/780) to being [needed by chromatic](https://github.com/ni/nimble/actions/runs/8164295790/job/22319288123#step:14:50).
- Updated renovate to make a single PR for npm updates that we can let update to latest.
  - Angular and it's related packages are ignored
    - `^@angular`, `ng-packagr`, `zone.js`, `typescript`
  - Pinned packages are ignored
    - `@microsoft/fast-foundation` and `@microsoft/fast-react-wrapper` due to [fast issue](https://github.com/microsoft/fast/pull/6906)
    - `comlink` as it is partially vendored into workers and needs a controlled and pinned update
    - `remark-gfm` as [updating breaks mdx](https://github.com/storybookjs/storybook/issues/24743#issuecomment-1799810754)
- Removed npm version specific install that was now downgrading the npm version accidentally.
- Removed file permission workaround [no longer needed to prevent warnings](https://github.com/actions/upload-pages-artifact/issues/45).
- Updated the beachball patch-package

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Relying on CI.

## ✅ Checklist


package.json
```diff
@@ -48,5 +48,12 @@
     "cross-env": "^7.0.3",
     "patch-package": "^6.4.7",
     "playwright": "1.40.0"
+  },
+  "overrides": {
```
rajsite, line 52: Note to self, need to recreate the lockfile before merging to main for overrides to work (merging updates from other branches breaks that update)

rajsite: I backed out the override. Seems to behave inconsistently after merges so don't want to deal with it.

.github/renovate.json
```diff
@@ -20,6 +20,30 @@
       "matchDepTypes": ["dependencies", "devDependencies"],
       "enabled": true
     },
+    {
+      "groupName": "npm dependencies bump",
+      "matchManagers": ["npm"],
+      "rangeStrategy": "update-lockfile",
+      "matchDepTypes": ["dependencies", "devDependencies"],
+      "matchPackagePatterns":[
+        "^@microsoft/fast",
```
rajsite, line 29: Actually maybe we should just exclude our manually managed / pinned packages and let everything else update. I'll switch to that.

.github/renovate.json
```diff
@@ -14,12 +14,30 @@
       "enabled": false
     },
     {
```
jattasNI, line 16: Wanna change the timing of the lock update to be more frequent now or save it for later?

jattasNI (approved):

.github/renovate.json
rajsite: Can save it for later. I want to see it make PRs we can just click merge on without headache a couple times before upping it. Might try it on depedency dashboard a couple times.

PR #1905, opened 2024-03-06
Dependency Updates March 2024

rajsite:
# Pull Request

## 🤨 Rationale

Gets all dependencies up-to-date (as reported by `npm outdated`) except the following:
- `@angular/*`, `zone.js`, `ng-packagr`, `typescript`: Coupled to current Angular version
- `@microsoft/fast-foundation` and `@microsoft/fast-react-wrapper`: Pinned due to https://github.com/microsoft/fast/pull/6906
- `playwright`: Latest version on nuget 1.41.2 not aligned with npm 1.42.1. There are some 1.42 beta releases on nuget so guess something is going on there. Didn't look into it more.
- `remark-gfm`: Updating it to latest caused storybook build errors and only recommendation seems to be [avoid upgrading `remark-gfm` to v4](https://github.com/storybookjs/storybook/issues/24743#issuecomment-1799810754).

## 👩‍💻 Implementation

Updated and fixed build & lint.

## 🧪 Testing

Relying on CI.

## ✅ Checklist


atmgrifter00 (approved):

packages/nimble-components/src/anchor-menu-item/specs/README.md
```diff
@@ -61,16 +61,10 @@ _The key elements of the component's public API surface:_
...
-        download="${x => x.download}"
-        href=${x => (x.disabled ? null : x.href)}
-        hreflang="${x => x.hreflang}"
-        ping="${x => x.ping}"
-        referrerpolicy="${x => x.referrerpolicy}"
-        rel="${x => x.rel}"
-        target="${x => x.target}"
-        type="${x => x.type}"
-    >
+    <a download="${x => x.download}" href="${x" ="">
```
jattasNI, line 64: This is harder to read and probably no longer syntactically correct. Can we revert it? I'm guessing it came from prettier?

packages/nimble-components/src/switch/styles.ts
```diff
@@ -107,8 +107,9 @@ export const styles = css`
         border-radius: calc(var(--ni-private-switch-indicator-size) / 2);
         margin: calc(
             calc(
```
jattasNI, line 109: Really prettier, this is better? 😆

But probably not worth fighting it.

packages/nimble-components/src/utilities/tests/matrix.ts
```diff
@@ -21,7 +21,7 @@ export const sharedMatrixParameters = () => ({
             hidden: true
         }
     }
-} as const);
+}) as const;
```
jattasNI, line 24: Can we just remove the outer `()`?

jattasNI: Same for `storybook.ts`

jattasNI (approved):
I don't think you need to wait for any other reviewers once you address my minor feedback.

packages/nimble-components/src/utilities/tests/matrix.ts
rajsite: Nope, you can't, it's an arrow function returning an object, have to wrap as an expression
Following is ambiguous:
`() => {test:2}`
Make it an expression:
`() => ({test:2})`

packages/nimble-components/src/switch/styles.ts
rajsite: Yea, think I'm going to disable prettier on html (something like 1/3rd of all html templates have prettier disable) but for CSS it hasn't been completely unhelpful yet. Though yea, don't particularly like this CSS change but don't feel like fighting it.

packages/nimble-components/src/anchor-menu-item/specs/README.md
rajsite: It came from prettier but it's also spec text and not production code so don't feel like it's worth the effort. Probably an example of too much implementation detail in spec anyway.

PR #1907, opened 2024-03-06
Allow icon column to be configured as fixed width

### LINKED ISSUE #1873: Fixed width support for icon column

mollykreis:
# Pull Request

## 🤨 Rationale

This change is part of #1873 and follows the decisions made in #1992

This change adds the ability to configure the `width-mode` of the icon column. When set to `iconSize`, the column is a fixed width of 32px, cannot be resized, and hides its sorting & grouping indicators in the header.

## 👩‍💻 Implementation

There were a number of changes that needed to be made within nimble to facilitate this change.

1. Column Internals
    - Add boolean properties of `resizingDisabled` and `hideHeaderIndicators`
1. Table Update Tracker
    - Mark `columnWidths` as being changed when `resizingDisabled` changes in `columnInternals`
1. Table Layout Manager
    - Update to deal with columns that have `resizingDisabled === true`
1. Table
    - Add logic to hide column dividers that can never be dragged to the left or to the right. This is accomplished by adding a `draggable` class to the dividers that can be dragged and making the presence of that class a condition in the CSS for showing the divider.
    - Add logic to set `indicators-hidden` on table columns where `hideHeaderIndicators === true`
1. Table Header
    - Add a boolean attribute of `indicators-hidden` that, when `true`, hides the sorting & grouping indicators in the column header.
1. Icon Column
    - Add `width-mode` attribute. When the attribute is changed, `columnInternals` is updated with appropriate values for `resizingDisabled`, `hideHeaderIndicators`, `minPixelWidth`, and `pixelWidth`.

## 🧪 Testing

Manually tested in storybook
Updated & added unit tests

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
```diff
@@ -76,17 +78,22 @@ export const iconColumn: StoryObj<IconColumnTableArgs> = {
...
+                <${iconXmarkCheckTag} title="Is a Simpson"></${iconXmarkCheckTag}>
+
                 <${mappingIconTag} key="fail" icon="${iconXmarkTag}" severity="error" text="Not a Simpson"></${mappingIconTag}>
                 <${mappingIconTag} key="success" icon="${iconCheckLargeTag}" severity="success" text="Is a Simpson"></${mappingIconTag}>
                 <${mappingSpinnerTag} key="calculating" text="Calculating"></${mappingSpinnerTag}>
                 <${mappingIconTag} key="unknown" text="Unknown"></${mappingIconTag}>
             </${tableColumnIconTag}>
             <${tableColumnIconTag} field-name="isChild" key-type="boolean">
-                Is Child
+                <${iconChartDiagramChildFocusTag} title="Is child"></${iconChartDiagramChildFocusTag}>
```
atmgrifter00, line 95: I think for maximum humor, we should group by `isChild`, properly mark Homer as "Is a Simpson", but mark Homer as "Calculating" for "Is a Child".

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -480,6 +533,68 @@ describe('Table Interactive Column Sizing', () => {
             });
         });
 
+        const resizingDisabledDividerDraggabilityTests = [
+            {
+                name: 'all dividers are draggable when no columns having resizing disabled',
```
atmgrifter00, line 538:
```suggestion
                name: 'all dividers are draggable when no columns have resizing disabled',
```

packages/nimble-components/src/table-column/icon/index.ts
```diff
@@ -20,6 +19,10 @@ import type { MappingConfig } from '../enum-base/models/mapping-config';
 import { MappingIconConfig } from '../enum-base/models/mapping-icon-config';
 import { MappingSpinnerConfig } from '../enum-base/models/mapping-spinner-config';
 
+const cellPadding = 8;
```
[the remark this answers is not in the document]
mollykreis: That's a good question. I did it this way since that was the existing pattern for defining the default minimum column width:
https://github.com/ni/nimble/blob/2c782847534b9af993d7e97e41d773f150b60793/packages/nimble-components/src/table-column/base/types.ts#L47-L60

I think the options are:
- leave this as-is
- add comments that these are the default values for `mediumPadding` and `iconSize`
- export the values from `table-column/base/types` (maybe with added comments there?) so that we are limiting the places we reference the default token values, perhaps with some values renamed to be more generic (e.g. `iconSize` rather than having both `groupIconSize` and `sortIconSize`).

Thoughts on a preferred approach? I'm leaning towards exporting the values from the existing shared `types` file. I originally had shied away from that because I didn't want to make this PR too big, but I don't think that change would actually add too much extra code.

packages/nimble-components/src/table/template.ts
```diff
@@ -109,6 +112,7 @@ export const template = html<Table>`
                                                 right
                                                 ${(_, c) => `${c.parent.layoutManager.activeColumnIndex === c.index ? 'column-active' : ''}`}
                                                 ${(_, c) => `${c.parent.layoutManager.activeColumnDivider === c.parent.getRightDividerIndex(c.index) ? 'divider-active' : ''}`}
+                                                ${(_, c) => `${c.parent.layoutManager.hasResizableColumnToLeft(c.index, c.parent.visibleColumns) ? 'draggable' : ''}`}
```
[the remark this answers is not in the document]
mollykreis: We actually don't care if a column to the right is resizable because the columns to the right will be pushed to the right if they can't be shrunk.

packages/nimble-components/src/table/models/table-update-tracker.ts
```diff
@@ -132,7 +132,8 @@ export class TableUpdateTracker<
                 changedColumnProperty,
                 'currentFractionalWidth',
                 'currentPixelWidth',
-                'minPixelWidth'
+                'minPixelWidth',
+                'resizingDisabled'
```
atmgrifter00, line 136: While I realize this is fairly innocuous, I'm curious if there's a practical reason why we have to track this change? I would have expected we only need to track property changes that can result in a visual width change of the columns.

packages/nimble-components/src/table-column/icon/index.ts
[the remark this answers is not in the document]
atmgrifter00: I think I also like the exporting idea. I'm not too worried about PR size. It was pretty easy to get through.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
mollykreis: I'm not against that change, but I'd rather not include that in this PR. Feel free to send one my way if you want to make those tweaks to this story.

atmgrifter00 (approved):

change/@ni-nimble-components-248b2e34-5be4-492b-8512-1e5226e10ca4.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Now that I've have a chance to play with this in storybook, I'm finding our UX decisions cause some clunky side effects. These are just my opinions but could we double check with @fredvisser that we're happy with these?

1. Only icons in the header. Icon headers are less effective than text headers to convey at a glance what the column represents. In our storybook examples I find myself hovering over the header to see the title. It's more annoying for columns where we don't have a great icon ("Is a Simpson") but I imagine that could happen for real app use cases (e.g. several columns displaying similar concepts). I'd lean towards allowing text headers and wider fixed widths (which opens a can of worms for how we/clients compute that width)
2. Narrow column results in lots of horizontal whitespace. In the screenshot below the icons end up really far from other columns. Obviously setting a smaller table width or adding more columns would help so maybe this is just an issue with our contrived example. But in this case I find myself wanting to bring the icon columns closer to the other columns and put the whitespace on the right edge.
4. No feedback in header when column is interactively sorted. I mentioned this one at your demo. I can live with it but it still bugs me so I'm mentioning it again. If we end up allowing icon columns to be wider we could bring back the sort icons when there's space for them.

image

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -138,27 +169,80 @@ export class TableLayoutManager<TData extends TableRecord> {
...
+     * `columnIndex` and moving to the left. If no resizable column
+     * is found, returns -1.
+     */
+    private getFirstLeftResizableColumnIndex(
+        columnIndex: number,
+        visibleColumns = this.visibleColumns
+    ): number {
+        for (let i = columnIndex; i >= 0; i--) {
+            const column = visibleColumns[i];
+            if (!column) {
```
jattasNI, line 204: How could `column` be falsy here? I don't think we can index outside the array; is there a case where columns in the array are undefined? Or are we just forced to handle this case because TypeScript can't determine whether it'll always be valid?

msmithNI (approved):

jattasNI (approved):

packages/nimble-components/src/table/models/table-layout-manager.ts
mollykreis: It really shouldn't be. TypeScript forces us to handle that case, but technically a caller could pass a `columnIndex` that is greater than the number of visible columns.

change/@ni-nimble-components-248b2e34-5be4-492b-8512-1e5226e10ca4.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "major",
+  "comment": "Make the icon column fixed size and remove its fractional-width and min-pixel-width configuration",
```
jattasNI, line 3: nit: could mention the new recommendation about header content for icon columns too

change/@ni-nimble-angular-6221e053-9114-4da5-a449-c8e19b91db59.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
jattasNI, line 2: Usual reminder to think through communication plan and who does client updates (if needed) for the breaking change.

jattasNI (approved):

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
```diff
@@ -76,17 +78,22 @@ export const iconColumn: StoryObj<IconColumnTableArgs> = {
             style="${isChromatic() ? '--ni-private-spinner-animation-play-state:paused' : ''}"
         >
             <${tableColumnTextTag} field-name="firstName" >
-                Name
+                First name
+            </${tableColumnTextTag}>
+            <${tableColumnTextTag} field-name="lastName" >
+                Last name
             </${tableColumnTextTag}>
```
rajsite, line 85: Can we make Last Name the last column and give it a large fractional width?

packages/nimble-components/src/table-column/icon/tests/table-column-icon.mdx
```diff
@@ -12,6 +12,8 @@ import { spinnerTag } from '../../../spinner';
 
 The <Tag name={tableColumnIconTag}/> column renders string, number, or boolean values as a Nimble icon or <Tag name={spinnerTag}/> in the <Tag name={tableTag}/>.
 
+The column is a fixed size and cannot be resized by a user. The column is wide enough for a single icon, so the column's header should be a nimble icon that has a `title` configured on it for accessibility.
+
```
rajsite, line 16: Another thing we discussed is usage guidance on avoiding having the icon column be the last column in a table.

packages/nimble-components/src/table/template.ts
```diff
@@ -97,6 +99,7 @@ export const template = html<Table>`
                                             class="header"
                                             sort-direction="${x => (typeof x.columnInternals.currentSortIndex === 'number' ? x.columnInternals.currentSortDirection : TableColumnSortDirection.none)}"
                                             ?first-sorted-column="${(x, c) => x === c.parent.firstSortedColumn}"
+                                            ?indicators-hidden="${x => x.columnInternals.minPixelWidth < defaultMinPixelWidth}"
```
rajsite, line 102: I'm iffy about this change. There is now some magical value table column users are unaware of that cause the indicators to disappear. Lets say I want my table columns to be flexible, the fractional width will be large but I set the minPixelWidth small to let the table user shrink the width as much as they want. So normally the column is large but the user interactively resizes smaller. Now unclearly at some point I configure all the status icons disappear.

mollykreis:
There is more on-going discussion about the exact behavior we want here, including whether or not we really want to force clients to make all their icon columns a fixed 32px. I'm moving this PR back to draft until that discussion is resolved.

change/@ni-nimble-components-248b2e34-5be4-492b-8512-1e5226e10ca4.json
mollykreis: This doesn't apply anymore to the change file description. The documentation for `width-mode` states that `iconSize` should only be used when the header and all mappings contain only icons.

change/@ni-nimble-angular-6221e053-9114-4da5-a449-c8e19b91db59.json
mollykreis: Not applicable anymore since this is no longer a breaking change.

angular-workspace/projects/example-client-app/src/app/customapp/customapp.component.html
```diff
@@ -266,7 +266,7 @@
                         key="unknown"
                         text="Unknown">
                     </nimble-mapping-spinner>
-                    Result
+                    <nimble-icon-check></nimble-icon-check>
```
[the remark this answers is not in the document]
mollykreis: No longer applicable.

packages/nimble-blazor/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -249,7 +249,7 @@
                         Key=@("unknown")
                         Text="Unknown">
                     </NimbleMappingSpinner>
-                    Result
+                    <NimbleIconCheck></NimbleIconCheck>
```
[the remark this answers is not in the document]
mollykreis: No longer applicable.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.stories.ts
mollykreis: There is now an additional column ("gender") that is the last column. I didn't adjust its fractional width, but I think the example shows the behavior adequately. Let me know if you disagree with the new example and still think the fractional widths should be updated.

packages/nimble-components/src/table/template.ts
mollykreis: Updated the logic so that a column explicitly tells the header (though `columnInternals`) that it wants the sort/group indicators to be hidden.

mollykreis:
@atmgrifter00, can you re-buddy this PR for me now that there has been a significant change in the direction of the feature?

atmgrifter00 (approved):

packages/nimble-components/src/table-column/base/types.ts
```diff
@@ -44,19 +44,18 @@ export type TableColumnSortOperation =
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface TableColumnValidity extends ValidityObject {}
 
-const groupIconSize = 16;
-const sortIconSize = 16;
-const spacing = 8;
-const menuDropdownSize = 24;
+export const columnIconSize = 16; // `iconSize` token
+export const columnSpacing = 8; // `mediumPadding` token
+const menuDropdownSize = 24; // `controlSlimHeight` token
```
rajsite, line 49: I'm a tad nervous about exporting all these in-case folks start depending on these unexpectedly. The defaultMinPixelWidth calculated like this has been a bit of a workaround / hack.

Can we instead export something like a `columnHiddenIndicatorsWidth` constant and move the calculation in here instead of exporting all the partial aliases to exisiting tokens?

packages/nimble-components/src/table-column/icon/index.ts
```diff
@@ -21,6 +27,9 @@ import { MappingIconConfig } from '../enum-base/models/mapping-icon-config';
 import { MappingSpinnerConfig } from '../enum-base/models/mapping-spinner-config';
 import { MappingText } from '../../mapping/text';
 import { MappingTextConfig } from '../enum-base/models/mapping-text-config';
+import { TableColumnMappingWidthMode } from './types';
+
+const iconSizeWidth = columnIconSize + 2 * columnSpacing;
```
rajsite, line 32: When this calculation is moved out can you give a name to the `2`. It's not immediately clear from source what that is for. Just extra padding?

packages/nimble-components/src/table-column/icon/tests/table-column-icon-matrix.stories.ts
```diff
@@ -80,6 +82,19 @@ const component = (): ViewTemplate => html`
             <${mappingIconTag} key="1" text="One" icon="${iconCheckTag}" severity="warning"></${mappingIconTag}>
             <${mappingIconTag} key="2" text="Two" icon="${iconCheckTag}" severity="error"></${mappingIconTag}>
         </${tableColumnIconTag}>
+        <${tableColumnIconTag}
+            field-name="code"
+            key-type="number"
+            width-mode="${_ => TableColumnMappingWidthMode.iconSize}"
```
rajsite, line 88: Shouldn't need a binding for just a constant

```suggestion
            width-mode="${TableColumnMappingWidthMode.iconSize}"
```

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -138,27 +169,80 @@ export class TableLayoutManager<TData extends TableRecord> {
...
+                return -1;
+            }
+            if (!column.columnInternals.resizingDisabled) {
```
rajsite, line 207: I didn't try and trace the logic too deeply but one thing that is not clear to me in this class is when we access the cached state vs the live state of these properties. My initial thought is that after caching all these code paths should generally access the cache value instead of the current columnInternals value. If that's the case can we double check the columnIInternals accesses in the file? If not curious how to make it clear for future authors.

msmithNI (approved):

packages/nimble-components/src/table/models/table-layout-manager.ts
mollykreis: I've updated the code a bit to hopefully make things clearer. Now, the cached state only includes the initial width values, which are not available in the column/columnInternals itself. Any other state, such as minimum width or resizing disabled is read from the column.

rajsite: That helps! 👍

rajsite (approved):

Issue #350, opened 2022-02-11
Placeholder for nimble-select when nothing is selected


gokulakrishnan-ni:
## 😯 Problem to Solve

Currently, the nimble dropdown displays the first item in the list as the default selection but the corresponding form control value doesn't reflect the displayed value.

## 💁 Proposed Solution

By default, if the form control value
1. doesn't match any of the values in the dropdown list, the dropdown shall display a placeholder which is passed as a property value
2. matches any of the value in the list, the corresponding item shall be displayed as the selected value

jattasNI:
Note that Blazor support is tracked separately #1952 so I think we can close this once we're satisfied with the web component and Angular integration.

fredvisser:
Looks great in Storybook and example Angular app!

Issue #791, opened 2022-10-19
Sectioned Select


TJ-G:
## 😯 Problem to Solve
I don't see anything tagged as 'client-request' or 'enhancement' that matches this, but please feel free to close if I missed an existing request similar to this.

When a user selects a notebook (say from the analyze slide-out in the files grid), they see all notebooks published to workspaces they have access to. One key thing here is that notebooks are published to workspaces for access control. Notebooks with the same name may be published to different workspaces and, currently, will show up in the dropdown without any way to differentiate which one is in what workspace.

## 💁 Proposed Solution
I'd like the ability to section/segment/group the items in a combo-box and give each section a header. This would let me group the notebooks by what workspace they're in, and let the user know the name of that workspace by using that as the header for each section.

It looks like we can sort of get this behavior by inserting disabled items as the header, but that means we can't actually have disabled items in the list, and they don't really stand out as a 'header' and are hard to read.

![image](https://user-images.githubusercontent.com/7433612/196754153-c30cefd5-6d40-4639-b265-9d4214c5c87a.png)

m-akinc:
You can actually include elements other than `nimble-list-option` in the combobox's content, e.g. `<h3>Workspace 1</h3>`. Does that fully address your use case?

TJ-G:
Oh, interesting... yes, I think it would, especially if we can stylize them a bit.

Thanks for letting me know that's already possible!

rajsite:
> You can actually include elements other than `nimble-list-option` in the combobox's content, e.g. `<h3>Workspace 1</h3>`. Does that fully address your use case?

@m-akinc @atmgrifter00

I don't see any stories or tests describing that use case. Is it tested (maybe I'm missing it)? Do we know how things like tab order and keyboard interactions, etc behave? What kind of content are we supporting, just decorative or interactive (can you dump a textbox, or a nimble-button in there)?

If we say it's supported we need testing and user guidance of what is supported. In addition we need documentation and examples in storybook.

Generally speaking we do not support placing arbitrary content in slots. Only certain use-cases are supported and we should add tests if we want to say we support additional use cases.

jattasNI:
One other workaround available today would be to append additional information to each item's text. e.g. "Notebook 1 (Default Workspace)"

jattasNI:
[...]

PR #1912, opened 2024-03-07
Angular Select placeholder integration

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #350 

## 👩‍💻 Implementation

1. Expose the necessary Angular attributes for the `NimbleListOptionDirection` to support the placeholder use-case
2. Changes to nimble-components to address a problem discovered during early integration attempts for a particular SLE usage. 
    - It was discovered that there are scenarios that will add `nimble-list-options` dynamically to the Angular template contents, followed immediately with setting a form value (that the `Select` was bound to) to the dynamically added option's value. This results in situation where the `SelectControlValueAccessor` processes the set value (which for the `Select` _requires_ that option to be present on the web component side) prior to the `Select` adding the option to its set, which would result in the `Select` view ignoring that value and setting it to empty string instead.
    - The solution to this is to leverage the `connectedCallback` override of the `NimbleListOption` to allow it to have its parent, through an available interface if implemented, process it. All the `Select` does, is if the option isn't already part of its `options`, it simply pushes it on to it. There is no concern with this interfering with the normal process of the option being added as part of the `slottedOptionsChanged` callback, that will simply result in overwriting the options (which is fine).
    - Applied this solution to the combobox too
3. Exported the `SelectPageObject` from Angular, so that we can also leverage it in SLE as needed. Removed a lot of unnecessary `async` aspects of the `SelectPageObject` public methods.
4. Updated example app to use the feature and removed extra items from select to align with Blazor app and avoid Lighthouse performance degradation.

## 🧪 Testing

Added unit tests in nimble-components for the new internal interface the `Select` is using. Added the boilerplate unit tests in Angular for the new attributes. Also added a couple of Angular tests one of which is representative of the original failing scenario from the integration effort (and the other is more for completion).

## ✅ Checklist


mollykreis (approved):

packages/nimble-components/src/select/index.ts
```diff
@@ -668,6 +670,17 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
 
+    /**
+     * @internal
+     */
+    public registerOption(option: ListOption): void {
+        if (this.options.includes(option)) {
+            return;
+        }
+
+        this.options.push(option);
```
rajsite, line 681: Is there an ordering assumption with options? If so, how can we guarantee there isn't an issue here?

atmgrifter00: Yeah, this could result in an order difference between the DOM and what the `Select` displays, which could corrupt the state of things. Perhaps the more rigorous approach is for this implementation to do a `querySelectorAll` on its parent, and find where it's supposed to be inserted into the options. I'll look into this. If I can't find a good solution (and verify that things can get in a bad state), the fallback will be to do a conditional (if possible) `querySelectorAll` when we set the value.

angular-workspace/projects/ni/nimble-angular/src/directives/select/testing/select.pageobject.ts
```diff
@@ -0,0 +1,3 @@
+import { SelectPageObject } from '@ni/nimble-components/dist/esm/select/testing/select.pageobject';
```
jattasNI, line 1: We want to export page objects from a separate entry point so that clients can enforce that they only be used from test code, not production code. 

The way we've done this for table and rich text is to create an entry point for the component and an entry point for the page object. If we follow that pattern we'd have to move the select which would be a breaking change. We want to do this eventually but maybe now's not the time?

An alternative might be to create the select page object in the correct long term location but leave the select where it is for now. But I'm not sure if that'll cause problems if things in the main nimble-angular entry point try to use it.

packages/nimble-components/src/patterns/dropdown/types.ts
```diff
@@ -22,3 +23,17 @@ export const DropdownAppearance = {
...
+ * @internal
+ *
+ * This interface is used to register options with their parent once their
+ * 'connectedCallback' method is run. This allows for the "owner", like the
+ * Select, to have its value set to that newly registered option earlier than it
+ * might otherwise in certain situations. One such scenario is in an Angular
+ * reactive form, where the form value is set to an option immediately after
+ * dynamically adding it.
+ */
+export interface DropdownOwner {
```
jattasNI, line 37: I think I'd go with `ListOptionOwner` since that's the type that the method in the interface takes.

angular-workspace/projects/ni/nimble-angular/src/directives/select/testing/select.pageobject.ts
atmgrifter00: I was able to create the `select/testing` entry point and use the page object from the entry point in the existing select control value accessor test.

change/@ni-nimble-angular-07596cf2-5025-4bb7-8744-b44330dd560a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add selected and hidden attributes to NimbleListOptionDirective",
```
jattasNI, line 3: Could mention the page object here too

packages/nimble-components/src/list-option/index.ts
```diff
@@ -51,6 +52,17 @@ export class ListOption extends FoundationListboxOption {
...
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        if (this.isDropdownOwner(this.parentElement)) {
+            this.parentElement.registerOption(this);
+        }
+    }
+
+    private isDropdownOwner(parent: unknown): parent is ListOptionOwner {
+        return typeof (parent as ListOptionOwner).registerOption === 'function';
+    }
```
jattasNI, line 65:
```suggestion
    public override connectedCallback(): void {
        super.connectedCallback();
        if (this.isListOptionOwner(this.parentElement)) {
            this.parentElement.registerOption(this);
        }
    }

    private isListOptionOwner(parent: unknown): parent is ListOptionOwner {
        return typeof (parent as ListOptionOwner).registerOption === 'function';
    }
```

packages/nimble-components/src/list-option/index.ts
```diff
@@ -51,6 +52,17 @@ export class ListOption extends FoundationListboxOption {
...
+
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        if (this.isDropdownOwner(this.parentElement)) {
+            this.parentElement.registerOption(this);
+        }
+    }
+
+    private isDropdownOwner(parent: unknown): parent is ListOptionOwner {
+        return typeof (parent as ListOptionOwner).registerOption === 'function';
```
jattasNI, line 64: Some questions about types:
1. Why is `parent` of `unknown` type rather than `HTMLElement`? Isn't this only called with `this.parentElement: HTMLElement`?
1. Do we need to handle the case where `parent` is undefined? Or since this is called from connectedCallback we know it'll always be valid?

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -281,16 +281,24 @@ describe('Select', () => {
...
+        const registerOptionSpy = spyOn(
+            element,
+            'registerOption'
+        ).and.callThrough();
+        registerOptionSpy.calls.reset();
+        element.insertBefore(newOption, element.options[0]!);
+
+        expect(registerOptionSpy.calls.count()).toBe(1);
+        expect(element.options).toContain(newOption);
+        expect(element.selectedIndex).toBe(0);
```
rajsite, line 298: From offline discussion we should make a tech debt issue to track this assertion as a known issue, that synchronous updates do not behave like native:

```
var select = Object.assign(document.createElement('select'));
var option1 = Object.assign(document.createElement('option'));
var option2 = Object.assign(document.createElement('option'), {selected: true, hidden: true, disabled: true});
select.append(option1);
select.append(option2);
console.log(select.selectedIndex); // 1, correct behavior
```

packages/nimble-components/src/select/index.ts
```diff
@@ -668,6 +673,17 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
...
```
rajsite, line 684: For the tech debt item that is created about synchronous state can you link to that from here as well? And discuss in the issue that there can be state inaccuracies if options are inserted in a different order than just appending.

packages/nimble-components/src/list-option/index.ts
```diff
@@ -51,6 +52,17 @@ export class ListOption extends FoundationListboxOption {
             .map(node => node.textContent?.trim())
             .join(' ');
     }
+
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        if (this.isDropdownOwner(this.parentElement)) {
```
jattasNI, line 58: When we do #791 do you think we'll end up with `list-option`s that aren't direct children of a `select`/`combobox`? Anything to think through or write tests for now? Or do we just need to try to remember to think about this when we work on that?

For example:
```html
<nimble-select>
   <nimble-list-group>Texas
      <nimble-list-option>Austin</nimble-list-option>
      <nimble-list-option>San Antonio</nimble-list-option>
   </nimble-list-group>
   <nimble-list-group>Not Texas
      <nimble-list-option>Tokyo</nimble-list-option>
   </nimble-list-group>
</nimble-select>
```

But then again we might [copy Shoelace's flat pattern](https://shoelace.style/components/select#grouping-options). Or not, it might make it hard to filter by all options in a section.

```html
<sl-select>
  <small>Section 1</small>
  <sl-option value="option-1">Option 1</sl-option>
  <sl-option value="option-2">Option 2</sl-option>
  <sl-option value="option-3">Option 3</sl-option>
  <sl-divider></sl-divider>
  <small>Section 2</small>
  <sl-option value="option-4">Option 4</sl-option>
  <sl-option value="option-5">Option 5</sl-option>
  <sl-option value="option-6">Option 6</sl-option>
</sl-select>
```

packages/nimble-components/src/select/index.ts
rajsite: From offline discussion, there is a temporal time where it is possible for the select.options state to be in an invalid state. i.e. a new option marked selected is added to the middle of the options list. But this does not impact Angular because the current CVA implementation only looks at the value property on options and creates a value using a unique incrementing id. It doesn't care about the selectedIndex reported by the select where the nimble-select can report incorrect values at times.
We will track the nimble temporal state issues in a tech debt issue mostly to document it but it should not cauuse problems for angular.
Also could not think of alternatives that would not be a potential performance concern at the moment. Discussed options were:
- modify select and selectedIndex to be accessors that on access would access all the child options
- in connectedCallback query all child items and set options to current state (for options inserted one after another, likely expensive / redundant work done on every connectedCallback of children, also does not improve the synchronous behavior since it only runs after connect to the dom so at best a partial solution)

packages/nimble-components/src/select/index.ts
```diff
@@ -668,6 +673,17 @@ export class Select extends FormAssociatedSelect implements ErrorPattern {
         }
     }
 
+    /**
+     * @internal
+     */
+    public registerOption(option: ListOption): void {
```
rajsite, line 679: From offline discussions we will apply this same behavior to all our listbox implementations (select, combobox, and user-mentions-listbox) so they will have the same behavior and tests (with link to tech debt issue)

packages/nimble-components/src/list-option/index.ts
```diff
@@ -51,6 +52,17 @@ export class ListOption extends FoundationListboxOption {
             .map(node => node.textContent?.trim())
             .join(' ');
     }
+
+    public override connectedCallback(): void {
```
jattasNI, line 56: I have some mild concern that we'll break other components that contain list options like the combobox. Could you take a look at their test coverage and ensure it's good enough to be confident we're not breaking anything? The obvious codepath of concern is when the parent is not a `ListOptionOwner`. Presumably `connectedCallback()` just runs and does nothing. But do we want to ensure that works for dynamic cases like the tests you added to `select.spec.ts`?

packages/nimble-components/src/list-option/index.ts
rajsite: > But then again we might copy Shoelace's flat pattern

I think our goal should be to copy patterns from DOM first, i.e. the former with a group element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

packages/nimble-components/src/select/index.ts
atmgrifter00: Created #1915. I also updated `Combobox` to be a `ListOptionOwner`.

packages/nimble-components/src/list-option/index.ts
atmgrifter00: So, I've updated the `Combobox` to actually implement `ListOptionOwner`. The only other case of interest is the `RichTextMentionListbox`. If the `connectedCallback` implementation were to have a problem with non-`ListOptionOwner` types, then at the moment, I would expect Chromatic tests to fail (which they don't). Do you feel this is sufficient?

packages/nimble-components/src/list-option/index.ts
atmgrifter00: 1) I've updated the `ListOptionOwner` interface to extend `HTMLElement`, which allows me to get rid of the `unknown` typing.
2) I'm handling the case where `parent` is `null` (as `parentElement`'s type is `HTMLElement | null`.

packages/nimble-components/src/list-option/index.ts
atmgrifter00: I prefer we not bother with making this implementation robust enough to handle the inevitable feature that #791 will introduce, as we still don't have an agreed upon design in place for handling that.

packages/nimble-components/src/list-option/index.ts
jattasNI: I also did some manual testing of the rich text mention listbox and didn't see any issues. LGTM!

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -281,16 +281,29 @@ describe('Select', () => {
         await disconnect();
     });
 
-    it('changing textContent of selected option updates display of select', async () => {
+    it('option added directly to DOM synchronously registers with Select', async () => {
```
jattasNI, line 284: Was it intentional to remove the previous test `changing textContent of selected option updates display of select`?

jattasNI (approved):

packages/nimble-components/src/select/tests/select.spec.ts
rajsite: Don't think so, betting it was a merge issue, I re-added the test and tweaked the synchronous issue comments below

packages/nimble-components/src/list-option/index.ts
rajsite: Created an issue to track the tech debt we had discussed of the rich text list box not aligning with our other listbox implementations: https://github.com/ni/nimble/issues/1926

rajsite (approved):

Issue #1538, opened 2023-09-14
Placeholder text in nimble-table columns


m-akinc:
## 😯 Problem to Solve

Need to support placeholder text in table columns.
Related: #1511

LIMS use case where missing date values should render "--":
![image](https://github.com/ni/nimble/assets/7282195/9ba3d3a5-a86f-41fc-9847-63d64f3ad7c3)

SLE Executions grid where a missing Notebook value is rendered **as a link** with text "Notebook not available" in italics:
![image](https://github.com/ni/nimble/assets/7282195/775d3ff1-5302-4cb5-8e92-a69dcf24d804)

SLE Systems grid grouping by "Group" value, where some systems have no "Group" value:
![image](https://github.com/ni/nimble/assets/7282195/1d7c9a86-1ee8-4481-b4c7-93533586a2f4)

Need to support different placeholder text in different columns. Cannot standardize on a single placeholder.

Open questions:
1. Should we support group header text that is different from the cell text? 
    a. E.g. "--" in cell but "No value" in group header.
    b. E.g. empty cell in an icon column, but "No value" in group header.
2. Should placeholder text be dimmed? Only in cells, but not group headers?
3. Should placeholder text always be italicized?
4. Does placeholder text use the same alignment of cells with values, e.g. number-text column with right alignment?

fredvisser:
Looking at the open questions:

1. Should we support group header text that is different from the cell text?
a. E.g. "--" in cell but "No value" in group header.
b. E.g. empty cell in an icon column, but "No value" in group header.

If it's technically straightforward – **yes**. "No value" communicates more clearly than "--" in the group header.

![Image](https://github.com/ni/nimble/assets/1458528/7362cbed-4693-4e21-8f03-4f83dd422334)

@NIbokeefe - thoughts on the below questions?
2. Should placeholder text be dimmed? Only in cells, but not group headers?
3. Should placeholder text always be italicized?
4. Does placeholder text use the same alignment of cells with values, e.g. number-text column with right alignment?

fredvisser:
After talking with Brandon and Jesse –

1. Should we support group header text that is different from the cell text?
- Yes, "No value" makes sense for the group header
2. Should placeholder text be dimmed? Only in cells, but not group headers?
- Yes, placeholder text should be dimmed (standard placeholder token value) in the data rows, but not in group headers.
3. Should placeholder text always be italicized?
- No, the placeholders don't need to be italicized
[...]

PR #1914, opened 2024-03-08
Add placeholder text to group rows

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #1538

This PR adds support for placeholder text in group rows within the table. There is follow on work that will need to be done before the linked issue is resolved to add placeholders to cells, but that is not included in this PR.

This PR is scoped to the following behavior changes to columns:
- text column
    - Show "Empty" on group rows for empty string values
    - Show "No value" on group rows for `null` and `undefined` values
- date-text column
    - Show "No value" on group rows for `null` and `undefined` values
- number-text column
    - Show "No value" on group rows for `null` and `undefined` values
- duration-text column
    - Show "No value" on group rows for `null` and `undefined` values
- icon column
    - Show "No value" on group rows for `null` and `undefined` values
- enum-text column
    - Show "No value" on group rows for `null` and `undefined` values
- anchor column
    - Show "Empty" on group rows for empty label values
    - Show "No value" on group rows for `null` and `undefined` label values

## 👩‍💻 Implementation

- Added new labels to the table label provider for _Empty_ and _No value_
- Added a `applyPlaceholderTextIfNeeded` function to `TableColumnTextGroupHeaderViewBase` that concrete classes can call to apply a placeholder
- Update concrete instances of `TableColumnTextGroupHeaderViewBase` to call the new function

## 🧪 Testing

- Wrote a number of new tests for each column to make sure the appropriate group row text was rendered for different values
- Updated matrix tests to ensure each column included all possible placeholder group values in the matrix test
    - Note: This meant creating a new matrix story for the text column because those matrix stories had never been created

## ✅ Checklist


packages/nimble-components/src/label-provider/base/tests/label-provider-stories-utils.ts
```diff
@@ -63,7 +63,7 @@ export const labelProviderMetadata: Meta<LabelProviderArgs> = {
                 }
                 ${tableTag} {
                     /* Make the table big enough to remove vertical scrollbar */
-                    height: 550px;
+                    height: 650px;
```
m-akinc, line 66: Can this be based on `x.labelTokens.length` so that (a) we don't have to update this when adding new labels, and (b) we don't have unnecessarily big snapshots for less populated tables?

packages/nimble-components/src/table-column/text-base/group-header-view/index.ts
```diff
@@ -15,7 +19,30 @@ export abstract class TableColumnTextGroupHeaderViewBase<
...
-    public text = '';
+    public text = tableGroupRowNoValuePlaceholderLabel.getValueFor(this);
+
+    /**
+     * Sets `this.text` to the appropriate placeholder if `groupHeaderValue` warrants it.
+     * @param groupHeaderValue The value for the group
+     * @returns `true` if `this.text` was set to a placeholder, `false` otherwise.
+     */
+    protected applyPlaceholderTextIfNeeded(
+        groupHeaderValue: TableFieldValue
```
m-akinc, line 36: Why doesn't this just use `this.groupHeaderValue`?

packages/nimble-components/src/table-column/text-base/group-header-view/index.ts
```diff
@@ -15,7 +19,30 @@ export abstract class TableColumnTextGroupHeaderViewBase<
...
     @observable
-    public text = '';
+    public text = tableGroupRowNoValuePlaceholderLabel.getValueFor(this);
+
+    /**
+     * Sets `this.text` to the appropriate placeholder if `groupHeaderValue` warrants it.
+     * @param groupHeaderValue The value for the group
+     * @returns `true` if `this.text` was set to a placeholder, `false` otherwise.
+     */
+    protected applyPlaceholderTextIfNeeded(
```
m-akinc, line 35: I don't like that derived classes have to "opt in" to this behavior by calling it directly. I considered having the base provide a default implementation of `groupHeaderValueChanged()` that called `applyPlaceholderTextIfNeeded()`, but then it needs to return a boolean for overrides to check to determine whether to do any further formatting. Maybe it's now justified to have this base class define `groupHeaderValueChanged()`, `columnConfigChanged()`, and `updateText()` (abstract or no-op):

```ts
    protected columnConfigChanged(): void {
        if (!this.applyPlaceholderTextIfNeeded()) {
            this.updateText();
        }
    }
    protected groupHeaderValueChanged(): void {
        if (!this.applyPlaceholderTextIfNeeded()) {
            this.updateText();
        }
    }
    protected abstract updateText(): void;
```

packages/nimble-components/src/label-provider/table/index.ts
```diff
@@ -38,7 +40,9 @@ const supportedLabels = {
     groupSelectAll: tableGroupSelectAllLabel,
     rowSelect: tableRowSelectLabel,
     rowOperationColumn: tableRowOperationColumnLabel,
-    rowLoading: tableRowLoadingLabel
+    rowLoading: tableRowLoadingLabel,
+    groupRowNoValuePlaceholder: tableGroupRowNoValuePlaceholderLabel,
```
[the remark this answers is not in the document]
mollykreis: Great point. I'll update these names.

packages/nimble-components/src/label-provider/base/tests/label-provider-stories-utils.ts
mollykreis: Sounds reasonable to me. I'll update it here, and then, if we're happy with the approach I take, I can update the other stories that take a similar approach.

packages/nimble-components/src/table-column/enum-text/tests/table-column-enum-text.spec.ts
```diff
@@ -359,7 +366,7 @@ describe('TableColumnEnumText', () => {
         async function setupInvalidMappings(): Promise<ModelInvalidMappingsFixture<Table<SimpleTableRecord>>> {
             const source = new ModelInvalidMappings();
             const result = await fixture<Table<SimpleTableRecord>>(html<ModelInvalidMappings>`
-                <${tableTag} style="width: 700px">
+                <${tableTag} ${ref('table')} style="width: 700px">
```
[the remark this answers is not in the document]
mollykreis: The typing in these tests was a little off, so I was mostly just trying to make the compiler happy. The root of the problem is that the tests were assigning an `ModelInvalidMappings` instance to a variable typed as `Model`, so they needed to be identical. However, since they are identical, I've now gotten rid of `ModelInvalidMappings` and `ModelInvalidMappingsFixture` in favor of using `Model` and `ModelFixture` in the invalid setup (since ultimately they were already implicitly required to be the same).

change/@ni-nimble-components-561538cd-7588-4219-9fd7-1fe50633d0f7.json
m-akinc, line 1: I'm assuming we're not intending to use italics for the placeholder text, so I would change the PR description to use "Empty"/"No value" instead of _Empty_/_No value_.

m-akinc (approved):

packages/nimble-components/src/label-provider/table/index.ts
```diff
@@ -89,6 +93,12 @@ export class LabelProviderTable
     @attr({ attribute: 'row-loading' })
     public rowLoading: string | undefined;
 
+    @attr({ attribute: 'group-row-placeholder-no-value' })
```
jattasNI, line 96: I learned that there is work to do in Angular and Blazor for new label provider strings too. See #1927. It probably makes sense to do in this PR but would also be fine in a follow up.

jattasNI (approved):

packages/nimble-components/src/table-column/text-base/group-header-view/index.ts
```diff
@@ -15,7 +19,44 @@ export abstract class TableColumnTextGroupHeaderViewBase<
...
+    /**
+     * Sets `this.text` to the appropriate placeholder if `groupHeaderValue` warrants it.
+     * @returns `true` if `this.text` was set to a placeholder, `false` otherwise.
+     */
+    private applyPlaceholderTextIfNeeded(): boolean {
+        if (
+            this.groupHeaderValue === null
+            || this.groupHeaderValue === undefined
+        ) {
+            this.text = tableGroupRowPlaceholderNoValueLabel.getValueFor(this);
```
rajsite, line 53: when a token is referenced in the binding of a template then I believe the binding will be updated when the token value changes. We could either move this to the template or subscribe to the tokens and observe them manually.

Doing it in the template may be a little tricky mixing with hasOverflow, didn't play with it.

Should probably also test the group header element responding to label-provider updates. Could either have tests for base specifically or put it in group-header-view-text and call it covered

change/@ni-nimble-components-561538cd-7588-4219-9fd7-1fe50633d0f7.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Thanks for breaking it up, this is a nice chunk of the work

packages/nimble-components/src/label-provider/table/index.ts
mollykreis: I was planning on doing it in a different PR, but there isn't a reason not to include it here. I'll go ahead and do that.

packages/nimble-components/src/label-provider/base/tests/label-provider-stories-utils.ts
```diff
@@ -61,9 +62,9 @@ export const labelProviderMetadata: Meta<LabelProviderArgs> = {
                 p {
                     font: var(${bodyFont.cssCustomProperty});
                 }
-                ${tableTag} {
+                ${tableTag}.${x => x.labelProviderTag} {
                     /* Make the table big enough to remove vertical scrollbar */
-                    height: 550px;
+                    height: ${x => `calc((34px * ${x.labelTokens.length.toString()}) + 32px)`};
```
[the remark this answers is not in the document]
mollykreis: Yep. I just wanted to make sure we were happy with this pattern before I adopted it in multiple places.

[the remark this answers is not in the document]
mollykreis: I ended up changing the pattern a little bit, but I've now adopted it in the icon and status table stories.

msmithNI (approved):

atmgrifter00 (approved):

rajsite (approved):

PR #1916, opened 2024-03-11
Wafer Map Experimental Event Coordinator

munteannatan:
# Pull Request

## 🤨 Rationale

I tried to expand on the strategy switching mechanism from the renderer, and reused it in the coordinator

This approach has the benefits of isolating the experimental code while in development, and an easy way to switch to the new version by changing just the imports and switching mechanism and removing the old versions after they become obsolete

## 👩‍💻 Implementation

because arquero causes typescript errors I implemented an alternative search and posted a PR with fixes in arquero repo https://github.com/uwdata/arquero/pull/346
changed the zoom handler to catch events on the component, not just the canvas
removed zooming border of around 100px when zooming in the wafer after PO decision, resulting in less complex code and touch controls enabling.

## 🧪 Testing

existing tests are passing
created new hover unit tests with wafer and data manager mocks

## ✅ Checklist


packages/nimble-components/tsconfig.json
```diff
@@ -20,6 +20,7 @@
         // so imports with side-effects are explicit
         "importsNotUsedAsValues": "error",
         "noImplicitOverride": true,
+        "skipLibCheck": true,
```
atmgrifter00, line 23: Discussed with @rajsite, and as expected we wouldn't be able to submit this. Some possibilities for moving forward include:
- Have an acceptable alternative to using arquero that meets all of the necessary performance requirements
- Providing our own typing for arquero that addresses the known typescript errors.
- Addressing the typescript issues in arquero itself by submitting a PR in arquero's repo.

@rajsite, are there are any other options worth considering?

munteannatan: I implemented an alternative for index searching.
I also created a PR with some type fixes in arquero. https://github.com/uwdata/arquero/pull/346
If you have time for some insights over there please let me know

packages/nimble-components/src/wafer-map/modules/event-coordinator.ts
```diff
@@ -12,25 +13,35 @@ export interface EventCoordinatorCallbacks {
  */
 export class EventCoordinator {
     private readonly zoomHandler;
-    private readonly hoverHandler;
+    private readonly stableHoverHandler;
+    private readonly expHoverHandler;
```
atmgrifter00, line 17: Even though this is a temporary name, I still suggest ditching the "exp" short-hand for "experimental".

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
```diff
@@ -0,0 +1,79 @@
+import type { WaferMap } from '../..';
+import { PointCoordinates, WaferMapOriginLocation } from '../../types';
+
+/**
+ * HoverHandler deals with user interactions and events like hovering
+ */
+export class HoverHandler {
+    public constructor(private readonly wafermap: WaferMap) {}
+
+    public mousemove(event: MouseEvent): void {
```
atmgrifter00, line 10: Our current patterns elsewhere would suggest instead of providing event handlers on a model class for a consumer object to register with, we instead expose `connect` and `disconnect` implementations (called those names for convention's sake), and then in those implementations you would handle registering/unregistering the events against the already available reference to `waferMap`, and calling the `connect` and `disconnect` methods from the `WaferMap`'s `connectedCallback` and `disconnectedCallback` implementations. Similar to [this](https://github.com/ni/nimble/blob/aeaa29b6db90372e2c2adf3f88275b758f1e0771/packages/nimble-components/src/table/models/virtualizer.ts#L56).

atmgrifter00: Additionally, the ultimate name for the event handler function should likely be of the form of `onMouseMove`.

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+        this.wafermap.hoverDie = undefined;
+    }
+
+    public mouseout(): void {
+        this.wafermap.hoverDie = undefined;
+    }
+
+    private calculateDieCoordinates(
```
atmgrifter00, line 53: This almost seems like a generally useful function that could be needed for some other function beyond hovering. Do you think it makes sense to pull this out into a static utility class?

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
atmgrifter00, line 10: Reading through the rest of the code, I can't seem to find where this method is being used outside of a test. I feel like I'm missing something.

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+            WaferMapOriginLocation.bottomLeft,
+            undefined,
+            getDataManagerMockForHover(
+                { left: 0, right: 0, top: 0, bottom: 0 },
+                getScaleQuantile([1, 10], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
+                getScaleQuantile([1, 10], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+            ) as DataManager
+        ) as WaferMap;
+        hoverHandler = new HoverHandler(waferMock);
+        element.addEventListener('mousemove', event => hoverHandler.mousemove(event));
```
atmgrifter00, line 46: Assuming you made the changes that I suggested in the `HoverHandler` implementation, I would think this would have to change such that the method that is creating the `waferMock` would also handle creating the `hoverHandler` instance, and be responsible for calling its `connect` implementation.

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+        ({ element, connect, disconnect } = await setup());
+        await connect();
+        waferMock = getWaferMapMockHover(
+            tableFromArrays({
+                colIndex: Int32Array.from([1]),
+                rowIndex: Int32Array.from([1]),
+                value: Float64Array.from([1])
+            }),
+            zoomIdentity,
+            WaferMapOriginLocation.bottomLeft,
```
atmgrifter00, line 37: We're missing some tests for particular code paths in `HoverHandler` because we're using one state like this in the test, when the implementation also cares about `bottomRight`. Additionally, we're not using data values (that I can see) that would be leveraging the `Math.ceil` or `Math.floor` functionality. It seems like we should be creating a `parameterizeSpec` that allows us to test multiple variants of state with expected results.

packages/nimble-components/tsconfig.json
atmgrifter00: Could you leave a comment in the code where you intend to ultimately switch to using arquero? Ultimately, I would expect to have an issue created in Nimble that is blocked on the PR you submitted to arquero, which we can then reference in the TODO comment.

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
atmgrifter00: Thinking through this further, and after discussing with @jattasNI, we agree that the use of mocks to test this functionality seems unnecessarily convoluted. Is there difficulty in hosting an actual `WaferMap` instance in the DOM, with some test dies, and then issuing a `mousemove` event on the `WaferMap` itself?

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
munteannatan: I understand, but in our use case we attach and detach events based on the update type in the wafermap. should we call those connect and disconnect instead?

munteannatan: the event handler switches between the stable hover handler, which is the existing one, and the experimental, which is this one. based on which one is selected they get to handle the mousemove event

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
munteannatan: the experimental hover handler needs a data manager instance which is not implemented yet. if we create a wafer instance, I will still need to replace the existing data manager with a mock and remove the existing event handlers. These mocks are temporary to not increase the scope of this PR with the implementation for the data manager

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
munteannatan: I made some parametric tests, but it seems taht the scale quantile does not need floor or ceil. I will leave them there so we can switch it later with the data manager

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
atmgrifter00: Yeah, I'm not a fan of the current implementation being influenced by the need to switch between the two systems. Honestly, I'd rather we do away with the whole `EventCoordinator` class, as I'm not really sure it's offering a clear utility even from a separation of concerns perspective. Even the `EventCoordinatorCallbacks` interface that is declared isn't even used in that file, and should just move to the `types.ts` file.

I would prefer the `WaferMap` just handle registering/unregistering the events that it has handlers for (in its `connectedCallback` and `disconnectedCallback` implementations), and also move to the `connect`/`disconnect` paradigm I mention, and just call those methods temporarily when the update type is switched.

atmgrifter00: I see...but, as I replied in the other comment, I would prefer we refactor this.

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
atmgrifter00: At the very least we need to create an issue (or link to an existing one) and then link to that issue in the test code and explain that we intend to change the implementation. As is, it feels like this is an incomplete way to mock out dependencies. Having what seems like an arbitrary div in the DOM to stand-in for the wafer-map to issue events through seems janky. We should make it clear that the end goal is to just use a real `WaferMap` instance.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -168,12 +177,18 @@ export class WaferMap extends FoundationElement {
         this.canvasContext = this.canvas.getContext('2d', {
             willReadFrequently: true
         })!;
+        this.hoverHandler.connect();
+        this.experimentalHoverHandler.connect();
```
atmgrifter00, line 181: Isn't this problematic in that both the `hoverHandler` and `experimentalHoverHandler` are now responding to the same event?

atmgrifter00 (approved):

change/@ni-nimble-components-163379f4-09c6-4606-95f4-05917dd50ee0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Created new hover event for the new diesTable api and changed the zoom event",
+  "packageName": "@ni/nimble-components",
+  "email": "33986780+munteannatan@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
rajsite, line 7: There are not any package.json changes in this PR so I wouldn't expect any lock file changes.
May need to bring in the lock file from main

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            element.addEventListener('mousemove', event => hoverHandler.onMouseMove(event));
+            element.dispatchEvent(
+                new MouseEvent('mousemove', {
+                    clientX: 15,
+                    clientY: 15
+                })
+            );
+            processUpdates();
+            expect(waferMock.hoverDie).toEqual(value[1]);
+        });
```
rajsite, line 120: Lets use the parameterizeSpec patterns instead: https://www.npmjs.com/package/@ni/jasmine-parameterized

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            [
+                WaferMapOriginLocation.topRight,
+                {
+                    index: 1,
+                    x: 2,
+                    y: 2
+                }
+            ]
+        ];
+        parameters.forEach(value => {
```
rajsite, line 86: Lets use the parameterizeSpec patterns instead: https://www.npmjs.com/package/@ni/jasmine-parameterized

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+            element.dispatchEvent(
+                new MouseEvent('mousemove', {
+                    clientX: 15,
+                    clientY: 15
+                })
+            );
+            processUpdates();
+            expect(waferMock.hoverDie).toEqual(value[1]);
+        });
+    });
```
rajsite, line 121: Lets use the parameterizeSpec patterns instead: https://www.npmjs.com/package/@ni/jasmine-parameterized

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -190,8 +205,12 @@ export class WaferMap extends FoundationElement {
         if (this.validity.invalidDiesTableSchema) {
             return;
         }
+        this.renderer = this.diesTable === undefined
```
rajsite, line 208: We are using `waferMap.diesTable === undefined` pretty often now as a signal for the experimental renderer. CAn we add an `@internal` method like `waferMap.isExperimentalRenderer()` to do that check. We should replace all the spots we are doing that check with the helper method.

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
rajsite: Should look something like:

```ts
const testCases = [
    {name: WaferMapOriginLocation.bottomLeft, expectedDie: undefined},
    {name: WaferMapOriginLocation.topLeft, expectedDie: undefined},
    {name: WaferMapOriginLocation.bottomRight, expectedDie: undefined},
    {name: WaferMapOriginLocation.topRight, expectedDie: undefined},
] as const;
parameterizeSpec(testCases, (spec, name, value) => {
    spec(`will return undefined when mouse moved out of range from ${name}`, () => {
        waferMock.originLocation = value.name;
        hoverHandler = new HoverHandler(waferMock);
        element.addEventListener('mousemove', event => hoverHandler.onMouseMove(event));
        element.dispatchEvent(
            new MouseEvent('mousemove', {
                clientX: 15,
                clientY: 15
            })
        );
        processUpdates();
        expect(waferMock.hoverDie).toEqual(value.expectedDie);
    });
});
```

Note:
- How the testCases data is marked `as const`
- How the testCases fields are well-named (`name`, `expectedDie`) instead of indexes in an array

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
atmgrifter00, line 86: Sorry this escaped me the first time...shouldn't this be a `parameterizeSpec`?

packages/nimble-components/src/wafer-map/tests/hover-handler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    it('will return undefined when mouse moved out of range', () => {
+        const parameters: [WaferMapOriginLocation, HoverDie | undefined][] = [
+            [WaferMapOriginLocation.bottomLeft, undefined],
+            [WaferMapOriginLocation.topLeft, undefined],
+            [WaferMapOriginLocation.bottomRight, undefined],
+            [WaferMapOriginLocation.topRight, undefined]
+        ];
+        parameters.forEach(value => {
```
atmgrifter00, line 108: Again, this should likely be a `parameterizeSpec`.

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
```diff
@@ -0,0 +1,99 @@
...
+        const dieCoordinates = this.calculateDieCoordinates(this.wafermap, {
+            x: invertedPoint[0],
+            y: invertedPoint[1]
+        });
+        const colIndex = this.wafermap
+            .diesTable!.getChild('colIndex')!
+            .toArray() as Int32Array;
+        const rowIndex = this.wafermap
+            .diesTable!.getChild('rowIndex')!
+            .toArray() as Int32Array;
```
rajsite, line 50: We should not need all this `as` casting, we already reviewed how we can type `Table<>` correctly. You can see in this example how TypeScript will know the types if we type it correctly: https://stackblitz.com/edit/arrow-table-generics?file=src%2Fmain.ts

image

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -211,12 +230,19 @@ export class WaferMap extends FoundationElement {
...
+            this.zoomHandler.connect();
         } else if (this.waferMapUpdateTracker.requiresRenderHoverUpdate) {
             this.renderer.renderHover();
         }
     }
 
+    /**
+     * @internal
+     */
+    public isExperimentalRenderer(): boolean {
```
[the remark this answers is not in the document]
munteannatan: I left this because we are also checking the type of the dies table. When using the function it will consider that undefined is still part of the type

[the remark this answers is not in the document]
rajsite: ahh, makes sense, it's not related to render modes but actually inspecting table data. Makes perfect sense!

jattasNI (approved):

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
munteannatan: I made the generic table but now the precision and bitWidth validity checks give type erros. should we keep them? Is there a way to make thwm work?

rajsite: Conceptually yes. These are web components so a user could leverage them in a JS application without type checking in which case we need the validator for runtime type checking. All the typescript work helps TypeScript users and us as maintainers to find errors at compile time via static analysis.

Looking at the CI build errors my guess is it is related to the extra fields added to the type:
```
export interface WaferRequiredTypeMap {
    [key: string]: DataType;
    [key: number]: DataType;
    [key: symbol]: DataType;
    colIndex: Int32;
    rowIndex: Int32;
    value: Float64;
}
```
ie the:
```
    [key: string]: DataType;
    [key: number]: DataType;
    [key: symbol]: DataType;
```

Either way I'd suggest we not add those yet, there are still concerns about that API that are not resolved and it does not seem necessary to include those yet.

munteannatan: removing them leaves this error: Type 'T' does not satisfy the constraint 'TypeMap'.
  Type 'WaferRequiredTypeMap' is not assignable to type 'TypeMap'.

packages/nimble-components/src/wafer-map/types.ts
```diff
@@ -79,3 +87,12 @@ export interface WaferMapValidity extends ValidityObject {
...
 }
+
+export interface WaferRequiredTypeMap {
+    [key: string]: DataType;
+    [key: number]: DataType;
+    [key: symbol]: DataType;
+    colIndex: Int32;
+    rowIndex: Int32;
+    value: Float64;
+}
```
rajsite, line 98: I sunk it down and played with it locally and I don't think we actually need to specify the index signatures if we do the following (I recommend going this way comments and all):

```suggestion
// Apache arrow probably should not be using a Record and index types on TypeMap
// because in strict checking they end up required.
// See: https://github.com/apache/arrow/issues/12663#issuecomment-1088244575
// We can work around that issue by using a type alias instead of an interface
// Where index signatures are looser.
// See: https://github.com/microsoft/TypeScript/issues/15300#issuecomment-1317901527
// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
export type WaferRequiredFields = {
    colIndex: Int32,
    rowIndex: Int32,
    value: Float64
};
```

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
```diff
@@ -0,0 +1,102 @@
...
+    PointCoordinates,
+    WaferMapOriginLocation,
+    WaferRequiredTypeMap
+} from '../../types';
+
+/**
+ * HoverHandler deals with user interactions and events like hovering
+ */
+export class HoverHandler<T extends WaferRequiredTypeMap> {
+    public constructor(private readonly wafermap: WaferMap<T>) {}
```
rajsite, line 12: We don't need to propagate the generic  into all of these classes. Most if not all of these classes at most care about the required fields which are already satisfied by the default type you specified here:
```ts
export class WaferMap<
    T extends WaferRequiredFields = WaferRequiredFields
> extends FoundationElement {
```

Almost none of these classes actually care if there are extra fields specified by the user, they only operate on the required fields. So we don't need to propagate the generic everywhere, only where it's needed (when we need to know more than just the required fields)

```suggestion
import {
    PointCoordinates,
    WaferMapOriginLocation
} from '../../types';

/**
 * HoverHandler deals with user interactions and events like hovering
 */
export class HoverHandler {
    public constructor(private readonly wafermap: WaferMap) {}
```

So lets not add the generic to all these. It makes it more complex without any benefit

munteannatan: I removed the generics from the modules but there are still type mismatches when initializing them

rajsite: Yea that's interesting. I sunk the changes locally and I think it has to do with Apache Arrow's slightly strange typing on table fields. I pushed a change that does a couple things:
- Made an accessor that performs the `as` cast so that they are not sprinkled in many places
- Updated mocks so that they cast to the mocked type before returning, this let us avoid lots of `as` casts in tests
- Modified the validator code to use references to the field objects instead of indexes in the arrays. This helped improve TS inferencing so many non-null assertions (`!`) could be removed and made it easier to read.

rajsite (approved):

Issue #495, opened 2022-04-12
Chromatic testing of hover/focus states


m-akinc:
## 🧹 Tech Debt
We would like to capture regressions in the visual styling of interactive states, like hover or keyboard focus.

jattasNI:
After a couple rounds of investigation it looks like the Chromatic approach isn't going anywhere soon. Closing this in favor of #705; we can always revive it if needed.

jattasNI:
[Chromatic announced a beta](https://www.chromatic.com/blog/how-to-visual-test-ui-using-playwright/) for a feature which allows you to take snapshots from within a Playwright test and have Chromatic manage them and track diffs. I think we could use this to write a test that hovers and focuses each of our input components in succession and takes snapshots to ensure that the hover and focus visuals are consistent.

We'd have to figure out how this relates to our matrix tests. In an ideal world, we'd test the hover and focus states combined with all the other state combinations. However, that'd be a pretty big explosion in the number of snapshots we need to take, since you can only hover or focus one thing at a time. Perhaps as a stopgap we'd just have tests for each input component in its default configuration in the 3 themes x the 3 states (hover, focus, hover+focus). We could expand that for specific components that are known to look different in other states, like button appearances. But omit testing for uninteresting states that look identical to each other.

A possible next step might be to prototype this on an innovation day.

m-akinc:
I need the addon to accept a [bunch of fixes](https://github.com/chromaui/storybook-addon-pseudo-states/pull/103), then we can update to that version and write our tests.

rajsite:
@fredvisser think we can leverage our chromatic support to get them to look at the PRs mert put up? https://github.com/chromaui/storybook-addon-pseudo-states/pulls/m-akinc

PR #1919, opened 2024-03-11
Add interaction stories for buttons, number field, and anchor

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #495

## 👩‍💻 Implementation

I had to make some changes in `storybook-addon-pseudo-states` to support our styling, but now we can pull in that fixed release and add some initial visual tests.

The way the addon typically works is that you set story parameters to select which component(s) should have the pseudo-state styling, e.g. 
```js
ButtonStory.parameters = {
  pseudo: {
    hover: ["#one", "#two", "#three"],
    focus: true, // means apply to all components
    active: ".some-class",
  },
}
```
However, this does work well with our matrix story approach. The effect of these parameters is just to set certain classes on the indicated elements. So, instead, we can bypass the story parameters and directly apply the classes to our elements. For example, any element with the class `hover` or `pseudo-hover-all` will be styled by the addon as if hovered.

NOTE: Because we have to put the classes directly on the elements in our Story html, we can't put them on any shadow DOM elements, e.g. the inc/dec buttons of the number field. However, puting `pseudo-<state>-all` on a host element will cause all elements in its shadow DOM to be styled as if they have that state. This can create some visuals that are practically impossible (e.g. a number field and both of its inc/dec buttons all having keyboard focus at the same time), but they are still useful from a visual testing standpoint.

The pseudo-states supported by the addon are:
- `:hover`
- `:active`
- `:focus-visible`
- `:focus-within`
- `:focus`
- `:visited`
- `:link`
- `:target`

It is only interesting to test states that we specifically style for, so generally, that is hover, active, and some form of focused. Because our components use both `:focus-visible` (`focusVisible` from FAST) and `:focus-within`, I'm not distinguishing the two, and I'm setting classes for both whenever we want to see "focused" styling. We could potentially test all combinations of these pseudo-states, but that seems unnecessary, especially considering that we want to test these states in _conjunction_ with other appearance/config variations. So I'm limiting combination testing to just hovered+active, since that is a common use case for mouse-based interaction. To summarize, that means I'm proposing we test:
- hover (only)
- active (only)
- focused (only)
- hovered and active

Initially, I'm adding new test matrix stories for:
- anchor
- anchor-button
- button
- menu-button
- number-field
- toggle-button

These seemed to be some of the more interesting components in terms of having distinct keyboard-focused and/or active states, (in addition to hover, which applies to most/all? components).

Some combinations of interaction states with control attribute values aren't interesting, e.g. a disabled control with focus or active interaction state. In order to exclude certain combinations, we had to refactor the `createMatrix` utility function and introduce a new `createMatrixInteractionsFromStates` function.

## 🧪 Testing

Storybook visual tests

## ✅ Checklist


m-akinc:
@mollykreis I would ask Malcolm, but he's out this week, so can you please buddy for me?

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
```diff
@@ -35,36 +38,53 @@ const partVisibilityStates = [
...
-        <${menuTag} slot="menu">
-            <${menuItemTag}>Item 1</${menuItemTag}>
-            <${menuItemTag}>Item 2</${menuItemTag}>
-        </${menuTag}>
     </${menuButtonTag}>
 `;
 
 export const menuButtonThemeMatrix: StoryFn = createMatrixThemeStory(
     createMatrix(component, [
+        interactionStates.slice(0, 1),
```
mollykreis, line 75: I don't think we should make assumptions about the order of the states in the `interactionStates` array. It looks like most, if not all, of the tests are wanting to use the "no interaction" state separately, so why not take that out of the `interactionStates` array so that the tests for interaction states don't have to skip the first entry?

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
```diff
@@ -76,6 +80,17 @@ export const numberFieldThemeMatrix: StoryFn = createMatrixThemeStory(
     ])
 );
 
+export const numberFieldInteractionsThemeMatrix: StoryFn = createMatrixThemeStory(
+    createMatrix(component, [
+        interactionStates.filter(x => x[0] && !x[0].includes('Active')),
+        disabledStates,
+        hideStepStates.slice(0, 1),
+        valueStates.slice(1, 2),
+        errorStates.slice(0, 2),
```
mollykreis, line 89: Similar feedback here -- can we come up with a better approach than calling `slice` to pick specific states that we want to test. Some of the downsides I see to this approach are that it is very difficult to look at this code and understand what set of states you are trying to tests and that this code becomes unnecessarily tied to the order in which states are added to arrays.

I think the `filter` approach you take in the toggle button tests would be more readable, especially with a comment saying something like _only include states where XYZ_ or _skip states with XYZ_.

change/@ni-nimble-components-ceebd404-0fb3-43e5-8f98-04f4eed64531.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: It looks like we are currently testing combinations that we don't care about -- particularly Focused + Disabled. I wouldn't expect our disabled controls to be focusable, so it seems like we'd be adding extra noise in our chromatic diffs if those states change. Can you think of a good way to skip a state like that?

packages/nimble-components/src/toggle-button/tests/toggle-button-matrix.stories.ts
```diff
@@ -47,32 +52,45 @@ type CheckedState = (typeof checkedStates)[number];
...
         partVisibilityStates,
         checkedStates,
         disabledStates,
         appearanceStates
     ])
 );
 
+export const toggleButtonInteractionsThemeMatrix: StoryFn = createMatrixThemeStory(
+    createMatrix(component, [
+        interactionStates.filter(x => x[0] && !x[0].includes('Active')),
```
mollykreis, line 86: It isn't clear to me why we need to skip the `Active` state in these tests. Perhaps a comment could help.

m-akinc: I think I made the mistake of thinking toggle button didn't have active styling, but it does, so I will include active states, too.

change/@ni-nimble-components-ceebd404-0fb3-43e5-8f98-04f4eed64531.json
m-akinc: One option would be to split the interaction states between two stories, where one story would include disabled states with just hover interactions, and the other would include other interactions without disabling. But that creates more stories and doesn't scale well.

Instead, I added an optional `filter` parameter to `createMatrix` which can be used to exclude any unwanted combination of states from the rendered story. For the interaction stories, I can pass a simple filter function that excludes all disabled configs, except if the interaction state is hover (i.e. not active or focused).

There are possibly cases in other non-interaction visual tests that we would want to exclude, too. For example, I added a filter to the menu button matrix test to exclude open + disabled combinations, since that's not really a realistic state.

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
```diff
@@ -76,6 +82,21 @@ export const numberFieldThemeMatrix: StoryFn = createMatrixThemeStory(
     ])
 );
 
+export const numberFieldInteractionsThemeMatrix: StoryFn = createMatrixThemeStory(
+    createMatrix(
+        component,
+        [
+            interactionStates.filter(x => !x[0].includes('Active')), // skip irrelevant active states
```
mollykreis, line 89: Aren't the active states relevant to the inc/dec buttons?

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
```diff
@@ -76,6 +82,21 @@ export const numberFieldThemeMatrix: StoryFn = createMatrixThemeStory(
...
+            disabledStates,
+            hideStepStates.filter(x => x[0] !== 'Hide Step'), // always show inc/dec buttons
+            valueStates.filter(x => x[0] !== 'Placeholder'), // don't test placeholder text
```
mollykreis, line 92: Is there a reason to skip the placeholder tests? Just the assumption that nothing with the placeholder will change in styling based on the interaction?

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
```diff
@@ -76,6 +82,21 @@ export const numberFieldThemeMatrix: StoryFn = createMatrixThemeStory(
...
+            errorStates.filter(x => x[0] !== 'Error No Message'), // don't test error state w/o error text
```
mollykreis, line 93: Similar question here. Is there a reason to skip these tests? Is it just to reduce the number of tests being run since the assumption is that testing with the error messages tests everything (and more) that the "Error No Message" would have tested?

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
```diff
@@ -35,39 +40,83 @@ const partVisibilityStates = [
...
+            appearanceStates
+        ],
+        // Only interaction relevant to disabled controls is hover
+        (
+            interactionState: InteractionState,
+            _partVisibilityState: PartVisibilityState,
+            _openState: OpenState,
+            disabledState: DisabledState
+        ) => {
+            return disabledInteractionsFilter(
```
mollykreis, line 114: Could this use the `disabledInteractionsFilter` directly if the order of the states were changed?

packages/nimble-components/src/toggle-button/tests/toggle-button-matrix.stories.ts
```diff
@@ -47,32 +54,60 @@ type CheckedState = (typeof checkedStates)[number];
...
+            appearanceStates
+        ],
+        // Only interaction relevant to disabled controls is hover
+        (
+            interactionState: InteractionState,
+            _partVisibilityState: PartVisibilityState,
+            _checkedState: CheckedState,
+            disabledState: DisabledState
+        ) => {
+            return disabledInteractionsFilter(
```
mollykreis, line 103: Can this use `disabledInteractionsFilter` directly if the states were reordered?

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
m-akinc: Right. Just to reduce noise, since value vs placeholder don't seem like interesting/distinct test cases.

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
m-akinc: Correct. Just to reduce noise.

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
m-akinc: Yes, I'll reorder them.

packages/nimble-components/src/toggle-button/tests/toggle-button-matrix.stories.ts
m-akinc: Yes, I'll reorder them.

packages/nimble-components/src/number-field/tests/number-field-matrix.stories.ts
mollykreis: Could you add a bit of justification to these comments so that looking at the code it's a bit more clear _why_ the filter is written the way it is rather than just _what_ the filter is accomplishing.

mollykreis (approved):

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
```diff
@@ -35,39 +40,72 @@ const partVisibilityStates = [
...
+            appearanceStates
+        ],
+        // Disabled and open is not a valid state
+        (
+            _interactionState: InteractionState,
+            disabledState: DisabledState,
+            _partVisibilityState: PartVisibilityState,
+            openState: OpenState
+        ) => {
+            return disabledState[0] !== 'Disabled' || openState[0] !== 'Open';
```
rajsite, line 92: I don't think I like this filter feature concept.. Our goal is that all states a control can be put in should behave sanely and consistently (i.e. minimally tested via matrix tests), even if when set they are not fully supported by usage guidance. The behavior should be well-known even if not recommended. This filter concept on the matrix tests encourages the opposite, putting gaps in our testing. We should discuss this outside of PR comments, i.e. post stand-up topic, etc. FYI @jattasNI

packages/nimble-components/.storybook/main.js
```diff
@@ -23,7 +23,8 @@ export const addons = [
         }
     },
     '@storybook/addon-a11y',
-    '@storybook/addon-interactions'
+    '@storybook/addon-interactions',
+    'storybook-addon-pseudo-states'
```
rajsite, line 27: This plugin seems active on every story, and it's behavior does not seem very good / useful too be always enabled.

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
m-akinc: From our discussion in the team meeting, it sounds like the two main points are:
- the main matrix test for a component should not exclude specific combinations, because we need to ensure that no configuration causes things to blow up
- excluding certain configurations from interaction matrix tests is fine, but using a filter function makes it harder to understand exactly what is being excluded

I introduced the filter functionality because, for interactions on disabled controls, I only wanted to test the hover behavior. I.e. there isn't a way for a disabled control to be active and/or focused. I can think of a few directions to go in:
1. Use the filter behavior, like I have in this PR.
2. Break the one interaction test into two: one for disabled, and one for enabled. I don't like that for two reasons: (1) it increases the number of snapshots, and (2) conceptually, it seems like an arbitrary split.
3. Test all interaction states of disabled controls, even if they don't make sense. I dislike this because it makes the snapshot noisier to read.
4. Collapse the `interactionStates` and `disabledStates` matrix dimensions into one, with values like: `enabled`, `enabled-hover`, `enabled-hover-active`, `enabled-active`, `enabled-focus`, `disabled`, and `disabled-hover`. I don't like this idea.

My vote would be to stick with option 1, but I'm open to other ideas or arguments.

packages/nimble-components/.storybook/main.js
m-akinc: I've disabled the toolbar item for the addon in all stories.

packages/nimble-components/src/menu-button/tests/menu-button-matrix.stories.ts
rajsite: I'm working on a 5th alternative which is to break-up matrix into smaller, targeted pieces that compose together but it's still WIP but FYI I'm pushing to this branch directly

rajsite: @m-akinc updated the PR with my proposed approach. Has the following characteristics:
- Interaction states are currently like background colors. Most controls are not going to need control specific integration so made a helper similar to the theme matrix helper.
- Switched from filtering of states to opting into states for each mode of interactions. I think this approach is easier to follow as the states used are explicit. I don't have to jump back and forth between the set of states and mentally running the filter to know what to expect. The states are all listed explicitly.
- There no longer is ordering dependence across dimensions, i.e. the interaction and disabled dimensions don't have to be first and second. Allows for controls to place things in an order that makes sense for the control. For example the menu-button making open states the first dimension.

Let me know if you have any questions or concerns. If not I'll do a final pass before approve.

rajsite: Took another pass through and feel pretty confident with the pattern so resolving the comment. But if want to discuss feel free to re-open 👍

rajsite (approved):

change/@ni-nimble-components-ceebd404-0fb3-43e5-8f98-04f4eed64531.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
```
rajsite, line 2: FYI @mollykreis @jattasNI I mainly did a source code review. I didn't do a deep analysis of if the new interaction matrix stories to see if the subset of states they cover give us confidence in thee coverage or that the stories align with design specs and should be the approved baselines.

jattasNI: I did a pass at looking at the coverage of states. It looked mostly good to me. The only thing I wasn't sure about was the usefulness of the "active" (but not "hovered") state for some components. Is there a user interaction that can hit that state? For example, tabbing to a button and pressing Space/Enter doesn't cause it to render in the active state. Maybe clicking on a component and then dragging away so it's no longer hovered? Ultimately I don't have a problem keeping it out of completeness.

Regarding whether the baselines should be approved, I think Mert mentioned he showed this to Brandon earlier but I might be confusing this PR with another one. If not then this would be a good opportunity to do that again (I assume we did it the last time we updated these styles significantly), but I'd be fine with the PR completing before that happens since this isn't changing anything, just snapshotting our current behavior.

jattasNI (approved):

change/@ni-nimble-components-ceebd404-0fb3-43e5-8f98-04f4eed64531.json
m-akinc: > Maybe clicking on a component and then dragging away so it's no longer hovered?

I think that might be the only way, but I'm inclined to leave it in.

The one off-spec thing I noticed was some of the anchor styling, though the spec has us using a different color for visited links, and I wonder if we actually want that. There are still a couple other differences that I will address in another PR, and I'll ask Brandon about the visited links.

change/@ni-nimble-components-ceebd404-0fb3-43e5-8f98-04f4eed64531.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Add interaction stories for buttons, number field, and anchor",
```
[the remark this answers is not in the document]
m-akinc: I added a sentence about `createMatrixInteractionsFromStates`. There was nothing to delete.

PR #1921, opened 2024-03-11
Update typescript to 4.9.5

m-akinc:
# Pull Request

## 🤨 Rationale

When running `npm i` on `main`, I was seeing a warning about an incompatible peer dependency of cosmiconfig. Updating typescript to version 4.9.5 (the last release before 5.0) fixes this. This version of typescript [is compatible](https://angular.io/guide/versions#actively-supported-versions) with Angular 15.2.

## 👩‍💻 Implementation

Updated dependency version in each `package.json` and lock file.

## 🧪 Testing

Successful build of nimble repo.

## ✅ Checklist


m-akinc:
@jattasNI do you mind being a first-pass reviewer on this?

angular-workspace/package.json
```diff
@@ -60,6 +60,6 @@
     "ng-packagr": "^15.2.2",
     "playwright": "1.40.0",
     "rollup": "^4.12.0",
-    "typescript": "~4.8.2"
+    "typescript": "~4.9.5"
```
jattasNI, line 63: [Angular's docs](https://angular.io/guide/versions#actively-supported-versions) show that this version of TypeScript is compatible with Angular 15.2 👍. Could you mention that in the PR description?

change/@ni-jasmine-parameterized-65d6b534-8705-4960-a2f3-3a8b73e8ba96.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jattasNI, line 2: Since we're only updating `devDependencies` my understanding is that this only impacts the version we build/test against but doesn't impact the published package or requirements for clients. If that's the case then I think we can set `"type": "none"` in all of these files. (It'd be good to get Milan's confirmation on this)

jattasNI (approved):

change/@ni-jasmine-parameterized-65d6b534-8705-4960-a2f3-3a8b73e8ba96.json
m-akinc: That's a good point. I'll make the change, and if Milan or anyone else argues otherwise, I'll change it back.

rajsite: It has the potential to change the build output. The compiler is generating the output code (we don't ship .ts, we ship output from the tsc compiler). So a change in compiler version does impact the output an app consumes. It should be patch.

change/@ni-jasmine-parameterized-65d6b534-8705-4960-a2f3-3a8b73e8ba96.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: making a comment here jic, all these changefiles should be patch.

rajsite (approved):

rajsite:
@m-akinc this PR has sufficient review, don't need to wait on fred or trevor

TrevorKarjanis (approved):

TrevorKarjanis:
> @m-akinc this PR has sufficient review, don't need to wait on fred or trevor

JIT reviewing :sunglasses:

fredvisser (approved):

PR #1928, opened 2024-03-13
Fix intermittent RichTextMentionListbox test

### LINKED ISSUE #1891: Intermittent test: mention-listbox "should scroll the selected option into view when opened"
## 🧹 Tech Debt

This test was disabled because it fails intermittently on the pipeline. ~~It passes consistently locally.~~ It fails locally, too, when all tests are run. It passes when run alone. It's unclear why this test is intermittent, but similar tests (named the same) for select and combobox are not.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1891 

## 👩‍💻 Implementation

Calling `show()` on the `RichTextMentionListbox` results in its anchored region becoming un-hidden, and an intersection observer begins observing the new anchor element (i.e. what the anchored region is attached to). The observer closes the listbox when the anchor element is  scrolled out of view. Depending on the timing, the observer callback can run right after calling `show()`, and if the anchor element is not fully visible in the viewport, then the listbox can be closed before we get a chance to run the rest of the test. Previously, the test had been using the document element as the anchor element, but for some reason that I couldn't discern, the intersection observer would say it was not intersecting (with itself!). I changed the test to use a div (which stays fully in the viewport) as the anchor element, and we no longer run into an issue with the intersection observer prematurely closing the listbox.

## 🧪 Testing

Interestingly, focusing the intermittent test with `fit` would cause it to always succeed. I had to disable all other tests (with `xdescribe`) to run the flaky test by itself. I ran the test with my changes many times without any failures.

## ✅ Checklist


m-akinc:
@mollykreis could you please buddy this for me?

mollykreis (approved):

jattasNI (approved):

rajsite (approved):

PR #1929, opened 2024-03-13
Improve the Wafer Map web worker class to support offscreen rendering

Razvan1928:
# Pull Request

## 🤨 Rationale

Wafer Map's canvas is the main component used to draw the actual dies on a wafer

image

The old wafer renders the canvas on the main thread, for the new one we want to render the canvas inside the web worker

## 👩‍💻 Implementation

The main changes were done in matrix-renderer.ts, worker-renderer.ts and index.ts
matrix-renderer.ts - holds the logic on how the canvas drawn, most of the changes are straightforward besides the typed array parsing algorithm from drawWafer(), but a comment explaining it can be found there
worker-renderer.ts - holds the logic on how worker is setup
index.ts - creates the worker and transfers the canvas control from main thread to the worker

less important changes were done in WaferMapTests.ts from the Blazor component and zoom-handler.ts
WaferMapTests.ts - after adding a new wafer map tag inside the wafermap template.ts the blazor tests were not able to distinguise between the 2 tags, I added IDs to each tag and changed the page.Locator to search for the ID and not for the tag name
zoom-handler.ts - zoom behavior was managed inside update() method from index.ts, as the constructor of ZoomHandler takes an wafer map as input we implemented the observable design pattern using Notifier from Fast Element, similar to how it is done for table nimble component

## 🧪 Testing

Added a few unit tests and a chromatic test

## ✅ Checklist


packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -247,6 +269,49 @@ export class WaferMap extends FoundationElement {
                 this.renderer.drawWafer();
             }
             this.zoomHandler.connect();
+            if (this.dataManager instanceof ExperimentalDataManager) {
```
atmgrifter00: Can you pull this implementation for the if block into a private method, just for readability and consistency's sake?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -246,12 +268,63 @@ export class WaferMap extends FoundationElement {
...
                 this.renderer.drawWafer();
             }
+            await this.drawWorkerWafer();
             this.zoomHandler.connect();
         } else if (this.waferMapUpdateTracker.requiresRenderHoverUpdate) {
             this.renderer.renderHover();
         }
     }
 
+    private async drawWorkerWafer(): Promise<void> {
```
atmgrifter00, line 278: I think I would just call this `drawWafer`...the fact that a worker is being used is an implementation detail.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,128 @@ import { expose } from 'comlink';
...
+    public scaledColIndex = new Float64Array([]);
+    public scaledRowIndex = new Float64Array([]);
+    private scaleX: number = 1;
+    private scaleY: number = 1;
+    private baseX: number = 1;
+    private baseY: number = 1;
+    private dieDimensions: Dimensions = { width: 1, height: 1 };
+    private transform: Transform = { k: 1, x: 0, y: 0 };
+    private topLeftCanvasCorner: { x: number, y: number } = { x: 0, y: 0 };
+    private bottomRightCanvasCorner: { x: number, y: number } = { x: 500, y: 500 };
```
atmgrifter00, line 26: Where do these initial values come from?

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: did it image

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: indeed, I renamed it

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -5,15 +5,17 @@ import { HoverDieOpacity } from '../../types';
...
 
     public updateSortedDiesAndDrawWafer(): void {
         // redundant function for backwards compatibility
         this.drawWafer();
     }
 
     public drawWafer(): void {
-        // rendering will be implemented in a future PR
+        this.wafermap.workerOne.setTransform(this.wafermap.transform).then(() => {
+            this.wafermap.workerOne.drawWafer().then(() => { }, () => { });
```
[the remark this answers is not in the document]
Razvan1928: It was a work around to not make drawWafer in that moment to run async, now I deleted them the .then part and added async/await 
image

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -74,11 +78,21 @@ export class WaferMap extends FoundationElement {
     @attr({ attribute: 'color-scale-mode' })
     public colorScaleMode: WaferMapColorScaleMode = WaferMapColorScaleMode.linear;
 
+    /**
+     * @internal
+     */
+    public workerOne!: Remote<MatrixRenderer>;
```
atmgrifter00, line 84: I would prefer a different suffix than `One`. Does this member even need a suffix?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -74,11 +78,21 @@ export class WaferMap extends FoundationElement {
...
+
     /**
      * @internal
      */
     public readonly canvas!: HTMLCanvasElement;
 
+    /**
+     * @internal
+     */
```
atmgrifter00, line 93: Why are we exposing two different canvas elements on the API surface (particularly when they are the same type)?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -74,11 +78,21 @@ export class WaferMap extends FoundationElement {
...
+    public readonly canvasOne!: HTMLCanvasElement;
```
atmgrifter00, line 94: Why are we exposing two different canvas elements on the API surface (particularly when they are the same type)?

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Once we had 2 workers, meanwhile we deleted the second. Indeed it doesn't make sense now that we have only one to name it workerOne, I will deleted the suffix

packages/nimble-components/src/wafer-map/tests/create-matrix-renderer.spec.ts
```diff
@@ -18,19 +18,28 @@ describe('MatrixRenderer worker', () => {
...
+            Uint32Array.from(testData.rowIndexes)
+        );
+        expect(resolvedDieMatrix.values).toEqual(
+            Float64Array.from(testData.values)
         );
     });
 
     it('emptyMatrix should empty the dieMatrix', async () => {
-        const testData = [4, 5, 6];
+        const testData = { colIndexes: [4, 1, 2], rowIndexes: [54, 54, 62], values: [8.12, 9.0, 0.32] };
```
atmgrifter00, line 37: If the test data contents don't really get checked against the specific values you are using (across multiple tests), then can we just declare a `testData` const ouside of the tests that we can use in multiple tests?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -244,14 +266,65 @@ export class WaferMap extends FoundationElement {
...
+                .getChild('colIndex')
+                ?.toArray() as number[],
+            rowIndexes: this.diesTable
+                .getChild('rowIndex')
+                ?.toArray() as number[],
+            values: this.diesTable.getChild('value')?.toArray() as number[]
+        } as WaferMapMatrix;
+        await this.workerOne.updateMatrix(waferMapMatrix);
+        await this.setupWorker();
+        await this.workerOne.drawWafer();
```
atmgrifter00, line 293: This method isn't `async` right?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -244,14 +266,65 @@ export class WaferMap extends FoundationElement {
...
+        const waferMapMatrix = {
+            colIndexes: this.diesTable
+                .getChild('colIndex')
+                ?.toArray() as number[],
+            rowIndexes: this.diesTable
+                .getChild('rowIndex')
+                ?.toArray() as number[],
+            values: this.diesTable.getChild('value')?.toArray() as number[]
+        } as WaferMapMatrix;
+        await this.workerOne.updateMatrix(waferMapMatrix);
```
atmgrifter00, line 291: This method isn't async right?

atmgrifter00, line 291: There seem to be a bunch of `await` calls on methods being called on `workerOne`, but _none_ of them are actually `async` right?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -244,14 +266,65 @@ export class WaferMap extends FoundationElement {
                 this.dataManager.updateDiesRenderInfo();
                 this.renderer.updateSortedDiesAndDrawWafer();
             } else if (this.waferMapUpdateTracker.requiresDrawnWaferUpdate) {
-                this.renderer.drawWafer();
+                await this.renderer.drawWafer();
             }
+            await this.drawWafer();
             this.zoomHandler.connect();
```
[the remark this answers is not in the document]
Razvan1928: indeed, I left it there since the beginning to be sure it will be always called, I moved it to a better place 
image

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Did a bit of research, the Comlink wrapper makes the methods appear as asynchronous from the main thread perspective, though this doesn't make them to run asynchronously inside the worker, good catch! I'll make them async

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -74,11 +78,21 @@ export class WaferMap extends FoundationElement {
...
+
     /**
      * @internal
      */
     public readonly canvas!: HTMLCanvasElement;
 
+    /**
+     * @internal
+     */
+    public readonly workerCanvas!: HTMLCanvasElement;
```
atmgrifter00, line 94: I still don't know why we are exposing both `canvas` and `workerCanvas` on the public API. Why is there a need for two distinct canvas elements? If there are different setup requirements for the element between the two scenarios, can't we just perform the appropriate setup as needed when we switch between them?

packages/nimble-components/src/wafer-map/index.ts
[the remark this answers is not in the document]
atmgrifter00: Just to make sure I'm not missing something, are you intending to move `this.zoomHandler.connect()`, or were you referring to something else?

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: I was preparing an explanation for this, when sharing the same canvas both wafers, stable and experimental, stop working, not sure why, I'll give an update

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -244,14 +267,64 @@ export class WaferMap extends FoundationElement {
                 this.dataManager.updateDiesRenderInfo();
                 this.renderer.updateSortedDiesAndDrawWafer();
             } else if (this.waferMapUpdateTracker.requiresDrawnWaferUpdate) {
-                this.renderer.drawWafer();
+                await this.renderer.drawWafer();
             }
             this.zoomHandler.connect();
```
atmgrifter00, line 272: While I understand the change you made, I'm still curious about how the `zoomHandler` actually operates. Why is it necessary to recreate the `ZoomBehavior` for all updates tracked by `requiresEventUpdate`? That tracked event seems to be a catch-all for most other tracked state changes (for some reason), including things that don't seem related to state that would require a re-creation of the `ZoomBehavior`.

packages/nimble-components/src/wafer-map/tests/create-matrix-renderer.spec.ts
Razvan1928: indeed, I moved it outside the tests
image

packages/nimble-components/src/wafer-map/index.ts
atmgrifter00: So, I think you may have been too hasty with the changes you've made. After talking with @rajsite, it's fine and expected that the worker method implementations need not be written as `async` if there is nothing fundamentally `async` about them (which there isn't), and it is also expected that you must await those methods exposed via the Comlink wrapper.

Razvan1928: Indeed, I'll revert the changes

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Meanwhile I managed to use only one canvas

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Explanation can be found in a different comment bellow

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Indeed it is not needed to connect and disconnect at every update, only when the width/height of the canvas is needed, I will update the code accordingly

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: I left them there to avoid the undefined kind of errors. I deleted the default values and put ! at the end of the properties, not sure if this is a proper solution

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -192,14 +201,25 @@ export class WaferMap<
...
-        this.canvasContext = this.canvas.getContext('2d', {
-            willReadFrequently: true
-        })!;
         this.hoverHandler.connect();
         this.experimentalHoverHandler.connect();
         this.zoomHandler.connect();
+        const { matrixRenderer } = await createMatrixRenderer();
+        this.worker = matrixRenderer;
+        if (this.diesTable !== undefined) {
+            const offscreenOne = this.canvas.transferControlToOffscreen();
```
atmgrifter00, line 212: Can we rename `offscreenOne` to `offscreenCanvas`?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,128 @@ import { expose } from 'comlink';
...
+    public scaledRowIndex = new Float64Array([]);
+    public canvas!: OffscreenCanvas;
+    public context!: OffscreenCanvasRenderingContext2D;
+    private scaleX!: number;
+    private scaleY!: number;
+    private baseX!: number;
+    private baseY!: number;
+    private dieDimensions!: Dimensions;
+    private transform!: Transform;
+    private topLeftCanvasCorner!: { x: number; y: number; };
```
atmgrifter00, line 25: Yeah, it's not proper to declare everything as a non-nullable value if they are not at least initialized in the constructor (or in some other specific use-cases like elements that are defined in a FAST template). To me, the alternative is to provide expected defaults for some of the values (like `scaleX` and `scaleY` which I presume can be 1), and for others where there is no obvious default (like `topLeftCanvasCorner`) then the two choice are:
- Provide the values as constructor parameters (and note that there could even be a data structure that makes sense to create that captures all of the necessary initial values in one object).
    - This option doesn't seem applicable with the way the `MatrixRenderer` is initialized in practice.
- Leave the values as nullable, which then requires appropriate checks in the code against those fields.

atmgrifter00:
@Razvan1928 @munteannatan, with regards to the failing tests, the first point of order, I believe, is to remove the `async` signature applied to the `connectedCallback` in `wafer-map.ts`, as that method is not async, and nothing will await it. This will likely require moving the relevant async calls within into a private method and then executing them with `void` (i.e. `void (async () => { ... } )();`). Note that this isn't sufficient to prevent the race condition.

I'm not sure what mechanism is supposed to exist that will actually block execution until the `MatrixRenderer` is created, but this seems to be the core problem.

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: sure!

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -225,40 +246,33 @@ export class WaferMap<
...
-            }
-            this.zoomHandler.connect();
+        // zoom translateExtent needs to be recalculated when canvas size changes
+        this.dataManager = this.isExperimentalRenderer()
+            ? this.experimentalDataManager
+            : this.stableDataManager;
+        if (this.waferMapUpdateTracker.requiresContainerDimensionsUpdate) {
+            this.dataManager.updateContainerDimensions();
+            await this.renderer.updateSortedDiesAndDrawWafer();
+        } else if (this.waferMapUpdateTracker.requiresScalesUpdate) {
```
atmgrifter00, line 265: Is it appropriate for this to be all `else if` blocks? Is it possible for there to be more than one aspect to be tracked? Generally this is expected to be the case, which is why we needed an update tracker, because more than one thing can change state that should both be processed at the same time.

atmgrifter00:
> @Razvan1928 @munteannatan, with regards to the failing tests, the first point of order, I believe, is to remove the `async` signature applied to the `connectedCallback` in `wafer-map.ts`, as that method is not async, and nothing will await it. This will likely require moving the relevant async calls within into a private method and then executing them with `void` (i.e. `void (async () => { ... } )();`). Note that this isn't sufficient to prevent the race condition.
> 
> I'm not sure what mechanism is supposed to exist that will actually block execution until the `MatrixRenderer` is created, but this seems to be the core problem.

So it seems that the test are now passing because the `drawWafer` method is no longer in the code path of execution for the test (which I'm guessing is appropriate). However, I just want to be clear that the fundamental timing issues are still present, and that `connectedCallback` should not be declared as `async`. Ultimately, if you wrote a test that relied on state being set in `connectedCallback` after the `MatrixRenderer` was created (or tried to execute any method on the `MatrixRenderer`), that test would likely still fail.

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Discussed in private

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
  * This setup is used in the wafer-map component to perform heavy computational duties
  */
 export class MatrixRenderer {
-    public dieMatrix: Uint8Array = Uint8Array.from([]);
+    public colIndexes: Int32Array = Int32Array.from([]);
+    public rowIndexes: Int32Array = Int32Array.from([]);
+    public values = new Float64Array([]);
+    public scaledColIndex = new Float64Array([]);
```
atmgrifter00, line 15: Why are we using two different ways to initialize an empty array:
~~~
public colIndex: Int32Array = Int32Array.from([]);
and
public scaledColIndex = new Float64Array([]);
~~~

Additionally, this should be called `scaledColumnIndexes` (and below would be `scaledRowIndexes`).

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
  * This setup is used in the wafer-map component to perform heavy computational duties
  */
 export class MatrixRenderer {
-    public dieMatrix: Uint8Array = Uint8Array.from([]);
+    public colIndexes: Int32Array = Int32Array.from([]);
```
atmgrifter00, line 12: Let's go ahead and spell out `column`.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+    public context!: OffscreenCanvasRenderingContext2D;
+    private scaleX: number = 1;
+    private scaleY: number = 1;
+    private baseX: number = 1;
+    private baseY: number = 1;
+    private dieDimensions: Dimensions = { width: 1, height: 1 };
+    private transform: Transform = { k: 1, x: 0, y: 0 };
+    private topLeftCanvasCorner!: { x: number; y: number; };
+    private bottomRightCanvasCorner!: { x: number; y: number; };
+    private margin: { top: number; right: number; bottom: number; left: number; } = { top: 20, right: 20, bottom: 20, left: 20 };
```
atmgrifter00, line 28: Why not just default these values to 0, particularly if it's expected that the `margin` will invariably be set to some non-0 value before use anyway? Is a 0 margin value non-sensical? If it is, then I would prefer we use a named constant for the value 20. Is there anywhere else in the code that is using this value of 20?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+
+    public setColIndexes(colIndexesBuffer: Int32Array): void {
+        this.colIndexes = colIndexesBuffer;
+        const scaledColIndex = [this.scaleX * this.colIndexes[0]! + this.baseX + this.margin.left];
```
atmgrifter00, line 32: By making the non-null assertion of `this.colIndexes[0]!` you are making an assumption that the caller will always provide a non-empty array. This doesn't seem like a necessary assumption to make. An early return when the array is empty seems reasonable.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+        const colPosition = [0];
+        let prev = this.colIndexes[0];
+        for (let i = 1, length = this.colIndexes.length; i < length; i++) {
```
atmgrifter00, line 35:
```suggestion
        for (let i = 1; i < this.colIndexes.length; i++) {
```
I don't see a usage of the `length` variable other than within this `for` statement.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+            const xIndex = this.colIndexes[i];
+            if (xIndex && xIndex !== prev) {
+                const scaledX = this.scaleX * this.colIndexes[i]! + this.baseX + this.margin.left;
+                scaledColIndex.push(scaledX);
```
atmgrifter00, line 39: Shouldn't `scaledColIndex` (or, again, better named `scaledColumnIndexes`) be cleared at the beginning of this function?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
atmgrifter00, line 32: I think you should pull out this math into a helper function:

```
private calcuatedScaledColumnIndex(columnIndex: number): number {
    return this.scaleX * columnIndex + this.baseX + this.margin.left;
}
```
You can then call this method both here and below.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+    private dieDimensions: Dimensions = { width: 1, height: 1 };
+    private transform: Transform = { k: 1, x: 0, y: 0 };
+    private topLeftCanvasCorner!: { x: number; y: number; };
+    private bottomRightCanvasCorner!: { x: number; y: number; };
+    private margin: { top: number; right: number; bottom: number; left: number; } = { top: 20, right: 20, bottom: 20, left: 20 };
+
+    public setColIndexes(colIndexesBuffer: Int32Array): void {
+        this.colIndexes = colIndexesBuffer;
+        const scaledColIndex = [this.scaleX * this.colIndexes[0]! + this.baseX + this.margin.left];
+        const colPosition = [0];
```
atmgrifter00, line 33: `columnPositions`

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+            this.canvas.height
+        );
+    }
+
+    public drawWafer(): void {
+        this.context.restore();
+        this.context.save();
+        this.clearCanvas();
+        this.scaleCanvas();
+        if (this.topLeftCanvasCorner === undefined || this.bottomRightCanvasCorner === undefined) { return; }
```
atmgrifter00, line 140: It might be better to throw an `Error` instead of just returning, as this is indicative of an improper configuration, yes?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,158 @@ import { expose } from 'comlink';
...
+        for (let i = 0; i < this.scaledColIndex.length; i++) {
+            const scaledX = this.scaledColIndex[i]!;
+            if (
+                !(scaledX >= this.topLeftCanvasCorner.x
+                    && scaledX < this.bottomRightCanvasCorner.x)
+            ) {
+                continue;
+            }
+            for (let j = this.colIndexPositions[i]!,
+                length = this.colIndexPositions[i + 1] !== undefined ? this.colIndexPositions[i + 1]! : this.scaledRowIndex.length;
```
atmgrifter00, line 150: <nit> I think I would prefer that we pull the `length` variable out of the `for` statement, and maybe rename it so that it is less generic. I'll mention that I don't think this is very easy to understand. As written, it appears that the values in `this.colIndexPositions` have two meanings. One is as an index into `this.scaledRowIndex` (line 152), and the other as a _possible_ length to iterate over. Can you explain this? Perhaps in a code comment?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -77,6 +80,21 @@ export class WaferMap<
     @attr({ attribute: 'color-scale-mode' })
     public colorScaleMode: WaferMapColorScaleMode = WaferMapColorScaleMode.linear;
 
+    /**
+     * @internal
+     */
+    public isWorkerAlive = false;
```
atmgrifter00, line 86: Why is this state being kept here, when the only thing that cares about it is the `WorkerRenderer`?

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -225,40 +243,38 @@ export class WaferMap<
      * The updates snowball one after the other, this function only choses the 'altitude'.
      * The hover does not require an event update, but it's also the last update in the sequence.
      */
-    public update(): void {
+    public async update(): Promise<void> {
+        this.zoomHandler.connect();
```
atmgrifter00, line 247: Why is calling `zoomHander.connect()` the right thing to do on _every_ update? Doesn't this add an event listener on every call? This seems like another case where a class should be leveraging notifiers in order for it to update as needed. For the `ZoomHandler`, it seems like it should be observing the `WaferMap`'s `canvasWidth`, `canvasHeight`, and `transform` properties, and when they change, calling `createZoomBehavior()`.

If this is deemed correct, but too large to handle for this PR (as it's tangential), I think we need another issue filed against this specifically.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -225,40 +243,38 @@ export class WaferMap<
...
-                this.dataManager.updateLabelsFontSize();
-                this.renderer.updateSortedDiesAndDrawWafer();
-            } else if (
-                this.waferMapUpdateTracker.requiresDiesRenderInfoUpdate
-            ) {
-                this.dataManager.updateDiesRenderInfo();
-                this.renderer.updateSortedDiesAndDrawWafer();
-            } else if (this.waferMapUpdateTracker.requiresDrawnWaferUpdate) {
-                this.renderer.drawWafer();
+        // zoom translateExtent needs to be recalculated when canvas size changes
```
atmgrifter00, line 255: Is this comment relevant anymore?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: The difference between Int32Array.from([]) and new Int32Array([]) can be observed in a JS/TS environment that is not strictly configured, Int32Array.from([]) returns for sure an Int32Array while new Int32Array([]) because of the empty parameter it might be seen as a regular JS array and be inferred to an regular JS array (not sure why, didn't get deep into it). The different ways of declaration in code though doesn't have an explanation, I was experimenting with different kind of declaration when things weren't working and were unintentionally left. I will use .from([]) for all. Regarding the naming, I will rename the property.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: Renamed them

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: Did it
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: image
indeed, deleted length

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: sure, did it 
image
image
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: You are right! I threw an error
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: renamed it!
image

Razvan1928: image
renamed it!

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: seems like it is obsolete, I removed it

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: The smallest wafer that can be generated has 13 dies, a margin of 20 is suited for it, I will put it in a separate variable
, with 0 it looks like this
image

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: So we have 3 areas that the user might keep its mouse on.
image
Leaving `zoomHander.connect()` called on every update will cover all the areas. Trying to narrow the cases when zoom is connected while maintaining a similar behavior I arrived to the following:
Aria 1 - Can move `zoomHander.connect()` inside `hoverDieChanged()` observer, this leads to an unpleasant behavior as the user is able to zoom in for the first ~1 second while hovering after which if the user keep the mouse on the same die the observer will not be called again leading to not being able to zoom in.
Aria 2 - We don't have observers to track this area, though it doesn't contain any die I image a case when the user tries to zoom in near the corner and may have a hard time hovering a die considering we expect wafer up to 300K dies
Aria 3 - Don't think it's relevant to zoom

Calling `this.hoverHandler.disconnect()` at every `update()` makes the things complicated, though we still want to still want to delete the event listener.
Will leave the discussion open, @munteannatan any opinions?

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -7,13 +8,51 @@ import { HoverDieOpacity } from '../../types';
...
+        );
+        await this.wafermap.worker.setRowIndexes(
+            transfer(rowIndexes, [rowIndexes.buffer])
+        );
+        await this.drawWafer();
     }
 
-    public drawWafer(): void {
-        // rendering will be implemented in a future PR
+    public async drawWafer(): Promise<void> {
```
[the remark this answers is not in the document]
Razvan1928: Had to do many changes after @munteannatan pushed his latest pr, can you please take a look if the comment still stands? Now `await this.setupWorker()` is called inside `updateSortedDies()` which is called before `drawWafer()`
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: I named j and length into columnStartIndex and columnEndIndex and pulled columnEndIndex outside the for, will add a comment tommorow with more details

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: I dicussed with @munteannatan and moving both disconnect and connect inside the first if preserves the behavior 
image

Razvan1928: I tried to implement your feedback about observables, now canvasWidthChanged() and canvasHeightChanged() handles zooming
image

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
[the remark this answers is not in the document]
Razvan1928: (`updateSortedDies` is now `setupAndDrawWaferis`) I think I got your point, 
image
one concern might be calling `drawWafer()` before `setupAndDrawWafer()` was called therefor not being able to draw the wafer because the worker is undefined, so far we didn't have any scenario when this happened.

packages/nimble-components/src/wafer-map/index.ts
atmgrifter00: So, we talked off-line about this and @Razvan1928 will try and implement the `Notifier` pattern in the `ZoomHandler` model so that the `WaferMap` doesn't have to call `connect` and `disconnect` directly in this way.

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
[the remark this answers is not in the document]
atmgrifter00: I think at a minimum, the `drawWafer` method should be doing some validation checks that it _can_ render. This could simply mean doing a null check on the `worker` field, and if it is null, then throw an Error.

Having a non-internal public method that _can_ be called out of an expected, required order of another public method is fairly non-standard. Typically, required setup work is handled in a constructor (or in `connectedCallback`), but if that is not possible for technical reasons, then I think we must come up with a more robust validation strategy (like simply throwing an error).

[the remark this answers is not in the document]
Razvan1928: I added the check 
image

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Implemented!

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,172 @@ import { expose } from 'comlink';
...
+    public clearCanvas(): void {
+        this.context.clearRect(
+            0,
+            0,
+            this.canvas.width,
+            this.canvas.height
+        );
+    }
+
+    public drawWafer(): void {
```
atmgrifter00, line 145: @Razvan1928 is this implementation being verified through existing tests?

Razvan1928: Now that you are saying I added only one test for this pr, will add a few more

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
```diff
@@ -31,6 +37,17 @@ export class ZoomHandler {
     public disconnect(): void {
         zoom().on('zoom', null)(select(this.wafermap as Element));
         this.wafermap.removeEventListener('wheel', this.onWheelMove);
+        this.wafermapNotifier.unsubscribe(this, 'canvasWidth');
```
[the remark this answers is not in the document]
Razvan1928: did it!
image

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: Added 4 more tests, can be found in `matrix-renderer.spec.ts`

atmgrifter00: @Razvan1928, I'm unresolving this just to get some clarity. While I appreciate the added tests, I'm still curious if the `drawWafer` method (particular the one for the worker-renderer) is being tested anywhere.

atmgrifter00: After talking with Razvan, this implementation will be verified through Chromatic tests.

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
```diff
@@ -31,6 +37,17 @@ export class ZoomHandler {
...
+        this.wafermapNotifier.unsubscribe(this, 'canvasHeight');
+    }
+
+    public handleChange(source: WaferMap, propertyName: string): void {
+        if (
+            source === this.wafermap
+            && (propertyName === 'canvasWidth' || propertyName === 'canvasHeight')
+        ) {
+            this.createZoomBehavior();
```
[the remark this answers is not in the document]
Razvan1928: I added 2 tests regarding canvas width and height changes
image

atmgrifter00:
@munteannatan before approving, I would prefer that this PR has your blessing.

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -51,9 +51,11 @@ export class Prerendering {
         const isDieRenderInfo = (
             info: DieRenderInfo | null
         ): info is DieRenderInfo => info !== null;
-        this._diesRenderInfo = this.wafermap.dies
-            .map(die => this.computeDieRenderInfo(die))
-            .filter(isDieRenderInfo);
+        if (this.wafermap.diesTable === undefined) {
```
munteannatan, line 54: maybe you can use teh isExperimental() check here

munteannatan (approved):

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/modules/prerendering.ts
Razvan1928: yes I can, did it

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -1,4 +1,5 @@
 import { expose } from 'comlink';
```
rajsite, line 1: Make sure prettier is run on this file, the prettierignore needs to be updated so that prettier runs on build/generate-workers/source

https://github.com/ni/nimble/blob/833a75fbdfcc23fb9a9f7964fef957818cb0e9d7/packages/nimble-components/.prettierignore#L12-L17

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -8,16 +9,179 @@ import { expose } from 'comlink';
  * This setup is used in the wafer-map component to perform heavy computational duties
  */
 export class MatrixRenderer {
-    public dieMatrix: Uint8Array = Uint8Array.from([]);
+    public columnIndexes = Int32Array.from([]);
```
rajsite, line 12: There is a lot of state and operations being added to this file. Considerable work was done to break-up the logic for the main wafer-map code into different parts (data-manager, prerendering, rendering). Is there a clear idea of how the logic in here will be organized? We won't end up with one giant class will we? @Razvan1928 @munteannatan

msmithNI (approved):

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -279,6 +287,18 @@ export class WaferMap<
         return this.diesTable !== undefined;
     }
 
+    public async createWorker(): Promise<void> {
```
jattasNI, line 290: I'm guessing these are internal too, if so could you mark them with the same tag as other internal functions?

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -1,14 +1,65 @@
...
+        await this.drawWafer();
+    }
+
+    public async drawWafer(): Promise<void> {
+        if (this.wafermap.diesTable === undefined) {
+            return;
+        }
+
+        if (this.wafermap.worker === undefined) {
+            throw new Error('Worker is not defined');
```
jattasNI, line 39: It's pretty rare that we throw exceptions in Nimble because we don't ever want to leave clients components in a bad state. I just want to double check that a client hitting this would indicate a programming error and not a case we think anyone would likely hit.

jattasNI (commented):
Generally looking good but I'd like to take another look once the existing feedback is addressed.

change/@ni-nimble-blazor-1cbfc49b-865a-4368-ab38-15ee937d0420.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added an additional canvas in wafer map template caused the blazor tests to fail due to missing unique identifiers for canvases, adding different ids for the canvases and explicitly telling the tests to look for one canvas id solved the issue",
```
rajsite, line 3: @Razvan1928 I did not get a chance to finish reviewing this today but expect to look closer tomorrow. Some of the async setup patterns I think need to be looked at a bit more closely. In particular passing a Promise to Dom.queueUpdate seems strange as [queueUpdate is not async aware](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-element/src/dom.ts#L55). I don't have any concrete suggestions yet but that is the area I'm concerned about.

packages/nimble-components/src/wafer-map/template.ts
```diff
@@ -21,7 +21,8 @@ export const template = html<WaferMap>`
                 </g>
             </g>
         </svg>
-        <canvas class="wafer-map-canvas" ${ref('canvas')}></canvas>
+        <canvas id="main-wafer" ${ref('canvas')}></canvas>
```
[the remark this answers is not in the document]
Razvan1928: I switched the ids with classes as you recommended. My initial reasoning was that we'll not use these classes for more than one tag so they can be ids, though I don't thing it brings much value, being consistent along all the nimble components is seems more important.

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: Yes, they are meant to be used only inside wafer-map component, I modified them
image

change/@ni-nimble-components-f866e449-14a0-410d-be26-c5570bf1006d.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
Razvan1928: Seems like my last PRs with a glimpse of description left a mark on you 😂, the more detailed description is a product of your feedback from the performance review, I really appreciate you made yourself some time to give it 🙂

change/@ni-nimble-blazor-1cbfc49b-865a-4368-ab38-15ee937d0420.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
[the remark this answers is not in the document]
Razvan1928: Oh, so the only changes that should have a type are the ones that have impact on the user, I understand, I put 'none' there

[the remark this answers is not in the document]
jattasNI: Yeah, it's all about tracking what bits are included in the built npm package. Since we don't ship our tests, the build output isn't changed so we can mark it "none" and the beachball publishing tool won't even bother releasing a new version.

(for this PR it's actually irrelevant because the nimble-components change will trigger a new NimbleBlazor package anyway but still good to be correct)

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: Considering the changes that we have in plan to push for the wafer map despite the ones inside this pr (adding text on dies, coloring the dies and dies highlighting) I expect the file to reach 300-350 lines, is it too much or should we try to split it into multiple files?

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
jattasNI: I thought about this some more and I think I'd advocate for removing the check for `undefined` and the `throw`. Instead you can use the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (`!`) to tell the compiler that you know the value can never be undefined. Even if that assertion turns out to be incorrect, at runtime JS will throw its own error when something tries to use the undefined value so you'll still get equivalent runtime behavior as you have now, but with less code.

If you want to avoid using `!` everywhere, you could do `const worker = this.wafermap.worker!;` once at the beginning of the method and then use the `worker` variable throughout.

jattasNI: A slightly bigger refactor to consider would be for this renderer to own the worker rather than the wafer map. This class is the one that does most of the work on the worker so it's the one that cares about its lifecycle the most. Why should it have to keep asking the component for something that it could own itself?

From a quick scan I think this would be a better direction, but I'll leave it up to you and @munteannatan to evaluate if it is feasible.

packages/nimble-components/src/wafer-map/template.ts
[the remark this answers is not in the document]
jattasNI: @Razvan1928 I'm not seeing that change here or in tests which are still looking up elements by ID. Is it possible it hasn't been pushed yet?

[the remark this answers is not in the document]
Razvan1928: should be in this commit https://github.com/ni/nimble/pull/1929/commits/2c85816fca26aac064c9fb12b5192f5165e17059 , maybe you need to refresh?

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
Razvan1928: I deleted the null-check and throwing and refactored the code so worker-renderer owns the worker, the changes can be seen here https://github.com/ni/nimble/pull/1929/commits/e9d99ae9ae3ccc5d1972be552efaceb0b0c919ee

packages/nimble-components/src/wafer-map/tests/wafer-map-matrix.stories.ts
```diff
@@ -179,6 +180,12 @@ const componentWaferWithGridDimensions = ([
 >
 </${waferMapTag}>`;
 
+const componentWaferMapWorkerCanvas = (): ViewTemplate => html`<${waferMapTag}
```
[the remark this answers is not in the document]
Razvan1928: We'll keep that in mind, unfortunately I can reproduce none of the chromatic tests using the current worker rendered wafer, but the next pr will add coloring and text for each die and we will be able to match most of the existing tests.

packages/nimble-components/src/wafer-map/template.ts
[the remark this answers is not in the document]
jattasNI: I see it now, thanks. Either my browser or my brain needed a refresh 😆

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
jattasNI: I think the config change you made is a bit too aggressive; now we will be running prettier on any build output that's JSON or TS, which might include files generated by other tools that are never reviewed by humans. Can you exclude just the directory that is in source control, `build/generate-workers`?

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
jattasNI: Thanks, I like how that turned out.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
rajsite: ... it's not about number of lines, it's about separation of respponsibilities and consistency.

If "MatrixRenderer" is analogous in responsibilities to ["RenderingModule"](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/wafer-map/modules/rendering.ts) then right now if I compare the two they look very different. Rendering Module does not have all state dumped in one file, instead element specific state is on a class representing the Elements state (WaferMap), there is a reference to dataManager which references computations and preerendering, etc.

I'd like to understand at a high-level why this Rendering class would be so different and not try to orgnaize state and logic similarly.

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -1,14 +1,65 @@
...
+        if (this.wafermap.diesTable === undefined) {
+            return;
+        }
+        await this.setupWorker();
+        const columnIndexes = Int32Array.from(
+            this.wafermap.diesTable.getChild('colIndex')!.toArray()
+        );
+        const rowIndexes = Int32Array.from(
+            this.wafermap.diesTable.getChild('rowIndex')!.toArray()
+        );
```
rajsite, line 27: If I understand correctly this results in iterating across the data (potentially pretty slowly?) and making a copy into a new array. I'm a bit surprised by making a copy, is that intentional long-term?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: I'm trying to find a way to get rid of that much state, but I don't yet have a solution, I'm gonna explain you the situation and maybe find a solution together. (or leave it as it as, we'll see) Most of the properties of `matrix-renderer.ts` are the same as the original wafermap, so passing a wafer through the constructor would get rid of a lot of state, but I cannot pass a WaferMap object to `matrix-renderer.ts` as the directory `nimble-components\build` is build before `nimble-components\src` and cannot reference any class from there. Creating a new `WaferMap` type inside `nimble-components\build` directory doesn't sound like a good idea. `nimble-components\build\generate-workers` should be built before because rollup packs up the `dist` folder of `generate-workers` and `index.ts` moves it to `nimble-components\src\wafer-map\workers\matrix-renderer.ts`, file without which by the time `src/wafer-map` is build will cause an error, as `src/wafer-map/modules/create-matrix-renderer.ts` references `nimble-components\src\wafer-map\workers\matrix-renderer.ts`. Regarding the separation of concerns, we somehow didn't think about doing it for `matrix-renderer.ts` , I think we can split the file similarly to how wafer-map is organized rendering/prerendering/data-manager etc, will try to do a prototype to see how it works. Unfortunately @munteannatan is OOO until Monday and can't give advise.

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: you are right, solved in this commit https://github.com/ni/nimble/pull/1929/commits/0486013906ea5f2169c549c9cf6e1d396840d2df

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
rajsite: Discussed in person and clarified that the goal isn't necessarily to share the exact same code but to follow the same patterns. It does not make sense to try and leverage the Wafer Map Element code in the worker since there is no DOM in the worker, etc. Instead the goal is to have similar organizational constructs by splitting the state and logic into different parts.

This stackblitz shows a high-level example of splitting up the state: https://stackblitz.com/edit/comlink-split-data?file=src%2Fmain.ts,src%2Fouter-state.ts,src%2Finner-state.ts
- The example has an outer object that creates an inner state object (there could be multiple such as data-manager, prerendering, rendering, etc)
- Then the inner state object is proxied to the main thread. The main thread is able to set state on the inner state objects and invoke methods on the outer state object which uses the inner state synchronously in the same thread.

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
Razvan1928: I wasn't remembering correctly, I modified the code as you suggested https://github.com/ni/nimble/pull/1929/commits/5c74f692678137e2ab454535360e750a544c259f

change/@ni-nimble-blazor-1cbfc49b-865a-4368-ab38-15ee937d0420.json
Razvan1928: Solved in this commit https://github.com/ni/nimble/pull/1929/commits/d1dd606b21eb59b19f9a30d6eb5771ed64effd25

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
Razvan1928: As was discussed yesterday with @munteannatan this will be handled in a different PR

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -241,12 +249,15 @@ export class WaferMap<
     public update(): void {
         this.validate();
         if (this.isExperimentalUpdate()) {
-            this.experimentalUpdate();
+            this.currentTask = this.experimentalUpdate().then(
+                () => {
+                    this.currentTask = undefined;
+                },
+                () => {}
+            );
```
rajsite, line 257: We should avoid using `.then` and prefer async / await

In this case can you just clear currentTask at the end of `experimentalUpdate`? make sure all the code paths clear it correctly

packages/nimble-components/src/wafer-map/modules/wafer-map-update-tracker.ts
```diff
@@ -93,16 +93,27 @@ export class WaferMapUpdateTracker extends UpdateTracker<typeof trackedItems> {
...
+            this.wafermap.currentTask.then(
+                () => {
+                    DOM.queueUpdate(() => {
+                        this.wafermap.update();
+                        this.untrackAll();
+                        this.updateQueued = false;
+                    });
+                },
+                () => {}
+            );
```
rajsite, line 116: Rewrite this to avoid `.then` and preefer async/await

something like
```ts
void async () => {
    await this.wafermap.currentTask;
    DOM.queueUpdate(() => {
        this.wafermap.update();
        this.untrackAll();
        this.updateQueued = false;
    });
}();
```

rajsite (commented):
Initial feedback, still need to review some bits

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -1,14 +1,64 @@
...
+import { createMatrixRenderer } from '../create-matrix-renderer';
+import type { MatrixRenderer } from '../../../../build/generate-workers/dist/esm/source/matrix-renderer';
 
 /**
  * Responsible for drawing the dies inside the wafer map, adding dieText and scaling the canvas
  */
 export class WorkerRenderer {
+    private worker!: Remote<MatrixRenderer>;
+
+    private isWorkerAlive = false;
```
rajsite, line 13: I don't think we need this isWorkerAlive stuff anymore since we are tracking the async lifecycle better and waiting for tasks to finish before queueing new tasks. My goal is to not need it anymore

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
```diff
@@ -33,6 +38,24 @@ export class ZoomHandler {
...
+        }
+    }
+
+    private observerWaferMap(): void {
+        if (this.wafermapNotifier) {
+            this.wafermapNotifier.unsubscribe(this, 'canvasWidth');
+            this.wafermapNotifier.unsubscribe(this, 'canvasHeight');
+        }
```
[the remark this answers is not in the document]
Razvan1928: Seems like it is true on every change on the wafer map data, ex: number of dies, orientation, colorScale etc, but zooming or resizing doesn't make the condition true
`this.wafermapNotifier` looks like this
image

packages/nimble-components/src/wafer-map/modules/wafer-map-update-tracker.ts
Razvan1928: sure! solved here https://github.com/ni/nimble/pull/1929/commits/2acda5dfbd4a393c9553f0e0a9f06d78d7ee0baf

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: sure! solved in this commit https://github.com/ni/nimble/pull/1929/commits/b1e08c9a676a5824cc848a4ad604cbf2659637c7

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
Razvan1928: indeed we no longer need that, I replaced the isWorkerAlive with a simple check if the worker is undefined, this is needed so we will not create new workers when we resize the wafer. https://github.com/ni/nimble/pull/1929/commits/8b6424051fe71d896b00acb65e9bc0021f637e22

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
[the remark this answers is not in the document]
rajsite: I don't think I understand what you're suggesting. Please look closely at the constructor logic.

It creates the readonly wafermapNotifier then immediately calls observerWaferMap. The constructor is the only call to observeWaferMap. It is called once by the constructor. Why does it need this conditional check? It is only ever called exactly once per instance of ZoomHAndler and we know the state of it.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -241,12 +249,13 @@ export class WaferMap<
     public update(): void {
         this.validate();
         if (this.isExperimentalUpdate()) {
-            this.experimentalUpdate();
+            // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
+            void (() => {
+                this.currentTask = this.experimentalUpdate();
+            })();
```
rajsite, line 255: Please make sure you understand the use-case of the void expression. It seems to be used for no reason here. The void expression is used for floating promises: https://typescript-eslint.io/rules/no-floating-promises/#ignorevoid

This use-case is not a floating promise. We are saving the result of the promise to be used, not completely ignoring the result.

```suggestion
            this.currentTask = this.experimentalUpdate();
```

packages/nimble-components/src/wafer-map/modules/wafer-map-update-tracker.ts
```diff
@@ -93,16 +93,26 @@ export class WaferMapUpdateTracker extends UpdateTracker<typeof trackedItems> {
...
+        this.updateQueued = true;
+        if (this.wafermap.currentTask === undefined) {
             DOM.queueUpdate(() => {
                 this.wafermap.update();
                 this.untrackAll();
                 this.updateQueued = false;
             });
+        } else {
+            void (async () => {
+                await this.wafermap.currentTask;
```
rajsite, line 108: The [timeline](https://github.com/ni/nimble/issues/2034) for `Render Dies` states "error handling of rejection" however there is no error handling of rejection in this PR (or tests for failure states).

i.e. the `await this.wafermap.currentTask` line can `throw` do to an Error being thrown in the worker, etc. But it's not being handled here.

@Razvan1928 @munteannatan The pattern we discussed is at least initially reporting the error to users via the validity flags. The options I can think of are:
- address it in this PR and then check the `Render Dies` task
- do not address it in this PR (and do not close the `Render Dies` task) but make a follow-up PR that does do the error handling to finish the `Render Dies` task.

packages/nimble-components/src/wafer-map/index.ts
Razvan1928: You are right, somehow I didn't notice I no longer needed the void expression, solved in this commit https://github.com/ni/nimble/pull/1929/commits/3c777a28161edf98ae30dc14454ba9c570c45906

packages/nimble-components/src/wafer-map/modules/zoom-handler.ts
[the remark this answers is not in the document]
Razvan1928: I think I understood now, being called when `ZoomHandler` the function call will:
- As you said, be called only once per instance
- The condition `if (this.wafermapNotifier)` will always be true (but only and only once)
- There is no point to unsubscribe as `canvasWidth` and `canvasHeight` weren't subscribed. I replaced the call of `observerWaferMap()` inside the constructor with only subscribing to `canvasWidth` and `canvasHeight` and moved the `observerWaferMap()` to be called in `handleChange()` https://github.com/ni/nimble/pull/1929/commits/a8abae25521e45cd28665f52a1e604460241f36e
please correct me if I'm wrong

packages/nimble-components/src/wafer-map/modules/wafer-map-update-tracker.ts
Razvan1928: As was discussed on teams we will do the 'error handling of rejection' in a different pr

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -1,14 +1,82 @@
...
+        await this.matrixRenderer.setScaling(scaleX, scaleY);
+        await this.matrixRenderer.setBases(
+            this.wafermap.experimentalDataManager.horizontalScale(0)!,
+            this.wafermap.experimentalDataManager.verticalScale(0)!
+        );
+        await this.matrixRenderer.setMargin(
+            this.wafermap.experimentalDataManager.margin
+        );
+
+        if (this.wafermap.diesTable !== undefined) {
```
rajsite, line 45: I think all these checks for diesTable are undefined are incorrect because they result in the worker not receiving any message.

But the user could intentionally be trying to clear the current wafer that is being drawn. Now if you set diesTable to undefined nothing happens right (the wafer canvas stays the same even though the data was cleared)?

Maybe we are being too clever with the diesTable property and renderer. We should consider adding a boolean property like `use-experimental-renderer` and just be explicit for the user to opt-in. 

I think there are two outcomes and they should not block this PR:
- setting and clearing the diesTable should behave well
- (up to y'all but I think it will help) we should make an explicit boolean configuration to switch rendering of the wafer map.

If that makes sense can you track it under `Render Dies` in the tracking issue? https://github.com/ni/nimble/issues/2034

packages/nimble-components/src/wafer-map/modules/prerendering.ts
rajsite: was that change made? I don't see it here

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
```diff
@@ -1,14 +1,82 @@
...
  * Responsible for drawing the dies inside the wafer map, adding dieText and scaling the canvas
  */
 export class WorkerRenderer {
+    private matrixRenderer!: Remote<MatrixRenderer>;
+
     public constructor(private readonly wafermap: WaferMap) {}
 
-    public drawWafer(): void {
-        // rendering will be implemented in a future PR
+    public async setupWafer(): Promise<void> {
```
rajsite, line 15: @Razvan1928 pushed a few changes to split setup and draw wafer into different responsibilities. If it looks good ready to go!

packages/nimble-components/src/wafer-map/tests/wafer-map-matrix.stories.ts
```diff
@@ -179,6 +180,12 @@ const componentWaferWithGridDimensions = ([
 >
 </${waferMapTag}>`;
 
+const componentWaferMapWorkerCanvas = (): ViewTemplate => html`<${waferMapTag}
+    :diesTable="${() => wafermapDiesTableSets[0]}"
+    :apiVersion="${() => 'experimental'}"
```
rajsite, line 185: The wafer-map class does not have a property named apiVersion. I don't think this does anything and this line can be removed

rajsite (approved):

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
Razvan1928: So we have a flag on the wafer-map called isExperimentalRenderer() but it is checked before calling setupWafer(), inside update() from index.ts there is a 
`
        if (this.isExperimentalUpdate()) {
            this.currentTask = this.experimentalUpdate();
            return;
        }
`
and the call stack is update() -> this.experimentalUpdate() -> this.workerRenderer.setupWafer();

so we can assume every time this.workerRenderer.setupWafer it is an experimental wafer

Regarding the undefined checking and the worker not receiveing messages, I replaced that check with 
image
now the wafer is cleared when diesTable is undefined

packages/nimble-components/src/wafer-map/tests/wafer-map-matrix.stories.ts
Razvan1928: You are right! I deleted it

packages/nimble-components/src/wafer-map/modules/prerendering.ts
Razvan1928: It was made but at that time I removed it because it was causing some errors in the storybook, I don't remember in detail. I will solve this conversation, the explanation is given in another conversation.

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
Razvan1928: Took a look on the commits, everything looks great!

packages/nimble-components/src/wafer-map/modules/experimental/worker-renderer.ts
rajsite: Reopening this question. @Razvan1928 

- isExperimentalUpdate just checks that diesTable is undefined, are you sure that is sufficient? Did you do any manual verification?
- are there tests to verify the behavior change? Ie the wafer if drawn, then the data is changed to undefined, and the wafer is cleared?

If there are not tests there needs to be a follow-up PR

PR #1931, opened 2024-03-13
Fix missing element names in some Storybook docs

m-akinc:
# Pull Request

## 🤨 Rationale

Some tag names in a few component docs were blank, because the wrong attribute was being set.

## 👩‍💻 Implementation

Should be `<Tag name=... >` rather than `<Tag of=... >`.

## 🧪 Testing

Storybook

## ✅ Checklist


jattasNI (approved):
No need to wait for other reviewers, this is a trivial change. @rajsite 

I'm not sure what's up with that `npm ci` build failure but it's clearly not related to your change. I don't have any suggestions other than requeuing.

rajsite (approved):

Issue #1906, opened 2024-03-06
Accent appearance variant for nimble-menu-button


codrutaserban:
# 🙋 Feature Request

## 😯 Problem to Solve

In sl-lib-angular a split-button is implemented with a nimble-button and a nimble-menu-button. This split button can be used as a primary button too, and for this, each nimble component needs appearance-variant to be set to accent. 

## 💁 Proposed Solution

Add appearance-variant attribute to nimble-menu-button, as it's available for nimble-button.

<img width="242" alt="Screenshot 2024-03-07 at 10 00 12 AM" src="https://github.com/ni/nimble/assets/1458528/6499148e-45bd-4dbb-be97-d6bf3b11a86b">

Design is available in [Figma](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-77205&mode=design)

TrevorKarjanis:
@m-akinc The variant was not added to the [NimbleMenuDirective](https://github.com/ni/nimble/blob/main/angular-workspace/projects/ni/nimble-angular/src/directives/menu-button/nimble-menu-button.directive.ts).

m-akinc:
> @m-akinc The variant was not added to the [NimbleMenuButtonDirective](https://github.com/ni/nimble/blob/main/angular-workspace/projects/ni/nimble-angular/src/directives/menu-button/nimble-menu-button.directive.ts).

Thanks, the issue shouldn’t have been closed yet, as the new attribute needs both Angular and Blazor support. I’ll have a PR up for that later today.

PR #1934, opened 2024-03-14
Support primary and accent variants for toggle and menu buttons

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes: #1906 

## 👩‍💻 Implementation

- The ask was for the accent appearance variant on the menu button, but because the menu button uses the toggle button internally, I added the `appearance-variant` attribute to both toggle and menu button. 
- I added support for both primary and accent appearance variants.
- When adding the new attribute states to the toggle and menu button matrix stories, I moved some common button state arrays to a new file in `patterns/buttons/tests`.
- While examining visual designs/colors, I found and fixed several issues:
    - The `DigitalGreenDark` token value was incorrect, based on the Figma document. I updated this value, which affected block-accent button color and the rich-text mention font color.
    - Anchors in active, prominent, and prominent+active states were using the wrong color tokens.
    - The icon color for outline-accent buttons should match the text color, but didn't.
    - After consulting with Brandon, he directed me to change the active/pressed colors for the block-primary and block-accent button appearances so they would be consistent with other button active/pressed colors. Previously, they were slightly darker versions of the primary and accent colors (see below). This allowed me to remove the two, now-unused tokens: `buttonFillActivePrimaryColor` and `buttonFillAccentActiveColor`.
image

To avoid increasing the size of this PR, Angular/Blazor support will be added in a follow-up PR.

## 🧪 Testing

Lots of manual state styling testing in Storybook. Also Chromatic visual tests.

## ✅ Checklist


packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -139,39 +133,42 @@ export const styles = css`
...
     }
 
     @layer disabled {
         :host([disabled]) {
-            color: ${buttonLabelDisabledFontColor};
             cursor: default;
         }
 
         :host([disabled]) .control {
+            color: ${buttonLabelDisabledFontColor};
```
m-akinc, line 161: `buttonLabelDisabledFontColor` is the same as `rgba(${actionRgbParitalColor}, 0.3)`, so I replaced it.

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -1,20 +1,34 @@
...
-import { styles as buttonStyles } from '../patterns/button/styles';
+import {
+    buttonAppearanceVariantStyles,
+    styles as buttonStyles
+} from '../patterns/button/styles';
+import { appearanceBehavior } from '../utilities/style/appearance';
+import { ButtonAppearance } from './types';
 
 export const styles = css`
+    @layer base, base-selected, hover, focusVisible, active, disabled, top;
```
m-akinc, line 18: By inserting a new `base-selected` layer, I can define the default styling for a toggled-on button, but still let standard hover, focus, active, and disabled styling take precedence.

packages/nimble-components/src/toggle-button/styles.ts
```diff
@@ -76,15 +45,37 @@ export const styles = css`
...
-            border-color: rgba(${fillSelectedRgbPartialColor}, 0.3);
-            box-shadow: none;
-        }
     }
-`;
+`.withBehaviors(
+    appearanceBehavior(
+        ButtonAppearance.outline,
+        css`
+            @layer base-selected {
```
m-akinc, line 53: Need to override the `base-selected` styling for the outline-accent appearance, so that it keeps the same accent color for the label and icons, even when toggled-on.

m-akinc:
@msmithNI do you mind buddying this?

change/@ni-nimble-components-083cf44c-68b4-43f6-b689-efd81c8228b4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
msmithNI, line 2: Since `buttonFillActivePrimaryColor` and `buttonFillAccentActiveColor` were previously exported design tokens, it seems like that makes this a `major` change (and we should note removing those tokens in the change notes).

msmithNI (approved):

change/@ni-nimble-components-083cf44c-68b4-43f6-b689-efd81c8228b4.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Haven't reviewed the rest yet but in regards to the following PR comment:

> buttonFillActivePrimaryColor and buttonFillAccentActiveColor (good riddance to inconsistent token names).

Those are inline with :token naming guidance:
https://github.com/ni/nimble/blob/8bb2c33f05e8e9a3db8a86804eabadbd7bc5d033/packages/nimble-components/CONTRIBUTING.md#L489-L498

element: button
part: fill
state: active, primary vs accent, active (the states are [sorted alphabetically](https://github.com/ni/nimble/blob/8bb2c33f05e8e9a3db8a86804eabadbd7bc5d033/packages/nimble-components/CONTRIBUTING.md#L497))
token_type: color

m-akinc: It's good that it wasn't accidental, but it still looks and feels inconsistent. If we had used "AppearanceVariantAccent" and "AppearanceVariantPrimary" instead of just "Accent" and "Primary", that would have solved the inconsistency at the cost of having reasonable name lengths. A better solution might be to give "Active", "Hover", "Selected", "Disabled" their own category separate from all other state descriptors. The other states category could maybe be renamed to **variant**. Now that I've removed those two tokens, it looks like the rest are consistent with the modified scheme I described, so we could change our guidance if we want.

The only tokens I'm not quite clear on are those like `titlePlus1DisabledFontColor`. Apparently "Plus1" is not a state, but rather part of "titlePlus1" (which is the **part**). If so, then it's still consistent with my suggested change. If not, then it was already nonconforming.

rajsite: title-plus1 is the element / part correct.

I don't think I follow your proposal for the naming scheme. May help to give a concrete proposal. Feel free to make a separate PR if you feel passionate about it.

m-akinc: I think it's a worthwhile change to make, so I created #1947

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -5,16 +5,25 @@ import {
     createUserSelectedThemeStory,
```
jattasNI, line 5: The button mdx has a big section about how to use the appearance modes and appearance variants but we don't include that section for the menu button, toggle button, or anchor button even though they now (or in some cases already did) support the same configuration.

Do you know if the guidance is the same for all types of buttons or if there are some states that are not recommended for some button types? You may need to consult with Brandon about this. If the guidance is different then we should document it in each component page. If it's the same then we should add something to each page. I'd be fine with it being a link to the button docs. The other alternative would be to find a way to share it but swap out the components used, which may not be worth the effort.

jattasNI (approved):

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
m-akinc: For now, I've gone ahead and added that styling section to each of the other button docs. @NIbokeefe is it reasonable to show the same guidance about appearances and appearance variants for anchor, menu, and toggle buttons?

I made some minor edits that came up while doing this. E.g. removed unhelpful `Styling / Theme` section from button docs, made a couple minor wording edits to the button appearance section, and improved wording of spinner's `Styling / Theme` section.

NIbokeefe: @m-akinc 
CC: @jattasNI 

Yes, we can use this solution consistently for all buttons, Action, Toggle, Anchor and Menu.  I've updated the buttons on my end in Figma to reflect those updates.

packages/nimble-components/src/patterns/button/tests/states.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const appearanceStates: [string, string | undefined][] = Object.entries(
+    ButtonAppearance
+).map(([key, value]) => [pascalCase(key), value]);
+export type AppearanceState = (typeof appearanceStates)[number];
+
+export const appearanceVariantStates: [string, string | undefined][] = Object.entries(ButtonAppearanceVariant).map(([key, value]) => [
+    pascalCase(key),
+    value
+]);
```
rajsite, line 22: You lose `as const`-ness typing of the enum states this way. Also it's not easier to read and not really changing often so don't think being explicit is a maintenance burden. Think it's better to have the stronger type

image

Compared to

image

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -139,39 +133,42 @@ export const styles = css`
     @layer active {
         .control:active {
             box-shadow: none;
+            color: ${buttonLabelFontColor};
```
rajsite, line 136: I'll point it out in one spot and let you update all the styles.ts files touched and recheck. Can you make sure these are in [DOM order based on the selector target](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order)? 

image


```
.control: active {}
.control:active::before {}
... [part start] {}
... [part end] {}
```

packages/nimble-components/src/toggle-button/styles.ts
rajsite: Do you need to actually add a new named layer? Can't we just use `base` again? it should make any selector in here win.

packages/nimble-components/src/toggle-button/styles.ts
rajsite: if we do need the new layer "selected" to me was ambiguous with "active" / "focus" which are the current "user selected" thing to interact with.

This comment makes me think more "pressed" going with the aria naming. And I'd leave off the 'base' prefix.

rajsite, line 18: This kind of override is really hard for me to follow mentally. II don't actually know how redefining `@layer` works. Sucks that it's also hard to inspect in dev tools. If I manually serialize the adoptedStylesheet in storybook with:

```js
console.log(Array.from($0.shadowRoot?.adoptedStyleSheets ?? []).flatMap(sheet => Array.from(sheet.rules).flatMap(rules => rules.cssText.split(';').join(';\n'))).join('\n\n') || 'none');
```

I get:

```
@layer base, base-selected, hover, focusVisible, active, disabled, top;


@layer base, hover, focusVisible, active, disabled, top;


@layer base {
  :host([hidden]) { display: none;
 }
  :host { display: inline-flex;
 }
  :host { background-color: transparent;
 height: var(--ni-nimble-control-height);
 color: var(--ni-nimble-button-label-font-color);
 font: var(--ni-nimble-button-label-font);
 cursor: pointer;
 outline: none;
 border: none;
 box-sizing: border-box;
 vertical-align: middle;
 }
  .control { background-color: transparent;
 height: 100%;
 width: 100%;
 border: var(--ni-nimble-border-width) solid transparent;
 box-sizing: border-box;
 color: inherit;
 border-radius: inherit;
 fill: inherit;
 display: inline-flex;
 align-items: center;
 justify-content: center;
 gap: 4px;
 cursor: inherit;
 font: inherit;
 outline: none;
 margin: 0px;
 padding: 0 var(--ni-nimble-standard-padding);
 position: relative;
 transition: box-shadow var(--ni-nimble-small-delay) ease-in-out,
                border-color var(--ni-nimble-small-delay) ease-in-out,
                background-size var(--ni-nimble-small-delay) ease-in-out;
 background-size: 100% 100%;
 background-repeat: no-repeat;
 background-position: center center;
 }
  :host([content-hidden]) .control { aspect-ratio: 1 / 1;
 padding: 0px;
 }
  .control::before { content: "";
 position: absolute;
 width: 100%;
 height: 100%;
 pointer-events: none;
 box-sizing: border-box;
 outline: transparent solid 0px;
 color: transparent;
 background-clip: content-box;
 transition: outline var(--ni-nimble-small-delay) ease-in-out;
 }
  .content { display: contents;
 }
  :host([content-hidden]) .content { display: inline-block;
 height: 1px;
 width: 1px;
 position: absolute;
 margin: -1px;
 clip: rect(1px, 1px, 1px, 1px);
 clip-path: inset(50%);
 overflow: hidden;
 padding: 0px;
 }
  [part="start"], [part="end"] { display: contents;
 --ni-nimble-icon-color: var(--ni-nimble-button-label-font-color);
 }
}

@layer hover {
  .control:hover { border-color: var(--ni-nimble-border-hover-color);
 box-shadow: 0px 0px 0px var(--ni-nimble-border-width) var(--ni-nimble-border-hover-color) inset;
 }
}

@layer focusVisible {
  .control:focus-visible { border-color: var(--ni-nimble-border-hover-color);
 box-shadow: 0px 0px 0px var(--ni-nimble-border-width) var(--ni-nimble-border-hover-color) inset;
 }
  .control:focus-visible::before { outline: var(--ni-nimble-border-width) solid var(--ni-nimble-border-hover-color);
 outline-offset: -3px;
 }
}

@layer active {
  .control:active { box-shadow: none;
 color: var(--ni-nimble-button-label-font-color);
 border-color: var(--ni-nimble-border-hover-color);
 background-image: linear-gradient(
                var(--ni-nimble-fill-selected-color),
                var(--ni-nimble-fill-selected-color)
            );
 background-size: calc(100% - 2px) calc(100% - 2px);
 }
  .control:active [part="start"], .control:active [part="end"] { --ni-nimble-icon-color: var(--ni-nimble-button-label-font-color);
 }
  .control:active::before { outline: none;
 }
}

@layer disabled {
  :host([disabled]) { cursor: default;
 }
  :host([disabled]) .control { color: var(--ni-nimble-button-label-disabled-font-color);
 box-shadow: none;
 background-image: none;
 background-size: 100% 100%;
 }
  :host([disabled]) .control::before { box-shadow: none;
 }
  :host([disabled]) slot[name="start"]::slotted(*), :host([disabled]) slot[name="end"]::slotted(*) { opacity: 0.3;
 --ni-nimble-icon-color: var(--ni-nimble-button-label-font-color);
 }
}

@layer top {
  @media (prefers-reduced-motion) {
  .control { transition-duration: 0s;
 }
}
  :host([content-hidden]) [part="end"] { display: none;
 }
}

@layer base-selected {
  .control[aria-pressed="true"] [part="start"], .control[aria-pressed="true"] [part="end"] { --ni-nimble-icon-color: var(--ni-nimble-button-label-font-color);
 }
  .control[aria-pressed="true"] { background-color: transparent;
 color: var(--ni-nimble-button-label-font-color);
 background-image: linear-gradient(
                var(--ni-nimble-fill-selected-color),
                var(--ni-nimble-fill-selected-color)
            );
 border-color: rgba(var(--ni-nimble-fill-selected-rgb-partial-color), 0.3);
 }
}

@layer disabled {
  :host([disabled]) .control[aria-pressed="true"] { background-image: linear-gradient(
                var(--ni-nimble-fill-selected-color),
                var(--ni-nimble-fill-selected-color)
            );
 border-color: rgba(var(--ni-nimble-fill-selected-rgb-partial-color), 0.3);
 }
}

@layer base {
  .control { border-color: rgba(var(--ni-nimble-action-rgb-partial-color), 0.3);
 }
}

@layer disabled {
  :host([disabled]) .control { border-color: rgba(var(--ni-nimble-border-rgb-partial-color), 0.3);
 }
}

@layer base {
  :host([appearance-variant="primary"]) .control { border-color: var(--ni-nimble-action-rgb-partial-color);
 }
  :host([appearance-variant="accent"]) [part="start"], :host([appearance-variant="accent"]) [part="end"] { --ni-nimble-icon-color: var(--ni-nimble-button-accent-outline-font-color);
 }
  :host([appearance-variant="accent"]) .control { border-color: var(--ni-nimble-button-border-accent-outline-color);
 color: var(--ni-nimble-button-accent-outline-font-color);
 }
}

@layer active {
  :host([appearance-variant="accent"]) .control:active [part="start"], :host([appearance-variant="accent"]) .control:active [part="end"] { --ni-nimble-icon-color: var(--ni-nimble-button-accent-outline-font-color);
 }
  :host([appearance-variant="accent"]) .control:active { color: var(--ni-nimble-button-accent-outline-font-color);
 }
}

@layer base-selected {
  :host([appearance-variant="accent"]) .control[aria-pressed="true"] [part="start"], :host([appearance-variant="accent"]) .control[aria-pressed="true"] [part="end"] { --ni-nimble-icon-color: var(--ni-nimble-button-accent-outline-font-color);
 }
  :host([appearance-variant="accent"]) .control[aria-pressed="true"] { color: var(--ni-nimble-button-accent-outline-font-color);
 }
}
```

So the source code order matches the dom order well which has been a goal. But the order of the `@layer` definitions was surprising to me. And that the toggle buttons while first were not overridden? Or maybe the first one always wins? I don't know how it works.

Instead of overriding the `@layer` definition I'd rather have one clearly defined in the scope. I'm okay either:
- Merging all the layers into the top one in the button so button has some unused layers
- Making separate `css` tags for each layer config (button vs button+toggle) and updating each user of the button styles to explicitly pull iin only one `@layer` config.

packages/nimble-components/src/toggle-button/styles.ts
m-akinc: Yes, if we just use the `base` layer, then the default "accent" styling takes precedence over the "pressed" styling because of specificity:
image
We want the "pressed" styling to always override the default styling regardless of specificity, which is what CSS layers are designed to do.

packages/nimble-components/src/patterns/button/tests/states.ts
m-akinc: I had just copied these as they were. I'll change them.

packages/nimble-components/src/toggle-button/styles.ts
m-akinc: It's actually pretty straightforward. As the second "NOTE" in [this section of the spec](https://drafts.csswg.org/css-cascade-5/#layer-empty) states, layer ordering is defined by first occurrence of the layer name. So we can always declare the layer order before any other imported content without worrying about the imported content affecting our specified order. The only caveat is that if the imported content declares any layers that we didn't, those will be ordered after (take precedence over) the layers we declared. So, supposing we introduce a new layer in the shared button CSS, and we didn't update the toggle-button `styles.ts`, that layer would be applied last, which is probably wrong. For this reason, your recommendation seems like a good idea. I'll just add the new layer to the shared button styles.

m-akinc: Unfortunately, I think "pressed" is ambiguous with "active", since a button is "active" while it is pressed. I suggest that we instead call it "checked", since that term matches the classes and attribute applied in that state. Plus, if we later introduce similar layers for other controls with binary state, that name will make sense.

packages/nimble-components/src/patterns/button/styles.ts
```diff
@@ -106,11 +104,7 @@ export const styles = css`
             ${accessiblyHidden}
         }
 
-        [part='start'] {
-            display: contents;
-            ${iconColor.cssCustomProperty}: ${buttonLabelFontColor};
-        }
-
+        [part='start'],
```
rajsite, line 107:
```
.control
.control::before
start
content
end
```

It does mean the selectors for start and end should be split. That's fine.

rajsite (approved):

fredvisser (approved):

TrevorKarjanis (approved):
I'm approving just because I can.

PR #1946, opened 2024-03-19
Created Computations and Data Manager Experimental

munteannatan:
# Pull Request

## 🤨 Rationale

Created new computations experimental component with changed scales and diesTable iteration for finding the bounding box until arquero fixes.

The new scales will be able to be inverted without the need to create separate scales for the hover handler.

## 👩‍💻 Implementation

changed the scales and iteration over the diesTable
integrated the changes and the strategy swap with the wafermap and the experimental hover

## 🧪 Testing

copied spec classes for the new experimental modules.
updated hover handler spec.
adapted the data manager to the previous renderer for visual tests in storybook.

## ✅ Checklist


packages/nimble-components/src/wafer-map/modules/experimental/computations.ts
```diff
@@ -0,0 +1,231 @@
...
+            .getChild('colIndex')!
+            .toArray() as Int32Array;
+        const rowIndex = this.wafermap.diesTable
+            .getChild('rowIndex')!
+            .toArray() as Int32Array;
+
+        const minPoint = { x: colIndex[0]!, y: rowIndex[0]! };
+        const maxPoint = { x: colIndex[0]!, y: rowIndex[0]! };
+
+        // will replace iterating with arquero derive after fixing errors
```
atmgrifter00, line 147:
```suggestion
        // will replace iterating with arquero after fixing errors
```
Also, I think it's worth linking to the PR in arquero here.

packages/nimble-components/src/wafer-map/tests/experimental-computations.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+            expect(computationsModule.margin).toEqual({
+                top: 4,
+                right: 54,
+                bottom: 4,
+                left: 54
+            });
+        });
+    });
+
+    describe('with top left originLocation', () => {
```
atmgrifter00, line 100: It feels like groups of these tests could be handled as a `parameterizeSpec`. For this `describe` and the following three, the only differences are the provided orientation, and the provided expected results for the horizontal and vertical ranges. These should all be able to be captured as `parameterizeSpec` parameters.

Consider other `parameterizeSpec` cases as well for providing a variety of input parameters and expected results.

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/tests/utilities.ts
```diff
@@ -201,6 +215,27 @@ export function getWaferMapMockComputations(
...
+export function getWaferMapMockComputationsExperimental(
+    diesTable: Table = getWaferMapDiesTable(),
+    originLocation: WaferMapOriginLocation,
+    canvasWidth: number,
+    canvasHeight: number,
+    validity: WaferMapValidity = {
+        invalidGridDimensions: false,
+        invalidDiesTableSchema: false
+    }
+): Pick<
```
rajsite, line 227: Use the same patterns as pushed to thee other PPR to reduce `as` casts, i.e. return WaferMap instead of Pick<WaferMapp> see: https://github.com/ni/nimble/pull/1916/commits/15ad11c33712a29e11d60ba44611cac969f0d637#diff-f8244bcd1e327bb58976a036662c316f8d3bf6b6e2f3428ebc374ca882985d10R163

And discussion here:
https://github.com/ni/nimble/pull/1916#discussion_r1537864808

rajsite (approved):

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -51,7 +51,29 @@ export class Prerendering {
...
+        )
+            .map(row => {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                return {
+                    x: row.colIndex,
+                    y: row.rowIndex,
+                    value: row.value,
+                    tags: []
+                };
+            })
```
rajsite, line 76: this all looks super sketchy. approving assuming it will be heavily cleaned up and will not have unsafe assignments and `as` casting is the following PR. If that is not the case should look more carefully and try and clean-up

packages/nimble-components/src/wafer-map/tests/experimental-computations.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+            left: 4
+        };
+        beforeEach(() => {
+            const waferMock = getWaferMapMockComputationsExperimental(
+                getWaferMapDiesTable(),
+                WaferMapOriginLocation.topLeft,
+                100,
+                100
+            );
+            computationsModule = new Computations(waferMock as WaferMap);
```
rajsite, line 27: `as` cast not needed after cleanup

packages/nimble-components/src/wafer-map/tests/experimental-computations.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    describe('with rectangular canvas', () => {
+        beforeEach(() => {
+            const waferMock = getWaferMapMockComputationsExperimental(
+                getWaferMapDiesTable(),
+                WaferMapOriginLocation.topLeft,
+                200,
+                100
+            );
+            computationsModule = new Computations(waferMock as WaferMap);
```
rajsite, line 69: `as` cast not needed after cleanup

packages/nimble-components/src/wafer-map/tests/experimental-computations.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    parameterizeSpec(testCases, (spec, name, value) => {
+        spec(
+            `with ${name} originLocation should have expected horizontal range and vertical range`,
+            () => {
+                const waferMock = getWaferMapMockComputationsExperimental(
+                    getWaferMapDiesTable(),
+                    value.name,
+                    100,
+                    100
+                ) as WaferMap;
```
rajsite, line 133: `as` cast not needed after cleanup

packages/nimble-components/src/wafer-map/tests/experimental-data-manager.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+        );
+        const diesWithPartialOpacity = dataManagerModule.diesRenderInfo.filter(
+            x => !x.fillStyle.endsWith(',1)')
+        );
+        expect(dies.length).toEqual(diesWithPartialOpacity.length);
+    });
+
+    // skipped until prerendering is refactored
+    xit('should have all dies inside the canvas with margins', () => {
+        for (const dieRenderInfo of dataManagerModule.diesRenderInfo) {
```
rajsite, line 147: This test should [not have conditional asserts](https://github.com/ni/javascript-styleguide/issues/125#issuecomment-1631721812) from looping over a structure not defined in the test. Instead precalculate the expected array and assert the array at once. If not cleaning up now include in the "skipped" comment that it needs to be cleaned up.

packages/nimble-components/src/wafer-map/tests/experimental-data-manager.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+    it('should have as many dies as provided', () => {
+        expect(dataManagerModule.diesRenderInfo.length).toEqual(
+            getWaferMapDies().length
+        );
+    });
+
+    it('should have label with suffix for each die', () => {
+        for (const dieInfo of dataManagerModule.diesRenderInfo) {
+            expect(dieInfo.text).toContain(dieLabelsSuffix);
+        }
```
rajsite, line 118: [Don't loop over asserts](https://github.com/ni/javascript-styleguide/issues/125#issuecomment-1631721812). Make sure to precalculate the array and assert its length so we know exactly how many asserts to expect.

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -152,7 +153,8 @@ describe('WaferMap', () => {
             expect(element.renderer instanceof RenderingModule).toBeTrue();
         });
 
-        it('will use WorkerRenderer after supported diesTable change', () => {
+        // skipped until prerendering is refactored
+        xit('will use WorkerRenderer after supported diesTable change', () => {
```
jattasNI, line 157: By disabling existing tests, it looks like test coverage of the stable rendering path is decreasing as a result of this change. Assuming that's accurate, I'm surprised by that: I thought that the existing non-experimental functionality would remain the same and it's only the experimental functionality that has behavior gaps.

Is there a way to keep these cases covered while we refactor the experimental code path? Or am I missing something?

jattasNI (approved):

packages/nimble-components/src/wafer-map/modules/prerendering.ts
munteannatan: This is just an adaptor to see that both strategies produce the same results with the previous render. It's not required, I can remove it

packages/nimble-components/src/wafer-map/modules/experimental/hover-handler.ts
```diff
@@ -72,27 +75,30 @@ export class HoverHandler {
...
-            )
-        );
-        const y = yRoundFunction(
-            this.wafermap.dataManager.invertedVerticalScale(
-                mousePosition.y - this.wafermap.dataManager.margin.top
-            )
-        );
-        return { x, y };
+    ): PointCoordinates | undefined {
+        if (this.wafermap.dataManager instanceof DataManager) {
```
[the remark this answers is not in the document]
munteannatan: there is still a type check for typescript, but I made both checks

packages/nimble-components/src/wafer-map/modules/hover-handler.ts
```diff
@@ -56,28 +61,31 @@ export class HoverHandler {
...
-            )
-        );
-        const y = yRoundFunction(
-            this.wafermap.dataManager.invertedVerticalScale(
-                mousePosition.y - this.wafermap.dataManager.margin.top
-            )
-        );
-        return { x, y };
+    ): PointCoordinates | undefined {
+        if (this.wafermap.dataManager instanceof DataManager) {
```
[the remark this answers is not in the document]
munteannatan: there is still a type check for typescript, but I made both checks

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
munteannatan: I removed the prerendering part which was tripping these tests

PR #1947, opened 2024-03-19
Small change to design token naming scheme

m-akinc:
# Pull Request

## 🤨 Rationale

I encountered a couple design tokens with names that felt inconsistent:
`buttonFillActivePrimaryColor`
`buttonFillAccentActiveColor`

It seemed like the appearance variant ("accent" or "primary") should be in the same relative position within the name, but our naming convention dictated these names instead.

## 👩‍💻 Implementation

Treat common state descriptors ("active", "disabled", "hover", and "selected") as a separate category of name segment than other distinguishing descriptors (like "accent" and "primary"). So instead of our old structure:

`[element]-[part]-[state]-[token_type]`

we now have:

`[element]-[part]-[state]-[variant]-[token_type]`

Luckily, the only existing token that violates this updated naming scheme is `buttonFillAccentActiveColor` (which should instead be `buttonFillActiveAccentColor`) and that token is being removed in [another PR](https://github.com/ni/nimble/pull/1934).

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/CONTRIBUTING.md
+++ b/packages/nimble-components/CONTRIBUTING.md
@@ -490,12 +490,13 @@ To modify the generated tokens, complete these steps:
 
 Public names for theme-aware tokens are specified in `src/theme-provider/design-token-names.ts`. Use the following structure when creating new tokens.
 
-`[element]-[part]-[state]-[token_type]`
+`[element]-[part]-[state]-[variant]-[token_type]`
 
 1. Where **element** is the type to which the token applies (e.g. 'application', 'body', or 'title-plus-1').
 2. Where **part** is the specific part of the element to which the token applies (e.g. 'border', 'background', or shadow).
-3. Where **state** is the more specific state descriptor (e.g. 'selected' or 'disabled'). Multiple states should be sorted alphabetically.
-4. Where **token_type** is the token category (e.g. 'color', 'font', 'font-color', 'height', 'width', or 'size').
+3. Where **state** is one or more of the following interaction states: 'active', 'disabled', 'hover', or 'selected'. Multiple states should be sorted alphabetically.
+4. Where **variant** is any other distinguishing descriptor (e.g. 'accent', 'primary, or 'large'). Multiple variants should be sorted alphabetically.
+5. Where **token_type** is the token category (e.g. 'color', 'font', 'font-color', 'height', 'width', or 'size').
 
 ### Size ramp
 
--- a/packages/nimble-components/CONTRIBUTING.md
+++ b/packages/nimble-components/CONTRIBUTING.md
@@ -490,12 +490,12 @@ To modify the generated tokens, complete these steps:
 
 Public names for theme-aware tokens are specified in `src/theme-provider/design-token-names.ts`. Use the following structure when creating new tokens.
 
-`[element]-[part]-[state]-[variant]-[token_type]`
+`[element]-[part]-[interaction_state]-[other_state]-[token_type]`
 
 1. Where **element** is the type to which the token applies (e.g. 'application', 'body', or 'title-plus-1').
 2. Where **part** is the specific part of the element to which the token applies (e.g. 'border', 'background', or shadow).
-3. Where **state** is one or more of the following interaction states: 'active', 'disabled', 'hover', or 'selected'. Multiple states should be sorted alphabetically.
-4. Where **variant** is any other distinguishing descriptor (e.g. 'accent', 'primary, or 'large'). Multiple variants should be sorted alphabetically.
+3. Where **interaction_state** is one or more interaction states (e.g. 'active', 'disabled', 'hover', or 'selected'). Multiple values should be sorted alphabetically.
+4. Where **other_state** is any other distinguishing state (e.g. 'accent', 'primary, or 'large'). Multiple values should be sorted alphabetically.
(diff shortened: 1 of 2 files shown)


packages/nimble-components/CONTRIBUTING.md
```diff
@@ -490,12 +490,13 @@ To modify the generated tokens, complete these steps:
...
 Public names for theme-aware tokens are specified in `src/theme-provider/design-token-names.ts`. Use the following structure when creating new tokens.
 
-`[element]-[part]-[state]-[token_type]`
+`[element]-[part]-[state]-[variant]-[token_type]`
 
 1. Where **element** is the type to which the token applies (e.g. 'application', 'body', or 'title-plus-1').
 2. Where **part** is the specific part of the element to which the token applies (e.g. 'border', 'background', or shadow).
-3. Where **state** is the more specific state descriptor (e.g. 'selected' or 'disabled'). Multiple states should be sorted alphabetically.
-4. Where **token_type** is the token category (e.g. 'color', 'font', 'font-color', 'height', 'width', or 'size').
+3. Where **state** is one or more of the following interaction states: 'active', 'disabled', 'hover', or 'selected'. Multiple states should be sorted alphabetically.
```
jattasNI, line 497: nit: I prefer the "(e.g. ...)" format we had before since this may not be an exhaustive list. "Focused" and its variants are the only ones I can think of that might be missing, but I prefer to keep some wiggle room if a new state arises or we're forgetting one.

jattasNI (approved):
I love a good naming convention!

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -490,12 +490,13 @@ To modify the generated tokens, complete these steps:
...
+4. Where **variant** is any other distinguishing descriptor (e.g. 'accent', 'primary, or 'large'). Multiple variants should be sorted alphabetically.
```
rajsite, line 498: "state" vs "variant" / "distinguishing descriptor" does not have a clear concept to me. i.e. it's not clear what it is a variant of. The examples are more like variants of the element/part not variants of the interaction state. If it's modelled after the appearance variant concept it's clearer that in that context it's a variant of appearance.

I'd propose `interaction state` vs `appearance state`. I think the examples show how it's not over constrained to the appearance attribute concepts (i.e. by including 'large')

As an aside this kind of also parallels the [cascade guidance](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes) where the interaction states for the most part are before other states.

m-akinc: My intent was for "variant" to be a catch-all for however else we're distingushing this token from another (with the same element, part, and/or interaction state). It may not be an appearance-based distinction. As a contrived example, suppose our token is a distinct width or font property that we want to use when the locale is German (because German words tend to be longer and take up more space). That doesn't seem like an "appearance state" to me.

As a counter proposal, I'm changing these to `interaction_state` and `other_state`, but I'm happy to use something else if you don't like `other_state`.

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -490,12 +490,13 @@ To modify the generated tokens, complete these steps:
...
 
-`[element]-[part]-[state]-[token_type]`
+`[element]-[part]-[interaction_state]-[other_state]-[token_type]`
 
 1. Where **element** is the type to which the token applies (e.g. 'application', 'body', or 'title-plus-1').
 2. Where **part** is the specific part of the element to which the token applies (e.g. 'border', 'background', or shadow).
-3. Where **state** is the more specific state descriptor (e.g. 'selected' or 'disabled'). Multiple states should be sorted alphabetically.
-4. Where **token_type** is the token category (e.g. 'color', 'font', 'font-color', 'height', 'width', or 'size').
+3. Where **interaction_state** is one or more interaction states (e.g. 'active', 'disabled', 'hover', or 'selected'). Multiple values should be sorted alphabetically.
+4. Where **other_state** is any other distinguishing state (e.g. 'accent', 'primary, or 'large'). Multiple values should be sorted alphabetically.
```
rajsite, line 498:
```suggestion
3. Where **interaction_states** is one or more interaction states (e.g. 'active', 'disabled', 'hover', or 'selected'). Multiple values should be sorted alphabetically.
4. Where **remaining_states** the remaining, non-interaction states (e.g. 'accent', 'primary, or 'large'). Multiple values should be sorted alphabetically.
```

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -490,12 +490,13 @@ To modify the generated tokens, complete these steps:
 
 Public names for theme-aware tokens are specified in `src/theme-provider/design-token-names.ts`. Use the following structure when creating new tokens.
 
-`[element]-[part]-[state]-[token_type]`
+`[element]-[part]-[interaction_state]-[other_state]-[token_type]`
```
rajsite, line 493:
```suggestion
`[element]-[part]-[interaction_states]-[remaining_states]-[token_type]`
```

rajsite (approved):

packages/nimble-components/CONTRIBUTING.md
rajsite: gave an alternative and approved assuming we are aligned, if not can give a counter

PR #1950, opened 2024-03-20
Add cell placeholders

mollykreis:
# Pull Request

## 🤨 Rationale

This PR adds a `placeholder` configuration to the following table columns as described in #1869: 
- anchor
- date-text
- duration-text
- number-text
- text

This is part of #1538. The remaining work is to add the `placeholder` attributes to the Angular and Blazor wrappers for each column.

As part of this PR, I also included best practices established for each column in the storybook docs. These are adapted from the placeholder HLD.

## 👩‍💻 Implementation

I implemented this change in a similar way as #1914. Specifically, I updated `TableColumnTextCellViewBase` to have its own implementation of `columnConfigChanged` and `cellRecordChanged` that calls a new `applyPlaceholderTextIfNeeded` function. This required me to create a `TableColumnTextBaseCellRecord` interface to enforce that the cell records used by `TableColumnTextCellViewBase` have a `value` property and to create a `TableColumnTextBaseColumnConfig` interface to enforce that the column configs used by `TableColumnTextCellViewBase` have an optional `placeholder` property.

## 🧪 Testing

- Updated/created unit tests
- Updated matrix tests to test each relevant column with & without a placeholder

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text-matrix.stories.ts
```diff
@@ -42,18 +42,26 @@ const alignmentStates: [string, string | undefined][] = Object.entries(
 ).map(([key, value]) => [pascalCase(key), value]);
 type AlignmentState = (typeof alignmentStates)[number];
 
+const placeholderStates: [string, string | undefined][] = [
+    ['With Placeholder', 'Custom placeholder'],
+    ['', undefined]
+];
```
m-akinc, line 48: Now that I've made this suggestion several times, I think it might be worthwhile to move this to `utilities/tests/states.ts`.

```suggestion
const placeholderStates = [
    ['With Placeholder', 'Custom placeholder'],
    ['', undefined]
] as const;
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text-matrix.stories.ts
```diff
@@ -42,18 +42,26 @@ const alignmentStates: [string, string | undefined][] = Object.entries(
...
+
 // prettier-ignore
 const component = (
-    [alignmentName, alignment]: AlignmentState
+    [alignmentName, alignment]: AlignmentState,
+    [placeholderName, placeholder]: PlaceholderState
 ): ViewTemplate => html`
     <label style="color: var(${controlLabelFontColor.cssCustomProperty}); font: var(${controlLabelFont.cssCustomProperty})">
-        Number Text Table Column with ${alignmentName} alignment
+        Number Text Table Column with ${alignmentName} alignment ${placeholderName}
```
m-akinc, line 57: Looks like our convention is to use title case for these labels.
```suggestion
        Number Text Table Column With ${alignmentName} Alignment ${placeholderName}
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.mdx
```diff
@@ -20,11 +20,16 @@ based on the value of the `lang` token, which can be set via the [nimble-theme-p
...
 
-{/* ## Usage */}
+## Usage
 
-## Angular Usage
+### Best Practices
 
-### Number Format Pipe
+-   Avoid mixing `undefined` and `null` as values for the same field. When grouping this will lead to two groups (one for `null` values and one for `undefined` values) that both have the text `"No value"`.
+-   If relevant to your data source, make sure to consider the IEEE 754 special cases of `-Inf`, `+Inf`, `-0`, `+0`, and `NaN`.
```
m-akinc, line 28: What is the guidance/best practice here? A client might infer that the column doesn't handle those values. What about these values do they need to consider?

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -739,50 +720,140 @@ describe('TableColumnNumberText', () => {
...
+                usesColumnPlaceholder: false
             }
         ];
 
         parameterizeSpec(testCases, (spec, name, value) => {
-            spec(`group row renders expected value when ${name}`, async () => {
-                await table.setData(value.data);
-                await connect();
-                await waitForUpdatesAsync();
+            spec(
```
m-akinc, line 772: Again, consider reusing test logic.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -739,50 +720,140 @@ describe('TableColumnNumberText', () => {
...
+            expect(pageObject.getRenderedCellTextContent(0, 0)).toBe(
+                placeholder
+            );
+
+            elementReferences.column1.placeholder = undefined;
+            await waitForUpdatesAsync();
+            expect(pageObject.getRenderedCellTextContent(0, 0)).toBe('');
+        });
+
+        it('setting placeholder to defined string updates cells from displaying placeholder to displaying blank', async () => {
```
m-akinc, line 827:
```suggestion
        it('setting placeholder to defined string updates cells from displaying blank to displaying placeholder', async () => {
```

packages/nimble-components/src/table-column/text/tests/table-column-text-matrix.stories.ts
```diff
@@ -36,25 +40,38 @@ const data = [
     }
 ] as const;
 
+const placeholderStates: [string, string | undefined][] = [
+    ['With Placeholder', 'Custom placeholder'],
+    ['', undefined]
+];
```
m-akinc, line 46:
```suggestion
const placeholderStates = [
    ['With Placeholder', 'Custom placeholder'],
    ['', undefined]
] as const;
```

packages/nimble-components/src/table-column/text/tests/table-column-text-matrix.stories.ts
```diff
@@ -36,25 +40,38 @@ const data = [
...
+type PlaceholderState = (typeof placeholderStates)[number];
+
 // prettier-ignore
-const component = (): ViewTemplate => html`
+const component = (
+    [placeholderName, placeholder]: PlaceholderState
+): ViewTemplate => html`
+<label style="color: var(${controlLabelFontColor.cssCustomProperty}); font: var(${controlLabelFont.cssCustomProperty})">
+    Text Table Column ${placeholderName}
+</label>
```
m-akinc, line 55:
```suggestion
    <label style="color: var(${controlLabelFontColor.cssCustomProperty}); font: var(${controlLabelFont.cssCustomProperty})">
        Text Table Column ${placeholderName}
    </label>
```

packages/nimble-components/src/table-column/text/tests/table-column-text.mdx
```diff
@@ -16,7 +16,13 @@ The <Tag name={tableColumnTextTag}/> column is used to display string fields as
...
 {/* ## Appearance Variants */}
 
-{/* ## Usage */}
+## Usage
+
+### Best Practices
+
+-   Avoid mixing `undefined` and `null` as values for the same field. When grouping this will lead to two groups (one for `null` values and one for `undefined` values) that both have the text `"No value"`.
+-   Avoid mixing empty string with `undefined`/`null`. The distinction when grouping between `"No value"` and `"Empty"` is not likely meaningful to a user.
+-   Avoid displaying whitespace values that are not empty string (`''`) as these values will be rendered as-is in group rows.
```
m-akinc, line 25: These all essentially boil down to:
-   Use a consistent representation for empty values in your data. Do not mix `null`, `undefined`, and/or the empty string (`''`).
    -   If you want cells to display placeholder text, use `null` or `undefined`. Empty strings are not replaced by placeholder text and will always render as a blank cell. 
    -   Avoid whitespace-only data values, as they can lead to confusing results when grouping or sorting.

This actually makes me wonder why we don't automatically normalize whitespace-only values to the empty string. Is there any reasonable argument not to?

packages/nimble-components/src/table-column/text/tests/table-column-text.spec.ts
```diff
@@ -237,40 +218,126 @@ describe('TableColumnText', () => {
...
             {
                 name: 'value is incorrect type',
                 data: [{ field: 100 as unknown as string }],
-                groupValue: ''
+                cellValue: '',
+                groupValue: '',
+                usesColumnPlaceholder: false
             },
             {
                 name: 'value is an empty string',
```
m-akinc, line 247: Maybe we should also have a case for a whitespace-only string.

packages/nimble-components/src/table-column/text/tests/table-column-text.spec.ts
```diff
@@ -237,40 +218,126 @@ describe('TableColumnText', () => {
...
             }
         ];
 
         parameterizeSpec(testCases, (spec, name, value) => {
-            spec(`group row renders expected value when ${name}`, async () => {
-                await table.setData(value.data);
-                await connect();
-                await waitForUpdatesAsync();
+            spec(
+                `cell and group row render expected value when ${name} and placeholder is configured`,
```
m-akinc, line 257: Consider reusing test logic.

packages/nimble-components/src/table-column/text/tests/table-column-text.spec.ts
```diff
@@ -237,40 +218,126 @@ describe('TableColumnText', () => {
...
+            expect(pageObject.getRenderedCellTextContent(0, 0)).toBe(
+                placeholder
+            );
+
+            column.placeholder = undefined;
+            await waitForUpdatesAsync();
+            expect(pageObject.getRenderedCellTextContent(0, 0)).toBe('');
+        });
+
+        it('setting placeholder to defined string updates cells from displaying placeholder to displaying blank', async () => {
```
m-akinc, line 311:
```suggestion
        it('setting placeholder to defined string updates cells from displaying blank to displaying placeholder', async () => {
```

packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.mdx
```diff
@@ -12,11 +12,28 @@ The <Tag name={tableColumnAnchorTag}/> column is used to display string fields a
 <Canvas of={tableColumnAnchorStories.anchorColumn} />
 <Controls of={tableColumnAnchorStories.anchorColumn} />
 
-## Target Configuration
+## Usage
+
+### Best Practices
```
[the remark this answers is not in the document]
mollykreis: I've reduced the information to what you've suggested. The full information is still in the HLD, so we can always refer back to it if we need more information and/or justifications for the best practices.

packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.mdx
```diff
@@ -12,11 +12,28 @@ The <Tag name={tableColumnAnchorTag}/> column is used to display string fields a
...
+
+-   Provide useful labels for well known urls. While an absent label will show the full URL for accessibility, it is useful to instead provide a clear and unique label to improve grouping.
```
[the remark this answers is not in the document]
mollykreis: No longer applicable since I've adopted a different suggestion that drastically changes the wording in the best practices.

packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.spec.ts
```diff
@@ -465,4 +452,160 @@ describe('TableColumnAnchor', () => {
...
+
+                    expect(pageObject.getRenderedCellTextContent(0, 0)).toBe(
+                        value.cellValue
+                    );
+                    expect(
+                        pageObject.getRenderedGroupHeaderTextContent(0)
+                    ).toBe(value.groupValue);
+                }
+            );
+        });
```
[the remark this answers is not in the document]
mollykreis: This doesn't align with our goal of keeping all `expect` statements within each actual test, but I will create a helper function that reduces the amount of duplication around setting a placeholder, setting the data, and waiting for that to be applied.

packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.spec.ts
```diff
@@ -16,41 +17,54 @@ interface SimpleTableRecord extends TableRecord {
...
-                    download="download value"
-                    group-index="0"
-                >
-                    Column 1
-                </${tableColumnAnchorTag}>
-                <${tableColumnAnchorTag}>
-                    Column 2
-                </${tableColumnAnchorTag}>
-            </nimble-table>`
+        html`<${themeProviderTag} lang="en-US">
```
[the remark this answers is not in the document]
mollykreis: I was thinking I needed it for the group labels to be correct from the label provider, but I don't think that's actually the case. I've removed it.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -518,86 +476,189 @@ describe('TableColumnDateText', () => {
...
+
+                            expect(
+                                pageObject.getRenderedCellContent(0, 0)
+                            ).toBe(value.cellValue);
+                            expect(
+                                pageObject.getRenderedGroupHeaderContent(0)
+                            ).toBe(value.groupValue);
+                        }
+                    );
+                });
```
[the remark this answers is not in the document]
mollykreis: Like the other case, I don't think we want to move the whole test to a helper function, but I've moved the setup to a helper function.

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.spec.ts
```diff
@@ -185,60 +185,146 @@ describe('TableColumnDurationText', () => {
...
-            spec(`group row renders expected value when ${name}`, async () => {
-                await table.setData(value.data);
-                await connect();
-                await waitForUpdatesAsync();
-
-                expect(pageObject.getRenderedGroupHeaderContent(0)).toBe(
-                    value.groupValue
-                );
-            });
+            spec(
```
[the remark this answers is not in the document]
mollykreis: Same here -- created helper function for setup

packages/nimble-components/src/table-column/icon/tests/table-column-icon.mdx
```diff
@@ -17,7 +17,15 @@ The <Tag name={tableColumnIconTag}/> column renders string, number, or boolean v
...
 <Controls of={tableColumnIconStories.iconColumn} />
 
-## Blazor Usage
+## Usage
+
+### Best Practices
+
+-   Avoid mixing `undefined` and `null` as values for the same field. When grouping this will lead to two groups (one for `null` values and one for `undefined` values) that both have the text `"No value"`.
+-   Avoid using values that do not correspond to a mapping for the column.
+-   To display an empty cell but have a non-blank group row, create a mapping of the record value to an `undefined` icon.
```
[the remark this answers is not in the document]
mollykreis: I'm not sure how mapping values to an `undefined` icon improves sorting behavior, but I included the rest of the suggestion.

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text-matrix.stories.ts
mollykreis: Moved to `utilities/tests/states.ts`.

packages/nimble-components/src/table-column/text/tests/table-column-text.mdx
mollykreis: I don't think the table should be modifying the data its given, which I think is what you are suggesting. Is that correct?

packages/nimble-components/src/table-column/mixins/placeholder.ts
```diff
@@ -0,0 +1,32 @@
...
+    TBase extends TableColumnWithPlaceholderConstructor
+>(base: TBase) {
+    /**
+     * The Mixin that provides a concrete column with the API to allow grouping
+     * by the values in that column.
+     */
+    abstract class ColumnWithPlaceholder extends base {
+        public placeholder?: string;
+
+        public abstract placeholderChanged(): void;
```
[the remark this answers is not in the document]
mollykreis: I debated on doing that. I was torn on whether it was better to have every column implement `placeholderChanged()` and do whatever they needed or to force each column that uses this mixin to have a `updateColumnConfig()` function. I landed on it being better to have columns implement `placeholderChanged()` with the flexibility of structuring the rest of their class however they wanted rather than require all columns with placeholders to implement updates through a `updateColumnConfig()` function.

I'm willing to be convinced otherwise.

packages/nimble-components/src/table-column/mixins/placeholder.ts
```diff
@@ -0,0 +1,32 @@
+import { attr } from '@microsoft/fast-element';
+import type { TableColumn } from '../base';
+
+// Pick just the relevant properties the mixin depends on (typescript complains if the mixin declares private / protected base exports)
+// Because the 'placeholder' mixin doesn't depend on any properties of the TableColumn, there are no properties to pick.
+type TableColumnWithPlaceholder = Pick<TableColumn, never>;
+// prettier-ignore
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
mollykreis: Looking at this comment again, I think you mean on line 9. Is that right? If so, no, that doesn't work because then a TypeScript error occurs: _A mixin class must have a constructor with a single rest parameter of type 'any[]'._

This isn't the prettiest code, but it's all following our established patterns for mixins, which are already used in `fractional-width-column.ts` and `groupable-column.ts`.

packages/nimble-components/src/table-column/icon/tests/table-column-icon.mdx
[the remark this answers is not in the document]
m-akinc: I thought that for both grouping and sorting, this helped by normalizing the value used to group/sort rather than a bunch of different values. But I might have gotten that wrong.

packages/nimble-components/src/table-column/mixins/placeholder.ts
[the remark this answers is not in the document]
m-akinc: Yeah, I was pointing at the suppression, thinking we might be able to use `unknown` instead of suppressing.

packages/nimble-components/src/table-column/text/tests/table-column-text.mdx
m-akinc: Do we not create a copy of the data passed to `setData` at any point? I wasn't quite sure. If not, and tanstack is sorting/grouping based on those values directly, then I guess we can't really do what I suggested.

m-akinc (approved):

packages/nimble-components/src/table-column/icon/tests/table-column-icon.mdx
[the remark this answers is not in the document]
mollykreis: You are right that since sorting is done by record value, having a bunch of different record values that display as the same thing (whether they display as the same mapped value or as an empty unmapped value) would be weird when it comes to sorting and grouping. However, I don't think having mappings for everything solves that problem. I'll add an additional bullet point here that says something like _Avoid having multiple values that map to the same label because grouping and sorting is done on the record value rather than the mapped label._

packages/nimble-components/src/table-column/text/tests/table-column-text.mdx
mollykreis: We do have our own copy of the data, but I don't think the fact we have a copy of the data makes it ok to change the data. Some things that could break if we "magically" change the data are:

- IDs of rows may not be as expected/specified by the client
- Mappings may not work as expected
- Parenting (using IDs) may not work correctly

Granted, those things would only break if people are doing odd things (like having IDs that are whitespace), but it would still be very difficult for a client to debug & understand the behavior if we're essentially using different data than the client gave the table. I think it would be much better for a client to normalize their data rather than the table trying to be too smart about it.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.mdx
```diff
@@ -16,7 +16,11 @@ The <Tag name={tableColumnDateTextTag}/> column is used to display date-time fie
```
jattasNI, line 16: The docs for the date and duration columns are showing errors for me in the latest storybook build.

image

https://60e89457a987cf003efc0a5b-wlnihdzmym.chromatic.com/?path=/docs/components-table-column-duration-text--docs
https://60e89457a987cf003efc0a5b-wlnihdzmym.chromatic.com/?path=/docs/components-table-column-date-text--docs

mollykreis: It should be fixed now. I also ended up moving the location of the shared docs for not mixing `undefined` and `null` so that it renders correctly in the docs.

packages/nimble-components/src/table-column/duration-text/tests/table-column-duration-text.stories.ts
```diff
@@ -94,9 +96,14 @@ export const durationTextColumn: StoryObj<TextColumnTableArgs> = {
...
+        },
+        placeholder: {
+            description:
+                'The placeholder text to display when the field value is `undefined` or `null` for a record.'
         }
     },
     args: {
-        fieldName: 'firstName'
+        fieldName: 'firstName',
+        placeholder: 'Unknown value'
```
rajsite, line 107:
```suggestion
        placeholder: 'Unknown time'
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -739,50 +720,137 @@ describe('TableColumnNumberText', () => {
...
             },
             {
                 name: 'value is specified and falsey',
                 data: [{ number1: 0 }],
-                groupValue: '0'
+                cellValue: '0',
+                groupValue: '0',
+                usesColumnPlaceholder: false
             }
         ];
```
rajsite, line 769:
```suggestion
        ] as const;
```

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.stories.ts
```diff
@@ -223,6 +228,7 @@ export const numberTextColumn: StoryObj<NumberTextColumnTableArgs> = {
         alignment: 'default',
         decimalDigits: 2,
         decimalMaximumDigits: undefined,
-        unit: 'volt'
+        unit: 'volt',
+        placeholder: 'Unknown value'
```
rajsite, line 232:
```suggestion
        placeholder: 'Unknown voltage'
```

packages/nimble-components/src/table-column/text/tests/table-column-text.spec.ts
```diff
@@ -237,40 +218,130 @@ describe('TableColumnText', () => {
...
+                usesColumnPlaceholder: false
+            },
+            {
+                name: 'value is a string containing only whitespace',
+                data: [{ field: '   ' }],
+                cellValue: '',
+                groupValue: '',
+                usesColumnPlaceholder: false
             }
         ];
```
rajsite, line 260:
```suggestion
        ] as const;
```

packages/nimble-components/src/table-column/text/tests/table-column-text.stories.ts
```diff
@@ -94,9 +96,14 @@ export const textColumn: StoryObj<TextColumnTableArgs> = {
...
+        },
+        placeholder: {
+            description:
+                'The placeholder text to display when the field value is `undefined` or `null` for a record.'
         }
     },
     args: {
-        fieldName: 'firstName'
+        fieldName: 'firstName',
+        placeholder: 'Unknown value'
```
rajsite, line 107:
```suggestion
        placeholder: 'Did not respond to request for comment'
```

packages/nimble-components/src/table-column/text/index.ts
```diff
@@ -22,6 +25,12 @@ declare global {
  * The table column for displaying string fields as text.
  */
 export class TableColumnText extends TableColumnTextBase {
```
rajsite, line 27: wait, just realized.. why are we not passing TableColumnTextColumnConfig to the parent class to fulfill the generic for the TableColumn base class? I'm guessing the typing of columnInternals isn't right.

May have been not a good idea to let TableColumn default to having ColumnConfig Type `unknown`.

packages/nimble-components/src/table-column/anchor/cell-view/index.ts
```diff
@@ -26,11 +26,25 @@ TableColumnAnchorColumnConfig
...
+
     /** @internal */
     public anchor?: Anchor;
 
     @volatile
     public get text(): string {
+        const displayedText = this.cellRecord?.label ?? this.cellRecord?.href;
+        if (
+            (displayedText === undefined || displayedText === null)
+            && this.columnConfig?.placeholder
```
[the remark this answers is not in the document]
rajsite: Actually maybe empty string should be allowed as thee placeholder string?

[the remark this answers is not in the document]
mollykreis: Sure, I can allow empty string as the placeholder and add some tests for it. I don't think it'll actually make a big difference from a client point of view (at least with our current columns) because values that would use the placeholder (`null`/`undefined`) get rendered as empty strings when the placeholder isn't configured. Therefore, an `undefined` placeholder and an empty placeholder would render the exact same way in a cell.

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -522,83 +480,177 @@ describe('TableColumnDateText', () => {
...
+                    usesColumnPlaceholder: false
                 }
             ];
 
+            async function initializeColumnAndTable(
+                data: readonly SimpleTableRecord[],
+                placeholder?: string
+            ): Promise<void> {
+                // Set a custom time zone so that the behavior of the test does not
+                // depend on the configuration of the computer running the tests.
```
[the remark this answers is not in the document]
mollykreis: The karma config looks like it is only used for Chrome, so I'm guessing we still want to keep the code in table-column-date-text.spec.ts so that the timezone is correct regardless of what browser is running the tests locally. Does that sound right? Should I still add the timezone to Chrome's karma config and the github action config?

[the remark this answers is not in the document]
rajsite: Yea we should keep it in the spec.ts 👍
Just trying to do as much as possible to keep environment across CI runs and CI vs local dev consistent (chrome configured the same for local tests in karma and CI configured so if agents run in different timezones)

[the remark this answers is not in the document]
rajsite: Could also do that in a small standalone PR so CI changes aren't impacting this one

packages/nimble-components/src/table-column/text/index.ts
mollykreis: We aren't passing the generic type because of our mixins. TypeScript doesn't support mixins + generics (see https://github.com/microsoft/TypeScript/issues/26154).

I still playing around with the types, but I haven't found a way to get the generic passed from a concrete implementation (say TableColumnText) all the way to the TableColumn base class.

rajsite: Doesn't have to block the PR as discussed at stand-up 👍

rajsite (approved):

packages/nimble-components/src/table-column/text/index.ts
mollykreis: Created #1964

PR #1954, opened 2024-03-20
Fix sizing test in WebKit

### LINKED ISSUE #1939: Disabled interactive column sizing test on WebKit
# 🧹 Tech Debt

In `packages/nimble-components/src/table/tests/table-column-sizing.spec.ts` the test `sizing table with a horizontal scrollbar does not change column widths until sized beyond current column pixel widths` is being skipped on WebKit because it fails with `Expected 499.984375 to be 500.`

The test either needs to be fixed and re-enabled or an issue filed upstream, e.g. to webkit, and this issue marked blocked.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1939 

## 👩‍💻 Implementation

I updated the test to use sizes that cause the columns to be whole numbers to avoid rounding precision issues as was previously seen. Additionally, the fractional widths now compute to values without repeating decimals, which contributes to the final widths being more consistent across browsers.

## 🧪 Testing

Ran `npm run test-chrome && npm run test-firefox && npm run test-webkit` with the test at hand focused and verified that the test passes in all three browsers

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f0730b00-2a69-4d06-8241-d97c34bf98b1.json
+++ b/change/@ni-nimble-components-f0730b00-2a69-4d06-8241-d97c34bf98b1.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Fix column sizing test to pass in webkit",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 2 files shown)


mollykreis:

atmgrifter00 (approved):

rajsite:
@jattasNI pretty straightforward so bypassing

rajsite (approved):

PR #1955, opened 2024-03-21
Created a new Experimental Prerendering module

munteannatan:
# Pull Request

## 🤨 Rationale

Created a new experimental prerendering module.

The previous prerendering oversaw calculating the `fillstyle` of each die and it's label.
We want to move these expensive operations to the web worker, and to make them faster.

## 👩‍💻 Implementation

I removed the code which prepared the dies for rendering.
I created a domain splitter for linear color scales. this will reduce the number of colors displayed but will increase the performance because we are no longer finding interpolations for each individual value in the whole domain.

Open Issues:
- given the problems with arquero, should we implement in-house highlight or leave it unimplemented until the issue gets fixed?

## ✅ Checklist


packages/nimble-components/src/wafer-map/modules/experimental/prerendering.ts
```diff
@@ -0,0 +1,85 @@
...
+        colors: string[],
+        values: number[]
+    } {
+        return this._colorScale;
+    }
+
+    private _colorScale!: {
+        colors: string[],
+        values: number[]
+    };
```
atmgrifter00, line 25: Why a type like this instead of doing something like:
```
interface IColorScaleMarker {
    color: string,
    value: number
}

export type ColorScale = IColorScaleMarker[];
```

Having two independent arrays that I would assume are supposed to be the same length seems an unnecessary type decision.

munteannatan: I was thinking that searching on a typed array for values will be faster in the worker, but I can try to use it like you suggested and see the impact

packages/nimble-components/src/wafer-map/modules/experimental/prerendering.ts
```diff
@@ -0,0 +1,86 @@
...
+    private _colorScale!: ColorScale;
+
+    private _labelsFontSize!: number;
+
+    private readonly fontSizeFactor = 0.8;
+    private readonly colorScaleResolution = 10;
+
+    public constructor(private readonly wafermap: WaferMap) {}
+
+    public update(): void {
```
atmgrifter00, line 28: This seems, to me, an awkward dependency pattern. It seems strange to have an `update` method that the call site actually has no semantic idea of what that `update` does. Calling `update` on the pre-renderer for every various `update<x>` method in the `DataManager` doesn't seem like an obvious requirement. Is it? How do know when to use/not use it?

The fact that there is nothing observable happening with respect to the changes to the `_colorScale` and `_labelsFontSize` properties is also concerning.

While I don't have a precise suggestion for a better direction to move in, I would recommend looking at other patterns we have that have _observable_ dependencies from one to the other, like the `ColumnInternals` or the `Virtualizer`.

munteannatan: Sure! I created this Issue to track these changes after we have all the pieces of the new update together https://github.com/ni/nimble/issues/1984

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -11,17 +12,21 @@ export class RenderingModule {
...
-            if (a.fillStyle > b.fillStyle) {
-                return 1;
-            }
-            if (b.fillStyle > a.fillStyle) {
-                return -1;
-            }
+        if (this.wafermap.dataManager instanceof DataManager) {
+            this.dies = this.wafermap.dataManager.diesRenderInfo.sort(
+                (a, b) => {
+                    if (a.fillStyle > b.fillStyle) {
```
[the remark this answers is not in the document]
munteannatan: This was just meant to cluster all the values with the same fill style together, so we do not have to change the canvas context so often. this became less of a limitation over time and even a bottleneck and is one of the reasons we are doing a new rendering startegy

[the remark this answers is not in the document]
atmgrifter00: I guess I still don't have a clear picture of whether this implementation is supposed to change. Do you intend to modify this so that you are making a numerical comparison, or is there going to no longer be a need to do this sort, or neither of those?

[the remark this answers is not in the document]
atmgrifter00: @munteannatan, just making sure you are aware I unresolved this comment thread.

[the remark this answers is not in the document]
munteannatan: Sorry, I did not get the notification in time.

In the new implementation we are firstly limiting the actual number of colors we display, like you see in this PR. This means a reduced number of fillstyles to change between.

Secondly, we are no longer storing the value for each die, we are working with a dictionary and keys instead, so we remove the data duplication as much as we can.

Thirdly, we are iterating over typed arrays instead of the list of dies, which is a lot faster and gives us improved rendering performance.

And finally, we are filtering dies out from rendering based on the zoom level. This is particularly important for big wafers, so we are implementing a caching technique where we are clustering based on the columns, so we can remove the whole column at once without iterating over the dies in the column at all, so the color clustering is no longer beneficial.

atmgrifter00 (approved):

jattasNI:
I'll be OOO this week, feel free to complete without my review.

packages/nimble-components/src/wafer-map/modules/hover-handler.ts
```diff
@@ -52,7 +52,9 @@ export class HoverHandler {
             return;
         }
 
-        this.wafermap.hoverDie = this.wafermap.dataManager.getWaferMapDie(dieCoordinates);
+        if (this.wafermap.dataManager instanceof DataManager) {
```
rajsite, line 55: I think this check should be similar to the one below, i.e. `!this.wafermap.isExperimentalRenderer()
            && this.wafermap.dataManager instanceof DataManager`

Instead of duplicating that everywhere we can add a method to the wafer map?

```ts
public isLegacyDataManager() {
    return !this.wafermap.isExperimentalRenderer()
            && this.wafermap.dataManager instanceof DataManager;
}
```

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -11,17 +12,21 @@ export class RenderingModule {
...
     public updateSortedDiesAndDrawWafer(): void {
-        this.dies = this.wafermap.dataManager.diesRenderInfo.sort((a, b) => {
-            if (a.fillStyle > b.fillStyle) {
-                return 1;
-            }
-            if (b.fillStyle > a.fillStyle) {
-                return -1;
-            }
+        if (this.wafermap.dataManager instanceof DataManager) {
+            this.dies = this.wafermap.dataManager.diesRenderInfo.sort(
```
rajsite, line 16: Calling sort will [mutate `diesRenderInfo` in place](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). I think that is surprising that rendering will mutate values in thee dataManager.

Maybe it should be a shallow copy, i.e. `[...this.wafermap.dataManager.diesRenderInfo].sort()`?
Or we should be explicit about mutating diesRenderInfo within dataManager instead of in the renderer.

packages/nimble-components/src/wafer-map/modules/rendering.ts
```diff
@@ -11,17 +12,21 @@ export class RenderingModule {
...
 
     public updateSortedDiesAndDrawWafer(): void {
-        this.dies = this.wafermap.dataManager.diesRenderInfo.sort((a, b) => {
-            if (a.fillStyle > b.fillStyle) {
-                return 1;
-            }
-            if (b.fillStyle > a.fillStyle) {
-                return -1;
-            }
+        if (this.wafermap.dataManager instanceof DataManager) {
```
rajsite, line 15: This should share the DataManageer check discussed above

packages/nimble-components/src/wafer-map/modules/prerendering.ts
```diff
@@ -51,7 +51,29 @@ export class Prerendering {
...
+        )
+            .map(row => {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                return {
+                    x: row.colIndex,
+                    y: row.rowIndex,
+                    value: row.value,
+                    tags: []
+                };
+            })
```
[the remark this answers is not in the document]
munteannatan: this was a merge issue from the previous branch, I removed this

packages/nimble-components/src/wafer-map/modules/rendering.ts
munteannatan: this appears changed because the type check of the data manager. this is the old rendering code which will get replaced

packages/nimble-components/src/wafer-map/modules/hover-handler.ts
munteannatan: it seems typescript does not see the type constraints if I'm using a function, should I try and make a separate update function and get rid of this switching mechanism?

rajsite (approved):

rajsite:
@munteannatan you don't need to wait for Jesse. I don't know about the remaining owners @DStavilaNI @zszilagy

PR #1958, opened 2024-03-22
Fix race condition in drawer animations

### LINKED ISSUE #1937: Disabled drawer test on Firefox due to animationend handler not being called
# 🧹 Tech Debt

The drawer test `should resolve promise if drawer completely opens before being closed` fails on Firefox with a timeout: `Failed: Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)`

It passes in other browsers. (Note this is a different issue than #1936)

It appears the `animationend` handler that we attach while showing the dialog is never called. 

This issue still needs to be have a reproduction created and be reported to Mozilla / webcompat.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1937 

## 👩‍💻 Implementation

Based on #1937, the issue is that the `animationend` handler isn't being called. That is true, but the root cause of the bug is that the close animation isn't starting (and therefore not ending). Specifically, the timing exposed in this test is:

- Open the drawer
- Wait for the drawer's open animation to complete
- Attempt to immediately start closing the drawer
    - The drawer doesn't actually close because the browser doesn't have time to fully process the `animating` class being removed from drawer in order to remove the animation in order for re-adding the `animating` class to cause the animation to start again.

I tried a few different things to try to resolve the issue:
- Update the test to have a delay between the drawer opening and closing
   - This fixed the test, but I decided that ultimately, the issue wasn't with the test but with the drawer element. There shouldn't be a time window (even if it is short) in which calling `close` doesn't work.
- Call `await DOM.nextUpdate` before starting the animation.
    - This worked as long as I called `DOM.nextUpdate` twice (not sure why it was needed twice). I decided against this because (1) the call was needed twice, which isn't ideal, (2) it forced previously synchronous code paths to be async, and (3) it would break a number of tests in nimble and likely even more client tests. While none of those are insurmountable issues, collectively, it was enough to make me question whether or not that was the best solution.
- Wrap the code that starts the animation in `setTimeout` with a timeout of 0.
    - Similar to the previous alternative, this would break a number of tests in nimble and likely even more client tests. This isn't insurmountable, but ideally we can find a solution that doesn't break client tests.
- Re-write the animation code
    - This isn't a feasible solution because [Firefox doesn't support programmatic animations of the `::backdrop` psuedo element](https://bugzilla.mozilla.org/show_bug.cgi?id=1770591).

Therefore, I landed on the solution to trigger a reflow by reading the offsetHeight of the dialog before adding the `animating` class to start the animation. One major downside of this approach is that the code itself doesn't make it clear why it is needed, but I think a comment explaining the line goes a long way in mitigating that concern. The big positive is that it doesn't have a large impact on the timing from a client or test perspective. Therefore, I expect significantly less (hopefully no) client repercussions associated with the change.

## 🧪 Testing

Ran all the drawer tests in chrome, FF, and webkit to make sure the change didn't introduce any more issues.

## ✅ Checklist


mollykreis:
@msmithNI, will you buddy this PR for me? I'm interested if you have any other ideas for how to solve this problem that are better than what I've implemented.

m-akinc (approved):

msmithNI (approved):

msmithNI:
> @msmithNI, will you buddy this PR for me? I'm interested if you have any other ideas for how to solve this problem that are better than what I've implemented.

The only other thought I had was that we could explore using the (JavaScript) Web Animations API instead of the current CSS animation approach. But I took a look at the dialog styles and that's probably not feasible due to the animations on the backdrop (I'm not aware of a JS equivalent to the `::backdrop` pseudo-selector).

rajsite:
> > @msmithNI, will you buddy this PR for me? I'm interested if you have any other ideas for how to solve this problem that are better than what I've implemented.
> 
> The only other thought I had was that we could explore using the (JavaScript) Web Animations API instead of the current CSS animation approach. But I took a look at the dialog styles and that's probably not feasible due to the animations on the backdrop (I'm not aware of a JS equivalent to the `::backdrop` pseudo-selector).

There is a pseudoElement option for element.animate: https://caniuse.com/mdn-api_element_animate_options_pseudoelement_parameter

Haven't tried it with `::backdrop` specifically

packages/nimble-components/src/drawer/index.ts
```diff
@@ -131,6 +131,12 @@ export class Drawer<CloseReason = void> extends FoundationElement {
     }
 
     private triggerAnimation(): void {
+        // Read the offsetHeight of the dialog to trigger a reflow. This guarantees that the browser
+        // has processed the 'animating' class being removed before trying to readd it, even if the previous
+        // animation has just finished. Otherwise, problems can occur. For example, trying to close the
+        // drawer immediately after the opening animation ends does not actually close the drawer.
+        void this.dialog.offsetHeight;
```
mollykreis, line 138: Moving comment to discussion:

>> The only other thought I had was that we could explore using the (JavaScript) Web Animations API instead of the current CSS animation approach. But I took a look at the dialog styles and that's probably not feasible due to the animations on the backdrop (I'm not aware of a JS equivalent to the ::backdrop pseudo-selector).

>There is a pseudoElement option for element.animate: https://caniuse.com/mdn-api_element_animate_options_pseudoelement_parameter
> 
>Haven't tried it with ::backdrop specifically

mollykreis: I tried using the javascript web animations API as suggested, but it turns out the `::backdrop` psuedo element isn't supported in FF. https://bugzilla.mozilla.org/show_bug.cgi?id=1770591

FYI - @rajsite

rajsite: Ahh dang, lovely. FWIW this isn't the first time this is being done for FF in the monorepo:

https://github.com/ni/nimble/blob/8a0315cd3a2789f38ad69d8fbeab0d88c085b253/packages/site/landing/circle.ts#L14-L16

Can't think of much else to do. Not sure if it's worth tracking as a tech debt item to switch to web animations when possible. Maybe doesn't hurt to capture the details in an item and link from here.

rajsite (approved):

PR #1959, opened 2024-03-25
Update date-text column tests to pass on all browsers

### LINKED ISSUE #1940: Disabled TableColumnDateText tests on WebKit due to date formatting differences
# 🧹 Tech Debt

Several TableColumnDateText tests have been disabled on WebKit due to errors like this:

`        Expected 'Dec 10, 2012 at 10:35:05 PM' to be 'Dec 10, 2012, 10:35:05 PM'.`

We need to either fix and re-enable the tests or file an issue upstream (e.g. to webkit) and mark this blocked.

Search for this bug ID in source for a complete list of such tests.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1940 

## 👩‍💻 Implementation

Added a page object function to get the expected cell text for a given value/locale combination. The function internally uses `Intl.DateTimeFormat` to determine the expected text, so that it will always return the correct value for a given browser. With this change, the tests still test all the column, cell, and group row formatting code without actually testing the nuances of the string returned from `Intl.DateTimeFormat`.

## 🧪 Testing

Ran the tests for the date-text column in WebKit on my machine (Windows 10) and on the pipeline to verify they pass on both.

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -235,14 +246,17 @@ describe('TableColumnDateText', () => {
             expect(pageObject.getCellTitle(0, 0)).toEqual('');
         });
 
-        // WebKit skipped, see https://github.com/ni/nimble/issues/1940
-        it('sets group header text to rendered date value #SkipWebkit', async () => {
-            await table.setData([
-                { field: new Date('Dec 10, 2012, 10:35:05 PM').valueOf() }
-            ]);
+        it('sets group header text to rendered date value', async () => {
+            table.style.width = '200px';
```
m-akinc, line 250: We don't need to set the width, right?

change/@ni-nimble-components-26aadfad-c24c-4ad0-ae8e-e75769665e95.json
m-akinc, line 1: So the issue being addressed is that `Intl.DateTimeFormat` returns different strings for Chromium and WebKit? I'm curious how it differs, but I think this approach is reasonable as long as our docs don't give specifics about the format that we display.

m-akinc (approved):

change/@ni-nimble-components-26aadfad-c24c-4ad0-ae8e-e75769665e95.json
mollykreis: That is correct.

Today our storybook docs say _By default, dates are formatted like "Jan 1, 2023, 12:00:00 AM"._ Depending on the reader's interpretation of "like" in that sentence, that may or may not be true. Do you think we should update that sentence to be less specific? Something like _By default, dates are formatted using `medium` `dateStyle` and `timeStyle` options with the Intl.DateTimeFormat._

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -78,32 +78,41 @@ describe('TableColumnDateText', () => {
             expect(column.checkValidity()).toBeTrue();
```
jattasNI, line 78: I pulled your branch and verified that the tests in this file pass on macOS 14.3 in Safari 17.3 and Playwright's Webkit browser.

jattasNI (approved):

packages/nimble-components/src/table-column/date-text/tests/table-column-date-text.spec.ts
```diff
@@ -78,32 +78,41 @@ describe('TableColumnDateText', () => {
...
         });
 
-        // WebKit skipped, see https://github.com/ni/nimble/issues/1940
-        it('changing fieldName updates display #SkipWebkit', async () => {
+        it('changing fieldName updates display', async () => {
+            const fieldValue = new Date('Dec 10, 2012, 10:35:05 PM').valueOf();
+            const anotherFieldValue = new Date(
```
rajsite, line 83: nit: can we use the names `expected` and `actual` for these where appropriate? Looks like we don't tend to save actual to named variables which is fine, but can we name these `expected` value vars as such, i.e. `expectedFieldValue`?

rajsite (approved):

change/@ni-nimble-components-26aadfad-c24c-4ad0-ae8e-e75769665e95.json
rajsite: Could include a note like "the exact formatting of the resulting date time string is browser-specific"

PR #1960, opened 2024-03-26
Table keyboard navigation HLD

msmithNI:
# Pull Request

## 🤨 Rationale

HLD for #1137 

## 👩‍💻 Implementation

Started with the HLD document from #1506 , and updated it to reflect our decision to standardize on the ARIA treegrid role (moved the info about the grid role/ previous IxD design to the Alternatives section), and added an explicit section with the specific interactions as they'll apply to the Nimble table.

Note: Currently there isn't a prototype that fully reflects the interactions detailed in this document, but that's in-progress.


packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+
+Since the entire group header row is a button, the explicit expand/collapse button in that row is not keyboard focusable.
+
+If a group header row is focused, pressing `Space` or `Enter` will expand or collapse that group.
+
+### ARIA roles
+
+Assigning appropriate ARIA roles to the various elements inside the table is critical for technologies like screen readers to work as intended. We believe the following ARIA designations will provide the expected behaviors:
+
+-   `role="rows"` : Provided to both the `TableRow` element and the element hosting all of the column header content
```
atmgrifter00, line 117:
```suggestion
-   `role="row"` : Provided to both the `TableRow` element and the element hosting all of the column header content
```

atmgrifter00: I think the `TableGroupRow` also needs to have this role (which it does now).

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+
+`Ctrl-Alt modifiers`: It appears to be a common convention for screen readers to leverage `Ctrl-Alt Arrows` to move the screen reader focus to each of the cells in a row. As such, it may be desirable for our implementation to explicitly not handle keyboard interactions that involve both modifier keys, and instead respond to any resulting focus change, and update any internal state as needed based on the newly focused element. This is what the ARIA `treegrid` example does. It's not clear whether this behavior will require special handling by us or not, but we should test this scenario.
+
+#### Row Selection
+
+If the table has `selection-mode` of `single` or `multiple`:
+
+-   If a row is focused, pressing `Space` will select/unselect that row
+-   If a row's selection checkbox is focused, pressing `Space` will toggle the checkbox, selecting/unselecting that row
+-   **(Open question)** If a single cell is focused, should pressing `Space` select/unselect that row?
```
fredvisser, line 103: I'd vote "no"

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+    Cons:
+
+    - Reliance on modifier keys when there are available workflows/keys that do not require them (i.e. strictly using `ArrowLeft` and `ArrowRight` keys with no modifiers)
+    - Mistypes (attmpted to use `Alt` modifier but didn't) shifts focus unexpectedly.
+
+## Open Issues
+
+-   Row selection: Should pressing `Space` when a single cell is focused, result in selecting that row?
+-   When focus leaves the table and later returns, should we return focus to the previously focused row, or restart focus on the top row/ header?
+-   Are we satisfied with the keyboard interactions dealing with the cell action menus? Specifically, they can be `Tab`'d to, but arrow keys will not stop at them specifically (since they're part of a cell, not their own cell). So it may require multiple key presses to focus and activate them (or the direct `Ctrl-Enter` key combo if the cell is focused).
```
atmgrifter00, line 180: I am personally fine with this.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+
+    Cons:
+
+    - Reliance on modifier keys when there are available workflows/keys that do not require them (i.e. strictly using `ArrowLeft` and `ArrowRight` keys with no modifiers)
+    - Mistypes (attmpted to use `Alt` modifier but didn't) shifts focus unexpectedly.
+
+## Open Issues
+
+-   Row selection: Should pressing `Space` when a single cell is focused, result in selecting that row?
+-   When focus leaves the table and later returns, should we return focus to the previously focused row, or restart focus on the top row/ header?
```
atmgrifter00, line 179: A point of reference is that the AzDO table (at least the one for the backlog) retains what row should be focused when the table receives focus again. I would be fine with this behavior.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+4. Expand/collapse of rows uses `Alt-ArrowUp/ArrowDown` (instead of simply `ArrowLeft` and `ArrowRight`)
+
+    Cons:
+
+    - Reliance on modifier keys when there are available workflows/keys that do not require them (i.e. strictly using `ArrowLeft` and `ArrowRight` keys with no modifiers)
+    - Mistypes (attmpted to use `Alt` modifier but didn't) shifts focus unexpectedly.
+
+## Open Issues
+
+-   Row selection: Should pressing `Space` when a single cell is focused, result in selecting that row?
```
atmgrifter00, line 178: Perhaps it's best not to, given that `Space` will be responsible for interacting with elements when the table isn't in navigation mode (and an interactive element has focus). Note that the ARIA tree-grid docs specifically call out `Shift-Space` to perform row selection if focus is currently on a cell. I think this should satisfy this need.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Implementation / Design
+
+We have decided to use the ARIA `treegrid` role/ pattern for our table component (as opposed to the `grid` role).
+
+### ARIA Treegrid guidance
+
+Some noteworthy aspects of the treegrid pattern are:
+
+1. All cells are focusable, even those that have no interactive elements. The left and right arrow keys, while in navigation modes, move focus from cell to cell. In the case where a cell has a single interactive element, the element can be given the focus instead of the containing cell.
```
mollykreis, line 25: >In the case where a cell has a single interactive element, the element can be given the focus instead of the containing cell.

Can you clarify this statement? I'm not sure if you mean
(a) The cell contains only one interactive element (but possibly other elements that are not interactive)
or
(b) The cell contains only one element, and that element is interactive

Based on the excerpt from the ARIA guidelines, I'm guessing you mean (b), but when I first read the sentence, I thought you meant (a).

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+> -   A cell contains one widget whose operation does not require arrow keys and grid navigation keys set focus on that widget. Examples of such widgets include link, button, menubutton, toggle button, radio button (not radio group), switch, and checkbox.
+> -   A cell contains text or a single graphic and grid navigation keys set focus on the cell."
+
+### Specifics for nimble-table
+
+#### Focusing the table, `Tab`/`Shift-Tab` behavior
+
+`Tab` or `Shift-Tab`, from neighboring elements on the page (according to tab order), will focus the table:
+
+-   If the table header contains any interactive/focusable elements, the 1st will be focused. Examples: 'Select all rows' checkbox, 'Collapse all groups' button
```
mollykreis, line 49: Will a column header itself be focusable? For example, in order to toggle the sort order of the column.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+-   Otherwise, the 1st table row will be focused
+-   (If no header elements are focusable, and the table has no rows, the table cannot be focused)
+
+As noted above, `Tab` / `Shift-Tab` key moves focus between the focusable elements in a row or the header (not the cells). When focus has reached the end (or beginning) of the row/header, a continued `Tab` (or `Shift-Tab`) will move focus outside of the table.
+
+Optionally, if a particular table row has been focused, then tab focus leaves the table and later returns, we may return focus to that row (rather than it restarting at the first row/ header).  
+**Open question**: Do we want to adopt this behavior?
```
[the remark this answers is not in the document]
mollykreis: A few related questions:
- Would we return to the row based on ID or based on index?
- What would happen if the row doesn't exist anymore? (I assume we could just decide to fallback restarting focus back at the first row/header)
- Should we scroll the focused row back into view if it isn't currently in view?

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### Navigating rows/cells
+
+Once focused, by default the table is in 'navigation mode', allowing the user to focus table rows and cells via the arrow keys.
+
+When a row is focused:
```
mollykreis, line 62: Can you include a statement about an action menu's visibility in the "When a row is focused" and "When a cell is focused" sections?

mollykreis, line 62: How do you move between focused rows? ArrowUp and ArrowDown?

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+    - Behavior that doesn't match mouse interactions may cause confusion or frustration that interrupts users' goals/workflows.
+
+4. Expand/collapse of rows uses `Alt-ArrowUp/ArrowDown` (instead of simply `ArrowLeft` and `ArrowRight`)
+
+    Cons:
+
+    - Reliance on modifier keys when there are available workflows/keys that do not require them (i.e. strictly using `ArrowLeft` and `ArrowRight` keys with no modifiers)
+    - Mistypes (attmpted to use `Alt` modifier but didn't) shifts focus unexpectedly.
+
+## Open Issues
```
mollykreis, line 176: Are there any concerns about the way the virtualizer works with regards to keyboard interactions? Some things I can think of are:
- Switching between focused rows may need to scroll (and possibly load a new set of visible rows). Is there anything special we need to think about for that?
- What happens when a user scrolls the table while something in the table is focused, but the focused row/cell moves out of view so there is no longer an HTML element associated with the row that should be focused?

fredvisser (approved):

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: This probably needs some discussion (I can bring it up at the team meeting tomorrow).

I wasn't originally thinking of making the column headers focusable, but that would mean that there wouldn't be a keyboard-accessible way of sorting columns, since we deprioritized implementing the column header menus. So maybe we should.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Yes, ArrowUp / ArrowDown. (It's mentioned later in this section but as a separate floating sentence, maybe I should merge it into this part.)

Meyer's original prototype also supported PageDown/PageUp to move by a control-height's amount of rows at once, so I'll probably list that as an option too.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: I was going to say (b), but I'm not sure that it's a hard & fast rule. If there's multiple interactive elements in the cell, we definitely won't try to focus any of them (just the cell).

But, for example, if a cell had a button or checkbox, plus some extra accompanying text (not the control label), it would probably still make sense to just focus the button/checkbox.

The approach I described in the nimble-table section below is actually closer to (a) too.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+        -   `Esc` is pressed: If the user had edited content (e.g. textbox), it may be reverted.
+        -   `Enter` is pressed: If the user was editing content or interacting with a control, those changes will be committed.
+-   For all cells with action menu buttons:
+    -   `Ctrl-Enter` will trigger the action menu button and open the action menu. Note: If the action menu button is the only interactive element in the cell, `Enter` will do the same thing.
+-   Pressing `Tab` will focus the 1st/ primary interactive element in the cell. Subsequent `Tab` presses will focus subsequent interactive elements in the cell, including the action menu button. `Shift-Tab` will focus the previous interactive element. Note: `Tab` will not cycle within the cell (it will go to the next focusable elements in the row, or focus will leave the table, as outlined in the "`Tab`/`Shift-Tab` behavior" section above).
+
+If a cell is focused and `ArrowDown` or `ArrowUp` is pressed, when there is an available row in the navigation direction, focus will move to the cell in the current focused column of the new row.
```
jattasNI, line 93: Related to [Molly's comment in the open issues section](https://github.com/ni/nimble/pull/1960/files#r1541750707) we should explicitly mention the plan for scrolling, both vertically and horizontally, if the cell the user is moving to is not currently visible.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+-   **(Open question)** If a single cell is focused, should pressing `Space` select/unselect that row?
+
+Since a row's selection checkbox is currently `role=gridcell`, it will be able to be navigated to with `LeftArrow`/`RightArrow`.
+
+#### Grouping
+
+Since the entire group header row is a button, the explicit expand/collapse button in that row is not keyboard focusable.
+
+If a group header row is focused, pressing `Space` or `Enter` will expand or collapse that group.
+
```
jattasNI, line 112: It'd be good to explicitly call out interactions that are not planned to be achieveable via the keyboard. In addition to [column sorting that Molly already mentioned](https://github.com/ni/nimble/pull/1960/files#r1541626331), the only other one I can think of is column resizing. I'm personally ok with that being out of scope.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+    - Behavior that doesn't match mouse interactions may cause confusion or frustration that interrupts users' goals/workflows.
+
+4. Expand/collapse of rows uses `Alt-ArrowUp/ArrowDown` (instead of simply `ArrowLeft` and `ArrowRight`)
+
+    Cons:
+
+    - Reliance on modifier keys when there are available workflows/keys that do not require them (i.e. strictly using `ArrowLeft` and `ArrowRight` keys with no modifiers)
+    - Mistypes (attmpted to use `Alt` modifier but didn't) shifts focus unexpectedly.
+
+## Open Issues
```
jattasNI, line 176: Are there expectations for how a screen reader tells the user what column they are interacting with? Do any of our proposed interactions allow a screen reader user to know the contents of a column header? Or is this out of scope? If it is out of scope and not already working, we should file an issue for it since https://github.com/ni/nimble/issues/1151 is closed already.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,180 @@
...
+-   If a row's selection checkbox is focused, pressing `Space` will toggle the checkbox, selecting/unselecting that row
+-   **(Open question)** If a single cell is focused, should pressing `Space` select/unselect that row?
+
+Since a row's selection checkbox is currently `role=gridcell`, it will be able to be navigated to with `LeftArrow`/`RightArrow`.
+
+#### Grouping
+
+Since the entire group header row is a button, the explicit expand/collapse button in that row is not keyboard focusable.
+
+If a group header row is focused, pressing `Space` or `Enter` will expand or collapse that group.
```
jattasNI, line 111: What do you think about also allowing `ArrowLeft` to collapse and `ArrowRight` to expand group rows? That would feel more in line with the proposal for hierarchical parent rows. I can imagine a use case where a user wants to collapse several nearby rows with a series of left+downs in quick sequence (in fact I've done this in other apps) but that would be interrupted if some of the rows are groups.

I can't think of any conflicts that this would introduce with other gestures you describe in this document, but obviously if it does then we should drop this idea.

jattasNI, line 111: If `selection-mode="multiple"` and a group row is focused I think there's a gesture conflict: will `Space` expand/collapse or will it select/deselect?

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,205 @@
...
+
+As noted above, `Tab` / `Shift-Tab` key moves focus between the focusable elements in a row or the header (not the cells). When focus has reached the end (or beginning) of the row/header, a continued `Tab` (or `Shift-Tab`) will move focus outside of the table.
+
+**Losing and regaining focus**:
+
+-   If a particular table row has been focused, then tab focus leaves the table and later returns, we will return focus to that row (based on row index), rather than it restarting at the first row/ header.
+-   If the previously focused row index is not currently in view, we'll scroll so that it's in view.
+-   If there is no longer a row at the previous row index (e.g. data updates resulted in fewer rows), we'll return focus to the first row (not the header).
+
+#### Navigating rows/cells
```
msmithNI, line 61: I noticed that the [ARIA treegrid keyboard interaction docs](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/#keyboardinteraction) lists additional keys not yet covered by this HLD: PageUp, PageDown, Home, End, Ctrl-Home, Ctrl-End (to navigate by larger #s of rows at once). I plan to add those in a subsequent HLD update.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Updated

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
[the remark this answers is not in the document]
msmithNI: HLD updated:
- Current plan is to return to the row based on index, not ID. I think that makes the most sense since keyboard navigation is basically navigating by row index also.
- If the given row index no longer exists after an update, specified that we'll restart focus at the 1st row (not header)
- The focused row will be scrolled to, if needed

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Removed from open issues

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Removed from open issues. Kept the wording that we'll return focus to the previously focused row.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Removed from open issues; Space with a focused cell won't result in row selecting.

Added a line about `Shift-Space` being an ARIA suggestion for selection in that case.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Updated HLD wording (closer to your (a) option)

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: As discussed at team meeting, added a section for column sorting.

> #### Sorting
> If at least one column can be sorted, all of the column headers will be keyboard focusable (the [ARIA grid examples](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/#examples) illustrate this idea), to allow for column sorting via the keyboard.
> In this scenario:
> -   `ArrowLeft` and `ArrowRight` will navigate between the selection checkbox and expand/collapse button, and each of the column headers, following similar semantics as the cell navigation (above).
> -   `Enter` will toggle column sorting (between ascending/descending/none, in the same way a mouse click does)
> -   `Shift-Enter` will add/remove the given column from sorting, while maintaining any other sorted columns (e.g. multi-sort), in the same way a Shift-mouse-click does

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Done

change/@ni-nimble-components-77b662a8-c850-49f0-aa24-90a0ba7c696a.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
msmithNI: Done

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Added a section stating column sizing via keyboard is out of scope.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: As the column headers already have `role=columnheader`, I'm not aware of any additional work that we should do regarding this. They'll also be directly focusable if any column can be sorted.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Seems reasonable. Added.

msmithNI: As expand/collapse is the primary action for group rows, I think that should win out for what `Space` does. Clarified that in the HLD.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Added a section about scrolling the table (generally anytime we focus/ refocus a row/cell, we'll scroll to it if needed).

I've mostly been considering vertical scrolling, not horizontal. I'll keep this conversation open for now to check if any further clarification is needed for horizontal scrolling.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Added a section about scrolling the table (generally anytime we focus/ refocus a row/cell, we'll scroll to it if needed).

I don't want to veer into implementation details too much for this HLD, but due to the virtualization, we'll probably need to keep track of the focused row/cell by index, and re-focus them as needed if they go out of view and become visible again. Meyer's earlier prototype had some of this working, so I think it's definitely non-trivial, but doable.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Moved the info about ArrowUp/Down into the sections about row/ cell navigation.

(I have a separate open thread about adding PageUp/Down to the HLD, so resolving this one)

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Removed this wording / removed from Open Issues

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,206 @@
...
+> -   When a cell contains _one widget_ whose operation _does not require arrow keys_, the grid navigation keys **will set focus on that widget**. Examples of such widgets include link, button, menubutton, toggle button, radio button (not radio group), switch, and checkbox.
+> -   When a cell contains _text or a single graphic_, the grid navigation keys will **set focus on the cell**."
+
+### Specifics for nimble-table
+
+#### Focusing the table, `Tab`/`Shift-Tab` behavior
+
+`Tab` or `Shift-Tab`, from neighboring elements on the page (according to tab order), will focus the table:
+
+-   If the table header contains any interactive/focusable elements, the 1st will be focused. Examples: 'Select all rows' checkbox, 'Collapse all groups' button
```
jattasNI, line 49:
```suggestion
-   If the table header contains any interactive/focusable elements, the 1st will be focused. Examples: 'Select all rows' checkbox, 'Collapse all groups' button, header of a sortable column
```

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,206 @@
+# Table Keyboard Navigation HLD
+
+## Problem Statement
+
+It is desirable for the table to support navigating and interacting with its contents with the keyboard alone. ARIA provides guidance on how to best accomplish this.
+
+## Links To Relevant Work Items and Reference Material
```
jattasNI, line 7: Do we have visual designs for a row being focused and a cell being focused? If so it'd be good to link to them here. If not we should put it on Brandon's radar.

jattasNI (approved):

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Resolving because I think the current wording covers horizontal scrolling too. Vertical scrolling will be via virtualizer APIs, horizontal scrolling will be via something like `Element.scrollIntoView()`.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: As mentioned on Teams, I can't load any of the old Adobe XD documents anymore (I always get a password error), so I'm not sure if there's any older designs for it. For prototyping I've been using the standard green border color. We can discuss it on Monday probably.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Updated

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Updated

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
mollykreis: I'm still a bit confused by this. The ARIA guidelines seem to be saying that the cell needs to be focusable in order for the non-focusable text to be read by a screen reader. However, the plan I'm reading in your statement is that neither the cell nor the non-focusable text will be focusable if the cell contains only one interactive element. Those statements seem in conflict with each other. Am I missing some aspect of what you are proposing here?

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
```diff
@@ -0,0 +1,215 @@
...
+-   `PageUp`: Navigate up by a table height's worth of rows. If a cell was selected, the matching cell (in the same column) of the new row will be selected.
+-   `PageDown`: Navigate down by a table height's worth of rows. If a cell was selected, the matching cell (in the same column) of the new row will be selected.
+-   `Home`: Navigate to the 1st row (if a row is selected), or the 1st cell of the row (if a cell is selected)
+-   `End`: Navigate to the last row (if a row is selected), or the last cell of the row (if a cell is selected)
+-   `Ctrl-Home`: Navigate to the first row. If a cell was selected, the matching cell (in the same column) of the new row will be selected.
+-   `Ctrl-End`: Navigate to the last row. If a cell was selected, the matching cell (in the same column) of the new row will be selected.
+
+Special case of focusing a cell:
+
+-   If a cell does not have an action menu, and only contains a single interactive/focusable element of the following types: button (and variants), switch, checkbox, or anchor/link: that child element will be focused _instead_ of the cell.
```
mollykreis, line 95: Whose job will it be to decide what needs to be focused (i.e. the cell or a specific element) and then focus it?

Note: I'm not looking to get too far into the implementation details here. I'm mostly wondering if this is something the table/row/cell can handle, or if we are needing to introduce a new API for the table column and/or cell view.

rajsite (approved):

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: We might need a new API for the `TableCellView`, similarly to where we ended up with the `focusedRecycleCallback()`. I still need to do some prototyping in this area to get a better idea of what the implementation should be like.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: I think our exact behavior here might depend on the implementation approach we go with. I'm also not aware of any specific examples of this (single interactive element plus extra content in a cell) for column types we currently know about.

- If we are just querying for specific element types in a cell (a single `nimble-anchor`/`nimble-button`/etc), I think we'd probably just go ahead and focus it if we find just one. I don't think we'd try to look for additional text in the cell and do something different if we find it.
- If we do have a new cell view API to handle the focusing behavior, I guess we'd just want some good documentation/guidance on when an element in the cell should get focused.

I did try out the NVDA screen reader some, and your reading of the ARIA guidelines seems accurate. If we focus an element inside a cell, that element (and it's label) will be read out loud, but any additional content in the cell not associated with that element will not be read. However there are some cases where it would be read (like the interactive element having an `aria-describedby` attribute that points to the other content).

So it's all a bit complicated, but I'm not sure we need to fully commit one way or another here (unless we expect this to come up for a known column type soon).

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: Current plan is to demo the current prototype to Brandon at one of the next team meetings (which uses a 2px border plus our standard border color around the active row/cell), and see if he has feedback.

packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md
msmithNI: If there ends up being API impact to the table/ cell, we can update the HLD with those details in a followup PR.

PR #1963, opened 2024-03-27
Fix table scroll tests on WebKit

### LINKED ISSUE #1942: Some table tests fail in WebKit and/or Safari due to focusedRecycleCallback
# 🧹 Tech Debt

These three tests fail in macOS Safari. The first one also fails in the WebKit Playwright build and has been disabled.

- Table > with connection > uses virtualization > and calls focusedRecycleCallback on focused cell views when a scroll happens
- Table > with connection > uses virtualization > and allows viewing the last rows in the data after scrolling to the bottom
- Table > with connection > uses virtualization > to render fewer rows (based on viewport size)

`Expected spy focusedRecycleCallback to have been called once. It was called 0 times.`

WebKit 605.1.15 (Mac OS 10.15.7)

We should either fix and re-enable them or file an issue upstream and mark this blocked.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1942 

## 👩‍💻 Implementation

WebKit doesn't appear to fire the `scroll` event when a scroll done by setting `scrollTop` on an element. Therefore, I've updated the page object to scroll using the `scroll()` function, which does cause the event to be fired. As a result, the table's virtualizer correctly detects the change and the tests pass.

## 🧪 Testing

Ran all the table tests in webkit locally and on the pipeline to verify they pass.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-e3b02c33-dd73-4a99-824b-cb0b19f706bb.json
+++ b/change/@ni-nimble-components-e3b02c33-dd73-4a99-824b-cb0b19f706bb.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix table scrolling test on WebKit",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


mollykreis:

msmithNI (approved):

rajsite (approved):

PR #1966, opened 2024-03-27
Angular column placeholder

mollykreis:
# Pull Request

## 🤨 Rationale

Adds Angular support for column placeholders, which is part of #1538.

## 👩‍💻 Implementation

Add `placeholder` to the 5 columns that support that property.

## 🧪 Testing

Added unit tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-angular-a238e91b-0109-4e07-b89a-aba8b2d19387.json
+++ b/change/@ni-nimble-angular-a238e91b-0109-4e07-b89a-aba8b2d19387.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add support for column placeholders",
+  "packageName": "@ni/nimble-angular",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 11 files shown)


mollykreis:

atmgrifter00 (approved):

atmgrifter00:
Should we add an example of the usage of this in the example project?

atmgrifter00 (commented):
Should we add an example of the usage of this in the example project?

rajsite (approved):

rajsite:
> Should we add an example of the usage of this in the example project?

I don't feel strongly. Wouldn't necessarily do it for every column, maybe 1-2 if done. Whichever way just a reminder to keep it consistent across with blazor

change/@ni-nimble-angular-a238e91b-0109-4e07-b89a-aba8b2d19387.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: Moving comment to a discussion:

>> Should we add an example of the usage of this in the example project?
>
> I don't feel strongly. Wouldn't necessarily do it for every column, maybe 1-2 if done. Whichever way just a reminder to keep it consistent across with blazor

mollykreis: If no one feels strongly, I lean towards not putting it in the example apps because I don't think it adds much value. There isn't anything framework-specific about this property, which is already exposed on all relevant storybook stories.

@atmgrifter00, was your question because you want this feature exposed in the example apps or because you just wanted to make sure we had considered it?

atmgrifter00: Just that we had considered it. Don't feel strongly either.

PR #1970, opened 2024-03-27
Consistent storybook matrix states refactoring

rajsite:
# Pull Request

## 🤨 Rationale

Makes the following bits more consistent in the storybook matrix tests:
- All the states enums use the normal enum pattern ([instead of programmatic enum definition which looses some typing](https://github.com/ni/nimble/pull/1934#discussion_r1534888401))
- All the states are defined first before the storybook metadata definition
- All the states use the naming convention of `blahStates` for the state dimension and `BlahState` for the individual state type
- File specific states were put in alphabetical order

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Should not result in any changes in stories except for one change in dialogs where I have the state name render on the page instead of the state value which is a long CSS string.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Hoping folks just copy and paste the patterns, didn't seem worth a doc update.

### DIFF
--- a/change/@ni-nimble-components-797c6409-b365-4d44-a205-575326a62af3.json
+++ b/change/@ni-nimble-components-797c6409-b365-4d44-a205-575326a62af3.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Matrix stories consistent strict enum pattern and organize states",
+  "packageName": "@ni/nimble-components",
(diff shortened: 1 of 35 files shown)


rajsite:
@jattasNI think the change is pretty non-controversial and minor. Bypassing owners but can address concerns in a follow-up.

PR #1972, opened 2024-03-28
Fix a couple issues with anchor styling

m-akinc:
# Pull Request

## 🤨 Rationale

I noticed a few differences between our anchor styling and the [Figma design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-76706&mode=design&t=tRlgHtHiOWeXNCIe-0):
1. Visited links have a distinct color (purple)
2. Keyboard focus is indicated by an outline rather than double underline
3. Disabled links should have underlines unless `underline-hidden` is set
4. Links with `underline-hidden` should not have underlines when clicked/active (despite mouse hover)
5. In the color theme, a clicked/active link should have the same, dim, gray color regardless of whether it has the `prominent` appearance. (Figma does not actually provide a design for the `prominent` appearance under the color theme, but we can extrapolate from the other themes.)

## 👩‍💻 Implementation

I am addressing issues 3-5. Tech debt issue #2011 tracks items 1 and 2.

To simplify things, I rewrote the anchor styles using CSS cascade layers.

## 🧪 Testing

Storybook snapshots

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-c92408f2-14ec-4254-9aac-0babfa7fbbd1.json
+++ b/change/@ni-nimble-components-c92408f2-14ec-4254-9aac-0babfa7fbbd1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix a couple issues with anchor styling",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchor/styles.ts
+++ b/packages/nimble-components/src/anchor/styles.ts
@@ -76,11 +76,6 @@ export const styles = css`
             color: ${linkActiveProminentFontColor};
         }
 
-        ${
-            '' /* It's probably cleaner to add :not(active) to the selector in the 'active' layer (rather
-                than resetting text-decoration back to the value from the 'default' layer), but
-                there's a bug in storybook-addon-pseudo-states where it doesn't handle :not() properly. */
-        }
         :host([underline-hidden]) .control:active {
             text-decoration: none;
         }
--- a/packages/nimble-components/src/anchor/styles.ts
+++ b/packages/nimble-components/src/anchor/styles.ts
@@ -2,102 +2,93 @@ import { css } from '@microsoft/fast-element';
 import { display } from '@microsoft/fast-foundation';
 import { focusVisible } from '../utilities/style/focus';
 import {
-    linkActiveDisabledFontColor,
     linkActiveFontColor,
     linkActiveProminentFontColor,
     linkDisabledFontColor,
     linkFont,
     linkFontColor,
-    linkProminentDisabledFontColor,
     linkProminentFontColor
 } from '../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('inline')}
+    @layer base, hover, focusVisible, active, disabled;
 
-    :host {
-        box-sizing: border-box;
-        font: ${linkFont};
+    @layer base {
+        ${display('inline')}
+
+        :host {
+            box-sizing: border-box;
+            font: ${linkFont};
+        }
+
+        .top-container {
+            display: contents;
+        }
+
+        [part='start'] {
+            display: none;
+        }
+
+        .control {
+            color: ${linkFontColor};
+            text-decoration: underline;
+        }
+
+        :host([underline-hidden]) .control {
+            text-decoration: none;
+        }
+
+        :host([appearance='prominent']) .control {
+            color: ${linkProminentFontColor};
+        }
+
+        .content {
+            pointer-events: none;
+        }
+
+        [part='end'] {
+            display: none;
+        }
     }
 
-    .top-container {
-        display: contents;
+    @layer hover {
+        .control:any-link:hover {
+            text-decoration: underline;
+        }
     }
 
-    [part='start'] {
-        display: none;
+    @layer focusVisible {
+        .control${focusVisible} {
+            outline: none;
+            box-shadow: inset 0px -1px;
+            text-decoration: underline;
+        }
     }
 
-    .control {
-        color: ${linkFontColor};
-        text-decoration: underline;
+    @layer active {
+        .control:active {
+            color: ${linkActiveFontColor};
+            text-decoration: underline;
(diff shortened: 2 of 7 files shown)


m-akinc:
@mollykreis mind buddying this for me?

packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -588,7 +588,7 @@ export const [
         element,
         DigitalGreenLight,
         DigitalGreenLight,
-        PowerGreen
+        hexToRgbaCssColor(White, 0.6)
```
mollykreis, line 591: When I look at the anchor in storybook for your PR, the prominent link on the color theme still looks green. Am I missing something?

packages/nimble-components/src/anchor/styles.ts
```diff
@@ -2,102 +2,93 @@ import { css } from '@microsoft/fast-element';
...
-        color: ${linkProminentDisabledFontColor};
-    }
+        :host([appearance='prominent']) .control:active {
+            color: ${linkActiveProminentFontColor};
+        }
 
-    .content {
-        pointer-events: none;
+        ${
+            '' /* It's probably cleaner to add :not(active) to the selector in the 'active' layer (rather
```
mollykreis, line 80: I don't think this comment is necessary. It is against nimble's guidelines to use the `:not` selector. Even if there wasn't a bug in the pseudo-states addon, I don't expect that we'd want to write the styles using `:not(active)`.

m-akinc: If it worked, I would have argued for using `:not` here, since it more directly represents what we want, but that's moot. I'll just remove the comment.

packages/nimble-components/src/theme-provider/design-tokens.ts
m-akinc: I'm only changing `linkActiveProminentFontColor`, so you'll see the difference only when clicking down on the link.

mollykreis (approved):

jattasNI (approved):

packages/nimble-components/src/anchor/styles.ts
```diff
@@ -2,102 +2,88 @@ import { css } from '@microsoft/fast-element';
...
     linkDisabledFontColor,
     linkFont,
     linkFontColor,
-    linkProminentDisabledFontColor,
     linkProminentFontColor
 } from '../theme-provider/design-tokens';
 
 export const styles = css`
-    ${display('inline')}
+    @layer base, hover, focusVisible, active, disabled;
```
rajsite, line 14: talked offline, two things to doublecheck:
- Are @layer definitions scoped to shadow root? Concerns are other `@` definitions like `@font-face` aren't
- Do we need to align styles with other text-based links that are not sharing the styles directly. i.e. rich-text-editor anchor styling and breadcrumb

m-akinc: > Are @layer definitions scoped to shadow root? Concerns are other `@` definitions like `@font-face` aren't

image

> Do we need to align styles with other text-based links that are not sharing the styles directly. i.e. rich-text-editor anchor styling and breadcrumb

- The breadcrumb item needed a change to hide the underline when clicking.
- The rich-text editor does not have interactive links and does not use `underline-hidden`, so no changes apply.
- The rich-text viewer had intentionally set the link color to be constant, regardless of interactions, but this seems wrong. The stated intent was "To ensure a consistent appearance, the font color is forced to the default link color regardless of the `href` attribute's presence." An inoperable/disabled link (i.e. no `href`) should render like other disabled links. By removing this bit of styling, we also get the standard interaction colors.

rajsite: I think what is currently there is intentional (and incomplete). The intention is that things that are not a supported link format are treated as plaintext. 

i.e. visually look like plaintext:
image

The proposed change makes it look like a disabled link:
image

For reasons[1] a link is emitted for this case instead of a plaintext span and it's partially styled to visually appear like plaintext. Maybe a comment update is needed to clarify the intention but I don't think the style change proposed in the PR gets closer to the goal.

[1] I believe the reason is that links can be either normal links or mentions. They are both represented in markdown as links. The way you differentiate is (partially) by the scheme. http and https are always normal links or mentions. Other schemes cannot be normal links but may be mentions, however mentions can be lazily / dynamically defined. So we can't throw away any "unsupported links" because what is supported changes dynamically. A previously unsupported URL scheme may become a supported mention scheme.

m-akinc: Reverted my change to rich-text viewer link styling.

rajsite: > Maybe a comment update is needed to clarify the intention

Didn't see a response to this statement so re-opening. Did you find the existing comment clear enough (so that during refactors and updates it's not lost)?

m-akinc: Sorry, I didn't read that closely enough. I realized that just reverting my change to the rich-text viewer styling isn't sufficient, because a non-href link still ends up with an underline. I tried to change the selector to affect only the non-href case (by adding `:not(:any-link)`) so I could then also remove the underline, but that selector doesn't seem to work on Chromium. Seems like this is an existing [bug](https://issues.chromium.org/issues/40623497). Instead, I modified the markdown parser to emit the `underline-hidden` attribute on a link that doesn't have an `href`. I modified the existing tests to verify this.

I have also updated the comment to be clearer about why we're styling links to look like plain text.

rajsite (approved):

change/@ni-nimble-components-c92408f2-14ec-4254-9aac-0babfa7fbbd1.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: > I am addressing issues 3-5. I will check with Brandon whether 1 and 2 are really the designs we want.

Is the result of that captured anywhere / in tech debt or features?

m-akinc: Created #2011 and updated description

Issue #1948, opened 2024-03-20
Align combobox implementation behavior with select.


atmgrifter00:
# 🧹 Tech Debt

The [initial attempt](https://github.com/ni/nimble/pull/1912/) to align the `Combobox` implementation with the `Select` resulted in a [bug in Angular](https://ni.visualstudio.com/DevCentral/_workitems/edit/2687002), where the options would ultimately remain hidden upon a reload of the DOM after they had been filtered out. The cause of this seems rooted in the new `registerOption` implementation, but it's unclear at the moment exactly why this is resulting in the new behavior.

Ultimately, as part of re-aligning the `Combobox` and `Select` implementations, the `Combobox` should adjust its implementation to no longer manipulate the `hidden` attribute directly while filtering, and instead leverage the now available `visuallyHidden` attribute of the `ListOption`.

atmgrifter00:
Ideally, as part of the work to re-align the `Combobox` with the `Select` we would also provide a test case that the linked AzDO bug above is representative of. This may be a bit tricky because it involves a re-loading of an Angular component that was hosting the `Combobox`.

jattasNI:
@atmgrifter00 we are trying to figure out how to prioritize this tech debt but I think it contains several issues that might have differing priorities. Could you take a pass at listing those and bucketing them into separate issues if we think we should fix them separately?

Brainstormed possible issues with Milan:
- registerOption
- visuallyHidden to track filtered items
- diacritic insensitive search

jattasNI:
Additional tech and/or architecture debt that might be covered by this:

1. coupled responsibilities of FAST foundation Listbox base class and Nimble concrete derived classes
2. better isolation of related component state. For example, should we have a separate keyboard manager and selection manager like we do for other components?
3. ensure synchronous state management to match the native select behavior (or be very clear about when we diverge from this). See #1915.

Re-adding the triage tag to trigger a discussion about whether these should be separate issues and what their priority is.

jattasNI:
@atmgrifter00 and the team agreed that he'll work on these issues this cycle and come up with a recommendation of how to split up the work once he's done some initial research. This will be be prioritized after the first 1 or 2 SLE combobox -> select migrations.

rajsite:
Also need to consider dynamic options equivalents for combobox (i.e. combobox analog to filter-mode manual)

jattasNI:
[...]

PR #1973, opened 2024-03-28
Fork FAST Combobox.

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1948 

As part of the effort to bring the `Combobox` implementation in alignment with the `Select`, namely with how we handle controlling the visibility of the options in the dropdown, as well as the need to register options as part of the `connectedCallback` hook, it makes sense to fork the FAST `Combobox` implementation into our own, as we did with the `Select` to make it more clear what is happening within our own implementation for the various new behaviors we require.

## 👩‍💻 Implementation

Essentially bringing over the current `Combobox` implementation from the FAST `archives\fast-element1` branch into our own. Copied the FAST `FormAssociatedCombobox` that our `Combobox` now extends.

## 🧪 Testing

Ran existing unit tests. Manual testing.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/combobox/index.ts
+++ b/packages/nimble-components/src/combobox/index.ts
@@ -1,14 +1,30 @@
-import { attr, html, observable, ref } from '@microsoft/fast-element';
+import {
+    DOM,
+    Observable,
+    attr,
+    html,
+    observable,
+    ref
+} from '@microsoft/fast-element';
 import {
     DesignSystem,
-    Combobox as FoundationCombobox,
-    ComboboxOptions
+    ComboboxOptions,
+    ComboboxAutocomplete,
+    SelectPosition,
+    ListboxOption,
+    DelegatesARIACombobox,
+    applyMixins,
+    StartEnd
 } from '@microsoft/fast-foundation';
 import {
     keyArrowDown,
     keyArrowUp,
     keyEnter,
-    keySpace
+    keyEscape,
+    keySpace,
+    keyTab,
+    limit,
+    uniqueId
 } from '@microsoft/fast-web-utilities';
 import { ToggleButton, toggleButtonTag } from '../toggle-button';
 import { errorTextTemplate } from '../patterns/error/template';
@@ -21,6 +37,8 @@ import type { DropdownPattern } from '../patterns/dropdown/types';
 import { DropdownAppearance } from '../patterns/dropdown/types';
 import type { AnchoredRegion } from '../anchored-region';
 import { template } from './template';
+import { FormAssociatedCombobox } from './models/combobox-form-associated';
+import type { ListOption } from '../list-option';
 
 declare global {
     interface HTMLElementTagNameMap {
@@ -32,18 +50,10 @@ declare global {
  * A nimble-styed HTML combobox
  */
 export class Combobox
-    extends FoundationCombobox
+    extends FormAssociatedCombobox
     implements DropdownPattern, ErrorPattern {
     @attr
     public appearance: DropdownAppearance = DropdownAppearance.underline;
-
-    /**
-     * The ref to the internal dropdown button element.
-     *
-     * @internal
-     */
-    @observable
-    public readonly dropdownButton?: ToggleButton;
 
     /**
      * A message explaining why the value is invalid.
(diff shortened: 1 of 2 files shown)


jattasNI (approved):

change/@ni-nimble-components-52555d4d-6a94-4763-9740-d90238ee16af.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: If you haven't already, could you double check that our unit tests are up to date with the FAST ones? Then call this out in the Testing section of the PR description.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -32,32 +50,52 @@ declare global {
...
+     * The open attribute.
+     */
+    @attr({ attribute: 'open', mode: 'boolean' })
+    public open = false;
+
+    /**
+     * Sets the placeholder value of the element, generally used to provide a hint to the user.
+     */
+    @attr
+    public placeholder = '';
```
rajsite, line 89: Can we avoid setting the placeholder to empty string? All `@att` marked properties will end up being reflected to the attribute which isn't what we want here. i.e. we end up with `<nimble-combobox placeholder="">`.

This is a [change in behavior from fast](https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/combobox/combobox.ts#L199)

```suggestion
    public placeholder?: string;
```

packages/nimble-components/src/combobox/index.ts
```diff
@@ -68,72 +106,234 @@ export class Combobox
...
+    /**
+     * @internal
+     */
+    @observable
+    public readonly dropdownButton?: ToggleButton;
+
+    /**
+     * The collection of currently filtered options.
+     */
+    public filteredOptions: ListboxOption[] = [];
```
rajsite, line 132: internal?

packages/nimble-components/src/combobox/index.ts
```diff
@@ -68,72 +106,234 @@ export class Combobox
...
+    private filter = '';
+
+    /**
+     * The initial state of the position attribute.
+     */
+    private forcedPosition = false;
 
-    // Workaround for https://github.com/microsoft/fast/issues/5123
-    public override setPositioning(): void {
+    public setPositioning(): void {
```
rajsite, line 218: this should be internal right? weird that it's not in fast..

packages/nimble-components/src/combobox/index.ts
```diff
@@ -148,78 +348,352 @@ export class Combobox
         }
     }
 
-    public override filterOptions(): void {
-        super.filterOptions();
+    public filterOptions(): void {
```
rajsite, line 351: internal? weird that it's public in fast

packages/nimble-components/src/combobox/index.ts
```diff
@@ -148,78 +348,352 @@ export class Combobox
...
+                -1,
+                this.options.length - 1,
+                next
+            );
+
+            // we only want to call the super method when the selectedIndex is in range
+            if (pinnedSelectedIndex !== this.selectedIndex) {
+                this.selectedIndex = pinnedSelectedIndex;
+                return;
+            }
```
rajsite, line 593: May certainly be missing it but is this new logic (don't think I see this bit in the original change)? Do we have test coverage for it?

rajsite (approved):

packages/nimble-components/src/combobox/index.ts
atmgrifter00: Yeah I only didn't mark it that way, because it wasn't originally in FAST. I'll change.

packages/nimble-components/src/combobox/index.ts
atmgrifter00: This logic is a slight update to the [original](https://github.com/microsoft/fast/blob/0c27d027ff6e8616ad4fddc17f4432aa7f6cbad0/packages/web-components/fast-foundation/src/combobox/combobox.ts#L547). I had to make the change that I did to avoid the Typescript issue of trying to modify a method parameter.

PR #1980, opened 2024-03-28
Disable ts deprecation warnings

rajsite:
# Pull Request

## 🤨 Rationale

Disables TypeScript deprecation warnings. Currently we get this warning in vscode all the time:
image

Which states that a TS feature we have enabled (and will be replaced by a pretty much compatible feature) will be removed in TS 5.5. We are pretty far from that because TS 5.5 is not even released yet (let alone up taken in a version of Angular we support).

We can disable the warning now as the ability to disable warnings was added in TS 4.9.5 which we recently updated to in #1921 

## 👩‍💻 Implementation

Disable TS deprecation warnings with [ignoreDeprecations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#deprecations-and-default-changes).

## 🧪 Testing

Disabled the warning and verified the vscode message is gone. Rely on CI for the rest.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f36f5e68-2f6a-4e85-bb86-9fcf7ad37a8d.json
+++ b/change/@ni-nimble-components-f36f5e68-2f6a-4e85-bb86-9fcf7ad37a8d.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Disable deprecation warnings",
+  "packageName": "@ni/nimble-components",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "none"
+}


packages/nimble-components/tsconfig.json
```diff
@@ -19,6 +19,7 @@
         // Require type only imports where possible
         // so imports with side-effects are explicit
         "importsNotUsedAsValues": "error",
```
jattasNI, line 21: Do you know if there's a reason we set `"importsNotUsedAsValues": "error"` in only two of our packages? Needn't be part of this PR, but should we consider enabling it everywhere?

jattasNI (approved):

packages/nimble-components/tsconfig.json
rajsite: It's most important in our tsconfigs that have imports with side-effects, i.e. anything that imports directly from nimble-components using the side-effect syntax, i.e. `import '@ni/nimble-components/dist/esm/button';`. So site and performance packages should probably have it to be safe.

Doesn't hurt to be consistent, I'll look at consistency across our tsconfigs. May be worth an internal private package to share that tsconfig inside nimble, could eventually lead to re-opening: https://github.com/ni/javascript-styleguide/pull/53

Think I'll put it on my personal list to play with. But if worth tracking as tecch debt lemme know and I can put that together.

PR #1986, opened 2024-03-29
Create `spright-components` package

### LINKED ISSUE #1974: Create `spright-components` package
# 🙋 Feature Request

## 😯 Problem to Solve

Part of setting up Spright.

## 💁 Proposed Solution

- Create `spright-components` package
  - Include single `spright-accordion` demo component
  - Build should build and publish it to npm
  - No Storybook initially (later PR)

m-akinc:
# Pull Request

## 🤨 Rationale

Resolves #1974

## 👩‍💻 Implementation

- Created `packages/spright-components`
  - has `spright-accordion` component with unit tests
  - `README` and `CONTRIBUTING`

Configuration includes some Storybook-related content (just to avoid having to remove and re-add later), but the actual Storybook support for `spright-components` will be a separate submission.

## 🧪 Testing

Builds and tests pass

## ✅ Checklist


packages/.eslintrc.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
+    root: true,
+    extends: '@ni/eslint-config-javascript',
```
m-akinc, line 3: Why are we not extending `eslint-config-typescript`?

packages/spright-components/.prettierrc.json
m-akinc, line 1: Seems like it would be nice to share this between `nimble-components` and `spright-components`, but the [recommended way to do that](https://prettier.io/docs/en/configuration.html#sharing-configurations) is by making an npm package for it. Not sure that's worth it.

m-akinc:
@jattasNI Not sure it makes sense for someone else to buddy this, so I thought you could take an initial look before having other people review.

packages/spright-components/src/all-components.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
jattasNI, line 1: At some point we'll need to figure out how the all-components approach will work if someone wants all of spright and all of nimble. Can I import both all-components files or will that result in errors from nimble elements being defined twice?

That doesn't necessarily need to be part of this PR.

packages/spright-components/src/accordion/index.ts
```diff
@@ -0,0 +1,27 @@
+import {
```
jattasNI, line 1: I'm slightly on the fence about including the accordion example. It hasn't gone through all the required steps to be in spright but because it's a real useful component I could see people starting to use it. 

If we need a component to prove our infrastructure is working I wonder if we should do something to make it obvious that this is a placeholder. Ideas:
1. give it an obvious flaw like incorrect styling (hot pink?)
2. give it a name that conveys its nature ("temporary-accordion"?)
3. replace it with a component that's obviously useless ("spright-rectangle").

packages/spright-components/.prettierrc.json
jattasNI: As we discussed offline, we think it will be useful to create some npm packages to store config that we want to share across packages in the monorepo. This will maintain our architectural philosophy of each package expressing its dependencies in its own package.json and not implicitly depending on anything outside its folder, making it easier to move packages outside the monorepo if desired.

High level attributes of these packages:
- should be small and scoped to a single purpose (e.g. lint config, karma config), not one big config package
- should follow ecosystem patterns (e.g. expose rulesets as eslint plugins, not as a file with an eslintrc)
- should be [marked private](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private) to start with

I don't have a good immediate sense about which packages that would result in. Candidates include eslint config, karma config, typescript config, prettier config, rollup config, and probably more. Do you want to take a stab at deciding or sync quickly with Milan to come up with a clear philosophy?

m-akinc: I figured the config refactoring should be submitted separately, so I created #2020. That covers the eslint config. I will look into the others. When all of that is in, we can return to this PR.

packages/spright-components/src/all-components.ts
m-akinc: I assume we're going to need an `all-components.js` in Spright for the sake of `SprightBlazor`, otherwise, I doubt we need to support that use case for clients. It's not likely we'll have tons of Spright components, nor that any given client app would want to import all of them.

packages/spright-components/.prettierignore
```diff
@@ -0,0 +1,29 @@
+# Visit and ignore all objects in the root. 
+# Then explicitly allow all TypeScript, JavaScript, Markdown, and JSON files in the root, src, and .storybook folders.
+/*
+!*.ts
+!*.js
+!*.md
+!*.mdx
+!*.json
+!src/
+!/.storybook/
```
m-akinc, line 10: Even though this PR doesn't include a Storybook, I figured it wouldn't be worth the bother to remove all the Storybook-related configuration just to add it back later.

change/@ni-spright-components-08e968ac-291f-4c66-949e-115ca77deef2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
jattasNI, line 2: Unless beachball behaves differently the first time it runs on a package, I suspect this will prevent it from publishing. Don't we want to publish it? Or are you holding off for some future milestone?

I also think we want to establish the changelog files since those need to be checked into source.

change/@ni-spright-components-08e968ac-291f-4c66-949e-115ca77deef2.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Create initial spright-components package",
+  "packageName": "@ni/spright-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
```
jattasNI, line 7: nit: I got thrown off by this statement in the PR description.
>Storybook for spright-components will be a separate submission.

There's a bunch of storybook stuff in this PR. No point in deleting it just to re-add it in a future PR, but could you make the PR description more accurately reflect the infrastructure that's being added?

packages/spright-components/CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
+# Contributing to Spright Components
+
+This package generally uses the same tooling and policies as Nimble components. See the [`nimble-components` CONTRIBUTING doc](../nimble-components/CONTRIBUTING.md) for an overview. For details on whether a component belongs in Nimble or Spright, refer to the [README](./README.md). This document describes ways in which Spright development differs from Nimble.
+
+The following table compares the requirements for components developed in Spright vs Nimble:
+
+|                      | Approved Spec | Unit Tests | SB\* Visual Tests | SB\* API Docs | SB\* Usage Docs | Approved\*\* VxD | Approved\*\* IxD | Angular/Blazor Support | Proper a11y | Minimal Tech Debt | Mobile Support |
```
jattasNI, line 7: nit: could you transpose this table so we don't have to scroll horizontally?

packages/spright-components/CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+🟢 = required\
+🟡 = optional\
+\*SB = Storybook\
+\*\*By an interaction and/or visual designer
+
+## Code ownership
+
+Code in Spright packages will be co-owned and reviewed by the Design System team and the contributing team.
+
+Design System team members remain as co-owners to enforce technical quality of implementation. They can (and will) give consulting on feature selection (api design for properties/attributes/methods/events, level of accessibility support for aria/keyboard/mouse, interaction design, visual design), while ultimately deferring to the Spright contributor for what to implement.
```
jattasNI, line 21:
```suggestion
Design System team members remain as co-owners to enforce technical quality of implementation. They can (and will) give consulting on feature selection (API design for properties/attributes/methods/events, level of accessibility support for ARIA/keyboard/mouse, interaction design, visual design), while ultimately deferring to the Spright contributor for what to implement.
```

packages/spright-components/CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+Spright component bug fixes and new features should be contributed by the team that needs them.
```
jattasNI, line 23:
```suggestion
Spright component bug fixes and new features should be contributed by the team that needs them. The Design System team will maintain infrastructure like pipelines and dependencies but may request assistance from contributing teams for component-specific issues.
```

packages/spright-components/README.md
```diff
@@ -0,0 +1,9 @@
+<div align="center">
+    <p><b>ni | spright</b></p>
+</div>
+
+# Spright Components
+
+[![NPM Version](https://img.shields.io/npm/v/@ni/spright-components.svg)](https://www.npmjs.com/package/@ni/spright-components)
+
+The `spright-components` package is intended for components that do not belong in `nimble-components` due to being: ["molecules"](https://atomicdesign.bradfrost.com/chapter-2/), product-specific, data-connected, or not "general utility". The primary component implementation should still be built on Nimble technologies (FAST, native web components, etc).
```
jattasNI, line 9:
```suggestion
NI-styled web components following [Nimble Design System](https://nimble.ni.dev) patterns but not part of the core design system.

The `spright-components` package is intended for components that do not belong in [`nimble-components`](/packages/nimble-components) due to being: ["molecules"](https://atomicdesign.bradfrost.com/chapter-2/), product-specific, data-connected, or not "general utility". The primary component implementations are built on Nimble technologies (FAST, native web components, etc).
```

packages/spright-components/README.md
```diff
@@ -0,0 +1,9 @@
+<div align="center">
+    <p><b>ni | spright</b></p>
+</div>
+
+# Spright Components
```
jattasNI, line 5: I'm debating whether we want to provide an equivalent of the Getting Started section in the nimble-components README. Would it be useful to say something like "Follow the instructions in the nimble-components README, replacing "nimble" with "spright" as appropriate"?

I'm guessing that will be obvious to most users of the package, but it couldn't hurt in case people stumble on Spright before Nimble.

packages/spright-components/src/all-components.ts
jattasNI: Yeah, we can work it out when we get to SprightBlazor.

packages/spright-components/src/demo/template.ts
```diff
@@ -0,0 +1,22 @@
...
+    <div
+        class="backplate ${x => x.shape}"
+        part="backplate"
+        style="${x => (x.fill ? `background-color: ${x.fill};` : undefined)}"
+    >
+        <a
+            class="link"
+            part="link"
+            href="${x => (x.link ? x.link : undefined)}"
+            style="${x => (x.color ? `color: var(--avatar-color-${x.color});` : undefined)}"
```
jattasNI, line 15: Where is `--avatar-color-...` set?

jattasNI (approved):

change/@ni-spright-components-08e968ac-291f-4c66-949e-115ca77deef2.json
rajsite: - Leaving as 0.0.1 is fine. Once the infra is set-up and ready for contributions we should bump it to 1.0.0.
- You don't need to check-in the empty changelog md and json files. The intention was that we should do a patch release so beachball starts generating them. I'm not sure if beachball will be confused by having the empty files so we can just remove them and let beachball handle them

packages/spright-components/README.md
rajsite: I think each public package should have a very clear README and CONTRIBUTING each with a Getting Started started section that is step-by-step and completable. I can own adding that after this PR goes in because I think nimble's and spright's need a clean-up for the latest era.

packages/spright-components/src/all-components.ts
rajsite: I think we should go ahead and import the nimble all-components from the spright all-components. The all-components biggest user is Blazor but it's not the only user [(codesearch)](https://codesearch.natinst.com/search/text?q=all-components-bundle%20-repo%3Animble%20-path%3A.cshtml&fold_case=auto&regex=true&context=true) [(GitHub)](https://github.com/search?q=org%3Ani+all-components-bundle+-repo%3Ani%2Fnimble+-repo%3Ani%2Fspright&type=code).

As-is I think a user that tries to use both nimble all-components-bundle and spright all-components-bundle in a page from separate builds could have issues because each all-components-bundle would load their own copy of FAST element, etc which is at best extra bulk and at worst (and I think expected) will cause conflicts and issues.

packages/spright-components/rollup.config.js
```diff
@@ -0,0 +1,60 @@
+import commonJS from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import sourcemaps from 'rollup-plugin-sourcemaps';
+import terser from '@rollup/plugin-terser';
+import replace from '@rollup/plugin-replace';
+import json from '@rollup/plugin-json';
+import nodePolyfills from 'rollup-plugin-polyfill-node';
```
rajsite, line 7: Was initially thinking we could remove things that are, i.e. table or rich-text specific but this build does need to be able to build all-components from nimble so all this stuff stays 👍

packages/spright-components/src/demo/template.ts
rajsite: Chatted with Jesse and lets just simplify the component further and give it a name that is a clear responsibly.

spright-rectangle

template.ts
```
<slot></slot>
```

styles.ts
```
display inline-block

:host {
  // border color tokens
}

slot {
   // font tokens
}
```

Scope of "demo" is unclear and could see it growing unnecessarily to demonstrate lots of things like custom tokens, accessibility patterns, component composition, plugins, whatever. Instead spright-rectangle is super scoped and small.

packages/spright-components/src/all-components.ts
rajsite: Also see the statement from the other comment about being derived from FAST Foundation but I think we just keep it even simpler and derive from FASTElement. I kinda expect long-term most things in FAST Foundation to be in Nimble instead of spright and even if extending from FAST Foundation nimble has plenty of examples already.

My understanding is nimble-rectangle is super simple and something just for infrastructure set-up that we can remove when we get a real spright component

packages/spright-components/src/demo/template.ts
rajsite: Also see the statement from the other comment about being derived from FAST Foundation but I think we just keep it even simpler and derive from FASTElement. I kinda expect long-term most things in FAST Foundation to be in Nimble instead of spright and even if extending from FAST Foundation nimble has plenty of examples already.

My understanding is nimble-rectangle is super simple and something just for infrastructure set-up that we can remove when we get a real spright component

change/@ni-spright-components-08e968ac-291f-4c66-949e-115ca77deef2.json
jattasNI: That was my misunderstanding / forgetting how beachball works. But Milan remembered better: I confirmed from some recent new library PRs in skyline that beachball can create it and we don't need it to be in this PR.

packages/spright-components/src/demo/template.ts
m-akinc: > Also see the statement from the other comment about being derived from FAST Foundation...

Not sure what you're referring to.

> ...but I think we just keep it even simpler and derive from FASTElement.

Needed to derive from `FoundationElement` in order to have a `compose` method to call.

rajsite: >> Also see the statement from the other comment about being derived from FAST Foundation...
> 
> Not sure what you're referring to.

That was my interpretation of this statement: https://github.com/ni/nimble/pull/1986#discussion_r1571370842

> I went with option 3, but I wanted something that derived from a FAST component, so I made spright-demo which is actually a not-very-functional user avatar.

That it was desired to extend a component that already existed in Fast Foundation (assumed that was the intention of "derived from a FAST component").

change/@ni-spright-components-08e968ac-291f-4c66-949e-115ca77deef2.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: waiting for a passing build to verify no regressions in build logs for new infra

packages/spright-components/src/demo/template.ts
m-akinc: Got it. 👍I thought you were directing me to refer to a comment _you_ made elsewhere, when you meant that you saw _my_ comment.

packages/spright-components/src/all-components.ts
```diff
@@ -0,0 +1,9 @@
...
+ * Import of all the web components available in Spright AND Nimble.
+ * Production applications are encouraged to import only components
+ * that are required instead of leveraging this file.
+ */
+
+import './rectangle';
+
+import '@ni/nimble-components/all-components';
```
rajsite, line 9: Flip the order so all nimble components are registered before spright components.
Nimble's library stuff is in a dist folder

```suggestion
import '@ni/nimble-components/dist/esm/all-components';

import './rectangle';
```

packages/spright-components/rollup.config.js
rajsite: Re-opening, guess things were removed from the roll-up. As stated since we are pulling in all nimble-components need the stuff from there.

The circular dependency stuff is missing: https://github.com/ni/nimble/blob/main/packages/nimble-components/rollup.config.js#L21

Resulting in build warnings: https://github.com/ni/nimble/actions/runs/8759928156/job/24043951459#step:12:1472

packages/spright-components/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
+# Contributing to Spright Components
+
+This package generally uses the same tooling and policies as Nimble components. See the [`nimble-components` CONTRIBUTING doc](../nimble-components/CONTRIBUTING.md) for an overview. For details on whether a component belongs in Nimble or Spright, refer to the [README](./README.md). This document describes ways in which Spright development differs from Nimble.
```
rajsite, line 3: Paths in markdown need to be relative to repo root
`/packages/nimble-components/CONTRIBUTING.md`
`/packages/spright-components/README.md`

packages/spright-components/src/rectangle/styles.ts
```diff
@@ -0,0 +1,21 @@
...
+    bodyFontColor,
+    borderColor,
+    borderWidth
+} from '@ni/nimble-components/dist/esm/theme-provider/design-tokens';
+
+export const styles = css`
+    ${display('inline-block')}
+
+    :host {
+        border: ${borderWidth} solid ${borderColor};
```
rajsite, line 14: nit / pref: lets use borderColorHover just to give it some pop

m-akinc: Yep. I made that same change in my other branch where I am adding the Storybook. I also added some other styling (hover, disabled) to make the stories a bit more educational, but we can discuss that in the next PR.

rajsite: I guess so, but like I started [here](https://github.com/ni/nimble/pull/1986#discussion_r1572889302) I'm not looking to maintain this component as an example and treating it as something to delete when we get a real component. It's just to help setup infra.

packages/spright-components/rollup.config.js
rajsite: FYI @m-akinc this comment is still open

rajsite (commented):
Waiting for comments to be addressed

rajsite (changes requested):
Needs correct package version management

packages/spright-components/package.json
```diff
@@ -0,0 +1,129 @@
...
+  "bugs": {
+    "url": "https://github.com/ni/nimble/issues"
+  },
+  "homepage": "https://github.com/ni/nimble#readme",
+  "dependencies": {
+    "@microsoft/fast-colors": "^5.3.1",
+    "@microsoft/fast-element": "^1.12.0",
+    "@microsoft/fast-foundation": "^2.49.6",
+    "@microsoft/fast-web-utilities": "^6.0.0",
+    "@ni/nimble-components": "*",
```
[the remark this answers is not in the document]
m-akinc: Thanks, I had been thinking of `"*"` as "latest", since it allows the newest version, but of course that's not what it means. Switched it back.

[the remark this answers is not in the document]
rajsite: "latest" does exist but we should use semantic version ranges to specify the version we actually test against. i.e. latest changes day-to-day and doesn't care about major versions, etc. If you want to make sure the mental model for how semver works is aligned we should discuss offline as you're setting up more packages.

rajsite (approved):

packages/spright-components/package.json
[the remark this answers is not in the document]
rajsite: This came up in the wild for a third party package using wildcard specifiers for dependencies as an example: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69355#issuecomment-2088994473

Even though the linkify-it package did a major version release to specify a breaking change users of markdown-it (directly or indirectly) were broken by the major change because markdown-it used a wildcard specifier.

PR #1949, opened 2024-03-20
Combobox registerOption reversion to address Angular bug

atmgrifter00:
# Pull Request

## 🤨 Rationale

[AzDO bug 'Nimble combo box | Issue in showing list options'](https://ni.visualstudio.com/DevCentral/_workitems/edit/2687002)

## 👩‍💻 Implementation

Simple reversion of problem code. I've also created a [tech debt item](https://github.com/ni/nimble/issues/1948) to track the work needed to ultimately re-align the `Combobox` and `Select` implementations.

## 🧪 Testing

Just removing the related test.

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #1991, opened 2024-04-01
Combobox CVA `addOption` optimization.

atmgrifter00:
# Pull Request

## 🤨 Rationale

This PR addresses the root issue behind the actual performance problem identified in [this AzDO issue](https://ni.visualstudio.com/DevCentral/_workitems/edit/2686999), that resulted in reverting a change in `nimble-components` that simply exposed the CVA flaw.

## 👩‍💻 Implementation

When the CVA adds options, it was always resetting the value of the `Combobox` in case the newly added option had a value assigned to it that matched the current assigned model value to the `Combobox`. This was so that it could update the display text to what the newly added option had.

However, there is no need to attempt to update the display value if the newly added option does not have a value that matches the current `Combobox` value. The fix is to simply put that guard around attempting to update the display value.

## 🧪 Testing

Unit tests still pass. Manual testing, and ran performance checks against a re-implementation of the `ListOptionOwner` by the `nimble-components` `Combobox`, and verified there was no slowdown.

## ✅ Checklist


angular-workspace/projects/ni/nimble-angular/src/directives/combobox/nimble-combobox-control-value-accessor.directive.ts
```diff
@@ -131,7 +131,10 @@ export class NimbleComboboxControlValueAccessorDirective implements ControlValue
         } else {
             this._displayTextToOptionsMap.set(option.text, [option]);
         }
-        this.updateDisplayValue();
+
```
rajsite, line 134: Wouldn't we also unrevert this change?
https://github.com/ni/nimble/pull/1949

Maybe it's worth testing tgz builds of both changes together in the end application to make sure a regression is not re-introduced?

atmgrifter00: Yes, I was just going to do that as a separate PR. I don't mind doing them both at the same time though. I'll test a tgz build with both changes.

atmgrifter00: I've now "unreverted" the change to the `nimble-components` `Combobox`. I've tested the tgz builds in the application with the noted regression and saw no performance change.

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -127,6 +127,33 @@ describe('Combobox', () => {
         await disconnect();
     });
 
+    fit('option added directly to DOM synchronously registers with Combobox', async () => {
```
rajsite, line 130:
```suggestion
    it('option added directly to DOM synchronously registers with Combobox', async () => {
```

rajsite (approved):

PR #1992, opened 2024-04-01
Update mapping column HLD

mollykreis:
# Pull Request

## 🤨 Rationale

As part of #1873, we've decided to make some changes to the mapping columns.

The changes are described in detail in the HLD, but the high-level changes are:
- `nimble-table-column-enum-text` and `nimble-table-column-icon` will be replaced with a single column named `nimble-table-column-mapping`
- mapping elements are reworked to support icon/spinner + text in the same cell
- there will be a configuration on the column to make it fixed at 32px

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -38,29 +38,30 @@ Below is an example of how these elements would be used within a `nimble-table`:
...
         Error Code
         <nimble-mapping-text key="1" label="A bad thing happened"></nimble-mapping-text>
         <nimble-mapping-text key="2" label="A worse thing happened"></nimble-mapping-text>
         <nimble-mapping-text key="3" label="A terrible thing happened"></nimble-mapping-text>
-    </nimble-table-column-enum-text>
+    </nimble-table-column-icon>
     <nimble-table-column-icon field-name="archived" key-type="boolean">
         Archived
         <nimble-mapping-icon key="true" icon="nimble-icon-database" label="Archived"></nimble-mapping-icon>
+        <nimble-mapping-icon key="false" label="Not archived"></nimble-mapping-icon>
```
mollykreis, line 58: Note: I added this `nimble-mapping-icon` element here since that makes it follow the best practices of having a mapping for each possible record value.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -102,30 +103,14 @@ _Props/Attrs_
 
 -   `field-name`: string
 -   `key-type`: 'string' | 'number' | 'boolean'
+-   `label-hidden`: boolean - When `true`, the `label` for a mapping will be rendered in a cell next to the specified icon. Note, that this attribute only affects what is rendered in a cell; a group row will always render both the icon and the label.
+-   `icon-width`: boolean - When `true`, the column will have a fixed width that makes the column the appropriate width to render only a single icon in the cell. This should only be set to `true` when the header contains a single icon (no text) and `label-hidden` is `true`. If set to `true` when `label-hidden` is `false`, the column's configuration will be considered invalid.
```
mollykreis: I'm open to suggestions about the name of this attribute. Particularly, I want to convey that when this attribute is set the column will not be resizable and will be very narrow. I also what to avoid confusion with anything more generic we add to columns to make them fixed-width/not resizable/pixel width.
	
Some other ideas I thought about were:
- fixed-width: I decided against this name because I didn't think it properly conveyed that the column would be narrow. I also thought it could be confused with more generic column sizing features we might add in the future.
- icon-only: This sounds more about what is being displayed than the size.
- fixed-icon-width: More verbose than `icon-width` with no clear advantage. It also sounds more like the icon is fixed width than the column.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -205,38 +147,33 @@ html<TableColumnMappingHeaderView>`${x => x.getMappingToRender().groupHeaderView
...
-<span
-    ${ref('span')}
-    @mouseover="${x => {
-        x.isValidContentAndHasOverflow = !!x.label && x.span!.offsetWidth < x.span!.scrollWidth;
-    }}"
-    @mouseout="${x => {
-        x.isValidContentAndHasOverflow = false;
-    }}"
-    title=${x => (x.isValidContentAndHasOverflow ? x.label : null)}>
+<span>
```
mollykreis, line 168: Note: I removed the `mouseover`/`mouseout` code from here because I didn't think it was particularly relevant to the plan here. We now have the `overflow` behavior that we'll use, and the HLD already says that we'll show the title if the label gets ellipsized.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -205,38 +147,33 @@ html<TableColumnMappingHeaderView>`${x => x.getMappingToRender().groupHeaderView
...
 </span>`;
 ```
 
 ### Grouping
 
-Grouping will be based on the record value. The grouping header will display the rendered icon/spinner/text. In the case of an icon/spinner, it will also be followed by the `label` text. This will disambiguate cases where multiple record values map to the same icon (assuming the labels are different).
+Grouping will be based on the record value. The grouping header will display the rendered label along with the icon/spinner, if configured.
 
-For values that do not match any mapping, we will display the raw data value. While this introduces inconsistency, it seems preferable to the alternative, which is having multiple, separate groupings with a blank header (well, with just the item count in parens). Even in the case where there is a default mapping, we would still end up with separate groups with the identical default mapped icon and/or text, which is just as bad.
+For values that do not match any mapping, the group row will display only the count. This is considered invalid data from the table's perspective and should be fixed within the client application.
```
mollykreis, line 177: Note: This is not a proposed behavior change. I'm simply updating the HLD to reflect the decisions we made while initially implementing the column.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -102,30 +104,14 @@ _Props/Attrs_
 
 -   `field-name`: string
 -   `key-type`: 'string' | 'number' | 'boolean'
+-   `display`: enum with possible values of `iconOnly`, `labelOnly`, and `iconAndLabel` - Specifies what is rendered in a cell for the column. Note, that this attribute only affects a cell's rendering; a group row will always render both the icon and the label.
```
mollykreis, line 107: I went back and forth on this attribute. I considered making it a boolean (`label-hidden`) to align with a number of our other components (e.g. `content-hidden` on buttons, `underline-hidden` on anchors, `title-hidden` on banners & dialogs, etc).

However, I decided that making it an enum was probably a better decision because:
1. We could do it in a backwards compatible way if we make the default value `iconOnly`
2. It allows us to add more configurations in the future without making a breaking change to the API.

mollykreis:

m-akinc (commented):
My primary feedback is that I'd like us to revisit the API (and remove/deprecate the existing elements) rather sticking to names that don't make as much sense anymore.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
mollykreis: This is now on the mapping element itself (specifically for icon and spinner mappings), and it is a boolean called `label-hidden`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The `nimble-table-column-enum-text` is a component that supports rendering specific number, boolean, or string values as mapped text. `nimble-table-column-icon` instead maps values to icons and/or spinners. The actual mappings are defined by child elements `nimble-mapping-icon`, `nimble-mapping-spinner`, and `nimble-mapping-text`.
+The `nimble-table-column-icon` is a component that supports rendering specific number, boolean, or string values as icon/label pairs and/or spinner/label pairs. The mappings are defined by child elements of `nimble-mapping-icon` and `nimble-mapping-spinner`.
```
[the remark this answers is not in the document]
mollykreis: Updated to `nimble-table-column-enum`.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -102,30 +104,14 @@ _Props/Attrs_
 
 -   `field-name`: string
 -   `key-type`: 'string' | 'number' | 'boolean'
+-   `display`: enum with possible values of `iconOnly`, `labelOnly`, and `iconAndLabel` - Specifies what is rendered in a cell for the column. Note, that this attribute only affects a cell's rendering; a group row will always render both the icon and the label.
+-   `icon-width`: boolean - When `true`, the column will have a fixed width that makes the column the appropriate width to render only a single icon in the cell. This should only be set to `true` when the header contains a single icon (no text) and `display` is `iconOnly`. If set to `true` when `display` is not `iconOnly`, the column's configuration will be considered invalid.
```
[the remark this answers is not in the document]
mollykreis: I'm marking this as resolved in favor of the existing conversation thread in this PR: https://github.com/ni/nimble/pull/1992#discussion_r1546812418

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
mollykreis: Combining conversation threads, @m-akinc posted [here](https://github.com/ni/nimble/pull/1992#discussion_r1546957248): 
>I have reservations about the name icon-width, because without proper context, it strongly implies a numeric value controlling the width of an icon (or icons). I realize that having the word "icon" in there is important, because naming it something like fixed-width would sound like it could be used even when the column/header contained text. Maybe icon-fixed-width?

mollykreis: I'm not opposed to `icon-fixed-width` either, but for now, I've updated my proposal to `fixed-to-icon-width`. It's a bit verbose, but I think it makes it more explicit that nothing about the icon is "fixed width" -- instead it's the column that is fixed to an icon's width. I also thought about using the term "locked" rather than "fixed" (i.e. `locked-to-icon-width`).

I'm still happy to take more opinions on this name.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -135,9 +121,9 @@ _Component Name_
 _Props/Attrs_
 
 -   `key`: string | number | boolean | undefined
--   `icon`: string - name of the Nimble icon element
+-   `icon`: string | undefined - name of the Nimble icon element. If `undefined`, no icon will be associated with the given `key`.
```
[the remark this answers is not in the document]
mollykreis: I've brought back the `nimble-mapping-text` element so that clients won't have to use `nimble-mapping-icon` even when they never want an icon. I've also added `nimble-mapping-placeholder` to cover some more use cases.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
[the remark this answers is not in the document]
mollykreis: Updated wording.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -135,9 +121,9 @@ _Component Name_
...
 -   `severity`: string - one of the supported enum values. Controls color of the icon.
--   `label`: string - localized value used as the accessible name and `title` of the icon. Will also be displayed in the group header.
+-   `label`: string - localized value associated with the given `key`.
```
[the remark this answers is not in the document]
mollykreis: I don't think creating two attributes for "value label" and "icon label" are necessary. We can update the docs if we want to make the expectation about these labels clear.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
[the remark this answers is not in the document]
mollykreis: We tend to not do "magical" things like that. We also don't do anything today with the content slotted in a column header. Therefore, I'm leaning towards leaving this as a property on the column so that clients can explicitly opt-in to this behavior. However, I'll leave the conversation open for more discussion.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -92,66 +96,69 @@ Cons:
...
+
 _Component Name_
 
 -   `nimble-mapping-icon`
 
 _Props/Attrs_
 
 -   `key`: string | number | boolean | undefined
--   `icon`: string - name of the Nimble icon element
+-   `icon`: string | undefined - name of the Nimble icon element. If `undefined`, no icon will be associated with the given `key`, but space will be reserved in the cell for an icon.
```
m-akinc, line 127: If a client doesn't want empty space reserved for missing icons, they can just use `nimble-mapping-text` instead, right?

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -161,37 +168,18 @@ _Props/Attrs_
...
-    <slot
-        ${slotted('mappings')}
-        name="mapping">
-    </slot>
-    <span class="header-content">
-        <slot></slot>
-    </span>
-</template>
-```
+The placeholder mapping element will display an empty cell. A group row associated with a placeholder mapping will display the mapping's text. The purpose of the placeholder mapping element is to allow clients to avoid cluttering their table with information that isn't particularly helpful to a user (e.g. that the state of a system is 'idle') while still having a good grouping experience that ensures group rows are not blank.
```
m-akinc, line 173: A few thoughts:
1. Grouping will still be based on the key values, which could be confusing if multiple placeholder mappings (i.e. with different keys) are used. Of course in that case, different `label`s should be used to mitigate confusion.
2. Initially I thought it might be better to just support `label-hidden` on `nimble-mapping-text` instead of introducing another mapping element, but a text mapping that hides its text is probably weird enough that a new element is warranted.
3. I don't think "placeholder" is the right term. In the table, we've used "placeholder" to mean text we show when a data value is missing. In this case, we're not showing any text, and it's not (necessarily) for missing values. I'd suggest we go with `nimble-mapping-blank` or `nimble-mapping-empty` since this is really for when the client explicitly wants blank cells.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -70,7 +74,7 @@ If multiple mappings in a column have the same key, an error flag will be set on
 
 If an invalid `icon` value is passed to `nimble-mapping-icon`, an error flag will be set on the column's validity object. An invalid `icon` value is any element that cannot be resolved or that does not derive from `Icon`.
 
-`nimble-table-column-icon` supports only `nimble-mapping-icon` and `nimble-mapping-spinner` as mapping elements. `nimble-table-column-enum-text` supports only `nimble-mapping-text`. Unsupported mappings will result in an error flag being set on the column's validity object.
+`nimble-table-column-enum` supports `nimble-mapping-icon`, `nimble-mapping-spinner`, `nimble-mapping-text`, and `nimble-mapping-placeholder` as mapping elements. Unsupported mappings will result in an error flag being set on the column's validity object.
```
m-akinc, line 77: This isn't really relevant anymore.
```suggestion
```

m-akinc (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
mollykreis: Yes, that is correct.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
mollykreis: I agree that putting `label-hidden` on `nimble-mapping-text` feels weird and that `placeholder` likely isn't the correct term either. I'll update this to `nimble-mapping-blank`.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -92,66 +94,69 @@ Cons:
...
--   `nimble-table-column-enum-text`
+-   `nimble-table-column-enum`
 
 _Props/Attrs_
 
 -   `field-name`: string
 -   `key-type`: 'string' | 'number' | 'boolean'
--   `fractional-width`: number (defaults to 1)
--   `min-pixel-width`: number (defaults to minimum supported by table)
+-   `fixed-to-icon-width`: boolean - When set, the column will have a fixed width that makes the column the appropriate width to render only a single icon in the cell. This should only be set when the header contains a single icon (no text) and none of the child mapping elements will result in text being rendered in a cell.
```
rajsite, line 107: If the specificity in the attribute name is there to enable API expansion, i.e. `fixed-to-something-else-width`, then maybe we make it an enum off the bat?

```ts
TableColumnEnumFixedWidthMode {flexibleWidth: undefined, iconWidth: 'icon-width'}
```

```html
fixed-width-mode="icon-width"
```

rajsite (approved):
API-wise looks good, just some Qs on naming

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -92,66 +94,69 @@ Cons:
...
 _Content_
 
--   column title (text or icon)
--   1 or more `nimble-mapping-text` elements
+-   column title (icon and/or text)
+-   1 or more `nimble-mapping-icon`, `nimble-mapping-spinner`, `nimble-mapping-text`, or `nimble-mapping-empty` elements
 
 #### Mapping element (icon):
 
+The icon mapping element will support displaying an icon, icon with text, or only text in a cell. If only text is displayed in a cell, space will be reserved for an icon so that the text associated with all icon mapping elements is aligned. A group row associated with an icon mapping element will always display the icon, if specified, and text.
```
jattasNI, line 116: It took me a minute to wrap my head around the distinction between an icon mapping with no icon and a text mapping. I think it makes sense, but we'll probably want to write usage guidance in storybook that explains which combinations of mappings are recommended. I'm guessing we wouldn't suggest mixing text mappings with text-only-icon mappings since they'd end up out of alignment.

No action needed in this PR, just a reminder for when you get to the implementation PR(s).

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -92,66 +94,69 @@ Cons:
...
--   column title (icon)
--   1 or more `nimble-mapping-icon` or `nimble-mapping-spinner` elements
-
-#### General mapping column element:
+#### Enum column element:
 
 _Component Name_
 
--   `nimble-table-column-enum-text`
+-   `nimble-table-column-enum`
```
jattasNI, line 101: I may have missed a discussion thread or offline chat, but why `nimble-table-column-enum` instead of `nimble-table-column-mapping`? In the [table column naming guidance](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-columns-hld.md#column-type-philosophy) we talk about `nimble-table-column-mapping` because the presentation of the column is a mapping element. Using `-enum` seems more  like data type terminology than presentation terminology to me, which we haven't done so far.

If this has already been debated an there's good rationale for `enum` then I don't want to prolong the discussion. But we should update that naming guidance to include this new example and remove that note about mapping columns.

jattasNI (approved):

change/@ni-nimble-components-e559fa6d-6507-478a-9e4e-889ea5cb3c9d.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
mollykreis: No one in SLE is using it today. I think there will still be plenty of integration work to do (or coordinate) as part of these API changes, but deleting that column shouldn't be part of it.

We'll need to delete the `sl-table` version of that column and remove it from storybook & the example app, but there shouldn't be any changes required in app code.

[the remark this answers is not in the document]
mollykreis: >Do we think columns that display an icon with text in their cells will want to also display an icon with text in their header?

Perhaps, but I think it'll actually be more common for those columns to only have text. I believe that @fredvisser is planning on creating some usage guidance around when we would expect icon-only columns or icon + text columns. If we have any specific thoughts on when the header should also be icon + text, perhaps we could add it there.

> If yes, do we need to do any work to enable that? I'm guessing it's already supported but we might want to add styling / docs / chromatic tests to ensure the header content is aligned with the cell content.

Technically, we already support a client doing that if they want. We just don't do any special alignment of the icon and text. Right now, a client would need to add their own padding between the icon and the text.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
mollykreis: We did discuss this in a meeting and settled on `nimble-table-column-enum`, but I don't actually remember why we preferred `-enum` rather than `-mapping`. @rajsite, do you remember? If I'm not mistaken, I think you were the one that suggested `-enum`.

change/@ni-nimble-components-e559fa6d-6507-478a-9e4e-889ea5cb3c9d.json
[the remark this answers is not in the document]
jattasNI: Cool, I'll resolve this for now and we can decide what guidance we want to provide for headers when we write that usage documentation.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
rajsite: I think I said that because since then we had renamed the column `enum` from `mapping` so seemed like just a stale doc.

I don't think that on their own either 'enum' or 'mapping' clearly represent presentations. Also, mappings at least initially were intended as generic models hence not owning a template / specific visual and components like enum / icon column or rich-text mention decide how to render specific mappings (and may do so differently).

If we say mappings are visual models (still don't own a specific view but are describing parameters of any(?) presentation of them) then I guess `nimble-table-column-mapping` is a better name since mappings are closer to presentations in that point-of-view. I think that makes adding things like "text-hidden" seem conceptually more aligned where before it was feeling maybe too table view specific.

Under that mental model I'm aligned on `nimble-table-column-mapping`

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping.md
```diff
@@ -92,66 +94,69 @@ Cons:
...
 
 _Props/Attrs_
 
 -   `field-name`: string
 -   `key-type`: 'string' | 'number' | 'boolean'
--   `fractional-width`: number (defaults to 1)
--   `min-pixel-width`: number (defaults to minimum supported by table)
+-   `fixed-width-mode`: enum - `flexibleWidth` (default) | `iconWidth` - When set to `iconWidth`, the column will have a fixed width that makes the column the appropriate width to render only a single icon in the cell. This should only be set when the header contains a single icon (no text) and none of the child mapping elements will result in text being rendered in a cell.
```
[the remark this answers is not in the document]
rajsite: Chatted with molly to clarify a couple things.

- The column should have the fractional width mixin for the fractional-width and min-pixel-width properties. They were unintentionally omitted.
- Rounded on keeping the enum but tweaking the names to avoid repeating "width", to use the term icon-size which aligns with the design token name, and to not give an explicit name of "flexible" to the default mode but just call it default (i.e. the standard sizing behavior from all other columns).

So concretely:
`width-mode: default | icon-size`

PR #1997, opened 2024-04-03
Improve TColumnConfig typing on columns that extend TableColumnTextBase

### LINKED ISSUE #1964: TableColumnTextBase loses type of TColumnConfig
# 🧹 Tech Debt
Columns that extend `TableColumnTextBase` don't enforce the expected type of their `TColumnConfig` because `TableColumnTextBase` is not generic. Making `TableColumnTextBase` generic, however, isn't straightforward because TypeScript doesn't support generics and mixins together: https://github.com/microsoft/TypeScript/issues/26154.

We should look into options that allow columns extending `TableColumnTextBase` to have `TColumnConfig` not use the default value of `unknown`.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1964 

## 👩‍💻 Implementation

`TableColumnTextBase` cannot simply be made generic because TypeScript doesn't support generics and mixins together: https://github.com/microsoft/TypeScript/issues/26154.

Therefore, each concrete instance of `TableColumnTextBase` should add its own mixins in a way that they get mixed into `TableColumnTextBase<TColumnConfig>` rather than `TableColumn`. To avoid code duplicate, I created a function, `mixinTextBase`,  that mixes in the appropriate mixins for `TableColumnTextBase`.

## 🧪 Testing

- Unit tests still pass
- Manually verified that the type of `this.columnInternals.columnConfig` is correct for a column using `mixinTextBase`
- Sanity tested examples in storybook to make sure columns still behave as expected

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table-column/text-base/index.ts
+++ b/packages/nimble-components/src/table-column/text-base/index.ts
@@ -1,5 +1,8 @@
 import { attr } from '@microsoft/fast-element';
 import { TableColumn } from '../base';
+import { mixinFractionalWidthColumnAPI } from '../mixins/fractional-width-column';
+import { mixinGroupableColumnAPI } from '../mixins/groupable-column';
+import { mixinColumnWithPlaceholderAPI } from '../mixins/placeholder';
 
 /**
  * The base class for table columns that display fields of any type as text.
@@ -14,4 +17,13 @@ export abstract class TableColumnTextBase<
         this.columnInternals.dataRecordFieldNames = [this.fieldName];
         this.columnInternals.operandDataRecordFieldName = this.fieldName;
     }
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/explicit-function-return-type
+export function mixinTextBase<TColumnConfig>() {
+    return mixinGroupableColumnAPI(
+        mixinFractionalWidthColumnAPI(
+            mixinColumnWithPlaceholderAPI(TableColumnTextBase<TColumnConfig>)
+        )
+    );
 }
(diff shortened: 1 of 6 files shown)


mollykreis:

atmgrifter00 (approved):

packages/nimble-components/src/table-column/text-base/index.ts
```diff
@@ -18,3 +18,12 @@ export abstract class TableColumnTextBase extends mixinGroupableColumnAPI(
         this.columnInternals.operandDataRecordFieldName = this.fieldName;
     }
 }
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/explicit-function-return-type
+export function mixinTextBase<TColumnConfig>() {
+    return mixinGroupableColumnAPI(
+        mixinFractionalWidthColumnAPI(
+            mixinColumnWithPlaceholderAPI(TableColumnTextBase<TColumnConfig>)
```
rajsite, line 26: Is it possible to take the class to mixin on as a parameter? i.e. something with type like `TBase extends TableColumnTextBase<TColumnConfig>`

My reasons are this method isn't really a "mixin" since it's not mixing in behavior on a provided class. It's more creating a TextBase class vs a mixin of behavior onto an exisiting class.

Other reason is that we don't need it now but if we wanted to have additional class layers this would be restrictive, i.e.

```
DateText -> TimeTextBase -> TextBase -> TableColumn
DurationText -> TimeTextBase -> TextBase -> TableColumn
NumberText -> TextBase -> TableColumn
```

We'd want to be able to do something like
```
DateText extends mixinTextBase(TimeTextBase) {}
DurationText extends mixinOther(mixinTextBase(TimeTextBase)) {}
NumberText extends mixinTextBase(TextBase) {}
```

mollykreis: I figured out how to get the typing to work correctly with the approach you suggested. I've also tested that `mixinTextBase` works correctly with a type that inherits from `TextBase`, such as `TimeTextBase` in your example.

Issue #1951, opened 2024-03-20
Ability to clear the selection from nimble-select


jattasNI:
# 🙋 Feature Request

## 😯 Problem to Solve

Now that we have placeholder items in select, there is no way to return to the initial "no selection" state. This is needed for use cases like a user selector where you want to unassign a work item.

## 💁 Proposed Solution

Some design systems do this using an 'x' button on the right of the input. But we need an IxD decision on whether this is the right approach and what the keyboard interaction would be.
<img width="747" alt="image" src="https://github.com/ni/nimble/assets/10500124/a05ee804-11f9-419b-8130-d204bd275ac7">

Clients need the ability to toggle whether the selection can be cleared.

jattasNI:
First step is to agree on interaction design so assigned to @fredvisser.

PR #1998, opened 2024-04-04
Add 'clear selection' workflow to Select IxD spec

fredvisser:
# Pull Request

## 🤨 Rationale

Interaction design spec update is a prerequisite for #1951.

Clients need the ability to allow the selection to be cleared.

## 👩‍💻 Implementation

- Updated the Figma [Nimble - IxD](https://www.figma.com/file/r2yGNQNVFdE7cBO9CyHmQx/Nimble---IxD?type=design&node-id=1-3&mode=design&t=XrrbLPuzJMjuQZtv-0) doc with new interactions
- Updated IxD spec with clear selection option

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -215,6 +226,7 @@ Typing into the filter bar filters the list options in the dropdown (not picture
 | `HOME`                 | Opens the dropdown and moves visual focus to the first list option                                                                                                                      |
 | `END`                  | Opens the dropdown and moves visual focus to the last list option                                                                                                                       |
 | _Printable characters_ | Opens the dropdown (if closed), moves visual focus to the filter input (if filterable) with the characters typed or to the first list option that matches the typed character or string |
+| `ESC`                  | Clears the selected item without moving focus (if clearable)                                                                                                                            |
```
fredvisser, line 229: There is very little consensus about the keyboard interactions in existing design systems:
1. The [majority](https://shoelace.style/components/select#clearable) of select/dropdown controls (that allow clearing the selected option) only support the 'clear selection button'. None of them support tab focus of the 'clear selection button'. 
2. A small [minority](https://dash.plotly.com/dash-core-components/dropdown) support `ESC`
3. A small [minority](https://style.monday.com/?path=/story/inputs-dropdown--dropdown-value-selection) support `Delete`.

| Approach | Pros | Cons |
| --- | ---| ----|
| Button only | Majority rule | Not keyboard accessible |
| `ESC` | Consistent with Combobox [ARIA guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) | Also used to close dropdown, and some users expect to use it to close slide-in dialogs |
| 'Delete` | Consistent with likely future multi-select/picker keyboard interactions | Low discoverability |

mollykreis (approved):

atmgrifter00 (approved):

packages/nimble-components/src/select/specs/IxD.md
```diff
@@ -188,7 +197,9 @@ When the dropdown is open, clicking anywhere outside of a list option (including
 
 ![ ](spec-images/select-23.png)
 
-> NOTE: NEEDS VISUAL DESIGN FOR MULTIPLE FOCUS STATES?
+If the select component allows clearing the selected item, the _Clear selection button_ will appear when an option is selected. Clicking the button will clear the selected item and revert to the placeholder state.
```
atmgrifter00, line 200: One thing that needs clarification is where the clear button should be placed relative to the error visible icon (when it is visible). I presume this is something that @NIbokeefe will specify?

fredvisser: image?

Yes, I'll ask Brandon to do a visual pass for the clear button, the various clear button states, and the error icon placement.

change/@ni-nimble-components-44c2b207-def9-472c-ae87-f6e48877185e.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Discussion topic, not blocking this PR: Would we ever want to add this feature to other inputs? Should we track that request? If it's low priority then I'm ok if we don't but just want to verify that it is low priority.
- text input: I've seen it used to clear search fields
- combobox: might be a natural change for "parity" with select (though users can achieve the same outcome with Delete)
- number: maybe for parity with text input but probably not a common use case
- text area and rich text editor: not likely useful

packages/nimble-components/src/select/specs/IxD.md
jattasNI: `ESC` feels natural enough to me and the alignment with the ARIA combobox pattern is compelling. I don't think the cons are dealbreakers.

jattasNI (approved):

change/@ni-nimble-components-44c2b207-def9-472c-ae87-f6e48877185e.json
fredvisser: I agree that we should add this to the combobox – clearing the Combobox with ESC isn't very discoverable today, and this will be more consistent and discoverable.

Filed https://github.com/ni/nimble/issues/2008

I also agree that this will be a requirement for a future search field, but that the other text areas don't need it.

change/@ni-nimble-components-44c2b207-def9-472c-ae87-f6e48877185e.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
```
jattasNI, line 2: You could consider mentioning this PR (or summarizing the proposal) on the [UX Design Discussion](https://teams.microsoft.com/l/channel/19%3A8786e8b04c62455798795d0a9d7aa362%40thread.tacv2/UX%20Design%20Discussion?groupId=9ee065d7-3898-4245-82f6-76e86084b8b1&tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e) channel. It's possible IxDs will have other ideas but even if they don't it'd be great to start engaging them.

packages/nimble-components/src/select/specs/IxD.md
rajsite: Maybe a mitigation for the con is that while the select is focused it always is capturing `esc`?
I can't remember if that would be in conflict with the future close watcher changes @m-akinc (but maybe we cross that bridge when we get to it)

rajsite (approved):

packages/nimble-components/src/select/specs/IxD.md
NIbokeefe: @fredvisser @atmgrifter00 @jattasNI 

In the select control section on [https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=9012%3A327273&mode=design&t=l9M1XTBQ0yvFnNny-1](url) Figma I've drawn some workflows and design options.

Screenshot 2024-04-10 at 4 37 42 PM

jattasNI: @NIbokeefe I'm guessing this was already your plan but in case not: let's get the team's feedback at our meeting today.

change/@ni-nimble-components-44c2b207-def9-472c-ae87-f6e48877185e.json
jattasNI: @fredvisser you'd suggested in a demo today that the SLE search field could have a clear button. I'm not sure how we'd recommend they do that today. Nimble doesn't yet offer a dedicated search field. The text input does include an [action slot](https://nimble.ni.dev/storybook/?path=/story/components-text-field--underline-text-field&args=label:Underline+Text+Field;type:text;appearance:underline;fullBleed:!false;value:;readonly:!false;disabled:!false;errorVisible:!false;errorText:Value+is+invalid;actionButton:!true;leftIcon:!false) which could contain a custom clear button, but it wouldn't exactly match the behavior we're proposing for the select (ESC would lose focus but not clear, action button would be keyboard focusable but clear button isn't).

Given this open question and the fact that the request is scope growth over sharing the existing search component sharing refactor, I'm not planning to drive this suggestion now with the implementing team. If you want to pick a direction track it as follow-up work, feel free.

jattasNI: #2021

PR #1999, opened 2024-04-05
Move column validator into columnInternals

### LINKED ISSUE #1389: Table columns should create validator as part of columnInternals

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1389

## 👩‍💻 Implementation

- Updated `TableColumn`, `ColumnInternals`, and `ColumnInternalsOptions` to be generic with `TColumnValidator extends ColumnValidator<[]>`
- Updated `ColumnInternals` to have a `validator` property whose type is enforced by the type of `TColumnValidator`
- Made `validator` a required input on `ColumnInternalsOptions`, so now every column has to specify a validator. In some cases this is a "default" validator with no validity keys, but making this required simplified a few different code paths. This included keeping columns with validators from having to use the `!` operator to assert that they knew that a validator had been set within their options.
- Updated columns to specify a validator and to use that validator through `ColumnInternals` rather than a class member on the column.

## 🧪 Testing

Existing tests still pass

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table-column/base/models/column-internals.ts
+++ b/packages/nimble-components/src/table-column/base/models/column-internals.ts
@@ -1,5 +1,5 @@
 import { uniqueId } from '@microsoft/fast-web-utilities';
-import { Observable, ViewTemplate, observable } from '@microsoft/fast-element';
+import { ViewTemplate, observable } from '@microsoft/fast-element';
 import { TableColumnSortDirection, TableFieldName } from '../../../table/types';
 import type { TableCell } from '../../../table/components/cell';
 import {
@@ -10,7 +10,7 @@ import {
 import type { TableGroupRow } from '../../../table/components/group-row';
 import { createGroupHeaderViewTemplate } from '../group-header-view/template';
 import { createCellViewTemplate } from '../cell-view/template';
-import { ColumnValidator } from './column-validator';
+import type { ColumnValidator } from './column-validator';
 
 export interface ColumnInternalsOptions<
     TColumnValidator extends ColumnValidator<[]> = ColumnValidator<[]>
@@ -82,12 +82,6 @@ export class ColumnInternals<
      */
     @observable
     public columnConfig?: TColumnConfig;
-
-    /**
-     * Whether this column has a valid configuration.
-     */
-    @observable
-    public validConfiguration = true;
 
     /**
      * The name of the data field that will be used for operations on the table, such as sorting and grouping.
@@ -185,16 +179,6 @@ export class ColumnInternals<
         this.delegatedEvents = options.delegatedEvents;
         this.sortOperation = options.sortOperation ?? TableColumnSortOperation.basic;
         this.validator = options.validator;
-
-        const notifier = Observable.getNotifier(this.validator);
-        notifier.subscribe(this);
-    }
-
-    /** @internal */
-    public handleChange(source: unknown, args: unknown): void {
-        if (source instanceof ColumnValidator && args === 'isColumnValid') {
-            this.validConfiguration = this.validator.isColumnValid;
-        }
     }
 
     protected fractionalWidthChanged(): void {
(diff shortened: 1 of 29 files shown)


mollykreis:

atmgrifter00 (approved):

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -163,14 +174,27 @@ export class ColumnInternals<TColumnConfig> {
...
         this.cellViewTemplate = createCellViewTemplate(options.cellViewTag);
         this.groupHeaderViewTemplate = createGroupHeaderViewTemplate(
             options.groupHeaderViewTag
         );
         this.delegatedEvents = options.delegatedEvents;
         this.sortOperation = options.sortOperation ?? TableColumnSortOperation.basic;
+        this.validator = options.validator;
+
+        const notifier = Observable.getNotifier(this.validator);
+        notifier.subscribe(this);
```
rajsite, line 190: ~~Not sure if there is an effective perf difference, but subscribe takes an optional property name you want to monitor right?~~

edit: not needed if removing prop in other comment

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -163,14 +174,27 @@ export class ColumnInternals<TColumnConfig> {
...
+    public constructor(options: ColumnInternalsOptions<TColumnValidator>) {
         this.cellRecordFieldNames = options.cellRecordFieldNames;
         this.cellViewTemplate = createCellViewTemplate(options.cellViewTag);
         this.groupHeaderViewTemplate = createGroupHeaderViewTemplate(
             options.groupHeaderViewTag
         );
         this.delegatedEvents = options.delegatedEvents;
         this.sortOperation = options.sortOperation ?? TableColumnSortOperation.basic;
+        this.validator = options.validator;
+
```
rajsite, line 188: ~~It's possible that the provided validator is initialized to be invalid I guess. Should we use its configured value?~~

```suggestion
this.validConfiguration = this.validator.isColumnValid;
```

~~Then we can remove the initialization since it would be set in constructor~~
```
@observable
public validConfiguration;
```
edit: not needed if removing prop in other comment

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -163,14 +174,27 @@ export class ColumnInternals<TColumnConfig> {
...
+        const notifier = Observable.getNotifier(this.validator);
+        notifier.subscribe(this);
+    }
+
+    /** @internal */
+    public handleChange(source: unknown, args: unknown): void {
+        if (source instanceof ColumnValidator && args === 'isColumnValid') {
+            this.validConfiguration = this.validator.isColumnValid;
+        }
     }
```
rajsite, line 198: Thinking about it again, this indirection seems unnecessary.  The columnIInternals.validConfiguration property was the observable property representing the validator state. Now the validator has its own observable property, so I don't see why columnInternals needs the observable validConfiguration property anymore. Things should just check columnInternals.validator.isColumnValid's observable prop.

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -10,8 +10,11 @@ import {
 import type { TableGroupRow } from '../../../table/components/group-row';
 import { createGroupHeaderViewTemplate } from '../group-header-view/template';
 import { createCellViewTemplate } from '../cell-view/template';
+import { ColumnValidator } from './column-validator';
 
-export interface ColumnInternalsOptions {
+export interface ColumnInternalsOptions<
```
rajsite, line 15: MentionInternals is following a similar pattern to columnInternals. In a follow-up PR the updated validator patterns should try and align there as well (or be captured in a tech debt if for some reason that is not relatively straightforward).

mollykreis: I'll take a look.

rajsite (approved):

PR #2003, opened 2024-04-09
Move validator into MentionInternals

mollykreis:
# Pull Request

## 🤨 Rationale

This is follow-up work from #1999 because we decided that we should make a similar change to how/when the validator is created for MentionsInternals.

## 👩‍💻 Implementation

- Include typed `RichTextMentionValidator` in `MentionInternalsOptions`
- Update `MentionInternals` to hold the typed reference to the validator passed in the options
- Updates to various components based on the above changes. This includes:
    - Update `RichTextMentionValidator` to no longer require a reference to `MentionInternals`
    - Remove `createValidator` from the `RichTextMention` base class
    - Update usages of validator

## 🧪 Testing

Ran unit tests

## ✅ Checklist


mollykreis:

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
```diff
@@ -22,10 +22,10 @@ export const baseValidityFlagNames = [
 export class RichTextMentionValidator<
     ValidityFlagNames extends readonly string[] = typeof baseValidityFlagNames
 > extends Validator<typeof baseValidityFlagNames | ValidityFlagNames> {
-    public constructor(
-        private readonly mentionInternals: MentionInternals,
-        configValidityKeys: ValidityFlagNames
-    ) {
+    @observable
+    public isRichTextMentionValid = true;
```
m-akinc, line 26: What is `isRichTextMentionValid` for (given we already have `isValid()`)? Do we need this as an observable?

m-akinc (approved):

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
mollykreis: That's a good point. I had added this here as a replacement to the `validConfiguration` observable in `MentionInternals`, but I never found anything that was listening to changes to that observable that needed to be updated. Do you know why `validConfiguration` was needed on `MentionInternals`?

packages/nimble-components/src/rich-text-mention/base/index.ts
```diff
@@ -133,7 +131,10 @@ export abstract class RichTextMention<
     }
 
     private patternChanged(): void {
-        this.validator.validate(this.mappingElements, this.pattern);
+        this.mentionInternals.validator.validate(
```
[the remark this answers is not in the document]
mollykreis: I'm not familiar with the rich-text code, so I'm not sure what the expectation here is. I specifically was trying not to change any functionality outside of the typing & ownership of the validator.

packages/nimble-components/src/rich-text-mention/base/models/mention-validator.ts
mollykreis: Removed.

packages/nimble-components/src/rich-text-mention/base/index.ts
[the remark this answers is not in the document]
rajsite: Tried tracing it a little bit and also not sure what to expect. My hunch is they shouldn't be coupled that way. Changing mapping configuration shouldn't cause pattern to be undefined, pattern just needs to be a parseable regex.

If there is concern can create a tech debt item (even better if can create incorrect behavior and treat as a bug) but either way inclined to not change that logic in this PR.

rajsite (approved):

jattasNI (approved):

Issue #2002, opened 2024-04-09
Filterable Select allows user to select option that is not visible in filtered list


atmgrifter00:
# 🐛 Bug Report

When using the `Select` with its `filterMode` set to `standard`, then after providing filter text, and filtering the existing set of options, if the last or first visible option isn't the actual last or first option respectively, then if a user presses the arrow up or down keys beyond what is visible, and presses `<Enter>`, a non-visible option will be selected.

## 💻 Repro or Code Sample

Just use the [Storybook `Select`](https://nimble.ni.dev/storybook/?path=/docs/components-select--docs) page to demonstrate the issue. After turning filtering on, click on the `Select` press '3', press `<ArrowUp>` and then `<Enter>`. You will see 'Option 1' gets selected.

## 🤔 Expected Behavior

After reaching the beginning or end of the filtered set of options in the dropdown when filtering is turned on, a user can not move selection past the selected item.

## 😯 Current Behavior

See above.

## 💁 Possible Solution

The issue was simply that our overridden implementations of `selectNextOption` and `selectPreviousOption` weren't considering the current filtered set of options. The solution is to make them do so.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #2005, opened 2024-04-09
Filterable Select arrow select fix

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #2002

## 👩‍💻 Implementation

We just needed a more robust implementation for the overrides of `selecteNextOption` and `selectPreviousOption` to consider only the visible options.

## 🧪 Testing

Added unit tests. Tested in Storybook.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/select/index.ts
+++ b/packages/nimble-components/src/select/index.ts
@@ -57,6 +57,10 @@ const isNimbleListOption = (el: Element): el is ListOption => {
     return el instanceof ListOption;
 };
 
+const isOptionSelectable = (el: ListOption): boolean => {
+    return !el.visuallyHidden && !el.disabled && !el.hidden;
+};
+
 /**
  * A nimble-styled HTML select.
  */
@@ -193,8 +197,6 @@ export class Select
     private _value = '';
     private forcedPosition = false;
     private indexWhenOpened?: number;
-    private firstVisibleOptionIndex?: number;
-    private lastVisibleOptionIndex?: number;
 
     /**
      * @internal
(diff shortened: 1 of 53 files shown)


packages/nimble-components/src/select/index.ts
```diff
@@ -665,7 +671,11 @@ export class Select
     public override selectPreviousOption(): void {
         // don't call super.selectPreviousOption as that relies on side-effecty
         // behavior to not select disabled option (which no longer works)
-        for (let i = this.selectedIndex - 1; i >= 0; i--) {
+        for (
+            let i = this.selectedIndex - 1;
+            i >= (this.firstVisibleOptionIndex ?? 0);
+            i--
+        ) {
             if (!this.options[i]?.disabled) {
```
rajsite, line 673: the issue is that disabled doesn't mean filtered anymore. Could we instead check for disabled || visually hidden?

rajsite (approved):

packages/nimble-components/src/select/index.ts
```diff
@@ -655,7 +659,8 @@ export class Select
         // don't call super.selectNextOption as that relies on side-effecty
         // behavior to not select disabled option (which no longer works)
         for (let i = this.selectedIndex + 1; i < this.options.length; i++) {
-            if (!this.options[i]?.disabled) {
+            const listOption = this.options[i]!;
+            if (isNimbleListOption(listOption) && isOptionSelectable(listOption)) {
                 this.selectedIndex = i;
```
jattasNI, line 664: There are several other places in the file where we programmatically set `this.selectedIndex = ...`. Could you scan them and see if there are other bugs lurking that should also call `isOptionSelectable()`?

I suspect they're mostly extreme corner cases like option attributes changing while the drop down is open, but worth a check.

jattasNI (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: The only other place that seemed possibly relevant is where we determine the initial default selected option. The new behavior is that we will only select a default option that isn't hidden (or `visually-hidden`, but that should never be possible, but it's not a harmful aspect of the general check).

PR #2006, opened 2024-04-09
Fix and re-enable intermittent mention-listbox test

### LINKED ISSUE #1891: Intermittent test: mention-listbox "should scroll the selected option into view when opened"

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #1891 

It's not clear why, but the mention-listbox test was awaiting `waitForUpdatesAsync` twice before asserting that the scroll position was correct. Other similar tests await `waitForUpdatesAsync` followed by `waitAnimationFrame`. The call to `waitAnimationFrame` is necessary because the scroll is queued with `requestAnimationFrame`.

## 👩‍💻 Implementation

Update the second call to `waitForUpdatesAsync` to a call to `waitAnimationFrame`.

## 🧪 Testing

Ran the test a number of times locally, though I'm not if this verifies much since I was never able to get the test to fail locally before this fix

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-bb8ab8b7-633d-4a78-9d87-e9ec1263eaff.json
+++ b/change/@ni-nimble-components-bb8ab8b7-633d-4a78-9d87-e9ec1263eaff.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "none",
   "comment": "Fix and re-enable intermittent mention-listbox test",
   "packageName": "@ni/nimble-components",
   "email": "20542556+mollykreis@users.noreply.github.com",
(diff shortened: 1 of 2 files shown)


mollykreis:

m-akinc (approved):
Fingers crossed

change/@ni-nimble-components-bb8ab8b7-633d-4a78-9d87-e9ec1263eaff.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
m-akinc, line 2: Should be `"none"`, right?

mollykreis:
```suggestion
  "type": "none",
```

rajsite (approved):

PR #2012, opened 2024-04-11
Add configuration for icon mappings and spinner mappings to show/hide their text within a table's cells

mollykreis:
# Pull Request

## 🤨 Rationale

This updates the `nimble-mapping-icon` and `nimble-mapping-spinner` elements to be aligned with the changes described in #1992.

## 👩‍💻 Implementation

- Added `text-hidden` to `nimble-mapping-icon` and `nimble-mapping-spinner`
    - I also added these to Blazor and Angular so that when the change is published clients in any framework can uptake the new package and have the opportunity to add `text-hidden` to mappings in order to keep their existing behavior
- Reworked the `MappingIconConfig` and `MappingSpinnerConfig` to be more aligned
    - They both create the appropriate `ViewTemplate` to render a spinner, icon, or blank
    - They both add the appropriate `title` and `aria-*` attributes based on the configuration
    - Neither adds css classes that are styled in the group header styles or cell view styles
- Update the templates for the icon column's cell view and group header share code for rendering icons and spinners (enabled because of aligning `MappingIconConfig` and `MappingSpinnerConfig`) and to render the text based on the `text-hidden` attribute

## 🧪 Testing

- Updated storybook & matrix tests
- Manually tested
- Wrote/updated a number of tests for the icon column
    - To enable this, created a page object for the icon column to encapsulate the logic around finding the icon/spinner and the text

## ✅ Checklist


mollykreis:

atmgrifter00 (approved):

msmithNI (approved):

packages/nimble-blazor/NimbleBlazor/Components/NimbleMappingIcon.razor.cs
```diff
@@ -21,4 +21,10 @@ public partial class NimbleMappingIcon<TKey> : NimbleMappingBase<TKey>
     /// </summary>
     [Parameter]
     public string? Text { get; set; }
+
+    /// <summary>
+    /// Gets or sets whether or not the mapping's text should be rendered in the table's cells.
+    /// </summary>
+    [Parameter]
+    public bool TextHidden { get; set; }
```
msmithNI, line 29: I think for consistency with our existing Blazor API, this should be nullable (`bool?`) - same for the property in `NimbleMappingSpinner.razor.cs`.

(I actually thought as-is this would require the property to always be set explicitly, but that doesn't look like the case, since ComponentsDemo has icon mappings without `TextHidden` specified.)

packages/nimble-components/src/table-column/enum-base/models/mapping-spinner-config.ts
```diff
@@ -1,10 +1,36 @@
+import { type ViewTemplate, html } from '@microsoft/fast-element';
 import { MappingConfig } from './mapping-config';
+import type { IconView } from './mapping-icon-config';
+import { spinnerTag } from '../../../spinner';
+
+export interface SpinnerView {
+    text?: string;
+}
+
+const createSpinnerTemplate = (textHidden: boolean): ViewTemplate<IconView> => {
```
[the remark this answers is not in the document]
mollykreis: The template created here (which parallels the template created for icons in `mapping-spinner-config.ts`) is just for the spinner. It will be used as part of the group-header-view's template or as part of the cell-view's template, and those templates are what control whether or not the text is rendered next to the icon/spinner. This template just makes sure the accessibility (i.e. tooltip, aria-hidden, aria-label) are correct for the visualization part of the view.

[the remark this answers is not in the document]
jattasNI: Ah, gotcha! I had completely missed that case was present in the matrix test. Looks good 👍

change/@ni-nimble-components-9b50b032-00a9-4a25-9089-308c223068ad.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
jattasNI: Now that I understand the state of the spinner better I realized you do have 2 already in the spinner mapping. If it's not too messy I think it would help to have both spinner and icon examples with both text hidden and not hidden. That's getting a bit like a matrix test which isn't ideal, but I think it's useful since we can't have controls on the mapping args table.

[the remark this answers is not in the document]
mollykreis: > If it's not too messy I think it would help to have both spinner and icon examples with both text hidden and not hidden.

Technically, the "Unknown" status of Abbey has `text-hidden` set on it. Though, I can see why that wouldn't be extremely clear since it also doesn't use an icon. I'm open to suggestions on how to make the storybook example clearer, but I'm concerned about doing too much to show every combination of configurations since I don't think we want it to deviate too far from whatever our best practices will end up being.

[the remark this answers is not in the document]
jattasNI: That's fair, I guess I was optimizing for what I want as a reviewer, not what would be most useful to a client. I agree that the latter is more important for the docs in the mdx. For the combinations that aren't best practices we could either have a separate client-visible story that demos them (but doesn't appear in the mdx) or reviewers can see them in the matrix story.

I would be fine to leave this as-is and revisit it when we've figured out our best practices.

jattasNI (approved):

packages/nimble-components/src/table-column/enum-base/models/mapping-spinner-config.ts
```diff
@@ -1,10 +1,36 @@
...
+    return html`
+        <${spinnerTag}
+            title="${x => (textHidden ? x.text : '')}"
+            aria-label="${x => x.text}"
+            aria-hidden="${_ => (textHidden ? 'false' : 'true')}"
+        >
+        </${spinnerTag}>
+    `;
```
rajsite, line 18: The goal should be creating as few of these as possible and instead relying on binding updates when we can. Is it possible to have textHidden be part of the IconView/SpinnerView state and place it on the config?

Then this template can be static instead of newly constructed per config, i.e.:

```
html`
        <${spinnerTag}
            title="${x => (x.textHidden ? x.text : '')}"
            aria-label="${x => x.text}"
            aria-hidden="${_ => (xtextHidden ? 'false' : 'true')}"
        >
        </${spinnerTag}>
    `;
```

packages/nimble-components/src/table-column/enum-base/models/mapping-icon-config.ts
```diff
@@ -13,19 +13,20 @@ export interface IconView {
...
+    textHidden: boolean
 ): ViewTemplate<IconView> => {
     if (icon === undefined) {
         return emptyTemplate;
     }
 
     return html`
         <${icon}
-            title="${x => x.text}"
+            title="${x => (textHidden ? x.text : '')}"
```
rajsite, line 25: same as the spinner we should move towards fewer unique instances instead of further. If we follow a similar patterns as the spinner and make x.textHidden part of the config we at least remove one dimension of uniqueTemplate construction. Longer-term we may want to cache the template instances per icon (but so far has seemed like an unneeded optimization).

packages/nimble-components/src/table-column/enum-base/models/mapping-spinner-config.ts
rajsite: Looks like textHidden is already on the cell views so shouldn't be too bad?

packages/nimble-components/src/table-column/enum-base/models/mapping-spinner-config.ts
rajsite, line 10:
```suggestion
const createSpinnerTemplate = (textHidden: boolean): ViewTemplate<SpinnerView> => {
```

packages/nimble-components/src/table-column/icon/cell-view/index.ts
```diff
@@ -27,19 +31,28 @@ export class TableColumnIconCellView
...
     public text?: string;
 
     @observable
-    public iconTemplate?: ViewTemplate<IconView>;
+    public visualizationTemplate?:
+    | ViewTemplate<IconView>
+    | ViewTemplate<SpinnerView>;
 
     @observable
     public visual?: 'spinner' | 'icon';
```
rajsite, line 47: Is visual still useful? can just check if `visualizationTemplate` is defined in the template where this was used?
Think it applies to cell-view and group-header-view

rajsite (approved):

change/@ni-nimble-angular-91d3afae-1d79-413d-9ca3-c0cd87b52ec7.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: Note: We should wait to merge this PR until we're ready to uptake the breaking change in SLE.

rajsite: Scanned the discussions from the spec update and don't think it came up but would we avoid a breaking change if we introduced `text-visible` instead of `text-hidden`?
If our usage guidelines are generally that the text should be hidden in an icon column and only shown for special cases (ambiguous icons or multiple mappings to the same icon) then maybe it would encourage good defaults and prevent a breaking change (I'm not sure what our usage guidance will be).

mollykreis: The breaking change didn't end up being a problem to uptake. Angular's strict template checking didn't mind having `text-hidden` on the mapping elements even though the directive change hasn't been merged yet. (If it were a non-boolean value where I had to set `text-hidden="foo"`, then it would've been a problem.)

With that knowledge, are you still in favor of potentially changing from `text-hidden` to `text-visible`? Should I try to get more clarity on our usage guidance to influence this discussion?

rajsite: I like good defaults and my hunch is that would be a better default but wouldn't hold anything up for it.

rajsite: Actually maybe not even sure about my hunch, might not even be worth the effort to rename.

mollykreis: @fredvisser - Do you have an idea of what the best practices will be for putting text next to the icons? Do you think the recommendations will make it more common to have text in the column with the icon or to have just the icon/spinner?

fredvisser: Yeah, our current plan is to default to `icon + text` for most columns. After standardizing on the icons and text we use across SLE apps, we should have a solid majority of icon columns configured this way.

mollykreis: Thanks. I'll leave this with `text-hidden` as currently implemented.

PR #2020, opened 2024-04-17
Move eslint config into `eslint-config-nimble`

m-akinc:
# Pull Request

## 🤨 Rationale

In preparation for adding the `spright-components` package, I'm moving the ESLint configuration into a private plugin `@ni-private/eslint-config-nimble`. This will allow easy sharing of the configuration.

## 👩‍💻 Implementation

- moved the configuration from `nimble-components/.eslintrc.js` and `nimble-components/src/.eslintrc.js` into two configs (`javascript` and `typescript`, respectively) in a new `eslint-config-nimble` package
- updated `nimble-components` to extend the `eslint-config-nimble` configs
- added missing dev dependencies for the `site` and `performance` packages

## 🧪 Testing

Spot checked locally.

## ✅ Checklist


packages/eslint-plugin-nimble/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
jattasNI, line 1: Could you add a couple-sentence README? Basically just your existing description below plus mention that it is built on the NI style guide (and link to that repo).

package-lock.json
```diff
@@ -4195,7 +4195,6 @@
       "version": "4.4.0",
       "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.4.0.tgz",
       "integrity": "sha512-1/sA4dwrzBAyeUoQ6oxahHKmrZvsnLCg4RfxW3ZFGGmQkSNQPFNLV9CUEFQP1x9EYXHTo5p6xdhZM1Ne9p/AfA==",
```
jattasNI, line 4197: A bunch of packages, including eslint-related ones, are changing from being dev dependencies to non-dev. That seems wrong on the surface but I'm not immediately sure the fix.

Maybe the eslint peerDependency of the new package is being satisfied automatically for existing packages by installing it as a non-dev dependency? If that's the issue then one possible fix could be explicitly listing eslint as a devDependency in existing packages. But I'm not sure if that's aligned with best practices (we don't explicitly depend on eslint in many of our NI package.jsons). Or maybe we need to keep the NI styleguide listed as a devDependency?

Or maybe it's not a problem?

Could you explore docs / other repos / impacts of different package configurations? @rajsite might have opinions or ideas too.

packages/eslint-plugin-nimble/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "@ni-private/eslint-plugin-nimble",
+  "version": "0.0.1",
```
jattasNI, line 3: Do you or @rajsite know how versioning will work for local private packages? I guess beachball won't be involved since it isn't published anywhere? So do we ever need to change this version number or the number listed in packages that depend on this?

packages/performance/.eslintrc.cjs
```diff
@@ -1,18 +1,6 @@
...
-        'node_modules',
-        'dist'
-    ],
-    rules: {
-        // Enabled to prevent accidental usage of async-await
-        'require-await': 'error'
-    },
+    extends: 'plugin:@ni-private/nimble/javascript',
     overrides: [{
         files: ['*.ts'],
```
jattasNI, line 5: Could you share the thought process around sharing the JS part of this config but not the TS part?

jattasNI (commented):
Wow, quick turnaround on turning this idea into a PR! ⏩

packages/eslint-plugin-nimble/package.json
rajsite: Yea doesn't change, just needs to be defined. We mark our other private package "version": "1.0.0" so we should do the same

```suggestion
  "version": "1.0.0",
```

packages/eslint-plugin-nimble/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "private": true,
+  "description": "Common ESLint configuration for the Nimble repository",
```
rajsite, line 5: We should add an explicit `"type": "commonjs"` as that is what this package currently is.
At least I believe eslint doesn't support plugins that are type module.

packages/eslint-plugin-nimble/index.js
```diff
@@ -0,0 +1,203 @@
+const restrictedImportsPaths = () => [
```
rajsite, line 1: Each package should have a README.md

```
# Eslint Plugin Nimble

## Overview

Shared eslint configuration for FAST web component development.
```

packages/eslint-plugin-nimble/index.js
```diff
@@ -0,0 +1,203 @@
...
+    {
```
rajsite, line 2: Can we align more with the patterns shown in the https://github.com/ni/javascript-styleguide
i.e. instead of creating a plugin, create a shareable config (we aren't trying to make new rules, just configure them): https://eslint.org/docs/latest/extend/shareable-configs

Package name: `@ni-private/eslint-config-nimble`

```
/README.js
/package.json
/javascript.js
/typescript.js
```

packages/nimble-components/src/.eslintrc.js
```diff
@@ -1,168 +1,8 @@
...
-];
-
 module.exports = {
     root: true,
-    plugins: ['jsdoc'],
-    extends: [
-        '@ni/eslint-config-typescript',
-        '@ni/eslint-config-typescript/requiring-type-checking'
-    ],
+    extends: 'plugin:@ni-private/nimble/typescript',
```
rajsite, line 3: this will end up looking more like a normal configuration override / like using the ni shared configs instead of using a plugin

```suggestion
    extends: '@ni-private/eslint-config-nimble/typescript',
```

packages/performance/.eslintrc.cjs
rajsite: I don't think this file should be updated. The performance package is not authoring nimble components iin FAST so the normal '@ni/eslint-config-javascript' was sufficient.

packages/performance/package.json
```diff
@@ -1,26 +1,27 @@
...
-    "devDependencies": {
-        "@lhci/cli": "^0.13.0",
-        "typescript": "~5.1.6",
-        "vite": "^5.1.5"
-    },
-    "files": [
-        "dist",
-        "!*.tgz"
-    ]
+  "name": "@ni-private/performance",
```
rajsite, line 2: can you avoid the whitespace change in this file? We normally do 4 space indents so this doesn't look right switching to 2 space

package-lock.json
rajsite: Don't think its a problem that the lockfile expresses it this way. The best choices today are:
- the eslint shared configuration package mark them as peerDependency
- users of the shared config in the monorepo mark it as devDependency (although I really want npm to make a devPeerDependency concept to support the monorepo use-case. All eslint configs should be "peer" for npm installl in the monorepo at dev but should not be a production peer).

packages/performance/package.json
```diff
@@ -1,26 +1,27 @@
...
+    "lint": "eslint .",
+    "start": "vite src",
+    "performance": "lhci autorun"
+  },
+  "dependencies": {
+    "@ni/nimble-components": "*"
+  },
+  "devDependencies": {
+    "@lhci/cli": "^0.13.0",
+    "@ni-private/eslint-plugin-nimble": "^0.0.1",
```
rajsite, line 19: this would change to @ni/eslint-config-javascript

packages/performance/package.json
m-akinc: Actually, all of our package.json files use two-space indentations. I intentionally changed this one to match.

packages/eslint-plugin-nimble/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "repository": {
```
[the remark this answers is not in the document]
m-akinc: `package.json` is always included by default, and none of our other `package.json` include this self reference. `index.js` is also auto-included because of it being the default for the `main` module name. Why add these explicitly?

> Edit: in the other comment I suggest different files being exported so this should reflect that

I don't see another comment mentioning exported files.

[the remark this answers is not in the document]
rajsite: fair, double-checked and I was mixing up the `files` prop and the [`exports`](https://nodejs.org/api/packages.html#package-entry-points) prop where `package.json` has to be mentioned manually.

It's probably a good idea to add an exports as well (`files` is what is included in the tgz and `exports` is what is allowed to be resolved from the package when imported).

packages/performance/package.json
rajsite: Oh interesting, maybe npm is doing that when it updates them? If it's matching the others that works 👍

packages/eslint-config-nimble/typescript.js
```diff
@@ -0,0 +1,163 @@
...
+            'error',
+            {
+                paths: restrictedImportsPaths()
+            }
+        ],
+
+        // Enabled to prevent accidental usage of async-await
+        '@typescript-eslint/require-await': 'error'
+    },
+    ignorePatterns: ['.eslintrc.js'],
```
rajsite, line 66: huh, noticing it now but realize it was that way before. seems like the ignore patterns from the js should be here too, i.e. node_modules and dist:

```
    ignorePatterns: [
        '.eslintrc.js,
        'node_modules',
        'dist'
    ],
```

I don't actually know why `.eslintrc.js` is being ignored here, it can be formatted following our style rules too right?

m-akinc: The reason we didn't ignore `node_modules` and `dist` is because this config was used within the `src` directory, so neither of those directories would be present. But since this config no longer assumes a context of the `src` directory, we can add those ignores. I'm also removing the ignore for `.eslintrc.js`. I don't know why we would need to exclude it.

m-akinc: Turns out we were excluding `.eslintrc.js` for a reason. Without it, we get the following lint error:
> /home/runner/work/nimble/nimble/packages/nimble-components/src/.eslintrc.js
  0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/.eslintrc.js` using `parserOptions.project`: /home/runner/work/nimble/nimble/packages/nimble-components/tsconfig.json
However, that TSConfig does not include this file. Either:
> - Change ESLint's list of included files to not include this file
> - Change that TSConfig to include this file
> - Create a new TSConfig that includes this file and include it in your parserOptions.project
>
>See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file

I tried adding `".eslintrc.js"` to the `"include"` of `nimble-components/tsconfig.json` (as that seems like what the second suggestion is saying), but that didn't work. I'm just going back to having the `.eslintrc.js` exclude itself.

m-akinc:
I'm going to submit without @jattasNI's approval so that I can build on top of this. If he has any additional comments or concerns, I'll be happy to address them in a follow up.

PR #2025, opened 2024-04-19
Don't allow use of the nullish coalescing operator in templates

### LINKED ISSUE #1843: Evaluate volatile template bindings
## 🧹 Tech Debt
We need to evaluate our template bindings for conditional statements that need to be made into `volatile` getters.

One instance of this was found as a bug in the table and fixed (see #1839), but we need to do a more thorough evaluation of our existing bindings to see if any other ones need to be modified.

If there are only a handful of issues to be fix, this issue can cover fixing them. If there are many, we could split this into separate issues and tackle them over time.

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1843 

I did an evaluation of how observables work within FAST templates. A longer explanation of what is and is not allowed can be found [in the discussion within the linked issue](https://github.com/ni/nimble/issues/1843#issuecomment-2064663561). The only issue I found within nimble code was the usage of `??` in templates, which may not correctly be identified as a volatile binding. Therefore, I've rewritten the bindings that previously used `??` and added a lint rule for `template.ts` files that disallows usage of `??`.

## 👩‍💻 Implementation

See Rationale

## 🧪 Testing

All existing tests still pass
Verified that the new lint rule catches instances of `??` in `template.ts` files

## ✅ Checklist


packages/nimble-components/src/table/components/cell/template.ts
```diff
@@ -20,10 +20,10 @@ export const template = html<TableCell>`
                 @toggle="${(x, c) => x.onActionMenuToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>)}"
                 @click="${(_, c) => c.event.stopPropagation()}"
                 class="action-menu"
-                title="${x => x.actionMenuLabel ?? tableCellActionMenuLabel.getValueFor(x)}"
+                title="${x => (x.actionMenuLabel ? x.actionMenuLabel : tableCellActionMenuLabel.getValueFor(x))}"
```
rajsite, line 23: Is this an expected change in behavior? Nullish checks just undefined / null but truthiness also checks empty string. Might be a good change, dunno why one would want an empty string label..

packages/eslint-config-nimble/typescript.js
```diff
@@ -118,6 +118,15 @@ module.exports = {
                 '@typescript-eslint/indent': 'off'
             }
         },
+        {
+            files: ['template.ts'],
+            rules: {
+                'no-restricted-syntax': [
+                    'error',
+                    { selector: "LogicalExpression[operator='??']" }
```
rajsite, line 126: Can you comment and link to the issue for context on why this rule exists?

rajsite (approved):

change/@ni-nimble-components-0911f623-6d16-4c80-9b6d-0757470b2ca7.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: We probably need some docs for aggregate bindings not being allowed to use any conditional logic right?

packages/nimble-components/src/table/components/cell/template.ts
mollykreis: It is an intentional behavior change, but I can also write this a different way to avoid the behavior change (it would just be more verbose). I didn't see a reason why someone should create an empty label to override the default label, so I opted for making the logic simpler.

change/@ni-nimble-components-0911f623-6d16-4c80-9b6d-0757470b2ca7.json
mollykreis: I don't mind adding docs. I hadn't because @jattasNI [commented on the issue](https://github.com/ni/nimble/issues/1843#issuecomment-2065090524) saying that he wasn't sure they were needed. Do you think we should add something to the contributing docs about this?

jattasNI (approved):

packages/nimble-components/src/table/components/cell/template.ts
rajsite: makes sense! 🚢-it

change/@ni-nimble-components-0911f623-6d16-4c80-9b6d-0757470b2ca7.json
rajsite: Yea guess it's weird to document. I think both of those behaviors could reasonably considered bugs.
I at least captured it in the PR review common issues, guess we can just do that for now: https://github.com/ni/nimble/issues/1479#issuecomment-2066815162

Issue #824, opened 2022-11-15
Storybook property name inconsistencies


rajsite:
# 🐛 Bug Report

The property name documentation in storybook is inconsistent. For example see how in the following screenshot how some properties are kebab-case and some are camelCase.

## 💻 Repro or Code Sample

![image](https://user-images.githubusercontent.com/1588923/202005653-85e7d317-3266-43c0-bd00-c405f72a5830.png)

## 🤔 Expected Behavior

The storybook documentation should be consistent and make it clear when what is the difference between attributes, properties, events, slot configuration, generic storybook controls

## 😯 Current Behavior

See repro above.

## 💁 Possible Solution

 - By default show the attribute names and values, ie `content-hidden`
   - Alternative 1: Show attribute names with the CSS selection syntax to emphasize it is an attribute `[content-hidden]`
 - For properties that do not have corresponding attribute, prepend the name with a period `.checked`
   - Alternative 1: Use the prototype syntax common on mdn to emphasize it is a property `Switch.prototype.checked`
 - For other things that require a discussion in the description, us the format `Topic: Start Icon`
   - Alternatives, `Discussion: Start Icon`, wrap in quotes `"Start Icon"`, just rely on sentence case to distinguish from attributes and properties `Start Icon`

## 🔦 Context

Current documentation can lead to confusion as discussed in this [teams thread](https://teams.microsoft.com/l/message/19:wo8vmMKMsHfltKXxc0bczZo-X4JlQSV5VxpaRJdh13k1@thread.tacv2/1668457167619?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=9ee065d7-3898-4245-82f6-76e86084b8b1&parentMessageId=1668457167619&teamName=Design%20System&channelName=General&createdTime=1668457167619&allowXTenantAccess=false).

## 🌍 Your Environment

N/a

rajsite:
- Looked at the following custom element framework docs (@nate-ni are there other examples worth comparing to?):
  - FAST: https://explore.fast.design/components/fast-number-field
    - cons: story names do not align with property names, does not have additional descriptions
  - Shoelace: https://shoelace.style/components/alert?id=properties
    - pros: clearly separates properties, methods, events, slots
    - cons: dedicated docs page / not a storybook pattern, cannot directly play with a property to see the effect
  -  Carbon:
     - docs page: https://carbondesignsystem.com/components/number-input/code/ that links to storybook for code: 
[...]

PR #2027, opened 2024-04-19
Improve API docs for menu button component

jattasNI:
# Pull Request

## 🤨 Rationale

As part of #824 we want to improve the accuracy and comprehensiveness of our storybook API documentation.

## 👩‍💻 Implementation

Apply the following patterns to the menu button docs:

1. create sections in the args table for attributes, slots, and events. (We'd also add methods for components that expose interesting ones)
2. Ensure the title and description are accurate for each member
3. link to shared docs about component APIs. 

Other minor bug fixes:
1. hide the "patterns" page from public storybook
2. hide text from ✅ icon mapping in component status table

### Possible follow up work

1. update doc templates and the docs for other existing components similarly
1. more detailed API overview docs
    - explanation of event patterns in web components and how they map to frameworks
    - using forms instead of change events
1. see if we can remove the type annotation from control descriptions
2. [wait for further feedback] more curated examples and highlight most important APIs for each component 
1. [out of scope] I wish I could hide the Default column since we don't bother populating it but that doesn't seem to be possible https://github.com/storybookjs/storybook/issues/15780

## 🧪 Testing

Manual storybook interaction

## ✅ Checklist


packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +47,69 @@ const metadata: Meta<MenuButtonArgs> = {
...
         },
         menuPosition: {
+            name: 'menu-position',
+            description: 'The position of the menu relative to the button.',
             options: Object.values(MenuButtonPosition),
-            control: { type: 'radio' }
+            control: { type: 'radio' },
+            table: { category: 'attributes' }
+        },
+        toggle: {
```
rajsite, line 105: When you hover over the control area for these it shows "setup control".
image

Is that avoided when setting `control: false`? 
https://storybook.js.org/docs/essentials/controls#disable-controls-for-specific-properties

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +47,69 @@ const metadata: Meta<MenuButtonArgs> = {
...
+        endIcon: {
+            name: 'end',
+            description: 'Set \`slot="end"\` to include an icon after the text content. This icon will be hidden when \`content-hidden\` is set.',
+            table: { category: 'slots' }
+        },
+        menu: {
+            description: 'The `nimble-menu` to be displayed when the button is toggled.',
+            table: { category: 'slots' }
+        },
         appearance: {
```
rajsite, line 70: maybe in the global config we should sort alphabetically?
https://storybook.js.org/docs/essentials/controls#sorting-controls

packages/nimble-components/.storybook/manager.js
```diff
@@ -7,9 +7,10 @@ addons.setConfig({
         filters: {
```
rajsite, line 7: Wanna add a change in this PR to mark the ready key label-hidden in the component status table?

image

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +47,72 @@ const metadata: Meta<MenuButtonArgs> = {
...
     argTypes: {
+        label: {
+            name: 'default',
+            description: 'The text content of the button.',
+            table: { category: 'slots' }
+        },
+        icon: {
+            name: 'start',
+            description:
+                'Set `slot="start"` to include an icon before the text content. This icon will be hidden when `content-hidden` is set.',
```
mollykreis, line 58: > This icon will be hidden when `content-hidden` is set.

That isn't true. In fact, the `start` slot is the only thing that isn't hidden when `content-hidden` is set. Instead, this statement should probably be applied to the `default` slot of the button -- perhaps with a note that says that the default slot should still have content when `content-hidden` is set because it makes the button accessible with screen readers.

jattasNI: Good catch, fixed by leveraging the shared doc strings from the pattern (and tweaking those strings)

packages/nimble-components/.storybook/manager.js
```diff
@@ -7,9 +7,10 @@ addons.setConfig({
...
             patterns: item => {
```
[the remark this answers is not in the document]
jattasNI: I spent some time scanning docs for how to remove this but haven't found anything yet. I'll keep looking but since this is pre-existing (and I'm OOO soon) I'll propose not blocking the PR on it.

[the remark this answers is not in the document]
rajsite: Sure, we could also skip tracking it until someone notices and brings it up. Seems minor

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +50,73 @@ const metadata: Meta<MenuButtonArgs> = {
         }
     },
     argTypes: {
+        label: {
+            name: 'default',
+            description:
+                'The text content of the button. This will be hidden when `content-hidden` is set but should always be provided; see [Accessibility](#accessibility) for more info.',
```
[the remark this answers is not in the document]
jattasNI: Hopefully this works...between dev builds, PR builds and the deployed storybook multiplied by MDX files, stories, MD files and shared snippets, I feel like the right way to link is different everywhere.

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +51,73 @@ const metadata: Meta<MenuButtonArgs> = {
         }
     },
     argTypes: {
+        label: {
+            name: 'default',
+            description:
+                'The text content of the button. This will be hidden when `content-hidden` is set but should always be provided; see [Accessibility](./?path=/docs/components-menu-button--docs#accessibility) for more info.',
```
rajsite, line 57: I dunno if you see something different but in Firefox and Chrome clicking this link isn't working for me. It updates the URL but does not scroll. If I refresh the page it scrolls to the right spot.

Maybe for same page references we can just say See `Accessibility` section below?

```suggestion
                'The text content of the button. This will be hidden when `content-hidden` is set but should always be provided; see the `Accessibility` section for more info.',
```

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
```diff
@@ -49,27 +51,73 @@ const metadata: Meta<MenuButtonArgs> = {
...
+            table: { category: argsTableCategory.slots }
+        },
+        endIcon: {
+            name: 'end',
+            description: endIconDescription,
+            table: { category: argsTableCategory.slots }
+        },
+        menu: {
+            description:
+                'The [nimble-menu](./?path=/docs/components-menu--docs) to be displayed when the button is toggled.',
```
rajsite, line 72: this link is working well for me in the chromactic.com published storybook in FFX and Chrome 👍

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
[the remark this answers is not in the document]
rajsite: One of the many nightmares of dealing with iframes and in this case reflecting routing between the parent and child :(

packages/nimble-components/src/utilities/tests/storybook.ts
```diff
@@ -161,3 +161,12 @@ export const disableStorybookZoomTransform = `
...
 `;
+
+export const argsTableCategory = {
+    attributes: 'Attributes',
+    events: 'Events',
+    methods: 'Methods',
+    slots: 'Slots'
+} as const;
+export type SpinnerAppearance =
+    (typeof argsTableCategory)[keyof typeof argsTableCategory];
```
rajsite, line 172: What's going on here? Should either remove this bit (notice the `SpinnerAppearance`) definition and just export an object or if we want to represent as an enum we need to move to a types.ts file and fix the naming. I'd probably just go with `ApiCategory` since it's not really about the args table but about the component apis.

rajsite (approved):

packages/nimble-components/src/menu-button/tests/menu-button.stories.ts
jattasNI: I was seeing the same thing but had thought it was maybe just a macOS bug. Changed to match your suggestion.

packages/nimble-components/src/utilities/tests/storybook.ts
jattasNI: Oops, mindless copy-paste bug, I simplified as suggested.

Issue #2022, opened 2024-04-17
Tabbing away from filterable select can cause value to change when it opens a second time


mollykreis:
# 🐛 Bug Report

Tabbing away from the nimble-select while filtering can cause the value of the select to unexpectedly change when the select is opened the next time.

## 💻 Repro or Code Sample

1. Open a filterable select (such as in storybook)
2. Type some text into the filter so that an option other than the currently selected option becomes highlighted in the dropdown
3. Tab away from the component
4. Notice that the select still shows the original value (expected because no new value was committed during the previous interaction)
5. Tab back to the nimble-select
6. Press 'Enter' to open the select
7. Notice the select's value immediately changes to the highlighted option from step 2

## 🤔 Expected Behavior

Entering text in the select's filter and then closing the select should not leave the component in a state that causes the value to change the next time the select opens.

## 😯 Current Behavior

The select gets in a state where the value changes immediately when it becomes open again.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

Issue #1926, opened 2024-03-12
Align rich-text mention listbox implementation behavior


rajsite:
# 🧹 Tech Debt

In discussions with @atmgrifter00 in #1912, as part of preventing arbitrary differences between all of our Foundation ListBox implementations it was found that the implementation in the mention list box:

1. Does not handle filteredOptions in a way that aligns with the select (as filtering is done the list of options is mutated which state-wise seems unexpected)
2. The mention listbox does not handle the synchronous option registration behavior added to select and combobox in #1912.

PR #2029, opened 2024-04-22
Select - no longer update value while navigating dropdown

atmgrifter00:
# Pull Request

## 🤨 Rationale

While in the midst of implementing the [clear select](https://github.com/ni/nimble/issues/1951) feature, I recognized that the fact that we were updating the value while navigating the dropdown was both confusing (since we don't update the display value) and resulting in an awkward implementation for the feature.

## 👩‍💻 Implementation

While we no longer want to update `selectedIndex` (and as a corollary to this, `value`) while a user navigates the dropdown, we still want the appropriate option in the dropdown to appear selected/highlighted. The styling was keying off of the state of the `aria-selected` attribute, which was being updated when the `selected` state of an option changed. Now, as the `selected` state is no longer changing, I've created a new attribute for the `ListOption`, `activeOption`, that the CSS for the `Select` leverages to manage the styling of the options.

To set the new attribute at the appropriate times required providing overrides to the various `select<>Option` (`Next`, `Previous`, `First`, `Last`) methods. In order to accomplish this we have to keep some additional internal state for what the current active index is (the index for which option should currently be highlighted).

It was also [discovered](https://github.com/ni/nimble/pull/2029#discussion_r1575359065) during the implementation of this fix that there was an issue stemming from the typeahead feature that allows a user to select an option (when there is no filter, or when the dropdown is closed) by typing characters while the `Select` has keyboard focus. This required forking some FAST `Listbox` implementations to prohibit updates to the `selectedIndex` while the dropdown was open.

Additionally, since I was already updating related implementation, I was able to tackle [this issue](https://github.com/ni/nimble/issues/2022) as well (where tabbing away from the `Select` could end up with unexpected display).

## 🧪 Testing

Largely this just required updates to tests that used to expect the `selectedIndex`/`value` to change while the dropdown was open and a user performed a navigation within the dropdown. Beyond that I provided new tests for the `<Tab>` behaviors, new tests to verify the typeahead behaviors, and a slight update to the pageObject for reporting what the selected option is (based on whether or not it's open). This last point could be subject for debate.

## ✅ Checklist


atmgrifter00:
@mollykreis would you mind buddying this?

packages/nimble-components/src/select/index.ts
```diff
@@ -757,11 +812,10 @@ export class Select
 
         if (this.open) {
             this.initializeOpenState();
-            this.indexWhenOpened = this.selectedIndex;
```
atmgrifter00, line 760: This just moved into `initializeOpenState`.

change/@ni-nimble-components-9da2496e-e544-4019-a666-d7fca3046771.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: I'm seeing a bug when I play with this in storybook:
1. Open storybook for the select with a placeholder & filter-mode set to `standard`
2. Using the mouse, click the select to open it and then click "Option 3"
3. Using the mouse, click the select to open it again
4. Type "Zur" in the filter box
5. Press Escape to close the dropdown
6. Press Enter to open the dropdown again
7. Notice that the value in the select incorrectly updates to "Zurich" and "Option 3" is selected with the green highlight
8. Also notice that if you press the Up Arrow, "Option 4" becomes highlighted in green in addition to "Option 3" staying highlighted.

change/@ni-nimble-components-9da2496e-e544-4019-a666-d7fca3046771.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "major",
```
mollykreis, line 2: Do you expect this breaking change to have an impact on clients?

atmgrifter00: > Do you expect this breaking change to have an impact on clients?

Sorry, I meant this to be a patch change.

packages/nimble-components/src/select/tests/select.foundation.spec.ts
```diff
@@ -298,7 +298,7 @@ describe('Select', () => {
         await disconnect();
     });
 
-    describe("should NOT emit a 'change' event when the value changes by user input while open", () => {
+    describe("should NOT emit a 'change' event while open", () => {
```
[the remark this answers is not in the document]
atmgrifter00: Updated docs at top of file.

change/@ni-nimble-components-9da2496e-e544-4019-a666-d7fca3046771.json
atmgrifter00: This issue stemmed from the typeahead feature of the `Select`. My solution was to fork/override the relevant implementation from the FAST `Listbox` so that we can control when the `selectedIndex` is being updated, and additionally manage the selected state of the options as we do elsewhere while the dropdown is open.

mollykreis, line 1: You should mark #2022 as being fixed by this PR. You mention it in your PR description, but it isn't marked as being fixed.

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +944,23 @@ export class Select
         this.committedSelectedOption = options[this.selectedIndex];
     }
 
+    private toggleNewActiveOption(getNewActiveIndex: () => number): void {
```
mollykreis, line 947: The behavior of setting `aria-selected` on the active option seems incorrect to me, though I will admit that I haven't studied all the ARIA guidelines as much as I'm sure you have.

However, if the active option isn't considered the selected option until it is "committed", then shouldn't `aria-selected` follow that? That would mean that the value of the `aria-selected` option should always be in sync with the `value` and `selectedIndex` of the select. That also seems to align with the [Select-Only Combobox Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).

packages/nimble-components/src/select/index.ts
```diff
@@ -476,11 +481,14 @@ export class Select
                 o => !o.disabled
             );
             if (enabledOptions.length > 0) {
-                enabledOptions[0]!.selected = true;
+                const selectedOption = enabledOptions.find(
+                    o => o === this.committedSelectedOption
+                ) ?? enabledOptions[0]!;
+                selectedOption.ariaSelected = 'true';
```
mollykreis, line 487: I think it would be nice if you could scroll to the item that is being marked as active. How difficult would that be?

change/@ni-nimble-components-9da2496e-e544-4019-a666-d7fca3046771.json
mollykreis, line 1: With these changes, using up/down keyboard arrows no longer updates the scroll position of the listbox to keep the active element in view.

packages/nimble-components/src/select/index.ts
mollykreis, line 947: Is there a reason why `toggleNewActiveOption` takes a function rather than simply a number? It isn't clear to me why all the callers are creating functions to wrap the logic for determining the new active index.

packages/nimble-components/src/select/index.ts
```diff
@@ -722,6 +816,20 @@ export class Select
         }
     }
 
+    /**
+     * Fork of Listbox implementation to prevent placeholder option from being selected.
+     */
+    protected override getTypeaheadMatches(): ListboxOption[] {
```
mollykreis, line 822: Instead of forking the entire implementation, what if you called the super class and then further filtered the result? Then, we wouldn't have to maintain the regex logic within nimble. Something like this:

```ts
protected override getTypeaheadMatches(): ListboxOption[] {
    const matchingOptions = super.getTypeaheadMatches();
    return matchingOptions.filter(o => !o.hidden & !o.disabled);
}
```

Technically, my suggestion is a little less performant because it's filtering the list twice, but since the second time only filters a subset of items (and we're talking a somewhat limited number of options), I'm not sure if that's a real concern.

packages/nimble-components/src/select/index.ts
```diff
@@ -967,6 +1091,8 @@ export class Select
             return;
         }
 
+        this.indexWhenOpened = this.selectedIndex;
```
mollykreis, line 1094: This comment is for a few lines above, but it won't let me comment there since the code isn't changed.

What is the purpose of the `if (!this.open)` block within `initializeOpenState()`? Does this function get called when the select is closing? I ask because I noticed some duplicate logic between this if-statement and `openChanged`, and I'm not sure if that's intentional or not. Does `this.openActiveIndex = undefined;` need to be in both places as well, or is the duplication unnecessary?

packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -53,9 +54,19 @@ export class SelectPageObject {
     }
 
     public getSelectedOption(): ListOption | null {
+        if (this.selectElement.open) {
+            return this.getSelectedDropdownOption();
+        }
```
mollykreis, line 59: Perhaps I don't understand all the usages of `getSelectedOption`, but it seems misleading to return the active element when it hasn't actually been selected yet. Should this function be left unchanged from master and a new `getActiveOption` function be added to the page object?

change/@ni-nimble-components-9da2496e-e544-4019-a666-d7fca3046771.json
atmgrifter00: I think I did that. Once this PR is submitted, I think that issue will be marked as closed.

atmgrifter00: Provided a new `focusAndScrollIntoView` method which we now call from the `toggleNewActiveOption` method.

packages/nimble-components/src/select/index.ts
atmgrifter00: Done.

atmgrifter00: Not difficult, but testing it with a unit test seems particularly challenging given the fact that the option will be scrolled into view in an rAF call. I verified manually.

packages/nimble-components/src/select/index.ts
atmgrifter00: I think I was just over-complicating things in my head. Simplified now.

packages/nimble-components/src/select/index.ts
atmgrifter00: Done.

packages/nimble-components/src/select/testing/select.pageobject.ts
atmgrifter00: Done. I was concerned that there were going to be some callers in of this method that may have been relying on the initial behavior, but a) there aren't, and b) that isn't a valid reason to create a disingenuous method.

packages/nimble-components/src/select/index.ts
```diff
@@ -510,13 +520,21 @@ export class Select
...
+                    (
+                        this.options[this.selectedIndex]! as ListOption
+                    ).activeOption = true;
+                }
             }
 
-            if (this.indexWhenOpened !== this.selectedIndex) {
+            if (this.indexWhenOpened !== currentActiveIndex) {
                 this.updateValue(true);
+                this.selectedIndex = currentActiveIndex;
```
mollykreis, line 537: Is there ever a time that `this.selectedIndex` shouldn't be updated when `this.updateValue()` is called? Should the setting of `this.selectedIndex` be moved into `this.updateValue()` to ensure they always stay in sync?

packages/nimble-components/src/select/index.ts
```diff
@@ -571,35 +590,49 @@ export class Select
                 if (!this.open) {
                     break;
                 }
+
+                this.toggleNewActiveOption(
+                    this.indexWhenOpened ?? this.selectedIndex
```
mollykreis, line 595: I find the distinction between `this.indexWhenOpen` and `this.selectedIndex` a bit confusing now. Wouldn't those always be the same value since `this.selectedIndex` doesn't change while keying through the options?

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +931,41 @@ export class Select
         this.committedSelectedOption = options[this.selectedIndex];
     }
 
+    private toggleNewActiveOption(newActiveIndex: number): void {
```
mollykreis, line 937: Perhaps this should be called `updateNewActiveOption` since it isn't actually "toggling" the option.

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +931,41 @@ export class Select
...
+        const selectedIndex = this.openActiveIndex ?? this.selectedIndex;
+        const activeOption = this.options[newActiveIndex];
```
mollykreis, line 939: Should `activeOption = this.options[newActiveIndex]` move into the if-statement? Doesn't seem like it makes sense to do that assignment when you haven't checked that the `newActiveIndex` isn't -1.

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +931,41 @@ export class Select
...
+            }
+
+            if (this.open) {
+                this.openActiveIndex = newActiveIndex;
+                this.focusAndScrollActiveOptionIntoView();
+            } else {
+                this.selectedIndex = newActiveIndex;
+            }
+
+            this.ariaActiveDescendant = this.options[newActiveIndex]?.id ?? '';
```
mollykreis, line 953: You can use `activeOption` here instead of indexing into `this.options` again, right?

packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -138,6 +163,26 @@ export class SelectPageObject {
...
+                'character parameter must contain only a single character'
+            );
+        }
+
+        if (
+            this.selectElement.open
+            && this.selectElement.filterMode !== FilterMode.none
+        ) {
+            const filterInput = this.getFilterInput();
+            filterInput!.value = filterInput!.value + character;
```
mollykreis, line 178: nit: Why not put the `!` on the line above so it's only needed once?

```suggestion
            const filterInput = this.getFilterInput()!;
            filterInput.value = filterInput.value + character;
```

packages/nimble-components/src/select/index.ts
atmgrifter00: So, the idea is that `updateValue` should be called in response to `selectedIndex` changing, as its implementation is optimized to pull the value to set from the current `selectedIndex`. If there is a refactor to be done to provide a more confident synchronization between these two fields, I'd rather hold that for another time.

packages/nimble-components/src/select/index.ts
atmgrifter00: Good call! I've removed the `indexWhenOpened` field now. Hopefully, the result feels more understandable.

packages/nimble-components/src/select/index.ts
atmgrifter00: Renamed to `setActionOption`.

packages/nimble-components/src/select/index.ts
atmgrifter00: The code is set up to handle the case where the result of indexing with -1 results in an undefined value. There's been further code cleanup, however, that hopefully makes this intent more clear.

packages/nimble-components/src/select/index.ts
```diff
@@ -471,20 +473,23 @@ export class Select
...
-                enabledOptions[0]!.selected = true;
-            } else {
-                // only filtered option is disabled
-                this.selectedOptions = [];
-                this.selectedIndex = -1;
-            }
-        } else if (this.committedSelectedOption) {
-            this.committedSelectedOption.selected = true;
+        const enabledOptions = this.filteredOptions.filter(o => !o.disabled);
+        let selectedOptionIndex = this.filter !== ''
```
mollykreis, line 477: Perhaps `selectedOptionIndex` should be `activeOptionIndex` or `updatedActiveOptionIndex`. Calling it "selected" is a bit misleading since the option isn't actually being selected.

packages/nimble-components/src/select/index.ts
```diff
@@ -471,20 +473,23 @@ export class Select
...
+            ? this.openActiveIndex ?? this.selectedIndex
+            : this.selectedIndex;
+
+        if (
+            enabledOptions.length > 0
+            && (this.selectedIndex < 0
```
mollykreis, line 483: What is the purpose of the `this.selectedIndex < 0` clause here? Shouldn't the active option always be updated to something within the filtered list, regardless of whether or not the currently selected item is in the filtered list?

packages/nimble-components/src/select/index.ts
```diff
@@ -510,12 +515,19 @@ export class Select
...
+            let currentActiveIndex = this.openActiveIndex ?? this.selectedIndex;
             this.open = false;
-            if (this.selectedIndex === -1) {
-                this.selectedIndex = this.indexWhenOpened!;
+            if (currentActiveIndex === -1) {
+                currentActiveIndex = this.selectedIndex;
+                if (this.selectedIndex >= 0) {
+                    (
+                        this.options[this.selectedIndex]! as ListOption
+                    ).activeOption = true;
```
mollykreis, line 525: What's the purpose of setting `activeOption` to `true` here? Didn't the select just close? I would expect `activeOption` to be updated when the select opens again, but I wouldn't expect that attribute to need to stay up-to-date on options while the select is closed.

packages/nimble-components/src/select/index.ts
```diff
@@ -622,9 +645,36 @@ export class Select
...
+     * @internal
+     * Fork of Listbox implementation, so that the selectedIndex is not changed while the dropdown
+     * is open.
+     */
+    public override typeaheadBufferChanged(_: string, __: string): void {
+        if (this.$fastController.isConnected) {
+            const typeaheadMatches = this.getTypeaheadMatches();
+
+            if (typeaheadMatches.length) {
+                const selectedIndex = this.options.indexOf(
```
mollykreis, line 664: Similar to one of my other comments, can this be called something other than `selectedIndex`? Only in some code paths will this item actually become selected. Perhaps something more like `matchIndex` would be more appropriate.

packages/nimble-components/src/select/styles.ts
```diff
@@ -116,6 +118,18 @@ export const styles = css`
         overflow: auto;
     }
 
+    ::slotted([role='option'][active-option]) {
+        background-color: ${fillSelectedColor};
+    }
+
+    ::slotted([role='option']:not([active-option])) {
+        background-color: none;
```
mollykreis, line 126: This styling isn't working because `none` isn't a valid value for `background-color`. You could either do `background: none` or `background-color: transparent`.

It would be nice to have a matrix test to catch things like this. Could you write a test case where the select is open and has a different active option than the selected option?

packages/nimble-components/src/select/index.ts
```diff
@@ -571,35 +585,44 @@ export class Select
...
 
-                if (this.selectedIndex !== this.indexWhenOpened!) {
-                    this.options[this.selectedIndex]!.selected = false;
-                    this.selectedIndex = this.indexWhenOpened!;
-                }
+                currentActiveIndex = this.selectedIndex;
                 this.focus();
                 break;
             }
             case keyTab: {
```
mollykreis, line 600: I'm probably missing a nuance, but why do you need to handle the tab key here? What makes tab different than other events that lead to focusout?

packages/nimble-components/src/select/index.ts
```diff
@@ -961,13 +1078,9 @@ export class Select
...
     private initializeOpenState(): void {
-        if (!this.open) {
-            this.ariaExpanded = 'false';
-            this.ariaControls = '';
-            return;
-        }
-
+        this.openActiveIndex = this.selectedIndex;
         this.committedSelectedOption = this.options[this.selectedIndex];
+        this.setActiveOption(this.openActiveIndex ?? this.selectedIndex);
```
mollykreis, line 1083: You don't need the `?? this.selectedIndex` on this line, right? You just initialized `this.openActiveIndex = this.selectedIndex` two lines up.

packages/nimble-components/src/select/index.ts
atmgrifter00: Nope. I think this was just a result of me creating an unneeded implementation for testing purposes. I've now changed the page object `pressTabKey` method to also emit a `focusout` event.

packages/nimble-components/src/select/index.ts
atmgrifter00: Yeah, this is not needed.

packages/nimble-components/src/select/index.ts
atmgrifter00: Removed this check.

packages/nimble-components/src/select/styles.ts
atmgrifter00: Added a matrix story that navigates an open dropdown. It indeed would have caught that styling goof.

packages/nimble-components/src/select/index.ts
```diff
@@ -622,9 +622,36 @@ export class Select
...
+
+            if (typeaheadMatches.length) {
+                const activeOption = this.options.indexOf(
+                    typeaheadMatches[0] as ListOption
+                );
+                if (activeOption > -1 && !this.open) {
+                    this.selectedIndex = activeOption;
+                } else if (this.filterMode === FilterMode.none) {
+                    this.setActiveOption(activeOption);
+                }
```
mollykreis, line 648: Should this `else if` block check `this.open` and only call `this.setActiveOption` if the select is open? I'm guessing there isn't any harm in calling `this.setActiveOption` when it is closed, but it would probably make things more clear & consistent if we aren't updating `this.setActiveOption` if half of the code paths when the select is closed but not the other half.

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +897,40 @@ export class Select
         this.committedSelectedOption = options[this.selectedIndex];
     }
 
+    private setActiveOption(newActiveIndex: number): void {
+        const selectedIndex = this.openActiveIndex ?? this.selectedIndex;
```
mollykreis, line 901: How about `previousActiveIndex` instead of `selectedIndex` for this variable name?

packages/nimble-components/src/select/index.ts
```diff
@@ -837,6 +897,40 @@ export class Select
...
+        const activeOption = this.options[newActiveIndex];
+        if (isNimbleListOption(activeOption)) {
+            activeOption.activeOption = true;
+            if (selectedIndex !== newActiveIndex && selectedIndex > -1) {
+                (this.options[selectedIndex]! as ListOption).activeOption = false;
+            }
```
mollykreis, line 907: Is there ever a time where the cast of `this.options[selectedIndex]! as ListOption` isn't valid? One scenario I'm thinking of is when the set of slotted options changes while the select is open. To be robust, should you check `isNimbleListOption(selectedIndex)` as part of the if-statement?

packages/nimble-components/src/select/index.ts
```diff
@@ -948,7 +1042,7 @@ export class Select
 
     private clearSelection(): void {
```
mollykreis, line 1043: A few thoughts on this `clearSelection` function:
- The name doesn't seem appropriate now that it doesn't have to do with selection
- It seems odd to be clearing the activeOption value on each list option without updating `this.openActiveIndex` back to -1

Is this function even needed with the changes you made to `inputHandler()` so that it always calls `setActiveOption`? I believe that is the only caller of this function.

packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -56,6 +57,20 @@ export class SelectPageObject {
         return (this.selectElement.selectedOptions[0] as ListOption) ?? null;
     }
 
+    public getActiveOption(): ListOption | null {
+        return (
+            (this.selectElement.options.find(
+                o => (o as ListOption).activeOption === true
```
mollykreis, line 63: nit: You shouldn't need the `=== true` check here since `activeOption` is already a `boolean`.

packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -68,7 +83,11 @@ export class SelectPageObject {
             throw new Error('Select must be open to click selectedItem');
         }
 
-        this.clickOption(this.selectElement.selectedIndex);
+        const selectedOption = this.getActiveOption();
```
mollykreis, line 86: Should this function be renamed to `clickActiveItem`? You aren't clicking the selected item anymore.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -326,7 +324,64 @@ describe('Select', () => {
         const selectedOption = pageObject.getSelectedOption();
         selectedOption!.textContent = 'foo';
         await waitForUpdatesAsync();
-        expect(element.displayValue).toBe('foo');
+        expect(pageObject.getDisplayText()).toBe('foo');
+
+        await disconnect();
+    });
+
+    it('after forcing Select to be blank, user can arrow down to first available option in dropdown', async () => {
```
mollykreis, line 332: If you don't already have it, it would be good to have a similar test as this that leaves the select closed when pressing the down arrow key.

mollykreis (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: You're right, but this required a slight logic reorganization in `setActiveIndex` so that we _always_ update the `activeOption` state of the previous active index to `false`. (if different than the new one). It was weird, regardless, to have been doing that only when the new active index was valid.

packages/nimble-components/src/select/testing/select.pageobject.ts
atmgrifter00: Ehhh...maybe? I hesitate to do that at the moment, as this would technically be a breaking change (though I see no usages of this method outside of our repo). I'll leave open for others to comment on.

packages/nimble-components/src/select/index.ts
```diff
@@ -571,35 +576,30 @@ export class Select
                 if (!this.open) {
                     break;
                 }
+
+                this.setActiveOption(this.selectedIndex);
```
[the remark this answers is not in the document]
atmgrifter00: This is analogous to line 579-582 in the old code. Without this, if you had navigated to a different option, and then pressed <Esc> nothing after this line would clear the `activeOption` state from the option that had been navigated to. The other recourse (though I have now removed this) would be to retain the `clearSelection` method and call that, again, at the beginning of `inputHandler`.

Calling `setActiveOption` here seems the most optimal solution.

packages/nimble-components/src/select/index.ts
atmgrifter00: Made the check.

packages/nimble-components/src/select/index.ts
atmgrifter00: As discussed in person, we _do_ call `setActiveOption` for code paths when the dropdown is closed. This really meant that this implementation could simplify to unconditionally calling `setActiveOption`.

packages/nimble-components/src/select/index.ts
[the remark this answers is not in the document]
atmgrifter00: Discussed further, and the better solution is just to reset/clear the `activeOption` state when the dropdown closes.

packages/nimble-components/src/select/styles.ts
```diff
@@ -116,6 +118,18 @@ export const styles = css`
...
+        background-color: ${fillSelectedColor};
+    }
+
+    ::slotted([role='option']:not([active-option])) {
+        background: none;
+    }
+
+    ::slotted([role='option'][active-option]:hover) {
+        background-color: ${fillHoverSelectedColor};
+    }
```
rajsite, line 131: Can these styles be owned by the list option element itself? None of it seems coupled to the state of this element

rajsite (commented):
Didn't finish reviewing today, expect to Monday

rajsite (approved):

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -358,6 +427,33 @@ describe('Select', () => {
...
+                    selectValue = element.value;
+                    resolve(true);
+                });
+                pageObject.pressArrowDownKey();
+                void (async () => {
+                    await pageObject.clickAway();
+                })();
+            }),
+            waitForUpdatesAsync().then(() => false)
+        ]);
```
rajsite, line 450: This seems a bit convoluted, why is the Promise race needed?

Can we have a helper to simplify this? Even if the following simplification isn't possible I want to avoid mixing promise and async patterns in the same block.

```ts
const changePromise = pageObject.waitForChange()
await pageObject.clickAway();
await changePromise;
selectValue = element.value;
```

Can also use [`once`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options) to clean-up easier:

```ts
waitForChange () {
   return new Promise(resolve => {
      this.element.addEventListener('change', () => resolve(), {
         once: true
      });
   });
}
```

packages/nimble-components/src/select/styles.ts
atmgrifter00: The problem is with the middle style in particular, where we remove the background when there is no `active-option` attribute present. Currently, this will only work for the `Select`, as all other dropdowns that use the `ListOption` do not apply the `active-option` attribute, and would thus _always_ set the background to `none`, which is not what we want...yet.

Would you like me to create an issue that tracks the need to move these styles to the `ListOption` CSS?

rajsite: Either make an explicit one or capture as an item in the align list box implementations tech debt: https://github.com/ni/nimble/issues/1926

packages/nimble-components/src/select/styles.ts
```diff
@@ -116,6 +118,18 @@ export const styles = css`
         overflow: auto;
     }
 
+    ::slotted([role='option'][active-option]) {
+        background-color: ${fillSelectedColor};
+    }
+
+    ::slotted([role='option']:not([active-option])) {
+        background: none;
+    }
```
rajsite, line 127: Can we avoid the :not selector by switching the order and changing to color?
Main goal is to [avoid not selectors and rely on cascade](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes)

```suggestion
    ::slotted([role='option']) {
        background-color: transparent;
    }
    ::slotted([role='option'][active-option]) {
        background-color: ${fillSelectedColor};
    }
```

packages/nimble-components/src/select/styles.ts
atmgrifter00: Updated the existing tech debt issue.

jattasNI (approved):

packages/nimble-components/src/select/testing/select.pageobject.ts
jattasNI: Don't fear breaking changes, especially when they don't impact any existing clients! Nimble components major version is only at 28 so we still have plenty more versions to bump to before we run out of numbers.

packages/storybook/src/nimble/select/select-matrix.stories.ts
```diff
@@ -100,6 +102,29 @@ export const blankListOption: StoryFn = createStory(
...
+                });
+                select.dispatchEvent(arrowDownEvent);
+                await waitForUpdatesAsync();
+            }
+        )
+    );
+};
+
+export const navigateToDifferentOption: StoryFn = createStory(
+    html`<${selectTag} open>
```
rajsite, line 120: @atmgrifter00 fyi in #2073 we started setting explicit widths on select / combobox stories

```suggestion
    html`<${selectTag} open style="width: 250px;">
```

atmgrifter00: Gah! I'll have to address this in a follow-up PR. Sorry.

rajsite: I did it in this PR by merging the above comment, you're all good, was just an FYI for future story work

PR #2030, opened 2024-04-22
Add Storybook for `spright-components`

### LINKED ISSUE #1978: Storybook for `spright-components`
# 🙋 Feature Request

## 😯 Problem to Solve

`spright-components` should be documented via Storybook, like `nimble-components`.

## 💁 Proposed Solution

~~- Create new Storybook for `spright-components` (both component docs and visual tests)~~
~~- Should be built and published by the PR/CI build~~

Create a new private package for all the stories of both Spright and Nimble. We can initially create the new package with the Spright Rectangle stories, then later move all the stories from Nimble.

This issue should not be closed until Spright stories are being published and visual tests are running. I.e. after Nimble stories are moved to the storybook package.

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1978

## 👩‍💻 Implementation

Long-term, we want a single Storybook for both Nimble and Spright. This will mean pulling all the Storybook code and story files into a separate package that depends on `nimble-components` and `spright-components`. Initially, we'll create this separate package, but it will only contain the Spright stories. Later, we will move all of Nimble's stories into that package and start publishing from the new package instead of the Storybook in `nimble-components`.

- Created `storybook` with:
   - `.storybook` directory (copied from `nimble-components`)
   - a handful of Storybook-related utility files (copied from `nimble-components/src/utilities/tests`)
   - React wrappers for check and exclamation mark icons used by stories
   - standard story and visual tests for Spright Rectangle component
- Removed Storybook-related dependencies and configuration from `spright-components`

## 🧪 Testing

Built and ran Storybook locally.

## ✅ Checklist


.github/workflows/main.yml
m-akinc, line 1: Chromatic publish is not working because it is using the same Chromatic project token for both Storybooks, and it doesn't rebuild for the same commit.

packages/site/landing/index.html
```diff
@@ -25,7 +25,7 @@
             </div>
             <div class="cards">
                 <div class="card-wrapper">
-                    <a class="card" href="storybook/">
+                    <a class="card" href="storybook/nimble">
```
rajsite, line 28: early feedback:
- I don't want to break any of our current urls
- Duplicating some config feels like one thing, duplicating lots of source feels like another
- I'm curious if long-term we do actually want separate storybooks

Can you make a meeting to just brainstorm these things with @jattasNI and I?

.github/workflows/main.yml
m-akinc: Moot now that we're not publishing the Spright Storybook.

packages/site/landing/index.html
m-akinc: Updated description with new direction that we agreed upon.

change/@ni-spright-components-a88786af-f8fe-498a-bc10-fe223f2ffa2f.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: - This is marked as resolving the storybook issue. I think the agreement was migrating all of nimble to the separate package ultimately resolves that feautre. How should we stage / track it?
- There were concerns about developer workflow, what did you evalutate and what is future work?

packages/nimble-storybook/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@ni-private/nimble-storybook",
```
rajsite, line 2: we use the `nimble-` prefix for packages that are nimble specific, I'd say lets do just `storybook` similar to `site`, `performance`, etc.
```suggestion
  "name": "@ni-private/storybook",
```

packages/nimble-storybook/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "build": "tsc -p ./tsconfig.json && storybook build -o dist/storybook --webpack-stats-json",
+    "start": "storybook dev -p 6006",
+    "lint": "eslint .",
+    "format": "eslint . --fix"
+  },
+  "dependencies": {
+    "@microsoft/fast-react-wrapper": "^0.3.22",
+    "@ni/nimble-components": "^26.0.0",
+    "@ni/spright-components": "^0.0.2",
+    "@ni/nimble-tokens": "^6.13.2",
```
rajsite, line 17: Private packages that are not published can depend on monorepo packages with * to avoid unnecessary package file updates

```suggestion
    "@ni/nimble-components": "*",
    "@ni/spright-components": "*",
    "@ni/nimble-tokens": "*",
```

packages/nimble-storybook/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "private": true,
+  "type": "module",
+  "description": "Storybook for the Nimble repo",
+  "scripts": {
+    "build": "tsc -p ./tsconfig.json && storybook build -o dist/storybook --webpack-stats-json",
+    "start": "storybook dev -p 6006",
+    "lint": "eslint .",
+    "format": "eslint . --fix"
+  },
+  "dependencies": {
```
rajsite, line 13: I'm not sure the distinction between dev and devDep is meaningful here. Lets just move these all to devDep

packages/nimble-storybook/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "circular-dependency-plugin": "^5.2.0",
+    "js-beautify": "^1.15.1",
+    "remark-gfm": "^4.0.0",
+    "storybook": "^8.0.4",
+    "storybook-addon-pseudo-states": "^3.0.0",
+    "terser-webpack-plugin": "^5.3.10"
+  },
+  "devDependencies": {
+    "@chromatic-com/storybook": "^1.2.25",
+    "@ni-private/eslint-config-nimble": "^1.0.0",
```
rajsite, line 40: eslint configs should always be peerDep

packages/nimble-storybook/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "eslint-plugin-storybook": "^0.8.0",
```
rajsite, line 41: eslint configs should always be peerDep

packages/nimble-storybook/src/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
rajsite, line 1: What's the idea for nimble and spright living under this src/ tree?
Maybe we should have src/nimble and src/spright?

packages/nimble-storybook/package.json
m-akinc: Why not devDep?

packages/nimble-storybook/src/.eslintrc.js
m-akinc: Sure, that's reasonable. Probably some advantage to keeping them separate.

change/@ni-spright-components-a88786af-f8fe-498a-bc10-fe223f2ffa2f.json
m-akinc: I've updated the PR description so it won't close the issue. I've updated the issue to describe the new plan, and we can use it to track the rest of the work. I will move the Nimble stories after this PR goes in.

I found that making a change in a component does now require `npm run build-components` before it gets picked up by the local, running Storybook. That's not so bad. I was more concerned that you would have to stop and restart Storybook, since that takes a relatively long time.

packages/nimble-storybook/package.json
rajsite: See discussion https://github.com/ni/javascript-styleguide/issues/93

packages/storybook/src/spright/rectangle/rectangle-matrix.stories.ts
```diff
@@ -0,0 +1,59 @@
...
+import { createStory } from '../../utilities/storybook';
+import { hiddenWrapper } from '../../utilities/hidden';
+import {
+    disabledStates,
+    type DisabledState,
+    disabledStateIsEnabled
+} from '../../utilities/states';
+
+const metadata: Meta = {
+    title: 'Tests (Spright)/Rectangle',
```
jattasNI, line 20: We have logic in `manager.js` to exclude the tests folder from public storybook builds and I think it won't match the "(Spright)" pattern.

How does it look if we do "Tests/Spright/Rectangle"? Or "Tests/Spright Rectangle"? Or we can update the logic to match this pattern too.

rajsite: I'd vote for `Tests Spright/Rectangle` to avoid extra punctuation since it's used for links etc

change/@ni-spright-components-a88786af-f8fe-498a-bc10-fe223f2ffa2f.json
jattasNI: In the storybook work I just did I really benefitted from having "hot reload" and got frustrated when it (fairly frequently) stopped working and I had to rerun commands. Could we keep that a possibility by starting a `npm run build-components:watch` or something like that in a separate console before running `npm run storybook`? Could you test that it works and document the workflow?

packages/storybook/src/spright/rectangle/rectangle-matrix.stories.ts
m-akinc: I did the simplest thing, which is to change `startsWith('Tests/')` to `startsWith('Tests')`. I'm not worried about potential, future false positives like `Testsomething`, but if anyone is, I can change it to use a regex.

change/@ni-spright-components-a88786af-f8fe-498a-bc10-fe223f2ffa2f.json
m-akinc: > Could we keep that a possibility by starting a `npm run build-components:watch` or something like that in a separate console before running `npm run storybook`?

Yes, that works just fine. I've added a note about it in the Storybook package's README.

packages/storybook/src/spright/rectangle/rectangle-matrix.stories.ts
rajsite: I'd still like to avoid punctuation in the name

m-akinc: Sure, I hadn't refreshed to see your reply before I responded to Jesse.

packages/spright-components/.prettierignore
```diff
@@ -1,29 +1,20 @@
...
-/.storybook/*.*
-!/.storybook/*.js
-
 # Explicitly exclude the below files and folders
 dist
 node_modules
 package.json
 CHANGELOG.*
-src/utilities/tests/tests/fixture.spec.ts
-src/utilities/tests/fixture.ts
```
rajsite, line 28: Think we missed it in last PR but we should have a test setup in spright-components using the fixture tools. It's jasmine test specific not stroybook related

rajsite (approved):

Issue #1192, opened 2023-04-19
Migrate hard-coded tokens to new base tokens


rajsite:
## 🧹 Tech Debt

With the completion of #1183 we can now add new base design tokens to nimble in style dictionary. Some tokens we need to add are:
- [ ] The tokens defined in `design-tokens-static`, see discussion https://github.com/ni/nimble/issues/370#issuecomment-1233423198. After which we can remove design-tokens-static.
   - Update April 2024, We can actually just switch back to normal token usage and delete the static tokens once [CSS ::backdrop inheritance](https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element) is available in all our browsers (late 2024). 
- [x] The fallback font definitions:  https://github.com/ni/nimble/blob/7118a210663789426b00d7817371924b0d8ea0e0/packages/nimble-components/src/theme-provider/design-tokens.ts#L104-L122
   - fixed by merging fallback font into font and font-family in #2033

jattasNI:
Assigned to @rajsite to update the proposed pattern for fallback fonts. We shouldn't need to expose tokens, we can just include the fallback fonts in the primary font token values.

rajsite:
Remaining work captured in https://github.com/ni/nimble/issues/1835

PR #2033, opened 2024-04-23
Include Fallback font in font and font-family tokens

rajsite:
# Pull Request

## 🤨 Rationale

Found that the nimble landing page was not using the fallback fonts when loading on a slow network. The reason was that the fallback fonts were previously not included in the font-family token but were a separate token: https://github.com/ni/nimble/blob/bad9f4c573bca5528355ec93e10bafe12b76e386/packages/site/landing/styles.scss#L110

This change updates the `font-family` theme tokens to always be defined as the base token font family combined with the fallback font family. This impacts the definition of the `font-family` tokens (but `font` is effectively unchanged). From [codesearch for font-family usage](https://codesearch.natinst.com/search/text?q=nimble.*font.*family&fold_case=auto&regex=true&context=true) this will impact those usages but hopefully for the better via improved fallback fonts during page load.

In addition it removes the fallback font tokens from being explicitly exported to prevent accidental direct usage. I found one direct usage of the [fallback token in codesearch](https://codesearch.natinst.com/search/text?q=nimble.*fallback.*family&fold_case=auto&regex=true&context=true) but looks like they are leveraging the SCSS build so it should be identified as part of build in an upgrade.

Partially addresses https://github.com/ni/nimble/issues/1192

## 👩‍💻 Implementation

Went for defining the Fallback font by convention. It assumes that all our Font Families that are defined will also define a Fallback using a known convention of `[Font Family] Fallback`.

Alternatives: Create a bunch of tokens for the Fallback font families as initially discussed in the issue. It didn't really seem high value. The definition of the fallback fonts is completely encapsulated in the CSS files using CSS features. Outside the web context having the token for the fallback font is not useful unless we also encode all the state for the fallback (size-adjust, etc) and even that might be CSS-specific. And in the web context you shouldn't really use it directly and rely on the theme token.

## 🧪 Testing

Rely on CI.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. I think that the font README sufficiently covers the need for creating Fallback tokens.


change/@ni-nimble-components-1ec0d312-a8f8-4af4-8661-74647a231e9d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Remove fallback design tokens",
```
jattasNI, line 3:
```suggestion
  "comment": "Remove fallback design tokens. Use primary font tokens instead; they include fallback fonts after Nimble's font.",
```

jattasNI (approved):

fredvisser (approved):

PR #827, opened 2022-11-16
Examples HLD

rajsite:
# Pull Request

## 🤨 Rationale

HLD for the structure of nimble examples

## 👩‍💻 Implementation

See HLD

## 🧪 Testing

N/a

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2036, opened 2024-04-24
Create `SprightBlazor` project

### LINKED ISSUE #1976: Create `SprightBlazor` project
# 🙋 Feature Request

## 😯 Problem to Solve

Part of setting up Spright.

## 💁 Proposed Solution

- Create `spright-blazor` fake npm package with `SprightBlazor` project
  - Blazor API for `spright-accordion` demo component
  - Build should build and publish it
  - No example program initially (later PR)

m-akinc:
# Pull Request

## 🤨 Rationale

Partially addresses #1976

## 👩‍💻 Implementation

Initially I had created a separate `spright-blazor` folder/package with its own VS solution, but it seems more fitting to just add Spright projects to the existing solution. This required creating separate `package.json` files located in the project directories. The old `package.json` in the solution directory was changed into a private package, similar to our Angular workspace package.

I opted to leave the directory name as `packages/nimble-blazor`, but for consistency we might want to change it to `blazor-workspace` (matching the private package name). The same goes for the solution name.

## 🧪 Testing

- Spot tested `build`, `lint`, `format`, `pack`, and `test` commands from root of Blazor workspace.
- Tested that the example builds and runs.

## ✅ Checklist


change/@ni-nimble-blazor-92675eea-194d-4c8d-81ea-ff1049cd4f6c.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Gonna move it to draft. This change is in an unclear middle ground where it's behaving monolithic and like separate packages. The angular workspace isn't the ideal setup, I'd love for the angular tooling to be normal and exist like any other npm package instead of a custom workspace. Want to discuss it to make sure the responsibilities of the packages are clear if we go this way. Can discuss it post standup I guess since we didn't didcuss before the PR went to owners.

rajsite: points I want to discuss because I'll forget otherwise:
- the [eventual cdn packages](https://github.com/ni/nimble/pull/827/files#diff-0967e47790c3f683cc13aff0bdc76ecdcf42ebe20f0d9aadcb89b4fe7bf2ed5aR26) as owners of all-components
- long-term direction of actual separate packages, publishing an npm blazor package for efficient builds
- vs behavior of a monolithic workspace, top-level should be only thing participating, sub packages almost no deps or scripts

m-akinc: This feels like a good structure to me, but we can discuss and see if any better solutions come up. Also, I haven't been getting buddy reviews on these Spright PRs, because I don't think others would have much to say about these changes. If you were expecting me to get buddies on these, just let me know.

rajsite: The concern for these in particular is that they are large changes that come as a surprise. Buddying with others not as familiar with the infrastructure isn't the concern. The concern is big infra changes not discussed at all until owner's PR.

packages/nimble-blazor/NimbleBlazor/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@ni/nimble-blazor",
+  "version": "16.0.1",
+  "description": "Blazor components for the NI Nimble Design System",
+  "scripts": {
+    "pack": "cross-env-shell dotnet pack -c Release -p:PackageVersion=$npm_package_version --output ../dist",
```
m-akinc, line 6: The `pack` script needs to stay in the project `package.json` because it relies on the version number from the file.

rajsite: For both of them lets add the commands used to peerDependencies, i.e. cross-env-shell and rimraf

packages/nimble-blazor/README.md
```diff
@@ -2,9 +2,10 @@
     <p><b>ni | nimble | blazor</b></p>
 </div>
 
-# Nimble Blazor
+# Nimble and Spright Blazor
```
rajsite, line 5: In a follow-up PR lets rename the package and sln and split this into two READMEs in the right spots

packages/nimble-blazor/build/generate-hybrid/source/NimbleBlazor.HybridWorkaround.js
```diff
@@ -1,3 +1,3 @@
-import { afterStarted } from '../../../NimbleBlazor/wwwroot/NimbleBlazor.lib.module';
+import { afterStarted } from '@ni/nimble-blazor/wwwroot/NimbleBlazor.lib.module';
```
rajsite, line 1: Can you undo this change. It creates an unnecessary coupling to the blazor build running first.

packages/nimble-blazor/build/generate-hybrid/source/SprightBlazor.HybridWorkaround.js
```diff
@@ -0,0 +1,3 @@
+import { afterStarted } from '@ni/spright-blazor/wwwroot/SprightBlazor.lib.module';
```
rajsite, line 1: Can you undo this change. It creates an unnecessary coupling to the blazor build running first.

packages/nimble-blazor/build/copyResources.js
```diff
@@ -2,32 +2,36 @@ const path = require('path');
 const glob = require('glob');
 const fs = require('fs');
 
-const destinationDirectory = path.resolve(__dirname, '../NimbleBlazor/wwwroot');
-console.log(`Destination directory for blazor assets: "${destinationDirectory}"`);
+copyEverythingToDestination(process.argv[2], process.argv[3]);
```
rajsite, line 5: Can we either pass an enum configuration of what to build with hard coded paths or not do sub encodings, i.e. the package name but just the part after the org / the project name but we assume a specifiic folder layout.

For build scripts I'd prefer to keep it exact and simple and just do an enum / string

```
const target = process.argv[2];
console.log('selected target: ',  target);
if (target === 'nimble-blazor') {do hard coded paths}
else if (target === 'spright-blazor') {do hard coded paths}
else {throw new Error()}
```

Limiting the parameterization of build scripts makes them a lot easier to troubleshoot

packages/nimble-blazor/build/generate-hybrid/source/NimbleBlazor.HybridWorkaround.js
m-akinc: Requires turning of a lint rule about relative imports from other packages, but that's fine.

change/@ni-spright-blazor-21b29346-f22b-4578-81b8-7d76855eca6a.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Probably in a follow-up PR to avoid growing scope, but how does nuget / assembly dependencies work? I expect SprightBlazor needs a dependency on NimbleBlazor for enums, etc at a minimum and potentially also utilities like [AttributeHelpers](https://github.com/ni/nimble/blob/c56b8518de39d66a1c86837b160348def0ad992d/packages/nimble-blazor/NimbleBlazor/AttributeHelpers.cs). Though for utilities we could go the copy route or add an additional assembly for shared functionality, but the enums question still stands.

packages/nimble-blazor/NimbleBlazor/NimbleBlazor.csproj
```diff
@@ -50,10 +51,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Include="package.json" />
     <None Include="packages.lock.json" />
   </ItemGroup>
 
   <Target Name="CopyNimbleResources" BeforeTargets="BeforeBuild">
-    <Exec Command="npm run copy-resources" />
+    <Exec Command="cd ../ &amp;&amp; npm run copy-resources:nimble" />
```
rajsite, line 59: There is a [WorkingDirectory option](https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022#parameters), something like

```suggestion
    <Exec Command="npm run copy-resources:nimble" WorkingDirectory="../" />
```

packages/nimble-blazor/SprightBlazor/SprightBlazor.csproj
```diff
@@ -0,0 +1,61 @@
...
+    <Folder Include="wwwroot\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="package.json" />
+    <None Include="packages.lock.json" />
+  </ItemGroup>
+
+  <Target Name="CopyResources" BeforeTargets="BeforeBuild">
+    <Exec Command="cd ../ &amp;&amp; npm run copy-resources:spright" />
```
rajsite, line 59: There is a [WorkingDirectory option](https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022#parameters), something like

```suggestion
    <Exec Command="npm run copy-resources:nimble" WorkingDirectory="../" />
```

packages/nimble-blazor/package.json
```diff
@@ -1,49 +1,40 @@
...
     "format:cs": "dotnet format",
     "format:js": "eslint . --fix",
     "test": "dotnet test -c Release",
-    "pack": "cross-env-shell dotnet pack NimbleBlazor -c Release -p:PackageVersion=$npm_package_version --output dist",
-    "invoke-publish": "npm run invoke-publish:clean-nuget && npm run pack && npm run invoke-publish:nuget && npm run invoke-publish:npm -- ",
-    "invoke-publish:clean-nuget": "rimraf --glob \"dist/*.nupkg\"",
-    "invoke-publish:nuget": "cross-env-shell dotnet nuget push \"dist/*.nupkg\" -k $NUGET_SECRET_TOKEN -s \"https://api.nuget.org/v3/index.json\"",
-    "invoke-publish:npm": "echo \"noop command to swallow npm args\"",
-    "copy-resources": "node build/copyNimbleResources.js"
+    "copy-resources": "npm run copy-resources:nimble && npm run copy-resources:spright",
```
rajsite, line 23: Each package runs the commands separately, don't need this anymore for calling both

change/@ni-spright-blazor-21b29346-f22b-4578-81b8-7d76855eca6a.json
m-akinc: My assumption is that `SprightBlazor` won't need anything defined by `NimbleBlazor`. Presumably, anything it needs from Nimble will come from `nimble-components` (included in `spright-components`'s all-components bundle). Am I missing something? A client using Spright will likely need dependencies on both the `NimbleBlazor` and `SprightBlazor` and pull in both nugets.

If our `SprightBlazor` API eventually exposes some enum that is defined by `nimble-components`, it is probably simpler just to recreate the enum in the `SprightBlazor` namespace than to depend on the one from `NimbleBlazor`.

For utilities like AttributeHelpers, we can probably just copy them until there is critical mass for making a shared assembly. I don't see any others besides that one.

m-akinc: BTW, I tried adding Spright Rectangle to the Blazor example app, and to my surprise, it worked without any trouble. I was bracing myself to see some redefinition errors from using both the Nimble and Spright all-component bundles, but that didn't seem to happen. Probably needs some more investigation to understand why, or at least to be confident that clients can (always) use `SprightBlazor` and `NimbleBlazor` together without issue.

rajsite: > BTW, I tried adding Spright Rectangle to the Blazor example app, and to my surprise, it worked without any trouble.

Sure... they do a [naive sanity check](https://github.com/microsoft/fast/blob/9c6dbb66615e6d229fc0ebf8065a67f109139f26/packages/web-components/fast-element/src/components/fast-definitions.ts#L200) when registering, doesn't mean you are in a good state when you are trying to register things multiple times. And I strongly suspect object references that need to be shared between components, like design token references, will be broken / competing for defining tokens.

> If our SprightBlazor API eventually exposes some enum that is defined by nimble-components, it is probably simpler just to recreate the enum in the SprightBlazor namespace than to depend on the one from NimbleBlazor.

I'd like the Blazor owners to weigh in on that as well. I'm skeptical of that statement being made so absolutely but haven't tried to think it through.

packages/nimble-blazor/NimbleBlazor/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "invoke-publish": "npm run invoke-publish:clean-nuget && npm run pack && npm run invoke-publish:nuget && npm run invoke-publish:npm -- ",
+    "invoke-publish:clean-nuget": "rimraf --glob \"dist/NimbleBlazor.*.nupkg\"",
+    "invoke-publish:nuget": "cross-env-shell dotnet nuget push \"dist/NimbleBlazor.*.nupkg\" -k $NUGET_SECRET_TOKEN -s \"https://api.nuget.org/v3/index.json\"",
+    "invoke-publish:npm": "echo \"noop command to swallow npm args\""
```
rajsite, line 10: fyi, follow the pattern update here: https://github.com/ni/nimble/pull/2039/files#r1581350041

change/@ni-spright-blazor-21b29346-f22b-4578-81b8-7d76855eca6a.json
m-akinc: Remind me why Spright's all-components bundle has to include Nimble's? Seems like keeping them separate would work fine and avoid potential conflicts. Otherwise, we're implicitly tying each Spright version to a specific version of Nimble, which might result in confusing behavior differences based on whether a client is consuming nimble via `nimble-components` imports or Spright's bundle.

m-akinc: Got my answer over a chat. I had forgotten that these bundles include all dependencies, so even without including Nimble components, we would still end up with two copies of FAST, etc., which is no bueno.

atmgrifter00: > > BTW, I tried adding Spright Rectangle to the Blazor example app, and to my surprise, it worked without any trouble.
> 
> Sure... they do a [naive sanity check](https://github.com/microsoft/fast/blob/9c6dbb66615e6d229fc0ebf8065a67f109139f26/packages/web-components/fast-element/src/components/fast-definitions.ts#L200) when registering, doesn't mean you are in a good state when you are trying to register things multiple times. And I strongly suspect object references that need to be shared between components, like design token references, will be broken / competing for defining tokens.
> 
> > If our SprightBlazor API eventually exposes some enum that is defined by nimble-components, it is probably simpler just to recreate the enum in the SprightBlazor namespace than to depend on the one from NimbleBlazor.
> 
> I'd like the Blazor owners to weigh in on that as well. I'm skeptical of that statement being made so absolutely but haven't tried to think it through.

Yeah, I'm not sure how I feel about this. I guess at worst, a client will have to namespace the "same" enum for its Spright vs Nimble usages:

```html
    <SprightWidget Appearance="SprightBlazor.ButtonAppearance.Ghost">
        <NimbleButton Appearance="NimbleBlazor.ButtonAppearance.Outline">
    </SprightWidget>
```
For the most part, I'm guessing Spright APIs will typically not need enums that are defined in nimble-components, so the above scenario shouldn't be common.

m-akinc: > For the most part, I'm guessing Spright APIs will typically not need enums that are defined in nimble-components, so the above scenario shouldn't be common.

That was my gut feeling, too, but it seems fairly harmless to have SprightBlazor depend on NimbleBlazor (just in case). I found that a project dependency results in the Spright nuget having a dependency to the Nimble nuget version >= <SprightBlazor-version>. I guess this is pretty reasonable when the projects in your solution are at the same version, but doesn't make much sense for us. Instead, I put a package dependency on NimbleBlazor, since that supports specifying the version. It should get updated by Renovate along with the rest of the nuget dependencies.

rajsite: > It should get updated by Renovate along with the rest of the nuget dependencies.

If that is what we really on I expect that to not behave well. Renovate runs once a month and is intended for external dependencies, not packages inside the monorepo build. 

I'd still prefer managing that change on a separate PR.

rajsite: -I want it separate because of it acts weird it is easier to revert
- I want you to validate that it doesn't pull from Nuget in a local build but is resolving to the build in the workspace

rajsite: Even just syncing the branch and running npm install causes dirty changes because the lock file is out of sync:
image

Soo yea, please follow the initial suggestion and don't make the change in this PR.

m-akinc: Removed the dependency. Created #2047 to track the tech debt.

change/@ni-nimble-blazor-92675eea-194d-4c8d-81ea-ff1049cd4f6c.json
rajsite, line 1: Since we are modifying the scripts they depend on make sure to manually test the Blazor server and Blazor Hybrid example apps locally for regressions. They are not tested on CI.

packages/nimble-blazor/NimbleBlazor/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@ni/nimble-blazor",
+  "version": "16.0.4",
+  "description": "Blazor components for the NI Nimble Design System",
+  "scripts": {
+    "pack": "cross-env-shell dotnet pack -c Release -p:PackageVersion=$npm_package_version --output ../dist",
+    "invoke-publish": "npm run invoke-publish:setup && npm run invoke-publish:nuget && npm run invoke-publish:noop -- ",
+    "invoke-publish:setup": "rimraf --glob \"dist/NimbleBlazor.*.nupkg\" && npm run pack",
```
rajsite, line 8: You probably already did it, but JIC make sure to run the `invoke-publish:setup` commands locally to make sure they have the expected behavior in both packages, i.e. manually bump the version in the package.json and verify the output packages are using it.

change/@ni-nimble-blazor-92675eea-194d-4c8d-81ea-ff1049cd4f6c.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: Reviewed the build scripts and package changes. Relying on blazor owners for C# proj, sln, nuget, etc changes

rajsite (approved):

packages/nimble-blazor/package.json
```diff
@@ -1,49 +1,39 @@
...
   "files": [
     "!*"
   ],
   "devDependencies": {
     "@microsoft/fast-web-utilities": "^6.0.0",
     "@ni/eslint-config-javascript": "^4.2.0",
     "@ni/nimble-components": "*",
+    "@ni/nimble-blazor": "*",
     "@ni/nimble-tokens": "*",
+    "@ni/spright-blazor": "*",
```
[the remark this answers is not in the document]
m-akinc: The build does `require.resolve('@ni/nimble-components')` (and `spright-components`) in order to copy the bundle. Doesn't that imply we should have a dependency?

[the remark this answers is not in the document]
rajsite: we should have a dependency on `@ni/nimble-components` and `@ni/spright-component` yes. My comment was specifically about `@ni/nimble-blazor` and `@ni/spright-blazor`.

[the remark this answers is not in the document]
m-akinc: Sorry, my brain read those as `@ni/*-components`.

msmithNI (approved):

packages/nimble-blazor/Tests/SprightBlazor.Tests.Acceptance/WebHostServerFixture.cs
```diff
@@ -0,0 +1,46 @@
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Xunit;
+
+namespace SprightBlazor.Tests.Acceptance;
+
+public abstract class WebHostServerFixture : IAsyncLifetime, IDisposable
```
atmgrifter00, line 7: There is enough implementation between this class and `PlaywrightFixture` (maybe others?) where it would be nice if we could leverage a single implementation between the Acceptance Test projects in NimbleBlazor and SprightBlazor. This likely calls for the creation of another project.

atmgrifter00 (approved):

packages/nimble-blazor/Tests/SprightBlazor.Tests.Acceptance/WebHostServerFixture.cs
rajsite: Changed this PR to only partially address the Blazor feature and captured the remaining tasks there: https://github.com/ni/nimble/issues/1976
@m-akinc going forward lets continue to do that where it's appropriate to split work in a different PR and make sure it's tracked for the goal of smaller more targeted PRs.

change/@ni-nimble-blazor-92675eea-194d-4c8d-81ea-ff1049cd4f6c.json
rajsite, line 1: Botched the merge, creating comment to prevent merge

PR #2037, opened 2024-04-25
Add parameterize suite

### LINKED ISSUE #1827: Create a parameterizeSuite test helper mechanism.
## 🧹 Tech Debt
As [discussed in this PR](https://github.com/ni/nimble/pull/1738#discussion_r1454205260), it would be useful for us to create a mechanism for running several independent tests under the same set of parameterized values.

A [recent PR](https://github.com/ni/nimble/pull/1762#discussion_r1489995511) had the need of this, but had to implement an isolated pattern for it.

mollykreis:
# Pull Request

## 🤨 Rationale

Create `parameterizeSuite` function within `jasmine-parameterize` package.

Resolves #1827 

## 👩‍💻 Implementation

- Refactored the existing internal `parameterize` function within `jasmine-parameterize` so that it can be called for specs and suites
- Create `parameterizeSuite` function within `jasmine-parameterize` package
- Leverage the new `parameterizeSuite` function in one place within `nimble-components`

## 🧪 Testing

- Extended the testing of `parameterize` to cover both spec and suite usages
- Copied the `parameterizeSpec` tests and adapted for testing `parameterizeSuite`
- Made sure the `parameterizeSuite` function worked as expected by using it within the `nimble-select` tests

## ✅ Checklist


mollykreis:
@m-akinc, will buddy this PR for me?

packages/jasmine-parameterized/src/parameterize.ts
```diff
@@ -0,0 +1,83 @@
...
+        ) as (keyof typeof overrides)[];
+        if (
+            !overrideNames.every(overrideName => testCaseNames.includes(overrideName))
+        ) {
+            throw new Error(
+                'Parameterized test override names must match test case name'
+            );
+        }
+        if (
+            defaultTestType === it
```
m-akinc, line 63: nit: It's a little weird that we're assuming `defaultTestType` is either `it` or `describe` (i.e. never `fit`, `xit`, `fdescribe`, or `xdescribe`). I spent a while trying to figure out a way to let the function use either `it` or `describe` as needed (instead of passing in a `defaultTestType`), but I guess you just can't do that without run-time types.

m-akinc (approved):

packages/jasmine-parameterized/src/parameterize.ts
mollykreis: I wasn't a huge fan of that either, but I thought the trade-off of being able to single-source this function for specs & suites was worth that concession, especially since this is an internal function. However, it would probably make sense to throw an error if the default type isn't either `it` or `describe` since we are making that assumption within the code.

packages/jasmine-parameterized/src/parameterize.ts
```diff
@@ -0,0 +1,83 @@
...
+): void;
+export function parameterize<T extends object>(
+    testCases: T,
+    test: (spec: Suite, name: keyof T, value: T[keyof T]) => void,
+    defaultTestType: Suite,
+    overrides?: {
+        [P in keyof T]?: SuiteOverride;
+    }
+): void;
+export function parameterize<T extends object, U = Spec | Suite>(
```
rajsite, line 42: Two nits/pref

- Can we make the first parameter the type selector? i.e. make defaultTestType first? Then it'll feel more like the parameterize specific options are first followed by everything else it delegates to instead of being in the middle
- Instead of defaultTestType being a reference to either it or describe could we use an enum or even just a union of two string options? `'spec' | 'suite'`. I think an enum / string union is easier to do an exhaustiveness check over while lots of different types fulfill the contract of `defaultTestType: U`

change/@ni-nimble-components-d6494d37-cc3a-4756-9bb7-c4a8ea4d89c7.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Looks like all references of `eslint-disable-next-line @typescript-eslint/no-loop-func` would be cleaned up with this helper. Could we address those in the PR?

change/@ni-nimble-components-d6494d37-cc3a-4756-9bb7-c4a8ea4d89c7.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
```
rajsite, line 2: Should share at the next UI working group!

rajsite (approved):

change/@ni-jasmine-parameterized-544a8033-fa3b-4201-872c-b6f070a48513.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Can you update the README to document the usage: https://github.com/ni/nimble/blob/4d4c614052de37bf6f6f42f2d4c55e6ac87de3dd/packages/jasmine-parameterized/README.md#usage

packages/jasmine-parameterized/src/parameterizeSpec.ts
```diff
@@ -0,0 +1,51 @@
+import { parameterize } from './parameterize.js';
```
rajsite, line 1: for all these files, parameterizeSpec.ts, parameterizeSuite.ts, parameterizeSpec.spec.ts, parameterizeSuite.spec.ts
can you switch to the kebab case convention? parameterize-spec.ts, parameterize-suite.ts, parameterize-spec.spec.ts, parameterize-suite.spec.ts

packages/jasmine-parameterized/src/parameterize.ts
mollykreis: I ended up updating this in a different way based on another review discussion.

PR #2042, opened 2024-04-26
Refactored wafer map experimental structure and update cycle 

munteannatan:
# Pull Request

## 🤨 Rationale

The experimental component part needed to be more centralized with it's internal state and configuration.

The update needed to consider the new asynchronous calls and preserving the state between subsequent calls.

## 👩‍💻 Implementation

Centralized all the state management under computations.
Created a snapshot mechanism to preserve data state between async calls.
Combined all the worker configurations in two objects to make a minimal number of async calls.
 
## 🧪 Testing

merged most of the tests in the computations spec. 

## ✅ Checklist

## Open Issues

- worker type export to the wafer. right now, I copy and paste the file contents
- better structure/ update cycle?


packages/nimble-components/build/generate-workers/index.ts
```diff
@@ -39,4 +42,10 @@ const renderFilePath: string = path.resolve(
     'matrix-renderer.ts'
 );
 
+const workersFilePath: string = path.resolve(
+    workersDirectory,
+    'types.ts'
+);
+
 writeFile(renderFilePath, fileContent);
+writeFile(workersFilePath, typesCode);
```
rajsite, line 51: > worker type export to the wafer. right now, I copy and paste the file contents

I think this approach seems okay enough for now, did you run into any issues / have any concerns with this approach?

atmgrifter00 (approved):

packages/nimble-components/build/generate-workers/source/types.ts
```diff
@@ -20,3 +20,38 @@ export interface Dimensions {
     width: number;
     height: number;
 }
+
```
rajsite, line 23: I'll be OOO for the next week starting tomorrow but discussed some high level feedback with @atmgrifter00 that would be worth having a discussion on:
- Can we avoid all the optional types below in RenderConfig and TransformConfig
- Can we keep a bunch of the new state added to the top-level wafer map in smaller classes that encapsulate their algorithms that are later accessed by createSnapshot instead of putting them all on top-level wafermap
- Corresponding to the above, there are a lot of non-null assertions being added, can we avoid adding more and start cleaning some up?
- Is there a plan for addressing the intermittent test bug: https://github.com/ni/nimble/issues/2104
  - It seems likely that the test became intermittent due to this recent change: https://github.com/ni/nimble/pull/2095

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -14,58 +16,69 @@ import type {
  * This setup is used in the wafer-map component to perform heavy computational duties
  */
 export class MatrixRenderer {
-    public columnIndexes = Int32Array.from([]);
-    public rowIndexes = Int32Array.from([]);
     public values = Float64Array.from([]);
     public scaledColumnIndex = Float64Array.from([]);
```
atmgrifter00, line 20: We should rename `scaledColumnIndex` to `scaledColumnIndices` (and similarly `scaledRowIndex`). The term "indices" is used elsewhere in Nimble to refer to a collection of indexes (as opposed to just "indexes"). The only area where the term "indexes" is used is within production code is wafer-map related implementation.

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -24,10 +24,17 @@ import {
...
-import { HoverHandler as ExperimentalHoverHandler } from './modules/experimental/hover-handler';
+import { HoverHandler as ExperimentalHoverHandler } from './experimental/hover-handler';
 import { ZoomHandler } from './modules/zoom-handler';
+import { Computations } from './experimental/computations';
+import type {
+    ColorScale,
+    Dimensions,
+    Margin,
+    RenderConfig
+} from './workers/types';
```
atmgrifter00, line 37: I feel like I'm missing something. Where is the `workers` folder?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -14,58 +16,69 @@ import type {
...
+        colorScale: undefined
+    };
+
+    private transformConfig: TransformConfig = {
+        transform: undefined,
+        topLeftCanvasCorner: undefined,
+        bottomRightCanvasCorner: undefined
+    };
+
+    public calculateHorizontalScaledIndex(columnIndex: number): number {
```
atmgrifter00, line 43: Why is this and the next function `public`? The only places that are calling these methods functionally are in this class. The tests against these function seem unnecessary if we are validating the public methods that are ultimately calling these methods.

packages/nimble-components/build/generate-workers/source/types.ts
munteannatan: I changed the intermittent test to check for NaN instead of a specific value

atmgrifter00 (approved):

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -216,27 +219,28 @@ export class WaferMap<
...
-                await this.workerRenderer.drawWafer();
-            } else if (this.waferMapUpdateTracker.requiresDrawnWaferUpdate) {
-                await this.workerRenderer.drawWafer();
+                this.computations.colorAndTextUpdate();
+                setupWafer = true;
             }
-        } else if (this.waferMapUpdateTracker.requiresRenderHoverUpdate) {
-            this.workerRenderer.renderHover();
+            const snapshot = this.createSnapshot();
+            if (setupWafer) {
```
rajsite, line 238: Instead of tracking a separate flag here, is it possible to add something to the wafermaptracker like requiresWaferSetup?

something like
```
    public get requiresWaferSetup(): boolean {
        return (
            this.requiresContainerDimensionsUpdate
            || this.waferMapUpdateTracker.requiresScalesUpdate
            || this.waferMapUpdateTracker.requiresLabelsFontSizeUpdate
            || this.waferMapUpdateTracker.requiresDiesRenderInfoUpdate
        );
    }
```

The point of the wafermaptracker is to encapsulate that kind of conditional flag tracking logic.

rajsite (commented):
Looking great, just looking to align with existing updatetracker patterns

rajsite (approved):

PR #2045, opened 2024-04-29
[HLD] Dropdown option groups

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #791

## 👩‍💻 Implementation

This is just an HLD for the feature.


packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+
+_Question: Should matching against a group be configurable?_
+
+#### Combobox Guidance
+
+It should be noted that when using groups with a `Combobox`, it is advised that each option still be unique (for both its display and value). This is predicated on the [decision](https://github.com/ni/nimble/issues/1168#issuecomment-1568733159) that the `Combobox` will change it's behavior for platforms like Angular, where the model value will _always_ be the text that is in the display. This will make it impossible to disambiguate between options that have the same display value under separate groups.
+
+## Alternative Implementations / Designs
+
+Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client desires to display the group for the selected option that will be managed on the client side, including _where_ the display would be.
```
m-akinc, line 70: Must the client have their own way of figuring out the group from the selected value (e.g. include the group label in the option value), or should we provide some API for getting the (selected item's) group label?

m-akinc (approved):

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: My thought is that we should _not_ provide such an API. The native APIs offer no such thing (not that this is in and of itself a definitive reason to _not_ provide it, but it is a data point). I certainly wouldn't want an attribute API where a client is providing essentially redundant values for everything inside a group. This would then suggest a method the `Select` would provide to return the group for a particular option (as it would need to be `ListOption`-derived, and not value, since we provide no uniqueness guarantees for values), which is not a pattern we should promote.

change/@ni-nimble-components-dcbb4630-7675-4f1f-b6cd-7c7698e74598.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Initial HLD draft.",
```
mollykreis, line 3: This change description should be updated

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+        <option value="audi">Audi</option>
+    </optgroup>
+</select>
+```
+
+The Nimble `Select` (and `Combobox`) will follow the same approach, and provide a new custom element, `nimble-list-option-group`, that mirrors the `optgroup` behavior/API:
+
+```html
+<nimble-select name="cars">
+    <nimble-list-option-group label="Swedish Cars">
```
mollykreis, line 34: I'm curious why you opted for a `label` attribute here rather than providing a slot (either default or named) that would be for the group name? Is it because that's the way the native select is designed, or is there some other rationale?

As a note, using a slot approach might be more future proof with other features (such as adding an icon next to the text).

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+## Alternative Implementations / Designs
+
+Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client desires to display the group for the selected option that will be managed on the client side, including _where_ the display would be.
+
+## Open Issues
+
+As called out above:
+
+1. Should the `ListOptionGroup` component support slotting content such as icons into named slots?
+    - My thought is that initiailly we don't need to support this until it is asked for. So far, we have seen no requests for this.
```
mollykreis, line 77:
```suggestion
    - My thought is that initially we don't need to support this until it is asked for. So far, we have seen no requests for this.
```

mollykreis, line 77: I agree that it doesn't seem like we need to support slotting icons into the group label right now if we don't have requests or visual design for this. However, we also shouldn't design ourselves out of making this easy to add in the future.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: Yeah, I think the slot approach is probably the right path forward. Will change.

rajsite: I think compatibility with the optgroup api surface is useful, so being able to support `label` is handy. Is there significant complexity in supporting both? i.e. us `label` if given, otherwise use a named slot?

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Links To Relevant Work Items and Reference Material
+
+-   [Interaction Design](./IxD.md#groups)
+-   [Sectioned Combobox (issue 791)](https://github.com/ni/nimble/issues/791)
+-   [Prototype branch Storybook](https://60e89457a987cf003efc0a5b-cvmxlqwglb.chromatic.com/?path=/docs/components-select--docs) ([code](https://github.com/ni/nimble/tree/groupable-select))
+
+## Implementation / Design
+
+The HTML native `select` supports option groups with the native `optgroup` element. Configuring the `select` to use the `optgroup` takes the following form:
```
rajsite, line 15: html optgroup does not support nesting. Would we also only support one layer of nesting?

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: While I agree that compatibility with the `optgroup` API can be beneficial, I also think we should probably pick one at first, since the APIs would not conflict with one another. My current preference is the slotting approach, as that would be more consistent with other Nimble APIs, despite this suggesting that for the label we would need a named slot (which is _not_ a current paradigm I am aware of).

Ultimately I would be fine with either approach, but I'm just not sure there is a lot of value in supporting both initially. I can be convinced otherwise and can likely be swayed with a gentle breeze.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: I do not see a reason to support nesting, at least initially. While it's easy to envision a use-case that could leverage nesting, I'm willing to use the native APIs as a support guide, until we have calls to alter that decision.

packages/nimble-components/src/select/specs/option-groups-hld.md
rajsite: I'd prefer to not wait for the "use case of compatibility" so if we just are supporting one initially I'd go for label. And I'll advocate that supporting both label and slot is probably not expensive (at least slot is probably the more expensive of the two).

And agreed that the default slot would be more inline with our other apis 👍

packages/nimble-components/src/select/specs/option-groups-hld.md
rajsite: Agreed, I'd call it out that it is unsupported and we should _maybe_ consider testing that edge case to make sure we don't behave poorly (I think it would be natural to assume nesting is supported).

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+The `ListOptionGroup` class API should be fairly lightweight:
+
+```ts
+public class ListOptionGroup {
+    @attr
+    public label?: string;
+}
+```
+
+_Question: Should we support slotting icons (other content?) into the group display?_
```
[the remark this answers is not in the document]
fredvisser: I agree - Doing a survey of other design systems – DevExtreme, Shoelace, Monday.com (Vibe), Gusto (Workbench), RedHat (PatternFly) – no one supports anything other than a string for the group. Let's keep things simple until we have a compelling reason to do otherwise.

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Filtering
+
+As mentioned in the IxD, when a user has filtering enabled, typing into the filter input will match both individual options (as normal), as well as _all_ options that are under a group that matches the filter text.
+
+_Question: Should matching against a group be configurable?_
```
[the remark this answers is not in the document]
fredvisser: I agree – I challenged the other POs on the requirement to search groups, and they continue to be interested in that capability. However, it's not the common pattern, so we should be as consistent as possible so that our users understand the behavior and capability.

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+## Alternative Implementations / Designs
+
+Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client desires to display the group for the selected option that will be managed on the client side, including _where_ the display would be.
+
+## Open Issues
+
+As called out above:
+
+1. Should the `ListOptionGroup` component support slotting content such as icons into named slots?
+    - My thought is that initially we don't need to support this until it is asked for. So far, we have seen no requests for this.
```
fredvisser, line 77: I agree - 👍

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,79 @@
...
+2. Should matching against a group name for filtering be configurable?
+    - I'm waffling on the importance of this. Because of the very loose matching we provide for the `Select` filtering (and likely the `Combobox` eventually), it will be easy to match against several groups, particularly when very few characters have been entered, resulting in a bunch of options in the dropdown that don't match the filter itself, requiring more characters to be typed than usual to filter it down to a reasonable set. It is unclear how laborious this will make it for the user.
```
fredvisser, line 79: Is it possible to be stricter with group text matching, to avoid this concern?

atmgrifter00: While certainly possible, I'm not sure if having differing matching heuristics between groups and option text will be confusing or if we can reasonably manage expectations with documentation. I'm willing to submit to the consensus here.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: So, to be clear, I think if we were to take the dual-approach (or just the slot approach), the slot approach would be to use a named slot. I'm not sure how the default slot approach would work:

Default slot approach:
```html
<nimble-list-option-group>
    Group Name
    <nimble-list-option>Option 1</nimble-list-option>
    <nimble-list-option>Option 2</nimble-list-option>
<nimble-list-option-group>
```
I suppose the above would be feasible with some slot forwarding of the `nimble-list-option` elements (into a named slot in the `Select` template?), but is this a reasonable API for a client, versus:

Named slot approach:
```html
<nimble-list-option-group>
    <span slot="label">Group Name</span>
    <nimble-list-option>Option 1</nimble-list-option>
    <nimble-list-option>Option 2</nimble-list-option>
<nimble-list-option-group>
```
I suppose it won't take much effort to vet either approach out, so I guess my question is, which do we prefer?

packages/nimble-components/src/select/specs/option-groups-hld.md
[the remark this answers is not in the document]
atmgrifter00: Move to a 'Future Considerations' section.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: Moved this to a 'Future Considerations' section.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: Moved this to a 'Future Considerations' section.

packages/nimble-components/src/select/specs/option-groups-hld.md
fredvisser: Let's start with the same heuristic for filtering group and option text, and if we get feedback with real world data we can always tweak the heuristics to improve the user experience. No need to optimize prematurely…

m-akinc (approved):

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,92 @@
+# Select and Combobox option groups
+
+## Problem Statement
+
+It is not uncommon to want to categorize the options that are present in the dropdown for a `Select`/`Combobox` component. By supporting categories, users are then able to provide options that have the same display value, but under separate categories, helping avoid confusion.
```
fredvisser, line 5:
```suggestion
It is a common request to categorize the options that are present in the dropdown for a `Select`/`Combobox` component. By supporting categories, developers are able to logically group options, and help end-users to quickly find the relevant option. This will also support use cases that need the same display values under separate categories.
```

fredvisser (approved):

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,92 @@
...
+
+```html
+<nimble-list-option-group>
+    Swedish Cars
+    <nimble-list-option value="volvo">Volvo</option>
+    <nimble-list-option value="saab">Saab</option>
+</nimble-list-option-group>
+```
+
+If a client configures both the `label` attribute _and_ supplies slotted content, we will honor the slotted content.
```
rajsite, line 66: Is this arbitrary or is there a specific reason? I think implementation-wise it is easier to check if the label attribute is unset then a default slot (since any whitespace ends up in the default slot, [saw some devs complaining about that on twitter recently](https://twitter.com/LeaVerou/status/1785904086929346957)).

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Filtering
+
+As [mentioned in the IxD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/select/specs/IxD.md#groups), when a user has filtering enabled, typing into the filter input will match both individual options (as normal), as well as _all_ options that are under a group that matches the filter text.
```
rajsite, line 70: Don't think I see it in the IxD specfically, but will filtering all options from a group cause the group to be hidden or will the group headers persist or even group headers persist but indicate empty somehow?

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### Combobox Guidance
+
+It should be noted that when using groups with a `Combobox`, it is advised that each option still be unique (for both its display and value). This is predicated on the [decision](https://github.com/ni/nimble/issues/1168#issuecomment-1568733159) that the `Combobox` will change its behavior for platforms like Angular, where the model value will _always_ be the text that is in the display. This will make it impossible to disambiguate between options that have the same display value under separate groups.
+
+## Alternative Implementations / Designs
+
+Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client desires to display the group for the selected option that will be managed on the client side, including _where_ the display would be.
```
rajsite, line 78: > including _where_ the display would be.

@fredvisser maybe we tweak the language to say reach out to us for the use-case? My concerns are having usage guidance with more opinions about where or where not, i.e. don't try and push it into the select display text or icon slot or label, etc. Make it a separate visual next to the control, don't try and wedge it into the control, etc.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: I think my reasoning was along the lines of "We will honor the API that is more in-line with our preferred approach". However, I don't mind changing this. It does make the implementation more straightforward.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: I've added a remark stating that groups will be hidden when all options within the group have been filtered out.

change/@ni-nimble-components-dcbb4630-7675-4f1f-b6cd-7c7698e74598.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: nit: please update the PR title to clarify that this is an HLD, not an implementation

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,106 @@
...
+</nimble-select>
+```
+
+#### Visual separators
+
+The visual design calls for option groups to have a visual separator between them:
+
+![](./spec-images/option-group-separator.png)
+
+To accomplish this, a client will be required to insert an `<hr>` element between each group. This mirrors the API for the `Menu`.
```
jattasNI, line 51: This seems like something clients might commonly forget to do, leading to visual inconsistency. Can you add rationale for recommending this approach and/or add an alternative that says Nimble would do it automatically?

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,106 @@
...
+The `ListOptionGroup` class API should be fairly lightweight:
+
+```ts
+public class ListOptionGroup {
+    @attr
+    public label?: string;
+}
+```
+
+In addition to the `label` attribute, the `ListOptionGroup` will also support slotted labels. This will result in a configuration that looks like the following:
```
jattasNI, line 68: What's the rationale for supporting multiple APIs to provide a label? If we allow the slotted approach do we need to document / enforce restrictions?
1. can the slotted label be in any order or must it be before all options?
2. can you slot multiple labels?
3. can you slot a span containing a label? (This would likely be important to apply a targeted `i18n` attribute in Angular)
4. can you slot more complex HTML content like an icon plus a label?

My initial inclination is to avoid all these questions and only support the label attribute, but that may change when I hear the rationale for the slotted API.

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,106 @@
...
+
+At some point there may be a desire to provide extra visuals to a group. There already is such a [desire for the options](https://github.com/ni/nimble/issues/1135) themselves. However, there is no need to provide the API/implementation for this initially, as there is no current use-case demanding it.
+
+#### Hierarchichal groups
+
+It isn't difficult to imagine scenarios where a client may wish to have nested grouping within the `Select`. For example, a `Select` where a user could select a city, which was grouped under both State/Province/etc..., _and_ Country. However, this use-case is uncommon _and_ the native `select` does not support this behavior. We will consider supporting this when a known use-case arrives and there are no better alternatives.
+
+_Note: If a client constructs a DOM hierarchy with a group within another group we may need to provide some validity state on the `Select` explaining its invalidity, and we certainly need tests around this case to ensure we behave as intended._
+
+## Open Issues
```
jattasNI, line 106: What presence will the list option group have in the accessibility tree?

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,106 @@
...
+
+It is a common request to categorize the options that are present in the dropdown for a `Select`/`Combobox` component. By supporting categories, developers are able to logically group options, and help end-users to quickly find the relevant option. This will also support use cases that need the same display values under separate categories.
+
+## Links To Relevant Work Items and Reference Material
+
+-   [Interaction Design](./IxD.md#groups)
+-   [Sectioned Combobox (issue 791)](https://github.com/ni/nimble/issues/791)
+-   [Prototype branch Storybook](https://60e89457a987cf003efc0a5b-cvmxlqwglb.chromatic.com/?path=/docs/components-select--docs) ([code](https://github.com/ni/nimble/tree/groupable-select))
+
+## Implementation / Design
```
jattasNI, line 13: It'd be good to mention the plan for Angular and Blazor, even if it's just "no special considerations".

I scanned the [spec template](https://github.com/ni/nimble/blob/main/specs/templates/custom-component.md) and didn't see any other interesting missing sections besides a11y (mentioned below) but it couldn't hurt for you to review the template too in case anything else comes to mind.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: After the discussion that questioned how these elements would be removed when groups were filtered out, I'm now unsure in this as a direction. I'm going to reconsider what the right approach is.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: The slotted API approach was suggested because it was more consistent with other Nimble component APIs for how such things are configured (like a Button's text). The label attribute seemed like a snowflake API, despite its consistency with the native `optgroup` API. The multiple API approach was suggested by @rajsite I believe primarily to promote consistency with native APIs (which could be useful for those moving away from native components to our own) as well as acknowledging that to be consistent with other Nimble APIs would lean heavily towards a slotted approach.

Regarding your specific questions about what kind of restrictions we would have, I would expect we would be as lax with them as we are for the default slotted content in a button.

Admittedly, I don't feel too strongly with the approach we decide to take, and can be swayed easily (my original thought was just to do the `label` attribute).

jattasNI: We spoke offline and agreed:
1. keep both the attribute and default slot APIs
2. attribute would win over the slot
3. we'd document the common cases (single label before all the options, plus a span containing a label) and all the other cases above would be considered unsupported for now. They could become supported if people ask for them and we document them.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: I've added an 'Accessibilty' section that details this.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: The working decision is now to use the `label` attribute if present.

packages/nimble-components/src/select/specs/option-groups-hld.md
atmgrifter00: Done.

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,122 @@
...
+
+![](./spec-images/option-group-accessibility-tree.PNG)
+
+### Angular/Blazor Integration
+
+Standard implementation pattern for support.
+
+## Alternative Implementations / Designs
+
+Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client desires to display the group for the selected option that will be managed on the client side, including _where_ the display would be.
```
jattasNI, line 108: Double checking, are we satisfied that we provide good enough APIs for a client to implement this? I imagine it would look something like:
1. subscribe to the `change` event on the `select`
2. when it fires, use the event data or the value to find the `list-option` corresponding to the new value
3. navigate up the DOM of that option to find its group
4. get the title of that group either via `label` or `textContent`, depending on which one you set in your app

An alternative would be to add event data or helper methods to do some of this for the client. Probably low ROI, but worth considering.

jattasNI (approved):

packages/nimble-components/src/select/specs/option-groups-hld.md
fredvisser: 👍 - My plan with the usage docs is to require that apps do _not_ display the group information anywhere outside the dropdown, but I'm happy to hear new requirements…

```suggestion
Originally, there was a desire to display the group name for the selected option outside of the dropdown of the component. However, there is not precedent for this for other design systems that support groups, nor the native `select`. If a client would like to display the group for the selected option, please discuss this use case with the design system team.
```

rajsite (approved):

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,122 @@
...
+</nimble-select>
+```
+
+#### Visual separators
+
+The visual design calls for option groups to have a visual separator between them:
+
+![](./spec-images/option-group-separator.png)
+
+Unlike the `Menu`, which requires the client to slot `<hr>` elements under the `nimble-menu`, the visual separators between the groups will be generated automatically.
```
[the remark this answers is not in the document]
atmgrifter00: That is correct.

packages/nimble-components/src/select/specs/option-groups-hld.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Top-level Options
+
+While not a common use-case, clients can provide `nimble-list-option`s as siblings to `nimble-list-option-group`s. In such cases, the client will be expected to insert the appropriate `<hr>` element between a `nimble-list-option-group` and set of `nimble-list-option`s that immediately follow it.
```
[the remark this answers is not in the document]
atmgrifter00: This was incorrect. Thanks. The visual separators, again, will be part of the styling of the `list-option-group` elements.

PR #2050, opened 2024-04-30
Anchor menu item focus styling fix

### LINKED ISSUE #2028: nimble-anchor-menu-item getting outlined when the dropdown is opened
# 🐛 Bug Report

The first item in the menu option is outlined when it is an anchor-menu-item
image

## 🤔 Expected Behavior
No items should be highlighted when opening the menu.

## 😯 Current Behavior
The first item is outlined.

## 🔦 Context

We added two anchor-items `Dash boards` and `Data Spaces` under the `Visualize` menu to visualize the data tables

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2028

## 👩‍💻 Implementation

The root of the issue is that the focus outline was being shown any time focus was within the anchor menu item rather than being styled based on focus-visible. The difficulty is that the anchor menu item delegates focus to the inner anchor element, which means the anchor becomes focus-visible but the anchor menu item does not. Therefore, I reworked the styling a little bit so that the anchor element is the same size as the anchor menu item to allow the anchor to have the outline. A beneficial side effect of this change is that the click area of the anchor now exactly matches the hover & focus area of the menu item.

## 🧪 Testing

- Updated the menu button story to include an anchor menu item as the first item to verify the behavior
- Verified no chromatic changes were introduced

## ✅ Checklist


mollykreis (commented):
> In my SLE testing, I thought the keyboard focus ring on menu-items was consistent, on second look the keyboard focus takes a stop between menu items, and some of the items don't end up getting focus rings. Is this a separate issue?

@fredvisser, I couldn't reproduce the issue you are seeing in Chrome or Firefox. What browser were you using? Do you also see that without my change? I wouldn't have expected my styling change for the anchor-menu-item to change the behavior of the arrow keys.

fredvisser:
@mollykreis - Yep, it's a Safari specific issue… and not related to your change. I'll create a new bug to track it.

mollykreis:

packages/nimble-components/src/anchor-menu-item/styles.ts
```diff
@@ -55,6 +51,13 @@ export const styles = css`
         cursor: pointer;
         text-decoration: none;
         outline: none;
+        padding-left: 8px;
+        padding-right: 8px;
+    }
+
+    a${focusVisible} {
```
m-akinc, line 58: I was wondering why the old styling removed the ring when the menu item was pressed/active, since the new styling does not. The new styling is consistent with non-anchor menu items, so I'm not sure what the motivation was before. Anyway, seems correct now.

m-akinc (approved):

change/@ni-nimble-components-5102e613-18d0-4310-9a34-a0a1564227df.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: > A beneficial side effect of this change is that the click area of the anchor now exactly matches the hover & focus area of the menu item.

Doesn't look like the pointer matches the area though, it's subtle but you end up in a spot where it looks like you can click but nothing happens

pointershape

mollykreis: Good catch. That's specifically a problem with submenus due to the menu-item styling apply to the slotted menu. I've resolved the problem.

rajsite (approved):

PR #2067, opened 2024-05-02
Rename icon column to mapping column

mollykreis:
# Pull Request

## 🤨 Rationale

The icon column now supports various types of mappings (icon, spinner, and text), so we've decided that it should be renamed to the mapping column rather than the icon column.

## 👩‍💻 Implementation

- Rename the icon column, along with its cell view, group header view, and page object, to be called "mapping" rather than "icon"
- Update Angular and Blazor to be aligned with the new name
- Lots of updates to support the rename
- No functional changes other than renaming components, classes, files, etc.

## 🧪 Testing

- Sanity checked that the storybook docs, Angular example app, and Blazor example app all render as expected
- All existing tests still pass

## ✅ Checklist


mollykreis:

atmgrifter00 (approved):

rajsite (approved):

PR #2070, opened 2024-05-03
Table keyboard interaction (draft / partial implementation)

msmithNI:
# Pull Request

## 🤨 Rationale

#1137 ([HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md))

[Storybook build for this branch](https://60e89457a987cf003efc0a5b-pjtinxqdil.chromatic.com/?path=/story/components-table--table&args=data:HierarchicalDataSet;selectionMode:multiple)

## 👩‍💻 Implementation

The `KeyboardNavigationManager` class does the bulk of the work:
- Tracks what is currently focused / should be focused (`focusState` property which contains `focusType` - there are multiple `TableFocusType` values like `columnHeader`/`row`/`cell`/`cellActionMenu`/etc)
   - Note: If the user clicks in the table away from the current keyboard focus, `focusState` would end up out-of-date. There's code in `handleFocus()` to handle for this (see additional comments there)
- Listens for table keypresses to handle arrow key navigation, Tab/Shift-Tab presses
- If the user unfocuses and refocuses the table, re-focuses the appropriate elements in the table (`handleFocus()` again)
- Tab/Shift-Tab behavior should match the HLD / expectations for ARIA treegrid. (Basically, Tab/Shift-Tab go through the tabbable elements of the header row / current row, not including cells/column headers, until the end is reached, then the table blurs and elements before/after it will be focused with additional Tab presses.)
- Not implemented in this version:
  - Home/End/PageUp/PageDown keys
  - Ctrl-Enter / F2 to focus elements in a cell (action menu can still be reached via Tab/Shift-Tab)
  - Row selection via Space
  - Behavior when a row is focused, then user scrolls with the scrollbar, then keyboard focus returns to the table, is unexpected / probably not right

The overall approach of only setting tabindex=0 on the single element in the table we want focused is called [roving tabindex](https://www.freecodecamp.org/news/html-roving-tabindex-attribute-explained-with-examples/).


packages/nimble-components/src/anchor/index.ts
```diff
@@ -47,6 +48,10 @@ export class Anchor extends AnchorBase {
      */
     @attr({ attribute: 'contenteditable' })
     public override contentEditable!: string;
+
+    /** @internal */
+    @observable
+    public tabIndexOverride = 0;
```
msmithNI, line 54: Elements in the table that are focusable need a way for the table to override that behavior, for the roving tabindex approach to work. (This is for controls with `tabindex=0` explicitly on the control or something in its template, or that are implicitly focusable like `<a>`)

The current approach for this is:
- Add a `tabIndexOverride` property for those elements (defaults to `0` so that non-table uses work the same as they do now)
- In the element template, where `tabindex=0` was set previously, use `tabindex=tabIndexOverride` instead. (Note: Here in the anchor tabindex wasn't set explicitly before, but tabindex=0 is the default for `<a>`)
- In the templates inside the table, set `tabIndexOverride` to `-1` (i.e. not keyboard focusable by default)
- `KeyboardNavigationManager` will set the current focus in the table to `tabIndex`/`tabIndexOverride`=`0` explicitly before focusing it, and back to `-1` when it's unfocused.

packages/nimble-components/src/checkbox/template.ts
```diff
@@ -0,0 +1,45 @@
...
+    _: unknown,
+    definition: OverrideFoundationElementDefinition<CheckboxOptions>
+): ViewTemplate<Checkbox> => html<Checkbox>`
+    <template
+        role="checkbox"
+        aria-checked="${x => x.checked}"
+        aria-required="${x => x.required}"
+        aria-disabled="${x => x.disabled}"
+        aria-readonly="${x => x.readOnly}"
+        tabindex="${x => (x.disabled ? null : x.tabIndexOverride)}"
```
msmithNI, line 23: Note: This is FAST's template as-is except for the `tabindex` line.

packages/nimble-components/src/table-column/base/cell-view/index.ts
```diff
@@ -29,6 +29,14 @@ export abstract class TableCellView<
     @observable
     public recordId?: string;
 
+    /**
+     * Gets the child elements in this cell view that should be able to be reached via Tab/ Shift-Tab,
+     * if any.
+     */
+    public get tabbableChildren(): HTMLElement[] {
```
msmithNI, line 36: If a table column type will contain focusable/tabbable elements, our guidance would be:
- That column's cell view overrides this property, and returns the focusable element(s) in the array
- The focusable elements should be tabindex=-1 by default

`KeyboardNavigationManager` uses this when handling Tab/Shift-Tab for a given row, and to reflect the correct focus state if one of these elements is focused (`TableFocusType.cellContent`).

Currently `KeyboardNavigationManager` updates tabindex of these elements directly (except for the `tabIndexOverride` cases described previously). For arbitrary cell content, this might not be sufficient, since the element that should be tabindex=0 could be an element descendant in a shadow root or something similar. So we may need to add an additional API/method to handle focusing/unfocusing these children.

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -119,13 +126,38 @@ export const styles = css`
         --ni-private-table-cell-action-menu-display: block;
     }
 
+    nimble-table-cell${focusVisible} {
+        --ni-private-table-cell-action-menu-display: block;
+    }
+
+    /* TODO: Scope this to only hasDataHierarchy; ~4px more padding on left, so border doesn't touch expand/collapse button border */
+    nimble-table-cell:first-of-type${focusVisible} {
```
msmithNI, line 134: (Note: These 2 rules are handing the special case of making the cell border for the 1st column / hierarchy column wide enough to include the expand/collapse button. It probably needs another cleanup pass - I couldn't come up with something that worked with just margin/padding on the cell, but I got something working adding a `::before` pseudoselector too).

packages/nimble-components/src/table/template.ts
```diff
@@ -33,6 +33,7 @@ import {
 export const template = html<Table>`
     <template
         role="treegrid"
+        tabindex="0"
```
msmithNI, line 36: As everything in the table will now start as tabindex=-1, this tabindex=0 ensures the table can be keyboard focused initially. The table remains as tabindex=0 for almost all (see `KeyboardNavigationManager.blurAfterLastTab()` for an exception)

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+        private readonly table: Table<TData>,
+        private readonly virtualizer: Virtualizer<TData>
+    ) {
+        table.addEventListener('keydown', e => this.onCaptureKeyDown(e), {
+            capture: true
+        });
+        table.addEventListener('keydown', e => this.onKeyDown(e));
+        table.addEventListener('focusin', e => this.handleFocus(e));
+        table.addEventListener('focusout', e => {
+            console.log(
```
msmithNI, line 68: (These `console.log`s are just temporary/for debugging, for anyone that wants to get a better sense of what's getting focused/unfocused while interacting with the table in Storybook. It's not always obvious what's getting focused.)

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+            source === this.table
+            && args === 'rowElements'
+            && this.hasRowOrCellFocusType()
+            && this.inNavigationMode
+        ) {
+            this.focusCurrentRow(false);
+        }
+    }
+
+    public onVirtualizerChange(): void {
```
msmithNI, line 104: This is how we handle re-focusing the appropriate row after scrolling / virtualizer changes. Observing the `virtualizer.visibleItems` is another option, but we'd also need to observe each individual row and monitor for `dataIndex` changes.  
Possibly keeping a reference to the currently focused row (so we could just watch for `dataIndex` changes) is another option.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+        if (this.hasRowOrCellFocusType() && this.inNavigationMode) {
+            this.focusCurrentRow(false);
+        }
+    }
+
+    private readonly handleFocus = (event: FocusEvent): void => {
+        // User may have clicked elsewhere in the table (on an element not reflected in this.focusState). Update our focusState
```
msmithNI, line 111: As indicated by the code comments, I'm not super satisfied with all of this logic being in the `focusin` handler, I think it makes it harder to reason about the various focus/unfocusing states of the table.

If we only had to worry about keyboard navigation by itself (without mouse interactions), the responsibilities of this class would be pretty simple - basically just a state machine: `current focus state + a key press => new focus state`.  
But in practice, the user can click anywhere on the table (like interacting with an action menu), and then we still want keyboard navigation to behave in a sane way.

Here's 2 action menu scenarios that behave differently, that led to the current logic in this function:
- Table is not focused at all, then the user clicks a cell action menu button. Result: first the `MenuButton` is the active element, then a `MenuItem` is active (once the menu is open)
- Table is focused (via Tab), the 1st row is focused (Down Arrow), then the user clicks the action menu button on some other row (not the 1st). Result: the `MenuItem` is active. I didn't see any focus events get triggered with the actual MenuButton in this scenario, which is why I have code going from a `MenuItem` back to the linked action menu.

This function also handles setting the appropriate element in the table on initial focus, so there's a fair amount going on in this.

I think there might be better ways to handle all of this, so I'm looking for suggestions on different approaches here too.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+
+    private blurAfterLastTab(activeElement: HTMLElement): void {
+        // In order to get the desired browser-provided Tab/Shift-Tab behavior of focusing the
+        // element before/after the table, the table shouldn't have tabIndex=0 when this event
+        // handling ends. However it needs to be tabIndex=0 so we can re-focus the table the next time
+        // it's tabbed to, so set tabIndex back to 0 after a rAF.
+        // Note: In Chrome this is only needed for Shift-Tab, but in Firefox both Tab+Shift-Tab need this
+        // to work as expected.
+        this.table.tabIndex = -1;
+        window.requestAnimationFrame(() => {
```
msmithNI, line 661: `requestAnimationFrame` calls always need vetting/questioning, so opening a comment thread here. See code comments for why this is here (if the table was still tabindex=0 when the user did Shift-Tab, the overall table would get focused which isn't what we want).

I'm also not aware of a good way to determine what the next or previous focused element will be, if we wanted to focus it programmatically (the browser computes it based on semi-complicated tab order rules).  
So this code is just trying to ensure the browser can handle it: we blur the focused table sub-element, and then we don't `preventDefault()` on the Tab/Shift-Tab key event, so the browser will do the default behavior still (of focusing the next/previous element in tab order).

msmithNI:
@rajsite @jattasNI @atmgrifter00 I think I've now commented on the main areas of this code that I'd like feedback on, so feel free to start looking it over now. If there's any demos/walkthroughs of it that you'd like to see to understand something better (or any discussions that you think would be better as Teams calls) we can do that next week.

packages/nimble-components/src/anchor/index.ts
rajsite: I don't think I see from this description why we need a different name for tabindex. Fundamentally we want the components to match the native APIs as closely as possible. Why can't we implement the tabindex api correctly?

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+                            this.focusState.columnIndex = this.table.visibleColumns.indexOf(cell.column!);
+                            this.focusState.cellContentIndex = contentIndex;
+                            this.inNavigationMode = false;
+                        }
+                    }
+                }
+            }
+        }
+
+        if (event.target instanceof MenuItem) {
```
atmgrifter00, line 159: I'm guessing it's not possible to enter into this `if` block if above `if` block was entered? If not, can we make it more clear by making this an `if/else if` block? If it _can_ make sense to enter into both `if` blocks, are there ways for us to avoid duplicate efforts (possibly an over-optimization as iterating over the columns isn't a big deal).

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+                    this.focusState.focusType = TableFocusType.headerActions;
+                    this.focusState.headerActionIndex = this.focusableHeaderElements.headerActions.length - 1;
+                    this.focusHeaderElement();
+                    return true;
+                }
+                return false;
+            case TableFocusType.row:
+                if (this.isRowExpanded(focusedRow) === true) {
+                    this.toggleRowExpanded(focusedRow!);
+                    this.focusRowElement(focusedRow!);
```
atmgrifter00, line 278: Just curious, but in this case isn't the row already focused? What's the reason for setting focus again? I think this question applies for other cases as well.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+                    default:
+                        break;
+                }
+                let cellContentIndex = 0;
+                if (this.focusState.focusType === TableFocusType.cellContent) {
+                    cellContentIndex = this.focusState.cellContentIndex! + 1;
+                }
+                while (
+                    cellIndex >= 0
+                    && cellIndex < this.table.visibleColumns.length
```
atmgrifter00, line 501: Maybe we can make a couple of private helper methods to help with code readability (reduce function length). Somethine like `setNextTabFocusState(cellIndex: number)` and `setPreviousTabFocusState(...)`? Possibly even `getNextTabCellIndex` and `getPreviousTabCellIndex` that just encapsulates the switch logic.

packages/nimble-components/src/table/components/cell/styles.ts
```diff
@@ -36,4 +43,8 @@ export const styles = css`
         height: ${controlSlimHeight};
         align-self: center;
     }
+
+    .action-menu.focused {
+        display: block;
```
atmgrifter00, line 48: Can you explain why the `display` needs to change when the action menu is focused?

msmithNI: The `focused` class is added on the action menu button immediately before we call `focus()` on it. From what I can tell, `focus()` has no effect if the menu button is not visible when it's called, and we hide them by default.

As you Tab through potentially multiple menu buttons on the row, only the menu button that's focused (or associated with the focused cell) is shown:
MenuButtonFocus

packages/nimble-components/src/anchor/index.ts
msmithNI: At the time I'd added this, I'd run into issues trying to set `tabIndex` directly for these controls, due to them already setting `tabIndex` in their templates. I can take another stab at getting it working though.

Matching [what we do for the AnchorMenuItem](https://github.com/ni/nimble/blob/c1c3f73d50d327368f8aba56206208e93c22cfd1/packages/nimble-components/src/anchor-menu-item/index.ts#L73) (overriding `setAttribute()` and `get/set tabIndex`) didn't seem to be sufficient.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
...
+            // TableCell.onActionMenuButtonBlur() ensures that the CSS class is removed when the action menu
+            // button is blurred in all cases.
+            if (focusable) {
+                menuButton.classList.add('focused');
+            } else {
+                menuButton.classList.remove('focused');
+            }
+        }
+
+        if ('tabIndexOverride' in element) {
```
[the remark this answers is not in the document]
msmithNI: At the time that I'd added `tabIndexOverride`, setting `tabIndex` on the controls (where they're consumed from the table/row/cell templates) wasn't working, because the controls already set `tabIndex` in their own template and that was taking precedence. It was mainly added so that those controls would start out as `tabindex=-1` when used from the table, before they're focused via the keyboard.

I need to take another stab at getting everything working just via the existing `tabIndex` property though (Milan made the same comment).

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: You're right, this particular call was a no-op, I'll remove it. Calling `focusRowElement()` again is only needed when we're focusing something new inside the row.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: Agreed. I'll want to do another pass on code sharing/ helper methods if people are otherwise satisfied with the general approach on tracking focus state that I have here. I wanted to see if there were any large course corrections needed first though.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -0,0 +1,1025 @@
+/* eslint-disable no-console */
```
atmgrifter00, line 1: Here are the set of behaviors that I noticed that I believe are not ideal (unexpected):

Table with hierarchy data (single selection mode):
- If the first action menu has focus, I think pressing Shift-Tab should focus the row.
- If an action menu has focus (or any focusable element), pressing the Left or Right arrow keys should shift focus to the previous or next cells respectively if there are no other focusable elements in the cell.
- If the table has ever moved focus to something like an action menu, then after you click out of the table, and then tab to it, it will skip the header content and refocus the action menu. This means that it is no longer easy to just tab past the table at the top-level.

Table Anchor Column
- The anchor column cells do not seem like they can show the highlighted cell border. I admit that this may be intentional, and not sure if there is a better answer.

Possibly more to come...

packages/nimble-components/src/anchor/index.ts
```diff
@@ -1,4 +1,4 @@
-import { attr } from '@microsoft/fast-element';
+import { attr, observable } from '@microsoft/fast-element';
```
jattasNI, line 1: I noticed an unexpected behavior while testing:
1. tab focus to the table
2. down arrow a few times to a row
3. tab to give focus to an action menu in that row
4. enter to open the action menu, arrow to an item, enter to select it
5. shift tab to navigate between action menus

Unexpected behavior: it requires two shift-tabs to move focus from the first column's action menu to outside the table. If you skip step 4 it behaves as expected, where the first shift-tab moves focus from the first action menu to outside the table. Once it enters the bad state it seems to stay in that state until a page refresh.

Tested in macOS Firefox.

jattasNI, line 1: I'm not able to get row selection to work. From the HLD:
>If a row is focused, pressing Space will select/unselect that row 

When I focus an entire row, space either does nothing or scrolls the page. I tried this with both single select and multi select mode on the hierarchical table (again in macOS Firefox).

In multiselect mode I am able to focus the selection checkbox and press Space to toggle it.

packages/storybook/src/nimble/table/table.stories.ts
```diff
@@ -28,11 +28,23 @@ const metadata: Meta<BaseTableArgs> = {
     parameters: {
         actions: {
             handles: [
+                /*
```
jattasNI, line 31: Shot in the dark, but I recall an issue with the Actions add-on where it can get broken / slow trying to print large / complex event data. I think I recall fixing it by limiting the depth of the event data serialization. I can help search for that if you think that might be the issue and can't discover how to configure it on your own.

jattasNI (commented):
Early feedback from some storybook testing. Haven't looked at the code much yet.

packages/nimble-components/src/anchor/index.ts
msmithNI: Just not yet implemented (updated the PR description), but should be straightforward.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: Anchor Column: Intentional. This is the case of a cell containing a single focusable control, in which case we only focus the control, not the cell (according to the HLD).

I think our current anchor focus styling makes it a bit difficult to tell when the anchor is focused (at least compared to the focused cell styling), so we might end up revisiting that behavior. (I can demo it to the team and get opinions)

packages/storybook/src/nimble/table/table.stories.ts
jattasNI: Yep, that's what I was recalling. Too bad it doesn't help 😔

packages/nimble-components/src/anchor/index.ts
msmithNI: I can't reproduce this on Firefox in Windows, so you might have to demo it to me.

Can you tell where tab focus is after the first Shift-Tab, before you have to do it a 2nd time?

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: Noticed while demoing to Milan / Meyer today that this simplified approach has issues once you start scrolling down with just DownArrow (you end up with no focused row visible).
Adding a rAF works but is hacky, so I'll probably switch back to the original approach mentioned in the comment above.

packages/nimble-components/src/anchor/index.ts
jattasNI: Tab focus remains on the action menu button after the first Shift-Tab.

I tried in macOS Edge and didn't see the same bug.

I tried in macOS Safari and saw a different bug: regardless of whether you open any of the action menus, Shift-tab from the first action menu takes you to the focusable content AFTER the table rather than before. Similarly, tab from the last action menu takes you to the focusable content BEFORE the table rather than after. 

Happy to jump on a call and debug together, just ping me when you're ready to look at it.

msmithNI: Gave Jesse a new build to try and I think I have a fix for this behavior (will be in a future update).

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: I'm not sure if I agree on the 1st 3 points (behavior after an action menu is focused). To summarize the current behavior:  
Once an action menu is focused, you're no longer in row/cell navigation mode. You can return to that mode by pressing Esc, though. I'd characterize being able to press LeftArrow / RightArrow to go from a focused action menu to cell focus again as a behavior change from what's in the HLD, but if others agree I can look at adding that.

> If the table has ever moved focus to something like an action menu, then after you click out of the table, and then tab to it, it will skip the header content and refocus the action menu. This means that it is no longer easy to just tab past the table at the top-level.

I'm not sure how I'd reconcile that suggestion with our desire to return to the previously focused row once you tab past the table, and then tab back to it. Currently I'm characterizing interacting with the action menu the same way as interacting with a row/cell (in that we want to return focus to that same place when tabbing past and returning). I'm curious if others have an opinion on this point.

jattasNI: >our desire to return to the previously focused row once you tab past the table, and then tab back to it

I think that workflow is important. It'd be annoying to navigate through a bunch of rows and columns to get to the cell you want, then accidentally press tab or click one extra time causing the table to lose focus and all your navigation to be lost.

I'm ok with the current behavior, but would be open to ideas for returning the table to its initial state where it only has one tab stop. Right now the way I know to do it is to reverse all the navigation that got you to your current state (ESC to unfocus action menu, arrows to return focus to row / header). As an example idea, maybe if you keep pressing ESC the focus should move up the hierarchy (action menu -> cell -> row -> header)?

But I'm wary of designing this on the fly and increasing scope too much before we get this in the hands of actual users.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
jattasNI: A simple improvement might be to do a "clean code" pass and see if you can move each of the `if` blocks into their own well-named functions.

Another idea is to make `focusState` into a class rather than an object with individual fields being mutated. That might help make the state transitions more explicit. This might look something like:
1. make all the fields of the class `readonly`/getters from the outside
2. add public methods to set state in batches. e.g. 
```
focusCellActionMenu(rowIndex, columnIndex) {
   this.focusType = TableFocusType.cellActionMenu;
   this.rowIndex = rowIndex;
   this.columnIndex = columnIndex;
}
```
3. see if those public methods can enable this class to be even more explicit about state transitions. Maybe each one can be called from a navigation manager method that also updates the navigation manager's state.

I only looked at a couple of state transitions so I'm not sure how much this would help, might require prototyping one or two states in a branch to see if we like it better.

jattasNI: Ah, I see Meyer made similar suggestions in https://github.com/ni/nimble/pull/2070/files#r1591068527

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
jattasNI: I'm ok with the overall architecture of managing the state from this class. For me the state machine readability is the biggest course correction. I left some brainstorm-y ideas on that in this thread: https://github.com/ni/nimble/pull/2070/files#r1595622740

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
atmgrifter00: Yeah, I'm okay with the tab behavior. More than anything I wanted to check that we were fine with the `Table` entering into a state where it is no longer to just tab past it at the top-level. The AzDO table avoids this by not letting you tab through focusable elements in a row, and instead relies on the arrow keys (as it does not let you put cell focus in columns with no focusable elements). Screen readers like NVDA allow a user to navigate to non-focusable columns with Ctl-Arrow keys, which does not work without an active screen reader.

[This example](https://warsus.github.io/fancytree/demo/sample-aria-treegrid.html), (which I recognize is just some jo-shmo's idea of a treegrid) does what is currently implemented in ours.

For the first point on `Shift-Tab` putting focus on the row when the first action menu is focused, I think I just needed to become accustomed to pressing `Esc` to re-enter navigation mode.

packages/nimble-components/src/anchor/index.ts
msmithNI: Created a separate work item for the tabIndex work: #2094

msmithNI: Also removed all the tabIndexOverride code from this PR. Replaced with (temporary) `NotFocusableBehavior` and `getTabIndexTarget()`.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
msmithNI: Pushed the updates that remove that no-op call.

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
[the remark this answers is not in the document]
msmithNI: Removed all the tabIndexOverride code from this PR. Replaced with (temporary) `NotFocusableBehavior` and `getTabIndexTarget()`.

#2094 covers the tabIndex changes for this feature.

packages/nimble-components/src/anchor/index.ts
msmithNI: Pushed updates that should fix this.

packages/nimble-components/src/checkbox/template.ts
msmithNI: N/A (tabIndex changes separated out)

PR #2073, opened 2024-05-06
Control Minimum Width token applied to Select and Combobox

fredvisser:
# Pull Request

## 🤨 Rationale

The `sl-details-panel` restyling recently went in, and some apps use `nimble-select` and `nimble-combobox` controls in their details-panels. These dropdown controls were set with `min-width:250px`, and couldn't resize correctly down to 176px.

Before:
image

After:
image

## 👩‍💻 Implementation

- Added a new `controlMinWidth` token, set to 176px;
- Set the dropdown min-width to this value
- Updated Select and Combobox stories, so those instances didn't change sizes

## 🧪 Testing

- Manual testing of CSS changes with Work Orders app
- Storybook review
- Relying on Chromatic

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-a5b4c275-28d7-4509-af45-748fdd2f69fb.json
+++ b/change/@ni-nimble-components-a5b4c275-28d7-4509-af45-748fdd2f69fb.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added controlMinWidth token and set Select and Combobox controls to this new value (176px)",
+  "packageName": "@ni/nimble-components",
+  "email": "1458528+fredvisser@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/theme-provider/design-tokens.ts
+++ b/packages/nimble-components/src/theme-provider/design-tokens.ts
@@ -259,6 +259,9 @@ export const buttonBorderAccentOutlineColor = DesignToken.create<string>(
 export const controlHeight = DesignToken.create<string>(
     styleNameFromTokenName(tokenNames.controlHeight)
 ).withDefault('32px');
+export const controlMinWidth = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.controlMinWidth)
+).withDefault('176px');
 export const controlSlimHeight = DesignToken.create<string>(
     styleNameFromTokenName(tokenNames.controlSlimHeight)
 ).withDefault('24px');
--- a/packages/storybook/src/nimble/combobox/combobox.stories.ts
+++ b/packages/storybook/src/nimble/combobox/combobox.stories.ts
@@ -3,7 +3,6 @@ import { html, repeat } from '@microsoft/fast-element';
 import { withActions } from '@storybook/addon-actions/decorator';
 import type { HtmlRenderer, Meta, StoryObj } from '@storybook/html';
 import { listOptionTag } from '@ni/nimble-components/dist/esm/list-option';
-import { menuMinWidth } from '@ni/nimble-components/dist/esm/theme-provider/design-tokens';
 import { comboboxTag } from '@ni/nimble-components/dist/esm/combobox';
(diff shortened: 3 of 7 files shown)


packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -259,6 +259,9 @@ export const buttonBorderAccentOutlineColor = DesignToken.create<string>(
 export const controlHeight = DesignToken.create<string>(
     styleNameFromTokenName(tokenNames.controlHeight)
 ).withDefault('32px');
+export const controlMinWidth = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.controlMinWidth)
+).withDefault('176px');
```
rajsite, line 264: We have 3 listbox based controls, select, combobox and rich text mention.

rich text mention is already defined in terms of menuMinWidth which is the same value of 176px.
Instead of adding a new token can we instead keep all the listboxes in sync and use the menuMinWidth token? I think that the idea is that the menu styling and dropdown styling of select / combobox / rich-text are aligned so the use of menuMinWidth seems appropriate?

packages/storybook/src/nimble/combobox/combobox.stories.ts
```diff
@@ -104,7 +103,7 @@ const metadata: Meta<ComboboxArgs> = {
             appearance="${x => x.appearance}"
             value="${x => x.currentValue}"
             placeholder="${x => x.placeholder}"
-            style="width: var(${menuMinWidth.cssCustomProperty});"
+            style="width: 250px;"
```
rajsite, line 106: menuMinWidth is defined as 176px, not 250, isn't this a change that was being avoided?

https://github.com/ni/nimble/blob/9a1ebfa1220710515a4fca079f8b01932fa3e347/packages/nimble-components/src/theme-provider/design-tokens.ts#L307-L309

fredvisser: The instance of the `nimble-combobox` in Storybook specified a width of `176px`, but the component defined a `min-width:250px`, so the width was actually `250px` (and that works well for the Storybook preview). 

Since we're actually setting the min-width to `176px`, we're setting Storybook to `250px` so it's unchanged.

rajsite: Gotcha 👍

packages/nimble-components/src/theme-provider/design-tokens.ts
fredvisser: That was my first thought, but Brandon has consistently used `176px` as a minimum width for all controls, not just the dropdowns, so I could see the token being used across all components.

I assume it's a breaking change, should we just rename `menuMinWidth`? In a follow-on PR?

rajsite: Lets use menuMinWidth in this PR and save spreading it to controls and renaming it for future work / a tech debt. 

Most of our controls don't define it (and may not be trivial to, i.e. icon button mode of icon) and some are defining different values.

rajsite (approved):

packages/nimble-components/src/theme-provider/design-tokens.ts
jattasNI: Reminder to update the PR description to reflect that this PR is no longer adding a new token.

jattasNI (approved):

change/@ni-nimble-components-a5b4c275-28d7-4509-af45-748fdd2f69fb.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @fredvisser do we want to modify the matrix stories to specify width and avoid the visual diff or accept the changes with the new min-width? I don't think I feel strongly either way (but haven't look through the changes, there are a bunch)

rajsite: Actually scanning a couple things like this are making me a bit nervous about the stability of these:
image

PR #2074, opened 2024-05-07
Refactor common Blazor acceptance test code

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1976

## 👩‍💻 Implementation

I've created a new project in `blazor-workspace` to contain code shared between the Nimble and Spright acceptance test projects. I ran into issues with the page components not being found when running tests from the VS Test Explorer. Through trial and error, I discovered that the `Setup` instance needed to be created in the same assembly as the page components for them to be discovered. Normally this is governed by the `AppAssembly` property of the `Router` component (in `App.razor`), but that is not used by the test runner.

## 🧪 Testing

Tests pass.

## ✅ Checklist


packages/blazor-workspace/BlazorWorkspace.sln
```diff
@@ -27,6 +27,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SprightBlazor.Tests", "Test
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SprightBlazor.Tests.Acceptance", "Tests\SprightBlazor.Tests.Acceptance\SprightBlazor.Tests.Acceptance.csproj", "{8E335572-CD8B-4879-8760-73416CF103B1}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NimbleBlazor.Tests.Acceptance.Shared", "Tests\NimbleBlazor.Tests.Acceptance.Shared\NimbleBlazor.Tests.Acceptance.Shared.csproj", "{D94A4535-51B3-4E70-9582-3F3A44C7A798}"
```
m-akinc, line 30: I named the new project `NimbleBlazor.Tests.Acceptance.Shared`, which isn't great, since it isn't just for `NimbleBlazor` acceptance tests, but everything else I considered also had problems. I'm open to suggestions.

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/AcceptanceTestsBase.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace NimbleBlazor.Tests.Acceptance.Shared;
+
+[Collection(nameof(PlaywrightFixture))]
+public abstract class AcceptanceTestsBase
+{
+    private readonly PlaywrightFixture _playwrightFixture;
+
+    protected abstract Uri ServerAddress { get; }
+    protected abstract string ComponentLibraryInitializationTestJavaScript { get; }
```
m-akinc, line 12: These two abstract read-only properties are the only real difference from before the refactor.

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/MainLayout.razor
```diff
@@ -1,4 +1,5 @@
 ﻿@using Microsoft.AspNetCore.Components
+@using Microsoft.AspNetCore.Components.Web
```
m-akinc, line 2: Not sure why this wasn't needed before. I was getting various build errors and squiggles without it.

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/Startup.cs
m-akinc, line 1: I removed the `Configuration` property, because it didn't seem to be needed.

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/WebHostServerFixture.cs
```diff
@@ -20,6 +20,8 @@ public async Task InitializeAsync()
         ServerAddress = new Uri(addressFeature!.Addresses.First());
     }
 
+    protected abstract Startup StartupFactory(WebHostBuilderContext context);
```
m-akinc, line 23: Even though we don't use the `context` parameter, it's what the `UseStartup()` call below wants.

m-akinc:
@msmithNI would you mind buddying/first-reviewing this?

packages/blazor-workspace/BlazorWorkspace.sln
rajsite: Don't have a strong opinion since it's not a public assembly and I'm not familiar with conventions for C# namespaces or what the "problems" are in sharing (technical vs naming).
Some ideas:
- `BlazorWorkspace.Tests.Acceptance.Shared` (The goal of this assembly would be for sharing, seems redundant to add Shared in this case?)
- `BlazorWorkspace.Tests.Acceptance` (feels a little weird, like it should be tests, not test infra)
- `BlazorWorkspace.Testing.Acceptance` (like how we have Testing vs Tests in other packages)
- `Nimble.Testing.Acceptance`
- `NI.Testing.Acceptance`
- `Testing.Acceptance`

rajsite: Think I'm partial to `BlazorWorkspace.Testing.Acceptance` (tools of the workspace) or `NimbleBlazor.Testing.Acceptance` (tools shared by nimble blazor) but don't feel strongly

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/NimbleBlazor.Tests.Acceptance.Shared.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <PackageReference Include="NI.CSharp.Analyzers" Version="[2.0.21]" />
+    <PackageReference Include="System.ComponentModel" Version="4.3.0" />
+    <PackageReference Include="xunit" Version="2.8.0" />
+    <PackageReference Include="xunit.extensibility.execution" Version="2.8.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
```
rajsite, line 32: Make sure to set DisableImplicitNuGetFallbackFolder,

```
<PropertyGroup>
  <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
</PropertyGroup>
```

rajsite: @atmgrifter00 @msmithNI is there other configuration we need to add per project? We should make sure that is captured in contributing docs for when additional projects are added.

msmithNI: Good catch @rajsite .

I think the per-project configuration is something like:
- .NET version: should match other existing projects generally (`net6.0` for now)
- `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` and `<DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>`
- Reference to `NI.CSharp.Analyzers` (version spec should match other existing projects)

@m-akinc if you want to take a stab at putting that in the blazor-workspace contributing docs (in a new section for 'adding new projects') feel free, otherwise we (owners) can follow up on that point.

packages/blazor-workspace/BlazorWorkspace.sln
msmithNI: `BlazorWorkspace.Testing.Acceptance` seems reasonable to me, so I'd vote for that too.

(Having `NimbleBlazor` or `.Tests.` in the name doesn't seem right)

msmithNI (changes requested):
Everything looks mostly reasonable to me, I'd just like to take a look again after you rename the shared project. (Probably makes sense to ping Meyer to see if he's in agreement on the shared project name too, so you only need to do one update for that.)

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance.Shared/NimbleBlazor.Tests.Acceptance.Shared.csproj
m-akinc: Added to the contributing doc.

msmithNI (approved):

rajsite (approved):

packages/blazor-workspace/Tests/NimbleBlazor.Tests.Acceptance/SharedPlaywrightCollectionDefinition.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Xunit;
+﻿using BlazorWorkspace.Testing.Acceptance;
+using Xunit;
```
atmgrifter00, line 2: I guess I'm curious as to why this class couldn't move to the shared project.

m-akinc: If I move it to the shared project, we get the build error: `xUnit1041: Fixture argument 'playwrightFixture' does not have a fixture source (if it comes from a collection definition, ensure the definition is in the same assembly as the test)`.

atmgrifter00 (approved):

PR #2076, opened 2024-05-07
Improve Storybook API docs for menu component

jattasNI:
# Pull Request

## 🤨 Rationale

Continuing the work done in #2027 in service of #824. This PR covers the menu component.

Example of why this is needed: after documenting the `nimble-menu-item` `change` event I searched for usages in SLE and found dozens of instances where apps were using the keyboard inaccessible `click` event instead. I'm fixing them [in this PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/708606).

## 👩‍💻 Implementation

1. Refactored menu MDX and story files to follow patterns from previous PRs. I made some judgement calls about where to document child components vs examples; open to feedback on how it turned out!
1. Added more shared documentation strings and used them in several existing stories.

## 🧪 Testing

Manual storybook inspection

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/patterns/anchor/anchor-docs.ts
+++ b/packages/storybook/src/nimble/patterns/anchor/anchor-docs.ts
@@ -1,5 +1,5 @@
-export const hrefDescription = (componentName: string, includeDisable: boolean): string => `The URL that the ${componentName} points to. 
-${includeDisable ? 'To disable the control, remove the `href` attribute.' : ''}
+export const hrefDescription = (options: { componentName: string, includeDisable: boolean }): string => `The URL that the ${options.componentName} points to. 
+${options.includeDisable ? 'To disable the control, remove the `href` attribute.' : ''}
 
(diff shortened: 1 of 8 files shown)


packages/storybook/src/nimble/patterns/anchor/anchor-docs.ts
```diff
@@ -0,0 +1,5 @@
+export const hrefDescription = (componentName: string, includeDisable: boolean): string => `The URL that the ${componentName} points to.
```
rajsite, line 1: instead of boolean flags can you change this to an options object so it's easier to intuit what is being configured at the call site?

```
hrefDescription('anchor tree item', false)
```
vs

```
hrefDescription({ tagName: 'anchor tree item', includeDisableInstructions: false })
```

jattasNI: Good idea. I applied that pattern to the descriptions in `storybook.ts` too, just in case we find we need to parameterize them as we use them in other stories.

rajsite (approved):

fredvisser (approved):

PR #2077, opened 2024-05-07
Create nimble-mapping-empty component

mollykreis:
# Pull Request

## 🤨 Rationale

As part of #1992, we decided to create a `nimble-mapping-empty` component that would allow a cell to render as blank for a given mapping while still having text provided for use in group rows.

This PR creates that component within `nimble-components`. Support for Blazor and Angular will be added in a subsequent PR.

## 👩‍💻 Implementation

- Create `nimble-mapping-empty` component
- Add support for the component in the mapping column
- Update tests

## 🧪 Testing

- Extended the unit tests for the mapping column to include tests for empty mappings
- Add empty mapping to the matrix test for the mapping column

## ✅ Checklist


mollykreis:

packages/storybook/src/nimble/mapping/empty/mapping-empty.stories.ts
```diff
@@ -0,0 +1,35 @@
...
+import { hiddenWrapper } from '../../../utilities/hidden';
+
+const metadata: Meta = {
+    title: 'Internal/Mappings',
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    `The \`nimble-mapping-empty\` element defines a mapping from a data value to text. It is meant to be used as content of the \`nimble-table-column-mapping\` element
+                    when the table cell should not include the mapped value and the column is groupable.`
```
m-akinc, line 14: This mapping type is less obvious than others, so I think we should be more explicit about why a client would use this.
```suggestion
                    `The \`nimble-mapping-empty\` element can be provided as content of \`nimble-table-column-mapping\` to define data values that should render as empty table cells, but should be distinctly groupable. For each such data value, provide a \`nimble-mapping-empty\` element that maps the value to a distinct text value to group by.`
```

packages/storybook/src/nimble/mapping/empty/mapping-empty.stories.ts
```diff
@@ -0,0 +1,35 @@
...
+export const emptyMapping: StoryObj = {
+    render: createUserSelectedThemeStory(hiddenWrapper(html`<style></style>`)),
+    argTypes: {
+        key: {
+            control: false,
+            description: mappingKeyDescription('nothing')
+        },
+        text: {
+            control: false,
+            description: 'A textual description of the value. The table cells will render as empty, and the text will be displayed in a group header. This attribute is required.'
```
m-akinc, line 31: It looks like we have used the terms "group row" and "group label" in other documentation, so I went with "group row".
```suggestion
            description: 'A textual description of the value. Though this value is not displayed in the table cell, it determines how rows are grouped together and is displayed in the group row. This attribute is required.'
```

m-akinc (approved):

change/@ni-nimble-components-cde0ca9f-1259-4049-ae46-ead5829d5151.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: This change supercedes the need for a blank icon mapping right? Should we also revert the ability to specify an empty icon mapping? https://github.com/ni/nimble/pull/1874/files

packages/nimble-components/src/mapping/empty/index.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * Maps data values to text.
+ * One or more may be added as children of a nimble-table-column-mapping element. The
+ * mapping displays an empty cell and text-only group rows.
+ */
+export class MappingEmpty extends Mapping<MappingKey> {
+    @attr()
+    public text?: string;
```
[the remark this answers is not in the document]
rajsite: The `group-row-placeholder-empty` label is already used for values that are unmapped values. (i.e. enum mappings for 1,2,3 but you assign value 4). If you have a mapped value that you want to render blank I think providing a name makes sense, otherwise why bother?

> would probably rather we default the value of text to '',

Specifically to this implementation point, only boolean attributes should be assigned a value and that value is false. FAST will try to update the attribute to `text=""` and that can cause races with someone trying to set the attribute at the same time. 

If there is a use-case not being covered for allowing a mapping with the label not specified, we should use a different implementation strategy.

change/@ni-nimble-components-cde0ca9f-1259-4049-ae46-ead5829d5151.json
mollykreis: That's an interesting question. There is a slight difference in rendering between an empty mapping and an icon mapping without an icon. The matrix tests probably illustrate this better than I can explain, but basically, the group row reserves space for an icon on icon mappings but not for empty mappings. I'm not sure if one is better in some cases or not. I'll talk to Brandon and see if he has any thoughts on the alignment.

packages/nimble-components/src/mapping/empty/index.ts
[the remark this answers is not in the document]
mollykreis: > The group-row-placeholder-empty label is already used for values that are unmapped values. (i.e. enum mappings for 1,2,3 but you assign value 4).

That actually isn't the current behavior. We decided against having a default placeholder label on group rows for the mapping columns because we said that having values that didn't correspond to any mapping's key was incorrect usage of the column.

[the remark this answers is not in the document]
rajsite: Hmm, let's discuss it tomorrow. I'd like to understand what does happen for unmapped values and see if we want to loosen the restriction or have a placeholder mapping with no configuration which feels a bit weird.

packages/nimble-components/src/table-column/mapping/tests/table-column-mapping.spec.ts
```diff
@@ -27,22 +28,18 @@ interface SimpleTableRecord extends TableRecord {
...
+    icon?: string;
     textHidden?: boolean;
 }
 
-interface BasicTextMapping {
-    key?: MappingKey;
-    text?: string;
+interface SpinnerMapping extends BaseMapping {
+    textHidden?: boolean;
 }
```
rajsite, line 44: Can we postfix `Test` or something to these, `BaseMappingTest` `IconMappingTest` etc. I think just standalone it was a little confusing looking at the template.

packages/nimble-components/src/mapping/empty/index.ts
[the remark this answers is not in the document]
mollykreis: To capture the decision from our discussion:
- I will ensure that an empty string, null, or undefined value does not use any of the default placeholder strings.
- The existing behavior of showing nothing for unmapped values will be kept.

change/@ni-nimble-components-cde0ca9f-1259-4049-ae46-ead5829d5151.json
mollykreis: Talked with @fredvisser, and he said that text-only mappings would not be recommended with icon or icon+text mappings. Therefore, I'll plan to remove the blank icon mapping and leave the alignment of everything else as-is.

packages/nimble-components/src/table-column/mapping/tests/table-column-mapping.spec.ts
mollykreis: I've updated to add `Test` as a prefix since, at least to me, it makes more sense to call these "test mappings" rather than "mapping tests".

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -23,8 +24,8 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
 ### Best Practices
 
 -   Provide a mapping for every expected record value. Because grouping is performed on the record value, non-mapped record values can result in multiple groups without group labels.
-    -   To improve grouping behavior of values that don't correspond to icons, explicitly map those values to an `undefined` icon.
--   Avoid having multiple values that map to the same label because grouping and sorting is done on the record value rather than the mapped label.
+    -   To improve grouping behavior of a value that shouldn't be rendered in table cells, use a `nimble-mapping-empty` element to provide text for that value's group row.
```
rajsite, line 27: Can you use the `<Tag name={mappingEmptyTag}/>` helper for the element name

rajsite (approved):

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -23,8 +25,8 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
 ### Best Practices
 
 -   Provide a mapping for every expected record value. Because grouping is performed on the record value, non-mapped record values can result in multiple groups without group labels.
-    -   To improve grouping behavior of values that don't correspond to icons, explicitly map those values to an `undefined` icon.
--   Avoid having multiple values that map to the same label because grouping and sorting is done on the record value rather than the mapped label.
+    -   To improve grouping behavior of a value that shouldn't be rendered in table cells, use a <Tag name={mappingEmptyTag}/> element to provide text for that value's group row.
```
jattasNI, line 28: I didn't follow the discussion about when to use the empty mapping versus an icon mapping with no icon. Is this always the preferred approach? If so should we update the table-column-mapping story to use an empty mapping instead of the existing "Unknown" icon mapping?

jattasNI (approved):

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
mollykreis: We're going to remove support for icon mappings without an icon because we now have the empty mapping element. Therefore, empty mappings will always be the preferred approach. I can go ahead and update the story now, though, since even with both currently supported, the empty mapping would still be preferred.

I'm planning on removing `undefined` icon support from the icon column in a follow-up change since it's a breaking change, and I need to make adjustments to a few dependent apps before that change should go in.

Issue #1977, opened 2024-03-28
Spright Angular and Blazor example apps


m-akinc:
# 🙋 Feature Request

## 😯 Problem to Solve

Part of setting up Spright.

## 💁 Proposed Solution

- Angular Spright example app
- Blazor Spright example app

PR #2080, opened 2024-05-07
Add Spright rectangle component to Blazor example

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1977

## 👩‍💻 Implementation

Added the rectangle element to the bottom of the component in `Demo.Shared`. Nothing really noteworthy to point out.

## 🧪 Testing

Ran all of the example projects and made sure things rendered properly.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-blazor-b9a4025f-2fd1-4b3e-8295-540f02450064.json
+++ b/change/@ni-nimble-blazor-b9a4025f-2fd1-4b3e-8295-540f02450064.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add Spright rectangle component to Blazor example",
+  "packageName": "@ni/nimble-blazor",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
+++ b/packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
@@ -51,6 +51,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Include="CHANGELOG.json" />
     <None Include="package.json" />
     <None Include="packages.lock.json" />
   </ItemGroup>
--- a/packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
+++ b/packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
@@ -24,10 +24,13 @@
     <NoWarn>CS8669</NoWarn>
   </PropertyGroup>
 
+  <!-- 
+    Need to explicitly exclude json and config files from Content, or they will get copied to the output directory.
+    See: https://github.com/dotnet/sdk/issues/25587#issuecomment-1134731799
+  -->
   <ItemGroup>
-    <Content Remove="CHANGELOG.json" />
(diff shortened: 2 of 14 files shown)


packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
```diff
@@ -51,6 +51,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Include="CHANGELOG.json" />
```
atmgrifter00, line 54: Just curious, but what is the purpose of including this in the project?

atmgrifter00 (approved):

packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
rajsite: Also curious, i.e. there is also a CHANGELOG.md file in this folder, what files do we need to pick and choose for exclusion?

rajsite: There may potentially be a node_modules in the folder as well next to the package.json. Would that cause issues in the future?

m-akinc: I _thought_ that would prevent `CHANGELOG.json` from being copied to the output directory, which was ultimately causing a path conflict error when publishing the example. Turns out that didn't actually work, because despite VS showing that the file was set to "Do Not Copy", it sure as shit was getting copied. Eventually, I found [this](https://github.com/dotnet/sdk/issues/25587#issuecomment-1134731799), which after reading, I understood why we were already doing that for `package.json` and `package.lock.json`.

m-akinc: Milan, apparently (see link in previous response) we only need to explicitly exclude `.json` files, `.config` files, and `wwwroot`.

rajsite: Can you add a comment to the file referencing why specifically the json files are being excluded.

Also the `**\*.json` matcher does make me worry about node_modules folders where the comment may have been missed.

change/@ni-nimble-blazor-b9a4025f-2fd1-4b3e-8295-540f02450064.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Seems to be comment posting races so FYI make sure discussions resolved in this thread: https://github.com/ni/nimble/pull/2080#discussion_r1593182562

packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
rajsite: Can we just remove `**\*.json` instead of picking out individual json files?

m-akinc: Yes, that's a good idea. I verified that there aren't (currently) any `.json` files other than the three previously listed, so that should be safe for now. I decided to do the same thing for `.config` files, too, since the `node_modules` concern would apply to them, too.

rajsite (approved):

msmithNI (approved):

Issue #2084, opened 2024-05-08
Move angular-workspace into packages


rajsite:
# 🧹 Tech Debt

As apart of [creating SprightBlazor](https://github.com/ni/nimble/issues/1976) and forming blazor-workspace in the packages folder it was determined we likely could also move the angular-workspace to the packages folder for consistency.

- [x] Move `angular-workspace` into the packages folder. 
   -  As part of this also flatten the unnecessarily deep hierarchy, ie `/angular-workspace/packages/ni/nimble-angular` -> `/angular-workspace/nimble-angular`
- [x] Update external documentation (AzDo wikis) to reference the new nimble-angular location
- [ ] Ignore the [refactoring commit](https://github.com/ni/nimble/commit/e83c8ba9f06850acd6a303d257ec9033c0c5d115) in blame history: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

jattasNI:
Links are being updated in this PR so I checked off that task item. https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/749596

@rajsite I'll leave it up to you to decide whether you still want to do the git history cleanup or just close this out.

rajsite:
@jattasNI thanks for putting that up and naw lets close this

PR #2086, opened 2024-05-08
Move angular-workspace to packages folder

rajsite:
# Pull Request

## 🤨 Rationale

Moves `/angular-workspace` to `/packages/angular-workspace` and also shrinks paths within the workspace:
-  Moves `/projects/ni/nimble-angular` to `/nimble-angular`
-  Moves `/projects/ni/spright-angular` to `/spright-angular`
-  Moves `/projects/example-client-app` to `/example-client-app`

part of #2084 

## 👩‍💻 Implementation

Straightforward move and path updates. No refactors / other changes.
Marked change as patch to verify publish infra.

## 🧪 Testing

- Manually opened all the ng-package.json to verify VSCode detects the schema paths
- Rely on CI for the rest

## ✅ Checklist


packages/nimble-components/src/breadcrumb/specs/README.md
```diff
@@ -126,7 +126,7 @@ We have a few options:
         Custom App Page
     </nimble-breadcrumb-item>
     ```
--   (Alternate/Fallback option) We can prevent the RouterLink onClick from being called (with a click listener on a child element, or with a listener with useCapture=true). For an example/ prototype implementation [see the directives here](https://github.com/ni/nimble/tree/nimble-breadcrumb-prototype/angular-workspace/projects/ni/nimble-angular/src/directives/breadcrumb-item). (This prototype has the correct behavior for left-click, Ctrl-click, and middle-mouse click of breadcrumb item links, and href gets set correctly.)
+-   (Alternate/Fallback option) We can prevent the RouterLink onClick from being called (with a click listener on a child element, or with a listener with useCapture=true). For an example/ prototype implementation [see the directives here](https://github.com/ni/nimble/tree/d8d59a36a37baaccd0e45e9c6ad789023f48ce5d/angular-workspace/projects/ni/nimble-angular/src/directives/breadcrumb-item). (This prototype has the correct behavior for left-click, Ctrl-click, and middle-mouse click of breadcrumb item links, and href gets set correctly.)
```
rajsite, line 129: The only not completely mechanical change. Switched to a commit absolute url in the spec

.gitignore
```diff
@@ -124,3 +124,4 @@ package-lock.json
 
 # ignore renamed paths
 packages/nimble-blazor
+/angular-workspace
```
rajsite, line 127: Only other non-mechanical change. Ignoring the renamed path to ease transition for those switching between branches (any cruft like generated icons in old folders don't show up as local changes when sinking down)

PR #2087, opened 2024-05-08
Avoid Angular component ID collision test failure

m-akinc:
# Pull Request

## 🤨 Rationale

A build hit a test failure unrelated to the changes in that PR: https://github.com/ni/nimble/actions/runs/9009318573/job/24753280540?pr=2083

## 👩‍💻 Implementation

In one of the test files we forked from Angular (`router_integration_spec.ts`) there are two components, `AbsoluteSimpleLinkCmp` and `RelativeLinkCmp` that have identical `selector` and `template` values, which can lead to component ID collisions. We worked around this by adding an irrelevant `:not(head)` to `RelativeLinkCmp`'s selector, however it turns out that there is another definition of `RelativeLinkCmp` earlier in the file (with unmodified selector) that can still lead to an ID collision. I switched the selector "uniquification" to `AbsoluteSimpleLinkCmp`, so now it should not collide with either of the `RelativeLinkCmp` definitions.

## 🧪 Testing

Ran tests locally.

## ✅ Checklist


change/@ni-nimble-angular-e488db99-92c5-4954-a624-673ac1c31f32.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Why are these not issues for Angular? Is there something unique to us that is causing it to be a problem?

Why is it intermittent? How confident are we that it won't be a whack-a-mole?

m-akinc: Good questions. When we first forked that test file, we got an error about `AbsoluteSimpleLinkCmp` and `RelativeLinkCmp` potentially generating colliding IDs. I was surprised that Angular didn't have to work around it, too. It would take time to investigate, and I'm not sure it's worth it right now. Apparently, the issue is limited to those two components. This is an easy workaround to try, and if we never run into the error again, great. If we do, we can spend a little more time on it.

change/@ni-nimble-angular-e488db99-92c5-4954-a624-673ac1c31f32.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "none",
```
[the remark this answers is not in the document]
m-akinc: I assumed the `template` value was taken into account, too, in some form. Since `AbsoluteSimpleLinkCmp` and `RelativeLinkCmp` have nearly identical templates, and we've only had issues with them. But then again, the templates are not textually identical, and it's less likely Angular is using some parsed/normalized version of that string to determine equality. So maybe lots of these components have potential to collide. I would propose I make all of the selectors unique using `:not()`. That should avoid this for sure. Sound good?

change/@ni-nimble-angular-e488db99-92c5-4954-a624-673ac1c31f32.json
rajsite: I mean with that reasoning (we don't know why it's intermittent and we don't want to investigate unless it is frequently intermittent) then why is there any confidence in making the current change? Doesn't the same logic apply (ignoring the intermittent test policy for this specific case)?

Do the tests fail consistently without the "irrelevant :not(head)" selector? Maybe we should remove all of them to better characterize the intermittency? Maybe we should add them to all of them?

Might be worth a post stand-up discussion. Would like @jattasNI's take on it as well.

change/@ni-nimble-angular-e488db99-92c5-4954-a624-673ac1c31f32.json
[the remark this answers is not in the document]
rajsite: Replied in the other comment that we should have a discussion. I don't particularly like the idea that we have to make such big changes to forked tests we need to keep in sync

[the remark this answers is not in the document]
m-akinc: To address the original question, though the `template` string is not directly part of the [ID generation hash](https://github.com/angular/angular/blob/ace0a683f60eba3d30d11f241c59e37386b2b805/packages/core/src/render3/definition.ts#L698), it is indirectly involved via the included [number of bindings](https://github.com/angular/angular/blob/fedeaac8ba85002e34dbd6ffa682d7483d39b5fd/packages/core/src/render3/interfaces/definition.ts#L330), number of nodes, local refs, and pipes, etc. So we were only seeing the issue with those two components, because they were the only ones with the same selector and the same numbers of those other properties.

change/@ni-nimble-angular-e488db99-92c5-4954-a624-673ac1c31f32.json
m-akinc: Apparently Angular is setting the global property `ngDevMode` to `false` for their test runs. I couldn't find exactly where in their build/test infrastructure this is happening, but I'm very confident. The ID collision error is [only reported when `ngDevMode` is undefined or truthy](https://github.com/angular/angular/blob/ace0a683f60eba3d30d11f241c59e37386b2b805/packages/core/src/render3/definition.ts#L723). By adding `window.ngDevMode = false;` or `enableProdMode()` to the test file, I am able to avoid the ID collision error, but it interferes with other tests being run. They fail with errors like `Cannot redefine property: ɵfac` and `Cannot configure the test module when the test module has already been instantiated. Make sure you are not using ``inject`` before ``TestBed.configureTestingModule``.`. If I add an `afterAll()` that sets `ngDevMode` back to undefined, it reduces the number of failures down to a handful, but doesn't fully solve the problem.

It's also not fully clear to me if there are any potential downsides to running in production mode. Given that we now have an explanation for why the issue is limited to just a couple of the components in the test file (see my response in the other thread), I think the change in this PR is probably our best option.

rajsite: Why tests are intermittent is still a good question but the overview given at stand-up and here gives more clarity into why we see the issues and Angular tests don't. We can follow the suggestion for the updated tweak in the PR 👍

rajsite (approved):

PR #2096, opened 2024-05-10
Fix issue with clicking option in dropdown not updating display.

atmgrifter00:
# Pull Request

## 🤨 Rationale

[AzDO bug: Filterable nimble select does not select the option selected on click](https://dev.azure.com/ni/DevCentral/_workitems/edit/2738356)

## 👩‍💻 Implementation

Just a minor logic glitch that wasn't caught because an existing test wasn't robust enough.

## ✅ Checklist


jattasNI (approved):
I'm ok with you submitting this without @rajsite' review. Small logic change and high severity bug.

rajsite (approved):

PR #2108, opened 2024-05-15
Fix text field start slot color

mollykreis:
# Pull Request

## 🤨 Rationale

An icon put in the `start` slot of the `nimble-text-field` had the incorrect color based on the visual design spec. Specifically, the icon had been the text color, but it should have a 60% opacity applied to it.

## 👩‍💻 Implementation

Updated the styling of the `start` slot

## 🧪 Testing

Manually tested across all 3 themes in storybook

## ✅ Checklist


mollykreis:

packages/nimble-components/src/text-field/styles.ts
```diff
@@ -118,7 +117,11 @@ export const styles = css`
 
     slot[name='start']::slotted(*) {
         flex: none;
-        ${iconColor.cssCustomProperty}: currentcolor;
+        opacity: 0.6;
```
atmgrifter00, line 120: Just checking, but this means that _any_ content put in the start slot (including plain text) would get the 60% opacity applied to it. I'm guessing this is fine, as our guidance is that only icons should go in this slot. If we don't have explicit guidance along these lines should we add that to the docs?

atmgrifter00 (approved):

packages/nimble-components/src/text-field/styles.ts
mollykreis: Yes, that is correct. Any content put in the start slot would get these opacities applied.

Our docs say _To place an icon at the far-left of the text-field, set `slot="start"` on the icon._ Do you think that's explicit enough?

atmgrifter00: Yeah, it's probably fine. Ultimately, I expect our docs to evolve to describe usage guidance at a higher level than in those Storybook settings, which aren't as discoverable.

packages/nimble-components/src/text-field/styles.ts
```diff
@@ -118,7 +117,11 @@ export const styles = css`
 
     slot[name='start']::slotted(*) {
```
rajsite, line 118: We have a bunch of other controls that have a "start icon next to some text", i.e. various buttons (button, anchor button, menu button), menu items, table headers, mapping column, tree item, tooltip and banner (fixed severity icons).
Do we need to drive any consistency around those?

jattasNI (approved):

rajsite (approved):

PR #2110, opened 2024-05-15
Support tabindex overriding for Button, MenuButton, ToggleButton, Checkbox, and Anchor

### LINKED ISSUE #2094: Need ability to override tabIndex on controls
Dependency of: table keyboard navigation ( #1137 )

## 😯 Problem to Solve

Currently several of our controls have a hardcoded `tabIndex` in their template. This is problematic for the table keyboard navigation feature, when we need to programmatically change tabIndex (for 'roving tabIndex' as we programmatically focus only a specific element).

The controls we initially need this for are those used in focusable parts of the table:
- Button
- Checkbox
- MenuButton
- ToggleButton (since it's used by MenuButton internally)
- Anchor

Desired behavior:
- Can set the `tabIndex` property or the `tabindex` property of a control, and it's reflected correctly to the internal parts of the control that are currently hardcoded as tabindex=0 (note that some elements don't have tabindex="0" explicitly, but that behavior is part of their element type, e.g. `<a>` and `<button>`).
- Example: Set `tabindex="-1"` on a `nimble-toggle-button`. Result: the `<div role="button" />`  that was previously tabindex=0 is now tabindex=-1 too. Later, setting tabindex=0 on the toggle button will also set the inner div back to tabindex=0.
- Current behaviors maintained:
   - Focusable controls are still focusable by default (until set to tabindex -1)
   - A control can be set to `tabindex="-1"` in it's initial HTML representation (template), and it respects that after it's attached/connected
       - Example: `<nimble-checkbox tabindex="-1"></nimble-checkbox>`, add to DOM, checkbox.tabIndex should be -1 not 0 (this doesn't work currently)

## 💁 Proposed Solution

Current plan (based on prototyping) is to override the tabIndex property and add `@attr`: 
```ts
@attr({ attribute: 'tabindex', converter: nullableNumberConverter })
public override tabIndex!: number;
```
In templates, if a control contains another focusable control, it should set tabIndex on the inner control, i.e.
`tabindex="${x => x.tabIndex}"`
In cases where tabIndex is currently set conditionally if not disabled  
(`tabindex="${x => (x.disabled ? null : 0)}"`), we need something closer to  
`tabindex="${x => (x.disabled ? null : (typeof x.tabIndex === 'number' ? x.tabIndex : 0))}"`

For some of these controls, we haven't yet forked / brought in the template from FAST, which we'll need to do as part of this work.

## 📋 Tasks

- Implement for controls:
  - Button (note: we need to bring in the template from FAST, following our standard patterns, e.g. tests too)
  - Checkbox (note: we need to bring in the template from FAST, following our standard patterns, e.g. tests too)
  - MenuButton
  - ToggleButton (since it's used by MenuButton internally)
  - Anchor
- Autotests
   - tabIndex is forwarded to internal parts as needed
   - With an initial representation of `<element-name tabIndex="-1"></element-name>`, the element's tabIndex will be `-1` after it's connected
- Manually test focus behavior in Chrome + Firefox + Safari, make sure works as expected

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2094

## 👩‍💻 Implementation

For each of `Button`, `MenuButton`, `ToggleButton`, 'Checkbox', and `Anchor`, added `tabindex` attribute, updated the template to forward the `tabindex` value, and wrote automated tests. Additionally for `Button` and `Checkbox`, forked template from FAST and copied over the FAST tests.

## 🧪 Testing

Ran tests. Manually tested in Storybook in Chrome and Firefox.

## ✅ Checklist


packages/eslint-config-nimble/typescript.js
m-akinc, line 1: We are forbidding the use of `DOM` from `fast-element` in tests, but we should make an exception for tests copied from FAST.

m-akinc:
@msmithNI Will you buddy this for me?

packages/nimble-components/src/toggle-button/template.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-restricted-syntax */
```
m-akinc, line 1: I don't understand why it complains about the `??` operator (`Using 'LogicalExpression[operator='??']' is not allowed.`). I disabled the rule to get around it.

packages/nimble-components/src/anchor/tests/anchor.spec.ts
```diff
@@ -80,6 +80,25 @@ describe('Anchor', () => {
...
+
+            expect(element.control!.getAttribute('tabindex')).toBe('-1');
+        });
+    });
+
+    it('should clear tabindex attribute from the internal control when changed to null', async () => {
+        await connect();
+
+        element.setAttribute('tabindex', '-1');
+        element.setAttribute('tabindex', 'null');
```
msmithNI, line 98: Setting the attribute to `'null'` for the test (the actual string `'null'`) seems strange to me (I don't think any code paths should do that in practice)

Verifying that we can clear tabIndex seems like a valid test, but that would just be `element.removeAttribute('tabindex')` instead, for this line.

(Same comment applies to the other control tests)

packages/nimble-components/src/button/tests/button.foundation.spec.ts
```diff
@@ -0,0 +1,669 @@
+/**
+ * Based on tests in FAST repo: https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/button/button.spec.ts
```
msmithNI, line 2: Can you update this to be a permalink? (same for the checkbox one)

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
```diff
@@ -121,6 +121,26 @@ describe('MenuButton', () => {
             expect(element.toggleButton!.disabled).toBeTrue();
         });
 
+        it('should default tabIndex on the toggle button to 0', async () => {
+            await connect();
+            expect(element.toggleButton!.getAttribute('tabindex')).toBeNull();
```
msmithNI, line 126: The test name doesn't match the assertion here - maybe `should not set tabIndex on the toggle button by default`, or `should not set tabIndex on the toggle button if not set on the menu button`,

packages/nimble-components/src/toggle-button/template.ts
msmithNI: This is our rule, added in #2025 .

Based on the [comment here](https://github.com/ni/nimble/issues/1843#issuecomment-2064663561), I think what you have here is OK though, because the binding isn't _only_ the null coalescing operator (you have `? :` too).

Can see if others want to avoid it entirely (vs. turning off the rule) or not.

msmithNI (changes requested):

packages/nimble-components/src/toggle-button/template.ts
m-akinc: Should we try to upstream a change to FAST that includes `??` in the regex used to identify volatile bindings?

packages/nimble-components/src/anchor/tests/anchor.spec.ts
m-akinc: That test was meant to verify that we are using the `nullableNumberConverter` for `tabindex`. I'm testing that if a binding ends up evaluating to `null`, it removes the attribute. I think `setAttribute('tabindex', 'null')` should be equivalent to a template with `tabindex="${x => x.foo}"` where `x.foo` evaluates to `null`. Is that wrong?

I suppose we could _also_ test that clearing the attribute with `removeAttribute('tabindex')` also reflects to the inner element.

packages/nimble-components/src/button/tests/button.foundation.spec.ts
m-akinc: I intentionally went with a non-permalink because it shows the branch name and can be used to diff with the current contents. But I guess that could be misleading if we're implying that's the version we forked. Will change.

packages/nimble-components/src/anchor/tests/anchor.spec.ts
msmithNI: As best as I can tell (stepping through the current checkbox binding code without these changes), `setAttribute('tabindex', 'null')` wouldn't be what happens. Since in templates we're generally binding to `tabindex` as an attribute, we go through FAST's [updateAttributeTarget()](https://github.com/microsoft/fast/blob/9c6dbb66615e6d229fc0ebf8065a67f109139f26/packages/web-components/fast-element/src/templating/binding.ts#L72), and then their [DOM.setAttribute()](https://github.com/microsoft/fast/blob/9c6dbb66615e6d229fc0ebf8065a67f109139f26/packages/web-components/fast-element/src/dom.ts#L208), which calls `removeAttribute()` on the element for null/undefined values.

I actually don't think we need the 'nullable' part of the converter, but I didn't see a plain number converter without that to use (without making one). In any case, I think the tests that do `removeAttribute()` are sufficient for how we'll be using tabindex.

msmithNI (approved):

change/@ni-nimble-components-573fc816-eaa7-4e07-b0ef-8bfd5ccff19e.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Think it'd be worthwhile to add some developer docs reminding us why/how we should do this for future components? I could imagine it going in spec templates or in nimble-components/CONTRIBUTING.md. It's kinda sorta similar to [this section of the latter](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#focus-delegation) so that could be a good place to put it, but I don't feel too strongly.

packages/nimble-components/src/anchor/index.ts
```diff
@@ -35,6 +35,14 @@ export class Anchor extends AnchorBase {
     @attr
     public appearance: AnchorAppearance;
 
+    /**
+     * @public
+     * @remarks
+     * HTML Attribute: tabindex
+     */
+    @attr({ attribute: 'tabindex', converter: nullableNumberConverter })
+    public override tabIndex!: number;
```
jattasNI, line 44: Double checking, do we need to update Angular / Blazor integrations to expose tabIndex on these components or is it available already because it's a native attribute?

jattasNI (approved):

packages/nimble-components/src/anchor/index.ts
m-akinc: From my experimentation, it works without adding anything to our Angular/Blazor wrappers.

change/@ni-nimble-components-573fc816-eaa7-4e07-b0ef-8bfd5ccff19e.json
m-akinc: I added it to that section of `CONTRIBUTING.md`.

packages/nimble-components/src/button/tests/button.spec.ts
```diff
@@ -1,11 +1,56 @@
+import { html } from '@microsoft/fast-element';
 import { Button, buttonTag } from '..';
+import { fixture, type Fixture } from '../../utilities/tests/fixture';
+import { waitForUpdatesAsync } from '../../testing/async-helpers';
 
 describe('Button', () => {
+    async function setup(tabIndex?: string): Promise<Fixture<Button>> {
+        return fixture<Button>(
+            html`<${buttonTag} ${tabIndex ? `tabindex=${tabIndex}` : ''}></${buttonTag}>`
```
rajsite, line 9: returning undefined / null to a fast attribute binding removes it which is my guess as to the funky syntax here. can you use a normal fast template?


```suggestion
            html`<${buttonTag} tabindex="${tabindexAttribute}"></${buttonTag}>`
```

See: https://www.fast.design/docs/fast-element/declaring-templates#attributes

> Example: Nullish value

> Some cases may occur where an attribute needs to have a value when used however not present if unused. These are different than boolean attributes by where their presence alone triggers their effect. In this situation, a nullish value (null or undefined) may be provided so the attribute won't exist in that condition.

packages/nimble-components/src/checkbox/template.ts
```diff
@@ -0,0 +1,41 @@
...
+ViewTemplate<Checkbox>,
+CheckboxOptions
+> = (_context, definition) => html`
+    <template
+        role="checkbox"
+        aria-checked="${x => x.checked}"
+        aria-required="${x => x.required}"
+        aria-disabled="${x => x.disabled}"
+        aria-readonly="${x => x.readOnly}"
+        tabindex="${x => (x.disabled ? null : x.tabIndex ?? 0)}"
```
rajsite, line 19: Can you add parenthesis here so it's less ambiguous? I can't tell if the behavior is
`(x.disabled ? null : x.tabIndex) ?? 0`
or
`x.disabled ? null : (x.tabIndex ?? 0)`

rajsite: Coupled with the other comment remove the eslint disable and move the logic to a method which has no issues with the `??` syntax.

packages/nimble-components/src/button/tests/button.spec.ts
```diff
@@ -1,11 +1,56 @@
+import { html } from '@microsoft/fast-element';
 import { Button, buttonTag } from '..';
+import { fixture, type Fixture } from '../../utilities/tests/fixture';
+import { waitForUpdatesAsync } from '../../testing/async-helpers';
 
 describe('Button', () => {
+    async function setup(tabIndex?: string): Promise<Fixture<Button>> {
```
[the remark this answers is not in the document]
rajsite: Actually these test patterns are inconsistent in all these files. Is there a specific reason? Can we just align on a single way of configuring the fixture in these tests (seems like the setAttribute approach is the easiest, don't think parameterizing the fixture is adding significant value here)?

packages/nimble-components/src/toggle-button/template.ts
rajsite: Yea there are two issues in the comment Malcolm linked thtat should be reported and fixed upstream. One is adding support for `??` and the other is that aggregate bindings don't ever work with the regex to mark a binding volatile.

packages/nimble-components/src/toggle-button/template.ts
```diff
@@ -14,7 +15,7 @@ ButtonOptions
     <div
         role="button"
         part="control"
-        tabindex="${x => (x.disabled ? null : 0)}"
+        tabindex="${x => (x.disabled ? null : x.tabIndex ?? 0)}"
```
rajsite, line 18: A couple lines below we set the disabled property of this `div` but the `div` element does not have a disabled property

rajsite: Can you also fix the whitespace for the comments below? Guessing someone just let prettier win

packages/nimble-components/src/button/tests/button.spec.ts
rajsite: The above comment would have us not parameterizing the fixture for this use-case so may be moot

rajsite (commented):
Most of the PR feedback applies to multiple spots but I did not comment on all of them. Make sure to address each instance of those issues.

packages/nimble-components/src/anchor/tests/anchor.spec.ts
```diff
@@ -80,6 +80,25 @@ describe('Anchor', () => {
...
+
+            expect(element.control!.getAttribute('tabindex')).toBe('-1');
+        });
+    });
+
+    it('should clear tabindex attribute from the internal control when cleared from the host', async () => {
+        await connect();
+
+        element.setAttribute('tabindex', '-1');
+        element.removeAttribute('tabindex');
```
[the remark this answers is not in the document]
m-akinc: Intent is to test that going from `tabindex` being set to removing it also removes it from the inner anchor. That's why that `setAttribute('tabindex', '-1')` is there.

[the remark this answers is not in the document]
rajsite: I'd recommend putting in assertions between those transitions

packages/nimble-components/src/checkbox/template.ts
m-akinc: Prettier won't let me add the parens (or otherwise format the logic with newlines to make precedence clearer). I could:
1. leave it as `return this.disabled ? null : this.tabIndex ?? 0;`
2. add the parens and a `// prettier-ignore`
3. split it into `if (this.disabled) { return null; } return this.tabIndex ?? 0;`

I went with 2, but let me know if you prefer otherwise.

packages/nimble-components/src/toggle-button/template.ts
m-akinc: I will keep a note to make those changes when I have nothing more valuable to do.

packages/nimble-components/src/anchor/tests/anchor.spec.ts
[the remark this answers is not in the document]
m-akinc: I noticed that I could move the `setAttribute` calls to before `connect()`, better indicating that it is a test precondition and separating it from the `removeAttribute` call. I generally try to avoid incidental assertions that are already covered by other tests (i.e. the `"should reflect value to the internal control for attribute tabindex"` test).

packages/nimble-components/src/toggle-button/template.ts
rajsite: Can you make that note an issue for reporting the bugs to FAST so we can track it?

m-akinc: Created #2127

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -375,6 +375,28 @@ const nimbleButton = Button.compose({
...
+```
+
+Then in the template, bind the focusable elements' `tabindex` to the host component's property:
+
+```html
+html<MyComponent
+    >`
+    <nimble-button ... tabindex="${x => x.tabIndex}"></nimble-button>
+    // or for an element that isn't focusable by default:
+    <div ... tabindex="${x => (x.disabled ? null : x.tabIndex ?? 0)}"></div>
```
rajsite, line 395: should tweak this snippet for the copy-pasters to incorporate feedback issues: prefer undefined, unambiguous operators, consistent typing, nullish in bindings.
Maybe it's not well known but arrow functions can be multiline (though usually a sign of doing too much work in template but potentially useful for examples)

```
<div tabindex = "${x => {
    const tabindex = x.tabIndex ?? 0;
    return x.disabled ? undefined : `${tabindex}`;
}">
```

packages/nimble-components/src/checkbox/index.ts
```diff
@@ -16,7 +17,23 @@ declare global {
...
+     */
+    @attr({ attribute: 'tabindex', converter: nullableNumberConverter })
+    public override tabIndex!: number;
+
+    /**
+     * @internal
+     */
+    public get resolvedTabIndex(): string | undefined {
+        const tabIndex = this.tabIndex ?? 0;
+        return this.disabled ? undefined : tabIndex.toString();
```
rajsite, line 34: nit / pref, using in a template for the explicit conversion is a little more compact
```suggestion
        return this.disabled ? undefined : `${tabIndex}`;
```

packages/nimble-components/src/checkbox/index.ts
```diff
@@ -16,7 +17,23 @@ declare global {
...
+     * @remarks
+     * HTML Attribute: tabindex
+     */
+    @attr({ attribute: 'tabindex', converter: nullableNumberConverter })
+    public override tabIndex!: number;
+
+    /**
+     * @internal
+     */
+    public get resolvedTabIndex(): string | undefined {
```
rajsite, line 32: because this is resulting in the attribute format I think we should model this on the attribute name

```suggestion
    public get resolvedTabindex(): string | undefined {
```

packages/nimble-components/src/checkbox/template.ts
```diff
@@ -0,0 +1,40 @@
...
+ViewTemplate<Checkbox>,
+CheckboxOptions
+> = (_context, definition) => html`
+    <template
+        role="checkbox"
+        aria-checked="${x => x.checked}"
+        aria-required="${x => x.required}"
+        aria-disabled="${x => x.disabled}"
+        aria-readonly="${x => x.readOnly}"
+        tabindex="${x => x.resolvedTabIndex}"
```
rajsite, line 18:
```suggestion
        tabindex="${x => x.resolvedTabindex}"
```

packages/nimble-components/src/toggle-button/index.ts
```diff
@@ -46,8 +46,24 @@ export class ToggleButton extends FoundationSwitch implements ButtonPattern {
...
     /** @internal */
     public readonly control!: HTMLElement;
+
+    /**
+     * @internal
+     */
+    public get resolvedTabIndex(): string | undefined {
+        const tabIndex = this.tabIndex ?? 0;
+        return this.disabled ? undefined : tabIndex.toString();
+    }
```
rajsite, line 66:
```suggestion
    public get resolvedTabindex(): string | undefined {
        const tabIndex = this.tabIndex ?? 0;
        return this.disabled ? undefined : `${tabIndex}`;
    }
```

packages/nimble-components/src/toggle-button/template.ts
```diff
@@ -7,30 +8,35 @@ import {
...
+// prettier-ignore
 export const template: FoundationElementTemplate<
 ViewTemplate<ToggleButton>,
 ButtonOptions
 > = (context, definition) => html<ToggleButton>`
     <div
         role="button"
         part="control"
-        tabindex="${x => (x.disabled ? null : 0)}"
+        tabindex="${x => x.resolvedTabIndex}"
```
rajsite, line 20:
```suggestion
        tabindex="${x => x.resolvedTabindex}"
```

packages/nimble-components/src/toggle-button/template.ts
```diff
@@ -7,30 +8,35 @@ import {
 } from '@microsoft/fast-foundation';
 import type { ToggleButton } from '.';
 
+/* eslint-disable @typescript-eslint/indent */
```
rajsite, line 11: each eslint-disable should have a corresponding eslint-enable so future additions after the section don't accidentally have the rule disabled. Can you add an eslint-enable for the rule after the template definition?

change/@ni-nimble-components-573fc816-eaa7-4e07-b0ef-8bfd5ccff19e.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: This change ended up improving our aria score from 81 to 84 in lighthouse @m-akinc @jattasNI 🎉

Looks like in numeric increment / decrement buttons we were setting tabindex="-1" but that wasn't reflect to the inner button which was then technically focusable. After this change it also gets tabindex="-1" and makes lighthouse happy ✔

image

rajsite (approved):

packages/nimble-components/src/toggle-button/template.ts
rajsite, line 1: Should not need the disable anymore

packages/nimble-components/CONTRIBUTING.md
m-akinc: Also fixing mangling by prettier that I hadn't noticed.

PR #2111, opened 2024-05-15
Select list option groups

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #791

This PR only introduces groups to the `Select`. Providing this feature for the `Combobox` will be handled separately.

## 👩‍💻 Implementation

Introduces the `ListOptionGroup` component. This component required some APIs that manage the visibility of visual separators between groups and other visuals. This management occurs by the `Select` handling when the `hidden` or `visibilityHidden` state of a `ListOptionGroup` changes and then updating the necessary states on both the group that changed its hidden state and the adjacent groups if present. The `ListOptionGroup` manages its own `visibilityHidden\hidden` state based on the aggregate state of its contained `ListOption`s.

I've also changed the `ListOption` to assign itself to the `option` slot that now exists in both the `ListOptionGroup` template and all of the various dropdown components (i.e. `Select`, `Combobox`, and `RichTextMentionListbox`). This follows a similar pattern to that of the `Tabs` component and its related pieces.

Out of scope:

- Explicit handling of groups within groups. It may be prudent to provide a more robust implementation that sets some validity state on the `Select` itself. For now, if a client provides a `nimble-list-option-group` within another `nimble-list-option-group`, nothing will get displayed in the dropdown, but no exception is thrown. The documentation currently states that this is unsupported.

## 🧪 Testing

Unit tests. Matrix tests.

## ✅ Checklist


packages/nimble-components/src/select/styles.ts
```diff
@@ -29,6 +30,25 @@ export const styles = css`
          error icon, and dropdown arrow because they are not "interactive" i.e. part of the tab order */ ''
     }
 
+    ::slotted(nimble-list-option-group.show-top-separator)::before,
```
mollykreis, line 33: Should all of the separator styling be in the list-option-group styles file? It seems like that would be the most reusable place for when grouping is added to other components, such as the combobox.

mollykreis, line 33: Why make `show-top-separator` and `last-visible-option-group` classes while `visually-hidden` is a property/attribute? It seems like those should be consistent. I also don't think we tend to add/remove css classes to elements created by a client app.

packages/nimble-components/src/select/index.ts
```diff
@@ -975,33 +990,96 @@ export class Select
      * @public
      */
     private filterOptions(): void {
-        const filter = this.filter.toLowerCase();
+        if (!this.$fastController.isConnected) {
```
mollykreis, line 993: There is a lot going on in this function. I think it would be good to consider ways to make it more readable. Some ideas I had to consider are:

- Simplify the code path when `filter` is empty.
- Don't create a local function (particularly one this long/complex) within another function
- Split the logic that updates the state on an option/group from the logic that determines if it matches the filter.
- Create some helper functions that encapsulate more of this logic
- [Unsure] Put the logic that hides a group if it has no visible children in list-option-group itself. I'm not sure if that makes sense or not

I'd be happy to brainstorm more concrete ideas around this, but I thought I'd at least put a few possible directions here for you to consider.

packages/nimble-components/src/select/template.ts
```diff
@@ -131,7 +136,7 @@ SelectOptions
                         class="scrollable-region">
                         <slot
                             ${slotted({
-                                filter: (n: Node) => n instanceof HTMLElement && isListboxOption(n),
+                                filter: (n: Node) => n instanceof HTMLElement && (isListboxOption(n) || slottedOptionGroupFilter(n)),
```
mollykreis, line 139: Is there a reason not to call this new function `isOptionGroup` to parallel the `isListboxOption` function already being called here?

mollykreis (changes requested):

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /** @internal */
+    @volatile
+    public get labelContent(): string {
+        if (this.label || !this.$fastController.isConnected) {
+            return this.label ?? '';
+        }
+
+        const nodes = this.labelSlot.assignedNodes();
+        return nodes.length > 0
```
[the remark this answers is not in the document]
atmgrifter00: You may be right. The idea behind the implementation as it is, was to provide a means of ellipsizing the text regardless of the API used. I suppose I can keep this implementation in place, and change the template to host the slotted content as provided (if the `label` isn't set), but then writing a unit test to actually check the rendered display content matches expectations gets a bit more complicated, but that shouldn't influence our design.

change/@ni-nimble-components-278f4b89-b1f8-402d-a3e7-ba0338b1b348.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
atmgrifter00: The problem was really that the selected styling wasn't being applied to the active option. The arrow keys and typeahead were still changing the active option, there just wasn't a visual reflection of it. This required (afaik) adding styles to the `ListOptionGroup` to handle the styling of slotted `ListOption`s. This is a concern because they currently rely on the new `activeOption` attribute, which is not implemented by the `Combobox`. I could not figure out how to style the slotted `ListOption`s (within a `ListOptionGroup`) from the `Select` style, which would avoid that concern.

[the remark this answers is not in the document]
mollykreis: This might run into the same issue of combobox & select needing different behavior, but could the list option know how to style itself when `activeOption` is set? Then the exact same behavior would be achieved regardless of whether or not the parent is a group or a select?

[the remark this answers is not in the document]
atmgrifter00: I've at least tried this before, but with the `ListOption` already styling itself with respect to `aria-selected='true'` has proven tricky to mix that with `activeOption` as well.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +984,89 @@ export class Select
...
 
-        this.options.forEach(o => {
-            if (isNimbleListOption(o)) {
-                if (!this.filteredOptions.includes(o)) {
-                    o.visuallyHidden = true;
-                } else {
-                    o.visuallyHidden = false;
+        const filteredOptions: ListOption[] = [];
+
+        let lastVisibleOptionGroupIndex = -1;
```
mollykreis, line 1014: I don't think you need to track `lastVisibleOptionGroupIndex` because you can set `lastVisibleElement.showBottomSeparator = true` when you find the next visible thing.

That would not only let you get rid of this variable and the check after the loop, but I think that would let you change the for loop to be `for (const element of this.slottedOptions)` rather than index based (which requires non-null assertion when indexing).

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +984,89 @@ export class Select
...
+                element.visuallyHidden = allOptionsHidden || element.hidden;
+
+                if (!allOptionsHidden) {
+                    element.showBottomSeparator = true;
+                    lastVisibleOptionGroupIndex = i;
+                    element.showTopSeparator = lastVisibleElement instanceof ListOption;
+                }
+                lastVisibleElement = allOptionsHidden
+                    ? lastVisibleElement
+                    : element;
```
mollykreis, line 1045: Instead of writing this logic in both if-statements, can you write it a single time for both?
```ts
for (...) {
    if (isGroup) {
       ...
    } else if (isOption) {
       ...
    }

    if (!element.visuallyHidden) {
        lastVisibleElement = element;
    }
}
```

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +984,89 @@ export class Select
...
+        const filteredOptions: ListOption[] = [];
+
+        let lastVisibleOptionGroupIndex = -1;
+        let lastVisibleElement: ListOption | ListOptionGroup | undefined;
+        for (let i = 0; i < this.slottedOptions.length; i++) {
+            const element = this.slottedOptions[i]!;
+
+            if (element instanceof ListOptionGroup) {
+                element.showTopSeparator = false;
+                element.showBottomSeparator = false;
```
mollykreis, line 1021: Do you need to reset these to `false`? They get set if the option is visible, and I think it would be good if the styling was written such that the separators aren't shown on a hidden group, regardless of the values of these properties.

packages/nimble-components/src/list-option-group/index.ts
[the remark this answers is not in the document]
atmgrifter00: As discussed out-of-thread, we managed to come up with what seems like a satisfactory solution to support both label attributes, and general slotted content.

packages/nimble-components/src/list-option-group/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+    groupHeaderTextTransform,
+    smallPadding
+} from '../theme-provider/design-tokens';
+
+export const styles = css`
+    ${display('flex')}
+
+    :host {
+        cursor: default;
+        justify-content: left;
```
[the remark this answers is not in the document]
atmgrifter00: Nope. Guessing copy/pasta.

packages/nimble-components/src/select/index.ts
atmgrifter00: Refactored fairly heavily. Hopefully these concerns are addressed satisfactorily.

packages/nimble-components/src/select/styles.ts
atmgrifter00: Transitioned this API to attributes, which there are only two of: `showTopSeparator` and `showBottomSeparator`.

packages/nimble-components/src/select/styles.ts
```diff
@@ -29,6 +30,25 @@ export const styles = css`
...
+        opacity: 0.1;
+    }
+
+    ::slotted(nimble-list-option-group.last-visible-option-group)::after {
+        width: 0px;
+        display: none;
+    }
+
+    ::slotted(nimble-list-option-group.show-bottom-separator)::after {
+        width: auto;
```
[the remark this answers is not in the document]
atmgrifter00: Changed.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1007,90 @@ export class Select
...
+            }
+
+            if (lastVisibleElement instanceof ListOptionGroup) {
+                lastVisibleElement.showBottomSeparator = true;
+            }
+
+            if (element instanceof ListOptionGroup && !elementHidden) {
+                element.showTopSeparator = lastVisibleElement instanceof ListOption;
+                element.showBottomSeparator = false;
+            }
```
mollykreis, line 1076: I think this whole block of code needs to be wrapped around a `!elementHidden` statement, rather than having that check being part of line 1073. As is, you will set `showBottomSeparator = true` on the last visible group, even if nothing is visible below it.

To see this in practice, you can look at your storybook page with filtering & grouping enabled. In the filter, type "1". Only the first group is shown, but it incorrectly has a bottom separator.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -91,6 +119,10 @@ To display placeholder text within the \`Select\` you must provide an option tha
 Any \`Select\` without a default selected option should provide placeholder text. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Use sentence casing and don't include punctuation at the end of the prompt.
 `;
 
+const groupedDescription = `
+To group options in a \`Select\`, you can use the \`nimble-list-option-group\` element. This element should be placed within the \`Select\` and contain the \`nimble-list-option\` elements that you want to group. Note that a \`nimble-list-option-group\` placed within another \`nimble-list-option-group\` is not supported. The \`label\` attribute of the \`nimble-list-option-group\` element will be used as the group label. Alternatively, text can be provided next to the \`nimble-list-option-group\` element to serve as the group label.
```
mollykreis, line 123: Can you use the exported tag names here rather than hard-code element tags in the description?

packages/nimble-components/src/select/index.ts
```diff
@@ -376,6 +396,11 @@ export class Select
             case 'hidden': {
                 if (isNimbleListOption(sourceElement)) {
                     sourceElement.visuallyHidden = sourceElement.hidden;
+                } else if (isNimbleListOptionGroup(sourceElement)) {
+                    sourceElement.listOptions.forEach(e => {
```
mollykreis, line 400: I think this is the only code that updates `visuallyHidden` on child list options when a group is hidden. If so, you should make sure everything works correctly when a group that is already hidden is slotted into the component. Since this is a change handler, I wouldn't expect this code path to be hit in that case. I suspect that may be problematic.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1009,92 @@ export class Select
...
+                }
+
+                if (element instanceof ListOptionGroup) {
+                    element.showTopSeparator = lastVisibleElement instanceof ListOption;
+                    element.showBottomSeparator = false;
                 }
             }
-        });
+
+            lastVisibleElement = elementHidden ? lastVisibleElement : element;
```
mollykreis, line 1082: This line can move into the `if (!elementHidden)` block and be simplified to `lastVisibleElement = element`.

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -28,17 +29,53 @@ type FilterModeState = (typeof filterModeStates)[number];
...
+                ${when(() => slottedLabel, html`
+                    <span>Group 1</span>
+                `)}
+                <${listOptionTag} value="2" disabled>Option 1</${listOptionTag}>
+                <${listOptionTag} value="3">Option 2</${listOptionTag}>
+            </${listOptionGroupTag}>         
+            ${when(() => optionsOutsideGroup ?? false, html`
+                <${listOptionTag}>Option Not in Group</${listOptionTag}>
+            `)}
+            <${listOptionGroupTag}  ${!slottedLabel ? 'label="Group 2 with a ridiculously long label that does\'t fit"' : ''}>
```
mollykreis, line 63:
```suggestion
            <${listOptionGroupTag}  ${!slottedLabel ? 'label="Group 2 with a ridiculously long label that doesn\'t fit"' : ''}>
```

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -28,17 +29,53 @@ type FilterModeState = (typeof filterModeStates)[number];
...
+                ${when(() => slottedLabel, html`
+                    <span>Group 2 with a ridiculously long label that does't fit</span>
```
mollykreis, line 65:
```suggestion
                    <span>Group 2 with a ridiculously long label that doesn't fit</span>
```

mollykreis (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: Line 1093 (the `getGroupedOptions` implementation) handles this case. I don't think I have test coverage for this though so I'll add it.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -91,6 +119,10 @@ To display placeholder text within the \`Select\` you must provide an option tha
 Any \`Select\` without a default selected option should provide placeholder text. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Use sentence casing and don't include punctuation at the end of the prompt.
 `;
 
+const groupedDescription = `
```
jattasNI, line 122: When elements are used with specific child elements I am moving our docs to a pattern of creating separate stories for those child elements. You can see the pattern in the menu and tab stories now. It allows us to explicitly document the API of the child elements rather than trying to cram a bunch of info in a catch-all control that doesn't match to an actual attribute / slot.

I've [locally started refactoring the select story](https://github.com/ni/nimble/compare/api-docs-inputs?expand=1) to get ready to use some of these patterns but haven't quite written a story for the list option yet (or the group, obviously).

This is mostly a heads up that I will want to change this and that we should coordinate. You are welcome to start following that pattern now, wait for me to come through and adopt it, or plan to adopt it up yourself later once I've laid the groundwork.

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -174,3 +211,48 @@ export const selectBelowOpenNoFilterDarkThemeWhiteBackgroundWithPlaceholder: Sto
...
+);
+
+export const selectGroupedWithSlottedLabelColorThemeWhiteBackground: StoryFn = createFixedThemeStory(
+    component(positionStates[0], FilterMode.standard, true, true, false, true),
+    colorThemeDarkGreenBackground
+);
+
+export const selectGroupedWithSlottedLabelDarkThemeWhiteBackground: StoryFn = createFixedThemeStory(
+    component(positionStates[0], FilterMode.standard, true, true, false, true),
+    darkThemeBlackBackground
```
jattasNI, line 257: The names of these stories don't match the names of the variables being passed for the theme.

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -28,17 +29,53 @@ type FilterModeState = (typeof filterModeStates)[number];
...
+
+const optionsOutsideGroupStates = [false, true] as const;
+type OptionsOutsideGroupState = (typeof optionsOutsideGroupStates)[number];
+
 // prettier-ignore
 const component = (
     [position, positionStyle]: PositionState,
     filterMode: FilterModeState,
-    placeholder?: PlaceholderState
+    placeholder?: PlaceholderState,
```
jattasNI, line 42: What do you think of moving these parameters into an options object? That would make it easier to not provide some of them and it would make the calls more readable because you'd see the names of the parameters at the call.

`component({postionState: positionStates[0], filterMode: FilterMode.standard, placeholder: true, grouped: true})`

jattasNI (commented):
Initial feedback from looking at Storybook, still need to review most of the code.

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,121 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'nimble-list-option-group': ListOptionGroup;
+    }
+}
+
+/**
+ * A nimble-styled HTML listbox option
+ */
+export class ListOptionGroup extends FoundationElement {
```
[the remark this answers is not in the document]
atmgrifter00: Yeah, I say we defer until there is a need. As the group itself isn't interactable, I don't see it as a huge UX issue for its display to not change to a "disabled" look.

packages/storybook/src/nimble/select/select.stories.ts
atmgrifter00: I'll make a pass at adopting it.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -91,6 +119,10 @@ To display placeholder text within the \`Select\` you must provide an option tha
...
+To group options in a \`Select\`, you can use the \`${listOptionGroupTag}\` element. This element should be placed within the \`${selectTag}\` and contain the \`${listOptionTag}\` elements that you want to group. Note that a \`${listOptionGroupTag}\` placed within another \`${listOptionGroupTag}\` is not supported. The \`label\` attribute of the \`${listOptionGroupTag}\` element will be used as the group label. Alternatively, text can be provided next to the \`${listOptionGroupTag}\` element to serve as the group label.
```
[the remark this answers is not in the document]
atmgrifter00: Updated.

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * The hidden state of the element.
+     *
+     * @public
+     * @defaultValue - false
+     * @remarks
+     * HTML Attribute: hidden
+     */
+    @attr({ mode: 'boolean' })
+    public override hidden = false;
```
jattasNI, line 37: Is it supported to mark a group `hidden`? If so, can you make sure:
1. the attribute is documented (could happen in the follow up PR where you add a story for the group element)
2. the behavior is well defined. I assume it hides the items under the group too?
3. we have a visual test covering the behavior (could just add a hidden group to the matrix story template)
4. we have unit test coverage as needed. Looks like you already have coverage for filtering of hidden groups but might also want something that asserts setting `hidden` causes `visuallyHidden` to update, for example.

If it's not intended to be a supported API then this should be marked internal.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1399,6 +1426,173 @@ describe('Select', () => {
...
+        it('can select an option within a group', async () => {
+            await clickAndWaitForOpen(element);
+            pageObject.clickOption(1);
+            expect(element.value).toBe('two');
+            expect(element.selectedIndex).toBe(1);
+        });
+
+        it('clicking a group does not close the dropdown', async () => {
+            await clickAndWaitForOpen(element);
+            const group = element.querySelector('[role="group"]');
```
jattasNI, line 1456: Could you take a pass at moving group-related helper functions in the select page object? They could be useful to clients too. Anything does a `querySelector...('[role="group"]')` or interacts with the result of that would be a good candidate to move.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1009,92 @@ export class Select
         this.updateListboxMaxHeightCssVariable();
     }
 
+    private isOptionHidden(option: ListOption): boolean {
```
jattasNI, line 1012: nit: given that we have so many concepts of "hidden" now (option attribute, option filtered out, group attribute) it would be helpful to give this a more precise name. `shouldBeVisuallyHidden()`? `isOptionHiddenOrFilteredOut()`?

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public clickHandler(e: MouseEvent): void {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+    }
+
+    protected slottedElementsChanged(): void {
+        this.slottedElements.forEach(e => {
+            if (e instanceof ListOption) {
+                e.slot = 'options-slot';
```
rajsite, line 105: can we have the options themselves own configuring their slot?
Also we don't usually have the slot name include the word "slot", maybe just "option" (analogous to `<nimble-tab slot="tab">`)?

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    public override hidden = false;
+
+    /**
+     * @internal
+     * This attribute is required to allow use-cases that offer dynamic filtering
+     * (like the Select) to visually hide groups that are filtered out, but still
+     * allow users to use the native 'hidden' attribute without it being affected
+     * by the filtering process.
+     */
+    @attr({ attribute: 'visually-hidden', mode: 'boolean' })
```
rajsite, line 46: Talked offline but I'm generally concerned about how much internal state is on list-option-group and list-option.
The concern is around option and group element recycling.
Some thoughts:
- Should we clear internal state on connect / disconnect (option or group re-used between controls or re-ordered)
- Should we clear internal state for any public attribute change (value, selected, hidden)? (options are updated in-place to represent new values, i.e. in manual filtering we only populate the first 5 options and as the filter changes the state of the options change but they are not attached / detached / re-ordered)
- Is WeakMap a useful tool for managing state of the options in the select itself instead of on the option / group instances? (may not be since we want to leverage internal state as attributes for styling)
- Is [sibling combinators](https://developer.mozilla.org/en-US/docs/Web/CSS/Next-sibling_combinator) a useful tool for a different approach to state management? i.e. groups just reflect an attribute that says they are empty and use sibling combinators to declaratively style bottom / top separator instead of relying on imperative "show" configuration

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    public visuallyHidden = false;
+
+    /**
+     * @internal
+     */
+    @attr({ attribute: 'show-top-separator', mode: 'boolean' })
```
rajsite, line 52: double check alignment with part naming: https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#attribute-common-name-patterns

`top-separator-visible`, `bottom-separator-visible`, etc

packages/storybook/src/nimble/select/select.stories.ts
atmgrifter00: This will actually happen in a follow up PR.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1399,6 +1426,173 @@ describe('Select', () => {
...
+        });
+
+        it('can select an option within a group', async () => {
+            await clickAndWaitForOpen(element);
+            pageObject.clickOption(1);
+            expect(element.value).toBe('two');
+            expect(element.selectedIndex).toBe(1);
+        });
+
+        it('clicking a group does not close the dropdown', async () => {
```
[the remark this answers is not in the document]
atmgrifter00: I wrote a bad test coupled with an incomplete implementation for a false positive. 🤦‍♂️

packages/nimble-components/src/list-option-group/tests/list-option-group.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    it('if label attribute is provided, labelContent is set to label attribute', async () => {
+        const { element, connect, disconnect } = await setup();
+        await connect();
+
+        expect(element.labelContent).toBe('Group 1');
+
+        await disconnect();
+    });
+
+    it('if label attribute is undefined, and slotted label is provided, labelContent is set to slotted label', async () => {
```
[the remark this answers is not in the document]
atmgrifter00: Added test.

packages/nimble-components/src/list-option-group/tests/list-option-group.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+        slottedLabel2.textContent = 'Slotted Label 2';
+        element.appendChild(slottedLabel2);
+        await waitForUpdatesAsync();
+
+        expect(element.labelContent).toBe('Slotted Label Slotted Label 2');
+
+        await disconnect();
+    });
+
+    it('if label attribute and slotted label are provided, labelContent is set to label attribute', async () => {
```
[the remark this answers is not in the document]
atmgrifter00: Discussed off-line, but I've added some tests to check the `title` behavior for the first and third scenarios listed, and I've added a test to cover the second scenario.

packages/nimble-components/src/list-option-group/index.ts
atmgrifter00: I've added some more tests. Documentation will come in a follow-on PR. Interesting aspect related to this is that I've created a new internal page object class that provides some APIs related to groups.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: I've added a couple of public APIs to the exported page object, but I've also create an internal page object that encapsulates all of the `querySelector` calls related to groups.

packages/nimble-components/src/list-option-group/index.ts
atmgrifter00: So I've done a re-factor that may allay some of the concerns presented here (but perhaps not).

The main thrust of the re-factor is removing the management of the separator visual states from the `filterOptions` implementation in order to support scenarios like manual filtering with groups. This is now controlled by responding to `hidden` and `visuallyHidden` state changes on the groups and options themselves. The `visuallyHidden` state of the group is now managed by the group itself, and not within the `filterOptions` implementation, again to support the manual filtering with groups case.

Sibling combinators do not seem possible to manage the necessary visual states as there is no CSS syntax to handle skipping over hidden groups.

So my summary of responses to the bullet points are (and I'm certain there could be more discussion to be had):
- I don't think this is necessary at the moment without identifying a use-case that won't function properly.
- Same as above
- I don't believe this is necessary due to how the `visuallyHidden` state gets reset when needed already
- Sibling combinators won't work because we do rely on styling to use the internal `visuallyHidden` state.

Unless we think the current implementation within this PR is problematic for known use-cases, I propose we move forward with it, and if there is an identified scenario outside of known use-cases that is problematic, I think this should be tracked by a Tech Debt issue.

@jattasNI @rajsite @mollykreis

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        const nodes = this.labelSlot.assignedNodes();
+        return nodes
+            .filter(node => node.textContent?.trim() !== '')
+            .map(node => node.textContent?.trim())
+            .join(' ');
+    }
+
+    private readonly hiddenOptions: Map<ListOption, boolean> = new Map();
```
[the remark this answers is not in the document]
atmgrifter00: 🤦‍♂️...yes, it should be a `Set`. I tracked the hidden options instead of the visible options mainly because I preferred this line of code for readability:
```
this.visuallyHidden = this.hiddenOptions.size === this.listOptions.length;
```
I'm fine changing it if you have a stronger preference.

packages/nimble-components/src/list-option-group/tests/list-option-group.spec.ts
```diff
@@ -0,0 +1,433 @@
...
+
+    it('group becomes hidden when all options are hidden', async () => {
+        const { element, connect, disconnect } = await setup();
+        await connect();
+        const option = document.createElement('nimble-list-option');
+        element.appendChild(option);
+        await waitForUpdatesAsync();
+        expect(element.visuallyHidden).toBeFalse();
+
+        element.removeChild(option);
```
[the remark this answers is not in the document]
atmgrifter00: Not sure what was going on in my head here. Updated test description.

packages/nimble-components/src/select/index.ts
```diff
@@ -376,10 +403,26 @@ export class Select
             case 'hidden': {
                 if (isNimbleListOption(sourceElement)) {
                     sourceElement.visuallyHidden = sourceElement.hidden;
+                    this.updateAdjacentSeparatorState(sourceElement);
+                } else if (isNimbleListOptionGroup(sourceElement)) {
+                    sourceElement.listOptions.forEach(e => {
+                        e.visuallyHidden = sourceElement.hidden;
+                    });
+                    this.filterOptions();
```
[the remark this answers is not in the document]
atmgrifter00: Good catch. I've changed it to now be called in both cases, and added a test that failed without the change.

change/@ni-nimble-components-278f4b89-b1f8-402d-a3e7-ba0338b1b348.json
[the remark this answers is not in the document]
jattasNI: I am seeing this bug again in the latest build: the arrow keys don't work when there are option groups.
https://60e89457a987cf003efc0a5b-urkleipuso.chromatic.com/?path=/docs/components-select--docs

jattasNI, line 1: There are some possibly unexpected diffs in the latest Chromatic build:
1. select components in a lot of stories got narrower
2. the listbox component is no longer rendering any items

packages/nimble-components/src/combobox/template.ts
```diff
@@ -76,7 +76,7 @@ ComboboxOptions
                 ?disabled="${x => x.disabled}"
                 ${ref('listbox')}
             >
-                <slot
+                <slot name="options"
```
jattasNI, line 79: Can you update the PR description to explain this change (and the corresponding one in mention-listbox)? And just to double check, this isn't client-facing, right? Clients can still slot combobox options in the default slot?

packages/nimble-components/src/select/template.ts
```diff
@@ -22,9 +22,14 @@ import {
     filterSearchLabel
 } from '../label-provider/core/label-tokens';
 import { FilterMode } from './types';
+import { ListOptionGroup } from '../list-option-group';
 import { buttonTag } from '../button';
 import { iconTimesTag } from '../icons/times';
 
+const isListOptionGroup = (n: HTMLElement): boolean => {
```
jattasNI, line 29: This function is duplicated in a few places in this PR. Can we single-source it?

Also, the[ FAST equivalent `isListboxOption`](https://github.com/microsoft/fast/blob/00bccd7812ca072fc2efdf39af7762bbfdc90846/packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts#L19) doesn't only do an `instanceof` check; it also checks the `role`. This seems more flexible since it would allow slotting new types of elements as long as they meet the API contract. Should we do the same? I think we do that for other element types. Or is our API contract too specific to our group element, so we'll never be able to support something like a native `<optgroup>`?

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -87,6 +115,10 @@ To display placeholder text within the select you must provide an option that ha
 Any select without a default selected option should provide placeholder text. Placeholder text should always follow the pattern "Select [thing(s)]", for example "Select country". Use sentence casing and don't include punctuation at the end of the prompt.
 `;
 
+const groupedDescription = `
+To group options in a \`Select\`, you can use the \`${listOptionGroupTag}\` element. This element should be placed within the \`${selectTag}\` and contain the \`${listOptionTag}\` elements that you want to group. Note that a \`${listOptionGroupTag}\` placed within another \`${listOptionGroupTag}\` is not supported. The \`label\` attribute of the \`${listOptionGroupTag}\` element will be used as the group label. Alternatively, text can be provided as content of the \`${listOptionGroupTag}\` element to serve as the group label. Text provided as content can be either plain text or as content of a \`<span\`> element.
```
jattasNI, line 119: I moved away from using the class name to refer to elements in one of my doc updates.
```suggestion
To group options in a select, you can use the \`${listOptionGroupTag}\` element. This element should be placed within the \`${selectTag}\` and contain the \`${listOptionTag}\` elements that you want to group. Note that a \`${listOptionGroupTag}\` placed within another \`${listOptionGroupTag}\` is not supported. The \`label\` attribute of the \`${listOptionGroupTag}\` element will be used as the group label. Alternatively, text can be provided as content of the \`${listOptionGroupTag}\` element to serve as the group label. Text provided as content can be either plain text or as content of a \`<span\`> element.
```

packages/nimble-components/src/select/template.ts
atmgrifter00: We are being more strict because we use this check to access APIs on the `ListOptionGroup` class (namely the `visuallyHidden` attribute). At the moment, I'm not sure of a better approach.

packages/nimble-components/src/list-option-group/index.ts
rajsite: Can we double check and align the plurality? Looks like we are doing `slot name="options"` here vs the exiting pattern in the tab control which is `slot="tab"`

change/@ni-nimble-components-278f4b89-b1f8-402d-a3e7-ba0338b1b348.json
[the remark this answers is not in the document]
atmgrifter00: Yep. I've corrected the issue and added a matrix test similar to the one I had made for options with no groups.

atmgrifter00: 1. Yep, this is intentional
2. We've talked through this off-line and we will be deleting the listbox component as nothing is using it.

packages/nimble-components/src/select/template.ts
atmgrifter00: Talked off-line and we're good with using explicit `instanceof` checks for now. I've single-sourced the implementation.

packages/nimble-components/src/combobox/template.ts
atmgrifter00: Updated the description. And for posterity, yes, this isn't client-facing. A client can still slot options into the default slot.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1017,131 @@ export class Select
         this.updateListboxMaxHeightCssVariable();
     }
 
+    private updateAdjacentSeparatorState(
```
jattasNI, line 1020: I spent a long time staring at this function trying to succinctly describe the rules for when separators should be visible. It seems like it should be simpler to understand.

I think it boils down to:
1. if I'm a group, show the bottom separator if there is a nextElement
2. if I'm a group, show the top separator if the previous element is an option

Among the things that would be good to clean up:
1. why do we set `element.bottomSeparatorVisible` regardless of `isOptionOrGroupVisible(element)` but only set `element.topSeparatorVisible` if `element` is visible?
2. instead of having one method encode the logic for both this element and next/previous and for both groups and options, have a lower-level method that sets the state for a given group and call that method on the 3 relevant groups (this, previous, next)
3. find a way to make the check-if-I'm-a-group-then-set-state pattern less verbose (1 line instead of 3). Possibly another helper function or maybe this will go away after you do 2.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1017,131 @@ export class Select
...
-            );
+                groupOptions.forEach(option => {
+                    option.visuallyHidden = groupMatchesFilter
+                        ? false
+                        : this.isOptionHiddenOrFilteredOut(option);
+                    if (!option.visuallyHidden) {
+                        filteredOptions.push(option);
+                    }
+                });
+            } else if (isNimbleListOption(element)) {
```
jattasNI, line 1124: Is there a strategy for when to use `isNimbleListOption(element)` vs `instanceof ListOption`? Same for `instanceof ListOptionGroup`. I think the code would be easier to read if we stayed at the same level of abstraction within a given file / method. I would vote for helper functions everywhere, and only doing instanceof checks within the helper function.

atmgrifter00: Agreed and updated as suggested.

packages/nimble-components/src/select/index.ts
```diff
@@ -969,39 +1017,131 @@ export class Select
...
-                return (
-                    !option.hidden
-                    && diacriticInsensitiveStringNormalizer(option.text).includes(
-                        normalizedFilter
-                    )
+        const filteredOptions: ListOption[] = [];
+        for (const element of this.slottedOptions) {
+            if (element instanceof ListOptionGroup) {
+                if (element.hidden) {
+                    break; // no need to process hidden groups
```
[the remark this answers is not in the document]
atmgrifter00: Yes...this should be `continue`.

packages/nimble-components/src/select/index.ts
atmgrifter00: I've now refactored this, and hopefully it is more clear.

packages/storybook/src/nimble/select/select-matrix.stories.ts
```diff
@@ -134,6 +135,17 @@ export const navigateToDifferentOption: StoryFn = createStory(
 
 navigateToDifferentOption.play = playFunction;
 
+export const navigateToDifferentOptionWithGroups: StoryFn = createStory(
+    html`<${selectTag} open style="width: 250px;">
```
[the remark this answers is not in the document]
atmgrifter00: Added an explicit height to a surrounding `div` element.

jattasNI (approved):

packages/nimble-components/src/list-option-group/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        if (!this.$fastController.isConnected) {
+            return '';
+        }
+
+        const nodes = this.labelSlot.assignedNodes();
+        return nodes
+            .filter(node => node.textContent?.trim() !== '')
+            .map(node => node.textContent?.trim())
+            .join(' ');
```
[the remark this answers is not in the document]
rajsite: Chatted offline with @atmgrifter00 and the group text serialization is proportional to the existing text serialization cost of options. It's not really unique to this PR.

Future potential options:
- More efficient filtering algorithm (on successive additive character inputs only filter against the current filter set)
- Usage guidance around number of instantiated DOM options / group elements (10s to low-100s) and dynamic / custom filtering for larger sets (high-100s to 1000s)

rajsite (approved):

rajsite:
@atmgrifter00  I think the richtext changes are trivial and non-breaking so don't need to wait for @vivinkrishna-ni

packages/nimble-components/src/list-option-group/tests/list-option-group.spec.ts
```diff
@@ -0,0 +1,463 @@
...
+        await waitForUpdatesAsync();
+        expect(element.visuallyHidden).toBeTrue();
+        await disconnect();
+    });
+
+    it('option that is removed and then updates its hidden state, does not update group', async () => {
+        const { element, connect, disconnect } = await setup();
+        await connect();
+        const option1 = document.createElement('nimble-list-option');
```
rajsite, line 155: we aren't super consistent but can you do a pass and try and use `listOptionTag` for these?

packages/nimble-components/src/list-option-group/index.ts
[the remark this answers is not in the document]
atmgrifter00: Will add usage guidance once we submit implementation for custom filtering.

PR #2113, opened 2024-05-15
Update anchor focus styling

### LINKED ISSUE #2011: Verify anchor design and update styling if necessary
# 🧹 Tech Debt

There are a couple differences between our anchor styling and the [Figma design](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-76706&mode=design&t=tRlgHtHiOWeXNCIe-0):

1. Visited links have a distinct color (purple)
2. Keyboard focus is indicated by an outline rather than double underline
    - ~~note: we should push back on this visual design since the text could flow across multiple lines, which would look weird if we try to draw a box around it~~
    - An outline will work fine – 
Screenshot 2024-05-09 at 2 37 18 PM

These were not the designs when we implemented the anchor. We should confirm whether they are what we want (with Brandon) and implement if so.

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #2011

## 👩‍💻 Implementation

Squared the styling of the anchor with the [design in Figma](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=8729-84171&t=ZWhzibWRmDT2XoD7-0):
- Keyboard focus: we have decided to keep the double underline, but change the color of the link so it is more noticeable.
- `appearance="prominent"` no longer does anything under the Color theme (updated Storybook docs)
- Fixed transparency (0.6 -> 0.75) for active link color under the Color theme

## 🧪 Testing

Manually tested in Storybook. We have existing visual tests for this.

## ✅ Checklist


m-akinc:
@mollykreis Would you please buddy?

change/@ni-nimble-components-e3d5df53-15c3-4fe1-8552-ad4db8dbddbd.json
m-akinc, line 1: Here are some experiments with different approaches for styling the focus box:
image

- Using `outline` provides a better result when a link spans multiple lines:
   - single outline around entire link area (vs overlapping boxes for each line)
   - `box-shadow` (without inset) and `border` do not close the left/right sides of each line
- `border` shifts content when toggled, which is a non-starter
- Using a negative `outline-offset` with `outline` (or `box-shadow` with `inset`) ensures that the box is visible regardless of container's overflow handling (though it does overlap edges of text)

For the considerations above, I'm proposing using `outline` with negative `outline-offset`.

mollykreis: The code itself seems fine, but I think it would be good to get PO and Visual Design sign off on the concession being made here that the outline will overlap the text. I personally think the overlapping border/text looks bad (and makes it hard to read the link), but whether or not it's acceptable isn't my decision to make.

I'm wondering if we should consider a different focus style (such as going back to the underline but making it a more prominent color, such as green).

m-akinc: Reaching out to Brandon and Fred by email for input.

NIbokeefe: 1. Can the outline be applied with the setting stroke > outside to provide more space between outline and text?
2. Do we have any control over padding?
3. Are we expecting to have a lot of large bodies of wrapping text with links within them?  Do we have any in product examples today?

It's good to see the extreme and what causes this design to break. I don't know how common this setup is for in product scenarios.

m-akinc: 1. Yes, with `outline`, we can use `outline-offset` to expand or contract the box in all directions at once (no independent left/right/top/bottom). The `box-shadow` gives us similar control, but I don't think we can do the same using `border` without adding padding.
2. We do not want to add padding, because links must blend smoothly into blocks of text. We don't want a link to have more space around it than regular text would.
3. Long, wrapping links like in the picture above are not the biggest concern (though we should be able to handle that case reasonably). It's more the case I mentioned in my email: if a table cell contains a link that doesn't fully fit in the cell (this is a real, SLE use case), we want to be able to ellipsize it and still be able to see a box around it when focused. There are technical hurdles to that, and forcing the box to be within the link's bounding box is one solution.

m-akinc: We decided to stick with the double underline, but change the color of the link to make it more visible.

change/@ni-nimble-components-e3d5df53-15c3-4fe1-8552-ad4db8dbddbd.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @m-akinc the chromatic stories don't seem to reflect the source changes. Reminder about https://github.com/ni/nimble/issues/2118

packages/storybook/src/nimble/anchor/anchor.stories.ts
```diff
@@ -47,7 +47,7 @@ const metadata: Meta<AnchorArgs> = {
             options: Object.keys(AnchorAppearance),
             control: { type: 'radio' },
             description:
-                'Set to `prominent` to make the anchor appear in a different color than normal text.'
+                'Set to `prominent` to make the anchor appear in a different color than normal text. This has no effect under the Color theme.'
```
jattasNI, line 50: nit: while you're here would you mind adding a style to set the color of the non-link text using a token? It looks bad in dark mode right now. 
image

jattasNI (approved):

rajsite (approved):

packages/storybook/.storybook/preview.js
```diff
@@ -105,5 +105,5 @@ configureActions({
     depth: 1
 });
 
-// Update the GUID on this line to trigger a turbosnap full rebuild: 3e3470f8-2a6c-4474-b409-409c1dfbecea
+// Update the GUID on this line to trigger a turbosnap full rebuild: 354fdf0e-bdb4-4dd9-a6e4-21bf71c6efe6
```
rajsite, line 108:
```suggestion
// Update the GUID on this line to trigger a turbosnap full rebuild: 3e3470f8-2a6c-4474-b409-409c1dfbecea
```

PR #2114, opened 2024-05-15
Use display helper consistently for all components

rajsite:
# Pull Request

## 🤨 Rationale

This PR updates all defined components to use a custom version of the `display` helper. The goal of this PR is to be a refactor to use the same helper but should be effectively a no-op.

## 👩‍💻 Implementation

- Added a custom display helper
- Updated all elements to use it
- Updates elements that were missing it to have one with a value that reflects what shows up in storybook todaay as the computed display value
  - Utility elements that should never be visible / have visible content were set to `display: none;`

## 🧪 Testing

Rely on storybook to verify nothing changed.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated the `display` doc topic.

### DIFF
--- a/change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
+++ b/change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Switch to custom display helper",
+  "packageName": "@ni/nimble-components",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/icon-base/index.ts
+++ b/packages/nimble-components/src/icon-base/index.ts
@@ -28,8 +28,7 @@ export const registerIcon = (baseName: string, iconClass: IconClass): void => {
     const composedIcon = iconClass.compose({
         baseName,
         template,
-        styles,
-        baseClass: iconClass
+        styles
     });
 
     DesignSystem.getOrCreate().withPrefix('nimble').register(composedIcon());
--- a/packages/nimble-components/src/table/components/cell/styles.ts
+++ b/packages/nimble-components/src/table/components/cell/styles.ts
@@ -24,8 +24,6 @@ export const styles = css`
 
     .cell-view {
         overflow: hidden;
-        display: flex;
-        align-items: center;
     }
 
     .action-menu {
--- a/packages/nimble-components/src/table/components/cell/styles.ts
+++ b/packages/nimble-components/src/table/components/cell/styles.ts
@@ -1,5 +1,5 @@
 import { css } from '@microsoft/fast-element';
-import { display } from '@microsoft/fast-foundation';
+import { display } from '../../../utilities/style/display';
 import {
     controlHeight,
     controlSlimHeight,
--- a/packages/nimble-components/src/table/components/group-row/styles.ts
+++ b/packages/nimble-components/src/table/components/group-row/styles.ts
@@ -1,6 +1,6 @@
 import { css } from '@microsoft/fast-element';
-import { display } from '@microsoft/fast-foundation';
 import { White } from '@ni/nimble-tokens/dist/styledictionary/js/tokens';
+import { display } from '../../../utilities/style/display';
 import {
     applicationBackgroundColor,
     borderWidth,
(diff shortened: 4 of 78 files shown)


packages/nimble-components/src/icon-base/index.ts
```diff
@@ -28,8 +28,7 @@ export const registerIcon = (baseName: string, iconClass: IconClass): void => {
     const composedIcon = iconClass.compose({
         baseName,
         template,
-        styles,
-        baseClass: iconClass
```
rajsite, line 32: This was for Fast dependency injection (i.e. tagFor) and isn't configured right anyway. The idea is you register a base class to dependency inject the concrete implementation. These `iconClass` are the concrete implementation classes, not base classes, they don't need to be registered to provide for a specific base class. So removed.

packages/nimble-components/src/table/components/cell/styles.ts
```diff
@@ -24,8 +24,6 @@ export const styles = css`
 
     .cell-view {
         overflow: hidden;
-        display: flex;
-        align-items: center;
```
rajsite, line 28: I moved the display configuration to the cell-view implementations. They should be responsible for their self and content styling

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -71,7 +71,6 @@ export const styles = css`
         padding-left: ${mediumPadding};
         ${userSelectNone}
         overflow: hidden;
-        display: flex;
```
rajsite, line 74: I moved the display configuration to the group-row implementations. They should be responsible for their self and content styling.

change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: something seems wrong... why are there 100% turbosnaps... that doesn't seem right

rajsite: https://github.com/ni/nimble/actions/runs/9104155344/job/25027443487#step:16:153

packages/spright-components/src/utilities/style/display.ts
```diff
@@ -0,0 +1,9 @@
+import {
+    type CSSDisplayPropertyValue,
+    // eslint-disable-next-line no-restricted-imports
+    display as foundationDisplay
+} from '@microsoft/fast-foundation';
+
+export const display: typeof foundationDisplay = (
```
jattasNI, line 7: Where did we land on code sharing vs duplication of utilities between Nimble and Spright? I see we have a few other duplicate files in `src/utilities`. Are we tracking the work to share them at some point?

If we're ok duplicating for now, then let's add the same comment as suggested in the nimble-components copy.

jattasNI (commented):
Code changes look good but I'm curious about the turbo snap behavior so I'll wait to approve.

vivinkrishna-ni (approved):

packages/spright-components/src/utilities/style/display.ts
rajsite: Don't think we have an agreed on policy quite yet. I don't really want nimble-components to become a component development infra / tools package + components package so right now we are duplicating utilities. I'll duplicate the comment for now 👍

packages/storybook/.storybook/preview.js
```diff
@@ -104,3 +104,6 @@ export default {
 configureActions({
     depth: 1
 });
+
+// Update the GUID on this line to trigger a turbosnap full rebuild: 3e3470f8-2a6c-4474-b409-409c1dfbecea
+// See https://www.chromatic.com/docs/turbosnap/#full-rebuilds
```
rajsite, line 109: @jattasNI added this to manually disable turbosnap for a build and looks like it helped trigger full snapshotting:
https://github.com/ni/nimble/actions/runs/9116910090/job/25066573475?pr=2114#step:16:54
image

So I think this PR is good. I'll hold off documenting, etc while investigating the behavior. I'll create an issue though.

rajsite: Marked automerge for once resolved (don't think we need to wait for all owners)

munteannatan (approved):

jattasNI (approved):

PR #2119, opened 2024-05-17
Color visited links distinctly

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #2011

## 👩‍💻 Implementation

- Changed color of visited anchors per the [design in Figma](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=8729-84171&t=ZWhzibWRmDT2XoD7-0).
- Created two new theme-aware tokens for this
- Added interesting "visited" cases to the anchor's interaction matrix story. This required updates to the `createMatrixInteractionsFromStates` utility function.

## 🧪 Testing

Added visual tests, and manually tested in Storybook.

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/utilities/matrix.ts
+++ b/packages/storybook/src/utilities/matrix.ts
@@ -110,16 +110,23 @@ export function createMatrixInteractionsFromStates<
     THover extends readonly unknown[],
     THoverActive extends readonly unknown[],
     TActive extends readonly unknown[],
-    TFocus extends readonly unknown[]
+    TFocus extends readonly unknown[],
+    TVisited extends readonly unknown[],
+    TVisitedCombinations extends {
+        plain: TVisited[],
+        active: TActive[],
+        focus: TFocus[]
+    }
 >(
     component: (
-        ...states: THover | TActive | THoverActive | TFocus
+        ...states: THover | TActive | THoverActive | TFocus | TVisited
     ) => ViewTemplate,
     states: {
         hover: THover[],
         hoverActive: THoverActive[],
         active: TActive[],
-        focus: TFocus[]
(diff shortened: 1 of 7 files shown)


m-akinc:
@mollykreis could you please buddy for me?

packages/storybook/src/utilities/matrix.ts
```diff
@@ -110,16 +110,23 @@ export function createMatrixInteractionsFromStates<
...
-        ...states: THover | TActive | THoverActive | TFocus
+        ...states: THover | TActive | THoverActive | TFocus | TVisited
     ) => ViewTemplate,
     states: {
         hover: THover[],
         hoverActive: THoverActive[],
         active: TActive[],
-        focus: TFocus[]
+        focus: TFocus[],
+        visited?: TVisitedCombinations
```
rajsite, line 129: Can you follow the existing pattern and flatten these out instead of making nested options objects?

```
visited?:
visitedActive?:
visitedFocus?:
```

packages/storybook/src/utilities/matrix.ts
```diff
@@ -144,6 +151,20 @@ export function createMatrixInteractionsFromStates<
             <p>Focus</p>
             ${createMatrixFromStates(component, states.focus)}
         </div>
+        ${states.visited ? html`
```
rajsite, line 154: The `when` directive would be more normal in a fast template. Coupled with the above these would split into separate `when` sections

m-akinc:
Based on group discussion, we're going to hold off on coloring visited links. Since we're not sure we want this, we won't keep around a work item for it.

PR #2122, opened 2024-05-18
Angular i18n enable

rajsite:
# Pull Request

## 🤨 Rationale

Enables the i18n attribute lint rule for the angular-workspace. The goal is to have some level of automation in detecting attributes that should not be marked i18n early to add to the javascript style guide angular eslint configuration.

## 👩‍💻 Implementation

- Enabled the rule with a local list of exceptions.
- Found that even though the i18n attributes rule is enabled for all files, it does not seem to run for our tests. Seems to be because our inline templates in tests are not defined at the top-level of the file. Created angular-eslint issue: https://github.com/angular-eslint/angular-eslint/issues/1825

## 🧪 Testing

Rely on CI.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Comments are in the eslint config, didn't seem like additional were neecessary.

### DIFF
--- a/packages/angular-workspace/.eslintrc.js
+++ b/packages/angular-workspace/.eslintrc.js
@@ -1,3 +1,5 @@
+const { ignoreAttributes } = require('@ni/eslint-config-angular/template/options');
+
 module.exports = {
     root: true,
     ignorePatterns: [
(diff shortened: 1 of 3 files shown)


packages/angular-workspace/.eslintrc.js
```diff
@@ -104,9 +106,27 @@ module.exports = {
             '@ni/eslint-config-angular/template'
         ],
         rules: {
-            // nimble-angular shouldn't provide user-visible strings so doesn't need i18n
-            // example-client-app is just for demo purposes and isn't anticipated to be localized
-            '@angular-eslint/template/i18n': 'off'
+            // Enable i18n template checking for the purpose of making sure to capture updates for the lint rules
+            // New entries should be added to the ignoreAttributeSets:
+            // See: https://github.com/ni/javascript-styleguide/blob/main/packages/eslint-config-angular/template/options.js
+            '@angular-eslint/template/i18n': [
```
jattasNI, line 112: Should we add a sentence to nimble-angular CONTRIBUTING explaining the goal to add attributes in javascript-styleguide vs here? I'm also fine with this comment plus a mention at standup, your call.

jattasNI (approved):

packages/angular-workspace/.eslintrc.js
rajsite: I think devs are more likely to discover it here when they hit the rule and for reviewers linking here in PRs would already be straightforward. Don't think CONTRIBUTING would give more context or be easier to link to so I think here is fine. I'll mention it at standup.

Issue #1866, opened 2024-02-23
Filterable Select should support dynamic options


atmgrifter00:
## 😯 Problem to Solve

There is a desire to have the `Select` be able to dynamically populate its options according to current filter input. This is primarily useful when all of the possible options are too numerous to include initially, and it's desirable to retrieve subsets of the options as needed by filtered input.

## 💁 Proposed Solution

There are a couple of aspects to supporting this feature properly:
- We need to supply an event that represents input in the filter.
- It is likely we want to create a new `filterMode` setting that represents a mode where we expect clients to provide the set of options dynamically. This is so that the `Select` does not perform its own filtering scheme against the dynamically provided options.

Unanswered Questions:
- What should the `Select` display as its options on successive openings (after providing a previous filter)?
    - Currently the filter string always starts as blank on open. Does this mean that in the new proposed filter mode, every time a user opens the `Select` it's up to the app author to define what options should be provided (is it always just the previous filtered set?)?
- Does the `Select` need to support the "loading" state while its options are being populated (Issue #1311)?

jattasNI:
One use case for this is the work order picker in SLE 
![image](https://github.com/ni/nimble/assets/10500124/65e5db24-e331-4f00-8d08-88016ba71626)

jattasNI:
Note that Blazor support for this feature was split off into #2253

PR #2125, opened 2024-05-20
[HLD] Select custom filtering.

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1866

## 👩‍💻 Implementation

This is just a spec update.


packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -98,18 +133,10 @@ There is also a desire to allow the [`ListOption` to contain complex content](ht
...
--   Would this feature be enabled through another mode on the `filterMode` enum (i.e. a `dynamic` or `custom` mode), or is it orthogonal to the `filterMode` API?
--   Are there challenges in having the filter work against local options in addition to retrieving new ones?
-
 ## Alternative Implementations / Designs
 
 None
 
 ## Open Issues
+
+-   Does the `Combobox` need to present a mirrored API to the `Select` for dynmically loaded options (i.e. an `isLoading` attribute along with the accompanying visuals, and a `filter-input` event)
```
fredvisser, line 142: We don't have that requirement today, but I could see supporting this in the future. I assume there are no issues with the API?

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+In order to support the use-case of clients supplying options to the `Select` based on the current search text, clients will need to set the `filterMode` to `custom`, and then subscribe to the `filter-input` event, the details of which will provide the current search text:
+
+```ts
+interface SelectFilterInputEventDetail {
+    filterText: string;
+}
+```
+
+Then, in the handling of the `filter-input` event, clients will need to perform the desired filtering using the supplied filter text, in addition to setting the `is-loading` attribute to `true` while the options are being loaded into the DOM, and then again to `false` when that process is complete.
```
fredvisser, line 65:
```suggestion
Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `is-loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete.
```

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
 export const FilterMode  = {
     none: undefined;
     standard: 'standard';
+    custom: 'custom';
 } as const;
 ```
 
 -   The `standard` filterMode will result in case-insensitive, diacritic-insensitive filtering.
--   `filterMode` will default to `none` so as not to affect existing clients.
+-   The `custom` filterMode will result in a search input being shown, but no automatic filtering will occur as the user types in the search box. Instead, it is expected that a client leverage the `filter-input`event described in the ['Dynamic Options'](#dynamic-options) section below.
```
fredvisser, line 49:
```suggestion
-   The `custom` filterMode will display a search input, but it won't automatically filter the options as the user types. Instead, clients can use the `filter-input` event described in the ['Dynamic Options'](#dynamic-options) section below to implement their own filtering logic.
```

fredvisser (approved):

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: I don't believe so.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+```ts
+interface SelectFilterInputEventDetail {
+    filterText: string;
+}
+```
+
+Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `is-loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete.
+
+_IMPORTANT_: When using the `custom` filter mode, clients are responsible for making sure the options that are placed in the DOM meet one of the following criteria:
```
jattasNI, line 67: Clarifying: how would we encourage clients to provide the filtered list of options? By adding / removing elements from the DOM? Marking filtered ones `hidden`?

packages/nimble-components/src/select/specs/filterable-select-hld.md
jattasNI, line 142: I'm not sure the HLD is the right place to hash this out. I think we should probably split it into a separate work item and declare it currently out of scope by listing it as a future consideration.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
 -   The `none` filter mode results in no search input being shown in the dropdown.
+-   `filterMode` will default to `none` so as not to affect existing clients.
 
 _Note: The `filterMode` isn't meant to mirror the `Combobox` `autocomplete` API, as they do serve slightly different purposes: The `autocomplete` for the `Combobox` ultimately helps set the actual value of the `Combobox` as the user types, and isn't necessarily performing any filtering (e.g. the `inline` mode). One possible concern, however, is that we are presenting an API that will allow different types of filter behaviors (i.e. case sensitive) that the `Combobox` does not support. Additionally, I am proposing diacritic insensitive filtering, which the `Combobox` also does not currently support, but I feel this is quite likely a better default experience._
 
+#### Dynamic options
+
+The `filterMode: custom` option allows clients to dynamically supply list options to the `select` control in response to user search text, by subscribing to the `filter-input` event.
```
jattasNI, line 57: Should we fire the event when the user first opens the input? This could be useful to trigger an initial network request for items. I could also imagine apps fetching the initial set when the page loads but that's not guaranteed.

Should we fire the event when the user types in the filter field if the filter mode is `standard`? I don't see a need but it'd be good to clarify.

Should Nimble debounce the event (I'm guessing not) or provide guidance about apps debouncing it?

jattasNI (approved):

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
     ...
     @attr({ attribute: 'filter-mode' })
     public filterMode = FilterMode.none;
+
+    /**
+     * Displays a message at the bottom of the dropdown to indicate more options
+     * are currently being loaded.
+     */
+    @attr({ attribute: 'is-loading' })
+    public isLoading = false;
```
mollykreis, line 36: Should this just be called `loading`? We don't tend to use the word "is" on our boolean attributes (e.g. `hidden`, `open`, `disabled`, etc).

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
     ...
 }
 
 // Provide enum for filterMode to allow for future modes
 export const FilterMode  = {
     none: undefined;
     standard: 'standard';
+    custom: 'custom';
```
mollykreis, line 44: Curious if you considered other words rather than `custom` here? Specifically, I'm wondering if `manual` would be a better term since you are requiring the client to "manually" filter the options. My only [slight] concern with "custom" is that it could be read to mean that a custom filter function/regex will be provided for matching, and that could either lead to client confusion or limit our API choices in the future.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Ultimately, I think we will provide guidance that steers clients towards managing which options are in the DOM in an "intelligent" way, meaning providing some reasonable initial set (that likely should always be available), and additional options that are discovered via filtering. It's possible that we could also suggest that, while filtering, it _could_ make sense to set the `hidden` state on the always present options as needed, but my guess is that this will typically just be an unnecessary complication to just managing the DOM presence of the filtered options.

packages/nimble-components/src/select/specs/filterable-select-hld.md
mollykreis: > Should we fire the event when the user first opens the input?

It would also be good to clarify what happens when the filter input is automatically cleared because the select closes. Would that use the same `filter-input` event? Or would that be a different event that fires when the select opens and/or closes?

atmgrifter00: Here are my initial takes:
- My initial guess is that while this event could be leveraged, I don't think the use-case you outlined is the expected pattern. Waiting for the initial opening of the dropdown before populating the `Select` with an initial set of options doesn't seem typical, and it just makes the implementation more complicated in having to disambiguate the _first_ time the dropdown is opened versus the other times (I'm asserting that making a network request every time the dropdown opens to retrieve the same set of options is an anti-pattern). If clients end up finding a need for this I say we address it as a follow-up.
- At the moment, I was intending to fire the event even when the filter mode was set to `standard`. I'm willing to alter this decision however.
- Nimble should _not_ do any debouncing, in my opinion, in that there really isn't anything to debounce. We don't want to artificially delay the event (i.e. wait 200 ms for another key press before firing the even), as this could just lead to an unnecessary delay. Debouncing actions such as these is pretty standard procedure in a framework like Angular when the intent is to make a network request based on a user typing. Let's just expect clients to continue to use standard procedure here.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: I'm not against the term `manual`. @jattasNI, @rajsite, thoughts?

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Copying Molly's comment for discussion:
> Should we fire the event when the user first opens the input?

It would also be good to clarify what happens when the filter input is automatically cleared because the select closes. Would that use the same `filter-input` event? Or would that be a different event that fires when the select opens and/or closes?

atmgrifter00: > It would also be good to clarify what happens when the filter input is automatically cleared because the select closes. Would that use the same `filter-input` event? Or would that be a different event that fires when the select opens and/or closes?

This is a great question. Would it be better to provide guidance to utilize RxJS patterns that will produce an observable set of options from an observable of options matching a current non-empty filter state, or if the filtered set is empty, then an observable that represents the "initial set" (but would be modified by selected options that weren't in initial set).

atmgrifter00: Talked it through with @mollykreis, and I think we're in agreement that firing the `filter-input` when the dropdown closes (containing an empty string for the filter text) is sufficient for clients to achieve what they need (and ultimately this is just an implementation detail as clients should already be handling what should happen when the filter input is empty).

packages/nimble-components/src/select/specs/filterable-select-hld.md
jattasNI: I'm struggling to think of existing native or Nimble APIs that we could use as precedence. So I'm left with personal opinions. I like `custom` marginally better than `manual` with the not-very-compelling rationale that "manual" for me has connotations of a human tediously filtering items by hand. 

Some other ideas:
- `filter-mode="event"` - you filter by responding to events. `input-event` or `filter-input-event` are possible variations
- `filter-mode="dynamic"` - the items can change dynamically. `programmatic` is a possible variation
- `filter-mode="client"` - the client provides the filtering. `client-provided` or `user-defined` are possible variations.

I'm not seeing a clear winner. My personal preferences are `dynamic` or `custom`.

mollykreis: I like `dynamic`, especially over `custom`. The places we use `custom` today tend to be in table columns where a number of other `custom-<configuration-option>` attributes are also used to refine the "custom" behavior. That's pretty different than what we're doing here.

atmgrifter00: Going with `dynamic`.

rajsite: Sorry to hop on late but to me "dynamic" is in contrast to "static" so as just a name it's not intuitive to me what filter-mode "dynamic" vs "standard" vs "none" mean.

What is dynamic is how the user is expected to treat options, but really options are always "dynamic"; we only show options that are slotted and what we show changes regardless of this attribute configuration.

Here the filtering is really "disabled". We aren't doing dynamic filtering, we are doing no filtering.

rajsite: I'd propose going with either `manual` or `custom` with the options given. And given the above examples of how we use custom to tweak config instead of completely change, I have a slight preference for `manual`

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
     ...
     @attr({ attribute: 'filter-mode' })
     public filterMode = FilterMode.none;
+
+    /**
+     * Displays a message at the bottom of the dropdown to indicate more options
+     * are currently being loaded.
+     */
+    @attr()
+    public loading = false;
```
rajsite, line 36: If my understanding is correct this is purely visual and not coupled to any other attribute. This is for just showing or hiding a part of the control and doesn't really have a behavior change. If that tracks then I'd vote for `show-loading`

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+The `filterMode: dynamic` option allows clients to dynamically supply list options to the `select` control in response to user search text, by subscribing to the `filter-input` event.
+
+```ts
+interface SelectFilterInputEventDetail {
+    filterText: string;
+}
+```
+
+Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete. A `filter-input` event will also be emitted with its `filterText` set to empty string upon closure of the dropdown if both the `filter-mode` was not set to `none` _and_ the filter input contains non-empty text.
```
rajsite, line 65: There are a lot of conditions in this sentence. I don't think we should be overly complex about when we fire `filter-input`. Can we specify it as whenever the user enters text in the filter text field or the select closes (clearing the filter) a filter-event is fired. Not coupled to any other attribute configuration (i.e. filter-mode manual or standard, doesn't matter, just matters about interaction with the text field).

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,58 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+_IMPORTANT_: When using the `dynamic` filter mode, clients are responsible for making sure the options that are placed in the DOM meet one of the following criteria:
+
+-   The option is some reasonable default option when there is no filter text provided _OR_
+-   The option display text matches the provided filter in an understandable way _OR_
+-   The option is a placeholder option
```
rajsite, line 71: It's clear to me the implications of this requirement. Conceptually changing out options is independent of the filter-mode. Programmatically a user could do it at any time. It shouldn't be unique to filter mode. And it's a declarative state, whatever is slotted in has a meaning and well-defined behavior.

The only thing I could think of is that user's should be aware that they need to keep whatever item is currently selected in the options list as selected, even if it doesn't match their custom filter algorithm so it continues to show as the current display text. Otherwise they will accidentally trigger a value change (but not a value change event).

And if they have something configured as a placeholder and they want it to keep showing as a placeholder, they need to makee sure it does not disappear in the list (and may need to make sure it stays first in the list / they don't mark multiple things `disabled hidden`).

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -55,12 +84,16 @@ export class LabelProviderCore
 
     @attr({ attribute: 'select-filter-no-results' })
     public selectFilterNoResults: string | undefined;
+
+    @attr({ attribute: 'select-loading-options' })
+    public selectLoadingOptions: string | undefined;
```
rajsite, line 89: discussed with meyers and seems like this and the above other labels, "selectFilterSearch", "selectFilterNoResults", "selectLoadingOptions" are being overly coupled to the select and could be more general concepts. The name proposed here is okay if we do not want a breaking change right now. If we can handle a breaking change we should consider "filterSearch", "filterNoResults", and "loading" for more general applicability (i.e. combobox, rich-text user mention, table filter, etc.)

rajsite (approved):
Discussed feedback with meyers and think on the same page with all the feedback so marking approve

packages/nimble-components/src/select/specs/filterable-select-hld.md
jattasNI: I can live with the emerging consensus of `manual`

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Updated.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Updated.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: I'm at least going to just have the new attribute be called `loading`. We can leave the name changes for the other attributes for another time/PR.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Update: The first sentence of the 'Dynamic options' section mentions the expected client implementation. This will also be provided in the API docs.

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Updated: I've added notes regarding debouncing, as well as when the `filter-input` event will be emitted.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+_IMPORTANT_: When using the `manual` filter mode, clients are responsible for the following:
+
+-   The currently selected option is included in the set of options placed in the DOM regardless of match against filter (and have its `hidden` state adjusted as needed)
+-   Placeholder options should not be filtered out
+
+##### Groups
+
+When using the `filterMode: manual` option along with [groups](./option-groups-hld.md), clients will also need to determine the appropriate matching behavior against the `ListOptionGroup` elements. Our guidance will be that filter text should be able to be matched against _any_ text in the group label, and when that matches, _all_ options within that group should be visible.
```
jattasNI, line 74: Based on #2111, empty groups will not be displayed, correct? If so, do you think it would be useful to mention that here? Something like this, though I acknowledge it's a bit wordy and maybe more informational than critical:
```suggestion
When using the `filterMode: manual` option along with [groups](./option-groups-hld.md), clients will also need to determine the appropriate matching behavior against the `ListOptionGroup` elements. Our guidance will be that filter text should be able to be matched against _any_ text in the group label, and when that matches, _all_ options within that group should be visible. Conversely, the select will automatically hide groups with no items in them. This means clients only need to conditionally remove items, not groups (though it's fine if they manually remove empty groups too).
```

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -88,25 +128,19 @@ The accessibility tree will report that the search `input` element should have i
 
 ### Future considerations
 
-#### Grouping/Metadata
+It's possible that we may want to improve/alter the discoverability of the fact that more options are available to be loaded when using the `manual` filtering option. This _could_ follow the design of what we see with the AzDO _Windows_-based user selector experiece where a display of the number of current results is shown at the bottom, and when a user begins typing in the filter input, a 'Show more results' button is shown at the bottom of the dropdown.
```
jattasNI, line 131: > the AzDO _Windows_-based user selector experiece

- typo in "experience"
- what do you mean by Windows-based? Is this not available in the web UI?
- all this background seems of questionable value. I'd vote to replace it with something more vague ("For example, the dropdown could include a "Show more results button"), but if you want to keep it then maybe a screenshot would be more helpful to explain it.

jattasNI (approved):

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: I've removed the description for a possible solution, and instead just outlined the problem a bit more clearly.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
     ...
     @attr({ attribute: 'filter-mode' })
     public filterMode = FilterMode.none;
+
+    /**
+     * Displays a message at the bottom of the dropdown to indicate more options
+     * are currently being loaded.
+     */
+    @attr()
+    public loading-visible = false;
```
mollykreis, line 36: nit: you should probably be consistent in how technically correct this property/attribute is with `filterMode` above.

```suggestion
    @attr({ attribute: 'loading-visible', mode: 'boolean' })
    public loadingVisible = false;
```

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+    filterText: string;
+}
+```
+
+Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete. The `filter-input` event is emitted any time the user types a character in the filter input and also when the dropdown is closed (in which case the `filterText` in the details will always be an empty string).
+
+_IMPORTANT_: When using the `manual` filter mode, clients are responsible for the following:
+
+-   The currently selected option is included in the set of options placed in the DOM regardless of match against filter (and have its `hidden` state adjusted as needed)
+-   Placeholder options should not be filtered out
```
mollykreis, line 70: There's only one placeholder option in a given select, right? Perhaps this could be written as _If a placeholder option exists for the select, it should never be filtered out._

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+interface SelectFilterInputEventDetail {
+    filterText: string;
+}
+```
+
+Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete. The `filter-input` event is emitted any time the user types a character in the filter input and also when the dropdown is closed (in which case the `filterText` in the details will always be an empty string).
+
+_IMPORTANT_: When using the `manual` filter mode, clients are responsible for the following:
+
+-   The currently selected option is included in the set of options placed in the DOM regardless of match against filter (and have its `hidden` state adjusted as needed)
```
mollykreis, line 69: > and have its `hidden` state adjusted as needed

It isn't clear to me what "as needed" means. Could you clarify when it would be needed? I'm guessing it should be marked as `hidden` if it doesn't match the current filter?

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+```ts
+interface SelectFilterInputEventDetail {
+    filterText: string;
+}
+```
+
+Then, when handling the `filter-input` event, clients should perform the desired filtering using the provided filter text. Additionally, they should set the `loading` attribute to `true` while the options are being loaded into the DOM, and then set it back to `false` when the loading process is complete. The `filter-input` event is emitted any time the user types a character in the filter input and also when the dropdown is closed (in which case the `filterText` in the details will always be an empty string).
+
+_IMPORTANT_: When using the `manual` filter mode, clients are responsible for the following:
```
mollykreis, line 67: The client is also responsible for making sure that options that don't match the filter are either removed from the DOM or marked as `hidden`. I'm surprised that isn't listed as one of the client responsibilities in `manual` mode. Even if it seems obvious, I think it would be good to have it clearly listed here in the bullet points of client responsibilities.

packages/nimble-components/src/select/specs/filterable-select-hld.md
```diff
@@ -20,29 +20,65 @@ Our clients have a need of a filterable dropdown component that does not allow a
...
+
+##### Groups
+
+When using the `filterMode: manual` option along with [groups](./option-groups-hld.md), clients will also need to determine the appropriate matching behavior against the `ListOptionGroup` elements. Our guidance will be that filter text should be able to be matched against _any_ text in the group label, and when that matches, _all_ options within that group should be visible. Conversely, the select will automatically hide groups with no items in them. This means clients only need to conditionally remove items, not groups (though it's fine if they manually remove empty groups too).
+
+_Notes_:
+
+-   When the `loading-visible` attribute is set, we will display localizable text at the bottom of the dropdown (defaulting to "Loading"), along with the spinner icon.
+-   The `Select` will not peform any debouncing as the user types into the filter input. It is expected that clients can perform any debouncing that is needed easily at the app level.
+-   The `filter-input` event will be emitted for all filter modes (except `none`)
```
mollykreis, line 80: These notes seem incorrectly placed under the _Groups_ section of this document since they aren't related to grouping.

mollykreis (approved):

packages/nimble-components/src/select/specs/filterable-select-hld.md
atmgrifter00: Updated.

PR #2130, opened 2024-05-21
Update banner button styles

mollykreis:
# Pull Request

## 🤨 Rationale

The "dismiss" button in the `nimble-banner` had the wrong styling in the following ways:
- It was the wrong size. It was 16x16, but it should have been 24x24.
- It was using the wrong icon. It was using the `xmark` icon, but it should have been using the `times` icon.
- The hover styling was wrong. It had custom styling to make it look like a ghost button with a fill on hover (not like any of our standard `nimble-button` appearance modes). Instead, it should be a standard `content-hidden`, `appearance="ghost"`, `appearance-variant="primary"` button.

Note: A number of these differences were because the design spec was updated after the banner was initially implemented.

## 👩‍💻 Implementation

1. Wrapped the contents of the banner's template in a theme provider hard-coded to the `color` theme
2. Updated storybook docs to specify that a slotted `action` button should use the `primary` appearance variant
3. Updated instances of the banner within nimble to set `appearance-variant="primary"` on slotted buttons

Those three changes allowed me to remove a lot of custom styling from the banner.

## 🧪 Testing

- Verified that the only chromatic change was due to the "dismiss" button

## ✅ Checklist


mollykreis:

change/@ni-nimble-components-db9e952a-a79d-4aa2-a8e9-b3455ee2b620.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update banner styling to align with visual design spec",
```
mollykreis, line 3: Assuming we're happy with the approach in this PR, I'll update the two instances in azdo of buttons slotted within the banner to have `appearance-variant="primary"` before submitting this PR. I've verified that having the existing styling + `appearance-variant="primary"` doesn't have any adverse side-effects.

change/@ni-nimble-components-db9e952a-a79d-4aa2-a8e9-b3455ee2b620.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
mollykreis, line 2: Thoughts on calling this a `patch` change. Technically, we are telling clients that they now need to add `appearance-variant="primary"` to buttons put in the `action` slot of the banner. Do we want to consider that a breaking change?

m-akinc:
> Note: A number of these differences were because the design spec was updated after the banner was initially implemented.

Thanks for exonerating me! 😁😅

packages/nimble-components/src/banner/styles.ts
```diff
@@ -85,6 +78,8 @@ export const styles = css`
         align-items: center;
         justify-content: center;
         align-self: flex-start;
+        margin-top: ${smallPadding};
+        ${controlHeight.cssCustomProperty}: ${controlSlimHeight};
```
m-akinc, line 82: slick 👍

packages/nimble-components/src/banner/template.ts
```diff
@@ -4,65 +4,70 @@ import { buttonTag } from '../button';
 import { iconExclamationMarkTag } from '../icons/exclamation-mark';
 import { iconInfoTag } from '../icons/info';
 import { iconTriangleFilledTag } from '../icons/triangle-filled';
-import { iconXmarkTag } from '../icons/xmark';
+import { iconTimesTag } from '../icons/times';
```
m-akinc, line 7: I still hate that we're using a math-named icon rather than the more generic "x mark" for a close button. I kind of want to alias it or duplicate it with the name "dismiss" or something like that.

change/@ni-nimble-components-db9e952a-a79d-4aa2-a8e9-b3455ee2b620.json
m-akinc: Personally, I think the visual difference is subtle enough that we can call this a patch, especially since you plan to update our SLE clients.

Did you consider creating a handler for when the slotted content changes and _forcing_ slotted buttons to one of the prescribed appearances and appearance variant? We already force slotted content to a 24px height, so it doesn't seem like such a strech to also impose other styling requirements (even if via an attribute).

m-akinc (approved):

packages/nimble-components/src/banner/template.ts
mollykreis: I don't disagree that "times" isn't the most intuitive name for an icon used to close things. Our docs already list "close" as one of its metaphors, and we don't have any conventions around exporting the same icon with different names. So, I'm not sure there is much actionable to do with this. I'll leave the conversation open in case any other reviewers have thoughts on this.

change/@ni-nimble-components-db9e952a-a79d-4aa2-a8e9-b3455ee2b620.json
mollykreis: I thought briefly about it, but we don't tend to update appearance attributes of slotted elements. I think there's a difference between styling slotted content to place & size it appropriately within another component and changing appearance attributes on it.

I also brought this up in a team meeting, and @rajsite agreed that adding best practice docs rather than mutating the client's element was aligned with how things are expected to work in nimble.

jattasNI: I'm ok with `patch` too.

packages/nimble-components/src/banner/template.ts
jattasNI: I agree it's poorly named, since names should describe the shape of the icon rather than their metaphors (unless there's only one relevant metaphor, which isn't the case here). I think the fix would be to rename it and provide good aliases rather than create a copy. But I'm not sure the best way to distinguish the two x icons. Maybe something like:
- `xmark` (status, fail) (unchanged from today)
- `xmark-2`or `xmark-thin` (times, close)

If we want to stage the deprecation we could temporarily duplicate it with a new name, but I'd want us to clean up the old copy once we've migrated important clients away from it. But it'd still be a breaking change to remove it.

I don't think any of that should be in this PR. If either of you feel strongly enough, let's file an issue and settle on names when we triage it.

packages/storybook/src/nimble/banner/banner.stories.ts
```diff
@@ -97,7 +98,7 @@ export const _banner: StoryObj<BannerArgs> = {
             options: Object.values(ActionType),
```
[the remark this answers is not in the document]
mollykreis: It looks like the styling is correct for the color theme. Technically, `severity` does have an effect in color theme -- it just doesn't change the background color of the banner. It still controls the icon in the banner. I'll clarify the docs.

change/@ni-nimble-components-db9e952a-a79d-4aa2-a8e9-b3455ee2b620.json
mollykreis: Azdo PRs have been submitted

packages/nimble-components/src/banner/template.ts
m-akinc: Filed #2139

Issue #2118, opened 2024-05-16
Chromatic Turbosnap not detecting component source changes


rajsite:
# 🧹 Tech Debt

In a PR that only has nimble-components source file changes (no package.json, etc or storybook package changes), chromatic turbosnap is not detecting the changes and is skipping all snapshots.

Workaround:

- Check the build log Chromatic Publish step for "Capturing 0 snapshots and skipping 580 snapshots." to know if all snapshots were skipped even if they were expected
- Tweak the `.storybook/preview.js` file comment at the bottom that has a guid you can update to make a change that will trigger a [full turbosnap rebuild](https://www.chromatic.com/docs/turbosnap/#full-rebuilds).

Acceptance:

- Changes to nimble-components should trigger precise snapshots for turbosnap
- Changes to nimble-tokens should trigger a full rebuild
- Remove the comment in `.storybook/preview.js` for triggering a rebuild

PR #2131, opened 2024-05-22
Storybook link directly to components source code

rajsite:
# Pull Request

## 🤨 Rationale

This PR is an abomination but hopefully fixes https://github.com/ni/nimble/issues/2118

Turbosnap relies on git source history to detect changes. As a separate package, the storybook build was correctly implemented to rely on the build output of nimble-components and spright-components instead of the non-public source.

However there is not a link that Turbosnap can detect from the source in git to the ephemeral dist output paths that ends up being used during build. So Turbosnap was now unable to follow the dependency path from a story file to specific component source files.

## 👩‍💻 Implementation

Do as minimally scoped a change as possible to enable storybook to link directly to non-public source code in the nimble-components and spright-components packages. Any other non-mechanical changes have comments in the PR.

## 🧪 Testing

I think [this build](https://github.com/ni/nimble/actions/runs/9194886405/job/25289396570#step:16:78) where I reverted a change that was only in the spright rectangle css source confirms that the approach will work. You can see turbosnap detected the original component source file changes for the rectangle story in the log:

```
Found 5 changed files:
  change/@ni-spright-components-a39f0ecb-4139-47f6-86f7-c22779ee1f08.json
  package.json
  packages/spright-components/src/rectangle/styles.ts
  packages/storybook/.eslintrc.cjs
  packages/storybook/README.md
...
Found affected story files:
  packages/storybook/src/spright/rectangle/rectangle-matrix.stories.ts [./src/spright/rectangle/rectangle-matrix.stories.ts]
  packages/storybook/src/spright/rectangle/rectangle.stories.ts [./src/spright/rectangle/rectangle.stories.ts]
  packages/storybook/src/spright/rectangle/rectangle.mdx [./src/spright/rectangle/rectangle.mdx]
...
✔ TurboSnap enabled
Capturing 8 snapshots and skipping 582 snapshots.
```

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated the storybook README with a warning this is not a general pattern to copy.


packages/storybook/src/utilities/matrix.ts
```diff
@@ -41,7 +41,9 @@ export function cartesianProduct<T extends readonly unknown[]>(
         ...states: readonly unknown[]
     ): void => {
         if (currentDimensions && currentDimensions.length >= 1) {
-            const [currentDimension, ...remainingDimensions] = currentDimensions;
+            const [currentDimensionOrUndefined, ...remainingDimensions] = currentDimensions;
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            const currentDimension = currentDimensionOrUndefined!;
```
rajsite, line 46: This one is super weird and I didn't figure it out. Typescript for some reason is now failing to infer that this value cannot be undefined when the storybook build runs (normally TS can see that we did a length check on the array so this is definitely not undefined). So I added the non-null assertion to help TS.

However the TS config as part of eslint-typescript (whish is the same tsconfig file) is still able to infer that the value is non-null so complains that we have an unnecessary non-null assertion. So I added an eslint disable for it.

packages/storybook/tsconfig.json
```diff
@@ -1,13 +1,28 @@
 {
     "compilerOptions": {
+      "declaration": true,
```
rajsite, line 3: storybook is now responsible for being able to build nimble-components and spright-components TypeScript source instead of just using build output so the tsconfig needs to align so there  is not weird behavior differences between the two during TS build.

I did not want to make sharing the TS config part of this change. I also want to see if switching to relying on the source actually fixes the turbosnap issues before leaning in harder on this pattern.

packages/storybook/src/global-type-extensions.d.ts
```diff
@@ -0,0 +1,28 @@
+// This file has modifications to global types used for nimble-components builds:
```
rajsite, line 1: In order to get nimble-components source to successfully build these globals defined in nimble-components needed to get redefined here (they are used only during source build). To me this is more of an indication that this pattern isn't great. I think we should try and enable `isolatedModules` in the tsconfig and then find an alternative pattern to these d.ts files. For now just copied the declaration from nimble to here.

packages/storybook/.eslintrc.cjs
```diff
@@ -14,7 +14,22 @@ module.exports = {
                 'import/no-extraneous-dependencies': [
                     'error',
                     { devDependencies: true }
-                ]
+                ],
+                // This rule is disabled in order to allow linking directly to nimble-components and spright-components source
+                // Other files in the repo should NOT directly link to source and should resolve with the package identifier instead
+                'import/no-relative-packages': 'off',
+                'no-restricted-imports': [
```
rajsite, line 21: Added a lint rule to prevent folks from accidentally relying on both build output and original source files from the components packages

packages/storybook/src/global-type-extensions.d.ts
jattasNI: Could you add comments at the top of both copies that points at the other copy and indicates they should be kept in sync?

It also looks like we are using a recent enough typescript version (5.1) to remove the `HTMLDialogElement` interface (4.8) but not quite the `NumberFormatOptions` one (5.5). Wanna try removing the dialog one, either in this PR or a follow up?

packages/storybook/src/utilities/matrix.ts
jattasNI: Can you add a brief comment to this effect (mostly to indicate that we thought about it, doesn't need to be exhaustive)?

Something like:
`// TypeScript and ESLint disagree about whether this can be null or undefined due to the way the this package links to nimble-components`

packages/storybook/tsconfig.json
jattasNI: Can you make sure we have a follow up task to track sharing config? Or add a note to #2118 and keep it open after this PR goes in? It seems highly likely they'll accidentally diverge over time.

jattasNI (approved):

packages/storybook/src/global-type-extensions.d.ts
rajsite: Added a comment and removed the dialog part in this PR

rajsite: Also made the components change type 'none' since it should be a nimble repo build time type checking change and not have an impact on published packages.

packages/storybook/tsconfig.json
rajsite: Created https://github.com/ni/nimble/issues/2133

package.json
```diff
@@ -19,7 +19,7 @@
     "validate:test:sequential": "npm run test --workspaces --if-present",
     "validate:lint-concurrent:nimble-components": "npm run lint-concurrent -w packages/nimble-components",
     "validate:test-concurrent:nimble-components": "npm run test-concurrent -w packages/nimble-components",
-    "storybook": "concurrently -n nimble,spright,storybook \"npm run build-components:watch -w @ni/nimble-components\" \"npm run build-components:watch -w @ni/spright-components\" \"npm run start -w @ni-private/storybook\""
+    "storybook": "npm run start -w @ni-private/storybook"
```
m-akinc, line 22: @rajsite Just noticed this no longer builds/watches nimble and spright components. It won't work anymore?

m-akinc: Great. Just confirmed. I think my nimble-components build was broken, and I made source fixes then ran `npm run storybook`. I expected it to build nimble-components, so I was surprised when it didn't.

PR #2117, opened 2024-05-16
Improve API docs for buttons and improve overall doc consistency

jattasNI:
# Pull Request

## 🤨 Rationale

As part of #824 we want to improve the accuracy and comprehensiveness of our storybook API documentation.

## 👩‍💻 Implementation

1. Update `button`, `toggle-button`, and `card-button` to follow new API doc pattern. (I'll do `anchor-button` along with other anchors).
2. Add and tweak wording on shared documentation strings and utilities
3. Tweak organization of MDX headings in all stories to put Appearance and Sizing under Styling
4. Add CONTRIBUTING guidance about referring to sections within docs (in previous PRs we gave up on trying to link to sections and started using bolded text instead).

## 🧪 Testing

Local storybook exploration

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2132, opened 2024-05-22
Improve API documentation for input components

jattasNI:
# Pull Request

## 🤨 Rationale

Ongoing efforts towards #824. This PR covers input components:
- select
- combobox
- radio group
- text area
- text field
- number field
- checkbox
- switch

## 👩‍💻 Implementation

Generally follow patterns from previous PRs like #2126 and #2117. A few interesting notes:
1. components have different behaviors regarding whether they sync their `value` property to an attribute. I documented the behavior I observed by adding a new table category for properties that don't have attributes (also used by checkbox indeterminate property).
2. not yet documenting a recommendation to use form association / CVAs instead of change events and value properties. Until I do so, the `placeholder` docs for select are outside any table category.
3. added another new table category for localizable strings.
4. not yet documenting list option API in the select and combobox stories. We (probably Meyer) are going to tackle this in a follow up after #2111.

## ✅ Checklist


packages/storybook/src/docs/component-apis.mdx
```diff
@@ -45,7 +45,7 @@ Or in a named slot:
 ## Attributes and Properties
```
jattasNI, line 45: I will wait for #2131 to go in before merging this.

Also conflicts with #2111 but we'll see which is ready first.

jattasNI:
@m-akinc would you mind being my buddy for API docs changes while Milan is out of town?

packages/storybook/src/nimble/text-field/text-field.stories.ts
```diff
@@ -65,31 +68,77 @@ const metadata: Meta<TextFieldArgs> = {
...
+        },
+        placeholder: {
+            description: placeholderDescription({ componentName: 'text field' }),
+            table: { category: apiCategory.attributes }
+        },
         type: {
             options: Object.values(TextFieldType),
-            control: { type: 'select' }
+            control: { type: 'select' },
+            description: 'They type of input to accept and render in the text field. This corresponds to [the `type` attribute of the native `input` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#type) though only a subset of values are supported.',
```
m-akinc, line 83: Shouldn't need to open the dropdown to see what values are supported. Alternatively, could use radio buttons instead.
```suggestion
            description: `The type of input to accept and render in the text field. This corresponds to [the \`type\` attribute of the native \`input\` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#type) though only the following values are supported:
   - text (default)
   - email
   - password
   - tel
   - url`,
```

m-akinc (approved):
Just to warn you, I left a lot of feedback about things that aren't part of your changes, so it's up to you if you want to make those changes in this PR, or create one or more tech debt items to track them. This documentation is starting to look pretty legit!

packages/storybook/src/nimble/checkbox/checkbox.stories.ts
```diff
@@ -29,13 +30,31 @@ const metadata: Meta<CheckboxArgs> = {
         </${checkboxTag}>
     `),
     argTypes: {
+        label: {
+            name: 'default',
```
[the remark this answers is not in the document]
jattasNI: I agree it would be good to distinguish actual names of slots/attributes from just descriptions like this. I will tackle this in a follow up PR. I'm tracking it in a task list in #824.

My only other idea besides "(default)" is "Default slot", under the hope that the space in its name makes it obvious that it's not `slot="Default slot"`.

packages/storybook/src/nimble/text-field/text-field.stories.ts
jattasNI: Changed it to a radio button.

jattasNI:
Bypassed @rajsite due to OOO.

PR #2140, opened 2024-05-29
Improve API documentation of table components

jattasNI:
# Pull Request

## 🤨 Rationale

Ongoing efforts towards #824. This PR covers the table and its columns.

## 👩‍💻 Implementation

Mostly following established patterns. Some interesting notes:
1. For the Table Column Configuration doc, we had divided the documentation into separate targeted stories. I decided these were useful but we also should have a single table describing the whole API of a column. I created the latter under the API section and moved the targeted stories to an Examples section. Let me know if it feels right!
2. The above resulted in a renamed story which appears as a Chromatic diff.

## 🧪 Testing

Interacting with the built Storybook.

## ✅ Checklist


jattasNI:
@m-akinc Here's another round of API docs for you to buddy. Thanks!

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
```diff
@@ -136,6 +136,96 @@ const metadata: Meta<SharedTableArgs> = {
...
+
+export const api: StoryObj<ApiTableArgs> = {
+    parameters: {
+        // Story used by documentation, not needed for visual comparison.
+        chromatic: { disableSnapshot: true }
+    },
+    argTypes: {
+        content: {
+            name: 'default',
+            description: 'Text or an icon added to the default slot of the column will be placed in the column header. See **Setting header content** for more information.',
```
m-akinc, line 161: I would suggest we link this and other similar references to the section they're referencing (i.e. `#setting-header-content`). Unfortunately, I can't get links to open in the same iframe. They instead load the iframe content as the top-level page, losing the navigation pane, etc. Seems like the default `target` value for a link should be `"_self"`, but it doesn't work that way. If I manually add `target="_self"` to the generated link in dev tools, then it works properly.

packages/storybook/src/nimble/table/table.mdx
```diff
@@ -9,9 +10,23 @@ The <Tag name={tableTag}/> is a component that offers a way to render tabular da
 For information about configuring table columns, see **Table Column Configuration** and **Table Column Types**.
```
m-akinc, line 10: > For information about configuring table columns, see **Table Column Configuration** and **Table Column Types**.

Can we link the referenced docs? Also, we no longer have a page called "Table Column Types", so maybe we just remove that.

packages/storybook/src/nimble/table/table.stories.ts
```diff
@@ -326,54 +334,101 @@ export const table: StoryObj<TableArgs> = {
...
+        },
+        columnConfigurationChange: {
+            name: 'column-configuration-change',
+            description: 'Event emitted when the user changes a column\'s width or sort order.',
+            control: false,
+            table: { category: apiCategory.events },
+        },
+        rowExpandToggle: {
+            name: 'row-expand-toggle',
+            description: 'Event emitted when the user expands or collapses a row in a table with hierarchy.',
```
m-akinc, line 428: Should we say that this is not emitted for group rows? (Or if it actually is, say that instead.)

m-akinc (approved):

packages/storybook/src/nimble/table/table.mdx
```diff
@@ -9,9 +10,23 @@ The <Tag name={tableTag}/> is a component that offers a way to render tabular da
...
+### Sizing
+
+The <Tag name={tableTag}/> should be sized explicitly or sized to fill the space of a parent container. The <Tag name={tableTag}/> does not currently support being styled with `height: auto`. The ability to auto size the table is tracked with [issue 1624](https://github.com/ni/nimble/issues/1624).
+
+{/* ## Usage */}
 
-## Delay-loaded Hierarchy
+## Examples
+
+### Delay-loaded Hierarchy
```
m-akinc, line 29: Should we use sentence case for these headings rather than title case? We used sentence case in the "Table Column Configuration" examples.

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
jattasNI: Yeah, Milan and I played with this before and came to the same conclusion. I've documented the best practice of marking headings as bold in the storybook CONTRIBUTING.

packages/storybook/src/nimble/mapping/icon/mapping-icon.stories.ts
```diff
@@ -23,28 +23,33 @@ export const iconMapping: StoryObj = {
...
             control: false,
-            description: 'The tag name of the Nimble icon to render, e.g. `nimble-icon-check`.'
+            description: 'The tag name of the Nimble icon to render, e.g. `nimble-icon-check`.',
+            table: { category: apiCategory.attributes },
         },
         severity: {
             control: false,
             description:
-                'Must be one of the values in the `IconSeverity` enum. Controls the color of the icon.'
+                'Must be one of the values in the `IconSeverity` enum. Controls the color of the icon.',
```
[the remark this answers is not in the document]
jattasNI: I don't have great ideas about how to do this either. But I added it to #824 to think some more about.

packages/storybook/src/nimble/mapping/icon/mapping-icon.stories.ts
```diff
@@ -23,28 +23,33 @@ export const iconMapping: StoryObj = {
...
+            table: { category: apiCategory.attributes },
         },
         text: {
             control: false,
             description: `A textual description of the value. The text will be displayed next to the icon in a cell if \`text-hidden\` is not set,
             or as the tooltip and accessible  name of the icon if \`text-hidden\` is set. The text is also displayed next to the icon
-            in a group header. This attribute is required.`
+            in a group header. This attribute is required.`,
```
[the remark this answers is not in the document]
jattasNI: Added to #824

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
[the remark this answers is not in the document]
jattasNI: I think the same could be said for every story that's embedded in an mdx page and also displayed below it in Storybook's table of contents. The only unique value I can see for some of them now is to see the event data in the Actions tab. This should be tackled holistically, so I added it to #824.

jattasNI:
Bypassing reviewer policy due to OOO.

Issue #2136, opened 2024-05-23
Prettier not running on storybook files


m-akinc:
# 🧹 Tech Debt

Prettier does not seem to be formatting on save like it used to when the story files were part of the `nimble-components` package.

m-akinc:
Check with Jesse to make sure there aren't painful merges.

fredvisser:
We need to update the .prettierignore file:

```
/*
!/packages
/packages/*
!/packages/storybook
!packages/nimble-components
```

PR #2144, opened 2024-05-31
Icon mapping doc update

fredvisser:
# Pull Request

## 🤨 Rationale

I had an action item to document best practices for when to use the different icon/mapping column configurations.

## 👩‍💻 Implementation

Added a table to the existing best practices section of the [Table Column: Mapping](https://nimble.ni.dev/storybook/iframe.html?viewMode=docs&id=components-table-column-mapping--docs#best-practices) docs.

Formatted the doc with prettier locally, as #2136 prevented it from happening automatically (and I refuse to format markdown tables manually)

## 🧪 Testing

Manual Storybook review

## ✅ Checklist


packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -25,35 +33,51 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
...
+-   Provide a mapping for every expected record value. Because grouping is
+    performed on the record value, non-mapped record values can result in
+    multiple groups without group labels.
+    -   To improve grouping behavior of a value that shouldn't be rendered in
+        table cells, use a <Tag name={mappingEmptyTag}/> element to provide text
+        for that value's group row.
+-   Avoid having multiple values that map to the same text because grouping and
+    sorting is done on the record value rather than the mapped text.
+
+#### Choosing a mapping style
```
jattasNI, line 69: For each place we refer to icon mappings in this guidance, it would be good to clarify that the guidance applies to spinner mappings too. If it's too wordy to include that everywhere, we could instead include a note at the beginning or end of the table, or even a separate row in the table. 

The subtlety is that "icon" is interchangeable with "spinner" for mappings but, I assume, NOT for headers. So maybe something like:

*Spinner mappings can be used in place of icon mappings throughout this guidance*

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -25,35 +33,51 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
...
+| **Icon-only, with an icon header** <br/> Use only when table width is at a premium and all the icons in the set have a clear and unambiguous meaning                                           | <img src={iconOnlyIconHeader} />             |
+| **Icon-only, with a text header** <br/> Use when header text provides enough context for all the icons in the set to have a clear and unambiguous meaning                                      | <img src={iconOnlyTextHeader} />             |
+| **Text-only, with a text header** <br/> Use when there isn't an icon for each mapping value                                                                                                    | <img src={textOnlyTextHeader} />             |
+| **Icon-only column beside a text-only column** <br/> Use to be compatible with existing columns, or when the end-user is likely to remove the text-only column in a custom table configuration | <img src={iconOnlyPlusTextOnlyTextHeader} /> |
```
jattasNI, line 77: How about a note about mixing mapping types? Could we explicitly recommend NOT mixing text mappings with icon mappings in the same column since it would result in misalignment? 

And what about mixing `text-hidden` values on mappings within the same column? We do it in the example above where the spinner mapping sets `text-hidden` but the icon mappings don't. But I don't think we'd recommend a mixture of icon mappings where some show text and some don't.

(I'm borderline on whether this is over explaining; I'm open to letting our clients use common sense and not documenting this at all)

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -25,35 +33,51 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
 <Controls of={tableColumnMappingStories.mappingColumn} />
```
jattasNI, line 33: I noticed a small bug in the API docs for `validity`.

`unsupportedMappingType: true when the column contains a mapping element other than nimble-mapping-text`

should be something like this

`unsupportedMappingType: true when the column contains a mapping element other than icon, spinner, text, or empty`

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -25,35 +33,51 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
...
+        for that value's group row.
+-   Avoid having multiple values that map to the same text because grouping and
+    sorting is done on the record value rather than the mapped text.
+
+#### Choosing a mapping style
 
--   Provide a mapping for every expected record value. Because grouping is performed on the record value, non-mapped record values can result in multiple groups without group labels.
-    -   To improve grouping behavior of a value that shouldn't be rendered in table cells, use a <Tag name={mappingEmptyTag}/> element to provide text for that value's group row.
--   Avoid having multiple values that map to the same text because grouping and sorting is done on the record value rather than the mapped text.
+| Best practice                                                                                                                                                                                  | Example                                      |
```
jattasNI, line 71: Should we include the name of the API to configure this state in the guidance somewhere (`text-hidden`)? It could get pretty wordy but I think it could be helpful. For example: "Icon and text" -> "Icon and text (`text-hidden=false`)" and "Icon-only" -> "Icon-only (`text-hidden=true`)".

An alternate idea would be new column(s) that display the mapping configuration (and, if you think it's useful, the column header configuration). But that might use a lot of space.

jattasNI (approved):
It would be good to get @mollykreis to review this too since she's thought through our mapping corner cases a lot.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -25,35 +33,51 @@ The <Tag name={tableColumnMappingTag}/> renders specific number, boolean, or str
...
+| ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
+| **Icon and text, with a text header** <br/> Default configuration where icons exist for the mapping set                                                                                        | <img src={iconTextTextHeader} />             |
```
mollykreis, line 73: nit: Can we call this the _preferred_ configuration rather than the _default_ configuration? I think the word "default" is a bit confusing in context of a page with API docs where there is a concept of a default behavior of the component.

```suggestion
| **Icon and text, with a text header** <br/> Preferred configuration where icons exist for the mapping set                                                                                        | <img src={iconTextTextHeader} />             |
```

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
mollykreis, line 77: I'm not sure how precise you care about these example snippets being, but the spacing between the icon & the text here doesn't look correct given that they are two distinct columns. I can help you get a different screenshot if you want.

mollykreis (approved):

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
fredvisser: Fixed in `packages/nimble-components/src/table-column/enum-base/tests/shared-storybook-docs.ts`

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
fredvisser: Added explicit guidance not to mix and match. The spinner is more of a special case, since it augments the primary column mapping. Added a row to the table to explain that.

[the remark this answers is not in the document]
fredvisser: The full treatment and audit of SLE can wait for the icon pattern doc, but in the meantime I added a line for spinners to the table, including a caution – "…but don't use to indicate a process is 'running' or 'waiting'". Doesn't give complete guidance/context, but should prompt a followup question

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
fredvisser: I tried to find a succinct way to add this to the table, but struggled in part because it's [not accurate](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML) to document `text-hidden=false` as equivalent to not adding the `text-hidden` attribute, correct? 

I'd argue that the attributes are named such that everyone will figure out the correct config without the docs spelling it out.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
fredvisser: Added explicit guidance not to mix and match except for spinners.

fredvisser: I created them in Figma, but you're right – it wasn't the correct spacing. I've updated all snippets to be better aligned, and approximately fixed the column spacing.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -28,40 +34,57 @@ string values as an icon, a spinner, text, or an icon/spinner with text in the
...
+### Best practices
+
+- Provide a mapping for every expected record value. Because grouping is
+  performed on the record value, non-mapped record values can result in
+  multiple groups without group labels.
+  - To improve grouping behavior of a value that shouldn't be rendered in
+    table cells, use a <Tag name={mappingEmptyTag}/> element to provide text
+    for that value's group row.
+- Avoid having multiple values that map to the same text because grouping and
+  sorting is done on the record value rather than the mapped text.
```
rajsite, line 68: Avoid having multiple values that map to the same text because the text is used for grouping and would result in multiple groups with the same text label.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -28,40 +34,57 @@ string values as an icon, a spinner, text, or an icon/spinner with text in the
...
+- Don't mix and match icon-only, and icon and text mappings in the same
+  column. Choose whether to enable the `text-hidden` attribute or not, and
+  stick with it.
```
rajsite, line 71: Choose whether to enable the `text-hidden` attribute for all mappings or none.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -28,40 +34,57 @@ string values as an icon, a spinner, text, or an icon/spinner with text in the
...
+| **Icon-only, with an icon header** <br/> Use only when table width is at a premium and all the icons in the set have a clear and unambiguous meaning                                           | <img src={iconOnlyIconHeader} />             |
+| **Icon-only, with a text header** <br/> Use when header text provides enough context for all the icons in the set to have a clear and unambiguous meaning                                      | <img src={iconOnlyTextHeader} />             |
+| **Text-only, with a text header** <br/> Use when there isn't an icon for each mapping value                                                                                                    | <img src={textOnlyTextHeader} />             |
+| **Icon-only column beside a text-only column** <br/> Use to be compatible with existing columns, or when the end-user is likely to remove the text-only column in a custom table configuration | <img src={iconOnlyPlusTextOnlyTextHeader} /> |
```
rajsite, line 83: > Use to be compatible with existing columns

That isn't very clear guidance. Is this no longer recommended or are there clear reasons to use this mode?
The way this is written is it just sounds legacy. If so we should say so and the alternative.

"Separate columns for icon and text is a legacy pattern. Ideally applications should use the **Icon and text, with a text header** pattern or another alternative above."

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -28,40 +34,57 @@ string values as an icon, a spinner, text, or an icon/spinner with text in the
...
+| **Spinner** <br/> Use with any other mapping style to indicate indeterminate progress, but don't use to indicate a process is 'running' or 'waiting'                                           | <NimbleSpinner/>                             |
```
rajsite, line 84: Aren't running or waiting indeterminate progress tasks? Why are those exceptions and what should I do in those cases? Have any examples of valid indeterminate progress tasks for using a spinner?

rajsite (approved):

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
fredvisser: That's a good point - there will likely be a few app-specific exceptions (like _running_) that are indeterminate, but will continue to use a 'running' icon, but it's not worth naming and explaining that here.

Issue #2034, opened 2024-04-24
Wafer Map High Performance Redesign


rajsite:
# 🙋 Feature Request

## 😯 Problem to Solve

The Wafer Map has new rendering target goals for drawing a large number of wafers.

## 💁 Proposed Solution

See the [Rendering++](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/wafer-map/specs/features/rendering.md) spec.

## 📋 Tasks

- [ ] Render Dies
   will have async await  for promises and error handling of rejection
- [ ] Refactor Model and state management
   will split state on responsibility and sync it between wafer and worker 
- [ ] Render text and colors (PR is drafted)
   will bring parity with the old renderer
- [ ] Remove legacy code
   will leave only the new update cycle inside the wafer map component
- [ ] Create error handlig
   will cover thread saturation and strict browser settings
- [ ] Sync rendering to animation frames
   will make updates smooth and will allow for rendering interuptions for large wafers

PR #2154, opened 2024-06-03
Created Wafer Map worker colors and text renderer

munteannatan:
# Pull Request

## 🤨 Rationale

We needed to display the colors from the color scale and the values as text on the offscreen wafer image.

## 👩‍💻 Implementation

Created a color map and stored indexes in place of values.
created separate draw text method and called it based on user settings and die size

## 🧪 Testing

Created new renderer test for hidden labels
Duplicated existing matrix UI tests
Small bug with grid dimensions, will investigate it

## ✅ Checklist


packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -66,6 +73,20 @@ export class MatrixRenderer {
         );
     }
 
+    private calculateColorIndex(value: number): number {
```
atmgrifter00, line 76: This isn't "calculating" so much as "finding" right?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -99,8 +120,20 @@ export class MatrixRenderer {
...
+        this.values = valuesBuffer;
+        this.colorIndices = new Int32Array(this.values.length);
+        for (let i = 0; i < this.values.length; i++) {
+            this.colorIndices[i] = this.calculateColorIndex(this.values[i]!);
+        }
+    }
+
     public setRenderConfig(renderConfig: RenderConfig): void {
         this.renderConfig = renderConfig;
+        this.colors = renderConfig.colorScale.map(category => category.color);
```
atmgrifter00, line 133: Why use the term "category" instead of "marker"?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -13,8 +13,13 @@ export class MatrixRenderer {
     public scaledColumnIndices = Float64Array.from([]);
     public scaledRowIndices = Float64Array.from([]);
     public columnIndicesPositions = Int32Array.from([]);
+    public colorIndices = Int32Array.from([]);
     public canvas!: OffscreenCanvas;
     public context!: OffscreenCanvasRenderingContext2D;
+    private colors: string[] = [];
+    private colorValues = Float64Array.from([]);
+    private readonly outsideRangeDieColor = 'rgba(218,223,236,1)';
```
atmgrifter00, line 21: Where does this color come from? Should we have a token for it?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -99,8 +120,20 @@ export class MatrixRenderer {
         }
     }
 
+    public setValues(valuesBuffer: Float64Array): void {
```
atmgrifter00, line 123: Beyond the suggested new matrix tests, the only place I could see a reasonable unit test for is the logic for when `drawText` is executed within `drawWafer` in the `WorkerRenderer`. I'm not sure how easy that is to set up, and if it's too expensive/laborious, then I think the matrix tests will suffice.

atmgrifter00 (approved):

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
munteannatan: This is the color used in the previous rendering for an empty die. It's light grey. Do tokens work in workers?

packages/nimble-components/build/generate-workers/index.ts
```diff
@@ -24,25 +24,34 @@ function writeFile(filePath: string, content: string): void {
...
+const matrixRendererTypesPath: string = resolveModulePath(matrixRendererTypes);
+const matrixRendererTypesCode: string = '/* eslint-disable @typescript-eslint/prefer-readonly */\n'
+    + `/* eslint-disable @typescript-eslint/explicit-member-accessibility */\n${fs.readFileSync(
+        matrixRendererTypesPath,
+        'utf-8'
+    )}`;
 
 const fileContent: string = `// eslint-disable-next-line no-template-curly-in-string
 export const workerCode = ${JSON.stringify(sourceCode)};
 `;
```
rajsite, line 45: Instead of putting the eslint-disable lines in the generated code can you add src/wafer-map/workers/ to the ignore list: 
https://github.com/ni/nimble/blob/5778a15ee6098377ed5044f014c9b18151313c57/packages/nimble-components/src/.eslintrc.js#L8

packages/storybook/src/nimble/wafer-map/wafer-map-matrix.stories.ts
```diff
@@ -1,8 +1,8 @@
 import type { StoryFn, Meta } from '@storybook/html';
 import { html, ViewTemplate } from '@microsoft/fast-element';
+import { tableFromArrays } from 'apache-arrow';
```
jattasNI, line 3: It's great that we're getting really close to parity between the experimental wafer map and the original one. I visually compared the stories in [the storybook build](https://60e89457a987cf003efc0a5b-vfitydsqhq.chromatic.com/?path=/story/tests-wafer-map--experimental-wafer-map-grid-dimensions-test) and noticed a few differences:
1. In "Wafer map dies and orientation test" there are slightly different colors on the die labeled "15" (might need to use an image sampling tool to see the difference)
2. "Wafer map grid dimensions test" there are some visual artifacts in the experimental version (random rectangles rendered outside the bounds of the wafer)
3. "Wafer map highlighted test" exists only for the original renderer, not the experimental one. Is highlighting still a feature gap?

## Issue 1
**Original**
image

**Experimental**
image

## Issue 2

**Original**
image

**Experimental**
image

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -116,6 +118,61 @@ describe('WaferMap', () => {
         });
     });
 
+    describe('worker renderer draw wafer action', () => {
```
jattasNI, line 121: Are we close to parity with unit test coverage for the original and experimental wafer maps now?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
rajsite: No we can't leverage tokens from workers. Ideally we would want all the style related things (not just this color but the default color scale, the font face, etc) to all come from tokens. The tokens should be subscribed to from the element and observed for changes. The changes should then be passed to the worker, maybe as part of renderConfig, etc.

I wouldn't make it a requirement for this PR to avoid growing scope. but we should address both pieces in follow-ups:
1. Don't hard code style in the worker and instead observe tokens on the wafer and pass to the worker (higher priority architecture pattern)
2. Use / create tokens that are theme-aware for colors, etc. (lower priority theme integration / awareness)

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
```diff
@@ -116,6 +118,61 @@ describe('WaferMap', () => {
...
+            element.dieLabelsHidden = true;
+            element.canvasWidth = 500;
+            element.canvasHeight = 500;
+            element.diesTable = tableFromArrays({
+                colIndex: Int32Array.from([]),
+                rowIndex: Int32Array.from([]),
+                value: Float64Array.from([])
+            });
+            await waitForUpdatesAsync();
+            await waitForUpdatesAsync();
```
rajsite, line 151: These double waitForUpdatesAsync have me a bit concerened. What is the timing that is happening here? Can you explain it?

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
munteannatan: I updated https://github.com/ni/nimble/issues/2034 with the mentioned feedback

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
munteannatan: This happens because I create a lot of mock async methods. the latest one which is draw text does not get called in the specified timeout

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
munteannatan: I think most of the code is covered. There are some drastic changes to the structure of the wafer and a lot of computation is moved in the worker or removed altogether, so I do not think there can be parity. But we repurposed some of the tests and created new tests for the worker. What is left should be covered by the image tests

packages/storybook/src/nimble/wafer-map/wafer-map-matrix.stories.ts
munteannatan: Issue 1 is expected to happen because we are taking the approach to slice the color interpolation for the linear color scale to increase performance.
Issue 2 was caused by the fac that the new vertical and horizontal scales did not have domain borders for which to filter out dies outside the bounding box. changed the approach to filter out dies before drawing

rajsite: > "Wafer map highlighted test" exists only for the original renderer, not the experimental one. Is highlighting still a feature gap?

@munteannatan what's the direction we are thinking for highlighting? If that is work we still expect can you track it in https://github.com/ni/nimble/issues/2034 ?

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
rajsite: So to give an example, for normal component work when we update an attribute, etc, that work is done asynchronously in an update queue and `waitForUpdatesAsync` is a way to wait for that queue to empty. It's not really being used as a timing function, it's knowing that specific work was added to the queue and we know it will complete. We "wait for the [DOM] updates queue to be emptied". That makes tests reliable, they are based on some specific thing we know is happening, not based on timing (ideally).

What I want to know is if this is reliable or if this is timing based and up to the worker running quickly or slowly (worker or whatever happens to take the time of 2 waitForAsync). If this is timing based then we should instead make a helper, etc, that can reliably wait for something, i.e. like a `waitForExperimentalUpdate()` helper that can `await waferMap.currentTask` or something.

We want to wait for a specific thing, one cycle of the task queue, one cycle of the wafer task, etc. instead of just 16ms or 16ms * 2 which may result in intermittent tests, etc.

rajsite (commented):
Implementation looks good, just have concerns about test pattern and making sure the tests run reliably.

packages/storybook/src/nimble/wafer-map/wafer-map-matrix.stories.ts
munteannatan: we hope to do the highlighting using arquero, but yes, we have to track it

packages/nimble-components/src/wafer-map/tests/wafer-map.spec.ts
munteannatan: I changed the second await for the current task and this seems to work. the two waitForUpdatesAsync were working because the first one was setting the current task and the second one was awaiting it when finding it was not undefined

jattasNI (approved):

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -187,5 +232,59 @@ export class MatrixRenderer {
...
+                const scaledY = this.scaledRowIndices[columnStartIndex]!;
+                if (
+                    !(
+                        scaledY >= this.transformConfig.topLeftCanvasCorner.y
+                        && scaledY < this.transformConfig.bottomRightCanvasCorner.y
+                    )
+                ) {
+                    continue;
+                }
+                let label = `${this.values[columnStartIndex] || 'NaN'}${this.renderConfig.dieLabelsSuffix}`;
```
rajsite, line 274: @munteannatan after enabling type checked linting it found a case here where theoretically `this.values[columnStartIndex]` could be `undefined`. I added the `|| 'NaN'` condition but not completely sure if that's how you want it to be handled.

rajsite (approved):

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
rajsite: @munteannatan I'm going to merge this PR in order to do some follow-up changes, but please consider what's the best thing to do here and handle in a follow-up, etc.

munteannatan: Thank you! I will try and check the behavior and make adjustments in a follow up.

PR #2155, opened 2024-06-03
Allow very long words in banner to break to another line

### LINKED ISSUE #2153: Long words can cause cut-off text in banner
# 🐛 Bug Report

Especially if horizontal space is limited, multi-line banner messages with long words can cause some text and icons to get cut off.

## 💻 Repro or Code Sample

Example:
image

Give any Nimble Banner a message with multiple words, one of which is long enough to not fit within the width the banner has to work with.

Note that our banner also has two buttons, which are not shown because of the long word:
image

## 🤔 Expected Behavior

The banner picks one or more places in the long word to split onto the next line. Other banner content to the right of the text, such as buttons, is visible.

## 😯 Current Behavior

See above screenshot.

## 💁 Possible Solution

Have the UI split onto a new line at appropriate places in long words.

## 🔦 Context

These error messages are coming from TestStand, so our simple sequencer doesn't have a good way to specify or update the text being displayed. It seems like the fix needs to come from the UI control.

## 🌍 Your Environment

* OS & Device: Windows PC
* Browser: WPF WebView2 control

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2153

## 👩‍💻 Implementation

- Set `overflow-wrap: anywhere` on the banner. This causes long words to first be wrapped to a new line, but if they still overflow, then the word is broken at the point it would overflow.
- Add a very long word to the banner's Chromatic tests
- Removed `white-space: nowrap` from the banner title styling because it was causing some/all of the action button/link and close control to be out-of-view when both the title and the button/link text were long. See Chromatic diff for details.

## 🧪 Testing

Storybook

## ✅ Checklist


m-akinc:
@atmgrifter00 would you mind buddying?

atmgrifter00 (approved):

packages/nimble-components/src/banner/styles.ts
```diff
@@ -33,6 +33,7 @@ export const styles = css`
         font-size: 12.8px;
         align-items: top;
         overflow: hidden;
+        word-break: break-word;
```
rajsite, line 36: mdn says `word-break: break-word` is deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
Looks like `overflow-wrap` should be configured: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#break-word

m-akinc: I think I had meant to use word-wrap rather than word-break, though I believe it's essentially equivalent to word-break/overflow-wrap. Confusing! I'll double-check.

m-akinc: Actually, this is the way I intended. `word-wrap` is deprecated, not `word-break`. From what I can tell, `word-break: break-word` and `overflow-wrap: break-word` are essentially equivalent. Maybe `overflow-wrap` is more common/recognized, though, so I can use that instead.

rajsite: > word-wrap is deprecated, not word-break.

The setting used in the PR, `word-break: break-word`, is what the MDN docs I linked is saying is deprecated:

image

Are the MDN docs incorrect? Any source for why that is not expected to be deprecated?

m-akinc: Oh, I didn't see that the `break-word` value was deprecated. It didn't say so in the [doc for that value](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#break-word). I searched a little for other confirmation that the `break-word` value is deprecated, and [caniuse](https://caniuse.com/mdn-css_properties_word-break_break-word) seems to back that up. Anyway, I had already switched to using `overflow-wrap: break-word`, so we should be good.

rajsite: The comment was resolved and I was reset but I don't see a change?

m-akinc: 🤦‍♂️Duh, I hadn't pushed the commit. Sorry.

jattasNI: Reminder to update the PR description too :-)

change/@ni-nimble-components-78757500-8f52-4913-b98c-dfc88b407979.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: >it was causing some/all of the action button/link and close control to be out-of-view when both the title and the button/link text were long

In [this chromatic diff](https://www.chromatic.com/pullrequest?appId=60e89457a987cf003efc0a5b&number=2155&activeElementId=comparison-test-key-63e2b3e0eb6fb7cdb2d44ef5-1200px&view=changes) it looks to me like the baseline is correct but the new build is not (buttons are clipping on their right edge). Am I looking at the wrong diff or misreading Chromatic or something?

rajsite (approved):

change/@ni-nimble-components-78757500-8f52-4913-b98c-dfc88b407979.json
m-akinc: Looks like the difference between `break-word` and `anywhere` is relevant here (i.e. for `anywhere`, "Soft wrap opportunities introduced by the word break are considered when calculating min-content intrinsic sizes"). Switched (and updated the PR description).

jattasNI (approved):

PR #2157, opened 2024-06-04
Select storybook update

atmgrifter00:
# Pull Request

## 🤨 Rationale

Just updating the `Select` storybook to conform to new standards.

## 👩‍💻 Implementation

Added new `ListOption` and `ListOptionGroup` storybooks that should be accessible from other component docs as needed.

## 🧪 Testing

N/A

## ✅ Checklist


packages/storybook/src/nimble/list-option/list-option.stories.ts
```diff
@@ -0,0 +1,60 @@
...
+} from '../../utilities/storybook';
+import { hiddenWrapper } from '../../utilities/hidden';
+
+const metadata: Meta = {
+    title: 'Internal/ListOption',
+    parameters: {
+        docs: {
+            description: {
+                component:
+                    'The `nimble-list-option` component is used to define the options that are available to the `nimble-select` component.'
```
jattasNI, line 15: At some point we should embed this story in the combobox mdx too. Your call between these options:
1. do this later and leave this as-is, hope we remember later to update its contents to be agnostic to what parent it's in
2. do this later but update it now to be agnostic about what parent it's hosted in
3. do this now

Regardless of which option you pick, we should use constants for the tag names.

jattasNI (approved):

packages/storybook/src/nimble/list-option/list-option.stories.ts
atmgrifter00: Made an attempt at making it agnostic.

packages/storybook/src/utilities/storybook.ts
```diff
@@ -196,7 +197,7 @@ export const errorVisibleDescription = 'When set to `true`, the `error-text` mes
 export const dropdownPositionDescription = (options: {
     componentName: string
 }): string => `Controls the position of the dropdown relative to the ${options.componentName}.`;
-export const optionsDescription = `The \`${listOptionTag}\` items for the user to select from.`;
+export const optionsDescription = (options: { forSelect: boolean }): string => `The \`${listOptionTag}\` items for the user to select from.${options.forSelect ? ` Each ${listOptionTag} can also be grouped using the \`${listOptionGroupTag}\` element.` : ''} `;
```
rajsite, line 200: Based on the text, maybe the option name should be `includeGrouping` since it's about grouping docs and not about select docs

rajsite (approved):

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -28,20 +28,22 @@ interface SelectArgs {
     optionsType: ExampleOptionsType;
```
jattasNI, line 28: I just noticed we don't actually document the `value` property on the select. I think we should, since we mention it in the initial summary and in the docs for the options. You can use the category `nonAttributeProperties` (not sure if that's the exact name) to indicate it's a property, not an attribute.

I still have a task in #824 to document a recommendation to use form association rather than events/value, but we should still document the existence of `value`.

jattasNI (approved):

packages/storybook/src/nimble/select/select.stories.ts
jattasNI: When you do this, you could consider removing the text about `value` from the summary at the top; it's probably more detail than is needed for someone who is just trying to understand what the control does.

PR #2159, opened 2024-06-04
Angular reduce test verbosity

rajsite:
# Pull Request

## 🤨 Rationale

With a single validate command running all validation the verbosity of the angular karma logs makes it difficult to scan logs and identify test failures vs lint failures etc.

## 👩‍💻 Implementation

Switched angular to use `karma-spec-reporter` with low verbosity settings like nimble-components has been.

## 🧪 Testing

Ran locally and rely on CI.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-angular-51664bbb-8d98-4796-991a-23f64c154a16.json
+++ b/change/@ni-nimble-angular-51664bbb-8d98-4796-991a-23f64c154a16.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Switch to low verbosity spec reporter",
+  "packageName": "@ni/nimble-angular",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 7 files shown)


rajsite:
It helps a lot and I think is uncontroversial so gonna override merge, but let me know if there are concerns @atmgrifter00 @jattasNI to address in follow-up

jattasNI:
I've spent a lot of time scrolling through test logs recently and haven't been loving this experience, so this seems like a good step (assuming it still gives a good summary when a test fails).

I've thought about pulling in a test reporter like [this one](https://github.com/marketplace/actions/test-reporter) but they require a bit of GitHub Action tweaking so I haven't gotten around to it; maybe next innovation day.

rajsite:
> I've spent a lot of time scrolling through test logs recently and haven't been loving this experience, so this seems like a good step (assuming it still gives a good summary when a test fails).
> 
> I've thought about pulling in a test reporter like [this one](https://github.com/marketplace/actions/test-reporter) but they require a bit of GitHub Action tweaking so I haven't gotten around to it; maybe next innovation day.

I don't really like coupling to GitHub Actions from random individuals and GitHub doesn't have a native test reporting / tracking feature. That just prints a markdown summary table and wouldn't get test failure / execution time history which I really would like. 

I think it would be neat to push our test results to SLE to be able to trace history or individual test steps etc. Another innovation day idea :D

Issue #1952, opened 2024-03-20
Support Select placeholder in Blazor


atmgrifter00:
# 🧹 Tech Debt

Supporting the placeholder feature in Blazor has implementation challenges related to how we need to not only expose the `Selected` property for users to set on a `ListOption`, but we also need to be able to handle changing the `Selected` state based on the value changing on the `Select`.

Currently, the `NimbleListOption` Razor component [supplies a `Selected` property](https://github.com/ni/nimble/blob/699749620c558e3119e22889a4c6f800a1dd4e51/packages/nimble-blazor/NimbleBlazor/Components/NimbleListOption.razor.cs#L24), but it's [not even being leveraged in the template](https://github.com/ni/nimble/blob/699749620c558e3119e22889a4c6f800a1dd4e51/packages/nimble-blazor/NimbleBlazor/Components/NimbleListOption.razor#L5), meaning that the `Selected` property currently on `ListOption` is basically broken/useless.

The current version of Microsoft's Fluent UI Blazor library seems to have [addressed these problems](https://github.com/ni/nimble/blob/699749620c558e3119e22889a4c6f800a1dd4e51/packages/nimble-blazor/NimbleBlazor/Components/NimbleListOption.razor#L5) (albeit in a seemingly goofy way). Their `FluenOption` now [binds the `selected` attribute](https://github.com/microsoft/fluentui-blazor/blob/86523213ac86c7772507074939cc4d3b900fdad1/src/Core/Components/List/FluentOption.razor#L10) of the web component to the `Selected` property on the Razor component as expected, and then also manages toggling that state via a handler that gets called not only when the option is clicked, but also [when the keydown handler runs](https://github.com/microsoft/fluentui-blazor/blob/86523213ac86c7772507074939cc4d3b900fdad1/src/Core/Components/List/ListComponentBase.razor.cs#L544) for the owner component.

jattasNI:
Removed tech debt label because this has client-facing impact so it's hard to prioritize against the backlog of internal dev-facing work.

PR #2161, opened 2024-06-05
ListOptionGroup integration to Angular/Blazor

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #791

## 👩‍💻 Implementation

Standard integration implementations for Angular/Blazor.

## 🧪 Testing

Standard unit tests for Angular/Blazor. Updated both example apps.

## ✅ Checklist

### DIFF
--- a/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
+++ b/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
@@ -152,21 +152,36 @@
             <div class="container-label">Select</div>
             <nimble-select filter-mode="standard" appearance="underline">
                 <nimble-list-option hidden selected disabled>Select an option</nimble-list-option>
-                <nimble-list-option>Option 1</nimble-list-option>
-                <nimble-list-option>Option 2</nimble-list-option>
-                <nimble-list-option>Option 3</nimble-list-option>
+                <nimble-list-option-group label="Group 1">
+                    <nimble-list-option>Option 1</nimble-list-option>
+                    <nimble-list-option>Option 2</nimble-list-option>
+                </nimble-list-option-group>
+                <nimble-list-option-group label="Group 2">
+                    <nimble-list-option>Option 3</nimble-list-option>
+                    <nimble-list-option>Option 4</nimble-list-option>
+                </nimble-list-option-group>
             </nimble-select>
             <nimble-select appearance="outline">
                 <nimble-list-option hidden selected disabled>Select an option</nimble-list-option>
-                <nimble-list-option>Option 1</nimble-list-option>
-                <nimble-list-option>Option 2</nimble-list-option>
-                <nimble-list-option>Option 3</nimble-list-option>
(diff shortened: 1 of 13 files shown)


packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -171,19 +171,35 @@
...
+                <NimbleListOptionGroup Label="Group 1">
+                    <NimbleListOption>Option 1</NimbleListOption>
+                    <NimbleListOption>Option 2</NimbleListOption>
+                </NimbleListOptionGroup>
+                <NimbleListOptionGroup Label="Group 2">
+                    <NimbleListOption>Option 3</NimbleListOption>
+                    <NimbleListOption>Option 4</NimbleListOption>
+                </NimbleListOptionGroup>
             </NimbleSelect>
+            <NimbleListOptionGroup></NimbleListOptionGroup>
```
rajsite, line 183: Can you make sure this aligns with Blazor? Two things:
- Don't think we need this empty listoptiongroup that is outside of any select
- Blazor does not seem to show placeholders, can't remember, are we able to use placeholders in blazor?

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
```diff
@@ -152,21 +152,36 @@
             <div class="container-label">Select</div>
             <nimble-select filter-mode="standard" appearance="underline">
                 <nimble-list-option hidden selected disabled>Select an option</nimble-list-option>
```
rajsite, line 154: general question (don't need to change anything in example app), can your placeholder be inside a group? I'd imagine so but didn't look at test coverage.

rajsite (approved):

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
atmgrifter00: That first one was just an "oopsy". For the second, we haven't enabled this because of  [this issue](https://github.com/ni/nimble/issues/1952).

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
atmgrifter00: Yeah that should work, but no, I haven't added explicit tests around this as this wouldn't follow our guidance (we should update our guidance docs for this).

jattasNI (approved):

msmithNI (approved):

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
rajsite: From offline discussions sounds like we will add some tests to validate behavior so that does not block this PR.

Issue #327, opened 2022-02-02
Support multi-color icons 🌈


jattasNI:
## 😯 Problem to Solve

The design team has started providing icons with colored accents but our current icon strategy [relies on setting the SVG `fill` to a single color](https://github.com/ni/nimble/blob/494f1e68dfa32945485b915fcf9b2f9124cc2742/packages/nimble-components/src/icon-base/styles.ts#L41). We'll need a way to specify multiple colors for an icon.

We also need to figure out whether the accent colors might need to change in response to themes or icon states like error and warning. [Original thread](https://teams.microsoft.com/l/message/19:wo8vmMKMsHfltKXxc0bczZo-X4JlQSV5VxpaRJdh13k1@thread.tacv2/1643835972184?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=9ee065d7-3898-4245-82f6-76e86084b8b1&parentMessageId=1643835972184&teamName=Design%20System&channelName=General&createdTime=1643835972184).

## 💁 Proposed Solution

I think @rajsite had ideas about how to leverage SVG features to make this happen, but I don't recall what they were.

Some expected colors:

![image](https://github.com/user-attachments/assets/a60df82a-a854-4b38-a10e-8108e4f17ae5)

jattasNI:
We added our first multicolor icon for #2046 but for now it uses fixed colors and doesn't respect theme or status. [Brandon suggested](https://github.com/ni/nimble/issues/2046#issuecomment-2159215000) that we'll probably want at least different colors for each theme, although the priority of doing that is still TBD.

PR #2165, opened 2024-06-05
Add icons for AI, clean, expand, collapse, debug

### LINKED ISSUE #2046: New icons for table expand – visual design request
# 🎨 Visual Design Request

1. The icons used in the 'table-fullscreen' and 'table-not-fullscreen' buttons are on the top-right of the screen, so the arrows should point up and to the right. Our current icons would work if the buttons were on the top-left, but alas. Can we make new icons with each turned 90º?

Screenshot 2024-04-29 at 10 12 24 AM
Screenshot 2024-04-29 at 10 13 05 AM

### LINKED ISSUE #2138: Broom icon
# 🎨 Visual Design Request

Can we add a broom icon?

## 📌 Request Background

We need a icon that suggest "clean".

## 🎯 Core Requirements

To look like a broom.

## 🍆 Non-requirements

It should be made out of wood not plastic.

jattasNI:
# Pull Request

## 🤨 Rationale

New icons are available. One of them is multicolored and we'd like to retain those colors.

Resolves #2138 
Resolves #2046

## 👩‍💻 Implementation

Generally followed CONTRIBUTING instructions for adding icons.

1. Naming:
   - the expand/collapse icons use Brandon's suggested names which are very free of metaphors
   - the clean icon uses "sparkles" to [match FontAwesome](https://fontawesome.com/search?q=sparkles&o=r)
   - the AI icon uses "sparkle-swirls" since we generally want to avoid metaphors but I didn't see a FontAwesome equivalent
   - the debug icon uses "debug" since descriptive names seemed silly ("shield-with-legs"?) and the metaphor seems unique
   - (open to feedback on all the above)
1. Synonyms: used the ones Brandon suggested
2. I stripped color from all icons except the AI icon. This doesn't match our CONTRIBUTING guidance and caused a test to fail. I'm proposing an update to the guidance and removing those asserts from the test. 
    - I considered establishing a convention where `<defs>` were allowed in svgs only if you explicitly tag them with a [custom attribute](https://stackoverflow.com/questions/15532371/do-svg-docs-support-custom-data-attributes) that proves you left them in intentionally. But I lean towards just relying on code review to enforce best practices.
    - Note: #327 would be another opportunity for us to improve this if we want themeable multicolored icons.
4. Minor update to use the new `npm run storybook` command in a CONTRIBUTING doc

## 🧪 Testing

Local storybook inspection

## ✅ Checklist


jattasNI:
I'll be OOO most of tomorrow. Someone is welcome to push this through if the builds pass (I know one client was hoping for the icons to be available this week).

packages/nimble-components/src/icon-base/tests/icons.spec.ts
```diff
@@ -16,20 +16,14 @@ describe('Icons', () => {
...
                 const svg = getSVGElement(value.data);
-                const paths = getPaths(svg);
                 expect(svg).toBeTruthy();
                 expect(svg.getAttribute('viewBox')).toBeTruthy();
                 expect(svg.getAttribute('height')).toBeNull();
                 expect(svg.getAttribute('width')).toBeNull();
-                expect(svg.querySelector('defs')).toBeNull();
-                for (const path of paths) {
-                    expect(path.getAttribute('style')).toBeNull();
-                }
```
rajsite, line 32: Do we think these are not generally useful? This is a parameterized spec, we could choose specific exceptions instead of disabling for all icons. i.e. one test with the more thorough tests and one with a smaller subset.

rajsite: pushed a change that preserves the style exclusion tests for other icons

rajsite (approved):

Issue #1862, opened 2024-02-23
Chromatic Menu matrix test has unexpected Firefox diff


m-akinc:
## 🧹 Tech Debt
After the recent Chromatic Capture Cloud update, the Menu matrix test has been showing unexpected gaps between some menus and their submenus:
https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=65d7e05b8ca63d78ed7c6906
![image](https://github.com/ni/nimble/assets/7282195/764b70f1-66f9-4d2d-8822-6f416d316d5c)
I cannot reproduce this in my local Firefox (version 123, as opposed to Chromatic's version 121).
Notably, the menus at the top of each theme section do not have the problem. This feels vaguely similar to the [issues we had with the Tooltip matrix story](https://github.com/ni/nimble/issues/1106).

jattasNI:
I saw this again in [this build](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=65fc5089025309d8632dc556). It does not reproduce for me in macOS Firefox in [the corresponding Storybook](https://60e89457a987cf003efc0a5b-bsxqygieoy.chromatic.com/?path=/story/tests-menu--menu-theme-matrix).

jattasNI:
[Another instance](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=66048fc2ade42a9f39e375db)

jattasNI:
The steps to work around this diff are captured [in this comment](https://github.com/ni/nimble/issues/1479#issuecomment-2341882851)

rajsite:
Haven't seen this in a while, closing

PR #2167, opened 2024-06-06
Select manual filtering.

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1866

## 👩‍💻 Implementation

This PR surfaces the necessary APIs for clients to be able to provide their own manual filtering of options in the `Select`. Additionally, clients can now set a `loading-visible` attribute on the `Select` to `true` to show a visual in the dropdown to indicate that it is in a loading state.

Not in this PR:
- Examples that actually exercise the expected client workflow in which they would apply their own custom filtering for options. Instead, there are simply unit tests that validate the necessary notifications for clients to use (in addition to tests regarding the loading visual in the dropdown). The Angular and/or Blazor integration will provide a more complete example of this feature.

## 🧪 Testing

Unit tests. Chromatic tests.

## ✅ Checklist


packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -833,412 +845,468 @@ describe('Select', () => {
...
         let pageObject: SelectPageObject;
 
-        beforeEach(async () => {
-            ({ element, connect, disconnect } = await setup());
-            element.style.width = '200px';
-            element.filterMode = FilterMode.standard;
-            await connect();
-            pageObject = new SelectPageObject(element);
-        });
+        describe('with filterMode = standard', () => {
```
atmgrifter00, line 848: All changes from here until line 1271 are indentation. Differ is a little confused in places.

atmgrifter00:
@m-akinc you mind buddying this?

packages/nimble-components/src/select/template.ts
```diff
@@ -151,11 +153,19 @@ SelectOptions
...
                         ></slot>
                     </div>
-                    ${when(x => (x.filterMode !== FilterMode.none && x.filteredOptions.length === 0), html<Select>`
+                    ${when(x => (x.filterMode !== FilterMode.none && x.filteredOptions.length === 0 && !x.loadingVisible), html<Select>`
                         <span class="no-results-label ${x => x.positionAttribute}">
                             ${x => filterNoResultsLabel.getValueFor(x)}
                         </span>
                     `)}
+                    ${when(x => x.loadingVisible, html<Select>`
+                        <div class="loading-container" @click="${(x, c) => x.ignoreClickHandler(c.event as MouseEvent)}">
```
m-akinc, line 162: Do we need `ignoreClickHandler` here?

packages/nimble-components/src/select/styles.ts
```diff
@@ -151,6 +151,28 @@ export const styles = css`
...
+
+    .loading-container.above {
+        margin-top: ${mediumPadding};
+        margin-bottom: 0;
+    }
+
+    .loading-container.below {
+        margin-top: 0;
+        margin-bottom: ${mediumPadding};
+    }
```
m-akinc, line 171: I don't see the loading container ever getting a class of `above` or `below`.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -831,413 +843,462 @@ describe('Select', () => {
...
+                await pageObject.openAndSetFilterText('o');
+                const filteredOptions = pageObject.getFilteredOptions();
+                expect(filteredOptions.length).toBe(6);
+            });
+
+            it('emits filter-input event when filter text is entered', async () => {
+                const filterInputEvent = jasmine.createSpy();
+                element.addEventListener('filter-input', filterInputEvent);
+                await pageObject.openAndSetFilterText('o');
+                expect(filterInputEvent).toHaveBeenCalledTimes(1);
```
m-akinc, line 1286: Validate event `filterText` property?

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -831,413 +843,462 @@ describe('Select', () => {
...
+            });
 
-            expect(option.visuallyHidden).toBeFalse();
+            it('emits filter-input event with empty filterText when dropdown is closed', async () => {
```
m-akinc, line 1289: Add test case that event is emitted when opening dropdown?

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -831,413 +843,462 @@ describe('Select', () => {
...
-            expect(option.visuallyHidden).toBeTrue();
+                expect(option.visuallyHidden).toBeFalse();
+            });
         });
 
-        it('option that is not hidden but visuallyHidden, if removed from DOM, and then re-added to DOM while dropdown closed, is visible in dropdown when opened', async () => {
-            const option = element.options[1] as ListOption;
-            await pageObject.openAndSetFilterText('Three');
-            expect(option.visuallyHidden).toBeTrue();
+        describe('with filterMode = manual', () => {
```
m-akinc, line 1263: Is there anything interesting to test about the `filterMode` being changed?

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1878,6 +1939,20 @@ describe('Select', () => {
...
             });
+
+            it('if all results are hidden, and loadingVisible is true, no results label is not shown', async () => {
+                element.loadingVisible = true;
+                await waitForUpdatesAsync();
+                await pageObject.openAndSetFilterText('abc');
+                expect(pageObject.isNoResultsLabelVisible()).toBeFalse();
+            });
+
+            it('does not automatically filter options when filterMode is manual', async () => {
```
m-akinc, line 1950: You have another test with the same name earlier in the file. Is it a duplicate?

packages/nimble-components/src/select/styles.ts
m-akinc, line 1: There's a slight spacing inconsistency when the loading indicator is shown alongside options. The loading indicator looks misaligned vertically, because it has different spacing between the text and the edge of the dropdown vs between the text and the adjacent option's text. This might be related to the margin styling I commented on earlier.
image
image

m-akinc, line 1: Looks like we need to prevent the loading indicator from shrinking vertically when there isn't enough space to show it and all the options:
image

m-akinc (approved):

packages/nimble-components/src/label-provider/core/label-token-defaults.ts
```diff
@@ -10,5 +10,6 @@ export const coreLabelDefaults: { readonly [key in TokenName]: string } = {
     popupIconWarningLabel: 'Warning',
     popupIconInformationLabel: 'Information',
     filterSearchLabel: 'Search',
-    filterNoResultsLabel: 'No items found'
+    filterNoResultsLabel: 'No items found',
+    loadingLabel: 'Loading'
```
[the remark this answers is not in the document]
atmgrifter00: @fredvisser? @jattasNI? I'm indifferent.

packages/nimble-components/src/select/template.ts
atmgrifter00: It's here so the dropdown isn't dismissed if a user clicks that visual. This avoids closing on it on misclicks.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: I don't believe so. There's no lingering state to be concerned about in either mode.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: We don't emit event on dropdown open. In the `openChanged` implementation you will see an early return if the dropdown is open.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: It's the same test, just done, additionally, when there are groups present, as that exercises a slightly different code path.

packages/nimble-components/src/select/styles.ts
atmgrifter00: Yep. This, and the fact that the styles were wrong as-is, was the reason for the vertical padding issues you mentioned.

packages/nimble-components/src/label-provider/core/label-token-defaults.ts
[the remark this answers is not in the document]
rajsite: Indifferent, other than if it is used we use the single character unicode ellipses instead of three periods.

```suggestion
    loadingLabel: 'Loading…'
```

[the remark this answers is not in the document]
jattasNI: I'd vote to include the ellipsis.

I did a very brief search for guidance in other design systems and found that [Adobe Spectrum suggests](https://spectrum.adobe.com/page/grammar-and-mechanics/#Ellipsis) to "use an ellipsis at the end of the text to communicate the progress of something that is in process, such as a waiting or loading state." So at least we won't be the only ones taking this approach.

packages/nimble-components/src/select/tests/select.spec.ts
jattasNI: No strong opinions about how it's achieved, but they should have different resolved names (the name that appends all the nested `describe` descriptions) so that you can tell which test is failing from a log file.

packages/storybook/src/nimble/select/select.mdx
```diff
@@ -52,6 +52,23 @@ the end of the prompt.
 
 <Canvas of={selectStories.placeholder} />
 
+### Dynamic Options
```
jattasNI, line 55: I've started standardizing on sentence casing for mdx headings
```suggestion
### Dynamic options
```

packages/storybook/src/nimble/select/select.mdx
```diff
@@ -52,6 +52,23 @@ the end of the prompt.
...
+listen for changes to the filter input. When accompanied with the `filterMode` set
+to `manual`, clients can perform their own filtering and provide the set of options
+to the `Select` dynamically.
+
+Typical steps for dynamic options are as follows:
+1. Set the `filterMode` to `manual` on the `Select`.
+1. Listen for the `filter-input` event on the `Select`.
+1. In the event handler, retrieve the filter text from the event and populate the DOM
+with the set of options that match the filter text by your own logic.
+    - It is important that one of the options in the DOM is the currently selected option.
```
jattasNI, line 70: 1. generally prefer attribute names
2. make docs "timeless"
3. use either `nimble-select` or just "select"

```suggestion
There are scenarios that require not all possible options to be available
initially, such as when the set of all options is too large to be loaded at once. 
In such cases, users are expected to provide filter text to narrow down the list 
of options. The select offers the `filter-input` event to allow clients to
listen for changes to the filter input. When accompanied with the setting `filter-mode`
to `manual`, clients can perform their own filtering and provide the set of options
to the select dynamically.

Typical steps for dynamic options are as follows:
1. Set the `filter-mode` to `manual` on the select.
1. Listen for the `filter-input` event on the select.
1. In the event handler, retrieve the filter text from the event and populate the DOM
with the set of options that match the filter text by your own logic.
    - It is important that one of the options in the DOM is the currently selected option.
```

packages/storybook/src/nimble/select/select.mdx
```diff
@@ -52,6 +52,23 @@ the end of the prompt.
...
+
```
jattasNI, line 71: In the HLD we have a bunch of useful guidance about how clients should do filtering (copied below). Can you summarize the relevant points and add them here? I think setting `loading-visible` is particularly important.

----

IMPORTANT: When using the manual filter mode, clients are responsible for the following:

- Either adding/removing options to/from the DOM that do or do not match the filter and/or marking existing options as hidden when they don't match.
- The currently selected option is included in the set of options placed in the DOM regardless of match against filter (and be marked as hidden if it doesn't match the filter)
- If a placeholder option exists for the select, it should never be filtered out.

Notes:

- When the loading-visible attribute is set, we will display localizable text at the bottom of the dropdown (defaulting to "Loading"), along with the spinner icon.
- The Select will not peform any debouncing as the user types into the filter input. It is expected that clients can perform any debouncing that is needed easily at the app level.
- The filter-input event will be emitted for all filter modes (except none)

Groups

When using the filterMode: manual option along with [groups](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/select/specs/option-groups-hld.md), clients will also need to determine the appropriate matching behavior against the ListOptionGroup elements. Our guidance will be that filter text should be able to be matched against any text in the group label, and when that matches, all options within that group should be visible. Conversely, the select will automatically hide groups with no items in them. This means clients only need to conditionally remove items, not groups (though it's fine if they manually remove empty groups too).

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: Yep. There is a describe block for "groups" that disambiguates them.

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -46,9 +51,13 @@ interface SelectMatrixStoryOptions {
 
 // prettier-ignore
 const component = ({
-    positionState, filterMode, placeholder, grouped, optionsOutsideGroup, slottedLabel
+    positionState, filterMode, loadingVisible, placeholder, grouped, optionsOutsideGroup, slottedLabel
 }: SelectMatrixStoryOptions): ViewTemplate => html`
-    <${selectTag} open position="${() => positionState[0]}" style="${() => positionState[1]}" filter-mode="${() => filterMode}" style="width: 250px;">
+    <${selectTag} open 
+        position="${() => positionState[0]}" 
+        style="${() => positionState[1]} width: 250px; ${isChromatic() ? '--ni-private-spinner-animation-play-state:paused' : ''}"
```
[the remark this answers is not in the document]
atmgrifter00: From what I can see it's working. Successive updates at least aren't triggering differences (that are unexpected).

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -7,6 +7,7 @@ import { FilterMode } from '../../../../nimble-components/src/select/types';
 import { createFixedThemeStory } from '../../utilities/storybook';
 import { sharedMatrixParameters } from '../../utilities/matrix';
 import { backgroundStates } from '../../utilities/states';
+import { isChromatic } from '../../utilities/isChromatic';
```
jattasNI, line 10: Reminder to take some care when Chromatic diffs. It looks like [the current build](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=6669161f7a2ba05516df3199) is exhibiting #1862 where sub menus are positioned far away from their parent, so we don't want to set that as the new baseline. My best advice is just to trigger a new build.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -99,14 +101,18 @@ const optionSets = {
...
 
 It is recommended that if the select has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`.
 `;
 const clearableDescription = `
 When the \`clearable\` attribute is set, a clear button will be displayed in the select when a value is selected. Clicking the clear button will clear the selected value and display the placeholder text, if available, or will result in a blank display.
 `;
 
+const loadingVisibleDescription = `
+When the \`loading-visible\` attribute is set, a loading spinner will be displayed in the dropdown of the select along with localizable text that defaults to "Loading". This is useful when the select is loading its options dynamically.
```
jattasNI, line 113:
```suggestion
When the \`loading-visible\` attribute is set, a loading spinner will be displayed in the dropdown of the select along with localizable text that defaults to "Loading…". This is useful when the select is loading its options dynamically.
```

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -235,6 +247,11 @@ export const select: Meta<SelectArgs> = {
             description: 'Emitted when the user changes the selected option.',
             table: { category: apiCategory.events },
             control: false
+        },
+        filterInput: {
+            description: 'Emitted when the user types in the filter input.',
```
jattasNI, line 252:
```suggestion
            name: 'filter-input',
            description: 'Emitted when the user types in the filter input.',
```

packages/nimble-components/src/select/index.ts
```diff
@@ -507,22 +510,30 @@ export class Select
...
+            ) {
+                activeOptionIndex = this.options.indexOf(enabledOptions[0]!);
+            } else if (enabledOptions.length === 0) {
+                activeOptionIndex = -1;
+            }
+            this.setActiveOption(activeOptionIndex);
+        }
+
+        if (this.filterMode !== FilterMode.none) {
+            this.emitFilterInputEvent();
```
jattasNI, line 535: In [the storybook build](https://60e89457a987cf003efc0a5b-cllvgqkkug.chromatic.com/?path=/story/components-select--select&args=filterMode:standard) in the Actions tab, I'm seeing the event be emitted twice for each key press. I know Storybook used to have bugs that would cause this but I haven't seen them recently. Can you check if that's a real issue or a Storybook issue? 

I'm also seeing `change` be fired twice so I suspect Storybook, but worth confirming.

packages/nimble-components/src/select/index.ts
```diff
@@ -853,6 +864,9 @@ export class Select
             this.filterInput.value = '';
         }
 
+        if (this.filterMode !== FilterMode.none) {
+            this.emitFilterInputEvent();
+        }
```
rajsite, line 869: Can this check be merged with the check above it? Or is there some subtle difference in these conditions?
```ts
if (this.filterInput) {...}
if (this.filterMode !== FilterMode.none) {...}
```
Aren't those the same check?

rajsite (approved):

packages/nimble-components/src/select/index.ts
atmgrifter00: Fixed.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -245,10 +242,26 @@ export const select: Meta<SelectArgs> = {
...
-        clearable: false
+        clearable: false,
+        loadingVisible: false
     }
 };
 
+export default metadata;
+
+export const select: StoryObj<SelectArgs> = {
+    ...metadata,
```
rajsite, line 253: I don't think we have had to do this before. And it runs contrary to my understanding of how metadata works in storybook. i.e. should already expect that metadata is shared between each story. Seems fishy

atmgrifter00: I'll try without it.

atmgrifter00: Removed.

packages/nimble-components/src/select/styles.ts
```diff
@@ -151,6 +151,37 @@ export const styles = css`
         align-items: center;
         padding: ${smallPadding} ${mediumPadding};
     }
+
+    .loading-container {
+        padding-left: ${mediumPadding};
+        padding-right: ${mediumPadding};
+        display: flex;
+        height: calc(${controlHeight} - ${smallPadding});
```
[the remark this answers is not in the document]
atmgrifter00: Yep, your `border-box` change fixed this.

PR #2168, opened 2024-06-06
Lint organization

rajsite:
# Pull Request

## 🤨 Rationale

Found that eslint rules and prettier rules were configured inconsistently resulting in different rules being enforced unexpectedly in different packages.

Cleaning up the build configuration resulted in more files being covered (not just src but build scripts, infra, etc), generated files being correctly ignored, and moving the nimble-components linting from our slowest validation step (5m58s) to one of the fastest (1m29s). Now linting is consistently faster than testing.

## 👩‍💻 Implementation

- Modified the `eslint-config-nimble/javascript` and `eslint-config-nimble/typescript` configurations to be generic rules on top of the NI style guide counterparts that can apply to every package within nimble. Provides a place for some of nimble monorepo's stricter rule configuration to be shared until upstreamed to the style guide.
- Added `eslint-config-nimble/components` to encapsulate rules specifically used for FAST component dev.
- Moved all storybook related eslint settings into that package
- Cleaned up pretter and eslint ignores (switching away from [`.eslintignore` files as they are deprecated](https://github.com/eslint/eslint/discussions/17342))

Also aligned on the following organization for eslint files:
- Each package should have ideally a single top-level `.eslintrc.js` file configured as `root: true`
   - Some packages may have child `.eslintrc.js` files for typescript project settings, but generally all rules should be configured top-level
- The top-level `.eslintrc.js` configs only extend rulesets for files matched in `overrides` sections; they do not extend other rule sets outside of `overrides`
- Each top-level config defines an `ignorePatterns` specific to that package. They should all ignore `node_modules` and likely also ignore `dist` among others that are packages specific
- Every JavaScript and TypeScript file should be linted, including config, build scripts, etc.

## 🧪 Testing

Ran locally and on CI. Also added back top level `npm run test` and `npm run lint` commands in case someone wants to run those individually instead of using validate which does both.

## ✅ Checklist


change/@ni-nimble-blazor-b9d8e727-ee91-4d52-8607-7839cca9ff9f.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Blazor linting improvements resulted in a very minor source file change

change/@ni-nimble-components-6ba4c099-b624-4368-b8f4-7fa4c78a590b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Switch to all nimble private linting",
```
rajsite, line 3:
```suggestion
  "type": "patch",
  "comment": "Minor wafer updates due to lint config changes",
```

munteannatan (approved):

packages/storybook/.prettierignore
```diff
@@ -1,26 +1,19 @@
...
 !src/
 !build/
 
 /src/**/*.*
-!src/**/*.ts
-!src/**/*.js
-!src/**/*.md
-!src/**/*.mdx
-!src/**/*.json
+!src/**/*.[ts,js,md,json]
```
jattasNI, line 9:
```suggestion
!src/**/*.[ts,js,md,mdx,json]
```

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/member-ordering */
```
jattasNI, line 1: nit, feel free to leave as-is, but it doesn't look like it'd be too hard to fix this suppression

packages/nimble-components/build/generate-workers/index.ts
```diff
@@ -33,7 +33,7 @@ const sourceCode: string = fs.readFileSync(modulePath, 'utf-8');
 const workerTypesPath: string = resolveModulePath(workerTypes);
 const typesCode: string = fs.readFileSync(workerTypesPath, 'utf-8');
 
-const fileContent: string = `// eslint-disable-next-line no-template-curly-in-string
+const fileContent = `// eslint-disable-next-line no-template-curly-in-string
```
jattasNI, line 36: This eslint-disable looks strange to me. It's not applying to the contents of this file, it's being emitted into the generated file. Do we actually lint the generated files? If so, could we put the disable inside the source contents instead? Or could we just lint the source files? Or even better, could we fix the code not to need to violate this rule?

All of this is pre-existing so it doesn't need to be fixed in this PR, just pointing it out in case there's an easy fix or you think it's worth tracking as tech debt / submitting a follow up PR.

packages/eslint-config-nimble/typescript.js
```diff
@@ -96,96 +44,14 @@ module.exports = {
...
-                '@typescript-eslint/naming-convention': [
-                    'error',
-                    {
-                        selector: 'objectLiteralProperty',
-                        format: ['PascalCase']
-                    }
-                ]
+                ],
+
+                'max-classes-per-file': 'off'
```
jattasNI, line 54: nit: Always nice to have an explanation for rule customizations
```suggestion
                // It's common to define helper classes for tests and it's more readable to do this in the same file
                'max-classes-per-file': 'off'
```

jattasNI (approved):

packages/nimble-components/build/generate-workers/index.ts
munteannatan: it's fixed in this pending PR https://github.com/ni/nimble/pull/2154

atmgrifter00 (approved):

packages/nimble-components/build/generate-workers/source/matrix-renderer.ts
rajsite: I had a couple more things to clean-up there I was keeping in my back pocket but went ahead and spelled it out in an issue for follow-up https://github.com/ni/nimble/issues/2178

PR #2180, opened 2024-06-07
Wafer generation script updates

### LINKED ISSUE #2178: Wafer map matrix-renderer build script and linting
# 🧹 Tech Debt

As part of #2168, it was identified that only non-type checking TypeScript lint rules were running on wafer map code. The PR enabled TypeScript type-checking eslint rules but to avoid conflicts with existing PRs were not updated.

Acceptance criteria:
- Wait for #2168  and #2154 to merge
- Update [matrix-renderer.ts](https://github.com/ni/nimble/blob/main/packages/nimble-components/build/generate-workers/source/matrix-renderer.ts#L1) to remove added eslint disables.
- Migrate[ nimble-components/build/generate-workers/index.ts](https://github.com/ni/nimble/blob/main/packages/nimble-components/build/generate-workers/index.ts) to just be a node JS file instead of a TS file to align with our other node build scripts.
  - Update the paths in the file to be based on `__dirname` instead of how they behave now based on relative paths from the build output directory (like our other node JS scripts). Ssince the script will no longer be built from TS to JS it will become easier to reason about the scripts behavior as part of repo infra.
  - Remove the `index.ts` build script from [tsconfig.json](https://github.com/ni/nimble/blob/main/packages/nimble-components/build/generate-workers/tsconfig.json) include and `generate-workers/.eslintrc.js`
  - This should change the generate-workers build output from having `dist/esm/source/[files]` to `dist/esm/[files]` so make sure that is accounted for (which it is desirable to not have a `source` directory in the build output, so that's good).

rajsite:
# Pull Request

## 🤨 Rationale

In #2168 it was found that lint / prettier settings for the worker source and generation scripts were partially enabled.
This PR addresses the follow-on work from that. Fixes #2178 

## 👩‍💻 Implementation

- Changed the generation script to be a JS file like our other generation scripts. Allows to skip typescript compilation and makes it easier to reason about relative paths in the repo for node.js scripting.
- Made the typescript compile only deal with worker source code and updated corresponding paths.
- Removed references to the `build` directory from nimble-components source code. It should be one-way; build scripts generate source code in source, source should not reach back into build scripts.

## 🧪 Testing

Rely on CI.

## ✅ Checklist


packages/nimble-components/build/generate-workers/index.js
```diff
@@ -0,0 +1,48 @@
...
+const typesDTSOutPath = path.resolve(workersDirectory, 'types.ts');
+const matrixRendererTypesOutPath = path.resolve(
+    workersDirectory,
+    'matrix-renderer.ts'
+);
+
+prepareDirectory(workersDirectory);
+writeFile(workerCodeOutPath, workerCode);
+writeFile(typesDTSOutPath, typesDTS);
+writeFile(matrixRendererTypesOutPath, matrixRendererTypes);
```
jattasNI, line 31: Based on the file name, looks like this variable name is for the .ts, not the .d.ts
```suggestion
const workerCodeOutPath = path.resolve(workersDirectory, 'worker-code.ts');
const typesTSOutPath = path.resolve(workersDirectory, 'types.ts');
const matrixRendererTypesOutPath = path.resolve(
    workersDirectory,
    'matrix-renderer.ts'
);

prepareDirectory(workersDirectory);
writeFile(workerCodeOutPath, workerCode);
writeFile(typesDTSOutPath, typesTS);
writeFile(matrixRendererTypesOutPath, matrixRendererTypes);
```

jattasNI (approved):
Thanks for doing this!

munteannatan (approved):
Thank you!

Issue #1326, opened 2023-06-26
Consistent box-sizing configuration


rajsite:
## 🧹 Tech Debt

We are currently a bit scattershot in where and how we configure the box-sizing mode of nimble-components.
We should be more consistent like with some helper that can do a minimal CSS reset in shadow roots for box-sizing.

We likely should also have recommendations for apps to configure their box-sizing modes more globally but that is outside the scope of this work item.

Proposal:

- [x] Have each application use a custom `display` helper in their `:host` styles #2114
- [ ] Update the display helper to also apply `box-sizing: border-box` to all elements in the shadow root consistently https://github.com/ni/nimble/pull/2184

jattasNI:
Assigned to @rajsite to add more detail to the proposal:
1. what mechanism do we use to do the css reset
2. what's the default value we want for box-sizing
3. what's the per-component work

PR #2184, opened 2024-06-10
Consistent box sizing border box

rajsite:
# Pull Request

## 🤨 Rationale

Switch to using `box-sizing: border-box;` consistently in all our components.
fixes https://github.com/ni/nimble/issues/1326

## 👩‍💻 Implementation

- Each component should use the `display` helper when defined which was made consistent in #2114. This PR leverages that helper to make sure each component sets the elements in the shadow root to use `box-sizing: border-box;` consistently
- Created separate selectors for targeting all children, i.e. `* {}`, the host itself, i.e. `:host {}`, and all before and after pseduo-elements so devtools stays clearer, specifically the selector for `:host` and the selector for `*` are clear in dev tools without the noise of the `::before` and `::after` selectors
- Cleared the manual configuration for `box-sizing`
- Cleaned up some padding configurations (now consistently included in the size inside the border) to be margin configurations (which are outside the border and not part of the size).

## 🧪 Testing

Rely on chromatic, explanation of some of the changes:
- Menu changes look correct, the minimum size now includes the border instead of the border going outside the configured size
- Select filter box changes look improved. The sizing of the filter box and the no options box now fit the control height instead of being larger than the control height
- Rich text changes align with figma (border included in the 40px of the footer section). Added the interaction design figma to the component status table but don't think there are strict visual design docs to reference.
- Recreated the PR from squashed commits so that reviewers can see Chromatic Test changes more easily.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
+++ b/change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Switch to custom box-sizing: border-box consistently",
+  "packageName": "@ni/nimble-components",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/utilities/style/display.ts
+++ b/packages/nimble-components/src/utilities/style/display.ts
@@ -5,8 +5,10 @@ import {
 } from '@microsoft/fast-foundation';
 
 /**
- * This utility will generate the appropriate display style, as well as a style rule
- * to hide the host element when its `hidden` attribute is set.
+ * Each element should use the display utility which will create styles to:
+ * - Set the `:host` display property
+ * - Respond to the `hidden` attribute set on `:host`
+ * - Configure `box-sizing` for `:host`, all elements in shadow root, and `::before` / `::after` pseudoelements
  */
 export const display: typeof foundationDisplay = (
     displayValue: CSSDisplayPropertyValue
--- a/packages/nimble-components/src/utilities/style/display.ts
+++ b/packages/nimble-components/src/utilities/style/display.ts
@@ -10,4 +10,4 @@ import {
  */
 export const display: typeof foundationDisplay = (
     displayValue: CSSDisplayPropertyValue
-) => `${foundationDisplay(displayValue)}`;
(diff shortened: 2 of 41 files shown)


packages/nimble-components/src/utilities/style/display.ts
```diff
@@ -10,4 +10,4 @@ import {
  */
```
jattasNI, line 10: Could you update the docs here and in the spright version of this file to mention the new behavior?

There's also a mention of it in `packages/nimble-components/docs/css-guidelines.md`

change/@ni-nimble-components-83fac5bb-4bc6-4067-951d-def8c7404d0c.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: If you can remember, could you take a look at the Chromatic diffs for systemlink-lib-angular once the Renovate PR uptakes this change? (Note that if there are diffs, approving them on the PR build doesn't seem to reset the baseline, you'll have to approve them on the subsequent build of master.

https://www.chromatic.com/builds?appId=62431f379ac921003aaca05a

And your call whether it's worth informing clients of this change with an announcement or post to the UI WG. If it's just 1-2 pixel diffs then maybe not. But could also be a nice teaching opportunity to educate devs about box-sizing CSS.

rajsite: > If you can remember, could you take a look at the Chromatic diffs for systemlink-lib-angular once the Renovate PR uptakes this change?

Will do.

> And your call whether it's worth informing clients of this change with an announcement or post to the UI WG. If it's just 1-2 pixel diffs then maybe not. But could also be a nice teaching opportunity to educate devs about box-sizing CSS.

Don't think it's super interesting yet. Based on the chromatic diffs pretty much only internals of controls are impacted or if you set a border on our controls manually (but I don't expect users to do that, or for us to support that behaving well).

Would be interesting if we aren't consistent in our apps or Angular / material isn't applying page level `box-sizing: border-box` config. I'll put it my back pocket to take a look and see if it's worth discussing but not planning to discuss it yet.

rajsite: Accepted the diff, only manifested in a one pixel tweak in the details pane: https://www.chromatic.com/build?appId=62431f379ac921003aaca05a&number=8583

PR #2185, opened 2024-06-11
Updated Blazor Wafer Map to use dies table input

munteannatan:
# Pull Request

## 🤨 Rationale

The existing dies input for the blazor wafer map uses json conversion to transmit data to the wafer map component.
This has caused some issues detailed here https://dev.azure.com/ni/DevCentral/_workitems/edit/2687395

## 👩‍💻 Implementation

created a new input for apache arrow table called dies table on the blazor wafer map
this input sends the data to the component as a binary
created an internal input on the wafer map component to receive the binary from blazor and convert it to apache table

created a render finished event which is emitted when the wafer component has rendered images on the canvas and not when hovering

## 🧪 Testing

updated the tests to reflect the changes

## ✅ Checklist


packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -169,6 +169,11 @@ export class WaferMap<
     @observable public dies: WaferMapDie[] = [];
     @observable public diesTable: Table<T> | undefined;
 
+    /**
+     * @internal
+     */
+    @observable public diesTableIPC: Uint8Array | undefined;
```
atmgrifter00, line 175: @rajsite and I talked through these changes a bit, and we both feel that a slight change of direction may be in order. First, it seems like we can do away with this new API, and handle the deserialization in the JS module file in Blazor. To do this, we will need to expose the necessary Apache APIs via the `WaferMap` type. The solution we discussed would involve creating a static member on `WaferMap` that exposed the `tableFromIPC` API. Like the following:
```
class WaferMap {
    ...
    public static Apache = {
        tableFromIPC: tableFromIPC;
    };
    ...
}
```
Then, in `NimbleBlazor.lib.module.js` you could do something like this:
```
WaferMap: {
        setDiesTable: async function (waferMapReference, data) {
           const diesTableIPC = window.customElements.get('nimble-wafer-map').Apache.tableIPC;
           const table = diesTableIPC(data);
           await waferMapReference.setData(table);
        }
        ...
}
```
Note the `setData` API on the component. This is the other major suggestion we think is prudent to do. This API accomplishes a few things:
- It consolidates the web component API, which makes it less confusing
- It mirrors the pattern we already have in place for the `Table`
- We can now synchronously validate the data within the `setData` implementation
- Because it is asynchronous, you can now await pending updates before processing the new data.

packages/blazor-workspace/NimbleBlazor/Components/NimbleWaferMap.razor.cs
```diff
@@ -118,6 +123,23 @@ public partial class NimbleWaferMap : ComponentBase
         }
     }
 
+    /// <summary>
+    /// Represents the input data, an apache arrow Table, which fills the wafer map with content
+    /// </summary>
+    [Parameter]
+    public RecordBatch? DiesTable
```
atmgrifter00, line 130: What's the reason for using `RecordBatch` here instead of the Apache `Table` type?

packages/nimble-components/src/wafer-map/index.ts
rajsite: Sounds like a good summary of what we discussed. Just updated the example to call the static `Arrow` instead of `Apache`

packages/blazor-workspace/NimbleBlazor/Components/NimbleWaferMap.razor.cs
munteannatan: The [ArrowStreamWriter](https://github.com/apache/arrow/blob/main/csharp/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs) does not support Tables from what I understand

atmgrifter00, line 130: So, I think we need to take the `async` pattern all the way, and do away with this property-based API for setting data on the `WaferMap`, and transition to the `async` method-based approach as we have with the `Table`.

packages/blazor-workspace/NimbleBlazor/Components/NimbleWaferMap.razor.cs
```diff
@@ -135,6 +138,19 @@ public partial class NimbleWaferMap : ComponentBase
         }
     }
 
+    /// <summary>
+    /// Sets the data in the wafer map.
+    /// </summary>
+    /// <param name="data">The input data, an Apache.Arrow.RecordBatch, which fills the wafer map with content</param>
+    public async Task SetDataAsync(RecordBatch data)
```
atmgrifter00, line 145: This doesn't have to be part of this PR, but I think the ideal would be to even get rid of the `Dies` property, and just have a method overload of `SetDataAsync` that takes an `IEnumerable<WaferDie>` as its parameter.

munteannatan: Yes, I think that is the direction, but also we have to remove it from the nimble component first

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -282,6 +288,14 @@ export class WaferMap<
         return this.diesTable !== undefined;
     }
 
+    public async setData(tableData: Table): Promise<void> {
```
jattasNI, line 291: If this is internal, let's add an `@internal` comment to it.

If it's part of the public API, let's add a description both in a method comment and in Storybook docs.

Same point for the `render-finished` event: if clients are expected to listen to that event, we should document it in Storybook. (We don't have a mechanism for marking events as internal other than just leaving them undocumented in storybook)

jattasNI (approved):

msmithNI (approved):

packages/blazor-workspace/NimbleBlazor/Components/NimbleWaferMap.razor.cs
rajsite: The goal of the SetDataAsync method is that it unlike property setting we know when it can be finished. So I was not expecting to see the render-finished event still exist. I'd expect the SetDataAsync method to finish when render has finished.

packages/nimble-components/src/wafer-map/index.ts
rajsite: I don't expect the render-finished event to continue to exist. I do expect setData to be public

packages/nimble-components/src/wafer-map/index.ts
```diff
@@ -282,6 +288,14 @@ export class WaferMap<
...
+        if (this.currentTask !== undefined) {
+            await this.currentTask;
+        }
+        this.diesTable = tableData;
+        this.waferMapUpdateTracker.queueUpdate();
```
rajsite, line 296: Conceptually we want this method to allow a user to set data and reliably wait for rendering to have finished. And it should be our primary entry point. So like the following:

1. Take either an Arrow table or old Wafer dies format
2. Assign to the right property (`diesTable` or `dies`)
3. Wait for that property update to be batched and queued ([similar to the table](https://github.com/ni/nimble/blob/d37ca77cda0729b42f05b4bac79be8ccc98ceb22/packages/nimble-components/src/table/index.ts#L280))
4. Wait for the `currentTask` created from the update to finish so that we know the side effect of setting the data has been resolved

Then we can avoid the `render-finished` event because we have a reliable way to know when the render associated with a `setData` is finished.

packages/blazor-workspace/NimbleBlazor/Components/NimbleWaferMap.razor.cs
rajsite, line 145: I think we need two overloads, one that takes the apache record format and one that takes the previous dies format.

packages/nimble-components/src/wafer-map/index.ts
rajsite, line 291: Talking to @atmgrifter00 we should make a tech debt to do data validation as part of `setData` after the `dies` and `diesTable` properties are removed. That way for validation related to data we can throw an error instead of have users needing to inspect the validity state (other validity based on properties would still be part of the validity state).

change/@ni-nimble-blazor-f8fbcdc6-38cc-4c51-8016-115c43182376.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @munteannatan can you give an update on status of the PR and expectations?
As an fyi most reviewers will be OOO W-F next week.

munteannatan: Sorry about this, we are also OOO until next week

change/@ni-nimble-blazor-f8fbcdc6-38cc-4c51-8016-115c43182376.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: @munteannatan I see that some changes were pushed but didn't see additional comments and there are unresolved issues so assuming it's still in work. Just FYI that @atmgrifter00 and I are OOO for the rest of this week so if comments are posted / reviews needed we won't be available until next week.

packages/nimble-components/src/wafer-map/index.ts
munteannatan: updated the list here https://github.com/ni/nimble/issues/2034

packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
```diff
@@ -47,6 +47,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Apache.Arrow" Version="16.1.0" />
```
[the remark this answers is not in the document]
jattasNI: @munteannatan there was a comment about setting an exact version for the npm `apache-arrow` dependency in `package.json` but it looks like it still a version range (`^16.1.0` versus `16.1.0`). Was that an oversight or is there rationale for keeping it as a range?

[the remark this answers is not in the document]
munteannatan: No reason, just an oversight, will fix it

jattasNI (approved):
@rajsite I verified to the best of my ability that the changes addressed your feedback. Because you're OOO today and there is a desire to complete this work this week, I'm going to bypass you but have assurances that @munteannatan and team can address any feedback next week.

PR #2189, opened 2024-06-12
Limit combobox and select dropdown to 10.5 visible options

### LINKED ISSUE #1870: Limit height of dropdown listbox to 10 1/2 elements
## 😯 Problem to Solve

The select and combobox listbox has a max-height set to prevent overflow beyond the browser viewport. However, when dealing with a long list of items, there is no specific guideline for limiting the height.

Considering that both the select and combobox controls support autocomplete/filtering, our usage guidance allows for a potentially large number of elements.

To improve user experience, we aim to avoid forcing users to visually scan lengthy dropdown lists. Therefore, we will display a maximum of 10 items at a time. Additionally, we should partially show the 11th item to indicate the presence of more items in the listbox.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1870

## 👩‍💻 Implementation

In `setPositioning`, where `maxHeight` of the listbox was already being calculated, now limit to maximum of 10.5 visible items. For a filterable select, the height of the filter adds to the height of the listbox, i.e. we still show 10.5 options. If a select contains grouped options, extra space is not reserved for group headers, meaning fewer than 10.5 options may be visible. This results in a simpler implementation and prevents height changes/jiggle when typing in the filter box and more than 10 options are matched.

## 🧪 Testing

Added visual tests for the height limiting behavior, including with/without filter input and above/below.

## ✅ Checklist


m-akinc:
@atmgrifter00 mind buddying this?

packages/nimble-components/src/combobox/index.ts
```diff
@@ -625,9 +632,27 @@ export class Combobox
...
             ? Math.trunc(currentBox.top)
             : Math.trunc(availableBottom);
+
+        const controlHeight = parseInt(
+            controlHeightToken.getValueFor(this),
+            10
+        );
+        const listboxInnerPadding = parseInt(
+            smallPadding.getValueFor(this),
+            10
```
atmgrifter00, line 645: This does default to base-10, so you can remove this parameter.

packages/nimble-components/src/select/index.ts
```diff
@@ -1008,9 +1013,29 @@ export class Select
...
+            10
+        );
+        const listboxInnerPadding = parseInt(
+            smallPadding.getValueFor(this),
+            10
+        );
+        const listboxGap = listboxInnerPadding; // both use smallPadding
+        const listboxBorderHeight = 2; // 1px top and bottom
+        const filterHeight = this.filterMode !== FilterMode.none ? controlHeight : 0;
+        this.maxHeight = Math.min(
```
atmgrifter00, line 1031: It feels like we could single-source all of this into a static helper function that takes the element to pull the tokens from as an argument.

m-akinc: I'm not 100% sure which part you think should go in a helper, but I made one that gets the dropdown height for 10.5 options. I put this in a new `helpers.ts` file under `patterns/dropdown`.

atmgrifter00: You could take a bit further, where you have `Select` implement `DropdownPattern` (which it should be doing anyway), and then move the `avaiableHeight` calculation into the static helper method,  making its `element` parameter typed to `DropdownPattern` as well (to access the `position`).

atmgrifter00 (approved):

packages/nimble-components/src/patterns/dropdown/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+    const listboxGap = listboxInnerPadding; // both use smallPadding
+    const listboxBorderHeight = 2; // 1px top and bottom
+    const filterHeight = filterVisible ? controlHeight : 0;
+    return (
+        controlHeight * maxVisibleOptions
+        + listboxGap
+        + listboxBorderHeight
+        + filterHeight
+        + listboxInnerPadding
+    );
```
rajsite, line 23: @atmgrifter00 @m-akinc Can we keep most of this in CSS? Particularly logic that is taking token values and just adding stuff together. We should be doing the minimal amount of work possible in JS and create private css custom properties that are used in the CSS for the minimal bits of calculation we can't do in CSS. Marking reject on the PR because I don't see any discussion related to keeping layout calculations out of JS so seems overlooked.

rajsite (changes requested):

packages/nimble-components/src/patterns/dropdown/helpers.ts
```diff
@@ -0,0 +1,24 @@
+import {
+    controlHeight as controlHeightToken,
+    smallPadding
+} from '../../theme-provider/design-tokens';
+
+const maxVisibleOptions = 10.5;
```
jattasNI, line 6: I guess the idea of "show an extra half item to indicate that there are more you can scroll to" breaks down when there are groups since our group items aren't the same height as our list items.

image

I'm not sure what, if anything, to do about it. I don't think it's worth it (or even possible) to do calculations that factor in how many groups are visible. I think our options are either:
1. leave it as is
2. decide that the extra half item isn't worth it since it won't work for some cases, and just pick a fixed pixel height instead of doing all these calculations

packages/nimble-components/src/patterns/dropdown/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function getDropdownMaxHeight(
```
jattasNI, line 8: The `loading-visible` attribute just went in. How should the presence of the loading item impact this height? Since it's not scrollable I'd vote to treat it similar to the search field, and have it add a fixed amount to the height.

packages/nimble-components/src/patterns/dropdown/helpers.ts
rajsite: Other ideas that came up:
1. Is there some other partial size that will "generally clip" like 10.75
2. [Scroll driven animations are not available in all browsers today](https://caniuse.com/mdn-css_properties_animation-timeline_scroll) but maybe in the future we can use them to make a useful style like [this example](https://x.com/jh3yy/status/1758284287189958973) ([codepen](https://codepen.io/jh3y/pen/oNVmzby) for supported browsers)

m-akinc: I think we leave it as is, since it still makes sense to pick a limit based on a number of visible options, and in the (more common?) case where options are not grouped, showing a partial item is nice.

packages/nimble-components/src/patterns/dropdown/styles.ts
m-akinc, line 1: Doesn't look very good when there's a loading indicator. Maybe we want to conditionally add a dividing line and/or modify the padding/alignment? FYI @NIbokeefe
image

packages/nimble-components/src/patterns/dropdown/helpers.ts
m-akinc: Yep. Even got to use `:has()` (which should be okay, since it was added to Firefox in Dec 2023)!

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
```diff
@@ -47,51 +45,71 @@ interface SelectMatrixStoryOptions {
...
 const component = ({
-    positionState, filterMode, loadingVisible, placeholder, grouped, optionsOutsideGroup, slottedLabel
+    positionState, filterMode, loadingVisible, placeholder, grouped, optionsOutsideGroup, slottedLabel, manyOptions
 }: SelectMatrixStoryOptions): ViewTemplate => html`
     <${selectTag} open 
-        position="${() => positionState[0]}" 
-        style="${() => positionState[1]} width: 250px; ${isChromatic() ? '--ni-private-spinner-animation-play-state:paused' : ''}"
+        position="${() => positionState}"
+        style="${() => (positionState === DropdownPosition.below ? 'margin-bottom: 120px;' : `margin-top: ${manyOptions ? 430 : 180}px;`)}"
+        style="${isChromatic() ? '--ni-private-spinner-animation-play-state:paused' : ''}"
```
jattasNI, line 58: In this screenshot it looks like animation has happened. Can you double check if this is working / needs to be applied elsewhere?

https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=6670cfca62bf533e2c2727e6

image

jattasNI (approved):

packages/storybook/src/nimble/select/select-opened-matrix.stories.ts
m-akinc: Didn't find this mistake anywhere else.

packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -139,12 +139,34 @@ export const styles = css`
...
+                    var(--ni-private-listbox-loading-indicator-height)
+            ),
+            calc(
+                var(--ni-private-listbox-available-viewport-height) -
+                    ${smallPadding}
+            )
+        );
         box-shadow: ${elevation2BoxShadow};
-        border: 1px solid ${popupBorderColor};
+        border: var(--ni-private-listbox-border-width) solid ${popupBorderColor};
```
rajsite, line 158: Can we use the existing `borderWidth` token instead of making a new token?

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: we have one more listbox implementation that is using a different size metric currently. I suspect we want all our dropdowns to behave the same: https://github.com/ni/nimble/blob/51aef514fe30e2cac694b885f3ebec000aed1a43/packages/nimble-components/src/rich-text/mention-listbox/styles.ts#L23

rajsite: @fredvisser should the rich text dropdown (which also shows a large number of items) behave similarly and increase from 5.5 to 10.5 items shown?
image

packages/nimble-components/src/patterns/dropdown/helpers.ts
rajsite: neat! yea definitely looking much better

packages/nimble-components/src/select/index.ts
```diff
@@ -1231,8 +1231,8 @@ export class Select
         this.$emit('filter-input', eventDetail, { bubbles: true });
     }
 
-    private maxHeightChanged(): void {
-        this.updateListboxMaxHeightCssVariable();
+    private availableViewportHeightChanged(): void {
```
rajsite, line 1234: looking at this again, I think the reason for this updatee the observable property -> listen for Changed -> call style.setProperty is a relic from when we did not own the template.

We should now just update the template to do the style binding.

```
<div class="listbox" style="--ni-private-listbox-available-viewport-height: ${x => x.availableViewportHeight}px;"
```

rajsite (commented):
Tweaks to clean-up old approach and qs about applying to all listboxes

packages/storybook/src/nimble/combobox/combobox-opened-matrix.stories.ts
```diff
@@ -45,28 +58,54 @@ if (remaining.length > 0) {
     throw new Error('New backgrounds need to be supported');
 }
 
+// prettier-ignore
```
rajsite, line 61: Are all these prettier-ignores really needed? How poorly is it formatting a couple lines?

m-akinc: Apparently they made no difference. Removed.

rajsite (commented):
Current changes look good, still awaiting outcome of open questions

packages/nimble-components/src/patterns/dropdown/styles.ts
NIbokeefe: > Doesn't look very good when there's a loading indicator. Maybe we want to conditionally add a dividing line and/or modify the padding/alignment? FYI @NIbokeefe image

@m-akinc @atmgrifter00 This was not my favorite design out of the bunch, I think _part_ of the solution was having the loading indicator at the top instead of the bottom.  Even with the loading indicator at the top some alternative choices needed to be made to finesse the design.

I spoke with @fredvisser about this issue, (correct us if we have it wrong) this a corner case to have populated items in combination to the loading indicator. Majority of the time it would be just the search and loading indicator shown.  So adding a conditional line would create an additional unnecessary line unless that condition is only when items are present.  

Corner case or not, if we want this to look better 100% of the time the loading works better above the items, so the consistency of it's location is always the same, directly below the search whether items are present or not.

m-akinc: > unless that condition is only when items are present

That was my thought.

> if we want this to look better 100% of the time the loading works better above the items

Consistency of location aside, I don't think moving it above the items solves the issue I'm pointing out here. It would still result in a partially clipped option (now at the top of the scroll area) when the top option is partially scrolled out of view. And presumably, it wouldn't change the vertical centering issue (would still look too near the items).

If we don't have a clear answer for these issues now but think it's worth addressing, I can just create a tech debt item to track this. And if we think a conditional separator line would at least be better than without, I can include that in this PR as a stop-gap.

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
fredvisser: I wouldn't change the rich text @mention dropdown just to maintain consistency with combobox and select. Those controls need to be general purpose, where the @mention dropdown is more narrowly tailored. Let's wait until we specific feedback for the @mention dropdown before changing it.

packages/nimble-components/src/patterns/dropdown/styles.ts
atmgrifter00: @NIbokeefe, also, just to push on the point of _why_ the loading indicator is likely better at the bottom when the dropdown is below, and at the top when the dropdown is above, is that the options are still interactive (i.e. clickable) while the indicator is present, and it would be annoying if right when a user was about to click the option they wanted that the loading indicator was removed, all of the options shift, and the user clicks what they didn't intend. Because of where the indicator is now, that won't happen.

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
rajsite: What's the usage specific reason that a list of users should have a different number of values shown? Don't think I follow. As far as I'm aware showing 5.5 items was arbitrarily defined. Guess I should ask what defines 10.5 as a good number and when to use one or the other?

rajsite: Also don't think the rich text impl is as robust and probably should align on being aware of the viewport size. If we do that, do we keep the work of it being intentionally a different number of items?

fredvisser: Max 10.5 rows for the select and combobox was an inference from the [select best](https://www.nngroup.com/articles/listbox-dropdown/) [practices](https://designsystem.digital.gov/components/combo-box/#guidance) that there shouldn't be more than 15 items in a (non-filterable) select. 

Virtually all design systems have a limit, but most don't explain or document the rationale. [IBM Carbon](https://carbondesignsystem.com/components/dropdown/usage/#scrolling-behavior) emphasizes showing the half row, and suggests 6.5 as a starting point. Our rows aren't very tall, so 10.5 is a similar height.

When we fix the rich text @mention dropdown to clip/scroll at the view boundary, it wouldn't be unexpected to maintain 5.5 as the limit. The @mention dropdown is invoked while typing, and is a search-first, keyboard-first workflow, so it's reasonable to make a different choice.

rajsite: For dev actions, I think it's clear that we should make the richtext mention listbox viewport aware like we do for select and combobox. @m-akinc I think we could do that either as part of this PR or to keep the review smaller / isolated we can do it in a follow-up PR (preference since sometime richtext changes end up more complex and would be easier to revert / manage standalone).

The discussion about how many rows to show for the richtext does not have to block the PR. I like the idea that the space of 10.5 of our current short rows is what we allocate. If / when we have double height rows for mention options (show user profile + email, etc) then we would show 5.5 options and fill a similar space.

@fredvisser I went and played with the current rich text mention support and I actually think showing more options might be useful because our filtering algorithm isn't very smart. It looks like we filter by any part of the string and then sort alphabetically by the first and then last name. If we filtered and sorted "good" then the option you want would always be early and I can see an argument for needing to show fewer options because it's probably always the first or second. But right now we filter kinda meh.

Like in the following it's pretty unintuitive that typing `@mar` doesn't show "mark" as like one of the first options. ([teams link to gif to avoid showing names public](https://nio365.sharepoint.com/:i:/s/SystemLinkDesignSystem/EagHOJ2PLxVMp2RavBcTxfMBiPuJXddyUkDuw73-r39_pA?e=w1jC13)). I think having the longer list like shown there helps with the idea that you still need to filter more / there are a lot of matches to deal with.

rajsite: Gonna try and resolve this thread quicker offline with @fredvisser

packages/nimble-components/src/patterns/dropdown/styles.ts
fredvisser: @m-akinc - We'll track the "Loading…" status separator as visual tech debt [here](https://github.com/ni/nimble/issues/2215), and not hold up this PR.

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
m-akinc: Sounds like we can address the richtext mention listbox separately from this PR.

m-akinc:
@NIbokeefe sorry, I reset your vote without thinking. No action needed on your part.

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
rajsite: > Sounds like we can address the richtext mention listbox separately from this PR.

@m-akinc can you make sure the work is captured. Maybe add it as a task in #1870 and update the PR description to say this PR partially resolves the issue

rajsite: > Gonna try and resolve this thread quicker offline with @fredvisser

Captured initial guidance on 10.5 vs 5.5 options from discussion here: https://github.com/ni/nimble/issues/2216

So we will keep rich text 5.5 options but it does need to be made viewport aware as discussed.

rajsite (approved):

change/@ni-nimble-components-878686fc-b8ba-42b4-a8d3-ac0bf0919b9d.json
m-akinc: Created new bug to track limiting the height of the richtext mention listbox to the viewport: #2217

PR #2190, opened 2024-06-12
Move column sorting properties to a column mixin

mollykreis:
# Pull Request

## 🤨 Rationale

Currently, there is no way for a table column to not support sorting because the sorting attributes (`sorting-disabled`, `sort-index`, and `sort-direction`) are defined on the base `TableColumn` class. The menu button column, however, should not allow sorting. Therefore, the three sorting attributes are being moved off of the base `TableColumn` class and into a mixin, that every existing column now uses.

## 👩‍💻 Implementation

**nimble-components**
- Create mixin for sortable columns
- Remove sorting properties from the base `TableColumn` class
- Add `sortingDisabled` to column internals

**nimble-angular**
- Remove sorting properties from the base column directive and into each column type directive
- Add support for `sortingDisabled` (including tests) on all columns since this hadn't been exposed through Angular

**nimble-blazor**
- Create `ISortableColumn` interface
- Move sorting properties from the base column component and into `ISortableColumn`
- Update all columns to implement `ISortableColumn`
- Add support for `SortingDisabled` since this hadn't been exposed through Blazor
- Update tests for sortable columns
- Update tests for base columns since it previously was only testing sorting properties. I've removed those and added tests for missing base properties of `ColumnId`, `ActionMenuSlot`, `ActionMenuLabel`, and `ColumnHidden`.

## 🧪 Testing

- Unit tests

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table-column/text-base/index.ts
```diff
@@ -36,6 +37,8 @@ export function mixinTextBase<
     TColumnValidator extends ColumnValidator<[]> = ColumnValidator<[]>
 >(base: TBase) {
     return mixinGroupableColumnAPI(
-        mixinFractionalWidthColumnAPI(mixinColumnWithPlaceholderAPI(base))
+        mixinFractionalWidthColumnAPI(
+            mixinColumnWithPlaceholderAPI(mixinSortableColumnAPI(base))
+        )
```
m-akinc, line 42: nit
```suggestion
        mixinFractionalWidthColumnAPI(
            mixinColumnWithPlaceholderAPI(
                mixinSortableColumnAPI(base)
            )
        )
```

m-akinc (approved):

packages/nimble-components/src/table-column/text-base/index.ts
mollykreis: Prettier does not agree with your suggestion 😄 

Assuming you don't feel really strongly about this specific formatting, I think I'll let prettier have its way here.

m-akinc: Sometimes I hate prettier.

packages/nimble-components/src/table-column/mixins/sortable-column.ts
```diff
@@ -0,0 +1,70 @@
...
+    TBase extends SortableTableColumnConstructor
+>(base: TBase) {
+    /**
+     * The Mixin that provides a concrete column with the API to allow sorting
+     * by the values in that column.
+     */
+    abstract class SortableColumn extends base {
+        public sortingDisabled = false;
+
+        public sortIndex?: number | null = null;
```
rajsite, line 23: Can we leave this uninitialized to undefined like it was previously in base? (prefer undefined over null)

```suggestion
        public sortIndex?: number | null;
```

rajsite: Hopefully we can do the same for groupIndex above as well

packages/nimble-components/src/table-column/mixins/groupable-column.ts
```diff
@@ -26,11 +26,7 @@ export function mixinGroupableColumnAPI<
         }
 
         public groupIndexChanged(): void {
-            if (typeof this.groupIndex === 'number') {
-                this.columnInternals.groupIndex = this.groupIndex;
-            } else {
-                this.columnInternals.groupIndex = undefined;
-            }
+            this.columnInternals.groupIndex = this.groupIndex ?? undefined;
```
rajsite, line 29: Can you do the same in the fractional-width-column mixins for fractionalWidthChanged and minPixelWidthChanged?

packages/nimble-components/src/table-column/mixins/tests/sortable-column.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+            delegatedEvents: [],
+            validator: new ColumnValidator<[]>([])
+        };
+    }
+}
+
+// prettier-ignore
+async function setup(): Promise<Fixture<TestTableColumn>> {
+    return fixture(columnName);
+}
```
rajsite, line 36: no action needed (unless the ignore was accidental): prettier-ignore only impacts the following "node" right, i.e. just this function.. how badly does it mess this function up lol?

packages/nimble-components/src/table/models/table-update-tracker.ts
```diff
@@ -118,9 +118,9 @@ export class TableUpdateTracker<
         ) {
             this.track('columnDefinition');
         } else if (
-            isColumnProperty(changedColumnProperty, 'sortingDisabled')
```
rajsite, line 121: no action needed (maybe a tech debt if it seems like a good direction): huh... half-thought but I think I like the idea of moving towards the table only relying on columnInternals and not the public api of columns. Seems like we are getting close.

change/@ni-nimble-angular-471f2c22-eef2-4640-9335-5fc12cb49c02.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Great job churning through this so quick. Wish there wasn't so much boilerplate needed in the wrappers 😫 but looks good 🚢

rajsite (approved):

packages/nimble-components/src/table-column/mixins/tests/sortable-column.spec.ts
mollykreis: I had copied this from the other mixin tests, but it doesn't look like it's actually needed. I'll remove it from all of them.

packages/nimble-components/src/table/models/table-update-tracker.ts
mollykreis: I think we are getting close. I don't have a strong opinion at this point.

rajsite: Feels nice academically, but I can't think of a concrete reason to do it yet, so no action needed if no strong opinions 👍

rajsite:
Gonna override merge since it's pretty mechanical and following existing patterns. If concerns @jattasNI @atmgrifter00 @msmithNI I'm sure we can address in a follow-up

Issue #2128, opened 2024-05-21
Create HLD template for table columns


mollykreis:
# 🧹 Tech Debt
We have a few different HLD templates for new components within nimble, but none of them are great fits for creating new table columns. As a result, each column HLD follows a slightly different format.

Some of the unique considerations of table columns are:
- sorting and grouping behavior
- sizing behavior
- placeholders (or lack thereof)
- cell view & group row views
- delegated events
- `focusedRecycleCallback` behavior
- integration with keyboard navigation
   - Focusable elements in cell view should be set to `tabindex=-1`, and those elements need to support setting their `tabIndex` property programmatically and having the right thing happen
   - Override `tabbableChildren` on the cell view to return focusable children of cell view

PR #2192, opened 2024-06-13
Create template for table column HLDs

m-akinc:
# Pull Request

## 🤨 Rationale

Addresses #2128 

## 👩‍💻 Implementation

Created template HLD. I considered puting this in `nimble-components/src/table/specs/table-column-specs` but opted to stick with the existing spec template directory.

## 🧪 Testing

NA

## ✅ Checklist


specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,134 @@
+# Table Column [Template]
+
+## Overview
```
jattasNI, line 3: Probably worth linking to https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-columns-hld.md#column-type-philosophy from one or more places in this template. It has useful info on when to create new columns and how to name them.

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,134 @@
...
+- *Relevant historical or background information*
+- *Link to relevant work items, related existing issues, etc.*
+
+### Features
+
+*A list of the key features unique to this component.*
+
+### Non-goals
+
+*A list of use cases, features, or functionality which are **not** goals for the component.*
```
jattasNI, line 18: Here are a couple ideas for additional content/instructions to incorporate. I think they're both already covered at a high level by sections in the template but it might be useful to augment some of the instructions as a reminder of the level of detail we're looking for. Just a suggestion though; your call if you can fit them in succinctly.

1. A common reason to create a new column would be to expose capability of an existing nimble component within the table. When we do that, we should evaluate each of the existing component's APIs and decide if they should be configurable at the row level, column level, or not at all. Could maybe request that those APIs be listed explicitly in "Non-goals" and "Column Element"? 
2. Every column should have 1 or more `xxx-field-name` attributes that populate per-row state from the data. Should we call that out specifically in "Column Element" or will it fall out naturally?

jattasNI (approved):

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Sizing
+
+*Will the column use a fractional width, fixed width, or both (via `width-mode` attribute)?*
+
+### Placeholder
+
+*Will the column support a `placeholder` for `undefined` and `null` values? Will the group header view display the standard strings for `undefined` and `null` values, or will there be something unique?*
+
+### Delegated Events
```
atmgrifter00, line 98: Do we need to provide descriptions for _why_ some of these things might be necessary (i.e. give examples)?

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,134 @@
...
+
+*Will the column be groupable? Describe any interesting use cases.*
+
+### Sizing
+
+*Will the column use a fractional width, fixed width, or both (via `width-mode` attribute)?*
+
+### Placeholder
+
+*Will the column support a `placeholder` for `undefined` and `null` values? Will the group header view display the standard strings for `undefined` and `null` values, or will there be something unique?*
```
mollykreis, line 96: I think this section could be simplified, which would also reduce some of the assumptions that I think are being made in these statements about the column and the developer's understanding of "standard" behavior.

For example, what if you said something like this:
_What is the behavior of cells and group rows when the value is `undefined` or `null`? Are there additional values (e.g. empty string) for which placeholders or default rendering need to be considered?_

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,134 @@
...
+
+*Will the cells delegate any events to the column element?*
+
+### Focus Recycling
+
+*Will the cell view need to override `focusedRecycleCallback()` to perform any actions when the cell is recycled while it has focus? If so, explain.*
+
+### Interactions
+
+*Will the cell render any interactive elements? You must override `tabbableChildren` on the cell view to return any focusable children. These elements must be initially configured with `tabindex=-1`, and the table will programmatically update `tabIndex` as needed. Make sure that any such elements properly forward the `tabIndex` value to shadow DOM elements.*
```
mollykreis, line 108: This section is a bit confusing to me. What would a spec writer be expected to put here? The question asked is a yes/no question and all the other information is informative for implementation but not necessarily the HLD itself.

specs/templates/table-column-hld.md
m-akinc: > Every column should have 1 or more xxx-field-name attributes that populate per-row state from the data. Should we call that out specifically in "Column Element" or will it fall out naturally?

I think that will fall out naturally. And in some rare cases, a column might not have such an attribute (like the button column I was proposing).

specs/templates/table-column-hld.md
m-akinc: Added an example `placeholder` use case.

specs/templates/table-column-hld.md
m-akinc: Merged your suggestion with a specific question and example about `placeholder` attributes.

specs/templates/table-column-hld.md
m-akinc: Worded it to be more instructive. Removed the implementation reminder about initializing `tabindex` to -1.

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Sizing
+
+*Will the column use a fractional width, fixed width, or both (via `width-mode` attribute)?*
+
+### Placeholder
+
+*What is the behavior of cells and group rows when the value is `undefined` or `null`? Will a `placeholder` be supported to display a specific interpretation for missing values (e.g. showing "Never run" in a date-time column of latest execution times)? Are there additional values (e.g. empty string) for which placeholders or default rendering need to be considered?*
+
+### Delegated Events
```
atmgrifter00, line 97: I guess what I was primarily concerned about with [this comment](https://github.com/ni/nimble/pull/2192/files#r1642954829) was the delegated events section. At the surface, particularly if we're talking about a someone who has low familiarity with other examples of columns, seeing this section title wouldn't be enough to know how to answer the given question (IMO). A brief description of why cells might delegate events to a column might be helpful.

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,133 @@
...
+
+*Will the cells delegate any events to the column element?*
+
+### Focus Recycling
+
+*Will the cell view need to override `focusedRecycleCallback()` to perform any actions when the cell is recycled while it has focus? If so, explain.*
```
atmgrifter00, line 103: In a similar vein to the delegated events section, it might be helpful to quickly mention why/when cell recycling happens, and an example of a column that reacts to this. It might be sufficient to simply link to existing docs if they exist.

atmgrifter00 (approved):

specs/templates/table-column-hld.md
m-akinc: Oh, my mistake. I've added some details on event delegation.

specs/templates/table-column-hld.md
m-akinc: Added a little more detail, and also a link to the issue that might make `focusedRecycleCallback` obsolete.

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,133 @@
...
+</nimble-table>
+```
+
+### API
+
+#### Column Element
+
+_Element Name_
+
+-   `nimble-table-column-*`
```
rajsite, line 44: We should either link to or move the column type philosophy stuff here: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-columns-hld.md#column-type-philosophy

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,133 @@
+# Table Column [Template]
+
+*Refer to the [table column type philosophy docs](/packages/nimble-components/src/table/specs/table-columns-hld.md#column-type-philosophy) for column creation policies and naming conventions.*
```
rajsite, line 3: maybe we should just move this line down to the element name section

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Delegated Events
+
+*Are there any events emitted by components in cells that a client might be interested in? Because clients cannot easily subscribe to every cell view, such events should be delegated to the column element so that it can emit an equivalent event for clients to subscribe to. Names of delegated events (e.g. `'click'`, `'keydown'`, etc.) should be listed in the `delegatedEvents` property of the column internals options, and the column should set up a listener to handle the `delegated-event` event that emits a new event for clients to listen for. For example, `click` events from the cell might be delegated to the column and re-emitted as `button-click` events. Describe here any events that will be delegated (e.g. `click`), the new event name(s) that will be emitted by the column (e.g. `button-click`), and any extra data that will be included in those event details (e.g. `recordId` field).*
```
rajsite, line 99: this is pretty lengthy, maybe some line breaks

```
*Are there any events emitted by components in cells that a client might be interested in? Because clients cannot easily subscribe to every cell view, such events should be delegated to the column element so that it can emit an equivalent event for clients to subscribe to. Names of delegated events (e.g. `'click'`, `'keydown'`, etc.) should be listed in the `delegatedEvents` property of the column internals options, and the column should set up a listener to handle the `delegated-event` event that emits a new event for clients to listen for.*

*For example, `click` events from the cell might be delegated to the column and re-emitted as `button-click` events. Describe here any events that will be delegated (e.g. `click`), the new event name(s) that will be emitted by the column (e.g. `button-click`), and any extra data that will be included in those event details (e.g. `recordId` field).*
```

Think it's also helpful to just have an example folks can edit instead of describing what an example would look like, maybe something like:
```
- delegated event name: `click` (name of event captured from the cell-view to delegate to the column)
- re-emitted event name: `button-column-click` (name of event emitted from the column)
- re-emitted event detail: `recordId` (extra details included)
```

rajsite (approved):

specs/templates/table-column-hld.md
m-akinc: There's a link to it at the top of the file. It will be essentially the first thing someone filling out the template will see (and then delete). There's no need to also link it from here as well, right?

specs/templates/table-column-hld.md
m-akinc: I put it right before the element name section, as it also gives guidance on general column type creation policy.

specs/templates/table-column-hld.md
m-akinc: Moved the existing link to right before the element name section.

specs/templates/table-column-hld.md
m-akinc: I gave an example written in sentences, as opposed to just a list.

specs/templates/table-column-hld.md
```diff
@@ -0,0 +1,143 @@
...
+-   `nimble-table-column-*-group-header-view`
+
+_Rendering_
+
+*Describe what the group header view will render.*
+
+### Sorting / Grouping
+
+*Will the column be sortable and/or groupable? If so, which `xxx-field-name` attribute will be assigned to  `operandDataRecordFieldName`? If sortable, which `TableColumnSortOperation` will be used? Describe any interesting use cases.*
+
```
mollykreis, line 87: Is this not an extra blank line? I'm surprised prettier isn't complaining.

mollykreis (approved):

specs/templates/table-column-hld.md
m-akinc: It is. We're not running prettier on the top-level specs folder. I manually ran the formatter on it, and it removed the extra line, as well as replaced all `*`s with `_`s for italicizing.

rajsite: Can you make a follow-up PR + tech debt to enable prettier there? Guessing it's not intentional

m-akinc: Created #2223

PR #2193, opened 2024-06-14
Update breadcrumb styling

### LINKED ISSUE #2163: Update breadcrumb styling
# 🙋 Feature Request

## 😯 Problem to Solve

1. The latest breadcrumb visual design swaps the color emphasis on the prominent version of the breadcrumb items. We should make the corresponding change in nimble.
image
1. The latest design also changes the keyboard focus state from a box, to the double-underline focus style that's consistent with links. 
image

## 💁 Proposed Solution

Implement the [latest design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=10133-130109) as spec'd in #2129

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2163

## 👩‍💻 Implementation

- `prominent` appearance now gives the current/last breadcrumb item an accent color (rather than all _but_ the last)
- breadcrumb item now uses same focus styling as anchor, i.e. double underline and accent color (rather than a double outline)
- breadcrumb item with missing/empty `href` is now the same color as a disabled link (rather than body font color)
- breadcrumb item now, like the anchor, remains underlined when clicking it (rather than the underline disappearing)
- spacing between breadcrumb item text and separators ('/') increased from 4px to 8px
- I did not want to include both `hover` and `hoverActive` in the breadcrumb interactions story, so I changed `createMatrixInteractionsFromStates` to handle empty arrays for any of the states.

## 🧪 Testing

Added visual tests for no-`href` states, as well as interaction states.

## ✅ Checklist


packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -1,81 +1,59 @@
...
+    linkDisabledFontColor,
+    mediumPadding,
     placeholderFontColor
 } from '../theme-provider/design-tokens';
-import { focusVisible } from '../utilities/style/focus';
+import { styles as anchorStyles } from '../anchor/styles';
 
 export const styles = css`
-    ${display('inline-flex')}
+    ${anchorStyles}
```
m-akinc, line 13: I considered forking the breadcrumb item template and changing it to use a `nimble-anchor` rather than a native anchor, but just reusing the anchor's styles does about the same thing without introducing extra elements in the DOM.

Most of the styles I removed from this file were redundant with the anchor styles.

packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -1,81 +1,59 @@
...
+        }
 
-    [part='end'] {
-        display: none;
-    }
+        :host(:not([href])) slot:not([name]),
+        :host([href='']) slot:not([name]) {
+            display: block;
+            margin-right: ${mediumPadding};
+        }
```
m-akinc, line 46: When there's no `href', there's no anchor (`.control`) element -- just the default slot for the text. We have to apply the `margin-right` that would have been on the anchor element.

packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -1,81 +1,59 @@
...
 
-    .content {
-        pointer-events: none;
-    }
+        .control {
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            margin-right: ${mediumPadding};
```
m-akinc, line 38: We now use `box-shadow` to draw the focus underline (comes from the anchor styles). I needed to change this from `padding-right` to `margin-right` so that the `box-shadow` wouldn't extend past the end of the text.

m-akinc:
@msmithNI would you mind buddying this?

msmithNI (approved):

packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -1,81 +1,59 @@
...
+            text-decoration: none;
+        }
 
-    [part='end'] {
-        display: none;
-    }
+        :host(:not([href])) slot:not([name]),
```
jattasNI, line 42: From our [CSS guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes)
>If you find yourself in complex logic with lots of :not() selectors it's possible the code should be reorganized to leverage the CSS cascade for overriding states.

I'm not saying this is overly complex, but it's still preferable to avoid `not` if practical. Could you take a stab at using the cascade instead or comment what makes that difficult?

jattasNI (approved):

packages/nimble-components/src/breadcrumb-item/styles.ts
m-akinc: There are two kinds of `:not()` usages here:
1. `slot:not([name])` - Used to style the default slot. Since breadcrumb item also has `start` and `end` slots, we have to disambiguate, and this is the most direct way to select just the default slot. 
2. `:host(:not([href]))` - Used to select when the breadcrumb item has no `href` set. Again, this is the most direct way to represent that.

If we use the cascade, we end up with more lines of CSS, which are more roundabout and harder to understand:
```css
slot {
    display: block;
    margin-right: ${mediumPadding};
}

:host([href]) slot,
slot[name="start"],
slot[name="end"] {
    display: unset;
    margin-right: unset;
}

:host([href='']) slot {
    display: block;
    margin-right: ${mediumPadding};
}
```
We can't combine the first and third rule blocks, because `:host([href]) slot` and `:host([href='']) slot` have the same specificity. That means order dictates precedence, so we have to make sure the `:host([href='']) slot` rule comes later. The fact we have to care about specificities (and I had to use an online calculator to check them) is another argument that this is more complex.

If I'm being a dope and there's a simpler way to write this, or there's an argument that the version without `:not()` is better, I'm open to hearing it. I also wish I could use `:not()` without spending extra time justifying it, but I guess that's a necessary evil.

rajsite: Some duplication is pretty much expected with the style conventions. We group by DOM order, not by the sets of configured properties, so start and end selectors get split up and often (not always) have some duplicated styles.

The complexity the style guide rules push to optimize for is picking an element in the DOM / template, jumping to the CSS, and seeing exactly what styles are expected on that element given the state of the page.

Stuff like the `slot:not([name])` makes that hard because one potentially has to jump to multiple styles to understand what applies to an element. By that understanding of complexity, low complexity is seeing `slot[name="start"]` applies to the "start slot" and high complexity is having to understand `slot:not([name])` is used to style the "start slot".

My gut tells me we can align with our CSS guidelines better. I'll take a stab at looking at the styles in a separate branch but fyi might be a couple days (don't think there is a specific urgency here). May be a good way to capture additional style guideline clarifications as well.

rajsite: > high complexity is having to understand `slot:not([name])` is used to style the "start slot".

Actually I think I even messed that up. `slot:not([name])` is to target a slot without a name, so the default slot 😂

packages/nimble-components/src/breadcrumb-item/styles.ts
rajsite: Scanning quickly it also looks like there are styles in the anchor pattern that might not be relevant to breadcrumb. So how we want to avoid duplicated / overrides in shared styles we should also avoid dead styles / styles that don't apply.

In the other comment I stated I'll take a look at the styles in a separate branch so I'll take a pass at evaluating the dead styles as well.

packages/nimble-components/src/breadcrumb-item/styles.ts
m-akinc: Yeah, I figured you meant "default slot". 😊

> We group by DOM order, not by the sets of configured properties

Good point. I hadn't followed that convention in my earlier reply, though since we're setting `display: none` on the `start` and `end` parts (in the anchor styles), we can leave out the styling for those slots. But there _is_ a `separator` slot I hadn't accounted for.

So if we avoid `:not()`, we actually end up with this:

```css
/* Since this does not refer to a specific element in the template, I guess it should go where the first instance of this element type occurs? */
slot { 
    display: block;
    margin-right: ${mediumPadding};
}

:host([href]) slot {
    display: unset;
    margin-right: unset;
}

:host([href='']) slot {
    display: block;
    margin-right: ${mediumPadding};
}

slot[name='separator'] {
    ...
    margin-right: unset;
}
```

For comparison, using `:not()`, we have:

```css
:host(:not([href])) slot:not([name]),
:host([href='']) slot:not([name]) {
    display: block;
    margin-right: ${mediumPadding};
}
/* no change needed to separator slot styling */
```

> The complexity the style guide rules push to optimize for is picking an element in the DOM / template, jumping to the CSS, and seeing exactly what styles are expected on that element given the state of the page. Stuff like the `slot:not([name])` makes that hard because one potentially has to jump to multiple styles to understand what applies to an element.

True, but importing shared styles (from the patterns directory) has that problem to an even greater extent, since you have to scan the entirety of another file. It's obviously a trade-off.

I agree that using `:not()` _can_ lead to the kind of complexity you're talking about, but in this particular case, since `slot:not([name])` refers to exactly one element in the template, it's not actually any more complex than a selector like `slot[name="start"]` (for example). If it's not obvious that `slot:not([name])` means "default slot", we could alway s add a comment.

packages/nimble-components/src/breadcrumb-item/styles.ts
m-akinc: As we discussed, I am no longer sharing styles with the anchor. I'm just copying the relevant parts.

packages/nimble-components/src/breadcrumb-item/styles.ts
m-akinc: Moot now. The padding is on the separator, which simplifies a lot.

packages/nimble-components/src/breadcrumb-item/styles.ts
m-akinc: It turns out all of this was essentially moot, because I can just move the padding to the separator rather than conditionally adding left/right padding to the link (or text) elements. 🤦‍♂️

packages/nimble-components/src/breadcrumb-item/styles.ts
```diff
@@ -1,81 +1,101 @@
...
-        text-decoration: none;
-    }
+        [part='end'] {
+            display: none;
+        }
+
+        .separator {
+            padding-left: ${mediumPadding};
+            padding-right: ${mediumPadding};
+        }
```
m-akinc, line 59: I'm not sure why we didn't go this route initially, but this seems way simpler and avoids the whole issue with styling the default slot, as well as the need to remove item padding via the parent breadcrumb's styles.

rajsite (approved):

Issue #787, opened 2022-10-18
Visual design for graph parts


jattasNI:
Sometime in Q4 a team will be incorporating a plotly graph into SLE. We would like it to use colors/fonts/spacing that aligns with other Nimble components.

This could be a lightweight pass to tell them which tokens to use for:
1. background color
2. text color, font, and size (axis and tick labels)
3. tick and grid line color and line thickness
5. plot colors
   - in this case the spec shows dots for each point but eventually clients will come along who want to connect them with lines
   - in this case there might only be one or two plots but we should provide a palette of default colors for many plots (a dozen?)
5. other graph parts like toolbars and legends
6. tooltips showing more information when hovering over a point

![](https://user-images.githubusercontent.com/10500124/196209932-c68aa72b-695a-4c9b-9a18-c50a950e3bd2.png)

jattasNI:
Another screenshot from an earlier visual design research project.
<img width="914" alt="MicrosoftTeams-image" src="https://user-images.githubusercontent.com/10500124/198382974-48092fe8-1847-46ea-b8a2-786a25e2deba.png">

jattasNI:
Brandon and Rick spoke about this at our team meeting today. Rick has some prior art for plot colors--when the time comes we can ask him for those values. Otherwise when this becomes a priority someone will have to collect values from the comps above and turn them into design tokens, filling in whatever gaps we find along the way.

rajsite:
Some discussion on creating plotly templates to implement theming: https://www.nelsontang.com/blog/2021-08-01-build-your-first-plotly-template

jattasNI:
Rick noted that there is no design for the color theme because it's impossible to pick plot colors that would work against every possible background color. For now we can just document that it's unsupported and use some placeholder colors (since the helpers in `design-tokens.ts` require that we provide colors for every theme).

rajsite:
> Rick noted that there is no design for the color theme because it's impossible to pick plot colors that would work against every possible background color. For now we can just document that it's unsupported and use some placeholder colors (since the helpers in `design-tokens.ts` require that we provide colors for every theme).

For future considerations of color theme maybe we should think of it like monochrome printing. Would use different line styles and maybe picking the order of line styles could come from nimble.

kroeschl:
Would it be possible to get tokens for graph colors beyond the existing `$ni-nimble-graph-gridline-color` token? Specifically, I'm looking for specific colors to use for e.g. different data series or different bars in a bar graph.

[...]

PR #2194, opened 2024-06-14
Define graph trace color tokens

m-akinc:
# Pull Request

## 🤨 Rationale

Addresses #787 

## 👩‍💻 Implementation

Added 8 color tokens for graph traces, e.g. `graphTrace1Color`.

Per the design, we cannot provide graph trace tokens that work generally for monochromatic applications. If the use case only requires a single trace, then it seems correct for that trace (`graphTrace1Color`) to be white for the color theme. If there are more traces, then based on what Rick wrote in his design, the client presumably has to override the token values based on the background color. So defaulting all the traces to white seems reasonable.

## 🧪 Testing

Token story

## ✅ Checklist


change/@ni-nimble-components-612b4007-c1f6-4b64-92c7-3fb0494c1002.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Is it just my eyes or do these colors look really similar? 😆

I'm not sure what the issue is; the theme-aware tokens look normal and the base colors in nimble-tokens look fine too. But the incorrect colors show up in all themes locally and in the Chromatic diff.

https://60e89457a987cf003efc0a5b-xuuatvbxpn.chromatic.com/?path=/docs/tokens-theme-aware-tokens--docs

image

jattasNI (changes requested):

change/@ni-nimble-components-612b4007-c1f6-4b64-92c7-3fb0494c1002.json
m-akinc: Whoops. Copy-paste error.

packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -204,6 +218,38 @@ export const graphGridlineColor = DesignToken.create<string>(
     styleNameFromTokenName(tokenNames.graphGridlineColor)
 ).withDefault((element: HTMLElement) => hexToRgbaCssColor(getColorForTheme(element, Black91, Black15, White), 0.2));
 
+export const graphTrace1Color = DesignToken.create<string>(
```
rajsite, line 221: Was not described in the issue, comments, and no prior discussion so guess we need to discuss in owner's PR (which I'd prefer not to do @m-akinc)?

I'd prefer the color theme to not be just completely broken. Using White for all the traces is just completely unusable. Maybe we should just pick either the light or dark colors as a temporary solution? And then we should capture usage guidance somewhere that says Color isn't fully supported yet. Thoughts @jattasNI @m-akinc

m-akinc: > Was not described in the issue, comments, and no prior discussion so guess we need to discuss in owner's PR (which I'd prefer not to do @m-akinc)?

I didn't think there was a need for discussion. Per the design, we cannot provide graph trace tokens that work generally for monochromatic applications:
image
If the use case only requires a single trace, then it seems correct for that trace (`graphTrace1Color`) to be white for the color theme. If there are more traces, then based on what Rick wrote in his design, the client presumably has to override the token values based on the background color. So defaulting all the traces to white seems reasonable. I'll add this rationale to the PR description.

rajsite: Our general idea is that even for unsupported configurations we don't completely fail and we give clear usage guidance. This is a complete failure where if we choose any color palette a user might be able to partially use the control in a poorly configured app this is just completely unusable.

> So defaulting all the traces to white seems reasonable.

I disagree with that assessment. 

>>  And then we should capture usage guidance somewhere that says Color isn't fully supported yet.
> I'll add this rationale to the PR description.

That concern is still not addressed.

rajsite: Another option: Use 8 different shades of white with trace 1 being full white. That way the I guess expected trace 1 color of white is still full white and there is some differentiation between traces 2-8 even if not great. And then documenting usage guidance.

Maybe we should have a Plot mdx that describes what tokens we have available for styling a plot and can also capture usage guidance.

m-akinc: Ok, I changed the color theme defaults to various opacities of white (using grayscale shades at full opacity looked odd). I also added a "Graph Styling" mdx that lists the graph-related tokens and gives usage guidance about overriding trace colors for the color theme.

packages/storybook/src/nimble/theme-provider/tokens.ts
```diff
@@ -0,0 +1,185 @@
...
+): ViewTemplate<TokenName> => {
+    const suffix = suffixFromTokenName(tokenName);
+    if (suffix === undefined) {
+        throw new Error(`Cannot identify suffix for token: ${tokenName}`);
+    }
+    const template = tokenTemplates[suffix];
+    return template;
+};
+
+export function getTokensStory(tokenNames: TokenName[]): StoryObj<TokenArgs> {
```
rajsite, line 114: storybook has a well established pattern for parameterizing stories, can we use that instead of making a new pattern?

packages/storybook/src/nimble/theme-provider/graphs.stories.ts
```diff
@@ -0,0 +1,29 @@
...
+    'graphGridlineColor',
+    'graphTrace1Color',
+    'graphTrace2Color',
+    'graphTrace3Color',
+    'graphTrace4Color',
+    'graphTrace5Color',
+    'graphTrace6Color',
+    'graphTrace7Color',
+    'graphTrace8Color'
+] as TokenName[];
```
rajsite, line 14: didn't dive deep but bet there is a way to avoid the `as` cast. please look into it

packages/storybook/src/nimble/theme-provider/tokens.stories.ts
```diff
@@ -1,198 +1,18 @@
...
-import { comments } from '../../../../nimble-components/src/theme-provider/design-token-comments';
-
-import {
-    bodyFont,
-    bodyFontColor,
-    groupHeaderFont,
-    groupHeaderFontColor,
-    groupHeaderTextTransform
-} from '../../../../nimble-components/src/theme-provider/design-tokens';
+import type { Meta } from '@storybook/html';
```
rajsite, line 1: chatted offline about parameterizing the stories based more on a normal story pattern instead of matrix story patterns, ie how appearance modes or status table is parameterized, etc.

packages/storybook/src/utilities/typed-object-keys.ts
```diff
@@ -0,0 +1,5 @@
+// Based on https://fettblog.eu/typescript-better-object-keys/
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface ObjectConstructor {
+    keys<T extends object>(o: T): (keyof T)[];
+}
```
rajsite, line 5: It's hard to rate these things but a solution that overrides how the keys property is defined globally is a little more of a hack than a single `as` cast. I think it's still worth trying to find an alternative to `as` casts but this is probably not the right direction to jump to first as a solution.

m-akinc: I don't think there's much else we can do. I did a fair bit of digging, and unless there's some alternative to `Object.keys()`, I think that's it.
https://alexharri.com/blog/typescript-structural-typing
https://fettblog.eu/typescript-better-object-keys/
If you search the repo for `Object.keys`, you'll see we are casting the result in plenty of other places.

I've reverted back to the `as` cast per your preference.

rajsite: Sounds reasonable, thanks for looking into it

rajsite (approved):

jattasNI (approved):

PR #2196, opened 2024-06-17
Remove extra whitespace from mapping docs

jattasNI:
# Pull Request

## 🤨 Rationale

https://nimble.ni.dev/storybook/?path=/docs/components-table-column-mapping--docs

image

## 👩‍💻 Implementation

Change to use the table element's name instead of its tag. We use plain text element names and code-formatted element tags pretty interchangeably throughout our docs.

I would mildly prefer to keep using the`<Tag>` but Prettier is insisting on putting an extra newline before it if it's the first thing on a line. I played with some [configuration options](https://prettier.io/docs/en/options.html) and couldn't find one to change that behavior.

Other options include moving this string to the `.stories.ts` file, adding inline prettier ignores, or debugging further. I went with something simple for a relatively minor bug that I didn't want to spend much time on unless it bothers us in more places, but if this makes you go "ick" I can reconsider.

## 🧪 Testing

In a local build the whitespace is gone

## ✅ Checklist


jattasNI:
Moved back to draft due to build failures

rajsite (approved):

PR #2197, opened 2024-06-17
Menu button page object

mollykreis:
# Pull Request

## 🤨 Rationale

Create menu-button page object that is exported from nimble-components and from nimble-angular.

## 👩‍💻 Implementation

- Create page object for the menu-button
- Create new `menu-button/testing` entry point from nimble-angular and export the page object from there
- Update menu button unit tests in nimble-components to use the page object where necessary
- Add `waitForEventAsync` utility function so that callers can wait for an event to be fired without having to have a dependency on `jasmine`

## 🧪 Testing

Unit tests still pass

## ✅ Checklist


mollykreis:

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,84 @@
...
+        );
+        this.clickMenuButton();
+        await toggleListener.promise;
+    }
+
+    public pressEnter(): void {
+        const event = new KeyboardEvent('keypress', {
+            key: keyEnter
+        } as KeyboardEventInit);
+        this.menuButtonElement.toggleButton!.control.dispatchEvent(event);
```
m-akinc, line 46: It's not a huge deal, but if keeping all these individual keypress funtions, I'd create a protected `sendKeypressEventToToggleButton` helper and turn all these `press___` functions into one-liners.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
```diff
@@ -190,14 +186,14 @@ describe('MenuButton', () => {
...
+            pageObject.clickMenuButton();
             expect(element.open).toBeFalse();
         });
 
         it('should close menu when toggle button is clicked while the menu is open', async () => {
             element.open = true;
             await connect();
-            element.toggleButton!.control.click();
+            pageObject.clickMenuButton();
             expect(element.open).toBeFalse();
```
m-akinc, line 197: Since they're in the page object API, we should presumably be using those instead:
```ts
    await pageObject.openMenu();
    pageObject.clickMenuButton();
    expect(pageObject.isOpen()).toBeFalse();
```

This applies to other places in this file, too.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
```diff
@@ -384,11 +367,12 @@ describe('MenuButton', () => {
             it('should open the menu and focus first menu item when the toggle button is clicked', async () => {
                 await connect();
                 const menuButton = value.getMenuButton(element);
+                const pageObject = new MenuButtonPageObject(menuButton);
```
m-akinc, line 370: Seems like all the tests (in this diff, at least) are repeating these steps. Can we put them in a `beforeEach`?

m-akinc (approved):

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,84 @@
...
+            throw new Error('Cannot close menu when it is not open');
+        }
+
+        const event = new KeyboardEvent('keydown', {
+            key: keyEscape
+        } as KeyboardEventInit);
+        this.menuButtonElement.region!.dispatchEvent(event);
+    }
+
+    public getTextContent(): string {
```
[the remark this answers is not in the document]
mollykreis: Do you think it makes sense to call this a "label"? It's the content of a button, not really a label.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: As a note, the implementations of these functions aren't all quite the same, which is why I decided to keep them separate. The enter & space keys get sent to the toggle button's _control_ and the up & down arrow keys get sent to the toggle button itself.

But, I'll take a pass at factoring out common logic to see if that makes the code more readable.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        const toggleListener = createEventListener(
+            this.menuButtonElement,
+            'toggle'
+        );
+        this.clickMenuButton();
+        await toggleListener.promise;
+    }
+
+    public pressEnter(): void {
```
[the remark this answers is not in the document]
mollykreis: It probably wasn't clear, so I'll update the docs comment for the page object ... the page object is really intended to be for interacting with the menu button itself (not the menu items). I guess the one exception is the function for pressing ESC on the menu, though technically, that's not a key press on the menu itself.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: Update: As I was looking at this, I actually remembered that there is another difference -- some events need to be `keypress` and others need to be `keydown`. Given the nuances of firing the events, I'm very happy to finally encapsulate the logic within the page object. However, I don't think there is a lot to gain by creating a helper function to create & dispatch the event.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
mollykreis: I could update the tests so that none (or very few) of them programmatically open/close the menu button setting `element.open` explicitly, but I'm not sure I see the advantage of that. Could you elaborate on why you think it would be better to use `await pageObject.openMenu()` rather than setting `element.open = true`?

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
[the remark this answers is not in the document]
m-akinc: I believe so. I checked the docs for [Material](https://m3.material.io/components/buttons/guidelines#9ecffdb3-ef29-47e7-8d5d-f78b404fcafe), [Fluent](https://fluent2.microsoft.design/components/web/react/button/usage), and [Carbon](https://carbondesignsystem.com/components/button/usage/) design systems, and they all refer to the text on a button as the "label", too.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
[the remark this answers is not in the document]
m-akinc: Yeah, I would expect a menu button page object to encompass all the interactions for the menu button, meaning both the button and the menu. Otherwise, wouldn't it just be a toggle button page object?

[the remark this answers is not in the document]
mollykreis: I think even without menu interactions, this is more than just a toggle button page object. Specifically, this includes:
- all the keypress/keydown code that needs to be called on the right part of the shadowroot
- having an async function that knows how to wait for the menu to fully open before returning

Perhaps if we had a page object for the `nimble-menu`, `nimble-menu-item`, and/or `nimble-anchor-menu-item`, we could expose a reference to those page objects here. However, I think a key thing to keep in mind is that the menu and its items are being slotted into the menu button by the client app. As a result, a client test can interact with those items directly in tests without having to make any assumptions about the internal workings or structure of the menu button.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
m-akinc: My thinking was that you've created a page object function that abstracts away the details of opening the menu, so that would be the preferred way to open it, but on second thought, `open` is part of the menu button API, so I guess that's just as valid. If that's the case, then the primary purpose of the page object function must be to allow waiting for the menu to actually be open. It's not clear to me why we sometimes need that, and other times can just call `open = true`. 

Also, I noticed that the page object's `openMenu()` opens via click. Should it instead use `open = true`, since that's what the old definitions of `openMenu()` were doing? If it is relevant that it opens via a click, maybe it should be named `clickMenuButtonAndWait()`.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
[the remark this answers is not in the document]
m-akinc: Good points!

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
mollykreis: My thoughts are:
- The page object should open the menu in an interactive way (through click or key events) since presumably almost any client test is trying to test how a user would interact with the page rather than via programmatic means
- In the end, a client calling `pageObject.openMenu()` doesn't care how the menu is opened. If they do care, they should call some other function such as `clickMenuButton` or `pressEnterOnMenuButton` so that they are explicitly controlling the precise interaction.
- These tests can be a little more efficient by setting `open = true` before calling `await connect()` with no further need to wait for the menu to render as open. Otherwise, they'd need to `await connect()` and then `await pageObject.openMenu()`.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        const toggleListener = createEventListener(
+            this.menuButtonElement,
+            'toggle'
+        );
+        this.clickMenuButton();
+        await toggleListener.promise;
+    }
+
+    public pressEnterOnMenuButton(): void {
```
atmgrifter00, line 45: Do we need the `OnMenuButton` suffix for these methods? I ask only because a semblance of a naming pattern has been established on the `SelectPageObject` for these kinds of actions, and they don't include a similar suffix (they're all `press<key>Key`).

atmgrifter00 (approved):

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: That's an interesting question. I added that suffix based on feedback from @m-akinc that it wasn't clear whether some of these functions were interacting with the button or menu/menu-items.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+
+        const toggleListener = createEventListener(
+            this.menuButtonElement,
+            'toggle'
+        );
+        this.clickMenuButton();
+        await toggleListener.promise;
+    }
```
jattasNI, line 43: I could see clients being unsure which of these they should use if their goal is just to open a menu to click an option. Could we provide some guidance via comments?

I'm guessing that guidance would be to prefer `openMenu()` since it [waits for the `toggle` event which implies the menu is actually visible](https://github.com/ni/nimble/blob/b3cf1500163c84dd7caee3ed350c36afd2979d1b/packages/nimble-components/src/menu-button/index.ts#L145). I can't think of a common use case for clients to want to use the click / keyboard methods so maybe we should discourage their use saying something like "only use this for fine-grained control over event timing".

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
rajsite: I think our page objects should align with our composition. So I'd expect different page objects for the menu-button, menu, and potentially menu items. And with references to those instances I'd expect the disambiguation at the reference naming vs in the api, i.e. `menuButtonPageObject.open();`, `menuPageObject.pressArrowDown();`, `menuItemPageObject.click()`. I could maybe see convenience apis on menu for interacting with menu items but from menu button all the way to menu items seems too far?

jattasNI (approved):

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: That's a good point. We could also make some of these functions internal if we don't think clients will have a reason to use some of them. I think that's a pattern we've done in other page objects, right?

rajsite: Other thoughts:
- We also did a thing where we made an internal pageObject that we didn't export. We could move some of this to an internal page object
- The Web Crypto api and the Signals api both have a sub reference called `subtle` to put things that aren't expected to be generally used. We could consider that pattern to make those operations still available outside nimble but indicate they aren't expected to be used often.

```ts
class MenuButtonPageObject {
    public constructor(
        protected readonly menuButtonElement: MenuButton
    ) {
        this.subtle = new MenuButtonSubtlePageObject(menuButtonElement);
    }
}
```

jattasNI: Yeah, for [select we have a separate internal page object](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/select/tests/select.pageobject.internal.ts) and for other page objects we've just added `@internal` comments. I'm fine with either approach if you can form a sense of which methods would be useful to clients.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    public closeMenuWithEscape(): void {
+        if (!this.isOpen()) {
+            throw new Error('Cannot close menu when it is not open');
+        }
+
+        const event = new KeyboardEvent('keydown', {
+            key: keyEscape
+        } as KeyboardEventInit);
```
rajsite, line 80: Could we have a helper for creating the KeyBoard event options or the whole KeyboardEvent to minimize the amount of `as` casts needed?

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: I'm leaning towards marking some functions as `@internal`. Upon further thought, I think creating a separate page object will likely be convoluted for cases like the menu button column page object where we want to expose the menu button page object as part of the API but also internally call internal functions on the menu button's page object.

rajsite: I suspect that using `subtle` wouldn't be an issue, just have circular references between them. could even keep the classes in the same file, but also don't feel the need to make a new pattern for this PR.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
m-akinc: Maybe a comment on those functions would be enough to make it clear that, say, if you open the menu and call `pressArrowDown()`, it's being pressed on the toggle button and not the menu. Both of these are parts of the menu button component, so it wasn't totally clear to me. Alternatively, if the page object had a `menuPageObject` property, that might also make it clearer that _that_ was the API for interacting with the menu.

mollykreis: Added short comment for each function and renamed the key press functions to be consistent with the select page object.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: Turns out that those casts weren't actually needed, so I've removed them.

packages/nimble-components/src/menu-button/testing/menu-button.pageobject.ts
mollykreis: I decided to mark quite a few functions as `@internal`.

rajsite (approved):

PR #2198, opened 2024-06-17
List option in group needs to register with Select

atmgrifter00:
# Pull Request

## 🤨 Rationale

Had a realization that list options that were in a group would not go through the `registerOption` routine (a fix for SLE Angular apps). While I'm not sure if there is any current use-case that would hit the problem, I went ahead and made the fix and the tests that demonstrate the problem.

## 👩‍💻 Implementation

Just made the private `getListOptionOwner` routine  on `ListOption` a bit more robust.

## 🧪 Testing

Wrote component and Angular tests modeled after existing tests but with a group as the option owner.

## ✅ Checklist

### DIFF
--- a/packages/angular-workspace/nimble-angular/src/directives/select/tests/nimble-select-control-value-accessor.directive.spec.ts
+++ b/packages/angular-workspace/nimble-angular/src/directives/select/tests/nimble-select-control-value-accessor.directive.spec.ts
@@ -4,6 +4,7 @@ import { AbstractControl, FormControl, FormGroup, FormsModule, ReactiveFormsModu
 import { SelectPageObject } from '@ni/nimble-angular/select/testing';
 import { NimbleSelectModule } from '../nimble-select.module';
 import { NimbleListOptionModule } from '../../list-option/nimble-list-option.module';
+import { NimbleListOptionGroupModule } from '../../list-option-group/nimble-list-option-group.module';
 import { processUpdates, waitForUpdatesAsync } from '../../../testing/async-helpers';
 import type { Select } from '../nimble-select.directive';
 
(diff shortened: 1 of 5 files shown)


packages/nimble-components/src/list-option/index.ts
```diff
@@ -66,9 +66,22 @@ export class ListOption extends FoundationListboxOption {
...
+
+        while (parent) {
+            if (this.isListOptionOwner(parent)) {
+                return parent;
+            }
+
+            parent = parent.parentElement;
         }
+
+        return null;
```
rajsite, line 84: prefer undefined over null

rajsite (approved):

packages/angular-workspace/nimble-angular/src/directives/select/tests/nimble-select-control-value-accessor.directive.spec.ts
```diff
@@ -336,4 +337,158 @@ describe('Nimble select control value accessor', () => {
...
+                    </nimble-select>
+                </form>
+             `
+        })
+        class TestHostComponent implements OnChanges {
+            @ViewChild('select', { static: true }) public select: ElementRef<Select>;
+
+            @Input() public selectValue?: { name: string, value: number };
+
+            public selectOptions: { name: string, value: number }[] = [
```
rajsite, line 432: prefer as const

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -500,6 +500,35 @@ describe('Select', () => {
...
+    it('option added directly to DOM under group synchronously registers with Select', async () => {
+        const { element, connect, disconnect } = await setupWithGroups();
+        await connect();
+        await waitForUpdatesAsync();
+        const newOption = new ListOption('foo', 'foo');
+        const registerOptionSpy = spyOn(
+            element,
+            'registerOption'
+        ).and.callThrough();
+        registerOptionSpy.calls.reset();
```
rajsite, line 512: do you need to reset a freshly made spy?

jattasNI (approved):

packages/angular-workspace/nimble-angular/src/directives/select/tests/nimble-select-control-value-accessor.directive.spec.ts
atmgrifter00: I modify the array below, so it doesn't allow `as const`. I've removed the redundant typing declaration however.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: Nope.

PR #2201, opened 2024-06-18
Update select page object

atmgrifter00:
# Pull Request

## 🤨 Rationale

Recognized that as part of attempting to migrate an SLE component to the `Select` to use the grouping feature that there was a useful pageObject API for retrieving the options for a group by providing the group's label. This PR adds that functionality to the `SelectPageObject`.

## 👩‍💻 Implementation

Simple implementation mirroring a similar existing one.

## 🧪 Testing

Added test for the new pageObject method.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-1d7d08c5-fc96-4796-bce3-650510487895.json
+++ b/change/@ni-nimble-components-1d7d08c5-fc96-4796-bce3-650510487895.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Adding new SelectPageObject API.",
+  "packageName": "@ni/nimble-components",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


rajsite (approved):

change/@ni-nimble-components-1d7d08c5-fc96-4796-bce3-650510487895.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
rajsite, line 2: @atmgrifter00 @jattasNI I can't remember, do we treat breaking changes on page objects as breaking api semver releases?

jattasNI: Yeah, we should treat it as part of our public API. So mark this `major` and add a note to the `comment` saying the old and new name of the method.

If we want to fudge it we could claim that the whole grouping feature is still "internal" / "incubating" since it's only been around for a couple weeks and no one is using it yet. But that'd be kind of a made up policy and there's no harm in major bumps so I say we do a major bump.

jattasNI (approved):

atmgrifter00:
Dang it! I thought I had pushed the changes to the change file, but didn't notice the failure before I squashed the changes.

NOTICE: This was technically a breaking change, as the `SelectPageObject` method `getGroupOptionLabels` was renamed to `getGroupOptionLabelsByIndex`.

PR #2172, opened 2024-06-06
Add keyboard navigation to table

msmithNI:
# Pull Request

## 🤨 Rationale

Resolves #1137 

## 👩‍💻 Implementation

See #2173 for the most up-to-date expected behavior of keyboard navigation that this PR implements. ([Direct link to updated HLD](https://github.com/ni/nimble/blob/96e1725fc7cc8b3fdf5f5197b090e8f95f86cd5f/packages/nimble-components/src/table/specs/table-keyboard-navigation-hld.md))

The `KeyboardNavigationManager` class does the bulk of the work:
- Tracks what is currently focused / should be focused (via properties `focusType`, and `row/column/cellContent/headerActionIndex`)
   - Note: If the user clicks in the table away from the current keyboard focus, the focus state would end up out-of-date. There's code in `handleFocus()` to handle this (see additional comments there)
- Listens for table keypresses to handle arrow key navigation, Tab/Shift-Tab, PgUp/PgDn/etc key presses
- If the user unfocuses and refocuses the table, re-focuses the appropriate elements in the table (`handleFocus()` again)
- Tab/Shift-Tab behavior should match the HLD / expectations for ARIA treegrid. (Basically, Tab/Shift-Tab go through the tabbable elements of the header row / current row, not including cells/column headers, until the end is reached, then the table blurs and elements before/after it will be focused with additional Tab presses.)

The overall approach of only setting `tabindex=0` on the single element in the table we want focused is called [roving tabindex](https://www.freecodecamp.org/news/html-roving-tabindex-attribute-explained-with-examples/).

To ensure that we control what elements in the table should get focused, this PR updates `tabindex` on many elements in the table:
- The table itself is `tabindex=0` to ensure that it can be focused via `Tab`
- Elements that are focusable by default such as buttons/anchors, are explicitly set to `tabindex=-1` to start with, until the keyboard navigation code wants to focus them (then they're set to `tabindex=0` just before being focused).

## 🧪 Testing

- Manual testing (mostly Chrome, some Firefox).
   - Earlier builds had some Safari testing, but we should explicitly re-test there before submission.
- Added autotests

Keyboard navigation can be used on any/all of the [existing table Storybook pages](https://60e89457a987cf003efc0a5b-gjoayfsquj.chromatic.com/?path=/story/components-table--table), once the table is focused.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.
[...]

PR #2204, opened 2024-06-20
Menu button column

mollykreis:
# Pull Request

## 🤨 Rationale

Create menu-button table column.

## 👩‍💻 Implementation

Create `nimble-table-column-menu-button` that creates a menu button in a column and allows slotting a menu into the table to be used with the menu buttons that are created in the column.

This is the first column that hasn't allowed grouping, so I also made a few changes to facilitate not needing to specify a group header view on a column.

## 🧪 Testing

- Manually tested
- Created unit tests for the column
- Created storybook page for the column
- Created matrix tests for the column, including tests with the menu being open

## ✅ Checklist


packages/nimble-components/src/table-column/menu-button/cell-view/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    public menuButton?: MenuButton;
+
+    /** @internal */
+    public valueSpan?: HTMLSpanElement;
+
+    /** @internal */
+    @observable
+    public hasOverflow = false;
+
+    // TODO: Override tabbableChildren getter
```
mollykreis, line 38: `tabbableChildren` will need to be implemented here once #2172 is merged.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
```diff
@@ -0,0 +1,23 @@
...
+        width: 100%;
+    }
+
+    .value-label {
+        margin-right: auto;
+        overflow-x: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        height: 100%;
+        align-content: center;
```
mollykreis, line 21: This `align-content: center` styling does not seem to be having any effect in chromatic, but I'm not sure why. The result is that the text is too close to the top of the menu button rather than being centered. The text alignment within the button looks correct on my machine in chromium, FF, and webkit. I'm open to suggestions of how to resolve the chromatic issue.

mollykreis:

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button-opened-matrix.stories.ts
```diff
@@ -0,0 +1,85 @@
+import type { StoryFn, Meta } from '@storybook/html';
+import { html } from '@microsoft/fast-element';
+import { waitForUpdatesAsync } from '@ni/nimble-components/src/testing/async-helpers';
+import { menuTag } from '@ni/nimble-components/src/menu';
+import { menuItemTag } from '@ni/nimble-components/src/menu-item';
+import { TablePageObject } from '@ni/nimble-components/src/table/testing/table.pageobject';
```
rajsite, line 6: We shouldn't be mixing import path formats, i.e. some directly to paths on disk and some via this format. Lets stick with direct import paths because we would not expect to ship typescript `src` from nimble directly via the package.

Also can you add this format to the block list (`@ni/nimble-components/*` and `@ni/spright-components/*`):

https://github.com/ni/nimble/blob/7450464c52f8dabfcac1f0fbc963a91dc23d4c4d/packages/storybook/.eslintrc.js#L52-L57

rajsite, line 6: We shouldn't be mixing import path formats, i.e. some directly to paths on disk and some via this format. Lets stick with direct import paths because we would not expect to ship typescript `src` from nimble directly via the package.

Also can you add this format to the block list (`@ni/nimble-components/*` and `@ni/spright-components/*`):

https://github.com/ni/nimble/blob/7450464c52f8dabfcac1f0fbc963a91dc23d4c4d/packages/storybook/.eslintrc.js#L52-L57

packages/nimble-components/src/table-column/base/cell-view/template.ts
```diff
@@ -28,6 +29,15 @@ export const createCellViewTemplate = (
...
             class="cell-view"
         >
+            ${repeat(
+        y => y.column?.columnInternals.slotNames || [],
+        html<string, TableCell>`
+                    <slot
+                        name="${(x, c) => uniquifySlotNameForColumn(c.parent.column!, x)}"
+                        slot="${x => x}"
+                    ></slot>
+                `
```
rajsite, line 39: CAn we do a prettier disable or something to improve the formatting here?

packages/nimble-components/src/table/template.ts
```diff
@@ -172,6 +175,12 @@ export const template = html<Table>`
                                             </slot>
                                         `)}
                                     `)}
+                                    ${repeat((x, c) => ((c.parent as Table).tableData[x.index]?.id ? (c.parent as Table).slotsByRecordId[(c.parent as Table).tableData[x.index]!.id] || [] : []), html<SlotMetadata>`
```
rajsite, line 178: this is pretty hard to understand, maybe a named method helper would be a bit clearer?

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -105,13 +115,13 @@ export class ColumnInternals<
...
      */
-    public readonly groupHeaderViewTemplate: ViewTemplate<TableGroupRow>;
+    public readonly groupHeaderViewTemplate?: ViewTemplate<TableGroupRow>;
 
     /**
      * Whether or not this column can be used to group rows by
      */
     @observable
-    public groupingDisabled = false;
+    public groupingDisabled = true;
```
atmgrifter00, line 124: This seems significant to change the defaults of two properties. Is it not? If it is, what are the repercussions? I would presume this would have some amount of cascading implementation, but maybe not?

mollykreis: Given our mixin pattern, I think it makes sense to disable grouping & sorting by default and only enable those features on a column if the column wants them enabled. The `mixinGroupableColumnAPI` has `groupingDisabled` and `mixinSortableColumnAPI` has `sortingDisabled`, both of which are initialized to `false`, which in turn sets these `ColumnInternals` properties to `false`. Therefore, I don't expect this change to have an impact on existing columns using grouping or sorting; it would just make a column not using those mixins (such as the menu-button column) have those features disabled by default.

packages/nimble-components/src/table/template.ts
mollykreis: I found a few ways to simplify this statement significantly, but I haven't figured out how to make this a named method helper without losing the observability of things.

mollykreis: I talked with @atmgrifter00, and he helped me find a better way to manage this state. It's now part of the `tableData` array, so that simplified things greatly.

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -31,13 +31,18 @@ export interface ColumnInternalsOptions<
...
 
     /**
      * The names of events that should be delegated from the cell view to the column.
      */
     readonly delegatedEvents: readonly string[];
 
+    /**
+     * The names of slots that need to be forwarded into a cell.
+     */
+    readonly slotNames?: readonly string[];
```
rajsite, line 44: since this gets initialized in the constructor and is concretely always an array I think we can leave off the optional check

```suggestion
    readonly slotNames: readonly string[];
```

mollykreis: This is the property on `ColumnInternalsOptions`, not `ColumnInternals`, so it doesn't get initialized in the constructor. Making it optional means that columns that don't care about slots don't need to explicitly set it to be an empty array.

rajsite: oops, my bad, makes sense

atmgrifter00 (approved):

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -0,0 +1,187 @@
...
+            name: 'field-name',
+            description:
+                "Set this attribute to identify which field in the data record contains the visible text value for each cell's menu button in the column. The field values must be of type `string`.",
+            control: false,
+            table: { category: apiCategory.attributes }
+        },
+        menuSlot: {
+            name: 'menu-slot',
+            description:
+                "The name of the slot within the `nimble-table` instance where the menu associated with the column's menu button will be provided",
```
jattasNI, line 130:
```suggestion
                "The name of the slot within the `nimble-table` instance where the menu associated with the column's menu button will be provided",
```
```suggestion
                "The name of the slot within the `nimble-table` instance where the menu associated with the column's menu button will be provided.",
```

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -0,0 +1,187 @@
...
+            table: {
+                disable: true
+            }
+        },
+        menuRef: {
+            table: {
+                disable: true
+            }
+        },
+        headerContent: {
```
jattasNI, line 144: We don't typically document the default slot for header content on other column types since it's covered by the Table Column Configuration story. Is there any reason to diverge from that pattern for this column?

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -0,0 +1,187 @@
...
+        storyArgs: MenuButtonColumnTableArgs,
+        e: CustomEvent<MenuButtonColumnToggleEventDetail>
+    ) => void;
+    menuRef: Menu;
+    headerContent: string;
+    toggleEvent: never;
+    beforeToggleEvent: never;
+}
+
+export const menuButtonColumn: StoryObj<MenuButtonColumnTableArgs> = {
```
jattasNI, line 92: It feels weird to me to open the menu button and select a new value but not have the displayed value update. I suspect that most of our client use cases will want to update the display. If that's true, I propose that:
1. we update this code to call `setData` in response to an item being selected
2. we add documentation saying that clients are responsible for doing this. I think that doc would fit either as a short sentence in the overview section or as a longer section under a `## Usage` heading.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from '@microsoft/fast-element';
+import { controlSlimHeight } from '../../../theme-provider/design-tokens';
+
+export const styles = css`
```
jattasNI, line 4: I was expecting the menu to be the same width as the column. I'm not sure why; maybe because I'm thinking of it like a select? And that would probably not be desirable in all cases, for example if the column is narrow then we wouldn't want the menu to be so narrow that its content is clipped.

I'm curious if we thought about this in the visual design or if we're just getting the default behavior. Assuming it's easy to control the menu width, could you check with Fred/Brandon/the team on what the desired behavior is?

image

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -0,0 +1,187 @@
...
+    sharedTableActions,
+    sharedTableArgTypes,
+    sharedTableArgs
+} from '../base/table-column-stories-utils';
+import {
+    apiCategory,
+    createUserSelectedThemeStory
+} from '../../../utilities/storybook';
+
+const simpleData = [
```
jattasNI, line 21: What do you think about a couple changes to make it easier to experiment with interaction corner cases?
1. add more data so we can see what happens if an open menu is scrolled out of view
2. add an action menu so we can play with keyboard navigation and multiple popups
3. enable selection so we can see how it renders when the row is selected

I'm on the fence; these would more be for me as a tester and might complicate the user-facing documentation / code sample. Maybe I should just pull your branch and try those things myself, but I'm curious if you think others would find them useful.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from '@microsoft/fast-element';
```
jattasNI, line 1: What is the state of the visual design right now? Assuming it's not final, do we have an item on Brandon's backlog to provide one? If we think it might be a little while then I guess we could put the docs in the Incubating section initially but I don't feel strongly that's necessary.

Personally I find the bold text seems to overemphasize the column relative to the others.

jattasNI (commented):
Just initial storybook and interaction review

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
mollykreis: I didn't realize that it was an established pattern. It just seemed to me that it had been left off by accident on all the columns. I don't mind removing it from here if we think the behavior is documented sufficiently in a different story that applies to all columns.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
mollykreis: This is the default behavior of the menu button itself. I can discuss with Fred and Brandon to see if we're happy with this behavior (both in the column & the stand-alone button).

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
mollykreis: Great question. Last I heard, the visual design was in work. However, that was a few weeks ago. I need to follow up with Fred and Brandon.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
mollykreis: 1. I've added a lot more data to the table (thanks Copilot for generating it all for me 😄)
2. I don't mind adding an action menu, but I wasn't sure how helpful that would be generally for a documentation point of view
3. Similar to the action menu. I don't mind adding selection to the table. I was just trying to be consistent with our other column type stories that don't do anything with selection.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
mollykreis: I've added a `## Usage` heading that mentions that the client needs to handle the `menu-button-column-beforetoggle` event by updating the menu items and handle events on the menu to update the data.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
jattasNI: Wow, Copilot knows the names of Simpsons characters I'd never heard of!

The more I think about 2 and 3, the less I'm able to justify them for clients so I'm ok to leave them off in the name of consistency and simplicity.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -0,0 +1,348 @@
...
+    toggleEvent: never;
+    beforeToggleEvent: never;
+    currentData: TableRecord[];
+}
+
+export const menuButtonColumn: StoryObj<MenuButtonColumnTableArgs> = {
+    parameters: {},
+    // prettier-ignore
+    render: createUserSelectedThemeStory(html<MenuButtonColumnTableArgs>`
+        <${tableTag}
```
rajsite, line 238: In the docs view the menu button isn't able to pop over the content. In combobox / select / menubutton we fix that using this helper:
https://github.com/ni/nimble/blob/2d017ba9169c8d83fb72bf98d6a5f657bd6f6e8d/packages/storybook/src/nimble/combobox/combobox.stories.ts#L106

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
```diff
@@ -1245,7 +1245,7 @@ describe('Table keyboard navigation', () => {
             name: interactiveColumnName
```
jattasNI, line 1245: I thought when we wrote the keyboard navigation tests we had a vision of parameterizing them so they could easily cover new column types that had interesting keyboard navigation behavior. But I don't remember the details and in a quick scan I don't see the place where we'd expect to do that. @msmithNI do you expect these tests should be changing more significantly for the new column type?

jattasNI (commented):
Part way through the implementation and tests

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
msmithNI: I ended up revamping the tests more based on feedback from Milan. Most of the tests I had for interactive column types were generic, and didn't really need to be run for each additional interactive column type. Those are the tests in this file now (which don't use the anchor column anymore, they use a test column type).

There's a handful of tests in [table-column-anchor.spec.ts](https://github.com/ni/nimble/blob/0f03ba3f181a96a585bb01aaf7da3ce74f6aac22/packages/nimble-components/src/table-column/anchor/tests/table-column-anchor.spec.ts#L650) that cover focusing the control via Tab, and blurring it via Esc, that would probably be good to add for this column type too. Others:
- Checking `tabbableChildren` in various configurations (examples also in the anchor test file)
- Any additional interactions specific to the column type (something with the menu open for this column type, potentially)

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
jattasNI: I took a stab at addressing this feedback in #2231. If we like that approach and it goes in before this PR, you'll need to update this column's MDX file to refer to the same shared link text.

packages/nimble-components/src/table/index.ts
```diff
@@ -249,6 +252,12 @@ export class Table<
     // the selection checkbox 'checked' value should be ingored.
     // https://github.com/microsoft/fast/issues/5750
     private ignoreSelectionChangeEvents = false;
+    // Map from external slot name to the record ID of the row that should have the slot and
+    // the uniquified slot that the slot should be given.
```
jattasNI, line 256:
```suggestion
    // the uniquified name that the slot should be given.
```

jattasNI (commented):
Finished main review, just waiting for open threads to resolve to take another look.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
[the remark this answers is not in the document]
mollykreis: Good catch! Interestingly, I can't see that behavior with the original colors I had in the example (I guess they aren't long enough strings). But, I see it with the longer colors, so I guess that should be another copilot win since it's the one that generated all of these colors 😄

packages/nimble-components/src/table/index.ts
mollykreis: Your correction actually isn't correct. Trying to write clear comments with overloaded terms ("slot" in particular) has been difficult, but I'll try to clarify this. What do you think about this?

_Map from the external slot name to the record ID of the row that should have the slot and the unique slot name that the slot should be slotted into._

packages/nimble-components/src/table-column/menu-button/tests/table-column-menu-button.spec.ts
```diff
@@ -0,0 +1,538 @@
...
+        it('when opening a menu button by pressing Enter', async () => {
+            menuButton.pressEnterKey();
+            expect(menuButton.isOpen()).toBeTrue();
+            await waitForUpdatesAsync();
+            const selection = await table.getSelectedRecordIds();
+            expect(selection.length).toBe(0);
+        });
+    });
+
+    describe('menu button interactions', () => {
```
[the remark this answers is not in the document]
mollykreis: My current thoughts are that 1 & 2 are overkill since that behavior is completely incapsulated within the menu/menu items. That behavior seems pretty far removed from the table column. But, I'll add a test for 3 to make sure that clicking on an open menu button fires all the expected events and doesn't change table selection.

rajsite (commented):
Only partial comments for implementation (got distracted tinkering with the overflow directive 😅)

packages/nimble-components/src/table-column/menu-button/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    private readonly onDelegatedEvent = (e: Event): void => {
+        const event = e as CustomEvent<DelegatedEventEventDetails>;
+        const originalEvent = event.detail.originalEvent;
+
+        if (
+            originalEvent.type === 'beforetoggle'
+            || originalEvent.type === 'toggle'
+        ) {
+            const newEventName = `menu-button-column-${originalEvent.type}`;
```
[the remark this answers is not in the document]
mollykreis: Created https://github.com/ni/nimble/issues/2238

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
mollykreis: Thanks for the suggestions. I've added quite a few additional tests following the patterns in the anchor column tests.

packages/nimble-components/src/table/index.ts
```diff
@@ -731,6 +760,14 @@ export class Table<
         this.tableUpdateTracker.trackColumnInstancesChanged();
     }
 
+    private removeActionMenuSlotsFromColumnRequestedSlots(): void {
+        for (const actionMenuSlot of this.actionMenuSlots) {
+            this.columnRequestedSlots.delete(actionMenuSlot);
+        }
+
+        this.refreshRows();
+    }
```
rajsite, line 769: I'm having a hard time understanding why the columnRequestedSlots may contain actionMenuSlots.
Is this a naming conflict guard? If so can that be checked at validation instead?

rajsite (approved):

packages/nimble-components/src/table/index.ts
mollykreis: Yes, it's to allow the same menu/slot to be used for both a menu-button column and an action menu. We could decide that we didn't want to support that and make the table invalid in that case, but it isn't clear to me why that would be better than allowing a single menu to be used for both. The other alternative to having this specific code is that I could look into rewriting some of the action menu slotting code to reuse some of the new slot tracking code that is being added in this PR.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
mollykreis: Closing this discussion since the current decision is to keep the menu width as-is.

packages/nimble-components/src/table/index.ts
rajsite: > but it isn't clear to me why that would be better than allowing a single menu to be used for both.

yea, fair

> The other alternative to having this specific code is that I could look into rewriting some of the action menu slotting code to reuse some of the new slot tracking code that is being added in this PR.

that does sound nice, I was on the fence about tableData.slots not being used for both or using tableData to hold both (tableData.columnSlots + tableData.actionMenuSlots) maybe as a tech debt and follow-on instead of holding this PR?

packages/nimble-components/src/table-column/menu-button/cell-view/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    /** @internal */
+    public valueSpan?: HTMLSpanElement;
+
+    /** @internal */
+    @observable
+    public hasOverflow = false;
+
+    public override get tabbableChildren(): HTMLElement[] {
+        if (this.menuButton) {
```
msmithNI, line 39: Can you doublecheck what this'll do if the table starts out with non-empty string data, then it's replaced with empty strings? 
i.e. [this test](https://github.com/ni/nimble/blob/3165257cb2f224227c4f8dfa348a3dc090feb8ff/packages/nimble-components/src/table-column/menu-button/tests/table-column-menu-button.spec.ts#L109), but first set data as `[{ field: 'value' }]`, then set it to `[{ field: '' }]`

I think in that case, `this.menuButton` will be initialized but not in the DOM, and we don't want `tabbableChildren` to include it in that case.

(That's why the [anchor cell view checks this.showAnchor, not just this.anchor](https://github.com/ni/nimble/blob/297df53de72dc291959ee85bcdff240d96f3b4b3/packages/nimble-components/src/table-column/anchor/cell-view/index.ts#L68))

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
mollykreis: Since it isn't straightforward to make these styling changes to the button, I'll create an issue to tackle this as a follow up task. This way, we can have the necessary discussions around the best way to achieve the desired styling and double check that everything looks as expected with focus/hover states while not blocking the next stages of the column work.

mollykreis: Created #2247

packages/nimble-components/src/table-column/menu-button/cell-view/index.ts
mollykreis: Good catch! You were right that `this.menuButton` is still initialized (which seems wrong), but I've updated it as you suggested and added a test.

packages/nimble-components/src/table/index.ts
mollykreis: Created https://github.com/ni/nimble/issues/2248 to track this.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
mollykreis: Updated to follow the pattern established in your PR.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button-matrix.stories.ts
```diff
@@ -0,0 +1,66 @@
+import type { StoryFn, Meta } from '@storybook/html';
```
jattasNI, line 1: Do you think it would be useful to add this column to Table Keyboard Focus Theme Matrix story? I'm guessing yes since the column has focusable content so it'd be good to validate how its focus state renders. Though we don't currently include anchor columns or action menus in that test so I might be misinterpreting its purpose. @msmithNI feel free to offer an opinion (I know he's OOO today so we can proceed with our best guess).

https://60e89457a987cf003efc0a5b-qjyaqfanzg.chromatic.com/?path=/story/tests-table--table-keyboard-focus-theme-matrix

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Canvas of={menuButtonColumnStories.menuButtonColumn} />
+
+## API
+
+<TableColumnApisLink />
+<Controls of={menuButtonColumnStories.menuButtonColumn} />
+<ComponentApisLink />
+
+## Usage
```
jattasNI, line 24: A couple other ideas for docs:
1. at UIWG Fred mentioned this will often be used like a select column. Should we have 1st class docs for this use case? That could mean:
   - updating the example to show a check icon on the currently selected item
   - describing the use case at a high level in the overview
   - describing how to configure the menu (e.g. which check icon to use, which menu item slot to put it in)
2. usage guidance for when to use this column and when to use the action menu

I think both of these are nice-to-have but not required. It's ultimately @fredvisser's call whether they're needed but feel free to do this in a follow up PR if you're unsure of the value and don't want to block this one.

jattasNI (approved):

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button-matrix.stories.ts
rajsite: I'd think this column should have it's own focus theme matrix test instead of putting all columns in a single test.

rajsite: Also probably reasonable to address in the style update follow-on

msmithNI: I don't think it's necessarily needed because the table doesn't add focus visuals to the controls inside cell views, the controls themselves provide that focus styling. Per-column matrix tests could still make sense though I guess.

packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
mollykreis: @atmgrifter00 helped me work through the problem. It looks better in chromatic now. @jattasNI - can you double check that it looks correct in safari as well?

jattasNI: Nice! Looks good to me!

image
image

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button-matrix.stories.ts
mollykreis: I'll add a note to #2247 that we should also consider adding a focus theme matrix for the column.

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
mollykreis: I'll follow up with @fredvisser on what, if any, additional docs we want to add or changes we want to make to the example. For now, I'm resolving the conversation to unblock the PR.

Issue #1582, opened 2023-10-02
nimble-button table column


jattasNI:
## 😯 Problem to Solve

We have received requests for a table column containing buttons. Example use cases:
1. a column of icon buttons that trigger a row to be deleted when clicked. This requirement comes from the Assets app. A possible part of the UX is to only show the button when the row is hovered (this would likely be more complicated than a normal button so we should negotiate whether it's necessary). 

   - [wireframe](https://www.figma.com/file/L8Xgic88jJYZ6SjTl49kzi/Move-assets-wireframes?type=design&node-id=0-1&mode=design&t=0cGTQYj9cKQZaOm6-0)
   - [requirements](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/docs/design-documents/Asset/Requirements/Asset-location-management.md&_a=preview)

![image](https://github.com/ni/nimble/assets/10500124/910d2566-2124-4923-9076-fabefcf7f521)

2. a column of icon buttons to mark a row as "favorite" (or "un-favorite"). This use case comes from the TestHub app. An interesting aspect is that the button content might change on each row (yellow star vs hollow star icon) and that the button icon itself might indicate hover state, rather than a square around it (though this still needs a VxD pass)
![image](https://github.com/ni/nimble/assets/10500124/35b7116c-d973-49fa-8a39-7a4358d81e5a)

3. In general one can imagine an app wanting to promote any item from the row action menu to the top level

## 💁 Proposed Solution

A column of nimble-button or maybe nimble-toggle-button

jattasNI:
Another use case for this is the SLE file upload table. It currently uses mat-table and has a button on each row to cancel a file upload. For now we could potentially replace that button with an action menu.

In that implementation the button only appears on hover and it goes away when the upload finishes. The former is probably negotiable but we'd likely need to support something like the latter (hide or disable button per-row).

![image](https://github.com/ni/nimble/assets/10500124/7e9118af-3a57-4efb-99ad-71bffaed7c99)

m-akinc:
> [button only appears on hover] is probably negotiable

I'm not an interaction designer, but that design seems problematic from the perspectives of discoverability and accessibility. 

I also wonder if an action menu isn't a better experience anyway than a separate column for a button that IMHO has questionable use anyway. What are the scenarios where a user would want to cancel the upload of a file that they just selected to upload?

Finally, as a side note, do we want to create a progress column, or do we expect this table to create separate columns for the spinner and numeric percentage?

[...]

PR #2205, opened 2024-06-20
Create HLD for button table column

m-akinc:
# Pull Request

## 🤨 Rationale

HLD for the feature in #1582 

## 👩‍💻 Implementation

HLD created.

## 🧪 Testing

NA

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-6354479c-4887-4a0c-a975-c678dbf4b269.json
+++ b/change/@ni-nimble-components-6354479c-4887-4a0c-a975-c678dbf4b269.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Create HLD for button table column",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/packages/nimble-components/src/table/specs/table-column-specs/table-column-button-hld.md
+++ b/packages/nimble-components/src/table/specs/table-column-specs/table-column-button-hld.md
@@ -1,0 +1,157 @@
+# Button Table Column
+
+## Overview
+
+A table column whose cells render an action button (`nimble-button`).
+
+### Background
+
+The primary ask has been from the Systems app to have an icon button that copies some long key related to a pending system. Another potential use case is the ability to mark a row as a "favorite", but it is not clear if we should have a separate toggle-button column type for those sorts of stateful use cases.
+
+- [Table button column (AzDO Feature)](https://ni.visualstudio.com/DevCentral/_workitems/edit/2530349)
+- [nimble-button table column (Nimble issue)](https://github.com/ni/nimble/issues/1582)
+
+### Features
+
+- Text and/or icon(s) in button
+- Option to hide buttons except when hovered or row/cell is focused (similar to action menu)
+- Keyboard focusable
+- Fixed-width
+
+### Non-goals
+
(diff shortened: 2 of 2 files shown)


change/@ni-nimble-components-6354479c-4887-4a0c-a975-c678dbf4b269.json
m-akinc, line 1: This initial draft is pretty simple, but there is one significant question that could blow up the scope. Specifically, whether to support stateful use cases like this:
image
If we need that, the API would probably involve our existing `nimble-mapping-icon` and `nimble-mapping-text` elements rather than a simpler, slot-based API. Also, supporting non-uniform buttons means we need sorting, grouping, and resizing (b/c text may vary), where otherwise we wouldn't. @fredvisser @jattasNI @rajsite initial thoughts?

m-akinc: Per discussion in team meeting:
- We do not want to support a button that can change its icon/text based on record data.
- We probably do not want to support action-menu-like visibility (`hide-unless-focused`) initially, as long as it doesn't make it harder to add at a later time.
- The API should not involve cloning elements. We should reference icons by name, like with the mapping element.
- We do not want to support icons in the end slot of a button (should update existing button docs to say so).
- We _may_ want to support hiding or disabling buttons on a row-by-row basis. There doesn't seem to be a use case right now, though.

jattasNI: Didn't get a chance to make this point at the meeting but I think it'd be helpful to include a section that describes the ideas that are initially out of scope and mentions that we think they're still achievable with API additions if they later become in scope. You could even consider sketching out those possible APIs or citing existing similar APIs, though wouldn't want to get into too much detail to avoid the sense that reviewers are approving those strawman APIs.

packages/nimble-components/src/table/specs/table-column-specs/table-column-button-hld.md
```diff
@@ -0,0 +1,155 @@
+# Button Table Column
+
+NOTE: No longer planned (at least, with this limited scope). Client use cases that would have used this form of the button column are served well enough by an action menu item.
```
m-akinc, line 3: I added this note since we're no longer planning to create this column type (in this proposed, limited scope). I think we should still submit this as a starting point/history in case we revisit the idea of a button column in the future. @jattasNI if you agree and approve, I'll just move it out of draft and submit.

jattasNI: Any objection to keeping your branch around indefinitely? If not, I think I'd rather link to the branch from the related issue and leave it unsubmitted, similar to an innovation day research project. Or maybe copy the contents of the HLD into a comment in the issue if you're worried about the branch disappearing.

I'd personally vote against having unapproved HLDs be committed to the repo since they'd be potential sources of confusion (even with a disclaimer at the top) and if we ever revive them I'd rather we look at the proposal with fresh eyes, using our up-to-date templates, than look at a diff versus a previously unapproved proposal.

m-akinc: Ok, I don't feel that strongly. I'll just leave the branch on the remote. Even if it ends up deleted somehow, there aren't really any critical insights in this doc--mostly just some mechanical stuff we'd have to rewrite.

PR #2206, opened 2024-06-21
Create HLD for custom column sorting in the table

mollykreis:
# Pull Request

## 🤨 Rationale

Create HLD for custom column sorting in the table

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
```diff
@@ -0,0 +1,31 @@
+# Custom table column sort order HLD
+
+## Problem Statement
+
+In some cases, a client may have the need to specify a custom sort order for a column. For example, a table displaying SystemLink tags may have a field in their records for the stringified tag but may not want to sort that column that stringified value. Instead, they want to provide their own sort order for tags that takes each tag's data type and value into account.
```
m-akinc, line 5:
```suggestion
In some cases, a client may have the need to specify a custom sort order for a column. For example, a table displaying SystemLink tags may have a field in their records for the stringified tag but may not want to sort that column by that stringified value. Instead, they want to provide their own sort order for tags that takes each tag's data type and value into account.
```

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
```diff
@@ -0,0 +1,31 @@
...
+
+[Azdo Feature 2665131 - Custom column sorting](https://dev.azure.com/ni/DevCentral/_workitems/edit/2665131)
+
+## Implementation / Design
+
+Columns that need support for a custom column sort order will extend their API to allow a client to optionally specify an additional field within the data that specifies the sort order of that column. This field is expected to be a numeric value that is the sorted data index of the record. When the table is sorted by this column, the table will use the `basic` sort operation and sort the records by the additionally specified field rather than the default `operandDataRecordFieldName` associated with the column.
+
+TanStack uses the same field internally for both sorting and grouping. Grouping by this custom sort field could likely lead to unexpected results since there isn't a guarantee that two values that render identically have the same sort sort index. Therefore, grouping and specifying a custom sort order will not be allowed on a column at the same time. If both are specified at the same time, the column will be considered invalid and the `invalidCustomSortWithGrouping` validation flag will be set on the column.
```
m-akinc, line 15: Another possibility might be to respond to setting the group index on the column by switching which column TanStack uses for grouping/sorting. I.e. when grouping, use the default `operandDataRecordFieldName`, otherwise use the `sort-by-field-name`. We still couldn't support both sorting _and_ grouping at the same time.

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
```diff
@@ -0,0 +1,31 @@
...
+
+In some cases, a client may have the need to specify a custom sort order for a column. For example, a table displaying SystemLink tags may have a field in their records for the stringified tag but may not want to sort that column that stringified value. Instead, they want to provide their own sort order for tags that takes each tag's data type and value into account.
+
+## Links to Relevant Work Items and Reference Material
+
+[Azdo Feature 2665131 - Custom column sorting](https://dev.azure.com/ni/DevCentral/_workitems/edit/2665131)
+
+## Implementation / Design
+
+Columns that need support for a custom column sort order will extend their API to allow a client to optionally specify an additional field within the data that specifies the sort order of that column. This field is expected to be a numeric value that is the sorted data index of the record. When the table is sorted by this column, the table will use the `basic` sort operation and sort the records by the additionally specified field rather than the default `operandDataRecordFieldName` associated with the column.
```
m-akinc, line 13: From a client perspective, a more convenient API might be to provide a custom sort function rather than to annotate their data with an extra field. I.e. change [`accessorFn`](https://github.com/ni/nimble/blob/d37ca77cda0729b42f05b4bac79be8ccc98ceb22/packages/nimble-components/src/table/index.ts#L1138) to:
```ts
(record: TableNode<TData>): TableFieldValue => {
    if (typeof column.columnInternals.customSortFunction === 'function') {
        return column.columnInternals.customSortFunction(record.clientRecord);
    }
    const fieldName = column.columnInternals.operandDataRecordFieldName;
    if (typeof fieldName !== 'string') {
        return undefined;
    }
    return record.clientRecord[fieldName];
}
```
The main drawback I can think of is that we'd be making it possible for clients to shoot themselves in the foot with a slow sort function, but I think documentation can mitigate that. Also, if they wanted the same speed as the other approach, clients could always pre-calculate the sort indices, add a new field to their record, and let the function return that field's value.

Are there other reasons not to prefer this API?

mollykreis: The approach you're suggesting goes against the guideline we've been following with the table that we should never call into client code when interacting with the table. Therefore, I don't think we'd want to allow the client to implement a custom sort function -- similar to the fact that we didn't allow them to implement a custom format function for string/number/date/duration columns. Additionally, I'm not sure how this would work for Blazor clients.

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
mollykreis: Yeah, technically we could do that, but I'm not sure how much that gets us.

Actually, if we really wanted to support grouping & sorting together on these columns we probably could do that. We'd just need to create two different TanStack columns for a given nimble column and then manage the state of each one. I'll add that as an alternative so that we can decide if we can weigh the additional complexity vs the additional functionality.

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
m-akinc: > the table that we should never call into client code when interacting with the table.

Is the concern (other than performance) that the client could do all sorts of things that might not be handled correctly in the middle of an operation, e.g. deleting rows/columns? Is it substantially different than the client registering event handlers?

> I'm not sure how this would work for Blazor clients.

Good point.

m-akinc (approved):

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
m-akinc: Let's see what others think, but if it's pretty cheap to support grouping and sorting (but not at the same time), that seems worthwhile. Supporting both at the same time sounds like it might add more complexity than it's worth.

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
jattasNI: Since clients are used to the custom sort function approach, it would be good to list that as an alternative so we have something to point to in case they ask why we changed that API.

And we might want to consider noting the difference in the migration guide for SLE (possibly low ROI, but we did get feedback about teams feeling blindsided by other behavior changes).

mollykreis: I've added the custom sort function idea as an alternative.

I'm not sure about putting it in the migration guide. This seems very column specific rather than general to the whole `sl-grid` to `sl-table` migration as a whole. But, if we think that would be helpful for clients, I can see where in that document this information might make sense.

atmgrifter00 (approved):

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
```diff
@@ -0,0 +1,57 @@
...
+
+TanStack uses the same field internally for both sorting and grouping. Grouping by this custom sort field could likely lead to unexpected results since there isn't a guarantee that two values that render identically have the same sort sort index. Therefore, grouping and specifying a custom sort order will not be allowed on a column at the same time. If both are specified at the same time, the column will be considered invalid and the `invalidCustomSortWithGrouping` validation flag will be set on the column.
```
jattasNI, line 15: This is possibly in the weeds / implementation detail, but what requirements and guarantees would we have for the values of the sort-by field?

1. Does it have to be an index (i.e a whole number from 0 to the number of records) or is any `Number` ok?
2. Is it ok for multiple records to have the same value? If so what is the tiebreaker to determine their sort order? The `field-name` / `label-field-name` display value? The original order of the data? No guarantees?
3. Do we need to define what happens if a value isn't provided (or is `null` / `NaN` / non-numeric, etc) for some records?

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
jattasNI: My 2 cents: 
If we really wanted both sorting and grouping, one option would be to put the burden on the client to provide equal `sort-by-field-name` values for records that should be grouped together. Sorting would fall back to whatever tiebreaker we decide (see my other thread) and grouping would behave as it does today. A downside of that would be increased complexity on client code for generating the sort by field values.

If we don't like that increased complexity (e.g. if we think clients will get it wrong) then I would lean towards not supporting grouping at all. It's not worth adding complexity if we don't have a requirement to allow sorting and grouping together. And for the types of mixed and not-well-normalized data in the column types we know about, grouping by the display value won't be particularly useful (e.g. separate groups for tags "1", "1.0", and "1%").

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
jattasNI, line 15:
```suggestion
TanStack uses the same field internally for both sorting and grouping. Grouping by this custom sort field could likely lead to unexpected results since there isn't a guarantee that two values that render identically have the same sort index. Therefore, grouping and specifying a custom sort order will not be allowed on a column at the same time. If both are specified at the same time, the column will be considered invalid and the `invalidCustomSortWithGrouping` validation flag will be set on the column.
```

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
mollykreis: Your suggestion of grouping on `sort-by-field-name` doesn't seem unreasonable from a behavior perspective. However, that still isn't free from the nimble side. Specifically, the group row's value comes from TanStack, but in your suggestion that would be the sort index (e.g. `0`, `1`, `2`, etc), which is not what we would want to have render in the group row. Therefore, we'd still have to add code that can figure out what the "actual" field value is, rather than the sort index.

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
mollykreis: The plan would be to use the existing `basicSortFunction`, which is used for our numeric-backed columns (e.g. number-text, duration). As a result:

1. Any `Number` is ok
2. There are no guarantees if you have multiple records with the same value. I'd guess the behavior is to maintain original order of the data, but we don't guarantee that (especially since that behavior would come from TanStack).
3. The sort function already has defined behavior for `null`, `undefined`, `NaN`. It also handles non-numeric cases, but it just wouldn't be locale-aware if you pass it strings. It uses the `>` operator for anything that isn't a special-cased value.

Feel free to look at the implementation of the sort function if you're curious:
https://github.com/ni/nimble/blob/d37ca77cda0729b42f05b4bac79be8ccc98ceb22/packages/nimble-components/src/table/models/sort-operations.ts#L50-L89

fredvisser (approved):

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
jattasNI: Regarding the migration guide, I don't feel strongly that it's necessary. There are likely only two teams doing migrations now: ours and the one who requested this feature. Both of us are well aware of the plan, so writing it down in another place isn't critical.

jattasNI (approved):

packages/nimble-components/src/table/specs/table-column-custom-sort-field-hld.md
```diff
@@ -0,0 +1,57 @@
...
+
+In some cases, a client may have the need to specify a custom sort order for a column. For example, a table displaying SystemLink tags may have a field in their records for the stringified tag but may not want to sort that column by that stringified value. Instead, they want to provide their own sort order for tags that takes each tag's data type and value into account.
+
+## Links to Relevant Work Items and Reference Material
+
+[Azdo Feature 2665131 - Custom column sorting](https://dev.azure.com/ni/DevCentral/_workitems/edit/2665131)
+
+## Implementation / Design
+
+Columns that need support for a custom column sort order will extend their API to allow a client to optionally specify an additional field within the data that specifies the sort order of that column. This field is expected to be a numeric value that is the sorted data index of the record. When the table is sorted by this column, the table will use the `basic` sort operation and sort the records by the additionally specified field rather than the default `operandDataRecordFieldName` associated with the column.
```
rajsite, line 13: > This field is expected to be a numeric value that is the sorted data index of the record

It doesn't actually have to be though right? It just has to be a set or orderable numbers (could be negative, could be fractional)? Maybe we can phrase it as a recommended pattern, but it's not required (and we don't depend on the data to actually be indexes right?)

rajsite (approved):

PR #2208, opened 2024-06-22
Use trace colors on landing page

rajsite:
# Pull Request

## 🤨 Rationale

Switched to using the graph trace colors for the circles on the landing page.

## 👩‍💻 Implementation

Switched to the trace tokens. Also put it in dark theme which gives lighter color traces.

## 🧪 Testing

Rely on CI.

## ✅ Checklist

### DIFF
--- a/packages/site/landing/index.html
+++ b/packages/site/landing/index.html
@@ -62,16 +62,19 @@
                 </div>
             </div>
         </div>
-        <svg class="circle-wrapper" id="circles" viewBox="0 0 1550 1550" width="1300" height="1300" stroke-width="50"
-            fill="transparent" stroke-linecap="round">
-            <circle cx="775" cy="775" r="750" />
-            <circle cx="775" cy="775" r="700" />
-            <circle cx="775" cy="775" r="650" />
-            <circle cx="775" cy="775" r="600" />
-            <circle cx="775" cy="775" r="550" />
-            <circle cx="775" cy="775" r="500" />
-            <circle cx="775" cy="775" r="450" />
-        </svg>
+        <nimble-theme-provider theme="dark">
+            <svg class="circle-wrapper" id="circles" viewBox="0 0 1550 1550" width="1300" height="1300" stroke-width="50"
+                fill="transparent" stroke-linecap="round">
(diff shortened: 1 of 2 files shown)


packages/site/landing/index.html
```diff
@@ -62,16 +62,19 @@
...
-            fill="transparent" stroke-linecap="round">
-            <circle cx="775" cy="775" r="750" />
-            <circle cx="775" cy="775" r="700" />
-            <circle cx="775" cy="775" r="650" />
-            <circle cx="775" cy="775" r="600" />
-            <circle cx="775" cy="775" r="550" />
-            <circle cx="775" cy="775" r="500" />
-            <circle cx="775" cy="775" r="450" />
-        </svg>
+        <nimble-theme-provider theme="dark">
```
rajsite, line 65: Traces in dark theme
image

Traces in light theme:
image

jattasNI: @NIbokeefe, wanna weigh in on whether you like this direction? I think it's technically cool to use our official graph tokens for the home page rings but if you prefer the aesthetic of the original design then I'd say it's your call.

NIbokeefe: @jattasNI @rajsite 

The dark themed traces look interesting because they are softer against the white.  The light theme looks too harsh.  Graph colors were designed to be viewed at very thin sizes, it doesn't hold up for these thick traces.

I still probably prefer the original NIBrand colors, but again the dark graph colors are #2 for me.  What prompted the possible changes?

rajsite: Just seemed neat to use the same colors someone would see on a graph / elsewhere in the design system. No real reason to change. 

Keeping the current brand colors is 100% Gucci. One question is would we use the same exact set of colors for an actual dark mode landing page (ie dark background etc)?

NIbokeefe: Gotcha, thanks!

Yes, I tested all the colors on dark mode and it looks great.   Here's what that looks like:

Screenshot 2024-06-27 at 11 17 53 AM

rajsite: Love it, let's keep it as-is!

PR #2214, opened 2024-06-25
Avoid phantom diffs from prettier-eslint

m-akinc:
# Pull Request

## 🤨 Rationale

Running `npm run format` would result in git detecting that `combobox/select.foundation.spec.ts` had been modified, but the diffs would show no changes.

## 👩‍💻 Implementation

The problem was that formatting was replacing some CRLF line endings with LF line endings.

`prettier-eslint` runs `prettier` followed by `eslint` to fix any formatting issues. The `xxx.foundation.spec.ts` files each had some lines like this:

```ts
const {
    element, connect, disconnect, option1, option2, option3
} = await setup();
```

`prettier` would reformat those lines like this, because they could fit on a single line without exceeding the line length limit:
```ts
const { element, connect, disconnect, option1, option2, option3 } = await setup();
```

Then `eslint` would run, and because these objects have more than five properties (the limit configured by the NI styleguide), it inserts newlines next to the curly braces, so it looks like the original file, except that the inserted newlines are Unix-style (LF) rather than Window-style (CRLF). This causes Git to detect a diff, but it's invisible. Even if you would submit the change, Git normalizes all the line endings to Unix-style upon check-in anyway, so there's no actual, commitable change.

### Solution

Ideally, we could configure `eslint` to use whatever line ending type is used in the rest of the file, but it seems to [only let you specify Unix or Windows](https://eslint.org/docs/latest/rules/linebreak-style). Instead, I am now reconfiguring the `object-curly-newline` rule to use an arbitrarily large `minProperties` value, so that `eslint` does not introduce newlines just because there are a certain number of properties in an object. `prettier` will still introduce newlines as needed, based on line length.

I also updated the npm `format` scripts to stop calling `eslint-fix` before `prettier-fix`. The latter already [does both a `prettier -fix` and `eslint -fix`](https://github.com/prettier/prettier-eslint?tab=readme-ov-file#this-solution), so the separate `eslint-fix` is redundant.

## 🧪 Testing

Ran `npm run format` on the whole repo, and only the two problem files were changed.

## ✅ Checklist

### DIFF
--- a/packages/eslint-config-nimble/typescript.js
+++ b/packages/eslint-config-nimble/typescript.js
@@ -34,14 +34,17 @@ module.exports = {
         'no-return-await': 'off',
         '@typescript-eslint/return-await': 'error',
 
-        // Increasing the value of minProperties (default is 6?) so that eslint doesn't
-        // introduce line breaks where prettier doesn't. If eslint introduces line breaks,
-        // they will be unix-style, which will cause pointless diffs in git.
+        // This rule's configuration is based on the NI javascript styleguide:
+        // https://github.com/ni/javascript-styleguide/blob/a1a6abd7adca7d9acd002705101b351d695b2442/packages/eslint-config-javascript/index.js
+        // The only difference is that we're increasing the value of minProperties (from 6) so
+        // that eslint doesn't introduce line breaks where prettier doesn't. If eslint introduces
+        // line breaks, they will be unix-style, which will cause pointless diffs in git.
         'object-curly-newline': ['error', {
-            multiline: true,
-            consistent: true,
-            minProperties: 1000
-        }]
+            ObjectExpression: { minProperties: 1000, multiline: true, consistent: true },
+            ObjectPattern: { minProperties: 1000, multiline: true, consistent: true },
+            ImportDeclaration: { consistent: true },
+            ExportDeclaration: { consistent: true }
+        }],
     },
     overrides: [
         {
--- a/packages/eslint-config-nimble/typescript.js
+++ b/packages/eslint-config-nimble/typescript.js
@@ -32,7 +32,16 @@ module.exports = {
         'require-await': 'off',
         '@typescript-eslint/require-await': 'error',
         'no-return-await': 'off',
-        '@typescript-eslint/return-await': 'error'
+        '@typescript-eslint/return-await': 'error',
+
+        // Increasing the value of minProperties (default is 5?) so that eslint doesn't
+        // introduce line breaks where prettier doesn't. If eslint introduces line breaks,
+        // they will be unix-style, which will cause pointless diffs in git.
+        'object-curly-newline': ['error', {
+            multiline: true,
+            consistent: true,
+            minProperties: 1000
+        }]
     },
     overrides: [
         {
--- a/packages/eslint-config-nimble/typescript.js
+++ b/packages/eslint-config-nimble/typescript.js
@@ -34,7 +34,7 @@ module.exports = {
         'no-return-await': 'off',
         '@typescript-eslint/return-await': 'error',
 
-        // Increasing the value of minProperties (default is 5?) so that eslint doesn't
+        // Increasing the value of minProperties (default is 6?) so that eslint doesn't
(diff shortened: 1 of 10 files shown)


packages/eslint-config-nimble/typescript.js
```diff
@@ -32,7 +32,16 @@ module.exports = {
         'require-await': 'off',
         '@typescript-eslint/require-await': 'error',
         'no-return-await': 'off',
-        '@typescript-eslint/return-await': 'error'
+        '@typescript-eslint/return-await': 'error',
+
+        // Increasing the value of minProperties (default is 6?) so that eslint doesn't
+        // introduce line breaks where prettier doesn't. If eslint introduces line breaks,
+        // they will be unix-style, which will cause pointless diffs in git.
+        'object-curly-newline': ['error', {
```
m-akinc, line 40: The docs say this rule (and all other formatting related rules, apparently) was deprecated in eslint 8.53.0 (we're on 8.57.0). The recommendation is to get the same rules from a different package, `@stylistic/eslint-plugin-js`, but I didn't want to go down that road right now. If this stops being supported or we get a bunch of ugly warnings in the future, we can deal with it then.

rajsite, line 40: at first glance this set of options seem pretty different than the styleguide config. Can you instead use the styleguide config with minimal changes and link to what it was forked from, i.e. https://github.com/ni/javascript-styleguide/blob/a1a6abd7adca7d9acd002705101b351d695b2442/packages/eslint-config-javascript/index.js#L397-L402

rajsite, line 40: This is a general eslint config rule, not a typescript-eslint specific rule (notice the name does not begin with `@typescript-eslint`) can you move this to the JS rule set?

rajsite: Oh there actually is a typescript eslint rule override: https://typescript-eslint.io/rules/object-curly-spacing/

1. So the current rule needs to go in the JS ruleset, and in the TS rules the JS needs to be disabled and the ts rule enabled, i.e.
  
   js rules:
    ```
    'object-curly-newline': ['error', {
        ...blah
    }]
    ```
  
    ts rules:
    ```
    'object-curly-newline': 'off',
    '@typescript-eslint/object-curly-newline': ['error', {
        ...blah
    }]
    ```
2. Can you create an issue in the ni style guide to enable the typescript-eslint object-curly-newline override rule

rajsite: Yea I'd expect that transition should happen in the ni style-guide. Could you create an issue to point out the deprecated rules there?

m-akinc: Created https://github.com/ni/javascript-styleguide/issues/147. I wasn't sure if we want to change the `minProperties` value for the JS and TS versions of that rule, so I didn't include that in the issue. My guess is that since our motivation for increasing it is specific to using `prettier-eslint` (and not just `eslint` by itself), we want to leave `minProperties` at 6 in the styleguide.

rajsite: Ahh, saw the issue close, google search did me dirty (and not reading close enough) lol

m-akinc: Whoops. Noticed a little too late that there _isn't_ a @typescript-eslint version of the rule. There is a typescript rule `object-curly-spacing`, but not `-newline`. I closed the issue that I had created.

Since the typescript config doesn't inherit from the javascript one (should it?), I have to configure the rule in both the ts and js configs.

rajsite: configuring in both is fine for now, there are challenges making them inherit not worth dealing with atm

rajsite (approved):

packages/eslint-config-nimble/typescript.js
m-akinc: Created https://github.com/ni/javascript-styleguide/issues/148

m-akinc:
@atmgrifter00 I removed the file (Angular example app file) that pulled you in as a code owner. Nothing left for you to review now.

jattasNI (approved):

PR #2218, opened 2024-06-26
Angular support and example for select custom filtering

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1866

## 👩‍💻 Implementation

Primarily this is just exposing the needed attribute for controlling the visibility of the loading indicator for the `Select` in Angular, accompanied with an update to the example application showing how to manage dynamic options.

However, in the making of the example it became clear that a change to the `Select` web component was needed to make the client experience cleaner. Now, when the `filter-input` event is issued as part of the action of changing the value, the `change` event will occur first, followed by the `filter-input` event. This required not tying the emitting of the `filter-input` event to the updating of the `open` attribute, as we will still update the `open` state prior to issuing the `change` event.

## 🧪 Testing

Wrote new tests validating the new expected event ordering. Added typical Angular tests for new attribute.

## ✅ Checklist


packages/nimble-components/src/select/tests/select.spec.ts
m-akinc, line 1: Might want to also verify that the events are only emitted once. I was a little iffy about whether we might be emitting an extra time in response to an ESC when looking at that code. I think I'm convinced that we don't, but it still might be worth asserting. Up to you.

packages/nimble-components/src/select/index.ts
m-akinc, line 1: I'm having a hell of a time trying to verify these changes do what they should just by reading the code. I think I've given similar feedback on a previous PR. 😅 The interplay of `value`, `selectedIndex`, `openActiveIndex`, `ListboxElement.activeOption`, `ListboxElement.activeIndex`, `firstSelectedOption`, `setActiveOption()`, and `updateValue()` is not easy to follow. For example, the `value` setter updates `selectedIndex`, but there are lots of other places where `selectedIndex` is assigned directly and followed up with `updateValue()` which in turn calls the `value` setter. I'm not smart enough for this. 🤯

I'm sure most of this logic is FAST's, and obviously you're not going to do some big refactor for this PR. So I guess I would ask if there's some way to tie calling `emitFilterInputEvent()` to a single place where `filter` is set to `''`. I assume that's really why we're emitting that event. Calling `emitFilterInputEvent()` because `this.open` was set to false, and we know that `openChanged()` sets `filter = ''` is not great. It also means we have to call that function in a bunch more places, and that implicit coupling is hard to maintain. Could we instead make sure `updateValue(true)` gets called _before_ setting `close = false`? I haven't looked too closely. I may try to look at this more tomorrow.

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -99,17 +103,39 @@ export class CustomAppComponent implements AfterViewInit {
...
+        { first: 'Selma', last: 'Bouvier' },
+        { first: 'Patty', last: 'Bouvier' },
+        { first: 'Abe', last: 'Simpson' },
+        { first: 'Ned', last: 'Flanders' },
+        { first: 'Maude', last: 'Flanders' },
+        { first: 'Rod', last: 'Flanders' },
+        { first: 'Todd', last: 'Flanders' }
+    ];
+
+    private readonly localSelectItems = this.availableSelectItems.slice(0, 5);
```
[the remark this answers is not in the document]
atmgrifter00: I don't feel too strongly about the current name, but I would use `defaultSelectItems` of the two suggestions, as they aren't supposed to represent just an "initial" set but the set that will _always_ be used (at least as part of the whole set) and not just what is there initially.

packages/angular-workspace/nimble-angular/src/directives/select/tests/nimble-select.directive.spec.ts
```diff
@@ -82,6 +82,17 @@ describe('Nimble select', () => {
...
+        it('has expected defaults for loadingVisible', () => {
+            expect(directive.loadingVisible).toBeFalse();
+            expect(nativeElement.loadingVisible).toBeFalse();
+        });
+
+        it('can use the directive to set loadingVisible', () => {
+            directive.loadingVisible = true;
+            expect(directive.loadingVisible).toBeTrue();
+            expect(nativeElement.loadingVisible).toBeTrue();
+        });
```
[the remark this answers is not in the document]
atmgrifter00: Agreed. This was just a copied pattern.

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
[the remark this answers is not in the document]
atmgrifter00: Went with `defaultDynamicSelectItems` to make it even more targeted in its use semantics.

m-akinc (approved):
We discussed my main concern about the maintainability of peppering the even emits throughout the code, and we had one potential idea to use `requestAnimationFrame` to delay the event. Going to approve so others can weigh in.

packages/nimble-components/src/select/index.ts
atmgrifter00: @rajsite @jattasNI, @m-akinc and I discussed this concern off-line, and while I explained that rearranging the ordering of setting `open = false` and calling `updateValue(true)` was the initial considered solution, an idea was brought up that to me seems more reasonable than what is currently in this PR: would we be fine with simply putting the call to emit the `filter-input` event within the `openChanged` callback in a rAF routine? This should provide us the necessary delay to allow `updateValue(true)` to run when needed, and feels more maintainable than the various scattered calls to emit the `filter-input` event.

rajsite: Think it's well noted that the current select code is difficult to maintain. Didn't we all pinky promise to not block implementation of features needed to move SLE onto the select with the understanding that afterwards we would refactor it?

> would we be fine with simply putting the call to emit the filter-input event within the openChanged callback in a rAF routine?

The filter event results in a bunch of complex state changes (by swapping out child options). Firing off an asynchronous task that runs at an unknown time / ordering in the future, that we don't track the completion of, etc that results in lots of state changes sounds like is going to be hard to debug.

I'd recommend finding a direction that doesn't add complexity to understanding the timing if possible.

atmgrifter00: The only alternative I can think of is my original approach called out above, which I honestly think I prefer. In summary, that was to change the ordering so that we call `updateValue(true)` first, followed by `open = false`. The considered drawback to this was fairly pathological, and I'm not sure actually can happen, but we wondered that if a user handled the `change` event, updated the DOM with new options, then while the dropdown was still open (hasn't been updated yet synchronously), that the dropdown would update with the new DOM options before closing. However, in all likelihood, the dropdown would be closed before any render updates would occur.

Again, I'm happier with that approach from an implementation perspective, but will concede the small risk of a visual glitch. Beyond a larger refactor, I have no other immediate alternative. For the sake of progress, I would suggest we pick either what I currently have in this PR, or make the noted switch to what is described above.

atmgrifter00: I've updated PR to switch when we update `open` (which will issue the `filter-input` event) and when we update the value during a user-initiated value change.

rajsite (approved):

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1989,6 +1989,79 @@ describe('Select', () => {
...
+                );
+                const filterInputEventListener = createEventListener(
+                    element,
+                    'filter-input'
+                );
+
+                const changeEventWasFirst = Promise.race([
+                    changeEventListener.promise.then(() => true),
+                    filterInputEventListener.promise.then(() => false)
+                ]);
```
rajsite, line 2007: I feel like a cleaner way to do this check would be to register the same spy to both and then do assertions on the spy.calls, i.e. checking the first and second calls had expected arguments

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1989,6 +1989,79 @@ describe('Select', () => {
...
```
rajsite, line 2027: Same as above, think sharing the spy and asserts on the calls might be cleaner

packages/nimble-components/src/select/index.ts
```diff
@@ -603,8 +604,12 @@ export class Select
...
-                this.open = !this.open;
-                if (!this.open) {
+                if (this.open) {
+                    commitValueThenClose = true;
+                } else {
+                    this.open = true;
+                }
+                if (this.open) {
                     this.focus();
                 }
```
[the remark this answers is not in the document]
atmgrifter00: Kept the logic in-line with what was before, so instead of checking `this.open` it now checks `commitValueThenClose` before calling `this.focus()`.

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -121,6 +147,31 @@ export class CustomAppComponent implements AfterViewInit {
...
 
+    public onDynamicSelectFilterInput(e: Event): void {
+        const event = e as CustomEvent<SelectFilterInputEventDetail>;
+        const filter = event.detail.filterText;
+        if (filter.length > 0) {
+            window.clearTimeout(this.dynamicSelectFilterTimeout);
+            this.dynamicSelect.loadingVisible = true;
+            this.dynamicSelectFilterTimeout = window.setTimeout(() => {
+                // do your custom filtering here
+                const filteredItems = this.availableSelectItems.filter(item => item.first.concat(item.last).toLowerCase().includes(filter.toLowerCase()));
```
jattasNI, line 158: Do we want to encourage using the diacritic insensitive filter by using it here?

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -121,6 +147,31 @@ export class CustomAppComponent implements AfterViewInit {
         alert(`${menuItemText} selected`);
     }
 
+    public onDynamicSelectFilterInput(e: Event): void {
+        const event = e as CustomEvent<SelectFilterInputEventDetail>;
+        const filter = event.detail.filterText;
+        if (filter.length > 0) {
+            window.clearTimeout(this.dynamicSelectFilterTimeout);
+            this.dynamicSelect.loadingVisible = true;
+            this.dynamicSelectFilterTimeout = window.setTimeout(() => {
```
jattasNI, line 156: When playing with this during the UI WG meeting I found it slightly unexpected that the original items that don't match the filter were still visible while the filter was "processing". Depending on how "real" we want this example to be, we could change that. Or we could decide that overcomplicates things and that the main purpose of the example is for our own testing. I'm fine either way.

packages/nimble-components/src/select/index.ts
```diff
@@ -635,14 +641,18 @@ export class Select
...
+            }
 
-        if (!this.open && initialSelectedIndex !== this.selectedIndex) {
-            this.updateValue(true);
+            if (initialSelectedIndex !== this.selectedIndex) {
+                this.updateValue(true);
+            }
         }
 
+        this.open = commitValueThenClose ? false : this.open;
```
jattasNI, line 654: nit: I'd vote for this since it removes ambiguity about whether setting `this.open` to its current value has any side effects. But admit it's at the cost of more lines so won't complain if you reject this suggestion.
```
if (comitValueThenClose) { 
    this.open = false; 
}
```

jattasNI (approved):

packages/nimble-components/src/select/index.ts
rajsite: +1 on being clearer about sideEffects. If no sideEffect is expected I like Jesse's suggestion. If one is expected we should add a comment about it

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
atmgrifter00: I'm not entirely worried about this, as often will be the case the actual filtering mechanism will be in a backend implementation that can not use that filter. That is, the code in question is likely to be replaced with an http request.

change/@ni-nimble-angular-0af6f496-c35f-451a-880c-98604b7dbadd.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
jattasNI: Oops, I meant for this comment to say "PR title", not description. @atmgrifter00 I just went ahead and changed that but feel free to update further.

PR #2221, opened 2024-06-27
Update `SelectPageObject` `clickOption` to only deal with visible options.

atmgrifter00:
# Pull Request

## 🤨 Rationale

While doing the integration work to begin leveraging a select with grouping in SLE, I recognized that an existing page object had an API for selecting options by index, where the index was related to only the visible options in the dropdown. I feel this is more intuitive for a pageObject API than what we currently had in our own, so I'm updating our pageObject implementation to follow this.

## 👩‍💻 Implementation

Simply finding the visible options with which to relate the provided `index` parameter to `clickOption` against.

## 🧪 Testing

Updated tests that had to adjust the index they were using to align with only the visible options.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/select/testing/select.pageobject.ts
+++ b/packages/nimble-components/src/select/testing/select.pageobject.ts
@@ -119,7 +119,8 @@ export class SelectPageObject {
         if (!selectedOption) {
             throw new Error('No option is selected to click');
         }
-        this.clickOption(this.selectElement.options.indexOf(selectedOption));
+        const visibleOptions = this.getVisibleOptions();
+        this.clickOption(visibleOptions.indexOf(selectedOption));
     }
 
     public async clickFilterInput(): Promise<void> {
@@ -131,13 +132,14 @@ export class SelectPageObject {
     }
 
     public clickOption(index: number): void {
-        if (index >= this.selectElement.options.length) {
+        const visibleOptions = this.getVisibleOptions();
+        if (index >= visibleOptions.length) {
             throw new Error(
                 '"index" greater than number of current displayed options'
             );
         }
 
-        const option = this.selectElement.options[index]!;
+        const option = visibleOptions[index]!;
         option.scrollIntoView();
         option.click();
     }
@@ -321,6 +323,12 @@ export class SelectPageObject {
 
     public getFilterInputText(): string {
         return this.selectElement.filterInput?.value ?? '';
+    }
+
+    private getVisibleOptions(): ListOption[] {
+        return this.selectElement.options.filter(
+            o => !((o as ListOption).hidden || (o as ListOption).visuallyHidden)
+        ) as ListOption[];
     }
 
     private getFilterInput(): HTMLInputElement | null | undefined {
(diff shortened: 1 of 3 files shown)


packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -131,13 +132,14 @@ export class SelectPageObject {
     }
 
     public clickOption(index: number): void {
```
jattasNI, line 134: Can you add a JSDoc comment that notes this behavior? And also mention what visible is defined to mean from the client's point of view. Not `hidden` and matches the filter? (I can't recall if `visually-hidden` is set by things other than filtering)

packages/nimble-components/src/select/testing/select.pageobject.ts
```diff
@@ -131,13 +132,14 @@ export class SelectPageObject {
...
             throw new Error(
                 '"index" greater than number of current displayed options'
             );
         }
 
-        const option = this.selectElement.options[index]!;
+        const option = visibleOptions[index]!;
         option.scrollIntoView();
         option.click();
     }
```
jattasNI, line 145: `clickOptionWithDisplayText(displayText: string)` below is using `clickOption()` but with an index into the entire options array, not the visible options array, so I think its behavior is now wrong.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -994,7 +994,7 @@ describe('Select', () => {
                 expect(element.value).toBe('one');
 
                 await pageObject.openAndSetFilterText('T'); // Matches 'Two' and 'Three'
-                pageObject.clickOption(2); // index 2 matches option with 'Three' text
+                pageObject.clickOption(1); // index 1 matches option with 'Three' text
```
jattasNI, line 997: Should we just change these to do `clickOptionWithDisplayText()`? It'd be more readable IMO (could remove the comment).

What's the use case for `clickOption()` via index for a client? Should we just make it `internal`?

atmgrifter00: The use case that is interesting for a client is when the display text is the same for the two options that they want to distinguish between. This can happen practically when using groups, so a client team wants to be able to click particular options that have the same name but are under different groups (for example). Using the index is an easy way to do this. We _could_ create a more direct API to click an option with a particular group parent, but at the moment I don't feel strongly compelled to do so, but would if you think that better.

jattasNI: Ah that makes sense. I'm good to keep it public then and agree that we can hold off on writing `clickOptionWithGroupAndDisplayText()`.

I still think `clickOptionWithDisplayText()` would generally be preferred for readability. If you agree then we could consider:
1. noting the use cases in the docs for both functions
2. updating these tests to use the preferred method (I didn't check if these use grouping)
3. ensure we still have good test coverage for `clickOption()` if we use it less in this file

jattasNI (approved):

rajsite (approved):

PR #2224, opened 2024-06-28
Allow text column and anchor column to be sorted by custom data field

mollykreis:
# Pull Request

## 🤨 Rationale

Add ability to sort a text column or anchor column by a custom data field. This implements the capability described in #2206.

This PR only includes the nimble-components changes. Angular and Blazor integration will be done in a separate PR.

## 👩‍💻 Implementation

- Create `mixinCustomSortOrderColumnAPI` that does the following for columns using it
    - Adds a `sort-by-field-name` attribute
    - Requires a validator with the `invalidCustomSortWithGrouping` flag
    - Updates the validator's `invalidCustomSortWithGrouping` flag based on the values of `sortByFieldName`, `columnInternals.groupIndex`, and `columnInternals.groupingDisabled`
- Use the mixin on the text column and anchor column
- Update storybook stories for text column and anchor column

## 🧪 Testing

- New unit tests for mixin

## ✅ Checklist


mollykreis:

packages/storybook/src/nimble/table-column/anchor/table-column-anchor.stories.ts
```diff
@@ -40,41 +41,66 @@ export default metadata;
 const simpleData = [
     {
         firstName: 'Ralph',
+        firstNameSortIndex: 3,
         lastName: 'Wiggum',
+        lastNameSortIndex: 3,
+        fullName: 'Ralph Wiggum',
```
rajsite, line 47: What about adding a field like address instead? Like Ralph is at `732 Evergreen Terrace` and address is something one could argue is more useful to sort by the street name than alphanumeric

packages/storybook/src/nimble/table-column/anchor/table-column-anchor.stories.ts
```diff
@@ -40,41 +41,66 @@ export default metadata;
...
     hrefFieldName: string;
     appearance: keyof typeof AnchorAppearance;
     underlineHidden: boolean;
     placeholder: string;
+    sortByFieldName: 'firstNameSortIndex' | 'lastNameSortIndex';
+    checkValidity: never;
+    validity: never;
 }
 
+const validityDescription = `Readonly object of boolean values that represents the validity states that the column's configuration can be in.
```
m-akinc, line 99: We repeat the main part of this description in a bunch of files. We should use a shared string that we just append the properties list to.

packages/storybook/src/nimble/table-column/anchor/table-column-anchor.stories.ts
```diff
@@ -132,6 +167,33 @@ export const anchorColumn: StoryObj<AnchorColumnTableArgs> = {
             description:
                 'The placeholder text to display when the label and href are both `undefined` or `null` for a record.',
             table: { category: apiCategory.attributes }
+        },
+        sortByFieldName: {
+            name: 'sort-by-field-name',
+            description:
+                'Set this attribute to identify a field to sort the column by. If not set, the column will sort by the `field-name` field. It is invalid to group by a column with `sort-by-field-name` configured.',
```
m-akinc, line 174:
```suggestion
                'Set this attribute to identify a field to sort the column by. If not set, the column will sort by the `label-field-name` field. It is invalid to group by a column with `sort-by-field-name` configured.',
```

m-akinc, line 174: Should state the expected data type.

packages/nimble-components/src/table-column/anchor/models/table-column-anchor-validator.ts
m-akinc, line 1: Should we share this class (and flag names) instead of duplicating it for two different columns?

m-akinc (approved):

packages/nimble-components/src/table-column/anchor/models/table-column-anchor-validator.ts
mollykreis: Maybe? I was on the fence about it because the validation flag is associated with a mixin, which, at least in my mind, means that the validation logic associated with it would make more sense in a validator mixin rather than a base validator class so that a column's validator could use the "custom column sort" validation without having to change its class hierarchy. However, coming up with a validator mixin pattern seems like over engineering at this point in time, especially because the classes are virtually empty since none of the validation logic actually happens in the validators. A validator using custom sorting only needs to add the `invalidCustomSortWithGrouping` flag to their list of validity flags.

packages/nimble-components/src/table-column/base/models/column-validator.ts
```diff
@@ -22,6 +22,30 @@ export class ColumnValidator<
         return this.getValidationFlags();
     }
 
+    public override track(
```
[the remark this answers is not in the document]
mollykreis: I agree, that's a lot better. I've updated the `Tracker` class and added tests for the changes.

packages/storybook/src/nimble/table-column/anchor/table-column-anchor.stories.ts
mollykreis: I've added a shared string that takes care of the entire description. I've used it in the text column and anchor column. If we're happy with the change, I can update other stories to use the new string as well.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,93 @@
...
+                this.columnInternals.operandDataRecordFieldName = this.sortByFieldName;
+                this.columnInternals.sortOperation = TableColumnSortOperation.basic;
+            } else {
+                this.columnInternals.operandDataRecordFieldName = this.getDefaultSortFieldName();
+                this.columnInternals.sortOperation = this.getDefaultSortOperation();
+            }
+        }
+
+        /** @internal */
+        public handleChange(_source: unknown, args: unknown): void {
```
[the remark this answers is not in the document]
mollykreis: hmm...interesting question. The tricky thing here is that if `this.updateOperandDataRecordFieldName()` is not called, then the column class needs to set `this.columnInternals.operandDataRecordFieldName = <newFieldName>;`, and that has nothing to do with custom sorting behavior. Therefore, it doesn't really feel like a private implementation detail of the mixin that should be completely hidden away from a column class, so I think it would make more sense to leave the code as-is. However, I could also see the argument that it's better to have the code in a single location, and column tests would (hopefully) catch any problems if the internals of the mixin changed such that the `this.columnInternals.operandDataRecordFieldName` wasn't updated at the correct time.

packages/nimble-components/src/table-column/text/index.ts
```diff
@@ -25,25 +26,45 @@ declare global {
...
+        return TableColumnSortOperation.localeAwareCaseSensitive;
+    }
+
+    protected override getColumnInternalsOptions(): ColumnInternalsOptions<TableColumnTextValidator> {
         return {
             cellRecordFieldNames: ['value'],
             cellViewTag: tableColumnTextCellViewTag,
             groupHeaderViewTag: tableColumnTextGroupHeaderViewTag,
             delegatedEvents: [],
             sortOperation: TableColumnSortOperation.localeAwareCaseSensitive,
```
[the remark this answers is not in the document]
mollykreis: Actually, making them property getters seems to break things with the mixin. I'm going to change them back to functions, at least for now.

In case anyone is interested, the problem I'm running into is this:
When the base class for a column is generated with the mixins, the type for it ends up containing the following lines:
```ts
readonly defaultSortFieldName: string | undefined;
readonly defaultSortOperation: TableColumnSortOperation;
```
even though those properties are defined in the mixin as:
```ts
public abstract get defaultSortFieldName(): string | undefined;
public abstract get defaultSortOperation(): TableColumnSortOperation;
```
As a result, trying to override them in the concrete class as property getters causes a TypeScript compile error due to [this TypeScript rule](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#properties-overriding-accessors-and-vice-versa-is-an-error).

packages/storybook/src/nimble/table-column/anchor/table-column-anchor.stories.ts
m-akinc: I like it. The only nit I have is that referring to the element by its tag name feels a little mechanical/unnatural. (We didn't use the tag name in any of the validity descriptions initially.) We could instead make the parameter `colloquialName` and pass in values like "column", "table", "editor", etc.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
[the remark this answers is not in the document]
m-akinc: Ah, I see what you're getting at. So if we wanted to go in the direction I suggested, we would really want to move `getDefaultSortFieldName`, `getDefaultSortOperation`, and a trivial implementation of `updateOperandDataRecordFieldName` to the `SortableColumn` mixin class, then make `CustomSortOrderColumn` extend it (actually, we can't extend it, but instead we add those member names to the `Pick<> at the top of the file?)`. Column types would then no longer call `this.updateOperandDataRecordFieldName()` or set `this.columnInternals.operandDataRecordFieldName = <newFieldName>`. However, they would all have to implement those two abstract members (or read-only properties). I think I like that.

The one reamaining loose end I can think of is that if a column that is groupable but not sortable would still have to set `operandDataRecordFieldName` itself. That doesn't seem too bad to me, especially given how (presumably) rare it would be to support only grouping without sorting.

All that said, there's nothing wrong with the code as it is now, so it's just a suggestion.

packages/nimble-components/src/table-column/text/index.ts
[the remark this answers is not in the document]
m-akinc: I tried adding similar abstract properties to the `SortableColumn` mixin class locally, and I'm not seeing the errors you are. I'm not sure what I'm doing differently.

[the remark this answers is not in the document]
mollykreis: I think I didn't see the problems when I built `nimble-components`. It wasn't until the full build ran and things tried to build on top of `nimble-components` such as `spright` and `nimble-angular`.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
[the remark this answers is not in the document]
mollykreis: What I realized reading your comment is that we don't enforce that the `CustomSortOrderColumn` mixin is only used by columns that also use the `SortableColumn` mixin. I don't think that's a huge deal, especially since I'm not sure how we'd go about enforcing that two mixins were used together.

But, that being said, I'm not sure how we'd `Pick` the fields you mention from the `CustomSortOrderColumn` since we can only `Pick` things from the base `TableColumn` class.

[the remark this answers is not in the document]
m-akinc: > we can only Pick things from the base TableColumn class.

Oh, right. Hmm. How important is it to keep sortability and custom-sortability separate? I can't really think of a reason why we'd need to disallow custom sorting for an otherwise sortable column. If there isn't one, we could just merge the two mixins.

This is sort of tangential, but when thinking of merging the mixins, it occurred to me that `sortByFieldName` might be better named `customSortFieldName`. Otherwise, it's kind of unintuitive that, for example, `sortByFieldName` is unset, and yet there is a field that you're sorting by.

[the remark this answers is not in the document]
mollykreis: > I can't really think of a reason why we'd need to disallow custom sorting for an otherwise sortable column.

Our philosophy tends to be that we should try to limit what a client can do unless we have a use case for it. I don't think we have any reason to allow custom sorting for numeric columns (duration, date, number-text). Those columns are all sortable, but I think a custom sort order would be pretty nonsensible.

As far as the name of the property, I don't have a strong opinion either way. @rajsite and @jattasNI, any opinions on the name `sortByFieldName` vs `customSortFieldName`.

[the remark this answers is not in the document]
jattasNI: I don't think `customSortFieldName` is significantly better than `sortByFieldName` so I vote to keep the decision unchanged from what was in the HLD.

change/@ni-nimble-components-a398657e-0eca-4a73-9948-71fd7466f4bd.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The test coverage looks quite solid, but it'd probably be worth scanning [table-sorting.spec.ts](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/tests/table-sorting.spec.ts) to see if there are any corner cases worth adding coverage for. I don't think we need to duplicate every test there with a custom sort order variation, but here are some ideas that could be valuable:
1. a test that asserts the actual order that rows end up in. I believe existing tests only cover TanStack configuration, unless I missed something.
2. same as above, but with two rows with equal custom sort index (we decided we're fine with whatever TanStack does in this case, but it could be good to assert about that behavior so we're aware if it changes.
3. a custom sort order equivalent of `can sort with have field names containing "."`. I don't recall why we added that test but presumably a similar corner case exists for the new custom sort order field.
4. two columns that point to the same custom sort order field. Nothing in particular makes me think this would fail, just a case that could be goofy depending on implementation choices.
5. any other cases you can think of after scanning the existing tests (hierarchy, sort order, etc)

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,96 @@
...
+            if (args === 'groupingDisabled' || args === 'groupIndex') {
+                this.updateCustomColumnSortingValidity();
+            } else if (args === 'dataRecordFieldNames') {
+                this.updateOperandDataRecordFieldName();
+            }
+        }
+
+        /** @internal */
+        public updateCustomColumnSortingValidity(): void {
```
jattasNI, line 66: Corner case to consider: a custom sort order but sorting is disabled. Should grouping be allowed in this case? It's admittedly a weird configuration, but a couple (weak) arguments in favor of supporting it:
1. we check whether grouping is disabled as part of the validity calculation so checking sorting feels symmetric
2. I can imagine a client wanting to toggle between custom sorting and grouping while leaving the configuration for both of those states set the whole time

If you see arguments against considering this configuration as valid, I don't feel strongly about supporting it. We could always add support if a client requests it.

jattasNI (commented):
Great job with the implementation and the buddy feedback! Just a couple corner cases to think through.

change/@ni-nimble-components-a398657e-0eca-4a73-9948-71fd7466f4bd.json
mollykreis: I hadn't added tests for these cases since none of the table code changed at all, and I also didn't change any of the ways we interact with TanStack. We're exercising the table's code path for `columnInternals.operandDataRecordFieldName` changing, which we already have tests for. I can't think of anything special to test for, other than `columnInternals.operandDataRecordFieldName` being set to the correct value, and the mixin tests verify that.

jattasNI: Thanks for that explanation. I think the argument that there are no table changes so we don't need table test changes is compelling, so I'm ok to close this with no action.

jattasNI (approved):

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
rajsite: Chatted with molly offline and the issue is that the operand config is used in both sorting and grouping so even though sorting may be temporary disabled, the issue is the same in that the right way to apply a configured sortByFieldName to grouping.

mollykreis: I've reverted the change I initially made in response to this discussion thread. If `sortByFieldName` is configured with grouping, the column is invalid, regardless of whether or not sorting is enabled or disabled.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,96 @@
...
+>(base: TBase) {
+    /**
+     * The Mixin that provides a concrete column with the API to allow sorting
+     * by a field other than the default for that column.
+     */
+    abstract class CustomSortOrderColumn extends base {
+        public sortByFieldName?: string;
+
+        /** @internal */
+        public notifier?: Notifier;
```
rajsite, line 28: Need more uniqueness to some of these properties. They are defined on the prototype chain so this would be exposed as textColumnRef.notifier and and textColumnRef.handleChange. Seems like it would be pretty easy to mixin on a class and conflict / override properties with these names.

We may want a separate handler object instead of adding handleChange to the class we are mixing in on.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,96 @@
...
+            } else if (args === 'dataRecordFieldNames') {
+                this.updateOperandDataRecordFieldName();
+            }
+        }
+
+        /** @internal */
+        public updateCustomColumnSortingValidity(): void {
+            const hasCustomColumnSorting = typeof this.sortByFieldName === 'string';
+            const isGrouped = !this.columnInternals.groupingDisabled
+                && typeof this.columnInternals.groupIndex === 'number';
```
rajsite, line 69: Do we really need the groupIndex check? Can't we just check if groupingDisabled is false?

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,96 @@
...
+                    'invalidCustomSortWithGrouping'
+                );
+            }
+        }
+
+        /** @internal */
+        public abstract getDefaultSortFieldName(): string | undefined;
+
+        /** @internal */
+        public abstract getDefaultSortOperation(): TableColumnSortOperation;
```
rajsite, line 86: I think I find this pattern a bit confusing / challenging to follow. The contract of who owns what state and what to do when changes happen is a little unclear.

Didn't walkthrough it completely but my hunch is that it'll be a bit cleaner if the concrete / leaf class owns the configuration of all the columnInternals properties (operandDataRecordFieldName, sortOperation, fieldName) and is leveraging the mixin to ask for advice instead of the mixin configuring some of the state and the concrete class other bits of state.

Something like:
```ts
abstract class CustomSortOrderColumn extends base {
    public sortByFieldName?: string;

    public getResolvedOperandDataRecordFieldName (initialOperandFieldName) {}
    public getResolvedSortOperation (initialSortOperation) {}

    abstract handleCustomSortOrderConfigurationChanged{} /* definitely not committed to this name*/
    /* still handle validity stuff and attr handling stuff*/
}


class Text {
    protected override getColumnInternalsOptions():
        return {
            ...
            sortOperation: this.getResolvedSortOperation(TableColumnSortOperation.localeAwareCaseSensitive),
        };
    }
      
    public override handleCustomSortOrderConfigurationChanged (){
        this.columnInternals.dataRecordFieldNames = [this.fieldName] as const;
        this.columnInternals.operandDataRecordFieldName = this.getResolvedOperandDataRecordFieldName(this.filedName)
        this.columnInternals.sortOperation = this.getResolvedSortOperation(localAwareCaseInsenitive)
    }

    override fieldNameChanged(): void {
        this.columnInternals.dataRecordFieldNames = [this.fieldName] as const;
        this.columnInternals.operandDataRecordFieldName = this..getResolvedOperandDataRecordFieldName(this.filedName)
        this.columnInternals.sortOperation = this.getResolvedSortOperation(localAwareCaseInsenitive)
    }
   
}
```

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
mollykreis: Sure, I can make that change. That's probably more robust since it will catch any validation errors if the column is groupable, not just if it happened to be grouped at the time.

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,110 @@
...
+                );
+            } else {
+                this.customSortOrderColumnNotifier?.unsubscribe(
+                    this.customSortOrderColumnChangeHandler
+                );
+                this.customSortOrderColumnNotifier = undefined;
+            }
+        }
+
+        public abstract handleSortByFieldNameChange(): void;
```
rajsite, line 74: nit / easily convinced otherwise: the implementers don't really care directly about SortByFieldName and in the future maybe there are other potential properties working together here.
What about `handleSortConfigurationChange`?

packages/nimble-components/src/table-column/anchor/index.ts
```diff
@@ -158,6 +174,13 @@ export class TableColumnAnchor extends mixinGroupableColumnAPI(
             placeholder: this.placeholder
         };
     }
+
+    private updateOperandDataRecordFieldName(): void {
```
rajsite, line 178: this is doing more than the name suggests, what about  `updateColumnInternalsSortConfig`?
Same applies to text column

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        /** @internal */
+        public handleCustomSortOrderColumnChange(
+            _source: unknown,
+            args: unknown
+        ): void {
+            if (args === 'groupingDisabled') {
+                this.updateCustomColumnSortingValidity();
+            }
+        }
```
rajsite, line 84: Is this simplified (i.e. avoid the notifier, handler, and all the conditional tracking only if sortBy is configured) if we use groupingDisabledChanged and an optional super invocation?

Something like:

```ts
public groupingDisabledChanged(_source, args) {
    super.groupingDisabledChanged?.(_source, args);
    this.updateCustomColumnSortingValidity();
}
```

rajsite (approved):

packages/nimble-components/src/table-column/mixins/custom-sort-order.ts
mollykreis: I don't see a straightforward way to achieve what you are suggesting. `groupingDisabledChanged` is not part of the super class, so I can't invoke it from the mixin, even optionally.

mollykreis: @rajsite - wanted to double check that you are fine with me merging this PR without making any changes for this suggestion.

rajsite: Current implementation is valid. I'm guessing it's a typing issue not a runtime issue but not worth spending extra cycles on it. LGTM and thanks for double checking.

rajsite (approved):

PR #2229, opened 2024-07-01
Avoid another instance of intermittent test component id collision

m-akinc:
# Pull Request

## 🤨 Rationale

I thought I had resolved this in #2087, but I just ran into another case of this Angular ID collision issue in [this build](https://github.com/ni/nimble/actions/runs/9751655260/job/26913710201?pr=2193). This time rather than being a collision for components `AbsoluteSimpleLinkCmp` and `RelativeLinkCmp`, it's a collision between two different definitions of `RelativeLinkCmp`:

```
Failed: NG0912: Component ID generation collision detected. Components 'RelativeLinkCmp' and 'RelativeLinkCmp' with selector 'link-cmp' generated the same component ID. To fix this, you can change the selector of one of those components or add an extra host attribute to force a different ID. Find more at https://angular.io/errors/NG0912
```

## 👩‍💻 Implementation

I'm still not clear on why this is intermittent, but I'm content to just add a uniquifying `:not(body)` to the selector for one of the `RelativeLinkCmp` components that can conflict.

## 🧪 Testing

None

## ✅ Checklist

### DIFF
--- a/packages/angular-workspace/nimble-angular/src/thirdparty/directives/tests/router_integration.spec.ts
+++ b/packages/angular-workspace/nimble-angular/src/thirdparty/directives/tests/router_integration.spec.ts
@@ -7146,7 +7146,9 @@ class DummyLinkCmp {
 class AbsoluteSimpleLinkCmp {
 }
 
-@Component({selector: 'link-cmp', template: `<a [routerLink]="['../simple']">link</a>`})
+// [Nimble] add irrelevant :not(body) to selector to force distinct component ID generation
+//  Solution based on suggestion from https://angular.io/errors/NG0912#debugging-the-error
+@Component({selector: 'link-cmp:not(body)', template: `<a [routerLink]="['../simple']">link</a>`})
 class RelativeLinkCmp {
 }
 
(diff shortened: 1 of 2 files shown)


packages/angular-workspace/nimble-angular/src/thirdparty/directives/tests/router_integration.spec.ts
```diff
@@ -7146,7 +7146,9 @@ class DummyLinkCmp {
 class AbsoluteSimpleLinkCmp {
 }
 
-@Component({selector: 'link-cmp', template: `<a [routerLink]="['../simple']">link</a>`})
+// [Nimble] add irrelevant :not(body) to selector to force distinct component ID generation
+//  Solution based on suggestion from https://angular.io/errors/NG0912#debugging-the-error
+@Component({selector: 'link-cmp:not(body)', template: `<a [routerLink]="['../simple']">link</a>`})
```
rajsite, line 7151: instead of whack-a-mole can we uniqify all the `selector: 'link-cmp'` selectors?
choose a pattern that's less random elements, maybe `link-cmp:not([nimble-unused-0])`, `link-cmp:not([nimble-unused-1])`, etc

m-akinc: I think the rest of the components with that selector have [different enough template characteristics](https://github.com/ni/nimble/pull/2087#discussion_r1596098105) that they generate different IDs (note that we've only ever seen collisions involving `RelativeLinkCmp` and/or `AbsoluteSimpleLinkCmp`), but it can't hurt to play it safe.

rajsite (approved):

Issue #1451, opened 2023-08-24
nimble-menu-button doesn't close when scrolled out of view


mollykreis:
# 🐛 Bug Report

The nimble-menu-button stays open when it is scrolled under another part of the page. This means the button itself is no longer visible, but the menu itself renders on top of the other content.

As an example, you can see that the menu is showing on top of the gray properties panel even though the menu is not visible.
![image](https://github.com/ni/nimble/assets/20542556/bded545a-fb74-4c64-8482-753b29eaaa59)

We should consider fixing this in a holistic manner that would also address other potentially problematic components, such as the nimble-select and nimble-combobox.

## 🤔 Expected Behavior

The menu associated with the menu-button should close when the menu-button is no longer visible.

## 😯 Current Behavior

The menu-button gets hidden behind other page content when the page scrolls, but the menu is shown on top of the other content. This leads to an odd "floating" menu that doesn't seem associated with anything on the page.

## 💁 Possible Solution

Use an intersection observer to close an anchored region whenever the anchor element goes out of view.

## 🔦 Context

This was originally filed as a bug within SLE: https://dev.azure.com/ni/DevCentral/_workitems/edit/2492542/

We should remove the workaround added to fix the bug once a fix is available within nimble.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

suseendran-ni:
The same issue we encountered while developing the `mention-list` dropdown in this [PR](https://github.com/ni/nimble/pull/1694)

Currently, we have handled it by creating an intersection observer for the `@` element in the editor. (Similar to the handling we did for this bug in `sl-comments-library`)
When the `@` character is intersected, the `mention-list` popup will be closed.

But during our discussion, @rajsite suggested not to implement the intersection observer as a fix. Instead, all the popups must be closed whenever there is a scroll event that happens within the page. This behavior must be consistent across all the nimble components.

rajsite:
> Instead, all the popups must be closed whenever there is a scroll event that happens within the page. This behavior must be consistent across all the nimble components.

[...]

PR #2236, opened 2024-07-02
Confine rich-text mention listbox to viewport

### LINKED ISSUE #2217: Rich-text mention dropdown should limit height to viewport
# 🐛 Bug Report

All Nimble dropdowns should be height-limited to avoid going out of the viewport. The rich-text mention dropdown is the only dropdown that doesn't yet do this.

Because that dropdown is limited to 5.5 options, it's very unlikely to extend outside of the viewport, but as long as it is quick/easy to add the viewport-limiting logic, we should.

## 🔦 Context

https://github.com/ni/nimble/pull/2189#discussion_r1655350758

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2217

## 👩‍💻 Implementation

Upon opening, the rich-text mention listbox now sets a `availableViewportHeight` property based on the distance between the bottom of the anchor element and the bottom of the window. Like with select and combobox, that property is bound to a CSS variable, which factors into the `max-height` for the listbox. I had to parameterize that CSS calculation with a couple new CSS variables to account for differences between the mention listbox and select/combobox.

Also, incidental to this change, I'm modifying the other mention listbox test that was disabled due to intermittency (#1891) in a way that might fix the intermittency. I'm re-enabling in hopes that it does.

## 🧪 Testing

Created a new test for the rich-text mention listbox, plus modified the existing select test to ensure that the test wasn't passing due to the default 10.5 item height limit.

## ✅ Checklist


packages/nimble-components/src/rich-text/mention-listbox/index.ts
```diff
@@ -119,6 +138,7 @@ export class RichTextMentionListbox extends FoundationListbox {
                     displayName: this.firstSelectedOption.text
                 };
                 this.emitMentionSelected(mentionDetail);
+                this.setOpen(false);
```
m-akinc, line 141: Previously, `emitMentionSelected` was calling `setOpen`, which doesn't seem right. I moved the `setOpen` calls out to the call sites.

packages/nimble-components/src/rich-text/mention-listbox/tests/mention-listbox.spec.ts
```diff
@@ -45,15 +49,24 @@ describe('RichTextMentionListbox', () => {
...
     // Intermittent: see https://github.com/ni/nimble/issues/1891
     xit('should scroll the selected option into view when opened', async () => {
         const model = new Model();
         const { connect, disconnect } = await setup500Options(model);
         await connect();
-        model.mentionListbox.show({
-            anchorNode: model.anchorDiv,
-            filter: ''
-        });
+        await showAndWaitForOpen(model.mentionListbox, model.anchorDiv);
```
m-akinc, line 69: I'm not sure if waiting for the anchored region's `loaded` event will fix the intermittency here, but I updated this test for consistency.

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -659,6 +659,10 @@ describe('Select', () => {
             const { element, connect, disconnect } = await setup500Options();
             await connect();
             await clickAndWaitForOpen(element);
+            element.listbox.style.setProperty(
+                '--ni-private-listbox-num-visible-options',
+                '10000'
+            );
```
m-akinc, line 665: Make sure we're exercising the viewport limit rather than the visible option limit.

m-akinc:
@mollykreis could I get a buddy, please?

packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -138,18 +138,24 @@ export const styles = css`
         flex-direction: column;
         overflow-y: auto;
         width: 100%;
+        --ni-private-listbox-num-visible-options: 10.5;
```
mollykreis, line 141: We don't tend to use abbreviations, such as "num" here. What would you think about _--ni-private-listbox-visible-options-count_ instead?

packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -138,18 +138,24 @@ export const styles = css`
...
+        --ni-private-listbox-anchor-element-gap: ${smallPadding};
         --ni-private-listbox-padding: ${smallPadding};
         --ni-private-listbox-filter-height: 0px;
         --ni-private-listbox-loading-indicator-height: 0px;
         max-height: min(
             calc(
-                ${smallPadding} + 2 * ${borderWidth} + ${controlHeight} * 10.5 +
+                var(--ni-private-listbox-anchor-element-gap) + 2 *
```
mollykreis, line 148: I find that the wrapping here makes visually parsing & understanding the logic really difficult. I assume prettier/eslint is fighting you here? Would it be reasonable to disable those for these calculations so that you could write it in a way where the different pieces of the calculation are more apparent. I'm thinking something like this:
```
gap +
2 * borderWidth +
padding +
controlHeight * numVisibleOptions +
filterHeight +
loadingIndicatorHeight
```

mollykreis (approved):

packages/nimble-components/src/rich-text/mention-listbox/tests/mention-listbox.spec.ts
mollykreis: Would it be worth re-enabling the test and seeing if the intermittency is gone now?

packages/nimble-components/src/rich-text/mention-listbox/index.ts
mollykreis: An alternative would be to rename `emitMentionSelected` to something a more generic like `handleMentionSelected` and have it emit & call `setOpen`. Then you wouldn't have to call `setOpen(false)` right after each call to `emitMentionSelected`.

m-akinc: I suppose leaving a method that does nothing but forward the argument to another method is silly. I could move `setOpen` back and move the event detail creation into it as well, but honestly, I think we can just do without this function that has two distinct responsibilities and is called in only two places.

packages/nimble-components/src/rich-text/mention-listbox/tests/mention-listbox.spec.ts
m-akinc: Guess it couldn't hurt. Let's do it.

change/@ni-nimble-components-905600e9-5f08-47e1-8b5f-345e25a2ccbc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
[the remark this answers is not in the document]
m-akinc: Ah, yes it is. I noticed that `max-height` we were setting in the shared styles was short by one listbox padding (4px).

change/@ni-nimble-components-905600e9-5f08-47e1-8b5f-345e25a2ccbc.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: This reveals a potentially problematic behavior where the dropdown can appear really small due to limited space and then scrolling the page keeps it small despite more space being available. This isn't a new problem and it exists for select (and probably other dropdowns) already. 

I think the fix is #1451 where we have to do something like dismiss these popups when the page or container scrolls. I don't think we need to do anything about it in this PR, I'm mostly just raising awareness and establishing a link to that work item.

If we wanted to do something in this PR, maybe we should set a min-height of at least one item? Or open the dropdown above if there's more space there? But those would probably just be band-aid fixes so probably not worth doing given that we'd have to apply them across all components.

https://github.com/ni/nimble/assets/10500124/5684daa0-1620-46bb-8160-e10b4e64fccb

jattasNI (approved):

rajsite (approved):

change/@ni-nimble-components-905600e9-5f08-47e1-8b5f-345e25a2ccbc.json
rajsite: Either way I don't think it needs to block this PR, fyi @m-akinc

PR #2242, opened 2024-07-08
Fix table phantom multi-select

### LINKED ISSUE #2222: Table select via keyboard sometimes behaves as if Shift is pressed when it isn't
# 🐛 Bug Report

In the SLE files app if I use keyboard navigation to select a table row, it sometimes behaves as if Shift is pressed (selecting multiple rows with a single checkbox toggle) even when it isn't.

I saw this a couple times in macOS Firefox but then after refreshing the page I haven't been able to reproduce it. I tried in other browsers and in Storybook and wasn't able to reproduce there. 

I'd recommend we do some code inspection to see if we can hypothesize any way that this could happen and that another person try to reproduce it too. If we find something then great, if not I have no problem closing this with no changes.

## 💻 Repro or Code Sample

https://dev.lifecyclesolutions.ni.com/files/
1. tab to the table
2. arrow to a row's selection checkbox
3. space to toggle it
4. arrow to a different row that's a few rows away
5. space to toggle it

## 🤔 Expected Behavior

6. the new row is selected but the ones in between aren't

## 😯 Current Behavior

6. sometimes all the rows in between are selected

## 🔦 Context
https://dev.lifecyclesolutions.ni.com/files/

## 🌍 Your Environment

macOS Firefox

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2222

Because the event emitted by the row (`row-selection-toggle`) does not include details about whether shift (or any other modifier key) was pressed as part of the selection toggle--and it can't, because the checkbox change event doesn't know--the table has to track the state of the shift key by listening for `keyup` and `keydown` events on the document. This breaks down if the shift key is let go while focus is outside the document (e.g. on the browser chrome). At that point, the table still thinks the shift key is depressed and behaves accordingly.

## 👩‍💻 Implementation

We were previously only updating the table's `documentShiftKeyDown` flag when the `shift` key was pressed/unpressed. But we can also look at the `shiftKey` flag of the event data for other keys' events. For example, if the `space` key is pressed to select a row, and `event.shiftKey` is `false`, then we can update `documentShiftKeyDown` to prevent the incorrect behavior.

However, we still have to handle the case where the selection is made by mouse click rather than pressing `space`. In that scenario, we may not get any `keydown`/`keyup` events that would clue us in that `shift` is no longer held down. My solution is to add a `blur` handler on the window that clears `documentShiftKeyDown` when the focus leaves the window. It would be better if we could detect focus leaving the document rather than window, but the document doesn't get a `blur` event, and `focusout` bubbles up from all elements in a document. Generally, if the window loses focus, we can expect that the user is no longer in the process of making a selection. However there are contrived use cases where this can result in the wrong behavior. For example:
1. an element in the window has focus
2. hold down `shift`
3. press `tab` repeatedly (`shift` is still pressed) until focus leaves the window, cycles through everything else, and returns to the table
4. click to select a row. The table will behave as if `shift` is not pressed, even though it is.

Note that in step 4, if `space` is pressed rather than mouse-clicking, the `keydown` for `space` will update `documentShiftKeyDown`, and things will behave properly.

## 🧪 Testing

Wrote a couple unit tests and manually tested in Storybook.

## ✅ Checklist


m-akinc:
@mollykreis will you please buddy for me?

mollykreis (approved):

packages/nimble-components/src/table/index.ts
```diff
@@ -338,6 +338,7 @@ export class Table<
         this.viewport.addEventListener('scroll', this.onViewPortScroll, {
             passive: true
         });
+        this.addEventListener('blur', this.onBlur);
```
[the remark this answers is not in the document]
m-akinc: Well, focus would have to leave the entire document, not just the table, but I realized I don't actually have to do that to be able to write a test that simulates it. In the process of writing those tests, I realized my fix didn't handle the case where the table doesn't initially have focus (e.g. user shift+tabs from some random element out of the document, then tabs back to the table). I switched my handler from a `blur` handler on the table to a `focusout` handler on the document to take care of that.

packages/nimble-components/src/table/index.ts
```diff
@@ -341,6 +341,7 @@ export class Table<
         this.keyboardNavigationManager.connect();
         document.addEventListener('keydown', this.onKeyDown);
         document.addEventListener('keyup', this.onKeyUp);
+        document.addEventListener('focusout', this.onFocusOut);
```
rajsite, line 344: focusout is a bubbling event so this should get triggered if any child has a focusout (and the event bubble up): https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event 

Which from the PR description I don't think is the intention. Sounds like the intention is to be tracking the window focus. If that's the case I think we should listen for the non-bubbling `blur` event instead.

m-akinc: Good catch. Looks like the document does not get a `blur` event, but interestingly, the window does. I think that's good enough for our purposes. Might be some corner cases with tables inside tabs or iframes, but I'm not sure how we can do better.

packages/nimble-components/src/table/index.ts
```diff
@@ -341,6 +341,7 @@ export class Table<
         this.keyboardNavigationManager.connect();
         document.addEventListener('keydown', this.onKeyDown);
         document.addEventListener('keyup', this.onKeyUp);
+        window.addEventListener('blur', this.onBlur);
```
rajsite, line 344: for consistency can we just listen for all of these on window?

packages/nimble-components/src/table/index.ts
```diff
@@ -350,6 +351,7 @@ export class Table<
         this.viewport.removeEventListener('scroll', this.onViewPortScroll);
         document.removeEventListener('keydown', this.onKeyDown);
         document.removeEventListener('keyup', this.onKeyUp);
+        window.removeEventListener('blur', this.onBlur);
```
rajsite, line 354: also jic lets remove all of these from window

rajsite (approved):

packages/nimble-components/src/table/tests/table-selection.spec.ts
```diff
@@ -1431,6 +1432,59 @@ describe('Table row selection', () => {
...
+                        expect(selectedRecordIds).toEqual(
+                            jasmine.arrayWithExactContents([
+                                simpleTableData[0].id,
+                                simpleTableData[3].id
+                            ])
+                        );
+                    });
+
+                    it('selects only the rows that SPACE was pressed on', async () => {
+                        const downArrowKeyEvent = new KeyboardEvent('keydown', {
```
jattasNI, line 1461: nit: wanna use [the `sendKeyDownEvent` test utility](https://github.com/ni/nimble/blob/f47da47d1ee6104e9e4c913c8e1fbca321825e9d/packages/nimble-components/src/utilities/tests/component.ts#L8) in these new tests instead of constructing and dispatching event objects?

There are plenty of existing places we don't do this, but it is nice to centralize it and abstract away the details if possible.

jattasNI (approved):

packages/nimble-components/src/table/tests/table-selection.spec.ts
m-akinc: I wasn't excited about adding a bunch of extra `waitForUpdatesAsync` calls, so I made a `sendKeyDownEvents` variant to send a sequence of keystrokes then wait once.

PR #2245, opened 2024-07-08
Align column header based on cell content alignment

### LINKED ISSUE #1750: Table headers should be right-aligned when data is right-aligned
# 🐛 Bug Report

During the `nimble-table-column-number-text` definition we decided to maintain left-alignment of headers for right-aligned data. This was a mistake, and we should fix that by consistently aligning headers with the data.

See this [table best practices doc](https://medium.com/design-with-figma/the-ultimate-guide-to-designing-data-tables-7db29713a85a) for more.

## 💻 Repro or Code Sample

Screenshot 2024-01-11 at 8 47 46 AM

https://nimble.ni.dev/storybook/?path=/docs/components-table-column-number-text--docs

## 🤔 Expected Behavior

Table header text should be right-aligned when data is right-aligned

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #1750 

## 👩‍💻 Implementation

`TableHeader` now has an `alignment` property that it uses to organize/align the header contents. This property gets bound to a new `headerAlignment` property on `ColumnInternals`. Its type is a new enum, `TableColumnHeaderAlignment`, which has values `left` and `right`. There is already an alignment enum for the cell contents (`TextCellViewBaseAlignment`), but I don't think we should directly couple cell and header alignment. 

Currently the only column that can have right-aligned cell content is the number-text column. It now updates `columnInternals.headerAlignment` whenever its cell content alignment might have changed.

The referenced bug doesn't really answer the question of where sort/grouping indicators (and future menu dropdown) should go when right-aligning the header. For now, I have assumed we should shift them to the left side of the column label.

## 🧪 Testing

Created new unit test cases.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-48e9cade-64e1-47a5-a616-c2e2a930a2b1.json
+++ b/change/@ni-nimble-components-48e9cade-64e1-47a5-a616-c2e2a930a2b1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Align column header based on cell content alignment",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table-column/base/models/column-internals.ts
+++ b/packages/nimble-components/src/table-column/base/models/column-internals.ts
@@ -1,6 +1,10 @@
 import { uniqueId } from '@microsoft/fast-web-utilities';
 import { ViewTemplate, observable } from '@microsoft/fast-element';
-import { TableColumnSortDirection, TableFieldName } from '../../../table/types';
+import {
+    type TableColumnHeaderAlignment,
+    TableColumnSortDirection,
+    TableFieldName
+} from '../../../table/types';
 import type { TableCell } from '../../../table/components/cell';
 import {
     TableColumnSortOperation,
@@ -152,6 +156,12 @@ export class ColumnInternals<
      */
     @observable
     public hideHeaderIndicators = false;
+
+    /**
+     * How to align the header content.
+     */
+    @observable
+    public headerAlignment?: TableColumnHeaderAlignment;
 
     /**
      * @internal Do not write to this value directly. It is used by the Table in order to store
--- a/packages/nimble-components/src/table-column/base/models/column-internals.ts
+++ b/packages/nimble-components/src/table-column/base/models/column-internals.ts
@@ -1,7 +1,7 @@
 import { uniqueId } from '@microsoft/fast-web-utilities';
 import { ViewTemplate, observable } from '@microsoft/fast-element';
 import {
-    type TableColumnAlignment,
+    TableColumnAlignment,
     TableColumnSortDirection,
     TableFieldName
 } from '../../../table/types';
@@ -161,7 +161,7 @@ export class ColumnInternals<
      * How to align the header content.
      */
     @observable
-    public headerAlignment?: TableColumnAlignment;
+    public headerAlignment: TableColumnAlignment = TableColumnAlignment.left;
 
     /**
      * @internal Do not write to this value directly. It is used by the Table in order to store
--- a/packages/nimble-components/src/table-column/base/models/column-internals.ts
+++ b/packages/nimble-components/src/table-column/base/models/column-internals.ts
@@ -1,7 +1,7 @@
 import { uniqueId } from '@microsoft/fast-web-utilities';
 import { ViewTemplate, observable } from '@microsoft/fast-element';
 import {
-    type TableColumnHeaderAlignment,
+    type TableColumnAlignment,
     TableColumnSortDirection,
     TableFieldName
 } from '../../../table/types';
@@ -161,7 +161,7 @@ export class ColumnInternals<
      * How to align the header content.
      */
     @observable
-    public headerAlignment?: TableColumnHeaderAlignment;
+    public headerAlignment?: TableColumnAlignment;
 
     /**
      * @internal Do not write to this value directly. It is used by the Table in order to store
--- a/packages/nimble-components/src/table-column/number-text/index.ts
+++ b/packages/nimble-components/src/table-column/number-text/index.ts
@@ -12,7 +12,10 @@ import {
 } from '@microsoft/fast-element';
 import { styles } from '../base/styles';
 import { template } from './template';
-import type { TableNumberField } from '../../table/types';
+import {
+    TableColumnHeaderAlignment,
+    type TableNumberField
+} from '../../table/types';
 import { TableColumnTextBase, mixinTextBase } from '../text-base';
 import { TableColumnSortOperation } from '../base/types';
 import { tableColumnNumberTextGroupHeaderTag } from './group-header-view';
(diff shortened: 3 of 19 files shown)


packages/nimble-components/src/table/components/header/styles.ts
```diff
@@ -26,6 +26,10 @@ export const styles = css`
         cursor: default;
     }
 
+    :host(.right-align) {
+        justify-content: flex-end;
```
m-akinc, line 30: I considered `flex-direction: row-reverse`, but that would also reverse the order of sort/grouping indicators, which we don't want.

m-akinc:
@atmgrifter00 do you mind buddying?

packages/nimble-components/src/table/types.ts
```diff
@@ -217,6 +217,18 @@ export interface TableRowState<TData extends TableRecord = TableRecord> {
     resolvedRowIndex?: number;
 }
 
+/**
+ * @internal
+ *
+ * Alignment of column header content
+ */
+export const TableColumnHeaderAlignment = {
```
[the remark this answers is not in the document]
m-akinc: That probably makes sense. If at some point we were to support different value sets (e.g. `center` for cells, but not for headers), we can separate the types at that time.

packages/nimble-components/src/table/components/header/template.ts
```diff
@@ -44,5 +47,9 @@ export const template = html<TableHeader>`
                 ></${iconTwoSquaresInBracketsTag}>
             `)}
         `)}
+
+        ${when(x => x.alignment === TableColumnAlignment.right, html`
```
atmgrifter00, line 51: Am I correct in saying that no matrix test has been updated after your change? If not, should there be a matrix test that exercises this use-case?

m-akinc: The number-text column matrix tests [did have a change](https://www.chromatic.com/pullrequest?appId=60e89457a987cf003efc0a5b&number=2245&activeElementId=comparison-test-key-64d2b2b73a35b8c89d2fe96d-1200px&view=changes)

atmgrifter00 (approved):

packages/nimble-components/src/table-column/base/models/column-internals.ts
```diff
@@ -153,6 +157,12 @@ export class ColumnInternals<
     @observable
     public hideHeaderIndicators = false;
 
+    /**
+     * How to align the header content.
+     */
+    @observable
+    public headerAlignment?: TableColumnAlignment;
```
rajsite, line 164: instead of making this undefined can this just be `TableColumnAlignment.left`? That seems to be the fallback being used below when undefined

packages/nimble-components/src/table/template.ts
```diff
@@ -109,6 +110,7 @@ export const template = html<Table>`
                                             ?indicators-hidden="${x => x.columnInternals.hideHeaderIndicators}"
                                             @keydown="${(x, c) => c.parent.onHeaderKeyDown(x, c.event as KeyboardEvent)}"
                                             @click="${(x, c) => c.parent.toggleColumnSort(x, (c.event as MouseEvent).shiftKey)}"
+                                            :alignment="${x => (typeof x.columnInternals.headerAlignment === 'string' ? x.columnInternals.headerAlignment : TableColumnAlignment.left)}"
```
rajsite, line 113: shouldn't need the typeof check after it's type is hardened on columninternals

change/@ni-nimble-components-48e9cade-64e1-47a5-a616-c2e2a930a2b1.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: > The referenced bug doesn't really answer the question of where sort/grouping indicators (and future menu dropdown) should go when right-aligning the header. For now, I have assumed we should shift them to the left side of the column label.

Looks like PO and Design have thought about... so guess should make sure the changes align with their expectations
image

Think it's strange to be in Owner's without those questions answered (I don't own the answers) so I'll mark the PR reject 🤷‍♀️

rajsite (changes requested):

change/@ni-nimble-components-48e9cade-64e1-47a5-a616-c2e2a930a2b1.json
fredvisser: Brandon and I discussed - the sorting and grouping indicators should be mirrored when the alignment shifts

image

change/@ni-nimble-components-48e9cade-64e1-47a5-a616-c2e2a930a2b1.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "Align column header based on cell content alignment",
```
jattasNI, line 3: Could you ensure we have a storybook test that shows both the sorting and grouping header icons in both left and right aligned columns? I'd be fine if that's a new test or a change to the existing numeric column matrix test.

jattasNI (commented):
Waiting for the test and for the icon order change Fred just requested.

packages/nimble-components/src/table/components/header/styles.ts
m-akinc: Turns out we _do_ want to reverse the sorting/grouping indicators, so now I'm using `flex-direction` instead.

packages/nimble-components/src/table/components/header/styles.ts
fredvisser, line 1: This didn't change in this PR, but it looks like the latest visual design uses a 4px gap between the icons. Is this something we could align with now? FYI @NIbokeefe 

image

image

packages/nimble-components/src/table-column/number-text/index.ts
```diff
@@ -180,6 +179,7 @@ export class TableColumnNumberText extends mixinTextBase(
                 alignment: this.determineCellContentAlignment(),
                 placeholder: this.placeholder
             };
+            this.columnInternals.headerAlignment = columnConfig.alignment;
             this.columnInternals.columnConfig = columnConfig;
```
rajsite, line 183: Curious why we should have both a number column specific `columnConfig.alignment` property and a shared across all columns `columnInternals.headerAlignment` property. Seems like the number-text column should just use the columnInternals value and the columnInternals value should be more generic (`columnInternals.alignment`).

Thoughts @m-akinc @mollykreis on if there is a reason to keep them separate?

m-akinc: The cell view doesn't have access to the full `ColumnInternals` object -- just the `TableColumnNumberTextColumnConfig`. I think that's the main reason. The table's template passes the alignment value from `ColumnInternals` to the header, and it can't access a property that is only in certain column types' config object.

jattasNI (approved):

packages/nimble-components/src/table-column/number-text/index.ts
rajsite: This enterprising individual thought otherwise:

https://github.com/ni/nimble/blob/f47da47d1ee6104e9e4c913c8e1fbca321825e9d/packages/nimble-components/src/table-column/base/cell-view/index.ts#L59

But it does make more sense for the cellView instances to not have direct coupling to column / columnInternals. I'll create a tech debt to break that coupling that was introduced with delegated events. So the approach in this PR seems good.

rajsite (approved):

packages/nimble-components/src/table-column/number-text/index.ts
m-akinc: 😅

rajsite: Created tech debt https://github.com/ni/nimble/issues/2258

packages/nimble-components/src/table/components/header/styles.ts
m-akinc: Looks like there's also a difference between the left/right header padding. I've updated that as well.

fredvisser (approved):

PR #2254, opened 2024-07-10
Refinements to table selection when using SHIFT

### LINKED ISSUE #2239: Unexpected multiselect behavior in table
# 🐛 Bug Report

I ran into some use patterns that resulted in behavior that seems wrong.

## 💻 Repro or Code Sample

In this clip, we can see that deselecting a row causes it to be "registered" as the starting point for a shift+click/space multi-select. I would have expected that shift+clicking Sam while nothing else is selected would only select Sam.
deselect_becomes_anchor

Here is another case where deselecting a row results in weird behavior. We deselect a row inside a selected range, then shift+click a row beyond the range. I'm not sure what the right behavior is, but I would probably have expected:
1. Only Sam is added to the selection (i.e. same as without the shift key)
2. Everything between Sam and Michael (in this clip there aren't any other rows) inclusive is added to the selection

The fact that Joe, which we just deselected, gets re-selected seems wrong.
deselect_becomes_anchor2

I looked for some documentation to use as the basis for what is and isn't correct behavior, but I couldn't find any. The [ARIA treegrid guidance](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/) doesn't even suggest using shift+space, but instead talks about shift+arrow.

mollykreis:
# Pull Request

## 🤨 Rationale

This PR addresses two shift-selection concerns:
1. Range selection using shift should not start on a row that is not selected
2. Shift+click to a row should deselect any rows not in the selected range, but Shift+Ctrl+click should allow a range to be added to the current select. (Note, this behavior is specific to clicking on the rows themselves in multi-select mode. Clicking on a selection checkbox will always behave as though Ctrl is held.)

Resolves #2239

## 👩‍💻 Implementation

The desired multiselect behavior is that a row will only be considered the start of a shift-select range if that row was being selected (either via selection checkbox or via clicking the row), but it will not be considered the start of a shift-select range if that row was deselected.

To achieve this behavior, I've updated the `MultiSelectionManager` to assign the current row ID to `shiftSelectStartRowId` only if the row is being selected. If the row is being deselected, `shiftSelectStartRowId` is reset to `undefined`.

## 🧪 Testing

- Manually tested in storybook, including keyboard selection
- Added new unit tests

## ✅ Checklist


mollykreis:

packages/nimble-components/src/table/tests/table-selection.spec.ts
```diff
@@ -1400,6 +1400,72 @@ describe('Table row selection', () => {
                             jasmine.arrayWithExactContents(expectedSelection)
                         );
                     });
+
+                    it('SHIFT + click only selects the clicked row when the previous click deselected the row clicking the selection checkbox', async () => {
```
m-akinc, line 1404: Had to read this several times to parse it. I think this is clearer.
```suggestion
                    it('SHIFT + click selects only the clicked row if the previous click deselected a row (by clicking the checkbox)', async () => {
```

packages/nimble-components/src/table/tests/table-selection.spec.ts
```diff
@@ -1400,6 +1400,72 @@ describe('Table row selection', () => {
...
+                        );
+                        const emittedIds = getEmittedRecordIdsFromSpy(
+                            shiftSelectListener.spy
+                        );
+                        expect(emittedIds).toEqual(
+                            jasmine.arrayWithExactContents(expectedSelection)
+                        );
+                    });
+
+                    it('SHIFT + click only selects the clicked row when the previous click deselected the row through CTRL + click', async () => {
```
m-akinc, line 1437:
```suggestion
                    it('SHIFT + click selects only the clicked row if the previous click deselected the row (by CTRL + click)', async () => {
```

m-akinc (approved):

change/@ni-nimble-components-e878875a-13bb-474b-b54e-76adb87bc3fa.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: The ctrl behavior is a little surprising to me. I see the following:
1. Click on a row
2. ctrl+ click on a different row to add to my selection instead of change
3. ctrl+shift+click on a different row. Expected: adds that range. Actual: Just adds the single row

https://60e89457a987cf003efc0a5b-fohguihpuy.chromatic.com/?path=/story/components-table--table&args=data:LargeDataSet;selectionMode:multiple

rangectrlclick

mollykreis: I believe that is existing behavior. I was able to reproduce it on our published storybook. I think it'll be as simple as switching the order of two if-statements in the code to fix (i.e. look for the `shift` case before looking for the `ctrl` case). If so, I can address that in this PR as well.

change/@ni-nimble-components-e878875a-13bb-474b-b54e-76adb87bc3fa.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: Will make a separate thread but chatted offline with Molly about having the mental model of:

1. Pressing ctrl key is required to add to your selection (so having a selection and not holding ctrl w/ shift clears the previous selection, this aligns with Windows explorer)
2. Using the checkboxes is a mental model of always using the ctrl key (they are always additive)

change/@ni-nimble-components-e878875a-13bb-474b-b54e-76adb87bc3fa.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Don't allow a range selection in the table to begin with a row that was deselected",
```
mollykreis, line 3:
```suggestion
  "comment": "Refinements to table selection when using SHIFT",
```

change/@ni-nimble-components-e878875a-13bb-474b-b54e-76adb87bc3fa.json
mollykreis: I've updated the code to have the behavior you described.

change/@ni-nimble-components-e878875a-13bb-474b-b54e-76adb87bc3fa.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Refinements to table selection when using SHIFT",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
jattasNI, line 7: Could we add a note to the keyboard navigation section of the table storybook to mention these features? I don't think we need to be too detailed in describing the exact behavior but we could at least mention that mutiselect is possible via clicking with Shift and Ctrl.

If we do this we might consider renaming that section to "Keyboard interaction" or just "Interactions"

mollykreis: I added a few sentences under "Accessibility" >> "Selecting multiple rows" that's a sibling of the existing "Accessibility" >> "Keyboard navigation" section.

packages/storybook/src/nimble/table/table.mdx
```diff
@@ -164,4 +164,12 @@ the current row have been traversed, `Tab`/`Shift-Tab` will move focus outside
 the table. When the table is focused again, navigation will resume from the
 previously focused element.
 
+### Selecting multiple rows
+
+When the table has a `selectionMode` of `multiple`, multiple rows can be
+selected using the selection checkboxes. Additionally, multiple rows can be
+selected by pressing the `CTRL` key while clicking each row, and a range of rows
+can be selected by pressing the `SHIFT` key while clicking on one row and then
+another.
```
rajsite, line 173: Can we describe the row interactions by breaking it down into four interactions:
- clicking a row: Selects just that row
- clicking a row followed by shift + click another row: Selects just that range of rows
- ctrl + click on a row: Adds additional single rows to selection
- ctrl + click on a row followed by ctrl + shift + click on a different row: Adds a range of rows to the selection

rajsite (approved):

rajsite:
Think we should have @jattasNI or @fredvisser verify the interactions on Safari mac + maybe also Chrome/ Firefox mac

packages/storybook/src/nimble/table/table.mdx
```diff
@@ -164,4 +164,18 @@ the current row have been traversed, `Tab`/`Shift-Tab` will move focus outside
...
+selected using the selection checkboxes. Additionally, clicking on rows has the
+following behavior:
+
+-   Clicking on a row selects just that row
+-   `CTRL` clicking on a row toggles the selection of that row while maintaining
+    the selection of other rows
+-   Selecting a row followed by `SHIFT` clicking on another row selects the
+    range of rows between the two rows
+-   Selecting a row followed by `SHIFT-CTRL` clicking on a different row adds
+    the range of rows to the existing selection
```
jattasNI, line 179: On macOS it's the Command key (`⌘`) rather than `CTRL`. Ctrl-click is treated as a right click.

jattasNI (approved):
Works well in the macOS browsers I tested (Safari and Firefox) other than the key difference I call out below.

PR #2255, opened 2024-07-10
Add combobox page object and clean up combobox tests

m-akinc:
# Pull Request

## 🤨 Rationale

A page object is useful for writing clearer, less error-prone unit tests. I was going to write new combobox tests and was motivated to follow the example of the select unit tests.

## 👩‍💻 Implementation

Based some of the page object on the select page object, but removed all functions that are not currently needed. We can easily add additional functions as needed.

Tests were adapted to use the page object. One duplicate test (copy-paste error) was removed. `parameterizeSpec` was employed in two places.

## 🧪 Testing

Unit tests pass.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
+++ b/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
@@ -22,7 +22,7 @@ export class ComboboxPageObject {
     /**
      * Gets the selectable options in the drop-down. Does not include options that are disabled or filtered out.
      */
-    public getOptions(): string[] {
+    public getFilteredOptions(): string[] {
         return this.comboboxElement.filteredOptions.map(x => x.text);
     }
 
--- a/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
+++ b/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
@@ -33,6 +33,9 @@ export class ComboboxPageObject {
         this.comboboxElement.control.dispatchEvent(inputEvent);
     }
 
+    /**
+     * Sets the input text and commits the value by pressing Enter. Will emit a 'change' event.
+     */
     public commitValue(text: string): void {
         this.setInputText(text);
         this.pressEnterKey();
--- a/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
+++ b/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
@@ -28,7 +28,7 @@ export class ComboboxPageObject {
         this.comboboxElement.control.value = text;
         const inputEvent = new InputEvent('input', {
             data: text,
-            inputType: asDelete ? 'deleteContent' : 'insertText'
+            inputType: asDelete ? 'deleteContentForward' : 'insertText'
         });
         this.comboboxElement.control.dispatchEvent(inputEvent);
     }
--- a/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
+++ b/packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
@@ -1,5 +1,6 @@
 import { keyEnter, keyArrowDown } from '@microsoft/fast-web-utilities';
 import type { Combobox } from '..';
+import { listOptionTag } from '../../list-option';
 import { waitForUpdatesAsync } from '../../testing/async-helpers';
 import {
     createEventListener,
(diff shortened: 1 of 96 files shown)


packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -1,259 +1,14 @@
...
-
-        await connect();
-
-        expect(element.getAttribute('tabindex')).toBe('0');
-        element.disabled = true;
-        await waitForUpdatesAsync();
-        expect(element.getAttribute('tabindex')).toBeNull();
-
-        await disconnect();
-    });
```
m-akinc, line 128: Removed `'should forward based on disabled state'` -- looks like a copy-paste leftover that is just a duplicate of the previous test.

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -264,377 +19,422 @@ describe('Combobox', () => {
...
+        afterEach(async () => {
+            await disconnect();
+        });
 
-    const filterOptionTestData: { autocomplete: ComboboxAutocomplete }[] = [
-        { autocomplete: ComboboxAutocomplete.inline },
-        { autocomplete: ComboboxAutocomplete.both }
-    ];
-    filterOptionTestData.forEach(testData => {
-        it('disabled options will not be selected by keyboard input', async () => {
```
m-akinc, line 627: Another `parameterizeSpec` replacement

m-akinc:
@atmgrifter00 would you buddy this, please? The diff is miserable, unfortunately. :(

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
```diff
@@ -0,0 +1,125 @@
...
+    public setInputText(text: string, asDelete = false): void {
+        this.comboboxElement.control.value = text;
+        const inputEvent = new InputEvent('input', {
+            data: text,
+            inputType: asDelete ? 'deleteContentForward' : 'insertText'
+        });
+        this.comboboxElement.control.dispatchEvent(inputEvent);
+    }
+
+    public commitValue(text: string): void {
```
atmgrifter00, line 36: Perhaps add some docs that state that calling this method will result in a `change` event being emitted?

atmgrifter00 (approved):

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
```diff
@@ -0,0 +1,128 @@
...
+    private readonly regionLoadedListener = createEventListener(
+        this.comboboxElement,
+        'loaded'
+    );
+
+    public constructor(protected readonly comboboxElement: Combobox) {}
+
+    public async waitForAnchoredRegionLoaded(): Promise<void> {
+        await this.regionLoadedListener.promise;
+    }
```
rajsite, line 23: thinking of the pageobject as abstracting details, this one feels strange to be in the public api. Could it instead be private and used in apis like clickAndWaitForOpen?

rajsite (approved):
Nice cleanup! Gave a quick look but relying on @jattasNI to review the page object api closer.

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    public async mouseoutInput(): Promise<void> {
+        this.comboboxElement.control.dispatchEvent(new MouseEvent('mouseout'));
+        await waitForUpdatesAsync();
+    }
+
+    public async prependOption(value: string, label: string): Promise<void> {
+        this.comboboxElement.insertAdjacentHTML(
+            'afterbegin',
+            `<nimble-list-option value="${value}">${label}</nimble-list-option>`
```
rajsite, line 107: make sure to use the listOptionTagName instead of hard-coding. That ensure the element is imported and registered.

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -1,640 +1,470 @@
...
-
-        expect(element.getAttribute('tabindex')).toBe('0');
-        element.disabled = true;
-        await waitForUpdatesAsync();
-        expect(element.getAttribute('tabindex')).toBeNull();
+    describe('with common setup', () => {
+        async function setup(): Promise<Fixture<Combobox>> {
+            // prettier-ignore
+            const viewTemplate = html`
+                <nimble-combobox>
```
rajsite, line 26: While we are here can you update this fixture and the others below to use the tagName imports instead of hard coding the names

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
m-akinc: I'm already waiting for the region loaded promise in the two click-and-wait-for-open functions, but if I also wait for it in the commit function, I think that works.

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+    createEventListener,
+    waitAnimationFrame
+} from '../../utilities/tests/component';
+
+/**
+ * Page object for the `nimble-combobox` component to provide consistent ways
+ * of querying and interacting with the component during tests.
+ */
+export class ComboboxPageObject {
```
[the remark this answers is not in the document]
m-akinc: > I don't see a getOptions() method here so it'd be good to add and ensure it has a test exercising it.

I didn't add that since `filteredOptions` is part of the combobox's public API, unless we forgot to mark it internal. The select has the same property, but it is internal.

> 2. get an array of all option values as string

Not sure if that's necessary if `filteredOptions` is part of the public API (e.g. `combobox.filteredOptions.map(x => x.value)`), but I guess there's no harm in adding it for convenience.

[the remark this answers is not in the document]
rajsite: I wouldn't expect filteredOptions to be part of the public API. We should mark that internal.

packages/nimble-components/src/combobox/testing/combobox.pageobject.ts
```diff
@@ -0,0 +1,183 @@
...
+        this.comboboxElement,
+        'loaded'
+    );
+
+    public constructor(protected readonly comboboxElement: Combobox) {}
+
+    /**
+     * Gets the selectable options in the drop-down. Does not include options that are disabled or filtered out.
+     */
+    public getOptions(): string[] {
```
jattasNI, line 25: nit: I'd vote for a name that disambiguates this from a (not currently existent, but possibly desired in future) method that returns all the child options, including the ones that are filtered out). I'd be fine with `getFilteredOptions()`, `getAvailableOptions()`, `getSelectableOptions()`, etc.

jattasNI (approved):

PR #2257, opened 2024-07-10
Remove table cell view focusedRecycleCallback

### LINKED ISSUE #2202: Consider removing table focusedRecycleCallback
# 🧹 Tech Debt

While wrapping up the table keyboard nav feature (and testing it with MenuButton column prototypes), we realized that the focusedRecycleCallback may now be redundant, since the keyboard nav code will focus the cell on table scroll (if the action menu button or focusable cell content was focused). That code that focuses the cell, will also result in the previously focused element being blurred.
So, if there's nothing additional we need to do for any column types other than blur, we may be able to remove `TableCellView` `focusedRecycleCallback()` and `Virtualizer` `notifyFocusedCellRecycling()`.

We may want to wait until the menu button ~~/ button~~ column types are done, once we're sure that the current code will handle all the various cases (in conjunction with table scroll / other virtualizer updates):
- Action menu focused, but not open
- Action menu open
- MenuButton focused, but not open
- MenuButton open

[PR discussion thread](https://github.com/ni/nimble/pull/2172#discussion_r1638579629)

msmithNI:
# Pull Request

## 🤨 Rationale

Resolves #2202 

## 👩‍💻 Implementation

- Remove `TableCellView` `focusedRecycleCallback()` and `TableRow` `closeOpenActionMenus()` (both were called on virtualizer updates).
   - `KeyboardNavigationManager` focuses the active cell on scroll in `handleChange()`, which has the same effect.
- Update column HLD template to remove section about focus recycling

## 🧪 Testing

- Manually tested table action menus and anchor + menu button columns with scrolling, ensured behavior is the same as before
- Tests were already present to ensure that interactive cell content (e.g. anchor) is blurred on data updates, and that menu buttons in cells are closed on scroll.
- Added new tests to ensure cell action menus are still closed after a scroll, and that interactive cell content (e.g. anchor) is blurred on scroll.

## ✅ Checklist


packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -215,12 +216,6 @@ implements Subscriber {
         }
     }
 
-    public handleFocusedCellRecycling(hadRowOrCellFocus: boolean): void {
-        if (hadRowOrCellFocus && !this.focusWithinTable) {
-            this.focusCurrentRow(false);
```
msmithNI, line 220: Note: The only case where we hit this code path was when an anchor was focused, and then a table scroll happened. Then, at the start of `table.handleFocusedCellRecycling();` focus was within the table, but it no longer was when this method was called (due to the anchor being blurred by its `focusedRecycleCallback`).

Now that the `focusedRecycleCallback` is gone, the code `handleChange()` which refocuses the cell does this. So this method goes away.

packages/nimble-components/src/table/index.ts
```diff
@@ -684,37 +683,6 @@ export class Table<
...
-        }
-        if (tableFocusedElement instanceof TableCellView) {
-            tableFocusedElement.focusedRecycleCallback();
-        }
-        if (this.openActionMenuRecordId !== undefined) {
-            const activeRow = this.rowElements.find(
-                row => row instanceof TableRow
-                    && row.recordId === this.openActionMenuRecordId
-            ) as TableRow | undefined;
-            activeRow?.closeOpenActionMenus();
```
msmithNI, line 710: This code can go away for the same reason `focusedRecycleCallback` does: focusing the cell after a scroll accomplishes the same thing.

msmithNI:
@mollykreis can you buddy this PR?

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
```diff
@@ -652,6 +617,54 @@ describe('Table keyboard navigation', () => {
                 expect(pageObject.getCell(0, 0).menuOpen).toBe(true);
             });
 
+            it('when a cell action menu is opened via keyboard, scrolling the table will close the action menu and focus the cell', async () => {
```
mollykreis, line 620: Can you double check that we have a test somewhere that ensures the action menu is closed when data is updated via `setData`?

mollykreis (approved):

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
msmithNI: Added that test.

atmgrifter00 (approved):

rajsite (approved):

change/@ni-nimble-components-b08c8f1b-bce8-4d5e-b9f0-a3fadb6c92c9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Breaking change: Table: remove TableCellView.focusedRecycleCallback() and TableRow.closeOpenActionMenus(). The table now focuses the active cell when the table is scrolled, which has the same effect (closing open menus and blurring active cell content).",
```
jattasNI, line 3: I would vote to call this a patch change. I don't think I'd consider `TableCellView` or `TableRow` to be public classes or either of these methods to be part of the public API. I guess technically we should have marked them `@internal` but if we don't have any known direct clients of them then I'd feel safe pretending they have been internal the whole time.

```suggestion
  "type": "patch",
  "comment": "Table: remove TableCellView.focusedRecycleCallback() and TableRow.closeOpenActionMenus(). The table now focuses the active cell when the table is scrolled, which has the same effect (closing open menus and blurring active cell content).",
```

jattasNI (approved):

change/@ni-nimble-components-b08c8f1b-bce8-4d5e-b9f0-a3fadb6c92c9.json
rajsite: Providing a `TableCellView` instance is part of the `TableColumn` / `ColumnInternals` API so it is a public API if we say making custom columns is part of the public API and not internal APIs.

rajsite: Marking `TableRow` internal makes sense but I think we want making custom columns to be part of the public API so I wouldn't mark `TableCellView` internal. I don't feel strongly about the change being patch or major, I'd certainly find it reasonable to say our public api for implementing custom columns is "incubating" so expect breaking changes.

rajsite: Went ahead and committed the suggested switch to patch so the PR can merge

PR #2260, opened 2024-07-11
Configure karma/jasmine to fail tests with no expectations

jattasNI:
# Pull Request

## 🤨 Rationale

A test with no expectations typically indicates a problem with the test. We'd like to detect this situation automatically.

See [similar PR for SystemLinkShared](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/750756)

## 👩‍💻 Implementation

Set [`client.jasmine.failSpecWithNoExpectations: true`](https://jasmine.github.io/api/5.1/Configuration) in each package's `karma.conf.js`

## 🧪 Testing

Locally verified that a test with no expectations would pass before but fails now.

## ✅ Checklist


jattasNI:
When verifying that this was working I realized we already have it, it's just configured differently than in SystemLinkShared. Abandoning

PR #2261, opened 2024-07-11
Remove reserved space for the table's "Collapse all" button when the table cannot have expandable rows

### LINKED ISSUE #2241: Table should allow removing reserved space for collapse all button
# 🙋 Feature Request

## 😯 Problem to Solve

In all tables, including simple tables, space is reserved for the "Collapse all" button in the header. In most cases, this isn't a problem, but in simple tables within constrained spaces, it would be ideal for that space not to be reserved.

image

## 💁 Proposed Solution

Two potential solutions:
1. Have a mode on the table to not reserve space for the "Collapse all" button. There probably needs to be validation that this mode isn't used with features that require the "Collapse all" button.
2. Don't reserve space for the "Collapse all" button when it isn't needed. This is likely when all columns have grouping disabled and when parentIdFieldName isn't configured on the table.

mollykreis:
# Pull Request

## 🤨 Rationale

When all table columns have grouping disabled and the table isn't configured for hierarchical data, then there isn't a way for a row to be collapsable. Therefore, the "Collapse all" button will never be shown, and the space for it does not need to be reserved.

When the space is not reserved for the "Collapse all" button, there now is a 16px margin between the edge of the table (or selection checkbox when `multiple` selection is enabled), rather than an 8px margin. This spacing is specified in the visual designs.

Resolves #2241 

## 👩‍💻 Implementation

The table checks to see if any column supports grouping or if `parentIdFieldName` is configured. If so, the space is still reserved for the "Collapse all" button. If not, space is not reserved in the header for the "Collapse all" button, and the table tells the row not to reserve space for an expand/collapse button.

## 🧪 Testing

- Manually tested in storybook, including resizing columns
- Added new matrix tests
- Added new unit tests

## ✅ Checklist


mollykreis:
@atmgrifter00, can I get your thoughts on these changes? I still need to add/update tests, including chromatic tests, but I wanted to get an opinion on the implementation direction before I invested time in testing.

packages/nimble-components/src/table/template.ts
```diff
@@ -67,7 +67,7 @@ export const template = html<Table>`
                                         ${ref('selectionCheckbox')}
                                         ${'' /* tabindex managed dynamically by KeyboardNavigationManager */}
                                         tabindex="-1"
-                                        class="${x => `selection-checkbox ${x.selectionMode ? x.selectionMode : ''}`}"
+                                        class="selection-checkbox"
```
mollykreis, line 70: Note: This change isn't part of the "Collapse all" modification. I just noticed that we were adding extra dynamic styling that wasn't necessary, so I removed it.

packages/nimble-components/src/table/template.ts
```diff
@@ -79,7 +79,7 @@ export const template = html<Table>`
                                 ${ref('collapseAllButton')}
                                 ${'' /* tabindex managed dynamically by KeyboardNavigationManager */}
                                 tabindex="-1"
-                                class="collapse-all-button ${x => `${x.showCollapseAll ? 'visible' : ''}`}"
+                                class="collapse-all-button ${x => `${x.showCollapseAll ? 'visible' : ''}`} ${x => `${x.canHaveCollapsibleRows ? '' : 'hidden'}`}"
```
atmgrifter00, line 82: This, I think, is the only part of the implementation that might merit re-thinking. At a glance it's odd to have one observable property, when true, set the CSS `visible` state, and a separate observable property, when false, set the CSS `hidden` state. I think it would be cleaner if we made a single observable property be enumerated. At the moment the right semantics are eluding me, as the three terms that leapt to mind were `show`, `hide` and `disable`, but the `hide` would map to "", while `disable` would map to `hidden`, which seems strange.

atmgrifter00, line 82: Would it be better to have a new volatile property that returned the right CSS setting:
```
@volatile
public get collapseButtonVisibility(): string {
    if (!this.canHaveCollapsibleRows) {
        return 'hidden';
    } else if (this.showCollapseAll) {
        return 'visible';
    }
    return '';
}
```
The thing that rubs me the wrong way is having two independent observables competing to set opposite CSS state.

atmgrifter00 (approved):

change/@ni-nimble-components-a8b6518e-7eb6-4450-8d81-4787ae2d1874.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Should we document the behavior in storybook? Maybe in the table styling section?

https://60e89457a987cf003efc0a5b-gzrnmbkymk.chromatic.com/?path=/docs/components-table--docs#styling

Might refer to it as "full-bleed" table styling to align with the text-field frameless full-bleed mode

packages/nimble-components/src/table/index.ts
```diff
@@ -184,6 +185,26 @@ export class Table<
...
+    @observable
+    public canHaveCollapsibleRows = false;
+
+    /**
+     * @internal
+     */
+    @volatile
+    public get collapseButtonVisibility(): string {
+        if (!this.canHaveCollapsibleRows) {
+            return 'hidden';
```
rajsite, line 200: Since we are leveraging the subtilties of `visibility: visible`, `visibility: hidden`, and `display: none` could we choose a different class name? Maybe `hidden-reduced-size`?

rajsite (approved):

change/@ni-nimble-components-a8b6518e-7eb6-4450-8d81-4787ae2d1874.json
jattasNI: I think it'd be good to do so. The main benefit for me is that clients might want this appearance but not realize they have to explicitly disable grouping (rather than just not configuring grouping).

jattasNI (approved):

PR #2264, opened 2024-07-12
Skip lighthouse dom-size audit

rajsite:
# Pull Request

## ✅ Checklist


rajsite:
Doesn't seem to impact the perf score, so while less noisy, doesn't help with false positives on perf issues

PR #2271, opened 2024-07-15
Consolidate table slot tracking

### LINKED ISSUE #2248: Consolidate slot tracking in table
# 🧹 Tech Debt
Currently, the table tracks what row/column needs to have the action menu slot, and the table separately tracks where slotted content for a column needs to be put. We should look into consolidating that tracking so that we only have one data structure containing slot placement information.

See discussion here: https://github.com/ni/nimble/pull/2204#discussion_r1669337893

mollykreis:
# Pull Request

## 🤨 Rationale

Previously there were two different ways that client-requested slots were tracked and dynamically created within rows of the table. This PR consolidates that code so that action menu slots and cell-requested slots all end up in `columnRequestedSlots`, and ultimately as part of `tableData`.

Resolves #2248 

## 👩‍💻 Implementation

When determining that action menu slots need to be created, the table now creates those slots in `columnRequestedSlots` rather than simply ensuring that slots with the same name don't already exist in `columnRequestedSlots`. The code that handles action menu slot names changing was also updated to remove the old slots from `columnRequestedSlots` and to add the new slots to `columnRequestedSlots` when the slot names change while the menu is open.

## 🧪 Testing

- Sanity checked that action menus still work as expected in storybook
- Verified existing unit tests still pass

## ✅ Checklist


mollykreis:

atmgrifter00 (approved):

packages/nimble-components/src/table/index.ts
```diff
@@ -762,9 +758,14 @@ export class Table<
         this.tableUpdateTracker.trackColumnInstancesChanged();
     }
 
-    private removeActionMenuSlotsFromColumnRequestedSlots(): void {
+    private updateRequestedSlotsForOpeningActionMenu(
+        openActionMenuRecordId: string
+    ): void {
         for (const actionMenuSlot of this.actionMenuSlots) {
-            this.columnRequestedSlots.delete(actionMenuSlot);
+            this.columnRequestedSlots.set(actionMenuSlot, {
```
rajsite, line 765: What about renaming `columnRequestedSlots` to just `requestedSlots`?

packages/nimble-components/src/table/template.ts
```diff
@@ -187,14 +187,6 @@ export const template = html<Table>`
...
                                     >
-                                    ${when((x, c: ExecutionContext<Table>) => c.parent.openActionMenuRecordId === c.parent.tableData[x.index]?.id, html<VirtualItem<HTMLElement>, Table>`
-                                        ${repeat((_, c: ExecutionContext<Table>) => c.parent.actionMenuSlots, html<string, Table>`
-                                            <slot
-                                                name="${x => x}"
-                                                slot="${x => `row-action-menu-${x}`}">
-                                            </slot>
-                                        `)}
-                                    `)}
                                     ${repeat((x, c: ExecutionContext<Table>) => (c.parent.tableData[x.index]?.slots || []), html<SlotMetadata>`
```
rajsite, line 190: inline with the other comment what about maintaining a similar name and calling this `tableData.requestedSlots`?

packages/nimble-components/src/table/index.ts
```diff
@@ -762,9 +758,14 @@ export class Table<
...
+                recordId: openActionMenuRecordId,
+                uniqueSlot: `row-action-menu-${actionMenuSlot}`
```
rajsite, line 767: This is kinda now a reserved columnId of `row-action-menu` right?
Should we go down the same uniqueness codepath?

```ts
    const rowActionMenuUniqueId = uniqueId('row-action-menu')
    uniqueSlot: uniquifySlotNameForColumnId(rowActionMenuUniqueId, actionMenuSlot)
```

rajsite (approved):

packages/nimble-components/src/table/index.ts
mollykreis: I could update it to go down the same uniqueness code path if we want for consistency, but I don't think it makes a difference in practice. The table's `actionMenuSlots` array has already removed duplicates between columns, so `row-action-menu-${actionMenuSlot}` will never result in duplicate slots created for action menus, and it will also never result in duplicates with column slots that are uniquified to something like `table-column-slot0-${requestedSlot}`.

Thoughts? Do you think I should make the code paths more similar for consistency?

rajsite: I'm sold to keep it as-is!

rajsite:
@mollykreis in the details refactoring that's mostly renaming, don't think we need to wait on @jattasNI

jattasNI (approved):

Issue #2107, opened 2024-05-14
Re-export utilities from nimble-components and nimble-tokens in nimble-angular


jattasNI:
# 🙋 Feature Request

## 😯 Problem to Solve

SLE clients should usually only need to use `nimble-angular`, not `nimble-tokens` or `nimble-components`. But there are currently some utilities they want that are not available. From [this search](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=%27%40ni/nimble-components/dist/esm%20path%3Aweb/workspaces&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/AssetManager/src/app/models/location-history-table-model.ts):

1. diacriticInsensitiveStringNormalizer
2. some table types (I suspect these are already exported and apps are just importing them from the wrong place)
3. some design tokens and theme provider classes (I suspect these are an exception to the rule since apps are doing non-standard things, but we should discuss)

## 💁 Proposed Solution

1. Discuss each case with the team
2. Re-export what should be re-exported
3. Clean up client apps to not depend on lower level packages

PR #2290, opened 2024-07-18
Include page object utility in nimble-components build output

jattasNI:
# Pull Request

## 🤨 Rationale

We added the combobox page object in #2255 and #2286. When I tried to use it in SLE I got a build error saying "Error: Module not found: Error: Can't resolve '../../utilities/tests/component' in '..../SystemLinkShared/node_modules/@ni/nimble-components/dist/esm/combobox/testing'"

That import is a dependency of the page object but because it lives in a folder called "tests", [the .npmignore file](https://github.com/ni/nimble/blob/main/packages/nimble-components/.npmignore) doesn't include it in the built package contents.

This also impacts the menu button page object.

## 👩‍💻 Implementation

1. Move `src/utilities/tests/component.ts` to `src/utilities/testing/component.ts` and update callers.
2. Add README.md to several utilities directories explaining their purpose. The general rule is that all of it should be considered internal. Some of it is used externally but that's tracked by tech debt like #2107.

Longer term we could probably catch issues like this earlier by explicitly exporting certain paths from the nimble-components package.json, which would mean that nimble-angular's build would have failed. This is tracked by #172.

## 🧪 Testing

Verified that the output of `npm run pack -w @ni/nimble-components` now includes the `testing/components` dependency. I will test this in SLE post-release and address any remaining issues in a follow-up.

## ✅ Checklist


packages/nimble-components/src/utilities/tests/README.md
```diff
@@ -1,3 +1,3 @@
 # Utilities Tests
 
-The `utilities/tests` folder contains test utilities to be used only by nimble-component library authors and are not intended to be exported externally.
+The `utilities/tests` folder contains test utilities to be used only by nimble-component tests and are not exported externally.
```
rajsite, line 3:
```suggestion
The `utilities/tests` folder contains test utilities to be used **only** by nimble-component tests as the directory is completely excluded from the published packages.
```

packages/nimble-components/src/utilities/testing/README.md
```diff
@@ -0,0 +1,3 @@
+# Utilities Testing
+
+The `utilities/testing` folder contains test utilities to be used by nimble-component page objects and tests. They are exported externally since public page objects need to use them but are not part of Nimble's public API.
```
rajsite, line 3:
```suggestion
The `utilities/testing` folder contains test utilities to be used by nimble-component page objects and tests. The utilities are used by public page objects but are not considered part of Nimble's public API.
```

packages/nimble-components/src/utilities/README.md
```diff
@@ -0,0 +1,5 @@
+# Utilities
+
+Utilities APIs are intended for Nimble use only and should be considered internal.
+
+Any existing usages by clients should be considered tech debt which should be fixed by moving the utility to a nimble-components directory that is considered public and re-exporting it from a framework package.
```
rajsite, line 5:
```suggestion
Utilities APIs are intended for Nimble use only and should be considered internal.
```

If we want to track the tech debt I think it should be a tech debt issue instead of in docs (frankly think someone adding to that tech debt has an equal if not lower chance of discovering it via markdown docs vs a tech debt issue).

PR #2291, opened 2024-07-18
Add dropdown "No results" placeholder for combobox and rich-text mention listbox

### LINKED ISSUE #1309: Placeholder for options list in combobox (and eventually select) when nothing matches filter
## 😯 Problem to Solve

When the user types a value in the combobox that doesn't match any options, it shows an empty dropdown list that looks odd.

image

In [this SLE PR](https://ni.visualstudio.com/DevCentral/_git/Skyline/pullrequest/499882) a client is implementing a workaround to show placeholder text instead.

(Note that there is a related issue #350 but they are different placeholders)

## 💁 Proposed Solution

We should bring the workaround from the linked PR into Nimble and add the ability to show a client-customizable "No items found" item instead of the empty list.

image

In that PR [Brandon suggested](https://ni.visualstudio.com/DefaultCollection/DevCentral/_git/Skyline/pullRequest/499882#1686852644) that the placeholder item should have a text color of Black91@60% to distinguish it from the valid items (90%) and the disabled items (30%). We'll need equivalent colors for themes besides Light.

m-akinc:
# Pull Request

## 🤨 Rationale

Part of #1309

The other part of that issue is updating clients to use this instead of their workarounds.

The issue specifies that the text should be configurable. This can be done by way of a `nimble-label-provider-core` element, but that has limitations, i.e. if you want to customize two different combobox/select/rich-text-listboxes on the same page, they have to be scoped to distinct theme provider elements. 

## 👩‍💻 Implementation

- added support for "No results found" placeholder to the combobox and rich-text mention listbox
- moved related styling to shared dropdown styling
   - reordered to follow DOM order
- fixed issue with combobox where options with `hidden` attribute would still be included in `filteredOptions` (preventing placeholder from showing)

## 🧪 Testing

- new unit tests for combobox and mention listbox
- new visual tests for combobox and mention listbox

## ✅ Checklist


packages/nimble-components/src/combobox/index.ts
```diff
@@ -340,13 +341,11 @@ export class Combobox
 
         const filter = this.filter.toLowerCase();
 
-        this.filteredOptions = this._options.filter(o => o.text.toLowerCase().startsWith(filter));
+        this.filteredOptions = this._options.filter(
```
m-akinc, line 344: Exclude hidden options.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -340,13 +341,11 @@ export class Combobox
...
+            o => o.text.toLowerCase().startsWith(filter) && !o.hidden
+        );
```
m-akinc, line 346: Exclude hidden options.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -340,13 +341,11 @@ export class Combobox
...
 
         if (this.isAutocompleteList) {
-            if (!this.filteredOptions.length && !filter) {
-                this.filteredOptions = this._options;
-            }
-
```
m-akinc, line 349: `filter` should not be `null`/`undefined`, so `!filter` should mean `filter === ''`. When the filter is empty, `startsWith('')` will let every option through, so the only way to get in this block is if there are no visible (i.e. not `hidden`) options. We don't need to reset it to `this._options`, and if all the options are `hidden`, resetting it is actually wrong.

m-akinc:
@atmgrifter00 will you buddy, please?

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -244,6 +244,33 @@ describe('Combobox', () => {
...
+
+        it('does not show "no items found" in dropdown when input is empty', async () => {
+            element.autocomplete = ComboboxAutocomplete.both;
+            await pageObject.clickAndWaitForOpen();
+            expect(pageObject.isNoResultsLabelVisible()).toBeFalse();
+        });
+
+        it('shows "no items found" in dropdown after hiding all options', async () => {
+            pageObject.hideAllOptions();
+            pageObject.setInputText('');
```
atmgrifter00, line 269: Just wondering, but is this line needed? Initially there is no value right?

atmgrifter00 (approved):

packages/nimble-components/src/combobox/tests/combobox.spec.ts
m-akinc: It was, because we need to trigger re-filtering, but I just changed `hideAllOptions` to call `filterOptions`.

packages/nimble-components/src/combobox/template.ts
```diff
@@ -10,7 +16,9 @@ import type { Combobox } from '.';
 import { anchoredRegionTag } from '../anchored-region';
 import { DropdownPosition } from '../patterns/dropdown/types';
 import { overflow } from '../utilities/directive/overflow';
+import { filterNoResultsLabel } from '../label-provider/core/label-tokens';
 
+/* eslint-disable @typescript-eslint/indent */
```
rajsite, line 21: Can you make sure all these `eslint-disable` have a bookend `eslint-enable`

change/@ni-nimble-components-d0cf5648-9e73-4eca-8266-c062a09c0dab.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: > if you want to customize two different combobox/select/rich-text-listboxes on the same page, they have to be scoped to distinct theme provider elements.

If in integration we are really finding that we need to customize this for every use case then we should probably add an attribute to make that configuration easier. Not something for this PR but a question we should clarify in a follow-up. I'd hope the generic "No items found" is good enough. But if not we should add an attribute instead of many theme-provider + label-providers all over.

packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -187,55 +227,33 @@ export const styles = css`
...
-    .selected-value {
-        flex: auto;
-        font-family: inherit;
-        text-align: start;
-        white-space: nowrap;
-        text-overflow: ellipsis;
-        overflow: hidden;
-        padding: 0px;
-        padding-left: ${mediumPadding};
+    .empty slot {
```
rajsite, line 240: pretty generic class name, including the class for the element makes it clearer


```suggestion
    .listbox.empty slot {
```

rajsite (approved):

change/@ni-nimble-components-d0cf5648-9e73-4eca-8266-c062a09c0dab.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: > reordered to follow DOM order

appreciate it! made it much easier to review 👏

change/@ni-nimble-components-d0cf5648-9e73-4eca-8266-c062a09c0dab.json
m-akinc: > Not something for this PR but a question we should clarify in a follow-up.

FYI @fredvisser

packages/nimble-components/src/combobox/template.ts
m-akinc: Got everything in `nimble-components`, I think.

change/@ni-nimble-components-d0cf5648-9e73-4eca-8266-c062a09c0dab.json
fredvisser: Good question - Milan and I discussed, and I'll make the usage/pattern guidance consistent with our current capabilities ("No items found"), and change the guidance and our implementation in the future if this ever becomes the most important thing we can do to improve usability.

m-akinc: So, to be clear, once this gets pushed to SLE, should I update the [existing, customized "No tags found" message](https://ni.visualstudio.com/DevCentral/_git/Skyline/pullrequest/499882) to use a `nimble-label-provider-core` instead, or should I remove the customization altogether?

rajsite: My understanding of the discussion would be to remove control-specific customizations of the "no items found" message and in the future if it is desired to have customized messages then have usage guidance to clarify if it's only for some cases or the general case and if the general case add an attribute to make config easier.

packages/nimble-components/src/rich-text/mention-listbox/tests/mention-listbox.spec.ts
```diff
@@ -44,62 +44,82 @@ describe('RichTextMentionListbox', () => {
...
-        listbox.show({
+    async function showAndWaitForOpen(filter = ''): Promise<void> {
+        const regionLoadedListener = createEventListener(element, 'loaded');
+        element.show({
             anchorNode: anchor,
-            filter: ''
+            filter
         });
         await regionLoadedListener.promise;
     }
```
jattasNI, line 65: Think it'd be worthwhile to include a test that the No Items Found item is not shown if the set of items updates to include an item that matches the filter? i.e. that we are correctly observing the child list options

jattasNI (approved):

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -244,6 +244,32 @@ describe('Combobox', () => {
             );
         });
```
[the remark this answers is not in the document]
m-akinc: It's unlikely we'd implement it in a way that this would be a distinct test case from the other cases where at least one option is visible, but it doesn't hurt to have it.

packages/nimble-components/src/combobox/template.ts
rajsite: I request it for refactoring purposes / modifications. Disable for a block of a file, not the rest of the file as stuff can be inserted later / moved around that may or may not intentionally be found.

> This conversation needn't block this PR so I left it Resolved.

@jattasNI @m-akinc this makes it very easy to miss since GitHub notifications are per PR not per mention (at least for me). I'd prefer the PR owner keep them open till merge and then re-open them so the open conversations are discoverable.

m-akinc: >  I'd prefer the PR owner keep them open till merge and then re-open them...

To confirm, you're not asking for a real change in how we've been operating--just that the author resolves threads when the resolving code gets pushed, then a subsequent commenter on a thread should unresolve it. Right?

rajsite: Right. Whoever merges the PR can temporarily resolve the unresolved, non-blocking comments to merge but should make sure they are re-opened (and ideally that they are temporarily resolved close to the merge, not significantly in advance).

PR #2294, opened 2024-07-19
Fix bug where user can't type disabled option into combobox

### LINKED ISSUE #2252: Combobox replaces text if you type a disabled value
# 🐛 Bug Report

As soon as you finish typing in the name/value of a disabled option, the combobox immediately replaces the text with the display name of a valid option.

## 💻 Repro or Code Sample
combobox_disabled_bug

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2252

## 👩‍💻 Implementation

The `options` getter should not revert to returning all the options (i.e. `_options`) when `filteredOptions` is empty. Doing so can lead to setting a `selectedIndex` that points to an option that is filtered out or disabled.

The expected behavior now is that a disabled option functions exactly like an enabled option, except that it cannot be clicked, cannot be focused in the dropdown, and appears dimmed. Like enabled options, t can be filtered to (if the input text matches it), and the user can set the control's value to it by typing it in fully.

## 🧪 Testing

Added new tests

## ✅ Checklist


change/@ni-nimble-components-5084859c-b0ef-4431-b684-b4733c6d7714.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Seems reasonable. @atmgrifter00 can you review as select / combobox expert and see if other edge cases, etc, in the logic to verify?

rajsite (approved):

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -307,21 +307,29 @@ describe('Combobox', () => {
...
                     element.autocomplete = name;
                     pageObject.setInputText('F');
                     await pageObject.clickAway(); // attempt to commit typed value
 
                     expect(element.value).not.toEqual('Four');
-                }
-            );
+                });
+
+                it('allows setting disabled option if typed in', () => {
```
atmgrifter00, line 327: This name seems a little misleading as we aren't allowing the _setting_ of the value to the disabled option, just that after completing typing of the same text as a disabled option, we don't immediately switch it to something else. Perhaps "will not switch option if disabled option typed in fully"?

packages/nimble-components/src/combobox/index.ts
```diff
@@ -152,7 +152,7 @@ export class Combobox
         let updatedValue = next;
 
         if (this.$fastController.isConnected && this.options) {
-            const selectedIndex = this.options.findIndex(
+            const selectedIndex = this.filteredOptions.findIndex(
```
atmgrifter00, line 155: A couple of possible issues:
- There is now an incongruity between what you're searching and the condition you use to enter the `if` as well as the arrays you are indexing below using the `selectedIndex` determined from `filteredOptions` (i.e. you are using this index when indexing `option`).
- Is this problematic for programmatic setting of `value`, where I expect `filteredOptions` may be empty? My guess is this is fine, as I think it will all ultimately resolve to `updatedValue` being set to `next` on line 166 and `selectedIndex` being set to -1, which I think is all proper.

packages/nimble-components/src/combobox/tests/combobox.spec.ts
m-akinc: > we aren't allowing the setting of the value to the disabled option

But typing in the option _does_ set the control's value. Am I misunderstanding you?

m-akinc: Changed the tests to do the same as the existing test (commit by clicking away).

packages/nimble-components/src/combobox/index.ts
```diff
@@ -159,11 +159,11 @@ export class Combobox
             const prevSelectedValue = this.options[this.selectedIndex]?.text;
             const nextSelectedValue = this.options[selectedIndex]?.text;
 
-            this.selectedIndex = prevSelectedValue !== nextSelectedValue
-                ? selectedIndex
-                : this.selectedIndex;
+            if (prevSelectedValue !== nextSelectedValue) {
+                this.selectedIndex = selectedIndex;
+            }
```
m-akinc, line 164: No need to reassign the same value. This is a little clearer in terms of intent.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -159,11 +159,11 @@ export class Combobox
...
 
-            updatedValue = this.firstSelectedOption?.text || next;
+            updatedValue = this.firstSelectedOption?.text || updatedValue;
```
m-akinc, line 166: This makes it clear that we're not actually changing the value if there's no `firstSelectedOption`.

packages/nimble-components/src/combobox/index.ts
m-akinc: Once again, I spent a lot more time that I'd like trying to trace exactly what was going on, but I settled on the fact that `options` shouldn't be returning the full set of options if there's a filter and there are no valid options to pick.

m-akinc: Oof, looks like a number of tests failed. Let me handle those before you review again.

change/@ni-nimble-components-5084859c-b0ef-4431-b684-b4733c6d7714.json
jattasNI, line 1: I'm not entirely convinced what the desired behavior would be for this case. A client might assume that setting an item as disabled would prevent the user from setting that as the value. Do we want to say that assumption is invalid? If so then what's the point of setting an item as disabled?

Some options I can think of:
1. Document that clients shouldn't set combobox items as disabled. Then we can choose whatever behavior we want if clients do this (within the bounds of good custom element behavior, ie we can't just crash). Maybe we'd decide to leave the behavior as-is or maybe we'd ignore the disabled attribute.
2. Document that the value can never be set to the value of a disabled item. Then we'd need to brainstorm the desired UX if the user types the value. While I agree the behavior reported in #2252 is clunky, it does achieve the goal of disallowing that value interactively.
3. Decide that disabled items are more of a suggestion to users, but there are valid use cases for allowing those values to be entered manually. Then this PR is going in the right direction.

It's probably worth a team / PO discussion to try to agree on the direction. @m-akinc

packages/nimble-components/src/combobox/index.ts
m-akinc: Turns out it was just one test failing for all three browsers. The "selected" class wasn't getting cleared from the option in `list` autocomplete mode when the filter stopped matching it. Fixed this by overriding `setSelectedOptions` and calling it from `selectedIndexChanged`.

change/@ni-nimble-components-5084859c-b0ef-4431-b684-b4733c6d7714.json
m-akinc: From the discussion, it seems like we're going with option 3. I filed #2302 for us to document usage patterns around disabled combobox options. We still want the control to behave sanely when a disabled option value is typed in, and that's what this PR aims to accomplish.

packages/nimble-components/src/combobox/index.ts
```diff
@@ -682,6 +683,18 @@ export class Combobox
         }
     }
 
+    /**
+     * Need to update even when options is empty.
+     * @internal
+     * @remarks
```
atmgrifter00, line 689: I might remark that this is essentially the same implementation as what was in FAST (w/o the guard).

packages/nimble-components/src/combobox/index.ts
```diff
@@ -575,6 +575,7 @@ export class Combobox
             }
 
             super.selectedIndexChanged(prev, pinnedSelectedIndex);
+            this.setSelectedOptions(); // the base class doesn't call this when no option is selected, but we need to
```
atmgrifter00, line 578: Is it worth describing the "why" of the "but we need to" part? I'm assuming this is because of the scenario where we end up with only a disabled option in the dropdown while typing?

atmgrifter00 (approved):

packages/nimble-components/src/combobox/index.ts
m-akinc: Actually it's because a number of properties won't get updated otherwise. I've expanded the comment to explain.

change/@ni-nimble-components-5084859c-b0ef-4431-b684-b4733c6d7714.json
jattasNI: Could you update the PR description to include a short summary of the expected behavior from the user's point of view?

jattasNI, line 1: I'm still seeing the buggy behavior when autocomplete is set to `none` in the current Storybook build.
https://60e89457a987cf003efc0a5b-nndckhaqkg.chromatic.com/?path=/docs/components-combobox--docs

https://github.com/user-attachments/assets/fe92ff9c-3bfd-45ce-bcc4-1cbab8d83dae

jattasNI (changes requested):

change/@ni-nimble-components-5084859c-b0ef-4431-b684-b4733c6d7714.json
m-akinc: So many code paths.... Fixed now. Had to make sure that when `options` is an unfiltered list, and we're searching for the index of some value, we treat disabled options as not found.

jattasNI (approved):

rajsite:
@atmgrifter00 it looks like there were changes since your last review, could you give it another check JIC?

packages/nimble-components/src/combobox/index.ts
```diff
@@ -153,9 +153,7 @@ export class Combobox
         let updatedValue = next;
 
         if (this.$fastController.isConnected && this.options) {
-            const selectedIndex = this.options.findIndex(
-                el => el.text.toLowerCase() === next.toLowerCase()
-            );
```
atmgrifter00, line 158: Looking at this implementation (recognizing that this is just copied from FAST), I'm concerned that this isn't really the behavior we want. While being case-insensitive makes sense for filtering, actually setting the `selectedIndex` to the first option found in a case-insensitive manner seems wrong. In fact, before your changes, I believe the `inputHandler` did things as I would expect, it would set the `selectedIndex` to the value considering case (but also would then set the `value` at the end, which could actually have resulted in buggy behavior).

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -349,25 +349,48 @@ describe('Combobox', () => {
...
+            {
+                name: 'allows committing disabled option from input (exact case)',
+                text: 'Four'
+            },
+            {
+                name: 'allows committing disabled option from input (with case difference)',
+                text: 'four'
+            }
+        ] as const;
+        parameterizeSuite(filterOptionSuiteData, (suite, name) => {
```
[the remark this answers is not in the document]
m-akinc: That was not expected. I've fixed that and added a test case.

jattasNI (commented):
Resetting my vote because I'd like to do one more round of testing based on the final implementation after the last bit of feedback is addressed

packages/nimble-components/src/combobox/index.ts
m-akinc: So is the expectation that the control will only coerce the value typed in (to match the casing of a dropdown option) when the `autocomplete` mode is `inline` or `both`? I.e. if I type in "s", it will autocomplete to "Sue". Otherwise, if I just commit the value "sue" (either by keyboard interaction or programatically), the control value stays "sue" and the `selectedIndex` is -1. Is that right?

m-akinc: I've updated the code (and one test) to work according to that understanding.

atmgrifter00: So, to be clear, the case of concern is when there is both the option "Sue" and "sue", and specifically when "Sue" is after "sue" in order. In this case, if a user types "Sue" I would expect the filtered list to show both "sue" and "Sue", but I would expect "Sue" to be highlighted (though this part is admittedly subjective). What we _don't_ want (and I don't think your changes allow this, but prior to them I believe this could occur) is that while typing "Sue", "Sue" would be highlighted as expected, but when committing the value, you would find that "sue" was selected instead.

atmgrifter00: Yes, I believe that is right. Admittedly, the scenario I thought I was concerned with can't really happen. I figured if you had both "Sue" and "sue", and typed "S", you would want "Sue" to the one that got selected. However, it appears that you actually can't even type in a capital letter if there is an option that is first matched that doesn't start with a capital letter. Personally, I think this is kinda crappy, but I'm not interested in holding this PR up for existing behavior that is admittedly a bit subjective.

packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -416,7 +475,7 @@ describe('Combobox', () => {
             element.value = 'two';
             await connect();
 
-            expect(element.value).toBe('Two');
+            expect(element.value).toBe('two');
```
atmgrifter00, line 478: You might even consider removing the `value` setters in the template as they are unused by the `Combobox`.

atmgrifter00 (approved):

jattasNI (approved):

PR #2295, opened 2024-07-19
Update jasmine parameterized docs

rajsite:
# Pull Request

## 🤨 Rationale

Updated the jasmine parameterize docs to align with other packages with Getting Started sections and a link from README to CONTRIBUTING.

Also did a major bump so the package jumps to `1.0.0` and can participate in proper semver.

Secret motivation was tired of seeing the badge be orange signifying a pre-1.0.0 release on the top-level readme:
image

## 👩‍💻 Implementation

See above

## 🧪 Testing

Rely on CI

## ✅ Checklist


packages/jasmine-parameterized/CONTRIBUTING.md
```diff
@@ -1,9 +1,11 @@
-# Contributing to Jasmine Parameterized
+# Contributing
 
-## Requirements
+## Getting Started
+
+Make sure you have the following requirements installed:
```
jattasNI, line 5: nit: for some reason this reads weird to me. I think this is slightly better, or could just delete the word entirely
```suggestion
Make sure you have the following prerequisites installed:
```

packages/jasmine-parameterized/CONTRIBUTING.md
```diff
@@ -1,9 +1,11 @@
-# Contributing to Jasmine Parameterized
+# Contributing
```
jattasNI, line 1: Looks like this CONTRIBUTING may have originally been modelled after the one from `xliff-to-json-converter`. You could consider making similar changes there. But that package is touched so infrequently that I don't mind if you leave it as-is.

jattasNI (approved):

packages/jasmine-parameterized/CONTRIBUTING.md
rajsite: Won't do it as part of this PR. Think it'll be on my back of mind tech debt plate to give the treatment to other READMEs. Angular-workspace just renders as a blank one lol

PR #2296, opened 2024-07-19
Fix menu item indentation bug with dynamically slotted icons

### LINKED ISSUE #2266: menu does not lay out correctly if an item's 'start' slot is dynamically provided
# 🐛 Bug Report

In the `nimble-menu`, the alignment of each `nimble-menu-item` is based on whether or not any of the items have an icon provided in their `start` slot. However, this code does not seem to account for an icon being specified dynamically, such as with an `ngIf` binding in Angular. If the icon is provided dynamically, the items don't indent to give space for the icon.

## 🤔 Expected Behavior

The `nimble-menu` along with its `nimble-menu-item` elements should render correctly when one of the menu items has an icon in the `start` slot, even if the icon was slotted after the initial rendering of the menu.

## 😯 Current Behavior

If the only `start` slot icons are dynamic on menu buttons (such as using an `ngIf` binding in Angular), the indention of the items is incorrect, and the menu renders like this:

image

### Workaround

To work around the issue in Angular, you can bind to the `style.visibility` rather than using `ngIf`. For example, this does not work
```html
<nimble-menu-item *ngFor="let color of colors" (change)="onColorSelected(color)">
    <nimble-icon-check *ngIf="color === currentColor" slot="start"></nimble-icon-check>
    {{ color }}
</nimble-menu-item>
```

but, this does work:
```html
<nimble-menu-item *ngFor="let color of colors" (change)="onColorSelected(color)">
    <nimble-icon-check [style.visibility]="color === currentColor ? 'visible' : 'hidden'" slot="start"></nimble-icon-check>
    {{ color }}
</nimble-menu-item>
```

## 🌍 Your Environment

* Version [e.g. 1.8.0]

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2266

## 👩‍💻 Implementation

Proper indentation relies on `startColumnCount` being set on the menu items. This is handled by [`FoundationMenu.setItems`](https://github.com/microsoft/fast/blob/1f9c97df9acae24be44f239bd66b5b2157c7f3e5/packages/web-components/fast-foundation/src/menu/menu.ts#L253). It needs to be called whenever the children of the slotted menu items change. To do this, we need to modify the template so that it watches for any added/removed icons that are children of its menu items (via `children` directive). I've forked the FAST `Menu` class (into separate file) and the template to modify them.

These modifications are in `menu.foundation.ts` rather than `index.ts` because this is the kind of bug fix that we would have upstreamed to FAST.

## 🧪 Testing

Added unit tests and forked FAST `Menu` tests.

## ✅ Checklist


m-akinc:
@mollykreis will you please buddy this change?

packages/nimble-components/src/menu/tests/menu.spec.ts
```diff
@@ -1,11 +1,64 @@
...
+
     it('should export its tag', () => {
         expect(menuTag).toBe('nimble-menu');
     });
 
     it('can construct an element instance', () => {
         expect(document.createElement('nimble-menu')).toBeInstanceOf(Menu);
     });
+
+    it('properly indents items when an icon is slotted or removed after connecting', async () => {
```
mollykreis, line 27: Should this test be split into multiple tests? It could be split by adding vs removing and for menu items vs anchor menu items. There are quite a few different things all be tested in this single test right now (though I understand it's all following the same code path).

packages/nimble-components/src/menu/index.ts
```diff
@@ -11,10 +12,51 @@ declare global {
...
 /**
  * A nimble-styled menu
  */
-export class Menu extends FoundationMenu {}
+export class Menu extends FoundationMenu {
+    private readonly baseItemsChanged: ItemsChangeHandler;
+
+    private readonly mutationObserver: MutationObserver = new MutationObserver(
+        mutations => {
+            for (const mutation of mutations) {
```
mollykreis, line 28: > we have to jump through a couple hoops to override/call them. I think this is still preferable to copying the whole class implementation over and modifying it.

I'm curious what it would look like to copy the class and modify it. How big of a change would that modification be. The current approach requires `as` casts and `eslint-disable` statements, which doesn't seem ideal.

m-akinc: I forked the Menu class from FAST and fixed all the lint issues. If you prefer the way I had it before, I can revert that commit.

mollykreis: I believe your recent change to fork the Menu class is more aligned with our strategy moving forward. The FAST tests need to be forked as well.

packages/nimble-components/src/menu/index.ts
```diff
@@ -14,7 +27,387 @@ declare global {
...
+        newValue: HTMLElement[]
+    ): void {
+        // only update children after the component is connected and
+        // the setItems has run on connectedCallback
+        // (menuItems is undefined until then)
+        if (this.$fastController.isConnected && this.menuItems !== undefined) {
+            this.setItems();
+        }
+
+        this.mutationObserver.disconnect();
```
mollykreis, line 180: I would probably be a good idea to disconnect the mutationObserver when the component is disconnected and start observing when the component is reconnected.

mollykreis (approved):

packages/nimble-components/src/menu/index.ts
m-akinc: I noticed that `setItems` is probably a better place to observe all the menu items, so I moved the calls there. Then I only had to `disconnect` in the `disconnectedCallback`.

packages/nimble-components/src/anchor-menu-item/index.ts
```diff
@@ -35,7 +35,7 @@ export class AnchorMenuItem extends AnchorBase {
      * @internal
      */
     @observable
-    public startColumnCount: MenuItemColumnCount = 0;
+    public startColumnCount?: MenuItemColumnCount;
```
m-akinc, line 38: Minor, but seems more appropriate not to have an `"indent-0"` class set on the item initially (though the normal `nimble-menu-item` initially has the class `"indent-undefined"` because FAST template doesn't handle `undefined` properly).

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
m-akinc, line 1: Unrelated to my changes (and I'm not sure it had any negative effect) but I fixed this anyway.

packages/nimble-components/src/menu/template.ts
```diff
@@ -0,0 +1,30 @@
...
+        slot="${x => {
+            if (x.slot) {
+                return x.slot;
+            }
+            return x.isNestedMenu() ? 'submenu' : '';
+        }}"
+        role="menu"
+        @keydown="${(x, c) => x.handleMenuKeyDown(c.event as KeyboardEvent)}"
+        @focusout="${(x, c) => x.handleFocusOut(c.event as FocusEvent)}"
+        ${children({ property: 'itemIcons', subtree: true, selector: ':scope > * > [slot="start"]' })}
```
m-akinc, line 25: This is the only functional change to the original template.

packages/nimble-components/src/menu/menu.foundation.ts
```diff
@@ -0,0 +1,403 @@
...
+        if (
+            this.$fastController.isConnected
+            && this.menuItems !== undefined
+            && (oldIcons.length !== newIcons.length
+                || oldIcons.filter(x => newIcons.includes(x)).length
+                    !== oldIcons.length)
+        ) {
+            this.setItems();
+        }
+    }
```
m-akinc, line 245: This is the only functional difference from the original.

packages/nimble-components/src/menu/tests/menu.spec.ts
```diff
@@ -8,4 +15,121 @@ describe('Menu', () => {
...
+                expect(item3.classList.contains('indent-1')).toBeTrue();
+            });
+
+            it('does not indent newly added item (without icon) when there were no items with icons', async () => {
+                const item3 = document.createElement(name);
+                element.appendChild(item3);
+                await waitForUpdatesAsync();
+
+                expect(item3.classList.contains('indent-0')).toBeTrue();
+            });
```
m-akinc, line 104: Seems like a weird test case, but I discovered that FAST's menu item code [apparently defaults the indentation level to 1](https://github.com/microsoft/fast/blob/1f9c97df9acae24be44f239bd66b5b2157c7f3e5/packages/web-components/fast-foundation/src/menu-item/menu-item.ts#L169) instead of 0, for some reason, so this can legitimately fail if we're not calling `Menu.setItems()` when new items are added.

packages/nimble-components/src/menu/tests/menu.spec.ts
```diff
@@ -8,4 +15,121 @@ describe('Menu', () => {
...
+                const icon = document.createElement(iconCheckTag);
+                icon.slot = 'start';
+                item1.appendChild(icon);
+                await waitForUpdatesAsync();
+
+                expect(item1.classList.contains('indent-1')).toBeTrue();
+                expect(item2.classList.contains('indent-1')).toBeTrue();
+            });
+
+            it('does not indent items when the only icon is removed', async () => {
```
jattasNI, line 119: Came here to request this test be added, discovered it already was. 🙌

packages/nimble-components/src/menu/menu.foundation.ts
```diff
@@ -0,0 +1,403 @@
+import { DOM, observable } from '@microsoft/fast-element';
```
jattasNI, line 1: nit: may as well add a comment linking to the original source in this file too like you did in the tests.

jattasNI (approved):

packages/nimble-components/src/menu/template.ts
```diff
@@ -0,0 +1,30 @@
...
+// prettier-ignore
+export const template: FoundationElementTemplate<
+ViewTemplate<Menu>
+> = () => html`
+    <template
+        slot="${x => {
+            if (x.slot) {
+                return x.slot;
+            }
+            return x.isNestedMenu() ? 'submenu' : '';
```
rajsite, line 20: returning `void 0` (which is just `undefined`) removes the slot attribute which is a different outcome than setting a slot to empty string: https://github.com/microsoft/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/menu/menu.template.ts#L15

Is there a reason for the effective change? Can we just return `undefined` which was the original behavior?

packages/nimble-components/src/menu/template.ts
rajsite, line 25: For this to be reliable we would need to include an attrbuteFilter for slot as that can change programmatically at a different time than dom insertion / removal: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#attributefilter

packages/nimble-components/src/menu/menu.foundation.ts
```diff
@@ -0,0 +1,403 @@
...
+        // Unintuitively, providing a selector (i.e. ':scope > * > [slot="start"]') to
+        // the children directive does not limit when/how often itemIcons gets updated.
+        // All it does is filter the elements that are assigned to itemIcons.
+        // To avoid unneccessary calls to setItems, we need to check if there was actually a change.
+        if (
+            this.$fastController.isConnected
+            && this.menuItems !== undefined
+            && (oldIcons.length !== newIcons.length
+                || oldIcons.filter(x => newIcons.includes(x)).length
+                    !== oldIcons.length)
```
rajsite, line 241: Can we use Sets to avoid this n*m search?

change/@ni-nimble-components-a9297ed4-57f8-4273-b530-3fd2554c28d5.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
m-akinc: This is what I see in our shipping Storybook:
menu-icon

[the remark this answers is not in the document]
jattasNI: Huh, I try exactly the same thing (macOS Firefox) and it indents correctly. But if you can reproduce it before your fix and not reproduce it after then I think that's sufficient testing for now.

packages/nimble-components/src/anchor-menu-item/index.ts
m-akinc: `startColumnCount` is assigned to menu items only by the menu, except for the initialization in `AnchorMenuItem` that I removed. I felt it would be more expected that the item not have a `startColumnCount` (and not have any `"indent-N"` class set on it) except when the menu assigns it. However, the regular `MenuItem` already (weirdly) applies an `"indent-undefined"` class before acted on by the menu, so that kind of undercuts my intent. I'm fine reverting back to initializing it to 0.

packages/nimble-components/src/menu/menu.foundation.ts
```diff
@@ -0,0 +1,405 @@
...
+        if (this.$fastController.isConnected && this.menuItems !== undefined) {
+            this.setItems();
+        }
+    }
+
+    private itemIconsChanged(oldIcons: Element[], newIcons: Element[]): void {
+        // Unintuitively, providing a selector (i.e. ':scope > * > [slot="start"]') to
+        // the children directive does not limit when/how often itemIcons gets updated.
+        // All it does is filter the elements that are assigned to itemIcons.
+        // To avoid unneccessary calls to setItems, we need to check if there was actually a change.
```
rajsite, line 238: Useful to know but not specific to this spot, it's general behavior of selector feature of children directive. Think we can leave it off or simplify to something like "detect if slotted icon element instances changed"

rajsite (approved):

Issue #2183, opened 2024-06-10
Built-in label for select and combobox


m-akinc:
# 🙋 Feature Request

## 😯 Problem to Solve

(Copied from [AzDO feature](https://ni.visualstudio.com/DevCentral/_workitems/edit/2688093))
Most Nimble controls offer built in labels but select and combobox don't, meaning clients have had to create and style labels manually.

This will also resolve some accessibility bugs:
https://github.com/ni/nimble/issues/94
https://github.com/ni/nimble/issues/280

## 💁 Proposed Solution

We should update their templates to include labels, document the feature (hopefully including guidance about when to use labels and when not), and coordinate changing client apps to use the feature.

## 📋 Tasks

- [x] HLD
- [x] Implementation in nimble-components
- [ ] Determine best way to document feature / provide usage guidance (work with @fredvisser on this)
   - Related questions from Jesse on PRs:
       - Should it always be provided? Only provided if the purpose of the control isn't obvious from its placeholder / default value? Do we need a way to provide it for a screen reader but not show it visibly?
       - Label vs placeholder vs neither
    - Team discussion 8/16: General guidance will be a label should always be provided, and placeholder shouldn't be used instead of the label. Still needs to be documented somewhere though.
- [x] Publicize feature: SLE demo meeting in SLE demos and/or UI Working Group
- [x] Publicize feature: UI Working Group
   - @jattasNI will add something on UIWG agenda about it
- [x] Come up with list of systemlink-lib-angular usages that need to be updated
   - https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/777610

PR #2299, opened 2024-07-23
HLD: Built-in labels for select/combobox

msmithNI:
# Pull Request

## 🤨 Rationale

HLD for #2183

### DIFF
--- a/packages/nimble-components/src/select/specs/built-in-labels.md
+++ b/packages/nimble-components/src/select/specs/built-in-labels.md
@@ -11,8 +11,9 @@ Built-in labels will reduce the effort needed for clients to consume these contr
 [Built-in label for select and combobox](https://github.com/ni/nimble/issues/2183)
 
 Related issues:
-- [Resolve label accessibility issues](https://github.com/ni/nimble/issues/94)
-- [Fix accessibility issues in Angular Example app](https://github.com/ni/nimble/issues/280)
+
+-   [Resolve label accessibility issues](https://github.com/ni/nimble/issues/94)
+-   [Fix accessibility issues in Angular Example app](https://github.com/ni/nimble/issues/280)
 
 ## Implementation / Design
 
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+
+### API (Slots)
+
+Default Slot:
+
+-   The `nimble-list-option`s in the control are placed in this slot currently.
+-   We will add support for specifying the label as text content in the same slot. For example:
+    ```html
+    <nimble-select>
+        Select Label Text
```
jattasNI, line 30: We'll need to support putting the label in a `<span>` in the default slot too. This would allow Angular clients to set an `i18n` tag on it to trigger its content to be localized. Without that they'd have to put the tag on the `<nimble-select>` which would trigger the HTML tags of the options to be localized too.

You can see an example of this in the [docs for the list option group default slot](https://nimble.ni.dev/storybook/?path=/docs/components-select--docs#list-option-group).

packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+        <nimble-list-option>A</nimble-list-option>
+        <nimble-list-option>B</nimble-list-option>
+    </nimble-select>
+    ```
+
+### Templates / Styling
```
jattasNI, line 36: Might be worth calling out that the label will be dimmed if the control is disabled since I think we've advertised that as one of the benefits of adding built-in labels. A lot of clients have gotten that wrong and forgotten to dim their custom labels.

jattasNI, line 36: A couple layout considerations:
1. if the label is not provided then should vertical space be reserved for it? I presume not, since that will maintain layout compatibility for existing clients and that matches what other labels do.
2. if the label doesn't fit in the width of the control what will happen? For our other components with labels it looks like they grow to fit the label.

packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+## Alternative Implementations / Designs
+
+**Label API:** We have several different paradigms for specifying label text in Nimble, in addition to the default slot:
+
+-   A `label` slot, with the expectation clients will add a `label` element in it (`nimble-radio-group`)
+-   A `label` attribute (`nimble-option-group`, though text content in the default slot is also supported)
+
+The default slot approach was chosen for this feature as it's the most common in Nimble, and the simplest to use.
+
+## Open Issues
```
jattasNI, line 75: Not sure if this is the place to figure this out (especially with Fred about to be OOO) but do we know what usage guidance we'll have for setting labels? Should it always be provided? Only provided if the purpose of the control isn't obvious from its placeholder / default value? Do we need a way to provide it for a screen reader but not show it visibly?

jattasNI (approved):

mollykreis (approved):

packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+
+`nimble-rich-text-mention-listbox` will not support this feature, meaning it will not contain a `label` element in its template. The current plan is to not address any ARIA label issues on it either (see Open Issues).  
+However, since we have shared dropdown styling, its template may need minor updates (e.g. to add the same parent/container `div` as select/combobox are getting), but we'll ensure it has the same visual appearance as before.
+
+## Alternative Implementations / Designs
+
+**Label API:** We have several different paradigms for specifying label text in Nimble, in addition to the default slot:
+
+-   A `label` slot, with the expectation clients will add a `label` element in it (`nimble-radio-group`)
+-   A `label` attribute (`nimble-option-group`, though text content in the default slot is also supported)
```
rajsite, line 71: This one aligns with the native optgroup api and is expected: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+### Impact to nimble-rich-text-mention-listbox
+
+`nimble-rich-text-mention-listbox` will not support this feature, meaning it will not contain a `label` element in its template. The current plan is to not address any ARIA label issues on it either (see Open Issues).  
+However, since we have shared dropdown styling, its template may need minor updates (e.g. to add the same parent/container `div` as select/combobox are getting), but we'll ensure it has the same visual appearance as before.
+
+## Alternative Implementations / Designs
+
+**Label API:** We have several different paradigms for specifying label text in Nimble, in addition to the default slot:
+
+-   A `label` slot, with the expectation clients will add a `label` element in it (`nimble-radio-group`)
```
rajsite, line 70: The `nimble-radio-group` requiring a slot label I'd consider an inconsistency / tech debt we should capture to improve.

rajsite (approved):

packages/nimble-components/src/select/specs/built-in-labels.md
msmithNI: Updated to state that.

packages/nimble-components/src/select/specs/built-in-labels.md
msmithNI: #2306

packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -0,0 +1,77 @@
...
+    </nimble-select>
+    ```
+
+### Templates / Styling
+
+Template updates:
+
+-   A new `label` element will be added at the root of the control template, which will reflect the default slotted content (minus the list options which are already handled elsewhere)
+-   A new container `div` will be added at the same level, to contain the control part (`div[part="control"]`)
+-   The element host will become `inline-block` so the label is laid out correctly. (Formerly `inline-flex`, which will move to the container `div`)
```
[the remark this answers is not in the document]
msmithNI: Updated to state that this was prototyped as `inline-block` but we should look at `inline-flex`/`inline-grid` instead.

packages/nimble-components/src/select/specs/built-in-labels.md
msmithNI: Good questions. Don't think we need to block this HLD on the answers, so added a task in the linked issue about docs/ usage guidance, and captured your questions there.

packages/nimble-components/src/select/specs/built-in-labels.md
msmithNI: Good points, added to the HLD. (1 - no reserved space if label not present; 2- said we'll probably grow to fit like the other controls)

msmithNI: Added

packages/nimble-components/src/select/specs/built-in-labels.md
msmithNI: Good point, added that detail.

PR #2300, opened 2024-07-23
Table data assignment bug

### LINKED ISSUE #2265: Cannot reassign same data instances to table when hierarchy is enabled
# 🐛 Bug Report

With the `nimble-table`, it is expected that you can pass the same array/object instances with modified property values to `setData()` and have those updates be reflected in the rendered table. This works as expected when hierarchy is not enabled -- in fact, there's [a test that verifies this works as expected](https://github.com/ni/nimble/blob/edc07e18e9698194267f268154d62fd81a9cb211/packages/nimble-components/src/table/tests/table.spec.ts#L395). However, when hierarchy is enabled (specifically when `parentIdFieldName` is configured), this does not work. It looks like the issue is that the `arrayToTree` code path does not ensure a copy of each data record is made in order for TanStack to recognize the change.

## 🤔 Expected Behavior

You should be able to write code like this and have the table reflect the changes:
```ts
const data= [{ a: 'hello' }];
table.setData(data);
// table renders 'hello'

data[0].a = 'world';
table.setData(data);
// table renders 'world' if parentIdFieldName is not configured
// table still renders 'hello' if parentIdFieldName is configured
```

## 😯 Current Behavior

See expected behavior

## 💁 Possible Solution

The `arrayToTree` copy code path in the table needs to ensure that a copy of the data records are made so that TanStack recognizes that there was an update to the data.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #2265 

## 👩‍💻 Implementation

The `arrayToTree` code path for setting data in the table does not make a copy of each record in the same way that the other `setData` code paths do. Therefore, I updated the `arrayToTree` function to copy the record as it turns the array into a tree.

## 🧪 Testing

- Added a new unit test that failed before my change and now passes
    - Note, there already is a similar test for non-hierarchical data, so I only added a new test for hierarchical data.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/models/array-to-tree.ts
+++ b/packages/nimble-components/src/table/models/array-to-tree.ts
@@ -55,7 +55,9 @@ export function arrayToTree<TData extends TableRecord>(
             orphanIds.delete(itemId);
         }
 
-        // add the current item's data to the item in the lookup table
+        // Add the current item's data to the item in the lookup table.
+        // Make a shallow clone of the record to avoid holding a reference to the client's data object.
+        // This also ensures that a data update will always result in a row's record being a new object.
         lookup[itemId]!.clientRecord = { ...item };
         const treeItem = lookup[itemId]!;
 
--- a/packages/nimble-components/src/table/models/array-to-tree.ts
+++ b/packages/nimble-components/src/table/models/array-to-tree.ts
@@ -56,7 +56,7 @@ export function arrayToTree<TData extends TableRecord>(
         }
 
         // add the current item's data to the item in the lookup table
-        lookup[itemId]!.clientRecord = item;
+        lookup[itemId]!.clientRecord = { ...item };
         const treeItem = lookup[itemId]!;
 
         // Add the index to the item
--- a/packages/nimble-components/src/table/models/array-to-tree.ts
+++ b/packages/nimble-components/src/table/models/array-to-tree.ts
@@ -55,10 +55,8 @@ export function arrayToTree<TData extends TableRecord>(
             orphanIds.delete(itemId);
         }
 
-        // Add the current item's data to the item in the lookup table.
-        // Make a shallow clone of the record to avoid holding a reference to the client's data object.
-        // This also ensures that a data update will always result in a row's record being a new object.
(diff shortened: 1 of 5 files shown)


mollykreis:

atmgrifter00 (approved):

jattasNI (approved):

packages/nimble-components/src/table/models/array-to-tree.ts
```diff
@@ -56,7 +56,7 @@ export function arrayToTree<TData extends TableRecord>(
         }
 
         // add the current item's data to the item in the lookup table
-        lookup[itemId]!.clientRecord = item;
+        lookup[itemId]!.clientRecord = { ...item };
```
rajsite, line 59: Could we add a comment here and in the other spots in data manager about why we are doing a shallow clone?

https://github.com/ni/nimble/blob/f24f5ea3938d8ba15e4f896090f64c0d5021c4f7/packages/nimble-components/src/table/models/data-hierarchy-manager.ts#L32

https://github.com/ni/nimble/blob/f24f5ea3938d8ba15e4f896090f64c0d5021c4f7/packages/nimble-components/src/table/models/data-hierarchy-manager.ts#L40

It's not necessary for the arrayToTree algorithm, it's for a side-effect elsewhere, so might be good to call out.

Also the non-hierarchy code path seems to get away with a shallow clone of the top-level array instead of shallow clones of each individual record. 

https://github.com/ni/nimble/blob/f24f5ea3938d8ba15e4f896090f64c0d5021c4f7/packages/nimble-components/src/table/index.ts#L947-L948

Just to confirm, is it necessary for the hierarchy code path to be different and require shallow clones of each record or is there something missed that could avoid that extra work?

mollykreis: The short answer is that we need to make a shallow clone of records in all cases that `setData()` is called, and we don't need to make a shallow clone of records in any other cases. Since `arrayToTree` gets called in multiple code paths, I've updated the approach for cloning to clone in `setData()` itself rather than in the data hierarchy manager.

Issue #2282, opened 2024-07-17
Create theme-unaware high-level tokens


rajsite:
# 🙋 Feature Request

## 😯 Problem to Solve

We currently have:
- theme unaware base tokens such as `$ni-nimble-base-black-85` that are static values defined as raw values
- theme aware high-level tokens such as `$ni-nimble-application-background-color` that are dynamic values that change at runtime based on CSS Custom Property values

We should consider having:
- theme unaware high-level tokens such as `$ni-nimble-application-background-color-light-theme`, `$ni-nimble-application-background-color-dark-theme`, `$ni-nimble-application-background-color-color-theme` that are static values defined as raw values
- we should then build the theme-aware high-level tokens from those theme-unaware high-lvel tokens

Use-cases:
- Mobiscroll does not support theming using dynamic CSS custom properties and needs statically defined SCSS values ([discussion](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/750620#1720039854))
- Gainsights does not support JavaScript execution which is needed for our theme aware high-level tokens to update dynamically. We could instead generate static CSS styles that use [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) to have top-level page theme awareness (but no concept of theme-providers to override themes on specific regions of a page).

## 💁 Proposed Solution

Partially described above but we need to determine:
- The naming scheme for theme unaware high-level tokens
- How to represent the theme-unaware high-level tokens in style dictionary
- Generation of static CSS that uses the `prefers-color-scheme` media query (current use-case being gainsights so should see what is actually supported first)

jattasNI:
https://github.com/ni/nimble/pull/2305/files/2699ec4760c024a1d150cea9b3df50c01992c46c revealed some of the challenges in doing this:
1. we need hierarchy in base tokens to avoid duplicating hex values and to provide clarity about which tokens are atomic color values and which are mapping to component parts / variants
2. we need a naming scheme that incorporates the many subtle parts / variants / themes / token types
3. we want a way to simplify the theme-aware token creation so that it doesn't require a single file importing hundreds of base tokens and manually mapping them

rajsite:
Related, making token values easier to copy for static css users (i.e. users that are not able to budget time towards scss tooling integration, ex LabVIEW HTML Diff Reports / Gainsights): https://github.com/ni/nimble/pull/2372

rajsite:
[...]

PR #2305, opened 2024-07-24
Add missing base color tokens for calendar theme aware tokens

aagash-ni:
# Pull Request

## 🤨 Rationale
Added the missing base color tokens required to create calendar theme-aware tokens according to the [mockup](https://www.figma.com/design/VS7WavgofSHJG8AUucW1NJ/Calendar-Workflow?node-id=712-21348&t=KRNNkEXzwSvmUQTA-0).

Figma Mockup: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/118969/Schedule-Test-Plans

Azdo Discussion for reference: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/750620#1720658507

## 👩‍💻 Implementation
- Added the necessary base colors for the calendar theme-aware tokens

Note: Calendar theme-aware tokens will be added in the follow-up PR

## 🧪 Testing
NA

## ✅ Checklist


change/@ni-nimble-components-92f16ac4-28b7-4ff8-a111-f9a46fbc86aa.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Hey @aagash-ni, @rajsite and I took a quick look at this draft PR to see the direction you're heading. We realized that this infrastructure will require some design work by the Nimble team which we're not ready to devote bandwidth to just yet.

** Near term recommendation **

1. any base tokens which are exposing new color hex values should be added using our existing base token naming conventions (I didn't check if there are any)
2. the SLE calendar integration can continue using base token infrastructure as it does today
3. I will update #2282 to capture concerns with the other parts of this PR, but for now we suggest not adding new base tokens or theme-aware tokens that are specific to the calendar / mobiscroll

** Longer term recommendation **

1. we'll work with @fredvisser to prioritize our team doing the design work
2. once the design work is done we can revive this concept and add the tokens in this PR to nimble (but following the new design) and then update the SLE integration to use them. We'd figure out ownership of that work when it's unblocked.

Hopefully this doesn't derail you too much, but please let us know if you disagree with this direction. And thanks for the draft PR, it was valuable in helping crystallize this proposal.

aagash-ni: Sure, Jesse.

To proceed with achieving the mockup:

- We'll add the required base tokens with the appropriate naming conventions today.
- Basically the event cards used in the Mobiscroll calendar are templated Angular components. To style these, we'll need the standard theme-aware tokens we typically use.
- For the event cards, I'll introduce a new set of theme-aware tokens (if they don't already exist) based on our needs.
- For calendar styling alone will continue with non theme aware tokens.

I hope this is fine for now to proceed. I'd like to hear your thoughts.

jattasNI: That generally seems ok. Regarding
>For the event cards, I'll introduce a new set of theme-aware tokens (if they don't already exist) based on our needs.

We'll just need to make sure these tokens fit well with our conventions. They might be able to follow similar patterns to [graph tokens](https://nimble.ni.dev/storybook/?path=/docs/components-graph--docs) but I'm just less sure of how we'd name them. I'm worried "event card" might too specific of a concept for Nimble but maybe it's ok or maybe we can find a more generic concept. I'd propose you put up a PR (or update this one) with a proposal that follows the [theme aware token naming guidelines](https://github.com/ni/nimble/blob/b814c437bff07dd565ca25facebc0a30e837b428/packages/nimble-components/CONTRIBUTING.md#naming) and we can discuss it in the PR.

aagash-ni: I have added some essential base tokens (currently unavailable) that are needed to create theme-aware tokens for the calendar. The theme-aware tokens for the calendar will be included in a follow-up PR.

Kindly have a look and let me share your thoughts @jattasNI @rajsite

change/@ni-nimble-tokens-b264b0ee-0471-4955-99c5-10e8b68dff74.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jattasNI, line 2: This is adding new APIs so it should be minor rather than patch
```suggestion
  "type": "minor",
```

packages/nimble-tokens/source/styledictionary/properties/colors.json
```diff
@@ -176,6 +176,30 @@
     },
     "NI Sea Dark 2": {
       "value": "#00685Fff"
+    },
```
jattasNI, line 179: Some questions about these colors:

1. Where do the names come from? Do they match values that @NIbokeefe is using in a Visual Design document?
2. What is the naming scheme? What do the numbers indicate? Do they mean the same thing as the numbers on existing tokens?
3. It looks like our pattern has been to use RGBA values rather than RGB values (i.e. to include "ff" for the alpha value at the end). Is there a reason not to do this for these colors?

Depending on the answers to these questions we should think about the naming convention. For example, if it's necessary to express these in RGB do we want to use "Opaque" to signify that? Should the number go at the end or next to the thing it's modifying? Should we sort them alphabetically / keep similar colors together / keep all the opaque colors together?

change/@ni-nimble-tokens-b264b0ee-0471-4955-99c5-10e8b68dff74.json
aagash-ni: Updated

change/@ni-nimble-tokens-b264b0ee-0471-4955-99c5-10e8b68dff74.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Note that these changes will cause a Chromatic diff because we have a story that displays all the base tokens. But the current diffs are incorrect as they show some colors from a previous version of this PR. This is likely due to a known issue with how Chromatic builds are triggered. To retrigger the snapshots you can modify the guid at the bottom of `packages/storybook/.storybook/preview.js` and include that change in your PR.

packages/nimble-tokens/source/styledictionary/properties/colors.json
jattasNI: Thanks for the explanation and the changes. The new color tokens seem good: they match existing patterns in that they are sourced from VxD, follow consistent naming, and specify values in RGBA.

I didn't completely follow your explanation about the remaining issue. I'll summarize my understanding; please clarify if I misinterpreted:
1. in the gif, the background of the card gets darker while it is being resized and the original size remains visible, both of which are undesirable
2. both are caused by using base tokens with transparency instead of opaque tokens
3. these new tokens are opaque and we can use them to generate theme-aware tokens with transparency to solve this problem

If that's accurate then I think the current approach is good.

Some more background in case that's not accurate: sometimes we have to compute opaque values whose values are equal to what would result if a given foreground color was made partially transparent and displayed over a given background color. We generally avoid doing this because we don't always know what the background color will be (for example the Color theme can be used with any background). But we've done it in a few places when necessary and we could talk about doing it here if needed.

change/@ni-nimble-tokens-b264b0ee-0471-4955-99c5-10e8b68dff74.json
jattasNI: Looks like you updated `dependentChangeType` but not `type`. The bump for the nimble-tokens package is controlled by `type` so it should definitely be `minor` because that's where the new APIs are being added. The bump for `nimble-components` package is controlled by `dependentChangeType`. It doesn't re-export these APIs and we're not adding any new ones, so it should be `patch`.

jattasNI (commented):
I'd like to take one more look once the remaining feedback is addressed but this is looking good.

packages/nimble-tokens/source/styledictionary/properties/colors.json
aagash-ni: > both are caused by using base tokens with transparency instead of opaque tokens

These issues arise because theme-aware tokens are generated by converting base tokens to hex using the `hexToRgbaCssColor` function, which returns the color in `rgba` format. This conversion causes problems, especially when these theme-aware tokens are utilized. 

image

To address this, I have added a few base tokens with the actual hex values. This approach helps avoid the issues that occur when the desired tokens are generated with transparency using `hexToRgbaCssColor` as `rgba`.

change/@ni-nimble-tokens-b264b0ee-0471-4955-99c5-10e8b68dff74.json
aagash-ni: Modified GUID and the storybook build got updated

jattasNI (approved):

jattasNI:
Changes are now pretty trivial, don't think this needs review from @rajsite or @fredvisser but feel free to comment afterwards or on the follow up PR. I marked this for auto-merge.

Issue #1228, opened 2023-05-08
Make textarea height optionally auto-resizable


pakdev:
## 😯 Problem to Solve

It would be nice if nimble-text-area offered auto-resize like [Angular Material](https://material.angular.io/components/input/overview#auto-resizing-textarea-elements) and [PrimeNG](https://primeng.org/inputtextarea#autoresize). This would allow the input to grow vertically vs. show scrollbars. See [this example](https://stackblitz.com/edit/angular-ivy-pcaccy?file=src/app/app.component.ts) implemented with a contenteditable div:

![textarea](https://user-images.githubusercontent.com/1248927/236914377-a4279bc0-60d0-4bed-b392-6c502fe6c132.gif)

## 💁 Proposed Solution

Add an attribute like `[autoResize]="true"`

rajsite:
Looks like there is some progress on autosizing textareas being implemented natively: https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/

nivetha-ni:
Hi, we would also love it if textareas supported auto-sizing for our PDV use case in SLE. We have a bunch of data selection options that we display in the textarea based on the selections the user makes.

![image](https://github.com/ni/nimble/assets/102650079/5f2f2156-6723-4ccc-9a9b-612e01404292)

It would be helpful if the textarea auto-sized based on the text shown and subsequently grows in height when user adds more selections.

rajsite:
This should be native with field-sizing: content eventually https://x.com/jh3yy/status/1799460095585308690?t=4md5aMDQYxUQNqcRNYYdng&s=19

pakdev:
Well it's been almost two years. We could have created a polyfill we could use _today_ but we're still stuck hoping for eventual field-sizing support

rajsite:
> Well it's been almost two years. We could have created a polyfill we could use _today_ but we're still stuck hoping for eventual field-sizing support

Since we are being cheeky

![Image](https://github.com/user-attachments/assets/3b84ac64-89a2-4b01-adf2-3225926698a5)

rajsite:
Chatted with Peter offline, while `field-sizing: content` isn't available in all browsers yet they may be able to workaround it temporarily in a demo by targeting the [CSS Shadow Part](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_shadow_parts) `control` with something like the following:

```css
nimble-text-area::part(control) {
    height: auto;
[...]

PR #2307, opened 2024-07-29
Add token representing the height of the table with all rows visible

### LINKED ISSUE #1624: nimble-table can't be styled to show all rows without a scrollbar
## 😯 Problem to Solve

There isn't a way to style the table to show all rows without a scrollbar unless you happen to know the pixel height in order to achieve that with your data set.

With the upgraded version of TanStack that will come when #1617 is resolved setting `height: auto` on the table renders no rows. There are a few workarounds that can be applied in the CSS for this, but even with those workarounds (or before the TanStack upgrade), a table with `height: auto` is extremely inefficient for the browser to render and can cause the browser to hang if the table is too large. This is because the table re-renders for every `overscan` number of rows set in the table's virtualizer. Currently, the value of `overscan` is 3, so the table will render with 0 rows, 3 rows, 6 rows, 9 rows, etc until all rows have been rendered.

## 💁 Proposed Solution

I don't have a concrete proposal, but here are a few possible ideas along with their limitations:

**Disable the virtualizer**
We could have a mode on the table to disable the virtualizer. Then a client could set `height: auto` and have the browser be able to render a single time with all the rows rather than re-render for each `overscan` number of rows. However, this doesn't account for styling that limits the height of the table (e.g. `height: auto; max-height: 700px`). In that case, the virtualizer still needs to be utilized to ensure that a large number of rows that can't fit within the `max-height` renders & scrolls performantly.

**Set a row count to render**
We could have an input on the table to configure how many rows should be rendered. This could be similar to the `rows` attribute on the `textarea`. Some questions with this approach are: 
    - If the row count on the table is set to a value larger than the number of records in the table, should space be reserved? Or, should the value represent the maximum number of rows to render? Implications of possibly having a dynamic number of group rows should be considered in this.
    - How does CSS styling of the height interact with setting this value?

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #1624

## 👩‍💻 Implementation

The table already knows the combined height of all the rows in the table because it uses this value to size the `table-scroll` container. Therefore, I was able to create a token that gets assigned a value within the scope of the table styling that has a value of the size of the height the table needs to show all rows. The token's name is `tableFitRowsHeight`.

The table also now sets a default max-height on itself to keep a client from accidentally using `tableFitRowsHeight` and trying to render thousands of rows at once.

To use the new token, a client would create their table exactly as before and add the following style:
```scss
nimble-table {
    height: $ni-nimble-table-fit-rows-height;
}
```

## 🧪 Testing

- Wrote chromatic tests that verify the height of the table is correct when its height is set to the new token value when:
    - There are 5 records, with & without grouping
    - There are 10 records, with & without grouping
    - There are 50 records, with & without grouping

## ✅ Checklist


mollykreis:
@atmgrifter00, @rajsite, can you take a look at this draft PR and let me know if you agree with the direction? If so, I'll add some tests for it and update the storybook docs referencing the issue being resolved.

atmgrifter00 (approved):

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: I think we are in the right direction. I haven't tried to fit all the things that come to mind together yet so I'm going to layout the puzzle pieces if you want to coalesce them:
- We also want this for textarea: https://github.com/ni/nimble/issues/1228
- And rich text: https://github.com/ni/nimble/pull/1446#discussion_r1310241464
- The feature is coming natively to browsers as a CSS property: https://developer.chrome.com/docs/css-ui/css-field-sizing
- The native feature has semantics about how it relates to other size settings: https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing#field-sizing_interaction_with_other_size_settings
- I didn't do any research into common attribute names for similar behavior in other design systems
- Want to choose an attribute name that will work well with the native concept, across the other controls, will behave consistently in sizing behavior among all of them, and we can pick up for rich text / text area trivially leveraging the native feature in the future.

mollykreis: I don't have answers for everything, but here are some notes from my investigation into some of these topics:

>I didn't do any research into common attribute names for similar behavior in other design systems
I didn't find any common attribute names. These are the attributes I did find:
- [Angular text area](https://material.angular.io/cdk/text-field/overview#automatically-resizing-a-textarea) uses `cdkTextareaAutosize` "attribute"
- [Shoelace text area](https://shoelace.style/components/textarea#expand-with-content) uses `resize` attribute with possible values of `"none"`, `"auto"`, or `"vertical"`
- [PrimeNG](https://primeng.org/inputtextarea#autoresize) has a boolean attribute `autoResize` attribute
- The text area didn't appear to be resizable in the Carbon design system

>Curious if we have a story of how it may interact with CSS resize which certainly makes sense for textarea and richtext and I think actually could make sense for a data table as well
- I tried out CSS resize on the table with & without auto sizing, and it worked well in both cases. Specifically, in both cases, the table correctly adjusted the visible area of rows based on the height of the table. I temporarily added a control to the table storybook page if you wanted to play around with it. So, I think the "story" of how the two features interact is that the table will auto size by default, but interactively changing the size of the table would take precedence over the automatically determined size. I think something similar would also make sense for other components.

packages/nimble-components/src/table/styles.ts
```diff
@@ -25,10 +26,24 @@ export const styles = css`
...
         height: 480px;
+        ${
+            /**
+             * Set a default maximum height for the table of 250 rows (including the header row) so
+             * that clients don't accidentally create a table that tries to render too many rows at once.
+             * If needed, the max-height can be overridden by the client, but setting a default ensures
+             * that the max-height is considered if a larger one is needed rather than being overlooked.
+             */ ''
+        }
+        max-height: calc(250 * ${controlHeight});
```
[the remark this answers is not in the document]
rajsite: Stepping back a bit, the use case we are really trying to help here is compact displaying of relatively small tables that fit in the viewport.

We aren't currently helping super long tables that would go outside the viewport and rely on some other container for scrolling. To support them we should have a mode where the intersection observer is tracking the whole viewport so that can be efficiently virtualized.

I'd say lets cap it at like 40 (i.e. number of rows that fit on my desktop landscape viewport). Something like the icon-table can just set `max-height: none;` to let it grow and handle it's own consequences (it wouldn't want to be virtualized anyway since it's goal is letting ctrl+f page search happen).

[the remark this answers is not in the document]
mollykreis: I'd be fine capping it at 40 rows by default.

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
mollykreis: @rajsite, given that I couldn't find any standard way for components to be marked as auto resizing, do you have any thoughts on how we should move forward? Any objections to using a boolean attribute of `autoHeight` as in this PR?

rajsite: The thing I'm on the fence about is how the api could grow and be consistent across controls:

Consistent apis:
- table today: auto-height
- table and other controls (textarea, etc) tomorrow: auto-height, auto-width
  - auto-width on table would size so all columns show
  - auto-height and auto-width would have to be implemented on top of `field-sizing: content` which on it's own grows height and width.

Inconsistent apis:
- table today: auto-height
- table tomorrow: auto-height, auto-width
- Other controls: fit-to-content? auto-resize?
  - Rely on user setting bounds on height and width, just fill the space

Another thing that just "feels" wrong is having styling, aka sizing, be part of the attribute api. I've fought pretty hard against that (i.e. attributes for "small" / "medium"/ "large" for spinners or dialog). I think ideally I'd like to say textarea / rich-text are just text areas. So in your style for sizing set `field-sizing: content` like you would any other text field. If you swap a native text-field for ours you can size it with CSS, you don't need to impact sizing and layout with a combination of attributes and CSS. (talked offline to express this concept more)

packages/nimble-components/src/table/styles.ts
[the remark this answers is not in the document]
rajsite: Other bit from talking offline is maybe we do a similar thing as combobox / select and have a half height offset, i.e. 40.5 so clearer to confirm additional rows are there

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
rajsite: Played around with the latest changes in storybook and it feels excellent. Implementation looks good, only nit would be a preference for the simpler name `--ni-nimble-table-fit-rows-height` which is a nice analog to the [`fit-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content) property.

Docs pattern can be similar to how we document dialog sizing style. And maybe a story test (separate from the matrix) with a couple sizes of data to show the property is behaving correctly (something like 5, 10, 45 or something)

packages/nimble-components/src/theme-provider/design-tokens.ts
```diff
@@ -503,6 +503,12 @@ export const spinnerLargeHeight = DesignToken.create<string>(
     styleNameFromTokenName(tokenNames.spinnerLargeHeight)
 ).withDefault('64px');
 
+// The token gets a default value of empty-string but is a value in
+// the table styles.
+export const tableFitToRowsHeight = DesignToken.create<string>(
+    styleNameFromTokenName(tokenNames.tableFitToRowsHeight)
+).withDefault('');
```
rajsite, line 510: Maybe could set it to the table's default height (480px) just so it's a real value. Don't feel strongly.

packages/storybook/src/nimble/table/table.stories.ts
```diff
@@ -263,9 +265,19 @@ const setSelectedRecordIdsDescription = `A function that makes the rows associat
...
 
+const fitRowsHeightDescription = `Style the table's with \`height: $${tableFitRowsHeight.name};\` to make the table's height grow to fit all rows.
+
+Note: The table has a default maximum height that will render approximately 40 rows to avoid performance problems. Use caution when overriding the maximum height
+of the table because this may lead to performance issues.`;
+
 export const table: StoryObj<TableArgs> = {
     // prettier-ignore
     render: createUserSelectedThemeStory(html<TableArgs>`
+        <style>
```
rajsite, line 276: Can you add the class to hide the style in the code view output?

packages/storybook/src/nimble/table/table.stories.ts
```diff
@@ -494,6 +511,7 @@ export const table: StoryObj<TableArgs> = {
         validity: undefined,
         checkValidity: undefined,
         tableRef: undefined,
+        fitRowsHeight: true,
```
rajsite, line 514: I'm on the fence of making this default true. I think it's generally a special case usage and more tempted to default to false to better represent the out of box behavior. Could be convinced either way.

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: icon story needs to clear it's max height

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Add token to allow the table's height to be configured to the height of all table rows",
```
rajsite, line 3: @mollykreis gave another round of feedback but wasn't sure if PR still in work / intended to still be in draft.

packages/storybook/src/nimble/table/table.stories.ts
mollykreis: I can be convinced either way as well. I decided to default it to `true` because it made the docs page read better without reserving a bunch of extra space for the table when the default data set has only a few rows. But, I agree that the common case will be not to use the new token.

rajsite: I did find it harder to read the docs for the large data set case since you have to scroll past a full screen of rows at that point. So if we do go this way lets set a max-height that's smaller.

mollykreis: Went ahead and made it default to `false` to keep existing behavior default.

packages/storybook/src/docs/component-status.stories.ts
```diff
@@ -559,11 +560,15 @@ const metadata: Meta<TableArgs> = {
     },
     // prettier-ignore
     render: createUserSelectedThemeStory(html<TableArgs>`
+        <style class="code-hide">
+            ${tableTag} {
+                height: var(${tableFitRowsHeight.cssCustomProperty});
+                max-height: unset;
```
rajsite, line 566: unset isn't likely to be what you need: https://developer.mozilla.org/en-US/docs/Web/CSS/unset
The initial value of the non-inheritable max-height property is none: https://developer.mozilla.org/en-US/docs/Web/CSS/max-height#formal_definition

We should just set it to none if that's the intended usage

packages/nimble-components/src/table/styles.ts
```diff
@@ -25,6 +28,16 @@ export const styles = css`
...
+        ${tableFitRowsHeight.cssCustomProperty}: calc(var(--ni-private-table-scroll-height) + ${controlHeight});
+        ${
+            /**
+             * Set a default maximum height for the table of 40.5 rows plus the header row so
+             * that clients don't accidentally create a table that tries to render too many rows at once.
+             * If needed, the max-height can be overridden by the client, but setting a default ensures
+             * that the max-height is considered if a larger one is needed rather than being overlooked.
+             */ ''
+        }
+        max-height: calc(${controlHeight} + (40.5 * (2 * ${borderWidth} + ${controlHeight})));
```
mollykreis, line 40: Since we are unconditionally setting the `max-height` here regardless of whether or not the client is trying to use the "fit-rows" feature, do we need to consider this a breaking change or update apps to make sure full-page tables don't have their heights accidentally limited?

Thoughts? @rajsite @jattasNI

rajsite: From a [basic search](https://dev.azure.com/ni/_search?action=contents&text=%22nimble-table%20%7B%22&type=code&lp=custom-Collection&filters=&pageSize=25&result=DefaultCollection/DevCentral/srd-port/GBmaster//src/BBA/Research%20Demos/Nimble%20Demo/APM-Start/src/app/home/welcome.component.scss) looks like usages outside systemlinkshared were setting a max-height. Don't have a clear idea about what CSS changes we would consider breaking changes but I think it makes sense to lean towards caution.

So I think tasks would be:
-  Checking our usages in systemlinkshared
- Mark breaking. Do we want to mark `dependentChangeType` as breaking as well? Seems to be what we are advocating (table's have a breaking usage change so treat this upgrade more carefully, including your table usage in Angular / Blazor)
- Description should be more specific about the breaking change and mitigations. i.e., table now predefines a max-height so make sure to override if you need a specific one.

rajsite (approved):

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -74,6 +74,7 @@ export const comments: { readonly [key in TokenName]: string } = {
     spinnerSmallHeight: 'Small height (16px) for a spinner component',
     spinnerMediumHeight: 'Medium height (32px) for a spinner component',
     spinnerLargeHeight: 'Large height (64px) for a spinner component',
+    tableFitRowsHeight: 'The height of the table when all rows are visible',
```
jattasNI, line 77: nit: Since the common case for most of our tokens is to have the same value across an entire application, it could be helpful to highlight that this one is different.
```suggestion
    tableFitRowsHeight: 'The height of the table when all rows are visible. It is set automatically to the correct value within the scope of each table.',
```

packages/nimble-components/src/table/styles.ts
```diff
@@ -25,6 +28,16 @@ export const styles = css`
 
     :host {
         height: 480px;
+        ${tableFitRowsHeight.cssCustomProperty}: calc(var(--ni-private-table-scroll-height) + ${controlHeight});
```
jattasNI, line 31: I think it would be worthwhile to add a unit test that ensures the `tableFitRowsHeight` token gets updated after a call that changes the number of rows. The Chromatic tests do a good job of covering the render behavior and they indirectly cover `setData()`. But explicit unit tests would be valuable so that it's tested closer to the implementation. You could also think about whether we'd want separate tests for each the cases that could impact this (e.g. calling `setData()` vs `setRecordHierarchyOptions()`).

jattasNI (approved):

packages/nimble-components/src/table/styles.ts
mollykreis: Added some tests that made sure the table's height is correct when styled with that token when data is set, grouping is applied, and when rows are collapsed.

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "breaking",
```
rajsite, line 2:
```suggestion
  "type": "major",
```

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Create token representing the table's height with all rows visible; **Breaking change:** the table now specifies a max-height; if a different max-height is required, it needs to be configured.",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "breaking"
```
rajsite, line 6:
```suggestion
  "dependentChangeType": "major"
```

change/@ni-nimble-components-a5fd15e6-373b-4e3c-b562-fc2e440df26c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Create token representing the table's height with all rows visible; **Breaking change:** the table now specifies a max-height; if a different max-height is required, it needs to be configured.",
```
rajsite, line 3: @mollykreis this PR looks ready to merge, dunno if you are waiting on anything

mollykreis: I'm planning on updating the `max-height` of SLE clients first.

rajsite: Ahh, makes sense 👍

packages/nimble-components/src/table/styles.ts
mollykreis: Change has been marked as `major`, and SLE tables have been updated.

PR #2308, opened 2024-07-29
Updates to table page objects

mollykreis:
# Pull Request

## 🤨 Rationale

Update the table page object to include functionality needed by some clients. Specifically,

- Add `getHeaderTextContent` to the page object so that clients don't have to access the header element to get the text content themselves
- Export `TableColumnMappingPageObject` from Angular
- Extends `TablePageObject` that's exported from Angular to include a new `waitForDataUpdatesToRender` to help clients wait for data to be rendered in the table when they are using the `data$` property on the Angular table directive, which makes the timing of `setData()` outside of the client's control

## 👩‍💻 Implementation

Both the implementation of `getHeaderTextContent` and `waitForDataUpdatesToRender` were copied from the `sl-table` page object in SystemLinkShared.

## 🧪 Testing

- Updated a test in nimble-components to use the new nimble-components page object test
- Wrote new Angular tests to test the Angular page object

## ✅ Checklist


packages/angular-workspace/nimble-angular/table/testing/table.pageobject.ts
```diff
@@ -1,3 +1,35 @@
...
+export class TablePageObject<T extends TableRecord> extends NimbleComponentsTablePageObject<T> {
+    private mostRecentSetDataPromise?: Promise<void>;
+
+    public constructor(tableElement: Table<T>) {
+        super(tableElement);
+
+        // Cache the most recent promise returned from calls to setData() on the nimble table
+        // so that we can appropriately wait for them to complete to know when the Angular
+        // data$ Observable has been applied to the table.
+        const originalSetDataFn = tableElement.setData.bind(tableElement);
```
mollykreis, line 22: As mentioned in the PR description, this functionality was copied from the `sl-table` page object.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -80,6 +80,14 @@ export class TablePageObject<T extends TableRecord> {
         );
     }
 
+    public getHeaderTextContent(columnIndex: number): string {
```
mollykreis, line 83: As mentioned in the PR description, this function was copied by the `sl-table` page object.

mollykreis:
@m-akinc, can you review buddy this PR for me?

m-akinc (approved):

packages/angular-workspace/nimble-angular/table/testing/table.pageobject.ts
```diff
@@ -1,3 +1,35 @@
...
\ No newline at end of file
+/**
+ * The page object for the `nimble-table` component to provide consistent ways of querying
+ * and interacting with the component during tests.
+ *
+ * This Angular version of the page object extends the nimble-components version to add the ability
+ * to wait for data updates to be applied to the table since the timing isn't easily observable
+ * when using the `data$` property.
+ */
+export class TablePageObject<T extends TableRecord> extends NimbleComponentsTablePageObject<T> {
```
jattasNI, line 13: >🧪 Testing
>
>None. Both the implementation of getHeaderTextContent and waitForDataUpdatesToRender were copied from the sl-table page object in SystemLinkShared.

Even when a page object method isn't needed by Nimble component tests we've still tried to write dedicated specs to cover them. It can be pretty basic, just a single test that verifies the method is sane, but it is helpful to catch breakages in Nimble's tests rather than client tests. I think we should still strive for this for these two methods.

 [This file shows the pattern for the table in nimble-components](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/tests/table-pageobject.spec.ts) which you could copy for the nimble-angular page object and add a test to for the nimble-components page object.

rajsite (approved):

atmgrifter00 (approved):

packages/angular-workspace/nimble-angular/table/testing/table.pageobject.ts
mollykreis: Wrote tests and updated PR description.

rajsite: Based on Jesse's comment I think I'd also expect tests in the nimble-components `table-pageobject.spec.ts` for the new `getHeaderTextContent`. But I'd also be curious if that's our general policy now for page object additions (not something I've been looking for explicitly in review).

mollykreis: I updated an existing test to use the new page object function, so I thought that would be sufficient.

https://github.com/ni/nimble/pull/2308/commits/32ddbbfc20b33d9216ddd18fa160bc330b744adf

packages/angular-workspace/nimble-angular/table/testing/tests/table-pageobject.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { Component, ElementRef, ViewChild } from '@angular/core';
```
rajsite, line 1: Could we name this file `table.pageobject.spec.ts`. I think that better aligns with our conventions for `file.ts` and `tests/file.spec.ts`

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -195,14 +195,12 @@ describe('Table', () => {
             await element.setData(simpleTableData);
             await waitForUpdatesAsync();
 
-            let headerContent = pageObject.getHeaderContent(0)!.firstChild;
-            expect(headerContent?.textContent).toEqual('stringData');
+            expect(pageObject.getHeaderTextContent(0)).toEqual('stringData');
```
rajsite, line 198: If we do add tests for the components page object can we also move it from:
`table/tests/table-pageobject.spec.ts` to
`table/testing/tests/table.pageobject.spec.ts`
to align with file name and test placement patterns a bit better.

packages/angular-workspace/nimble-angular/table/testing/table.pageobject.ts
rajsite: (making up on the spot) I think since we are making pageobjects public apis they probably warrant their own explicit tests. Future refactoring could remove the current coverage of the pageobject but clients may still rely on the behavior if it's not removed from the public api explicitly. I think I could be convinced either way, I don't think we have page object apis yet that are only used externally but those absolutely need coverage.

jattasNI: To date I have only enforced that all page object methods should get called somewhere by some Nimble test, not that they get called by a dedicated page object test. For this PR I'd advocate to hold the same bar, which with the new tests I believe is now being met.

I think your proposal likely makes sense as a policy going forward. It seems like it'd be easy for us to remove the only usage of a page object method and not notice that it's no longer tested. But there are likely details to work out so we should split the discussion into a tech debt issue. I'll update the notes on #265 and mark it for triage.

jattasNI (approved):

packages/nimble-components/src/table/tests/table.spec.ts
mollykreis: Based on @jattasNI's guidance, I'm not planning on adding tests to the component's page object. Therefore, I'm not planning on renaming the file as part of this PR.

PR #2309, opened 2024-07-29
Update menu button column styles

### LINKED ISSUE #2247: Finalize styling of menu button table column
# 🐛 Bug Report

There are some changes that have been proposed to the styling of the menu button column that are not being addressed in the initial submission (#2204). Specifically, the following changes have been proposed by Brandon and Fred:

- Make the button the standard height (32px) rather than the slim control height
- ~~Change the left/right padding within the button to 8px rather than 16px~~ _Conceding that the 16px padding is sufficient given that there isn't a supported way to update the padding._
- ~~Change the font of the button to no longer be bold so that the text doesn't stand out from the rest of the table~~ _Leaving the font bold because this is the convention within nimble to show that text is interactable (such as in a ghost button)_

~~These changes, however, are not easily achieved using supported styling on the button.~~ We also need to evaluate how a 32px button looks with hover and focus states of the row/cell in the table.

We should also look into creating a focus theme matrix for the column to have chromatic testing for the focus state of the button within the table.

## 🤔 Expected Behavior

The goal is to make the menu button look more like a select within the table.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #2247

The only item from #2247 that we agreed should be changed is the height of the menu button. It will no longer be styled with a height of `controlSlimHeight` and instead have its default height. Additionally, @fredvisser wanted the row's focus-visible outline to be centered correctly when making this change since the misalignment of the outline is significantly more apparent when there is a 32px control in a cell. Therefore, I've also address that in this PR.

## 👩‍💻 Implementation

- Remove styling of the menu-button's height
- Update the focus-visible styling of a row so that the green outline now symmetrically outlines the row. This now matches the outline height of each cell. Previously the top border of the outline was 2px too high because the row borders aren't symmetrical -- specifically, a row has a 2px top border and a 0px bottom border in order to have the horizontal dividers between rows.

## 🧪 Testing

Manually tested

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-334dab4c-65a4-4b4a-9a92-56976aee8c66.json
+++ b/change/@ni-nimble-components-334dab4c-65a4-4b4a-9a92-56976aee8c66.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update menu-button column styles",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
+++ b/packages/nimble-components/src/table-column/menu-button/cell-view/styles.ts
@@ -1,5 +1,4 @@
 import { css } from '@microsoft/fast-element';
-import { controlSlimHeight } from '../../../theme-provider/design-tokens';
 
 export const styles = css`
     :host {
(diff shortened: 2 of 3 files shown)


mollykreis:
@msmithNI, can you buddy this PR for me?

fredvisser:
Apologies - I accidentally marked this as ready for review 🤦‍♂️

fredvisser (approved):
The visual updates look great - 👍

The slight collision between the row or cell keyboard focus styling and the button hover styling is a good compromise.

rajsite (approved):

msmithNI (approved):

jattasNI (approved):

PR #2310, opened 2024-07-30
Add calendar theme aware tokens

aagash-ni:
# Pull Request

## 🤨 Rationale

Add required theme-aware tokens according to the mockup

Figma Mockup: https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/118969/Schedule-Test-Plans

## 👩‍💻 Implementation

- Added calendar theme-aware tokens for themes with respect to the mockup.
- Added `calendar.mdx` as a component doc and storybook file.

## 🧪 Testing
New story for calendar is added

## ✅ Checklist


packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,29 @@ export const comments: { readonly [key in TokenName]: string } = {
     mentionFontFamily: 'Font family for mention views',
     mentionFontSize: 'Font size for mention views',
     mentionFontWeight: 'Font weight for mention views',
-    mentionFontLineHeight: 'Font line height for mention views'
+    mentionFontLineHeight: 'Font line height for mention views',
+    calendarStaticEventBackgroundColor:
```
rajsite, line 297: Can you give screenshots of static event, dynamic event, and transient event? I don't have intuition of those event types

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,29 @@ export const comments: { readonly [key in TokenName]: string } = {
...
+    calendarTransientEventFontColor: 'Font color for transient calendar events',
+    calendarStaticEventHoverColor:
+        'Color while hovering static calendar events',
+    calendarDynamicEventHoverColor:
+        'Color while hovering dynamic calendar events',
+    calendarTransientEventHoverColor:
+        'Color while hovering transient calendar events',
+    calendarResourceSelectionBackgroundColor:
+        'Background color while calendar resource is selected',
+    calendarBlockedStrokeColor:
```
rajsite, line 317: is stroke meaningfully different than border? if not can we use the term border. if so can you explain / provide screenshots?

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,29 @@ export const comments: { readonly [key in TokenName]: string } = {
...
+        'Background fill to indicate the occupied slots in calendar resource rows',
+    calendarEventResizeRingColor:
```
rajsite, line 319: It's not clear to me what visual this is associated with. Can you provide a screenshot?

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,29 @@ export const comments: { readonly [key in TokenName]: string } = {
...
+    calendarStaticEventFontColor: 'Font color for static calendar events',
+    calendarDynamicEventFontColor: 'Font color for dynamic calendar events',
+    calendarTransientEventFontColor: 'Font color for transient calendar events',
+    calendarStaticEventHoverColor:
+        'Color while hovering static calendar events',
+    calendarDynamicEventHoverColor:
+        'Color while hovering dynamic calendar events',
+    calendarTransientEventHoverColor:
+        'Color while hovering transient calendar events',
+    calendarResourceSelectionBackgroundColor:
```
rajsite, line 315: I don't know what part of a calendar is a "resource selection" can you provide a screenshot for this intended usage?

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -293,5 +293,29 @@ export const comments: { readonly [key in TokenName]: string } = {
...
+        'Border color for transient calendar events',
+    calendarStaticEventFontColor: 'Font color for static calendar events',
+    calendarDynamicEventFontColor: 'Font color for dynamic calendar events',
+    calendarTransientEventFontColor: 'Font color for transient calendar events',
+    calendarStaticEventHoverColor:
+        'Color while hovering static calendar events',
+    calendarDynamicEventHoverColor:
+        'Color while hovering dynamic calendar events',
+    calendarTransientEventHoverColor:
+        'Color while hovering transient calendar events',
```
rajsite, line 314: What part are these hover colors for? border? background?

packages/nimble-components/src/theme-provider/design-token-names.ts
```diff
@@ -239,7 +239,25 @@ export const tokenNames: { readonly [key in TokenName]: string } = {
     mentionFontFamily: 'mention-font-family',
     mentionFontSize: 'mention-font-size',
     mentionFontWeight: 'mention-font-weight',
-    mentionFontLineHeight: 'mention-font-line-height'
+    mentionFontLineHeight: 'mention-font-line-height',
```
jattasNI, line 242: Let's try to get consensus on naming. [These are the guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#naming), with the expected format being `[element]-[part]-[interaction_states]-[remaining_states]-[token_type]`. And here's a screenshot showing some of these states (not using these tokens yet).
image
 
I'd propose these values for "element":
1. `calendar` - everything except the cards
2. `calendar-event` - a card in any of its states
3. `grab-handle` - the circle that you can click on to resize a calendar event (I suggest making this a separate "element" since I'm guessing it'd be the same styling if we allowed resizing other components).
4. I'm not sure what "calendar-resource-selection" is so not sure if it's an element or a part. Could you clarify with a screenshot?

Then `calendar-event` could have "remaining_states" like "static", "dynamic", "transient". Maybe "blocked" too? Does that behave like an event (a rounded rectangle within a time slot) or is it a part of a calendar? (I'll admit I'm on the fence about putting this at the end. It feels clunky but seems to fit our guidance best)

We should add the "part" for the "hover" tokens (background, I assume?) and grab-handle/resize-ring.

For example, this would yield:
- `calendar-event-background-static-color`
- `calendar-event-background-dynamic-color`
- ...
- `calendar-event-border-static-color`
- ...
- `calendar-event-static-font-color`
- ...
- `calendar-event-background-hover-dynamic-color`
- ...
- `grab-handle-background-color`

Thoughts @rajsite?

rajsite: Oh, missed this before my other comments, but yea this is more inline with what I would expect for the naming 👍

aagash-ni: > Then calendar-event could have "remaining_states" like "static", "dynamic", "transient". Maybe "blocked" too? Does that behave like an event (a rounded rectangle within a time slot) or is it a part of a calendar?

These states are just a different appearance of the events as below:

Static Event:

image

These are static events that represent the actual/defined data (In our case test plans data scheduled against systems )

Dynamic Event:

image

This represents that the event can be dynamically created/resize ( In our case its the event created by the user to select the timeline for the current test plan to be scheduled)

Transient Event:

image

These don't have specific state, They will just act as a reflection of the Dynamic events.


I have updated all the tokens names as per guidance, Please let me know if anything need to be updated.

packages/nimble-components/src/theme-provider/design-token-comments.ts
aagash-ni: Added here: https://github.com/ni/nimble/pull/2310#discussion_r1698434728

packages/nimble-components/src/theme-provider/design-token-comments.ts
aagash-ni: These are strokes to represent the specific slot is busy/occupied based on the combined availability of the selected system with respect to its scheduled testplans.

image

packages/nimble-components/src/theme-provider/design-token-comments.ts
aagash-ni: image

packages/nimble-components/src/theme-provider/design-token-comments.ts
jattasNI: Given the fact that "stroke" is often used to mean "border", I'd vote to change this to something like "stripe". So `calendar-blocked-stripe-color` (open to improvements on that suggestion)

packages/nimble-components/src/theme-provider/design-token-comments.ts
jattasNI: Let's go with a less domain-specific term like "row" instead of "resource". So `calendar-row-background-color`.

packages/nimble-components/src/theme-provider/design-token-names.ts
jattasNI: Generally looks good to me. I left a couple more comments on the threads Milan had opened. And one last point is whether the grab handle should be `calendar-grab-handle` or just `grab-handle`. I originally proposed the latter but on second thought I think I prefer to start with it scoped to calendar since we have fairly low confidence that the design will apply generally.

jattasNI (approved):

packages/nimble-components/src/theme-provider/design-token-comments.ts
rajsite: I'd go for `calendar-fill-blocked-color` the specific fill color is independent from the fill shape / pattern (stripe today, could be rectangular, tomorrow).

Is there an argument that this is a calendar-event type? If so I'd alternatively say `calendar-event-fill-blocked-color`, but either of these options sound sane to me.

packages/nimble-components/src/theme-provider/design-token-comments.ts
rajsite: And because this is a `selected` state of a calendar-row's background, how about `calendar-row-background-selected-color`

packages/nimble-components/src/theme-provider/design-token-comments.ts
jattasNI: I agree on `fill` rather than `stripe`. I'm less clear on the choice between the 3 variants. To me it's a bit different than an event and doesn't fill a whole row so I'd probably vote `calendar-fill-blocked-color`, but defer to @aagash-ni since you know the use case better.

fredvisser (approved):

packages/nimble-components/src/theme-provider/design-token-comments.ts
aagash-ni: Updated it as calendarRowBackgroundSelectedColor - `calendar-row-background-selected-color`

packages/nimble-components/src/theme-provider/design-token-comments.ts
aagash-ni: Updated it as calendarEventFillBlockedColor - `calendar-event-fill-blocked-color`

PR #2314, opened 2024-07-31
Fix Select arrow down behavior with hidden selected option

atmgrifter00:
# Pull Request

## 🤨 Rationale

While doing the integration work of updating the SLE user selector to use dynamic options, a non-ideal behavior became apparent in the case where a user would provide some filter text, filtering out the current selected user (which as our guidance states, requires the option to be present but hidden), and then press `ArrowDown`, and nothing would get selected. This would happen when the selected option was _below_ the other options in DOM order.

The two solutions for creating the expected behavior were 1) update the integration code to always put the hidden option at the top of the DOM order, or 2) change the Nimble Select behavior to provide the expected behavior in this case. I opted with 2 because it feels like a gap that any user of the Nimble Select could possibly hit.

## 👩‍💻 Implementation

Now handling the `ArrowDown` case directly in the key handling code of the `Select`.

## 🧪 Testing

Added some unit tests. Manually checked behavior in Angular example app with dynamic options.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/select/index.ts
+++ b/packages/nimble-components/src/select/index.ts
@@ -567,15 +567,33 @@ export class Select
      */
     public override keydownHandler(e: KeyboardEvent): BooleanOrVoid {
         const initialSelectedIndex = this.selectedIndex;
-        super.keydownHandler(e);
         const key = e.key;
         if (e.ctrlKey || e.shiftKey) {
             return true;
         }
 
+        if (key !== keyArrowDown) {
+            super.keydownHandler(e);
+        }
+
         let currentActiveIndex = this.openActiveIndex ?? this.selectedIndex;
         let commitValueThenClose = false;
         switch (key) {
+            case keyArrowDown: {
+                const selectedOption = this.options[this.selectedIndex];
+                if (this.open && isListOption(selectedOption) && !isOptionOrGroupVisible(selectedOption)) {
+                    if (this.openActiveIndex === this.selectedIndex) {
+                        this.selectFirstOption();
+                    } else {
+                        this.selectNextOption();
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/select/index.ts
```diff
@@ -567,15 +567,38 @@ export class Select
...
         }
 
+        if (key !== keyArrowDown) {
+            super.keydownHandler(e);
+        }
+
         let currentActiveIndex = this.openActiveIndex ?? this.selectedIndex;
         let commitValueThenClose = false;
         switch (key) {
+            case keyArrowDown: {
```
rajsite, line 582: didn't trace the code path but just to confirm, no weird issues with the other interactions to consider? up, home, end?

rajsite (approved):

packages/nimble-components/src/select/index.ts
fredvisser: Looks like `up` is [supposed](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#down_arrow) to select the last item in the list, but we haven't been doing that to date. 

This change now selects the item above the hidden selected menu-item, if my hacking is correct…

https://github.com/user-attachments/assets/43f1bfd2-0466-4eb8-9dc6-09418621f2ba

atmgrifter00: Yep, needed to handle `ArrowUp` as well. Home and End have no issues as they don't rely on the current selection for their behavior.

fredvisser (approved):

jattasNI (approved):

PR #2315, opened 2024-07-31
Select/combobox built-in labels

msmithNI:
# Pull Request

## 🤨 Rationale

Implementation for #2183 . (Work item includes figuring out usage guidance, so won't close it with this PR yet)

## 👩‍💻 Implementation

See updated HLD in this PR.

## 🧪 Testing

- Local testing in Storybook
- Added autotests to cover the select's new (internal) labelContent property and that it sets aria-label (similar to those in `list-option-group.spec.ts`)
- Storybook: Added the 'height test' stories present in other controls that have labels
- Inspected Chromatic diffs. The changes are all expected:
   - Disabled controls now have disabled labels
   - If a `width` style is set on the control (as it is in the theme matrix tests), labels will wrap to multiple lines if needed. This matches existing control behavior (see text area / number field stories)
- Verified pre-existing accessibility violations for [select](https://nimble.ni.dev/storybook/?path=/story/components-select--select) / [combobox](https://nimble.ni.dev/storybook/?path=/story/components-combobox--underline-combobox) are resolved

## ✅ Checklist

   - Updated Storybook docs to document label support for select/combobox

### DIFF
--- a/change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
+++ b/change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "nimble-select and nimble-combobox: Built-in label support (via default slot)",
+  "packageName": "@ni/nimble-components",
+  "email": "20709258+msmithNI@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/combobox/index.ts
+++ b/packages/nimble-components/src/combobox/index.ts
@@ -838,6 +838,7 @@ const nimbleCombobox = Combobox.compose<ComboboxOptions>({
                 @keydown="${(x, c) => x.toggleButtonKeyDownHandler(c.event as KeyboardEvent)}"
                 class="dropdown-button"
                 part="button"
+                aria-hidden="true"
                 aria-haspopup="true"
                 aria-expanded="${x => x.open}"
                 tabindex="-1"
--- a/packages/nimble-components/src/combobox/index.ts
+++ b/packages/nimble-components/src/combobox/index.ts
@@ -882,8 +882,6 @@ const nimbleCombobox = Combobox.compose<ComboboxOptions>({
                 class="dropdown-button"
                 part="button"
                 aria-hidden="true"
-                aria-haspopup="true"
-                aria-expanded="${x => x.open}"
                 tabindex="-1"
             >
                 <${iconArrowExpanderDownTag}
--- a/packages/nimble-components/src/combobox/template.ts
+++ b/packages/nimble-components/src/combobox/template.ts
@@ -34,6 +34,9 @@ ComboboxOptions
         @focusout="${(x, c) => x.focusoutHandler(c.event as FocusEvent)}"
         @keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
     >
+        <label part="label" class="label">
+            <slot></slot>
+        </label>
         <div class="control" part="control" ${ref('controlWrapper')}>
             ${startSlotTemplate(context, definition)}
             <slot name="control">
--- a/packages/nimble-components/src/select/index.ts
+++ b/packages/nimble-components/src/select/index.ts
@@ -199,6 +199,23 @@ export class Select
         return !(this.multiple || typeof this.size === 'number');
     }
 
+    /** @internal */
+    public labelSlot!: HTMLSlotElement;
+
+    /** @internal */
+    @volatile
+    public get labelContent(): string {
+        if (!this.$fastController.isConnected) {
+            return '';
+        }
+
+        const nodes = this.labelSlot.assignedNodes();
+        return nodes
+            .filter(node => node.textContent?.trim() !== '')
+            .map(node => node.textContent?.trim())
+            .join(' ');
+    }
+
     private _value = '';
     private forcedPosition = false;
     private openActiveIndex?: number;
--- a/packages/nimble-components/src/select/specs/built-in-labels.md
+++ b/packages/nimble-components/src/select/specs/built-in-labels.md
@@ -39,16 +39,15 @@ Default Slot:
 Template updates:
 
 -   A new `label` element will be added at the root of the control template, which will reflect the default slotted content (minus the list options which are already handled elsewhere)
--   A new container `div` will be added at the same level, to contain the control part (`div[part="control"]`)
--   Update element host `display` mode so the label is laid out correctly. (Formerly `inline-flex`, which will move to the container `div`. Prototyped as `inline-block`, but `inline-flex` / `inline-grid` is probably more in line with what we want for our controls layouts going forward.)
-
-Currently our shared dropdown styling (`patterns/dropdown/styles.ts`) applies focus/error-visible styling directly on `::host` - most of those styles will instead target the container `div`.
+-   Update shared dropdown styling as needed for label layout (`flex-direction` to `column`, move fixed height from `::host` down to the `div[part="control"]`)
 
 Other considerations:
 
 -   The label will get disabled visually if the control is disabled.
 -   If the label is not provided, vertical space should not be reserved for it.
--   Behavior if label length exceeds control width (most likely we'll grow to fit like our other controls)
+-   Behavior if label length exceeds control width will match our other controls:
+    -   If a fixed `width` style is not set, the control grows to fit the entire label on 1 line.
+    -   If a fixed `width` style is set, the label will wrap if needed to fit within that width.
 
 ### Accessibility / ARIA
 
@@ -67,8 +66,7 @@ Other considerations:
 
 ### Impact to nimble-rich-text-mention-listbox
 
-`nimble-rich-text-mention-listbox` will not support this feature, meaning it will not contain a `label` element in its template. The current plan is to not address any ARIA label issues on it either, but we should note the gap in [Align rich-text mention listbox implementation behavior](https://github.com/ni/nimble/issues/1926).  
-However, since we have shared dropdown styling, its template may need minor updates (e.g. to add the same parent/container `div` as select/combobox are getting), but we'll ensure it has the same visual appearance as before.
+`nimble-rich-text-mention-listbox` will not support this feature, meaning it will not contain a `label` element in its template. The current plan is to not address any ARIA label issues on it either, but we should note the gap in [Align rich-text mention listbox implementation behavior](https://github.com/ni/nimble/issues/1926).
 
 ## Alternative Implementations / Designs
 
--- a/packages/nimble-components/src/select/template.ts
+++ b/packages/nimble-components/src/select/template.ts
@@ -58,6 +58,7 @@ SelectOptions
         aria-disabled="${x => x.ariaDisabled}"
         aria-expanded="${x => x.ariaExpanded}"
         aria-haspopup="${x => (x.collapsible ? 'listbox' : null)}"
+        aria-label="${x => x.labelContent}"
         aria-multiselectable="${x => x.ariaMultiSelectable}"
         ?open="${x => x.open}"
         role="combobox"
@@ -69,6 +70,9 @@ SelectOptions
         @keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
         @mousedown="${(x, c) => x.mousedownHandler(c.event as MouseEvent)}"
     >
+        <label part="label" class="label">
+            <slot ${ref('labelSlot')}></slot>
+        </label>
         ${when(x => x.collapsible, html<Select>`
             <div
                 class="control"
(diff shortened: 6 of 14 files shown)


packages/nimble-components/src/select/specs/built-in-labels.md
```diff
@@ -39,16 +39,15 @@ Default Slot:
 Template updates:
 
 -   A new `label` element will be added at the root of the control template, which will reflect the default slotted content (minus the list options which are already handled elsewhere)
--   A new container `div` will be added at the same level, to contain the control part (`div[part="control"]`)
--   Update element host `display` mode so the label is laid out correctly. (Formerly `inline-flex`, which will move to the container `div`. Prototyped as `inline-block`, but `inline-flex` / `inline-grid` is probably more in line with what we want for our controls layouts going forward.)
-
-Currently our shared dropdown styling (`patterns/dropdown/styles.ts`) applies focus/error-visible styling directly on `::host` - most of those styles will instead target the container `div`.
+-   Update shared dropdown styling as needed for label layout (`flex-direction` to `column`, move fixed height from `::host` down to the `div[part="control"]`)
```
msmithNI, line 42: After further implementation / style tweaks, I found that I didn't need to add a new container div after all, so I simplified the implementation and HLD here. There's also no changes to the rich text mention listbox template, with the updated approach.

Initial issues I hit when prototyping were (1) to get the `div.control` to still size correctly, (2) focus/ focus-visible styling, including getting the underlines to still center, but those issues are fixed now.

packages/storybook/src/nimble/combobox/combobox.stories.ts
```diff
@@ -113,14 +115,20 @@ const metadata: Meta<ComboboxArgs> = {
             appearance="${x => x.appearance}"
             value="${x => x.currentValue}"
             placeholder="${x => x.placeholder}"
-            style="width: 250px;"
+            style="min-width: 250px;"
```
msmithNI, line 118: Note: Switching from `width` to `min-width` in these stories so that the control will grow with a long label, instead of wrapping the label (matches more of our other control stories).

packages/nimble-components/src/combobox/index.ts
```diff
@@ -881,6 +881,7 @@ const nimbleCombobox = Combobox.compose<ComboboxOptions>({
                 @keydown="${(x, c) => x.toggleButtonKeyDownHandler(c.event as KeyboardEvent)}"
                 class="dropdown-button"
                 part="button"
+                aria-hidden="true"
```
msmithNI, line 884: From the HLD, this resolves an accessibility violation.

>  -   Elements must only use supported ARIA attributes: `nimble-combobox` `nimble-toggle-button`: ARIA attribute is not allowed: `aria-expanded="false"`
>     -   Can be resolved via `aria-hidden="true"` on the dropdown toggle button, as it's not keyboard-focusable.

msmithNI:
@atmgrifter00 can you buddy this PR?

packages/nimble-components/src/select/template.ts
```diff
@@ -58,6 +58,7 @@ SelectOptions
         aria-disabled="${x => x.ariaDisabled}"
         aria-expanded="${x => x.ariaExpanded}"
         aria-haspopup="${x => (x.collapsible ? 'listbox' : null)}"
+        aria-label="${x => x.labelContent}"
```
atmgrifter00, line 61: I would have expected the `Combobox` to need this as well.

packages/nimble-components/src/select/index.ts
```diff
@@ -199,6 +199,23 @@ export class Select
         return !(this.multiple || typeof this.size === 'number');
     }
 
+    /** @internal */
+    public labelSlot!: HTMLSlotElement;
+
+    /** @internal */
+    @volatile
+    public get labelContent(): string {
```
atmgrifter00, line 207: Why doesn't the `Combobox` need this same implementation and set of tests?

packages/nimble-components/src/select/template.ts
fredvisser, line 61: I don't think the select [requires](https://stackoverflow.com/a/56775032/13331204) an explicit `aria-label`, because assistive technology can already infer the label from the slotted content. If that's the case, we should remove the `aria-label`.

I.e. The label and aria-label are both displayed in VoiceOver 

Screenshot 2024-08-01 at 10 30 25 AM

msmithNI: Without the `aria-label` set on the select, the select still has an accessibility violation about not having an accessible label. I ran a build with that change, [this Storybook build without aria-label](https://60e89457a987cf003efc0a5b-zgzmmrogzr.chromatic.com/?path=/story/components-select--select) shows the violation.
As I noted above, the select and combobox declare their ARIA role in different places (also in the shadow root for the combobox, vs on the template/element itself for the combobox), which may be the reason for the different behavior.

Good catch that the current changes can result in duplicate narration though. Since `aria-label` is set, I probably need to set `aria-hidden` on the `label` element. (I see that in [list-option-group which sets aria-label](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/list-option-group/template.ts#L19) too).

msmithNI: Added `aria-hidden` to the `label` element, since `aria-label` is used.

atmgrifter00: There's still a bit of logic present to produce the `labelContent` string. I presume we've concluded that the results produced by that logic are unneeded for the `Combobox`?

atmgrifter00 (approved):

packages/nimble-components/src/select/index.ts
msmithNI: As discussed in the other comment threads, combobox doesn't need `aria-label` set to resolve its label accessibility violation, so it doesn't need `labelContent` (only used to set `aria-label`).

packages/nimble-components/src/select/template.ts
msmithNI: The logic of `labelContent` is just used to compute the value for `aria-label`, whereas combobox will just use the label text directly for that.

In terms of the computed `labelContent` value vs. what the label shows visually, I checked that both handle plain text and  `span` elements. One minor difference I see is that `<span>A B C</span><span>D E F</span>` shows `A B CD E F` visually in the label (vs `labelContent` of `A B C D E F`). But the way the label is shown matches how other existing controls (number field/ text field/etc) show it, so I think it's OK.

change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Interesting that hovering over the label causes the control to enter the hover state and that clicking it behaves the same as clicking the control. This was slightly unexpected to me (I thought maybe I could select and copy the label text, but that's not possible). However it matches the behavior of labels on text and number field so I think it's the right call.

No action needed, just observing in case someone else sees this and finds it unexpected like I did.

https://github.com/user-attachments/assets/63e44712-1c99-44f1-9d22-c2013139e6b0

packages/nimble-components/src/select/index.ts
```diff
@@ -199,6 +199,23 @@ export class Select
...
+        if (!this.$fastController.isConnected) {
+            return '';
+        }
+
+        const nodes = this.labelSlot.assignedNodes();
+        return nodes
+            .filter(node => node.textContent?.trim() !== '')
+            .map(node => node.textContent?.trim())
+            .join(' ');
```
rajsite, line 216: We have this logic duplicated slightly differently a couple times. Made a tech debt to clean it up. Could help to add this example to the issue so it's found easier after merge: https://github.com/ni/nimble/issues/2322

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -154,6 +157,11 @@ const metadata: Meta<SelectArgs> = {
         </${selectTag}>
     `),
     argTypes: {
+        label: {
+            name: 'default',
+            description: `${slottedLabelDescription({ componentName: 'select' })}`,
```
jattasNI, line 162: Reminder to work with @fredvisser to figure out usage guidance on when clients should use label vs placeholder vs neither. Doesn't have to block this PR but it'd be good to have a decision before closing the feature.

packages/nimble-components/src/combobox/template.ts
```diff
@@ -34,6 +34,9 @@ ComboboxOptions
         @focusout="${(x, c) => x.focusoutHandler(c.event as FocusEvent)}"
         @keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
     >
+        <label part="label" class="label">
```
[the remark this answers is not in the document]
rajsite: Same issue for select

jattasNI (commented):
Just reviewed storybook docs and behavior, still need to look at the implementation.

packages/nimble-components/src/combobox/template.ts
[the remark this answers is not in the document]
rajsite: Looks like we aren't consistent across controls. Switch label is selectable but text area and text field labels aren't. Maybe create a bug to make consistent.

rajsite (approved):

packages/storybook/src/nimble/combobox/combobox.stories.ts
msmithNI: The theme matrix stories have wrapping labels because they set a fixed `width` on the control. This matches the other stories for controls with labels (and I think a fixed width makes sense for the theme matrix so everything stays in a grid).

rajsite: Think it's expected in matrix tests so the controls layout nicely. But for the doc stories it's allowed to grow

change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
rajsite: Looks like in chrome and firefox a labelled select gets focused when the label is clicked (instead of open) but hover state on the label does apply to control and label text is selectable in control: https://jsbin.com/viveyitene/2/edit?html,js,output

Made a separate comment about selection of label, hover seems aligned, don't feel strongly about click opening vs just focusing

packages/nimble-components/src/combobox/template.ts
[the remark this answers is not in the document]
rajsite: Based on the other comment we should have the issue align our controls on mouse selection, hover interactions, and click activation.

change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
rajsite, line 1: Looks like the Angular and Blazor example apps should be updated so the selects and comboboxes have labels. Feel free to do that in this PR or a follow-up

packages/nimble-components/src/combobox/template.ts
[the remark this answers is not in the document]
rajsite: That issue should also probably capture some docs on labelled control micro interactions to validate are consistent.

jattasNI (approved):

mollykreis (approved):

packages/nimble-components/src/combobox/index.ts
```diff
@@ -881,6 +881,7 @@ const nimbleCombobox = Combobox.compose<ComboboxOptions>({
...
                 aria-haspopup="true"
                 aria-expanded="${x => x.open}"
```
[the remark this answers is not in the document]
msmithNI: Removed aria props

change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
msmithNI: Updated Angular/Blazor examples with select/combobox labels.

packages/nimble-components/src/combobox/template.ts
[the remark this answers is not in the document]
msmithNI: I attempted to capture all of this discussion / document the inconsistencies in #2331 .

I considered moving the `user-select: none` style, but left it as-is for this PR, since the current behavior matches more of the other labeled controls, and it's currently a shared dropdown style. We can revisit as part of that issue, as needed.

change/@ni-nimble-components-18b7e8bd-7ece-49a0-a662-2a8424f6bfdd.json
msmithNI: Documented the current inconsistencies in #2331

packages/nimble-components/src/select/index.ts
msmithNI: Done

PR #2321, opened 2024-08-01
Exclude Apache Arrow and eslint from Renovate major updates

jattasNI:
# Pull Request

## 🤨 Rationale

This Renovate PR #2318 is failing builds because it tries to update eslint to 9.x but javascript-styleguide is only compatible with 8.x.

We also don't want Renovate to do major updates of apache arrow in either npm or nuget because it's pinned and needs to stay in sync between npm and nuget so should probably be manually handled.

## 👩‍💻 Implementation

Update Renovate config to exclude those dependencies. I got the names of the dependencies from `packages.json` for npm and `packages.lock.json` for nuget.

## 🧪 Testing

After this goes in we can retrigger the Renovate algorithm and see if the major bump excludes these.

## ✅ Checklist

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -27,7 +27,9 @@
       "matchDepTypes": ["dependencies", "devDependencies", "peerDependencies"],
       "excludePackagePatterns":[
         "^@angular",
+        "apache-arrow",
         "comlink",
+        "eslint",
         "ng-packagr",
         "playwright",
         "remark-gfm",
@@ -41,6 +43,7 @@
       "groupName": "nuget dependencies",
       "matchManagers": ["nuget"],
       "excludePackagePatterns":[
+        "Apache.Arrow",
         "NI.CSharp.Analyzers",
         "Microsoft.Playwright"
       ],


.github/renovate.json
```diff
@@ -27,7 +27,9 @@
       "matchDepTypes": ["dependencies", "devDependencies", "peerDependencies"],
       "excludePackagePatterns":[
         "^@angular",
```
jattasNI, line 29: Do we already have another process for manually updating arrow? If not, think it'd make sense to do it manually as part of https://github.com/ni/nimble/issues/1747 alongside the pinned playwright deps? If so I can update those instructions.

rajsite: Nuget doesn't have a peer dependency concept as far as I'm aware so maybe it's possible for the nimble-blazor and blazor app arrow versions to be out of sync? Not sure how assemblies resolve, can multiple different versions resolve in the same C# app? Either way I wouldn't expect any of that from keeping oour dependency up to date so yea lets have a manual task

rajsite (approved):

.github/renovate.json
jattasNI: Updated.

PR #2324, opened 2024-08-02
Remove icon-check and rename icon-check-large to icon-check

mollykreis:
# Pull Request

## 🤨 Rationale

There have been two "check" icons in nimble -- `nimble-icon-check` and `nimble-icon-check-large`. According to @NIbokeefe, the only icon we should be using (both within nimble and our clients) is `nimble-icon-check-large`. However, all client usages I could find and most nimble usages were `nimble-icon-check`. Since there are no valid use cases for the smaller icon, I've deleted it and renamed "check-large" to "check".

Note: As a bit of background, the smaller icon was originally added for use within the `nimble-checkbox` component. However, the visual design and implementation was updated at some point to use the larger icon. When this occurred, there were no longer any places that needed the smaller icon.

## 👩‍💻 Implementation

- Deleted the SVG for the old "check" icon
- Renamed the SVG for the old "check-large" icon to "check"
- Updated the few instances of "check-large" in nimble to "check"

## 🧪 Testing

- Verified there are no usages of "check-large" or "checklarge" in SLE
- Verified there are no unintentional chromatic diffs. Specifically, the `nimble-checkbox` visuals are not affected by this change.
- Verified that instances of `nimble-icon-check` in the Angular and Blazor example apps now correctly render the larger check icon

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-tokens-9e9f7223-3c56-4a3e-a547-a00214a35d34.json
+++ b/change/@ni-nimble-tokens-9e9f7223-3c56-4a3e-a547-a00214a35d34.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Remove icon-check and rename icon-check-large to icon-check",
+  "packageName": "@ni/nimble-tokens",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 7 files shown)


change/@ni-nimble-tokens-9e9f7223-3c56-4a3e-a547-a00214a35d34.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Remove icon-check and rename icon-check-large to icon-check",
+  "packageName": "@ni/nimble-tokens",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
```
rajsite, line 6: Guess this should be dependentChangeType major since it'll result in a breaking component removal in all frameworks

```suggestion
  "dependentChangeType": "major"
```

rajsite (approved):

jattasNI (approved):
I dug up some history and #2156 was the PR that updated the checkbox to use the larger icon. Looks like we're heading in a consistent direction, just commenting to establish a link in case we're looking through history down the road.

PR #2330, opened 2024-08-02
Update menu button column docs

mollykreis:
# Pull Request

## 🤨 Rationale

Update the storybook example and docs for the menu-button table column to include:
- A better example of the expected use case that shows the selected item with a check mark next to the selected menu item
- Documentation about using the menu-button column vs the action menu

Resolves the comments from https://github.com/ni/nimble/pull/2204#discussion_r1670905005

## 👩‍💻 Implementation

Update storybook story

## 🧪 Testing

Verified the updated story works as expected

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
+++ b/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
@@ -31,8 +31,8 @@ responsibility to do the following:
    which it was opened. This should be done by listening to the
    `menu-button-column-beforetoggle` event and updating the menu items when the
    `newState` is `true`.
-1. Ensure the currently selected value for the open menu exists as an item in the menu
-   and is marked as selected with the <Tag name={iconCheckTag}/>.
+1. Ensure the currently selected value for the open menu exists as an item in
+   the menu and is marked as selected with the <Tag name={iconCheckTag}/>.
 1. Update the data through a `setData()` call when a menu item is activated in
    order to update the text on the menu button in the column.
 
--- a/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
+++ b/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
@@ -3,6 +3,7 @@ import * as menuButtonColumnStories from './table-column-menu-button.stories';
 import ComponentApisLink from '../../../docs/component-apis-link.mdx';
 import TableColumnApisLink from '../../../docs/table-column-apis-link.mdx';
 import { tableColumnMenuButtonTag } from '../../../../../nimble-components/src/table-column/menu-button';
+import { iconCheckTag } from '../../../../../nimble-components/src/icons/all-icons';
 
 <Meta of={menuButtonColumnStories} />
 <Title of={menuButtonColumnStories} />
@@ -32,3 +33,17 @@ responsibility to do the following:
    `newState` is `true`.
 2. Update the data through a `setData()` call when a menu item is activated in
    order to update the text on the menu button in the column.
+
+### Menu Button Column vs Table Action Menu
+
+Both an action menu on a column and the <Tag name={tableColumnMenuButtonTag}/> provide ways to provide a menu associated with a row in the table. However, they are used in different contexts.
+
+An action menu should be used in the table when:
+-  the items in the menu are actions to be performed on the record
+-  selecting an item in the menu does not change the value of the record
+
+The <Tag name={tableColumnMenuButtonTag}/> should be used when:
+-  the items in the menu represent values that can be assigned to the record
+-  the currently selected value should be displayed in the table
+
+When using the <Tag name={tableColumnMenuButtonTag}/>, the currently selected value should also exist as an item in the menu and be marked as selected with the <Tag name={iconCheckTag}/>.
--- a/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
+++ b/packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
@@ -31,7 +31,9 @@ responsibility to do the following:
    which it was opened. This should be done by listening to the
    `menu-button-column-beforetoggle` event and updating the menu items when the
    `newState` is `true`.
-2. Update the data through a `setData()` call when a menu item is activated in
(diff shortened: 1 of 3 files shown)


mollykreis:

m-akinc (approved):

rajsite (approved):

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -316,34 +322,28 @@ export const menuButtonColumn: StoryObj<MenuButtonColumnTableArgs> = {
...
-                const item2 = document.createElement(menuItemTag);
-                item2.textContent = `[${recordId}] Green`;
-                item2.addEventListener('change', () => changeFavoriteColor('Green'));
-
-                const item3 = document.createElement(menuItemTag);
-                item3.textContent = `[${recordId}] Purple`;
-                item3.addEventListener('change', () => changeFavoriteColor('Purple'));
-
-                storyArgs.menuRef.replaceChildren(item1, item2, item3);
+                const menuItems: MenuItem[] = [];
```
rajsite, line 334: Can't we have this be more template driven? i.e. keep the state on the argsTable (like args.menuCurrentSelectedOption) and have the template use normal template patterns for managing this state?

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
```diff
@@ -3,6 +3,7 @@ import * as menuButtonColumnStories from './table-column-menu-button.stories';
 import ComponentApisLink from '../../../docs/component-apis-link.mdx';
```
jattasNI, line 3: This is unrelated to this PR but while testing it I noticed a bit of clipping at the left edge of the menu button text (only for "Yellow" since the Y seems to go further left than other letters). I only see this in Firefox, not Edge or Safari.

Happy to file a bug if you / @fredvisser think it's worth fixing but don't want to just knock it out.

image

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
```diff
@@ -32,3 +33,23 @@ responsibility to do the following:
...
+
+-   the items in the menu are actions to be performed on the record
+-   selecting an item in the menu does not change the value of the record
+
+The <Tag name={tableColumnMenuButtonTag}/> should be used when:
+
+-   the items in the menu represent values that can be assigned to the record
+-   the currently selected value should be displayed in the table
+
+When using the <Tag name={tableColumnMenuButtonTag}/>, the currently selected
```
jattasNI, line 53: I think this note would fit better as a bullet under the list above ("it is the client's responsibility to do the following")

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
```diff
@@ -32,3 +33,23 @@ responsibility to do the following:
    `newState` is `true`.
 2. Update the data through a `setData()` call when a menu item is activated in
    order to update the text on the menu button in the column.
+
+### Menu Button Column vs Table Action Menu
```
jattasNI, line 37: It would be good to link to this guidance from our docs on the table action menu. Unfortunately it looks like we don't actually have docs on the table action menu? From what I can find we document the slots / attributes for it but don't have a specific example for it.

The best place I can find would be a link from the table story's "Action menus" slot.

jattasNI (approved):

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.mdx
mollykreis: I can't see that behavior on Firefox (on Windows), so I'm wondering if it's a Firefox on Mac issue only. Since I can't reproduce it, I don't plan to knock it out as part of this PR. Filing a bug probably makes the most sense.

jattasNI: #2332

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
rajsite: Ah yea, would need a way to mark the properties observable. Think the easy way would be to make a small model class that is assigned to the args table with observable props. Could spend a sec on it but if not straightforward current behavior already an improvement 👍

rajsite: Yessss, it's beautiful, ty :)

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -290,6 +323,16 @@ export const menuButtonColumn: StoryObj<MenuButtonColumnTableArgs> = {
...
+        updateFavoriteColor: {
+            table: {
+                disable: true
+            }
+        },
+        currentRecord: {
+            table: {
+                disable: true
+            }
+        },
```
rajsite, line 335: @mollykreis  ah dang, sorry looked closer at the storybook and we need to disable menuState in the table as well
image

packages/storybook/src/nimble/table-column/menu-button/table-column-menu-button.stories.ts
```diff
@@ -310,40 +353,29 @@ export const menuButtonColumn: StoryObj<MenuButtonColumnTableArgs> = {
...
-                item2.textContent = `[${recordId}] Green`;
-                item2.addEventListener('change', () => changeFavoriteColor('Green'));
-
-                const item3 = document.createElement(menuItemTag);
-                item3.textContent = `[${recordId}] Purple`;
-                item3.addEventListener('change', () => changeFavoriteColor('Purple'));
-
-                storyArgs.menuRef.replaceChildren(item1, item2, item3);
+                const currentColor = storyArgs.currentRecord
+                    .favoriteColor as string;
```
rajsite, line 368: We can also create a tpye for the record data to avoid the as cast right?

```
        interface SimpleRecord extends TableRecord {
          id: string,
          firstName: string,
          lastName: string,
          favoriteColor: string;
        }
```

Issue #2322, opened 2024-08-01
Coalesce pattern for slot text content


rajsite:
# 🧹 Tech Debt

We have a couple patterns for serializing the textContent of a slot reference. We should make a helper for it so it's done consistently and efficiently (avoid unnecessary duplicate calls to textContent)

Examples:

https://github.com/ni/nimble/blob/186667b0dfe36930a9bce3efd711931799d12f84/packages/nimble-components/src/list-option/index.ts#L61-L64

https://github.com/ni/nimble/blob/186667b0dfe36930a9bce3efd711931799d12f84/packages/nimble-components/src/list-option-group/index.ts#L88-L92

https://github.com/ni/nimble/blob/186667b0dfe36930a9bce3efd711931799d12f84/packages/nimble-components/src/table-column/base/index.ts#L56-L59

https://github.com/ni/nimble/blob/5068a59f17f70fe822c3c61026d611c6f548c507/packages/nimble-components/src/select/index.ts#L212-L216

PR #2333, opened 2024-08-05
Assigned nodes text content

rajsite:
# Pull Request

## 🤨 Rationale

Fixes https://github.com/ni/nimble/issues/2322 by creating a shared helper for slot text content.

## 👩‍💻 Implementation

- Create a shared utility and use it

## 🧪 Testing

- Relying on chromatic and CI tests
- Found story changes to icon story from the checkmark change. Expect that the rename of the check icon didn't correctly retrigger a build of the story because our [externals ](https://github.com/ni/nimble/blob/a62f247447a1476a7825c905f3f861a3a21b03ba/.github/workflows/main.yml#L94-L97) aren't configured correctly. Re-triggered all the stories so we can approve new baselines.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/utilities/models/slot-text-content.ts
+++ b/packages/nimble-components/src/utilities/models/slot-text-content.ts
@@ -1,3 +1,6 @@
+/**
+ * Finds all text content within a slot and returns it as a space-delimited string.
+ */
 export const slotTextContent = (slot: HTMLSlotElement): string => {
     return slot
         .assignedNodes()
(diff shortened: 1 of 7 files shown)


packages/nimble-components/src/utilities/models/slot-text-content.ts
```diff
@@ -0,0 +1,7 @@
+export const slotTextContent = (slot: HTMLSlotElement): string => {
```
jattasNI, line 1: Could use a quick comment. For example:
```suggestion
/**
 * Finds all text content within a slot and returns it as a space-delimited string.
 */
export const slotTextContent = (slot: HTMLSlotElement): string => {
```

jattasNI (approved):

PR #2335, opened 2024-08-07
waitForEvent utility refactor

rajsite:
# Pull Request

## 🤨 Rationale

The previous `createEventListener` helper in nimble testing utilities leveraged a jasmine library feature which indirectly made jasmine a dependency of nimble as several of the page objects had a dependency on it and are part of the public api. This refactor breaks the jasmine dependency in the public api.

## 👩‍💻 Implementation

- The `waitForEvent` helper was updated to take an optional event handler callback. This deprecates the need for the `createEventListener` utility which was removed.
- The tests were refactored to create their own jasmine spies instead of relying on a utility to create them. This had several benefits:
  - Many tests were using the utility that created spies unnecessarily. They should have been using waitForEvent and now do so.
  - Several tests where trying to track ordering of events. This is very natural if the same spy is used in multiple event listeners which you can now do since the spy is provided.
  - Several tests were relying on the idea that the helpers can trigger events multiple times. This is not the case, they are written to only resolve once and to remove their event handlers after the first call
  - Now that the spy is constructed in the test, it is easier to fulfill it's generic and remove a bunch of `as` casting

## 🧪 Testing

Relying on existing test runs.

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
+++ b/packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
@@ -11,7 +11,7 @@ import {
     processUpdates,
     waitForUpdatesAsync
 } from '../../testing/async-helpers';
-import { createEventListener } from '../../utilities/testing/component';
+import { waitForEvent } from '../../utilities/testing/component';
 import { MenuButtonPageObject } from '../testing/menu-button.pageobject';
 
 class TestSlottedElement extends FoundationElement {}
@@ -30,39 +30,6 @@ async function setup(): Promise<Fixture<MenuButton>> {
 
 async function slottedSetup(): Promise<Fixture<TestSlottedElement>> {
     return fixture(composedTestSlottedElement());
-}
-
-/** A helper function to abstract adding a `beforetoggle` event listener, spying
- * on the event being called, and removing the event listener. The returned promise
- * should be resolved prior to completing a test.
- *
- * The function asserts that the menu button has the expected `open` value when the
- * `beforetoggle` is fired and that when the `beforetoggle` event is fired, the
- * `toggleSpy` has not been called.
- */
-function createBeforeToggleListener(
-    menuButton: MenuButton,
-    expectedOpenState: boolean,
-    toggleSpy: jasmine.Spy
-): {
-        promise: Promise<void>,
-        spy: jasmine.Spy
-    } {
-    const spy = jasmine.createSpy();
-    return {
-        promise: new Promise(resolve => {
-            const handler = (...args: unknown[]): void => {
-                expect(menuButton.open).toEqual(expectedOpenState);
-                expect(toggleSpy).not.toHaveBeenCalled();
-
-                menuButton.removeEventListener('beforetoggle', handler);
-                spy(...args);
-                resolve();
-            };
-            menuButton.addEventListener('beforetoggle', handler);
-        }),
-        spy
-    };
 }
 
 describe('MenuButton', () => {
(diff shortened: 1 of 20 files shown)


packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
```diff
@@ -256,86 +227,92 @@ describe('MenuButton', () => {
...
                 oldState: true
             };
-
-            pageObject.clickMenuButton();
-            await beforeToggleListener.promise;
-            expect(beforeToggleListener.spy).toHaveBeenCalledTimes(1);
-            const event = beforeToggleListener.spy.calls.first()
-                .args[0] as CustomEvent;
-            expect(event.detail).toEqual(expectedDetails);
-            beforeToggleListener.spy.calls.reset();
```
rajsite, line 334: This isn't a useful way to track the spy as createEventListener only fired the event once. the spy call count wouldn't have ever gone up

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -1477,16 +1480,16 @@ describe('Select', () => {
...
+                );
                 await pageObject.openAndSetFilterText('o');
                 await pageObject.clickAway();
                 expect(filterInputEvent).toHaveBeenCalledTimes(2);
                 expect(
-                    (
-                        (filterInputEvent.calls.argsFor(1)[0] as CustomEvent)
-                            .detail as SelectFilterInputEventDetail
-                    ).filterText
+                    filterInputEvent.calls.argsFor(1)[0].detail.filterText
```
rajsite, line 1492: This is much better. We lie to addEventListener about the handler to tradeoff for the spy getting richer type info. We may be able to create a feature request in jasmine to make createSpy easier to use with addEventListener but not going down that path for now.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
```diff
@@ -256,86 +227,92 @@ describe('MenuButton', () => {
...
-                .args[0] as CustomEvent;
+            const event = spy.calls.first().args[0];
             expect(event.detail).toEqual(expectedDetails);
         });
 
         it("should fire 'beforetoggle' event before the menu opens", async () => {
             await connect();
-            const toggleListener = createEventListener(element, 'toggle');
-            const beforeToggleListener = createBeforeToggleListener(
+            const spy = jasmine.createSpy<MenuButtonToggleEventHandler>();
```
rajsite, line 261: This ends up working nice since both events return the same type. Otherwise a shared spy would need something more generic like or some union type, etc.

```
type CustomEventHandler = (
    evt: CustomEvent
) => void;
```

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
rajsite: Was on the fence about renaming `waitForEvent` to something like `waitForSingleEvent`.
I'm thinking `createEventListener` is more misleading to believing it would potentially trigger multiple times / continuously compared to `waitForEvent` so didn't make that change.

packages/nimble-components/src/utilities/testing/component.ts
```diff
@@ -46,44 +46,22 @@ export async function sendKeyDownEvents(
...
 ): Promise<void> {
     return new Promise(resolve => {
-        const handler = (): void => {
-            element.removeEventListener(eventName, handler);
+        const handler = ((evt: T): void => {
+            callback?.(evt);
             resolve();
-        };
-        element.addEventListener(eventName, handler);
+        }) as EventListener;
```
rajsite, line 58: Here we can encapsulate the lie to addEventListener about the typing on the callback so waitForEvent is a little cleaner than using addEventListener directly.

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -1117,23 +1121,24 @@ describe('Table Interactive Column Sizing', () => {
...
             element,
-            'column-configuration-change'
+            'column-configuration-change',
+            spy
         );
         pageObject.dragSizeColumnByRightDivider(2, [1, 1, 1, 1]);
         await waitForUpdatesAsync();
-
-        expect(listener.spy).toHaveBeenCalledTimes(1);
+        await listener;
```
rajsite, line 1132: Some examples like this could have been racy as they were not explicitly waiting on the listener. This becomes a lot more evident as the waitForEvent returns the promise directly and we have linting for avoiding floating promises.

packages/nimble-components/src/menu-button/tests/menu-button.spec.ts
jattasNI: The name `waitForEvent` seems good enough to me. It returns a Promise, not an Observable or callback or something where there would be an expectation that it'd fire multiple times.

jattasNI (approved):

Issue #2304, opened 2024-07-23
Allow hiding the dialog backdrop on a NimbleDialog


epetersoni:
# 🙋 Feature Request

Provide a way for a `NimbleDialog` to hide the drop shadow.

## 😯 Problem to Solve

The drop shadow causes some issues with the way we're using a popup `NimbleDialog` with `InstrumentStudio`.

We're creating a modal WPF window that uses the same size as the WPF application and fill itself with a `WebView2` that is just showing the `NimbleDialog`. This allows Nimble to center the dialog content within the modal window.

* The drop shadow is inconsistent with other `InstrumentStudio` modal dialogs.
* There are issues if the main WPF window is partly offscreen. The modal window gets positioned to be entirely on screen, which makes the drop shadow visible over parts of whatever is behind the `InstrumentStudio` window. It seems like any solution to this would get complex.

## 💁 Proposed Solution

Provide a new property on the `NimbleDialog` that allows the drop shadow to be hidden.

The property could be a `bool` that allows the shadow to be shown or hidden.
Or maybe a floating point value allowing the opacity of the shadow to be set.

jattasNI:
Here's a screenshot illustrating the issue about the window being partly offscreen. The second webview in which the dialog is presented is not partly offscreen, so the overlay isn't positioned correctly. The suggestion from designers is that the dialog positioning would be ok if the overlay weren't visible.

<img width="1229" alt="Screenshot 2024-07-29 at 2 43 50 PM" src="https://github.com/user-attachments/assets/6ffe5406-8be1-4a94-947f-4406ce35c549">

jattasNI:
Assigned to @rajsite to document his concerns with exposing a first class API for this but also to see if we can find a workaround for this narrow use case.

searly:
@jattasNI @rajsite Should we be able to workaround this after the linked token change now? Can we get an updated version of the NimbleBlazor nuget published to try it out?

rajsite:
Sorry about that, I meant to get back to y'all last week 😅. The summary is:

- We aren't sure about adding this configurability as a supported feature yet.
  - A big part of supporting a new configuration is having crisp recommendations and usage guidelines
  - We wouldn't recommend disabling the backdrop for any browser running nimble apps or desktop running nimble apps using just blazor for the UI.
  - We are also not sure if it makes sense as a general feature we would recommend for all hybrid apps either yet. Some considerations:
[...]

PR #2339, opened 2024-08-08
Switch backdrop to using design tokens

### LINKED ISSUE #1835: Remove static design tokens when backdrop inheritance available
## 🧹 Tech Debt

We have a concept of static tokens used in the dialog and drawer because `::backdrop` pseudo-elements previously did not inherit from their owning element. This is fixed in spec in https://github.com/whatwg/fullscreen/issues/124 and is starting to [release in browsers](https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element).

As of March 2024, this is released in all browsers. That means we should be able to remove the `design-tokens-static` workarounds by Sept/Oct 2024:

https://github.com/ni/nimble/blob/f041266ec1ab47fff4e4d1908aaf1c40dade14de/packages/nimble-components/src/theme-provider/design-tokens-static.ts#L7-L18

rajsite:
# Pull Request

## 🤨 Rationale

Fixes #1835 by using the existing `modalBackdropColor` for dialogs and drawers now that CSS Custom Properties can participate in the DOM cascade.

The related goal is to provide a workaround for #2304 to allow temporarily overriding the backdrop color via CSS. For example:

```html
<nimble-dialog style="--ni-nimble-modal-backdrop-color: transparent;">
```

## 👩‍💻 Implementation

- Switch to the design token
- Drop the static token constants file
- I did not update the storybook docs to describe the workaround case or add testing specific to it. I don't think we would describe overriding the backdrop color as an explicitly supported configuration.

## 🧪 Testing

Relying on chromatic. Note, this actually fails to render the backdrop on Chromatic. The behavior change was introduced in [Chrome 122](https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element) but Chromatic infrastructure is on [Chrome 121](https://www.chromatic.com/docs/infrastructure-release-notes/#version-6).

## ✅ Checklist


packages/nimble-components/src/drawer/styles.ts
```diff
@@ -153,33 +149,4 @@ export const styles = css`
...
-`.withBehaviors(
-    /*
-     * We cannot use the modalBackdropColor token directly because the backdrop
-     * element is not a descendant of the nimble-theme-provider element.
-     */
-    themeBehavior(
-        Theme.light,
-        css`
-            dialog::backdrop {
-                background: ${modalBackdropColorThemeLightStatic};
```
rajsite, line 165: @jattasNI I think we could potentially support the hard coded values as a fallback: https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#defining_fallbacks_in_the_var_function

But I opted to just switch to the new pattern at the expense of chromatic diffs for the background for these stories (see PR description). I don't really feel strongly either way. I'd still like to clean out the fallback eventually so would keep the tech debt open if we used the fallback instead of what I proposed here.

rajsite: I couldn't find documentation on how often we expect Chromatic to upgrade so created https://github.com/chromaui/chromatic-cli/issues/1021

jattasNI: Looks like they've historically updated just once a year in January or February. Maybe your issue will spur them to update more often, but I'm not too optimistic. I think I read somewhere once that the upgrades are expensive for them because they do a bunch of validation of what tests break across all of their customers suites.

Assuming they'll update early next year I suppose that's a fairly long time for our Chromatic baseline to be wrong. But also annoying to carry the tech debt of using a fallback for that long. I'm landing near the fence too but ultimately I'm ok with the current approach. 🚀

jattasNI (approved):
I validated the backdrop still appears in macOS Safari, Firefox, and Chrome on my machine.

packages/nimble-components/src/drawer/styles.ts
rajsite: Think it also helps that the Chromatic specific rendering difference is easy to evaluate / doesn't impact the primary content of the control. Seems unlikely to make future visual changes to those components difficult to evaluate. ⛴️ing

Issue #2210, opened 2024-06-24
TableCellView tabbableChildren should support changing dynamically


msmithNI:
# 🧹 Tech Debt

Currently `TableCellView` `tabbableChildren` is assumed to be static (or at least, the number of elements in it not changing). This may not be true generally - changes to table/ column configuration may necessitate it changing (more/less interactive elements in a cell; alternatively, an interactive element may get disabled, in which case it's no longer tabbable, and should no longer be included).

We should consider making this property observable, and updating keyboard focus as needed if it changed (re-focus cell if it changes or the current focus is no longer valid, potentially).

msmithNI:
Of note, we may be avoiding a potential error case related to this just due to the focusedRecycleCallback right now - if an anchor is focused when data changes, it'll be blurred, in which case focus resets to the cell. So we shouldn't end up in a case where cellContentIndex is out of bounds even if tabbableChildren changed (anchor link present in initial data, not in updated data).

rajsite:
@fredvisser was mentioning the possibility of editable table cells where on interaction (click / touch / keyboard cell enter) the cell contents are replaced (or text -> numeric text field) and on commit / cancel switched back (numeric text field -> text). Need to prototype but possible this issue may need to be addressed for that.

PR #2340, opened 2024-08-09
Support dynamic tabbableChildren for TableCellView

msmithNI:
# Pull Request

## 🤨 Rationale

#2210 

## 👩‍💻 Implementation

Summary of changes:
- `TableCellView` `tabbableChildren` changed from a getter to be an `@observable` property
- Both anchor & menu button columns only have a tabbable/focusable element in cells in some cases, so opt them into this change.
- Create a new `dynamicRef` directive, used in the anchor/ menu button cell templates. The issue with the `ref` directive is that once a reference is initialized, it's never cleared. So, for example, if an anchor column starts out with valid URL data then gets data without URLs, `cellView.anchor` would still be set (to an element no longer in the DOM) with `ref`, whereas `dynamicRef` clears out the property
    - As an alternative, I considered the `children` directive with a filter, but it seemed like that would require adding yet another extra container div to the cell views, which didn't seem desirable.
- When `tabbableChildren` changes, the cell view fires a bubbling `'cell-tabbable-children-change'` event, handled by `KeyboardNavigationManager`, which focuses the cell instead (if the cell content was previously focused, and is no longer present)

Implementation concerns:
- `'cell-tabbable-children-change'` event
    - This works, but feels somewhat heavy handed (but I couldn't really think of a better alternative).
    - This fires for every visible row, not just the focused row (extra events firing that won't be handled)
    - `KeyboardNavigationManager` keeps track of focused rows/ cells via index, whereas most of the rest of the table code uses `recordId` and `columnId`. In order to only handle the event for the focused row, there's some new code in `KeyboardNavigationManager` that keeps track of the `columnId` / `recordId` if cell content is focused, which complicates that class a bit more.

UX concerns:
- In Chrome, if a focused element inside a cell is removed from the DOM, the table loses focus entirely. `KeyboardNavigationManager` tracks when the table gains/ loses focus, and won't programmatically set focus if the table doesn't already contain focus. So in this case (in Chrome) the user would need to manually refocus the table, at which point the cell would get focused. I discovered this fairly recently so haven't spent too much time pondering what it would take to have the keyboard nav code handle this specific case.
   - This didn't occur in Firefox / WebKit, which is why the new test only conditionally refocuses the table right now.

## 🧪 Testing

- Existing keyboard nav / tabbableChildren tests still pass
- Add new test ensuring that if tabbableChildren are changed (removed), focus reverts to the cell
- No tests for `dynamicRef` currently (we'd want some if we went with this approach)

## ✅ Checklist

N/A for now since this draft PR is to get feedback on this approach.

### DIFF
--- a/change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
+++ b/change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update TableCellView tabbableChildren to be an observable property",
+  "packageName": "@ni/nimble-components",
+  "email": "20709258+msmithNI@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
+++ b/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
@@ -1335,8 +1335,8 @@ implements Subscriber {
                     newColumnIndex,
                     true
                 );
+                return true;
             }
-            return true;
         }
 
         return false;
--- a/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
+++ b/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
@@ -40,6 +40,12 @@ interface TableFocusState {
     cellContentIndex?: number;
 }
 
+interface TableCellContentFocusState {
+    index: number;
+    columnId?: string;
+    recordId?: string;
+}
+
 /**
  * Manages the keyboard navigation and focus within the table.
  * @internal
@@ -49,13 +55,16 @@ implements Subscriber {
     private focusType: TableFocusType = TableFocusType.none;
     private headerActionIndex = -1;
     private rowIndex = -1;
-    private cellContentIndex = -1;
     private columnIndex = -1;
     private focusWithinTable = false;
     private isCurrentlyFocusingElement = false;
     private readonly tableNotifier: Notifier;
     private readonly virtualizerNotifier: Notifier;
     private visibleRowNotifiers: Notifier[] = [];
+    private readonly cellContentState: TableCellContentFocusState = {
+        index: -1
+    };
+
     private get inNavigationMode(): boolean {
         return (
             this.focusType !== TableFocusType.cellActionMenu
@@ -117,6 +126,10 @@ implements Subscriber {
             'cell-blur',
             this.onCellBlur as EventListener
         );
+        this.table.viewport.addEventListener(
+            'cell-tabbable-children-change',
+            this.onCellTabbableChildrenChange as EventListener
+        );
     }
 
     public disconnect(): void {
@@ -156,6 +169,10 @@ implements Subscriber {
         this.table.viewport.removeEventListener(
             'cell-blur',
             this.onCellBlur as EventListener
+        );
+        this.table.viewport.removeEventListener(
+            'cell-tabbable-children-change',
+            this.onCellTabbableChildrenChange as EventListener
         );
     }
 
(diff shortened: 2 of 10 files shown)


msmithNI:
@jattasNI @rajsite @atmgrifter00 interested in your thoughts on this implementation direction so far (but note the concerns I listed in the PR description).

change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @msmithNI: @jattasNI @rajsite @atmgrifter00 interested in your thoughts on this implementation direction so far (but note the concerns I listed in the PR description).

rajsite: The general direction looks reasonable to me.

For the `cell-tabbable-children-change` I'm a bit curious how often that event is firing. Is it firing once per row continuously as you scroll for example? That could potentially be excessive object creation / gc pressure on scroll.

> In Chrome, if a focused element inside a cell is removed from the DOM, the table loses focus entirely.

Do you have a screencap of how that manifests? Not exactly clear how to try out the UX for that situation and if it's a new change in behavior or existing.

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
```diff
@@ -1422,6 +1426,28 @@ describe('Table keyboard navigation', () => {
                     expect(blurSpy).toHaveBeenCalledTimes(1);
                     expect(currentFocusedElement()).not.toBe(cellContent);
                 });
+
+                it('and then the cell updates to no longer have tabbableChildren, the cell is focused instead', async () => {
```
atmgrifter00, line 1430: I presume if we went with this approach (which I think I'm fine with), that there are some other tests we still need to add, yes? Namely, that after the tabbableChildren change (but not completely removed, as this test covers that) that tabbing (and shift-tabbing) behave as expected.

packages/nimble-components/src/utilities/directive/dynamic-ref.ts
```diff
@@ -0,0 +1,62 @@
+import {
```
jattasNI, line 1: Does FAST have unit tests for their RefBehavior that we could leverage for patterns to write unit tests for this class?

packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
```diff
@@ -323,6 +340,28 @@ implements Subscriber {
...
+        event.stopPropagation();
+        const cellView = event.detail;
+        if (
+            this.focusType === TableFocusType.cellContent
+            && cellView.recordId === this.cellContentState.recordId
+            && cellView?.column?.columnId === this.cellContentState.columnId
+        ) {
+            if (
+                this.cellContentState.index >= cellView.tabbableChildren.length
+            ) {
```
jattasNI, line 355: I think you could collapse these two `if`s into one.

change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
jattasNI: No objections to the overall direction from me. (I left a couple comments while wrapping my head around the PR because I couldn't stop myself; neither one is existential, just normal code review feedback)

packages/nimble-components/src/table/tests/table-keyboard-navigation.spec.ts
msmithNI: Agreed, those would be good tests to add too.

change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
msmithNI: When `cell-tabbable-children-change` fires:
- With what's in this PR, there's some unnecessary firing as the table loads with initial data. If we omit firing when `!cellView.$fastController.isConnected`, we could prevent that though.
- During scrolling:
  - Anchor column: Yes, depending on what's in the cells. With the Storybook example, yes, since the recycled cells change between having valid links / no link / placeholder text, as you scroll (depending on the data).
  - Menu button: Yes with this current PR. But not if we skip firing when the FAST controller is not connected.

msmithNI: > Do you have a screencap of how that manifests?

Not currently, I was just going by the behavior of the new autotest failing in Chrome without the code to refocus the table. If we updated an example to periodically update the table data with an anchor column periodically switching between a valid URL / no URL, that would probably show it.

packages/nimble-components/src/utilities/directive/dynamic-ref.ts
msmithNI: Not that I've found, but writing our own should be pretty quick anyway.

change/@ni-nimble-components-8499a62b-64ea-41e0-877c-6bacca7ae0c8.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "Update TableCellView tabbableChildren to be an observable property",
```
rajsite, line 3: Closing the draft PR as the discussion is stale. FYI @msmithNI @fredvisser I'm betting this refactor is needed as part of the editable table cells implementation where I suspect new tabbable elements will be created dynamically
https://github.com/ni/nimble/pull/2477

PR #2341, opened 2024-08-09
Listbox padding

fredvisser:
# Pull Request

## 🤨 Rationale

Brandon noted that the visual design for select, combobox, and mention, specifies a 4px 'mat' around the options, even when the content is scrolling. 

## 👩‍💻 Implementation

Screenshot 2024-08-09 at 3 31 01 PM

https://github.com/user-attachments/assets/5df5ffa1-be2d-482c-90d2-4578e3f6c978

- Moved `padding: var(--ni-private-listbox-padding);` from the scrollable content, to the fixed listbox container
- Made the `combobox` template more consistent with the `select` template, so the shared styling at the same effect (i.e. moved the `scrollable-region` class into it's own div)

## 🧪 Testing

- Manual Storybook review
- Chromatic

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/patterns/dropdown/styles.ts
+++ b/packages/nimble-components/src/patterns/dropdown/styles.ts
@@ -220,6 +220,7 @@ export const styles = css`
         box-shadow: ${elevation2BoxShadow};
         border: ${borderWidth} solid ${popupBorderColor};
         background-color: ${applicationBackgroundColor};
+        padding: var(--ni-private-listbox-padding);
     }
 
     .listbox:has(.filter-field) {
@@ -247,7 +248,6 @@ export const styles = css`
     .listbox slot {
         display: block;
         background: transparent;
-        padding: var(--ni-private-listbox-padding);
     }
 
     .listbox.empty slot {
--- a/packages/nimble-components/src/patterns/dropdown/styles.ts
+++ b/packages/nimble-components/src/patterns/dropdown/styles.ts
@@ -243,7 +243,6 @@ export const styles = css`
 
     .scrollable-region {
         overflow-y: auto;
-        scroll-snap-type: y mandatory;
     }
 
     .listbox slot {
@@ -258,7 +257,6 @@ export const styles = css`
     ::slotted([role='option']),
     ::slotted(option) {
         flex: none;
-        scroll-snap-align: start none;
     }
 
     .no-results-label {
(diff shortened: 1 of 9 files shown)


packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -242,12 +243,12 @@ export const styles = css`
 
     .scrollable-region {
         overflow-y: auto;
+        scroll-snap-type: y mandatory;
```
rajsite, line 246: Tried the storybook for Chrome on Windows and it ends up feeling kinda clunky. The scroll on open is too slow and feels like it gets in the way. It also looks like Chrome on Windows using the mousewheel to scroll does not have kinetic scrolling or something. So it always stops then shifts again and feels pretty uncomfortable.

I tried to record a gif but it's not showing enough frames to see the issue. The shift on Option 47 is what I see on each option as I'm scrolling:

scrollsnapwinchrome

fredvisser: I don't see the same with a Windows 11 VM + mouse scroll wheel, but I've pulled the change out of this PR, and we can evaluate if these APIs are well enough supported to use in another PR.

packages/nimble-components/src/patterns/dropdown/styles.ts
```diff
@@ -220,6 +220,7 @@ export const styles = css`
         box-shadow: ${elevation2BoxShadow};
         border: ${borderWidth} solid ${popupBorderColor};
         background-color: ${applicationBackgroundColor};
+        padding: var(--ni-private-listbox-padding);
```
rajsite, line 223: on windows this change is resulting in the padding being around the scrollable region which seems off.

main:
image

pr:
image

jattasNI: There are also some suspicious Chromatic diffs. For example, the [Search field seems squished](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=66ba7d027a3c4e929521ac2c) and the [No items found item seems taller](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=66ba7d027a3c4e929521ac1a)

rajsite: @fredvisser fyi the latest commit does still seem to impact the scrollbar, notice the top and bottom of the list box:
image

fredvisser: Closing the loop - the initial behavior (padding around the scrollbar) is the intended design. 

However, @atmgrifter00 says he has this fix in a branch, so I'll abandon this PR.

PR #2342, opened 2024-08-09
Remove unnecessary font size precision

rajsite:
# Pull Request

## 🤨 Rationale

Aligns the font values with Figma:
image

Only real motivation was I was tired of seeing it in devtools 😅
image

## 👩‍💻 Implementation

Update font values with excess precision.

## 🧪 Testing

Rely on chromatic

## ✅ Checklist

### DIFF
--- a/packages/storybook/.storybook/blocks/story-layout.css
+++ b/packages/storybook/.storybook/blocks/story-layout.css
@@ -1,5 +1,3 @@
-@import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
-
 .story-layout-container {
     display: grid;
     align-items: center;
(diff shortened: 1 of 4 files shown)


packages/storybook/.storybook/blocks/story-layout.css
```diff
@@ -1,5 +1,3 @@
-@import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
```
rajsite, line 1: saw in devtools that SB was importing base tokens but they don't seem to be used so removed.

rajsite:
SB shows no effective change in usages (tokens table adjusted slightly but no actual component usage change). Gonna override @jattasNI and @fredvisser for trivial change

PR #2344, opened 2024-08-14
Fix Select Placeholder story and width doc/bug

### LINKED ISSUE #2336: Select component resizes when an option with long text is selected
# 🐛 Bug Report

The width of the select has started changing depending on which option is selected. It's supposed to stay the same width and show an ellipsis if the option is too long to fix.

## 💻 Repro or Code Sample

https://nimble.ni.dev/storybook/?path=/docs/components-select--docs

Change the default slot to "wide options" and toggle the value between a wide and a narrow option

https://github.com/user-attachments/assets/afdd61c0-b458-41e6-b6d9-af9cd7fa3f70

## 🤔 Expected Behavior

The component doesn't resize

## 😯 Current Behavior

The component resizes

## 💁 Possible Solution

I suspect this broke with #2315 because that PR touched the component's styling and a build from shortly before then doesn't exhibit the bug. FYI @msmithNI 

## 🌍 Your Environment

macOS Firefox

fredvisser:
# Pull Request

## 🤨 Rationale

Fixes #2336, `nimble-select` Placeholder story controls, and minor story formatting.

## 👩‍💻 Implementation

- Changed `min-width` to `width` to resolve control changing size when selecting longer options
- Refactored 'Placeholder' story (so the Story controls work correctly) by creating a set of options that includes a placeholder.
- Add basic styling guidance to Select docs

## 🧪 Testing

- Manual storybook review
- Chromatic

## ✅ Checklist


packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -240,6 +259,14 @@ const metadata: Meta<SelectArgs> = {
             description: 'Emitted when the user types in the filter input.',
             table: { category: apiCategory.events },
             control: false
+        },
+        width: {
+            name: 'width',
+            description:
```
jattasNI, line 265: I confirmed that width is working as I'd expect now. If it's set to `auto` or unset then setting a long label or picking a long option will grow the width. If it's set to a fixed length like a `px` or `%` value then long labels will wrap and long options will ellipsize. 👍

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -97,6 +112,7 @@ const getGroupedOptions = (): GroupedOptionArgs[] => {
 
 const optionSets = {
     [ExampleOptionsType.simpleOptions]: simpleOptions,
+    [ExampleOptionsType.placeholderOptions]: placeholderOptions,
```
jattasNI, line 115: Did you intend for this to show up as a selectable option in the args table? If so can you give it a Sentence case name?

image

jattasNI (approved):

packages/storybook/src/nimble/select/select.stories.ts
rajsite: I'm not really seeing the need to document this explicitly. Our controls behave like standard html elements. Setting the width sets the width. Why would we call it out only for this specific control?

If we have sizing style usage guidance we should document that in the mdx but the current description isn't really usage guidance, just a general description of how CSS width properties work.

fredvisser: Fair point - @jattasNI - thoughts on the limiting principle for this type of documentation?

The latest push removes the width args…

packages/storybook/src/nimble/select/select.stories.ts
fredvisser: Yes - done - 👍

packages/storybook/src/nimble/select/select.stories.ts
jattasNI: I am ok with removing the arg and agree that the description was pretty general.

Longer term I think sizing usage guidance under Styling>>Sizing would be useful for all components. If we wanted to try to start doing that here I think it could look something like this:
> This component's height is fixed and its width obeys standard CSS sizing rules. If the component has no explicit `width` (e.g. `width: auto`) then it will grow to fit both the label and the currently selected item. When it reaches its `max-width` the label will wrap and the selected item will ellipsize. Applications with fixed layouts that shouldn't change in response to user interaction should set an explicit width.

@fredvisser your call whether to include something like this now or save it for a more dedicated effort across all components. And if we are including it, @rajsite you're welcome to wordsmith with more precise CSS terminology.

rajsite: That is a good textual representation of what the control does in text form. As a user that tells me the control behavior I see is documented. That could be useful but seems like a lot of work to document and maybe is redundant? I think if someone really wanted to know those behaviors it should be covered in a matrix test and can see what long labels look like there (i.e. if their goal is validating what they see is expected).

If a user is coming to the style section to know how they should style a control I don't think all the detail helps. I just want to know what to do as quickly as possible so someone from Nimble or UX won't complain about how I sized the control.

I'd say simplify to something that gives both reasoning and guidance to build decisions on:
- By default the select control resizes to the selected option size which keeps the content visible and the control compact.
- In a grid layout consider giving the select a specific width to align the size with other grid items. The control should be wide enough so labels do not wrap.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -35,6 +35,7 @@ interface SelectArgs {
     value: string;
     change: undefined;
     filterInput: undefined;
+    width: string;
```
rajsite, line 38: not needed anymore

rajsite (commented):
Implementation looks good and like the placeholder story cleanup! Just docs discussions left

packages/storybook/src/nimble/select/select.stories.ts
fredvisser: Added the styling guidance as…

>## Styling

>By default, the select control resizes to the selected option size. In most
>layouts, consider giving the select a specific `width` to align with other
>items. Ensure the control is wide enough so labels do not wrap.

rajsite (approved):

Issue #2274, opened 2024-07-15
Intermittent "scroll selected option into view" test failures in webkit


m-akinc:
# 🧹 Tech Debt

An intermittent failure of the following `nimble-rich-text-mention-listbox` test was seen on webkit:

> RichTextMentionListbox should scroll the selected option into view when opened FAILED
> [test-concurrent:nimble-components] [test-webkit:verbose] 	Expected 0 to be greater than 8000.
> [test-concurrent:nimble-components] [test-webkit:verbose] 	Expected 0 to be close to 4.

There are similar tests for the combobox and rich-text mention listbox. Though they have not been seen to fail, it is likely they are intermittent as well and are being disabled under that assumption.

The tests will be disabled until this issue is resolved.

Failing run: https://github.com/ni/nimble/actions/runs/9947647052/job/27480698021

Issue #2272, opened 2024-07-15
Intermittent nimble-select test failure in webkit (should not confine dropdown)


mollykreis:
# 🧹 Tech Debt

There is an intermittent test failure in webkit related to the nimble-select:

>[test-webkit:verbose] Safari 17.0 (Linux x86_64) Select within a div should not confine dropdown to div with "overflow: auto" FAILED
>[test-webkit:verbose] 	Expected false to be true.

The test will be disabled until this issue is resolved.

Failing run: https://github.com/ni/nimble/actions/runs/9946563531/job/27477385164

Issue #2269, opened 2024-07-15
Intermittent "limit dropdown height to viewport" test failures in webkit


mollykreis:
# 🧹 Tech Debt

There are intermittent `nimble-select` and `nimble-rich-text-mention-listbox` test failures in webkit related to dropdown height :

> Select with 500 options should limit dropdown height to viewport FAILED
> [test-webkit:verbose] 	Expected false to be true.

> RichTextMentionListbox should limit dropdown height to viewport FAILED
> [test-concurrent:nimble-components] [test-webkit:verbose] 	Expected false to be true.

The tests will be disabled until this issue is resolved.

Failing runs:
https://github.com/ni/nimble/actions/runs/9914481179/job/27393640409
https://github.com/ni/nimble/actions/runs/9947647052/job/27480698021

PR #2349, opened 2024-08-16
Use spec reporter for CI test commands

jattasNI:
# Pull Request

## 🤨 Rationale

Occasionally we've seen intermittent tests due to components being pushed off screen by the karma-jasmine-html-reporter progress UI. While we've attempted to eliminate those by [placing the component under test at the top of the page](https://github.com/ni/nimble/blob/68e220a0e2c188fb76cc1868142db1aa0eef3a3d/packages/nimble-components/src/utilities/tests/fixture.ts#L148-L153), there is still a risk of behavior changes because the page length is changing. That progress UI is not visible for headless CI runs anyway, so I propose removing it.

This is inspired by #2269, #2272, #2274, but I'm planning to submit a few different attempts to fix those in separate PRs before re-enabling tests.

## 👩‍💻 Implementation

1. Update the dedicated karma configuration that we use for the `test` and `test-concurrent` commands that are run on the CI for `nimble-components` and `spright-components`. Now it uses only the `spec` reporter (prints tests to console) and not the `kjhtml` reporter (draws dots on the HTML page).
1. Rename the dedicated karma configuration to "headless" since it hasn't been "verbose" for a while.
1. Updated other headless commands (`test-chrome`, `test-chrome:watch`, etc) to use the same configuration. This means they'll start printing skipped tests to the console instead of just failing tests. Allowed some commands to be consolidated.
1. Considered making similar changes to karma files for other packages but they haven't exhibited similar intermittency so it didn't seem worthwhile.

## 🧪 Testing

- Inspected the build output and confirmed the test report is the same as before (shows skipped tests and presumably failed tests, but not passing ones).
- not statistically significant, but the validate step was about a minute faster than it had been for the last few runs

## ✅ Checklist


packages/nimble-components/karma.conf.ci.js
```diff
@@ -1,5 +1,5 @@
 /**
- * A karma config file that adds verbose reporting to the base configuration
+ * A karma config file that sets up reporting for CI builds
```
rajsite, line 2: I'd vote for "quiet", but mainly I want to avoid just having the bucket "ci" because that doesn't really tell me anything about how it runs different

rajsite: or maybe "headless"

jattasNI: I hadn't gone with headless because we weren't using it for all of our headless commands (`test-chrome`, `test-chrome:watch`, and other browser equivalents). And I was initially reluctant to do so because I thought those commands were run by devs and needed better live logging to the terminal.

But it looks like their live logging to the terminal is also quiet (only prints failing tests, no live feedback about how many tests have passed). So we could consolidate those commands:

`karma.conf.headless.js` called by `test-chrome` and `test-chrome:watch`. Delete `test-chrome:verbose` and update `test-concurrent` to call `test-chrome` instead.

The only behavior change would be that skipped tests would now be printed to the console when you call `test-chrome`, which I think I'm fine with.

rajsite (approved):

PR #2361, opened 2024-08-28
Disable table row hover styling while scrolling

### LINKED ISSUE #2312: Don't change row hover highlight during table scroll
# 🐛 Bug Report

The nimble-table moves the row hover highlight during page or table scroll, which adds unnecessary visual noise to the scrolling process. 

For example, compare nimble-table:
https://github.com/user-attachments/assets/00cb4136-b37e-4071-9bff-24d412e9f106

with ag-grid:
https://github.com/user-attachments/assets/24b4b240-dde6-496e-98f0-f1bffb81d41d

## 💻 Repro or Code Sample

https://nimble.ni.dev/storybook/?path=/story/components-table--table&args=data:LargeDataSet;selectionMode:single

## 🤔 Expected Behavior

[NNGroup suggests](https://www.nngroup.com/articles/timing-exposing-content/) that:

>For hover interactions, the timing guidelines for speed of visual feedback and exposing hidden elements must be broken down into more steps:
>Mouse cursor enters target area: display visual feedback within 0.1 seconds.
>Wait 0.3–0.5 seconds.
>If cursor remains stopped within target area, display corresponding hidden content within 0.1 seconds.
>Keep displaying the exposed content element until the cursor has left the triggering target area or the exposed content for longer than 0.5 seconds.

But we can evaluate if that is consistent with other tables.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2312

## 👩‍💻 Implementation

The TanStack virtualizer exposes an `isScrolling` property. We already observe the virtualizer for changes (which is triggered when `isScrolling` changes), so when that is triggered we can update a new observable property on our (Nimble) Virtualizer class. The table template uses that property to set/remove an `allow-hover` class on each row. I updated the row styling to only apply the hover styling when the `allow-hover` class is present.

I wanted the class to be `scrolling`, but that would require us to use `:not()` selectors in our styles, which our conventions say to avoid. Thus I inverted the semantics of the class and picked the affirmative `allow-hover` (as opposed to `not-scrolling`).

## 🧪 Testing

I don't think there's a practical way to really verifty the end behavior we want. Our visual tests can't take consistent snapshots while actually scrolling, and a unit test can't simulaneously scroll and assert the state. Furthermore, there doesn't seem to be a practical way to trigger the `:hover` pseudo-class in a unit test, so we can't test that our CSS is written properly. The best I could do was to mock the virtualizer's `isScrolling` property to return `true` and test that the `allow-hover` class does not get applied.

## ✅ Checklist


m-akinc:
@mollykreis would you mind buddying this?

packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -25,6 +25,10 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
     @observable
     public scrollHeight = 0;
 
+    public get isScrolling(): boolean {
```
m-akinc, line 28: I had to make this a property with a getter (rather than just a data field) to be able to mock it with `jasmine.spyOnProperty`. I then found that `@observable` doesn't seem to work on a property with only a getter, so I just added an observable, private field.

packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -117,6 +123,7 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
             scrollToFn: elementScroll,
             observeElementOffset,
             observeElementRect,
+            isScrollingResetDelay: 0,
```
m-akinc, line 126: This is basically a debounce duration ([defaults to 150 ms](https://tanstack.com/virtual/v3/docs/api/virtualizer#isscrollingresetdelay)) that controls how long after the last scroll `isScrolling` gets set to `false`. A delay of zero should make it feel more responsive (assuming the virtualizer's change handler isn't otherwise debounced/delayed).

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -478,6 +478,29 @@ describe('Table', () => {
             expect(element.checkValidity()).toBeTrue();
         });
 
+        it('allows row hover styling when not scrolling', async () => {
```
mollykreis, line 481: I question the value of these tests, particularly because they don't provide full code coverage and they required writing the implementation code in a non-standard way.

Is it possible to send scroll events to the appropriate element within the table rather than having to fake the value of `isScrolling`? We have tests that scroll the table and get the virtualizer to return different data, so it seems like it should be possible to have the virtualizer know that it is scrolling.

packages/nimble-components/src/table/template.ts
```diff
@@ -145,7 +145,7 @@ export const template = html<Table>`
                             ${repeat(x => x.virtualizer.visibleItems, html<VirtualItem<HTMLElement>, Table>`
                                 ${when((x, c: ExecutionContext<Table>) => c.parent.tableData[x.index]?.isGroupRow, html<VirtualItem<HTMLElement>, Table>`
                                     <${tableGroupRowTag}
-                                        class="group-row"
+                                        class="group-row ${(_, c) => (c.parent.virtualizer.isScrolling ? '' : 'allow-hover')}"
```
mollykreis, line 148: I'm wondering if adding this state with class is the approach we want. I can't think of any other place where we add a CSS class to a child element and then put the styling associated with that CSS class in the child element's styles. Generally, if we want state to be passed to the child, we use an attribute. Is there a reason to use a CSS class instead here?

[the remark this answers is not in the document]
m-akinc: Good catch. Thanks.

packages/nimble-components/src/table/tests/table.spec.ts
m-akinc: Yeah, in the PR description, I explained some of my reservations about these tests and the difficulties of writing more complete tests. These did not catch the fact that I missed adding styling for group rows. I can try to mitigate the concern of "writing the implementation in a non-standard way" by using scroll events instead of mocking the virtualizer, though I'm a little worried that approach is more prone to intermittency. But if we have existing tests doing this non-intermittently, then maybe it's fine.

packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -113,7 +116,7 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
             estimateSize: (_: number) => rowHeight,
             enableSmoothScroll: true,
             overscan: 3,
-            scrollingDelay: 5,
+            isScrollingResetDelay: 0,
```
m-akinc, line 119: This option [was renamed](https://github.com/TanStack/virtual/discussions/714#discussioncomment-9258710).

packages/nimble-components/src/table/models/virtualizer.ts
m-akinc: We were already setting `scrollingDelay` (which was renamed to `isScrollingResetDelay`) to 5. Might as well be 0, though.

mollykreis (approved):

packages/nimble-components/src/table/components/group-row/index.ts
```diff
@@ -66,6 +66,12 @@ export class TableGroupRow extends FoundationElement {
     @observable
     public animationClass = '';
 
+    /**
+     * @internal
+     */
+    @attr({ attribute: 'allow-hover-styling', mode: 'boolean' })
+    public allowHoverStyling = true;
```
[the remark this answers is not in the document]
m-akinc: Is that just a stylistic thing?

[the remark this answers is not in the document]
rajsite: Nope, it's a consequence of the definition of boolean attributes asserting that absence means false. See https://github.com/microsoft/fast/issues/5320#issuecomment-951157316 and lets capture in CONTRIBUTING docs if not there already.

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: You are going to hate me: I'm seeing some clunky behavior in Safari. 

When scrolling slowly, I still see flickers of row highlighting. Similar when using momentum scroll as the scroll slows down, but it's even worse because the flicker occurs over a row that I'm not hovered on.

You can see both effects in this video (first tab is the build from this PR, second tab is the public storybook).

Edge and Firefox on macOS both work great.

https://github.com/user-attachments/assets/fa59ec76-7619-45d3-bc1f-13b0929e1509

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: While using my mousewheel I get into a state (inconsistently) where the hover stops working:

hovereffect

m-akinc: Weird. I had seen something like that myself, but it was only when I got to the very top or bottom. This is clearly more pervasive.

packages/nimble-components/src/table/components/group-row/index.ts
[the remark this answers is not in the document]
m-akinc: Added the CONTRIBUTING note in PR #2373

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
m-akinc: Ugh. Looks like there are a couple browser bugs involved 😭
scrollend-bug-chrome
Above is Chrome. In some cases the `scrollend` event doesn't get fired. It seems like mouse movement is a necessary component of the bug.

On Firefox, we very frequently get a `scroll` event (indicated by a `.`) _after_ the `scrollend` event, which effectively looks like an open-ended scroll to TanStack.
scrollend-bug-firefox
https://codepen.io/m-akinc/pen/gONQawN

I haven't filed browser bugs, but I will.

In this case, I think a reasonable workaround is to simply sett `isScrolling` to false if we haven't gotten a virtualizer change notification in 300ms (we get one for each `scroll` event). Let me know if you still see any weirdness.

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
jattasNI: Looking a lot better! I've only noticed one issue so far: there was one time that a row got stuck in the hovered state even though the mouse wasn't over it. If I hovered other rows, I ended up with two rows showing the hover state.

I think it happened when I had right clicked on the table to open dev tools in a new window. When I later returned to the primary browser window it was in that state. But I haven't been able to reproduce those conditions again; I'll keep trying.

This isn't fatal or anything: just hovering over the row that was stuck was sufficient to put it back in a good state.

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
m-akinc: I added my report to an [existing Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1896515#c4).
Filed [this bug](https://issues.chromium.org/issues/364568425) to Chromium.

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
m-akinc: That doesn't sound like it would be related to the original problem or my change. I don't think there's any way for our hover styling to apply unless the browser thinks it is hovered (i.e. the `:hover` pseudo-state applies). My changes could only _reduce_ the cases where hover styling is applied, and I can't affect whether the browser does or doesn't register that an element is being hovered. My guess is we'd be equally likely to reproduce what you saw in `main`. Based on that reasoning, I'm going to mark this thread resolved.

packages/nimble-components/src/table/models/virtualizer.ts
m-akinc: Changed it to 250ms (which is also the default) to avoid the noisy flickering that Jesse reported in another thread.

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -478,6 +478,28 @@ describe('Table', () => {
...
+            await connect();
+            await element.setData(simpleTableData);
+            await waitForUpdatesAsync();
```
jattasNI, line 484: nit: is this `waitForUpdatesAsync();` needed? I'd expect that we could do all the setup and then only wait for it to take effect before asserting. Same for the first one (and maybe two) in the test below.

packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -47,6 +50,7 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
     private readonly viewportResizeObserver: ResizeObserver;
     private virtualizer?: TanStackVirtualizer<HTMLElement, HTMLElement>;
     private _pageSize = 0;
+    private isScrollingTimeout = 0;
```
jattasNI, line 53: nit: to me this name makes it seem like this is a duration but the value returned from setTimeout is actually an integer that uniquely identifies the timer. Maybe `isScrollingTimer` or `isScrollingTimeoutId`? MDN is a bit inconsistent but they sometimes use "timeout id" (and sometimes use timeout as you are, so maybe this is just my hangup).

packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -125,6 +129,15 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
         const virtualizer = this.virtualizer!;
         this.visibleItems = virtualizer.getVirtualItems();
         this.scrollHeight = virtualizer.getTotalSize();
+        this.isScrolling = virtualizer.isScrolling;
+        // There are multiple browser bugs that can result in us getting stuck thinking that we're scrolling.
```
jattasNI, line 133: Now that you've got bug reports, could you link to them from here?

Alternatively we could also file a Nimble issue that links to the browser issues and mark it blocked, then link to the Nimble issue. That probably gives us a slightly better chance of removing this workaround someday.

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -478,6 +478,28 @@ describe('Table', () => {
...
+        it('does not allow row hover styling while scrolling', async () => {
+            await connect();
+            await element.setData(simpleTableData);
+            await waitForUpdatesAsync();
+            column1.groupIndex = 0;
+            await waitForUpdatesAsync();
+
+            element.viewport.dispatchEvent(new Event('scroll'));
+            await waitForUpdatesAsync();
+            expect(pageObject.isRowHoverStylingEnabled()).toBeFalse();
```
jattasNI, line 500: If we wanted to check that hover styling got re-enabled when scrolling finished, what would that look like? I'm guessing:
1. dispatch `'scrollend'`
2. wait for a sufficient amount of time for the timers to finish reliably
3. assert

Assuming that step 2 is messy, I'm ok not testing this. But just wanted to make sure we thought about it.

jattasNI (approved):

packages/nimble-components/src/table/components/group-row/index.ts
```diff
@@ -66,6 +66,12 @@ export class TableGroupRow extends FoundationElement {
     @observable
     public animationClass = '';
 
+    /**
+     * @internal
+     */
+    @attr({ attribute: 'allow-hover-styling', mode: 'boolean' })
```
rajsite, line 72: nit: Including `styling` in the name feels weird / too leaky about the implementation. Can we do `allow-hover`?

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -133,6 +133,12 @@ export class TableRow<
     @observable
     public animationClass = '';
 
+    /**
+     * @internal
+     */
+    @attr({ attribute: 'allow-hover-styling', mode: 'boolean' })
```
rajsite, line 139: nit: Including `styling` in the name feels weird / too leaky about the implementation. Can we do `allow-hover`?

rajsite (approved):

change/@ni-nimble-components-8ca8762f-ad08-436d-980f-56d9f5ba9e39.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Disable table row hover styling while scrolling",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
```
rajsite, line 6: The current change do see to improve the mobile behavior too! https://github.com/ni/nimble/issues/2312#issuecomment-2273419492

While scrolling you no longer see the hover state, it just appears again a little after stopping. Feels like an improvement 👍

packages/nimble-components/src/table/models/virtualizer.ts
m-akinc: Agreed. Went with ` isScrollingTimer`.

packages/nimble-components/src/table/tests/table.spec.ts
m-akinc: I think this is a reasonable test to write. If it ends up being flaky, we can nix it.

PR #2364, opened 2024-08-29
Update Node version from 20 to 22

jattasNI:
# Pull Request

## 🤨 Rationale

We were on Node 20 which is still LTS but 22 is now current and will become LTS. [Skyline is switching to 22](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/786732) and [javascript-styleguide is too](https://github.com/ni/javascript-styleguide/pull/154) so it makes sense to align with what will be on dev machines.

## 👩‍💻 Implementation

Update GitHub Actions workflow and CONTRIBUTING.md

## ✅ Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
           global-json-file: packages/blazor-workspace/global.json
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           registry-url: 'https://registry.npmjs.org'
           cache: 'npm'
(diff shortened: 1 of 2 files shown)


rajsite:
Huh, we have a rennovate config to update the github action steps: https://github.com/ni/nimble/blob/21616fc7a1705f92f4e5774d6827debaed96a354/.github/renovate.json#L52-L56

But guess this would be updating the version passed to specific steps.
Nothing to do just a curiosity

rajsite (approved):

PR #2376, opened 2024-09-04
Remove combobox autocomplete modes

### LINKED ISSUE #2347: Combobox with autocomplete="both" filters out too many options
# 🐛 Bug Report

When the combobox is set to `autocomplete="both"`, it is changing its value to equal the item that autocompletes. This causes other items that should match the entered text to be filtered out.

## 💻 Repro or Code Sample

1. In the [Storybook combobox page](https://nimble.ni.dev/storybook/?path=/docs/components-combobox--docs), set `autocomplete` to `both` and choose "Many options"
2. Type "S"

https://github.com/user-attachments/assets/9556d96e-ffd3-4f48-a41f-3e29d1a6ff1e

## 🤔 Expected Behavior

All of the "Sue..." items should be visible

## 😯 Current Behavior

Only the first matching item, "Sue (1)", is visible.

## 💁 Possible Solution

@atmgrifter00 says 
>So I think I have an understanding of why the combobox behaves the way it does, and it basically involves us putting in some workarounds to make the combobox behave more like a text field, which FAST wasn't ready (or willing) to do. Unfortunately, we just had an oversight resulting in this bug, which the fix for is, annoyingly enough, likely going to involve some more state management for the filter text.
 

## 🔦 Context

This was originally reported as an SLE bug: [Bug 2670859](https://dev.azure.com/ni/DevCentral/_workitems/edit/2670859): Nimble combo box | Drop down does not filter the matched text correctly

That will likely be fixed by migrating to a filterable select. Another possible workaround is to use `autocomplete="list"`.

## 🌍 Your Environment

Not environment-dependent

jattasNI:
# Pull Request

## 🤨 Rationale

Fixes #2347. In that issue we decided to deprecate the combobox autocomplete modes `both` and `inline` because they are too aggressive about filtering out options that should match what the user typed but don't match what the component autocompleted. Short term, clients can either use `autocomplete="list"` or switch to a filterable `nimble-select`. Long term we may change combobox autocomplete to behave more like select.

## 👩‍💻 Implementation

1. Created a type for the autocomplete attribute in Nimble to replace the one from FAST
2. Deleted code from nimble-components that was only relevant to those modes
3. Angular and Blazor: trivial test and example app updates, plus updating the Blazor type that corresponds to the nimble-components one.
4. Storybook: remove docs and tests that reference those modes

In [this SLE PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/790806) I'm updating places that used "both" to now use "list".

## 🧪 Testing

- Updated tests that didn't care about the mode to use a mode that's still supported.
- Deleted tests that no longer apply
- the forked "foundation" test still uses the FAST type since it's testing FAST code that uses that type. There was one place where the test is relying on the Nimble element so it produced a type mismatch, but by luck that code was using modes that are still supported in both places, so some type tweaks allowed it to continue working.

## ✅ Checklist


packages/nimble-components/src/combobox/tests/combobox.spec.ts
```diff
@@ -525,7 +519,7 @@ describe('Combobox', () => {
             // prettier-ignore
             const viewTemplate = html`
                 <${comboboxTag}
-                    autocomplete="inline"
+                    autocomplete="list"
```
jattasNI, line 522: This change causes the test in this describe block to fail both with and without the other changes in this PR. I haven't yet investigated why: interactively it behaves correctly to scroll the selected item into view when `autocomplete="list"`.

atmgrifter00: So, the issue is that when you change to `list` the items in the dropdown are limited to what has been typed in the input, whereas before they were not. Changing this to `none` (or simply removing this attribute), in addition to a change needed in the `focusAndScrollOptionIntoView` implementation (see this [commit](https://github.com/ni/nimble/pull/2376/commits/c53bc58288b301d6bbceb5a6d9048532d357cb31)).

packages/nimble-components/src/combobox/index.ts
```diff
@@ -667,15 +632,12 @@ export class Combobox
...
     protected override focusAndScrollOptionIntoView(): void {
         if (this.open) {
-            if (this.contains(document.activeElement)) {
-                this.control.focus();
-                if (this.firstSelectedOption) {
-                    requestAnimationFrame(() => {
-                        this.firstSelectedOption?.scrollIntoView({
-                            block: 'nearest'
-                        });
+            if (this.firstSelectedOption) {
```
atmgrifter00, line 635: @jattasNI and @rajsitem the code I've removed here was partially responsible for keeping a test from passing. Essentially `document.activeElement` had not been updated to something contained within the `Combobox`. While I can't say I confidently know why that is, the removed logic still seems wholly unnecessary. I see no reason why we have to explicitly place focus on the `input` in the template when scrolling the selected option into view, nor a reason to gate this on the currently focused element already being with the `Combobox`. This method should just be concerned with scrolling the current selected item into view.

PR #2381, opened 2024-09-04
Fix animation jumping in right slide-in on Safari

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes AzDO bug [2323886](https://ni.visualstudio.com/DevCentral/_workitems/edit/2323886)

## 👩‍💻 Implementation

I went through all kinds of experiments with varying degrees of success, but ultimately I stumbled upon a very small, effectively unnoticeable change (for non-Safari browsers) that consistently avoids the bug, as far as I have seen. The bug seems to be related to some race condition between positioning/layout of the content div and calculation of animation frame translations. Apparently starting the div completely off-screen causes problems, so starting the slide-in from 95% off-screen is an effective workaround. Curiously, even 99%, 98%, etc. still exhibit the bad behavior. 

I have filed a [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=279148).

## 🧪 Testing

Tested with [this codepen](https://codepen.io/m-akinc/pen/PorBvaY) on a Mac at my desk.
I'm also having Jesse and Fred test on their own Macs.

## ✅ Checklist


rajsite (approved):
wild 🤪

jattasNI (approved):

PR #2385, opened 2024-09-05
Button label nowrap

### LINKED ISSUE #2384: Button label shouldn't wrap
# 🐛 Bug Report

The label text within the `nimble-button` should not wrap. Instead the button should be a consistent size (for a given label) in all layouts and view sizes. 

Screenshot 2024-09-05 at 8 51 54 AM

## 💻 Repro or Code Sample

1. Button [Storybook page with long button label](https://nimble.ni.dev/storybook/?path=/story/components-button--outline-button&args=label:Outline+Button+kjhask+jhdka+kasha+askj+haskjh+asdkjh+askjh+asdk+jh+asdkjh+adsk+haskjh+adskjh+adskjh+adskj+kjh+kjasdhk+jashksdajh+k+asjh+askjh+asdkjh+adskjh+adskjh+asdkj+askjahsd+kjasdh+kadsjh;appearance:outline;appearanceVariant:default;icon:!false;endIcon:!false;contentHidden:!false;disabled:!false)
2. Reduce the width of the window until the button clips the view
3. The label text will wrap

image

## 🤔 Expected Behavior

The button label text shouldn't wrap to a second line. Instead the button should be as wide as necessary to display the entire label text (along with the standard button padding)

Note: We do not need to add a max-width to the button or support label truncation with this change. Future requirements may differ, but for now the entire button label should always be shown.

fredvisser:
# Pull Request

## 🤨 Rationale

Fixes #2384 

Before:
Screenshot 2024-09-05 at 9 40 16 AM

After:
Screenshot 2024-09-05 at 10 43 06 AM

## 👩‍💻 Implementation

- Added `white-space: nowrap;` to shared button `:host` style

## 🧪 Testing

- Manual Storybook + Chromatic

## ✅ Checklist


change/@ni-nimble-components-dbd8f845-12ff-44bd-b2de-86145d22572c.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The button mdx has a section about sizing but it only mentions height. Should we add info about width too? This might be over-explained so feel free to trim it, but maybe something like: "The button will grow its width to fit its contents. Ensure its container has sufficient space for its potential content, including localized text. If an explicit width is necessary, prefer `min-width` over `width` or `max-width` to avoid content overflowing."

If that's valuable we should probably move it to `packages/storybook/src/nimble/patterns/button/styling-docs.mdx` so that it shows up for anchor/toggle/menu buttons too.

change/@ni-nimble-components-dbd8f845-12ff-44bd-b2de-86145d22572c.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "Add nowrap to button content styling",
```
jattasNI, line 3: >It looks like Chromatic is reporting that the icon button snapshot is a bit smaller than before. Perhaps, the snapshot was allocating space for the hidden label, and that size changed?

It's definitely weird but I poked around in the actual DOM and didn't see any issues so I'm fine to accept the diffs.

change/@ni-nimble-components-dbd8f845-12ff-44bd-b2de-86145d22572c.json
rajsite: >   If an explicit width is necessary, prefer min-width over width or max-width to avoid content overflowing."

I don't think I follow the intention here but may be wording. If I want an "explicit width" why prefer min/max-width instead of setting an explicit width? and how does setting a max-width avoid overflow?

jattasNI: Yep, my wording is unclear. I meant "prefer `min-width` and avoid both `width` and `max-width`".

This is admittedly still controversial / opinionated. But with the current implementation if the content width exceeds the explicit `width` or `max-width`, you get this:
image

So my supposition is that the only valid reason to set an explicit width is to add extra padding around the content. Until we change the overflow behavior to clip/ellipsize/shrink text, clients should never set a style that restricts the button from growing to fit its content. If that's the case, `min-width` achieves that but `width` and `max-width` don't.

I am open to holes being poked in that rationale!

fredvisser: I don't think adding width guidance in the Nimble Storybook adds _much_ value - ultimately, the patterns will drive how the button is used. E.g.

1. We have some [label guidance](https://stratus-storybook.ni.dev/iframe.html?viewMode=docs&id=patterns-links--docs#buttons-or-split-buttons) in the pattern docs, that should help keep labels short. If you follow the guidance, you won't need to worry about overflow.
2. For login pages or other places where button width grows to fill the view, we can document the best practice to use `min-width` within the pattern itself.

rajsite: on board with documenting via patterns 👍

rajsite (approved):

change/@ni-nimble-components-dbd8f845-12ff-44bd-b2de-86145d22572c.json
jattasNI: I guess I'd like a more holistic vision for Nimble styling docs. If we think this type of guidance belongs in patterns should we remove the Styling sections we have in Nimble storybook and populate the pattern backlog with work items / acceptance criteria to replace those docs?

Part of my struggle is that we're not currently prioritizing the work to fill out our patterns but I want an authoritative place to write down best practices so I can remember what they are and point to them in code reviews. I'll probably keep suggesting Nimble styling docs until I have another outlet for that.

Ultimately this is a bigger discussion than this PR and Fred has decision rights over the docs, so this PR is ready to 🚀

jattasNI (approved):

change/@ni-nimble-components-dbd8f845-12ff-44bd-b2de-86145d22572c.json
rajsite: > I'll probably keep suggesting Nimble styling docs until I have another outlet for that.

I think that's the right thing to do and it's the a that'll help crystalize it over time absent coordinated effort.

Issue #2379, opened 2024-09-04
[select] Selected option text content can get out of sync with displayed selected value


mollykreis:
# 🐛 Bug Report

When the text content of the selected item in the `nimble-select` is updated, the display value in the select isn't updated.

## 💻 Repro or Code Sample

1. Go to the `nimble-select` storybook page: https://nimble.ni.dev/storybook/?path=/docs/components-select--docs
2. In dev tools, update the text of the selected option
3. Notice that the text in the drop-down is updated but the text in the control is not

![image](https://github.com/user-attachments/assets/4b3b2948-ed3c-4ad1-a954-6a2feba9c757)

## 🤔 Expected Behavior

The display value in the `nimble-select` should always mirror the text content of the selected option.

## 😯 Current Behavior

The display value in the `nimble-select` can get out of sync with the text content of the selected option.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

m-akinc:
If it can't be fixed in a few hours, put it in the backlog

PR #2389, opened 2024-09-11
Select display value fix

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #2379

## 👩‍💻 Implementation

Added a `MutationObserver` to `ListOption` that will respond to `characterData` updates, and then emit an event that the `Select` now handles to update the display.

## 🧪 Testing

Added unit test (failed without fix). Manual verification.

## ✅ Checklist


change/@ni-nimble-components-56889b4d-259c-4017-b55f-97248f5f4db3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for selected option display text being updated in DOM not being reflected in display of select.",
```
mollykreis, line 3: Did you test whether or not filtering correctly handles the text changing in an option? I'm guessing not. If it is broken and it doesn't seem like it could be easily fixed within this PR, we should file a separate issue for it.

packages/nimble-components/src/list-option/index.ts
```diff
@@ -66,6 +66,13 @@ export class ListOption extends FoundationListboxOption {
         super.connectedCallback();
         const owner = this.getListOptionOwner();
         owner?.registerOption(this);
+        const observer = new MutationObserver(_ => {
+            this.$emit('display-changed');
+        });
+        observer.observe(this, {
```
mollykreis, line 72: Should you call `observer.disconnect()` in the `disconnectedCallback()`?

packages/nimble-components/src/list-option/index.ts
```diff
@@ -66,6 +66,13 @@ export class ListOption extends FoundationListboxOption {
         super.connectedCallback();
         const owner = this.getListOptionOwner();
         owner?.registerOption(this);
+        const observer = new MutationObserver(_ => {
```
rajsite, line 69: I'm pretty concerned about creating mutation observers for every option. Can we be more optimal like only for selected options? We are also emitting indiscriminately, I suspect that is a huge number of allocations and firing of junk events.

packages/nimble-components/src/list-option/index.ts
```diff
@@ -66,6 +66,13 @@ export class ListOption extends FoundationListboxOption {
...
+            this.$emit('display-changed');
```
mollykreis, line 70: I'm curious if you considered the alternative to emitting an event here of making `elementTextContent` observable. And if so, what the rationale was for choosing an event instead of an observable.

packages/nimble-components/src/list-option/index.ts
rajsite, line 72: This runs contrary to the above (because I want to avoid creating a huge number of unnecessary mutation observers) but just fyi FAST has a first class pattern for registering mutation observers via the `children` directive.

change/@ni-nimble-components-56889b4d-259c-4017-b55f-97248f5f4db3.json
atmgrifter00: It does. Ultimately when you filter it will re-query the `textContent` for each option, which will always retrieve what's in the DOM.

packages/nimble-components/src/list-option/index.ts
atmgrifter00: Updated to just observer the selected option.  Regarding emitting indiscriminately, I'll note that we only observe `characterData` changes on the option (and its basically non-existent subtree), and thus, I don't think the events will often fire, specifically for typical usages.

packages/nimble-components/src/list-option/index.ts
atmgrifter00: I've altered the implementation according to @rajsite's feedback, and I have included calling `disconnect` in the `disconnectedCallback` method for the `Select`.

packages/nimble-components/src/select/index.ts
```diff
@@ -1307,6 +1316,27 @@ export class Select
         this.setPositioning();
         this.focusAndScrollOptionIntoView();
     }
+
+    private observeSelectedOptionTextContent(): void {
+        if (this.selectedOptionObserver) {
+            this.selectedOptionObserver.disconnect();
+        }
+
+        if (this.selectedIndex === -1) {
```
mollykreis, line 1325: Do you need this check? Is there a case where `this.firstSelectedOption` would return a truthy value when `this.selectedIndex === -1`?

packages/nimble-components/src/select/tests/select.spec.ts
```diff
@@ -738,6 +738,22 @@ describe('Select', () => {
         await disconnect();
     });
 
+    it('when selected option textContent is updated directly in DOM, display text is updated', async () => {
```
mollykreis, line 741: Can you write one more test that verifies that this functionality works correctly if the display value wasn't directly text content (i.e. in a span instead)?

packages/nimble-components/src/select/index.ts
```diff
@@ -228,6 +229,13 @@ export class Select
         }
     }
 
+    public override disconnectedCallback(): void {
```
mollykreis, line 232: Don't you need to start observing again in the connected callback? If you disconnect and then reconnect without `selectedIndexChanged` being called, then the selected option wouldn't be observed anymore, right?

packages/nimble-components/src/list-option/index.ts
rajsite: I think we should try and frame all performance discussions for "typical usage" in terms of being used as part of virtualized scrolling in the table. In that case I'd expect character data in options to change quickly during scroll

packages/nimble-components/src/select/index.ts
```diff
@@ -228,6 +229,13 @@ export class Select
...
+        super.disconnectedCallback();
+        if (this.selectedOptionObserver) {
+            this.selectedOptionObserver.disconnect();
+        }
```
rajsite, line 236:
```suggestion
        this.selectedOptionObserver?.disconnect();
```

packages/nimble-components/src/select/index.ts
```diff
@@ -1307,6 +1316,27 @@ export class Select
         this.setPositioning();
         this.focusAndScrollOptionIntoView();
     }
+
+    private observeSelectedOptionTextContent(): void {
+        if (this.selectedOptionObserver) {
+            this.selectedOptionObserver.disconnect();
+        }
```
rajsite, line 1323:
```suggestion
        this.selectedOptionObserver?.disconnect();
```

packages/nimble-components/src/select/index.ts
atmgrifter00: This is actually just handling a special case for when the `Select` initializes, whereby it sets `selectedIndex` to '-1'. At this point, the `selectedOptions` for the `Select` will be `undefined`, which without that check will cause an access of `firstSelectedOption` to throw.

packages/nimble-components/src/list-option/index.ts
atmgrifter00: Point taken. Even so, I think the only extra optimization I could do is to create a new update tracker that would queue display updates. Not convinced the potential scenario of a a virtualized table would even demand such an optimization.

packages/nimble-components/src/select/index.ts
atmgrifter00: Yep, added a test for this as well.

packages/nimble-components/src/select/tests/select.spec.ts
atmgrifter00: Note that this required setting `childList: true` in the options for the mutation observer.

packages/nimble-components/src/list-option/index.ts
rajsite: The current PR changes look reasonable

rajsite (approved):

packages/nimble-components/src/select/index.ts
```diff
@@ -1307,6 +1316,26 @@ export class Select
...
+
+        if (this.selectedIndex === -1) {
+            return;
+        }
+
+        const selectedOption = this.firstSelectedOption;
+        if (selectedOption) {
+            this.selectedOptionObserver = new MutationObserver(() => {
+                this.updateDisplayValue();
+            });
```
rajsite, line 1331:
```suggestion
          if (!this.selectedOptionObserver) {
              this.selectedOptionObserver = new MutationObserver(() => {
                  this.updateDisplayValue();
              });
          }
```

rajsite: Or instead of constructing it lazily we could construct it once as a readonly prop. Guessing most select's will have an option marked selected so probably not avoiding much making it lazy

mollykreis (approved):

packages/nimble-components/src/select/index.ts
```diff
@@ -216,6 +216,14 @@ export class Select
     private _value = '';
     private forcedPosition = false;
     private openActiveIndex?: number;
+    private readonly selectedOptionObserver: MutationObserver;
```
rajsite, line 219: Betting the PR failures have to do with constructor timing (propertyChange callbacks for initialized props run before contructors). Need to capture notes on it but instead initialize here:

```
private readonly selectionOptionsObserver = new MutationObserver(() => {this.updateDisplayValue()});
```

change/@ni-nimble-components-56889b4d-259c-4017-b55f-97248f5f4db3.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: fyi this PR is still in draft, guessing that's not expected

packages/nimble-components/src/select/index.ts
atmgrifter00: I actually had to make its type `MutationObserver | undefined`, even when defined in this way, because of the issue you raise. Which also means using the null check operator where I access `selectedOptionObserver`.

packages/nimble-components/src/select/index.ts
```diff
@@ -216,6 +216,9 @@ export class Select
     private _value = '';
     private forcedPosition = false;
     private openActiveIndex?: number;
+    private readonly selectedOptionObserver? = new MutationObserver(() => {
```
jattasNI, line 219: I might be missing something but why does this type need to handle the mutation observer possibly being `undefined`?

packages/nimble-components/src/select/index.ts
```diff
@@ -736,6 +746,7 @@ export class Select
             this.setActiveOption(this.selectedIndex);
         }
         this.updateValue();
+        this.observeSelectedOptionTextContent();
```
jattasNI, line 749: Can you talk about the decision to put this call in `selectedIndexChanged()` rather than, say `updateValue()`? It looks like `updateValue()` is called in a bunch of other scenarios like slotted options changing and clearing the selection. Should we stop observing immediately in those cases too?

packages/nimble-components/src/select/index.ts
atmgrifter00: It's because the `selectedIndexChanged` callback will run prior to the constructor or even the assignment of this field, and at that time `selectedOptionObserver` is `undefined`. While I admit this looks a bit goofy, it's at least declaratively honest. The only way I could see to make this less "goofy" looking is to just declare it here, and then define it in `connectedCallback`. I don't have a strong opinion.

packages/nimble-components/src/select/index.ts
atmgrifter00: Conceptually, the idea is that we just want to react to the text content changing of the current selected option. This only needs to be changed (i.e. we need to observe a new option) when the actual selection changes, not when the display of the current selected option changes.

packages/nimble-components/src/select/index.ts
jattasNI: Makes sense, I'm fine to leave it as-is now that I understand the rationale.

packages/nimble-components/src/select/index.ts
atmgrifter00: While the corner-case you mention may have merit, to answer your larger question, I don't think this is a concern. The mutation observer doesn't react to synchronous changes to the DOM. Note that the only time calling `updateValue` results in the observed option becoming no longer selected is if the `selectedIndex` has been changed to -1, in which case there will be no option to observe, and the observer will be disconnected from the previous option (again before it ever bothers trying to update the display value, and even if it did it would be innocuous).

All that said, because of the corner-case you mention, I think calling `observerSelectedOptionTextContent` from within `updateValue` is a valid solution, and I'm not concerned with calling that method unnecessarily at times.

atmgrifter00: I've updated the implementation and added a test to cover the corner-case you mention (failed without the change).

jattasNI (approved):

PR #2391, opened 2024-09-11
Add usage guidance for select vs combobox vs radio

jattasNI:
# Pull Request

## 🤨 Rationale

I needed an example of component usage guidance for an NITech presentation

## 👩‍💻 Implementation

Capture my understanding of when to use select, combobox, and radio in each of their MDX files.

I moved the combobox description from storybook to mdx because it was annoying to format large markdown strings from code.

Updated language for existing usage guidance to match new pattern.
   - my secret motivation is working around a prettier bug that puts tags near the end of a line in a new paragraph, leading to ugly docs. I couldn't find a syntax to `prettier-ignore` in mdx files. [nor can](https://github.com/prettier/prettier/issues/12209) [other people](https://stackoverflow.com/questions/69311134/prettier-how-to-ignore-a-single-line)

image

## 🧪 Testing

Local storybook

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/anchor-button/anchor-button.mdx
+++ b/packages/storybook/src/nimble/anchor-button/anchor-button.mdx
@@ -20,9 +20,8 @@ import * as anchorButtonStories from './anchor-button.stories';
 An anchor button is a component with the visual appearance of a button, but it
 navigates like an anchor/link when pressed.
 
-If you want a button that triggers an action or event, use the
-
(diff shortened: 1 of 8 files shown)


rajsite (approved):

atmgrifter00 (approved):

rajsite:
@fredvisser straightforward change to docs that I don't think are controversial. Bypassing but if concerns can address in follow-up.

PR #2392, opened 2024-09-11
Fix two sources of references to detached rows

m-akinc:
# Pull Request

## 🤨 Rationale

While investigating a [memory leak issue](https://ni.visualstudio.com/DevCentral/_workitems/edit/2843309) reported in AzDO, we found a couple places in the Nimble table code that wasn't cleaning up observers/notifiers, causing us to hold references to disconnected row elements.

There is also a FAST bug that contributes to the leak (which is [getting fixed](https://github.com/microsoft/fast/pull/7023)), so the leak will still not be fixed until this PR and the FAST one are both in.

## 👩‍💻 Implementation

1. Rows observe columns, which results in there being a reference chain from each column to each row. When the row is detached, we must remove the column observers.
2. The keyboard navigation manager has logic for storing the row notifiers it creates, so that it can clean them up at the appropriate time. However, by a simple oversight, we were never actually adding the notifiers to the array.

## 🧪 Testing

Verified (in Chrome dev tools) that after these changes, detached rows are no longer referenced via the columns or keyboard navigation manager.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b9a584ff-3c6d-46d3-afde-b9af6c45ec0e.json
+++ b/change/@ni-nimble-components-b9a584ff-3c6d-46d3-afde-b9af6c45ec0e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix two sources of references to detached rows",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table/components/row/index.ts
+++ b/packages/nimble-components/src/table/components/row/index.ts
@@ -170,6 +170,11 @@ export class TableRow<
         return null;
     }
 
+    public override disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.removeColumnObservers();
+    }
+
     /** @internal */
     public onSelectionCheckboxChange(event: CustomEvent): void {
         if (this.ignoreSelectionChangeEvents) {
--- a/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
+++ b/packages/nimble-components/src/table/models/keyboard-navigation-manager.ts
@@ -171,6 +171,7 @@ implements Subscriber {
             for (const visibleRow of this.table.rowElements) {
                 const rowNotifier = Observable.getNotifier(visibleRow);
                 rowNotifier.subscribe(this, 'resolvedRowIndex');
+                this.visibleRowNotifiers.push(rowNotifier);
                 if (visibleRow.resolvedRowIndex === this.rowIndex) {
                     focusRowAndCell = true;
                 }


m-akinc:
@atmgrifter00 would you please buddy?

atmgrifter00 (approved):

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -170,6 +170,11 @@ export class TableRow<
         return null;
     }
 
+    public override disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.removeColumnObservers();
```
rajsite, line 175: From a quick (incomplete) search I found at least one other spot we are failing to clean-up notifiers in the disconnected callback: https://github.com/ni/nimble/blob/c4470de36c67bd8b08155fe4dedb1fce96972500/packages/nimble-components/src/table-column/enum-base/index.ts#L127

We should create a tech debt issue to evaluate our `getNotifier` & `subscribe` usages to make sure they clean-up correctly and address in a separate PR.

change/@ni-nimble-components-b9a584ff-3c6d-46d3-afde-b9af6c45ec0e.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @m-akinc @atmgrifter00 
Captured the following `Common items caught in PR review` topic on avoiding memory leaks: https://github.com/ni/nimble/issues/1479#issuecomment-2344585509

Feel free to add relevant details or other patterns that we should watch for.

rajsite (approved):

packages/nimble-components/src/table/components/row/index.ts
```diff
@@ -170,6 +170,11 @@ export class TableRow<
         return null;
     }
 
+    public override disconnectedCallback(): void {
```
jattasNI, line 173: Do we need to consider the lifecycle where a row element is disconnected and then reconnected? If yes, what would cause the columns to be re-observed in that case? Would `columnsChanged()` be called again?

jattasNI (approved):

packages/nimble-components/src/table/components/row/index.ts
m-akinc: Created #2393

packages/nimble-components/src/table/components/row/index.ts
rajsite: Good point, didn't think through it but if discussion about thinking of the full connect / disconnect lifecycle is important that should be captured in the common PR items memory leaks comment: https://github.com/ni/nimble/issues/1479#issuecomment-2344585509

m-akinc: I don't think we need to support that, given that `nimble-table-row` is just an implementation detail of the `nimble-table` (not exposed via the table's public API) and not otherwise intended for use by clients.

`columnsChanged()` does not get called upon re-connecting a row to the DOM. There could be other state in the `TableRow` (e.g. `columns`) that went stale while the component was disconnected, so supporting disconnect/reconnect would take more than simply adding a call to `observeColumns()` in `connectedCallback`.

rajsite:
@m-akinc thinking back historically on changes we have made to FAST Foundation and then having discovered:

- It fails their tests (which don't run automatically)
- It causes some other issue in full integration (like how we found the chrome crashes)

For the two PRs that are being merged into FAST archives:

https://github.com/microsoft/fast/pull/6960
https://github.com/microsoft/fast/pull/7023

- Do we feel confident that their tests pass and won't be an issue for releasing?
- Have we done a full test with local tgz builds in nimble and as a full integration in an SLE app (manual and CI testing)? Given that the memory leak is blocking table adoption I want us to do our best to avoid needing additional update cycles in FAST.

m-akinc:
@rajsite I think that has been done for the older fix, but not for the memory leak issue. The memory leak fix ought to have very limited ability to affect anything in an unexpected way, unless `unbind()` is called (unexpectedly?) at a time other than the node disconnecting, or if the binding observers have a bad reference, or something like that. Still, it's probably a good idea to do at least a quick sanity check, so I'll do that.

Issue #2019, opened 2024-04-16
Error state for nimble-radio-group


fredvisser:
# 💡 Updated Component

## 😯 Problem to Solve

We need to finish out our first-class form support for our core input components by supporting error states and text for all components commonly used in forms.

## 💁 Proposed Solution

Leverage the existing patterns/APIs to configure error state on `nimble-radio-group`

I.e. Add `error-visible` and `error-text` properties to the `nimble-radio-group`.

![Screenshot 2024-04-16 at 3 44 26 PM](https://github.com/ni/nimble/assets/1458528/92c5dbc1-2b27-40b4-8944-c2f61cfd3d46)

## 📋 Tasks

- [ ] ~~nimble-radio-group interaction design issue~~
- [ ] ~~nimble-radio-group interaction design spec~~
- [x] nimble-radio-group [visual design issue](https://github.com/ni/nimble/issues/1765)
- [x] nimble-radio-group [visual design spec](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-63148&mode=design)
- [ ] Update nimble-radio-group spec
- [ ] Update nimble-radio-group custom element
- [ ] Update nimble-radio-group unit tests
- [ ] Update nimble-radio-group Storybook tests
- [ ] Update nimble-radio-group Storybook docs
- [ ] Update nimble-radio-group Angular wrapper
- [ ] Update nimble-radio-group Blazor wrapper
- [ ] Update Angular example app
- [ ] Update Blazor example app

Issue #2018, opened 2024-04-16
Error state for nimble-checkbox


fredvisser:
# 💡 Updated Component

## 😯 Problem to Solve

We need to finish out our first-class form support for our core input components by supporting error states and text for all components commonly used in forms.

## 💁 Proposed Solution

Leverage the existing patterns/APIs to configure error state on `nimble-checkbox`

I.e. Add `error-visible` and `error-text` properties to the `nimble-checkbox`.

![image](https://github.com/ni/nimble/assets/1458528/7b9eaebf-b3bc-46ae-9ef1-06ac325e6d62)

## 📋 Tasks

- [ ] ~~nimble-checkbox interaction design issue~~
- [ ] ~~nimble-checkbox interaction design spec~~
- [x] nimble-checkbox [visual design issue](https://github.com/ni/nimble/issues/1765)
- [x] nimble-checkbox [visual design spec](https://www.figma.com/file/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?type=design&node-id=1295-63148&mode=design)
- [x] Update nimble-checkbox [spec](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/checkbox/specs/error-state.md)
- [x] Update nimble-checkbox custom element
- [ ] Update nimble-checkbox unit tests
- [x] Update nimble-checkbox Storybook tests
- [x] Update nimble-checkbox Storybook docs
- [ ] Update nimble-checkbox Angular wrapper
- [ ] Update nimble-checkbox Blazor wrapper
- [ ] Update Angular example app
- [ ] Update Blazor example app

Issue #610, opened 2022-06-20
Configurable / Additional heights for nimble controls


rajsite:
## 🧹 Tech Debt / Enhancement

Currently the height of nimble controls is defined by the `--ni-nimble-control-height` css custom property. Our control styles are written in such a way that they are dependent on `--ni-nimble-control-height` custom property such as with fixed css `calc()` statements that use the `--ni-nimble-control-height` property. This is further discussed in the "Arbitrary Sizing" section.

In addition, we currently do not support most of the additional component heights shown in Figma Specs. This is further discussed in the "Predesigned Sizing" section.

## Arbitrary Sizing

The design of the height of nimble controls being defined by the '--ni-nimble-control-height` css custom property may be unexpected for users as, for example, a user may expect to set the height via css of a control:

```html
<nimble-some-control style="height:50px"></nimble-some-control>
```

However, this may result in issues as by manually changing the height of the control, the `--ni-nimble-control-height` constant will not update in response and layout calculations using those values will layout incorrectly.

If a user really required changing the height of a control they have to do it via the `--ni-nimble-control-height` property:

```html
<nimble-some-control style="--ni-nimble-control-height: 50px"></nimble-some-control>
```

Some options to consider:

1) We state firmly that nimble controls have a fixed height that is not intended to be changed by any client. In that case we should treat it like padding and make it more difficult to override control height: https://github.com/ni/nimble/issues/598. We should also consider removing the token / marking the token internal (`--ni-nimble-internal-control-height`)
2) We allow the controls to have user-defined heights. The `--ni-nimble-control-height` token is only used for setting the default height of controls and is not used in CSS layout calculations. CSS is modified to avoid requiring `calc()` statements using `--ni-nimble-control-height`.
3) [not recommended] We support users configuring the height on controls using the `--ni-nimble-control-height` design token. I do not think we should encourage users to use a CSS property for sizing controls that is proprietary to the nimble design system. We should discourage / forbid users setting `--ni-nimble-control-height` directly.

## Predesigned Sizing

[...]

PR #2394, opened 2024-09-11
HLD for adding error state to checkbox and radio group

mollykreis:
# Pull Request

## 🤨 Rationale

HLD for adding error state to checkbox and radio group

This is the HLD for #2018 and #2019

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
+++ b/packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
@@ -47,7 +47,7 @@ In order to accomplish these changes, the `nimble-radio-group` template and the 
 
 #### Checkbox and Radio styling updates
 
-Both the `nimble-checkbox` and `nimble-radio` need some styling updates to align them with the current visual designs. These components have very similar styling needs, so there is likely opportunity to move much of the styles into a shared pattern. This will be considered during implementation.
+Both the `nimble-checkbox` and `nimble-radio` need styling updates to align them with the current visual designs. These components have very similar styling needs, so there is likely opportunity to move much of the styles into a shared pattern. This will be considered during implementation.
 
 The styling changes include (but may not be limited to):
 
@@ -56,9 +56,9 @@ The styling changes include (but may not be limited to):
     -   Verifying the label wraps as expected
     -   Ensuring the alignment of the checked/selected indicator is center aligned with the first line of text
 
-#### Other control updates
+#### Text field updates
 
-The `error-visible` visual design of other controls has been updated. Specifically, the spacing between the right edge of the control and the error icon has been updated to be 4px consistently across controls. This update impacts the `nimble-text-field`. There is already [a nimble issue](https://github.com/ni/nimble/issues/2090) for the spacing being incorrect on this control, so the work is already tracked. Ideally, this work would be done as part of the effort to add error states to the `nimble-checkbox` and `nimble-radio-group` so that the error icons are aligned across the controls when they are stacked.
+The visual design of the `nimble-text-field` has been updated such that the error icon is 4px from the right edge of the control rather than 8px. There is already [a nimble issue](https://github.com/ni/nimble/issues/2090) for the spacing being incorrect on this control, so the work is already tracked. Ideally, this work would be done as part of the effort to add error states to the `nimble-checkbox` and `nimble-radio-group` so that the error icons are aligned across the controls when they are stacked.
 
 ## Open Issues
 
--- a/packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
+++ b/packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
@@ -6,15 +6,14 @@ The `nimble-checkbox` and the `nimble-radio-group` need to have error states ass
 
 ### Background
 
-- [Visual design spec](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-63148&node-type=canvas&t=sJ2Fhm1vLoZ4zpsK-0)
-- [Nimble issue 2018: Error state for nimble-checkbox](https://github.com/ni/nimble/issues/2018)
-- [Nimble issue 2019: Error state for nimble-radio-group](https://github.com/ni/nimble/issues/2019)
+-   [Visual design spec](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-63148&node-type=canvas&t=sJ2Fhm1vLoZ4zpsK-0)
+-   [Nimble issue 2018: Error state for nimble-checkbox](https://github.com/ni/nimble/issues/2018)
+-   [Nimble issue 2019: Error state for nimble-radio-group](https://github.com/ni/nimble/issues/2019)
 
 Related issues that can and/or should be resolved as part of this effort:
 
-- [Nimble issue 2091: nimble-checkbox has incorrect height](https://github.com/ni/nimble/issues/2091)
-- [Nimble issue 2090: Error icon is incorrectly spaced in nimble-text-field](https://github.com/ni/nimble/issues/2090)
-
+-   [Nimble issue 2091: nimble-checkbox has incorrect height](https://github.com/ni/nimble/issues/2091)
+-   [Nimble issue 2090: Error icon is incorrectly spaced in nimble-text-field](https://github.com/ni/nimble/issues/2090)
 
 ## Design
 
@@ -26,8 +25,8 @@ Both the `nimble-checkbox` and the `nimble-radio-group` will be updated to be ab
 
 The `nimble-checkbox` will be updated to implement the `ErrorPattern` interface. This means it will have two new attributes:
 
-- `error-text` - string
-- `error-visible` - boolean
+-   `error-text` - string
+-   `error-visible` - boolean
 
 The `error-visible` boolean will control whether or not a red exclamation icon will be visible next to the checkbox label and whether or not the `error-text` is displayed below the checkbox label.
 
@@ -35,8 +34,8 @@ The `error-visible` boolean will control whether or not a red exclamation icon w
 
 The `nimble-radio-group` will be updated to implement the `ErrorPattern` interface. This means it will have two new attributes:
 
-- `error-text` - string
-- `error-visible` - boolean
+-   `error-text` - string
+-   `error-visible` - boolean
 
 The `error-visible` boolean will control whether or not a red exclamation icon will be visible next to each radio label and whether or not the `error-text` is displayed below all the radio controls.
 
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
```diff
@@ -0,0 +1,65 @@
+# Error State on Checkbox and Radio Group
```
mollykreis, line 1: Note: I wasn't sure the best place to put this HLD since is spanned 2 components. Therefore, I put it in the error pattern directory. I'm open to alternatives.

mollykreis:

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
m-akinc: I had similar uncertainty about where to put the "required input" HLD, as it also affects multiple components, and the pattern would potentially apply to future components in `spright-components`, too. I started my HLD at `nimble/specs/required-inputs/README.md`, but I'm not sure if that's a better place than you've chosen. (The only other (non-template) specs there are for `labels-and-localization` and `nimble-blazor`.)

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
```diff
@@ -0,0 +1,65 @@
...
+-   [Visual design spec](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-63148&node-type=canvas&t=sJ2Fhm1vLoZ4zpsK-0)
+-   [Nimble issue 2018: Error state for nimble-checkbox](https://github.com/ni/nimble/issues/2018)
+-   [Nimble issue 2019: Error state for nimble-radio-group](https://github.com/ni/nimble/issues/2019)
+
+Related issues that can and/or should be resolved as part of this effort:
+
+-   [Nimble issue 2091: nimble-checkbox has incorrect height](https://github.com/ni/nimble/issues/2091)
+-   [Nimble issue 2090: Error icon is incorrectly spaced in nimble-text-field](https://github.com/ni/nimble/issues/2090)
+
+## Design
```
m-akinc, line 18: What is the expectation/guidance for clients in terms of setting widths of checkboxes or radio groups? Currently, radio buttons in a group do not expand to the width of the group, so (in a vertical orientation) error icons would not line up unless the labels were all the same length. Will we automatically change `width` to `100%` when the button is within a group? Will we have guidance (or styling rules) to ensure the radio group width allows for some space between the longest radio button label and the error icon? Is it up to a client to ensure radio groups are wide enough that button labels never wrap? Otherwise, how does label wrapping affect icon placement?

What about radio groups with a horizontal orientation? Do we still put an error icon after each button label, and what spacing do we use?

For checkboxes, some of the same questions apply. Should the width default to `100%`? Is it okay if the client doesn't provide an error string, and the width of the checkbox results in huge separation between the icon and the label? E.g.
image

What is the wrapping behavior of the error text?

m-akinc, line 18: Currently, both checkboxes and radio buttons respond to clicks (and style for hover) anywhere in the component. If we change the widths to `100%`, that will pose a problem, as (I believe) ARIA guidelines say that only the control and its label should respond to clicks (not blank areas within the component's bounds). I think the same applies to hover styling.

Should have some notes about any changes that we'll need to make.

mollykreis: A lot of this information has now been incorporated into the HLD, but here are some answers:

> What is the expectation/guidance for clients in terms of setting widths of checkboxes or radio groups?

For the checkbox and vertical radio group, the expectation is that a client would set a width on it, just like the client would for any other control. I'm not entirely sure what the expectation is for a horizontal radio group. I don't believe we currently have any scenarios where that is the recommended orientation.

> Will we automatically change width to 100% when the button is within a group?

Yes, when the radio buttons are vertically oriented, they will be styled to have the same width as the group. I've updated the HLD.

> Will we have guidance (or styling rules) to ensure the radio group width allows for some space between the longest radio button label and the error icon?

The space between the label and the error icon will be part of the component's styling. The client won't have to do anything to accomplish this.

> Is it up to a client to ensure radio groups are wide enough that button labels never wrap? Otherwise, how does label wrapping affect icon placement?

The labels are allowed to wrap. The icon will always be center aligned with the first row of the text.

> What about radio groups with a horizontal orientation? Do we still put an error icon after each button label, and what spacing do we use?

There will still be an error icon after each button label using the same spacing. I think we need guidance in general about when to use the horizontal orientation of the radio buttons, but I don't think that has to be part of this effort of adding error states.

>What is the wrapping behavior of the error text?

As with the error text in other components, it does not wrap.

mollykreis: I've updated the HLD with these details, including some screenshots that show where the interactable part of the control is.

m-akinc (approved):
Love all the example images! They make everything very clear.

atmgrifter00 (approved):

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
jattasNI: I think both locations are sane and in line with our guidance and don't have a strong preference between them so I'm fine to leave each PR as-is.

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
```diff
@@ -0,0 +1,103 @@
...
+The `error-visible` boolean will control whether or not a red exclamation icon will be visible next to the checkbox label and whether or not the `error-text` is displayed below the checkbox label.
+
+#### Radio Group
+
+The `nimble-radio-group` will be updated to implement the `ErrorPattern` interface. This means it will have two new attributes:
+
+-   `error-text` - string
+-   `error-visible` - boolean
+
+The `error-visible` boolean will control whether or not a red exclamation icon will be visible next to each radio label and whether or not the `error-text` is displayed below all the radio controls.
```
jattasNI, line 73: Did we have discussions about whether the exclamation should appear on all the radio labels? My initial reaction from the screenshot with 3 radio buttons is that it looks like 3 separate errors even though there's just one error for the whole control.

It's possible this is the best we can do:
1. If we put the exclamation next to the first or last option I'd probably complain that it looks like only that option has an error. 
2. I don't think there's enough space to put it next to the label.
3. We could decide not to show the exclamation on the radio group and just rely on the error text, but that's inconsistent with other controls.

Just want to raise it in case there's rationale for this approach that we should capture and to see if anyone else has opinions.

jattasNI (approved):

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
mollykreis: I believe these alternatives were considered as part of creating the visual design. Perhaps @fredvisser and @NIbokeefe have some context they can provide here.

rajsite: > looks like 3 separate errors even though there's just one error for the whole control.

+1 for that concern. I also expect the interaction would feel pretty jarring for all of the indicators to appear / disappear at the same time.

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
```diff
@@ -0,0 +1,103 @@
...
+
+While `error-visible` will be a new public attribute on the `nimble-radio-group`, it will also be added as an internal attribute on the `nimble-radio`, which will be responsible for displaying the error icon next to its label when that attribute is set. The `nimble-radio-group` will be responsible for keeping the `error-visible` state of each child `nimble-radio` instance synchronized with its own `error-visible` state. It will do this by updating `error-visible` on each `nimble-radio` any time its own `error-visible` state changes or any time its slotted radios changes.
```
rajsite, line 75: > it will also be added as an internal attribute

Maybe a nit, but particularly if it's internal maybe we can be a property instead of an attribute. But I can easily imagine styling being based off the attribute so in  that case 100% internal attribute makes sense.

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Additional Work
+
+#### Checkbox and Radio styling updates
+
+Both the `nimble-checkbox` and `nimble-radio` need styling updates to align them with the current visual designs. These components have very similar styling needs, so there is likely opportunity to move much of the styles into a shared pattern. This will be considered during implementation.
+
+The styling changes include (but may not be limited to):
+
+-   Updating the control height to respect the `controlHeight` token rather than being sized to the text and/or checked/selected indicator
```
rajsite, line 91: When we leverage the token I want to encourage us to use it for default value configuration and not as part of calc expressions, etc. The intention is that normal css props like height are what we are configured off of and we aren't inventing a new api for sizing controls, we just have tokens for defaults. i.e. option 2 in this discussion https://github.com/ni/nimble/issues/610#issue-1277404188

rajsite (approved):

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
mollykreis: @fredvisser or @NIbokeefe, can you weigh in here on more of the rationale of the design and whether or not we want to revisit the current visual design?

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
mollykreis: I made it an attribute so that it can be styled off of, particularly using the [existing error pattern styles](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/patterns/error/styles.ts) that expects an `error-visible` attribute to control the visibility of the error.

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
mollykreis: That's an interesting point. In my current prototype, I have one `calc` using the control height because the top/bottom padding of the control changes based on the control height. Therefore, that would need to change if we wanted to avoid any `calc` statements with the height, but the replacement css is alluding me at the moment.

rajsite: I vaguely remember similar concerns for the spinner and I think that is where this comes from:
https://github.com/ni/nimble/blob/b338c2500cfe0974feeb3b00eb3e5a92cffdf227/packages/nimble-components/src/spinner/styles.ts#L25

Don't think it blocks the HLD but fair warning (without really understanding the concerns) that we may end up deviating from pixel-perfect spacing to accommodate whole pixel alignment on grids to good simplifying approximations.

packages/nimble-components/src/patterns/error/specs/checkbox-radio-error.md
mollykreis: This is no longer applicable. There are now 2 separate HLDs in this PR.

jattasNI (approved):

Issue #2100, opened 2024-05-13
Required State for all input controls


fredvisser:
# 🙋 Feature Request

## 😯 Problem to Solve

![image](https://github.com/user-attachments/assets/a6751718-f569-4463-8dcf-345a250e7510)

Input controls are often used in forms or property editor dialogs. Often the user is required to provide an input to the control before the form can be submitted or before the property changes can be saved.

With Nimble controls today, client applications will manually add … (Required) to the control's placeholder text or to the label.

We need to provide a standardized way for the application to let the end-user know that a control is required.

Input controls:
- Combobox
- Number Field
- Radio Group
- Select
- Text Area
- Text Field

Don't need to support:
- Checkbox
- Switch button

## 💁 Proposed Solution

https://www.w3schools.com/tags/att_input_required.asp

Per the [visual design](https://github.com/ni/nimble/issues/2072), we should add a `required` attribute to each input control that enables or disables the display of the visual state.

Need compatibility with existing [Angular form validation patterns](https://angular.io/guide/form-validation).

## 📋 Tasks

- [x] Visual design [link](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-47481)
- [ ] input control custom element changes
- [ ] input control unit tests
- [ ] input control Storybook tests
- [ ] input control Storybook docs
- [ ] input control Angular wrapper
- [ ] input control Blazor wrapper
- [ ] Add to Angular example app
- [ ] Add to Blazor example app

jattasNI:
This is available for someone to start researching. The ultimate goal is an HLD. That will likely be quick to research and write but if not, Fred would appriciate an earlier high-level estimate once we have an initial idea of how hard it will be.

NIbokeefe:
@fredvisser @jattasNI  @rajsite 

The request from visual design was to convert the Asterisk symbol from a font character into a vector path icon. 

-The icon is to be cropped to size 
-Provide alignment spec


Alignment of Asterisk with control label:

![Image](https://github.com/user-attachments/assets/b7b7e813-043c-408e-8336-3f0115623b9a)

**asterisk_5x5**
Tag: star, required, 

![Image](https://github.com/user-attachments/assets/8a6fa54c-249c-4218-9044-0d13f49a1a2b)

m-akinc:
The Nimble support for this is complete, thanks to Molly.

PR #2395, opened 2024-09-12
Create Required Inputs HLD

m-akinc:
# Pull Request

## 🤨 Rationale

HLD for #2100

## 👩‍💻 Implementation

Wrote HLD.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/specs/required-inputs/README.md
+++ b/specs/required-inputs/README.md
@@ -1,0 +1,66 @@
+# Required Input Controls
+
+## Problem Statement
+
+Input controls are often used in forms or property editor dialogs. Often the user is required to provide an input to the control before the form can be submitted or before the property changes can be saved.
+
+We need to provide a standardized way for the application to let the end-user know that a control is required.
+
+Initially, we need this for:
+- `nimble-combobox`
+- `nimble-number-field`
+- `nimble-radio`/`nimble-radio-group`
+- `nimble-select`
+- `nimble-text-area`
+- `nimble-text-field`
+
+## Links To Relevant Work Items and Reference Material
+
+- [Nimble issue](https://github.com/ni/nimble/issues/2100)
+- [AzDO Feature](https://ni.visualstudio.com/DevCentral/_workitems/edit/2732543)
+- [Figma design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-47481)
+
+## Implementation / Design
+
+### API
+
+Our FAST base components provide built-in forms support via the [same API](https://www.w3schools.com/tags/att_input_required.asp) exposed by the native HTML `input`:
+
+- `required`: boolean attribute whose presence indicates that a value must be provided to submit
+
+ For radio buttons/groups, only the button elements expose the `required` attribute. If any of the radio buttons in a group are marked required, the group is treated as required.
+
+### Visuals
+
+We will update our templates so that setting `required` causes a red asterisk to be displayed after the label, as per the visual design. For radio buttons/groups, the asterisk is on the group label, not individual button labels.
+
+### Form validation
+
+#### Vanilla HTML
+
+We would expect this to already work, since the FAST components we derive from provide forms support. Unfortunately, this support is broken for `nimble-combobox` and `nimble-radio`.
+
+- `nimble-combobox`: All that is needed is to forward the `required` attribute to the `input` in the template (which we have already forked from FAST). We will make this change.
+- `nimble-radio`:
+    - `required` on a radio button does not take other buttons in the group into account. It will result in a validation error if _that specific_ radio is unchecked.
+    - Validity is only updated when the value or checked state changes, so if the radio button is initially unchecked and remains unchecked, it will remain marked valid, even though it violates the `required` constraint.
+
+Because the `nimble-radio` support has never worked, would take significant work to fix, and FAST is no longer accepting submissions to the archive branch we use, we **will not fix radio button forms support** as part of this feature.
+
+Forms support already works for `nimble-number-field`, `nimble-select`, `nimble-text-area`, and `nimble-text-field`.
+
+#### Angular
+
+The `required` attribute only plays a role in validation for template-driven forms (with reactive forms, it is only used for accessibility purposes). Nimble inputs already support setting the `required` attribute via template, but we will add it to the directive wrappers for completeness.
+
+#### Blazor
+
+Typically, form validation is handled via a `DataAnnotationsValidator`, and an input is treated as required if it is bound to a model property annotated with `RequiredAttribute` (i.e. `[Required]`). The Nimble component's `required` attribute is still needed to turn on the visual affordance, though, so we will add it to the Razor API.
+
+## Alternative Implementations / Designs
+
+None
+
+## Open Issues
+
+None


m-akinc:
@msmithNI do you mind buddying?

msmithNI (approved):

atmgrifter00 (approved):

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Implementation / Design
+
+### API
+
+Our FAST base components provide built-in forms support via the [same API](https://www.w3schools.com/tags/att_input_required.asp) exposed by the native HTML `input`:
+
+- `required`: boolean attribute whose presence indicates that a value must be provided to submit
+
+ For radio buttons/groups, only the button elements expose the `required` attribute. If any of the radio buttons in a group are marked required, the group is treated as required.
```
mollykreis, line 31: Can you clarify this statement a bit to make sure different corner cases are considered? I'm specifically thinking of how attributes such as `hidden` or `disabled` on the `required` radio will affect whether or not the group is required. For example, is the group required if a `hidden` radio is marked as `required`?

mollykreis (approved):

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+#### Blazor
+
+Typically, form validation is handled via a `DataAnnotationsValidator`, and an input is treated as required if it is bound to a model property annotated with `RequiredAttribute` (i.e. `[Required]`). The Nimble component's `required` attribute is still needed to turn on the visual affordance, though, so we will add it to the Razor API.
+
+## Alternative Implementations / Designs
+
+None
+
+## Open Issues
+
```
jattasNI, line 65: This is out of scope for this HLD, but reminder that as part of this feature we'll want to consider the impact on clients and who will do the work to uptake this. It looks like [there are lots of apps using a "(required)" note](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=required%20path%3Aweb/workspaces%20ext%3Axlf&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/SystemsManagement/projects/systems/src/locales/messages.de.xlf) on their label and I assume we'll want to migrate those to use this. Could you make sure there's a user story or feature acceptance criteria tracking this?

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+None
```
jattasNI, line 66: Borderline on whether this needs to be in the HLD, but what is the impact on:
1. testing. Will this be a new matrix test state or can we get away with just one new test? I'm guessing the former since we'd want to test themes and interaction with other state like very long labels or readonly attributes.
2. documentation. Will we just document the `required` attribute on each component? Or add any new higher-level guidance?

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+### API
+
+Our FAST base components provide built-in forms support via the [same API](https://www.w3schools.com/tags/att_input_required.asp) exposed by the native HTML `input`:
+
+- `required`: boolean attribute whose presence indicates that a value must be provided to submit
+
+ For radio buttons/groups, only the button elements expose the `required` attribute. If any of the radio buttons in a group are marked required, the group is treated as required. It does not matter if the button with `required` is disabled, hidden, etc. It is effectively as if the attribute is on the group of buttons, rather than a specific one. This matches the behavior of the native `input` with `type="radio"`. 
+
+### Visuals
```
jattasNI, line 33: [This note on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required#attribute_interactions) suggests that for inputs marked `readonly`, `required` should have no effect. Would we hide the asterisk if both attributes are present?

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
+# Required Input Controls
+
+## Problem Statement
+
+Input controls are often used in forms or property editor dialogs. Often the user is required to provide an input to the control before the form can be submitted or before the property changes can be saved.
+
+We need to provide a standardized way for the application to let the end-user know that a control is required.
+
+Initially, we need this for:
```
jattasNI, line 9: For completeness I'll note that [`aria-required` is also supported for `tree` and `tree-grid` roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required#associated_roles). We don't support labels on either our tree or table component yet but we do have some clients that treat them as more complex selector inputs. I could imagine them wanting to force a user to make a selection and indicate that via an asterisk too.

Might be useful to explicitly note that those components are excluded for scope reasons, not because we forgot to consider them.

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### Blazor
+
+Typically, form validation is handled via a `DataAnnotationsValidator`, and an input is treated as required if it is bound to a model property annotated with `RequiredAttribute` (i.e. `[Required]`). The Nimble component's `required` attribute is still needed to turn on the visual affordance, though, so we will add it to the Razor API.
+
+## Alternative Implementations / Designs
+
+None
+
+## Open Issues
```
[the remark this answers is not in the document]
m-akinc: That note is pretty obsolete now, as it seems all the browsers have supported `required` for 10+ years. I'll add some language to say that we don't need to set `aria-required` because the FAST templates either use elements that support `required` (i.e. `input`, `select`, etc) or are already setting `aria-required`.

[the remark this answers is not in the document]
m-akinc: Actually, it seems I was wrong about FAST's existing support being complete. I'm going to have to take a closer look to find out what's going on.

specs/required-inputs/README.md
m-akinc: Added a note in the problem statement.

specs/required-inputs/README.md
m-akinc: Yes. I made a note.

specs/required-inputs/README.md
m-akinc: Added brief notes on testing (we'll have matrix test cases) and documentation (only attribute documentation is necessary).

specs/required-inputs/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Our FAST base components provide built-in forms support via the [same API](https://www.w3schools.com/tags/att_input_required.asp) exposed by the native HTML `input`:
+
+- `required`: boolean attribute whose presence indicates that a value must be provided to submit
+
+ For radio buttons/groups, only the button elements expose the `required` attribute. If any of the radio buttons in a group are marked required, the group is treated as required. It does not matter if the button with `required` is disabled, hidden, etc. It is effectively as if the attribute is on the group of buttons, rather than a specific one. This matches the behavior of the native `input` with `type="radio"`. 
+
+### Visuals
+
+We will update our templates so that setting `required` causes a red asterisk to be displayed after the label, as per the visual design. For radio buttons/groups, the asterisk is on the group label, not individual button labels.
```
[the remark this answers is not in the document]
m-akinc: > Do you know how you'll add the asterisk?

Added details.

> Will it be a shared pattern or is it all per-component CSS?

Shared. Made a note.

> Do all the component labels currently behave the same if the label exceeds the control width?

Doubtful. That's out of scope.

> Do we know where the asterisk will appear if the label wraps?

It will always be to the right of the label's bounding box. I've added details about this to the HLD.

> Anything we need to do to hide it from the accessibility tree?

Nope. As long as we keep it separate from the label, it should be ignored.
image

specs/required-inputs/README.md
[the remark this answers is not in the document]
m-akinc: I've done an audit of these six inputs we're interested in and detailed the gaps that we will (and won't) fix.

specs/required-inputs/README.md
m-akinc: Added acceptance criterion to [Feature 2732543](https://ni.visualstudio.com/DevCentral/_workitems/edit/2732543).

specs/required-inputs/README.md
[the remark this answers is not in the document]
jattasNI: Nice job with the in-depth research and breakdown!

jattasNI (approved):

specs/required-inputs/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+This will be implemented using a styled span containing the text "*". The span must be outside of the `label` element so that it is not included in the accessible name. We will wrap the two in a `div` so that we can force a horizontal layout (`flex-direction: row`). This structure will also ensure that whatever the wrapping/ellipsizing behavior of the label is, the asterisk will always appear at the far right:
+
+```html
+<div class="annotated-label">
+    <label part="label" class="label">
+        <slot></slot>
+    </label>
+    ${when(x => x.required, html`
+        <span class="required-indicator">*</span>
```
rajsite, line 47: Using text doesn't seem right to me. It's a visual indicator, like an icon, etc. I wouldn't expect it to show up as the text content of a label, copy and paste, etc. It's more of an abuse of text as a visual indicator. If the font was customized strangely I think I'd expect this to always be the same visual indicator and not treated as text. Can we make this an icon instead?

specs/required-inputs/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+- [Nimble issue](https://github.com/ni/nimble/issues/2100)
+- [AzDO Feature](https://ni.visualstudio.com/DevCentral/_workitems/edit/2732543)
+- [Figma design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-47481)
+
+## Implementation / Design
+
+### API
+
+Our FAST base components provide built-in forms support via the [same API](https://www.w3schools.com/tags/att_input_required.asp) exposed by the native HTML `input`:
```
rajsite, line 29: nit: lets reference mdn docs, they tend to be higher quality with spec references etc.

i.e. mdn covers the radio group discussion below as well: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

specs/required-inputs/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+- `required`: boolean attribute whose presence indicates that a value must be provided to submit
+
+For radio buttons/groups, only the button elements expose the `required` attribute. If any of the radio buttons in a group are marked required, the group is treated as required. It does not matter if the button with `required` is disabled, hidden, etc. It is effectively as if the attribute is on the group of buttons, rather than a specific one. This matches the behavior of the native `input` with `type="radio"`.
```
rajsite, line 33: we can afford the chars

```suggestion
For radio buttons / radio button groups, only the radio button elements expose the `required` attribute. If any of the radio buttons in a radio button group are marked required, the radio button group is treated as required. It does not matter if the radio button with `required` is disabled, hidden, etc. It is effectively as if the attribute is on the radio button group, rather than a specific one. This matches the behavior of the native `input` with `type="radio"`. 
```

specs/required-inputs/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Form validation
+
+#### Vanilla HTML
+
+When the user attempts to submit a form and a `required` input is missing a value, a browser-specific indicator reports the validation error:
+
+![Missing value indicator](missing-value.png)
+
+We would expect this to already work, since the FAST components we derive from provide forms support. Unfortunately, this support is broken or incomplete for multiple components:
```
rajsite, line 64: It would be nice to break these out into issues for tracking in-case someone does rely on native form validation. We have at least one known user ([ozone for web login](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/Ozone/WebLogin/src/app/app.component.html&version=GBmaster&line=8&lineEnd=32&lineStartColumn=1&lineEndColumn=16&lineStyle=plain&_a=contents) leveraging native form with `<nimble-text-field>` and `<nimble-button type="submit">`).

rajsite (approved):

specs/required-inputs/README.md
jattasNI: We talked about this at standup today. Consensus was that an icon was probably better.

Brandon would need to provide the icon as well as specs about how it should be layed out relative to the text. He's happy to do that, just ping him when it's needed.

specs/required-inputs/README.md
m-akinc: Created bugs #2415, #2416, and #2417

I did not create bugs for the accessibility issues for the combobox and select, since they are within scope for this feature.

rajsite:
fyi @m-akinc merging is blocked on menu story instability, see https://github.com/ni/nimble/issues/1479#issuecomment-2341882851

Issue #1509, opened 2023-09-11
Scrollable tabs


munteannatan:
## 😯 Problem to Solve

When the tabs displayed have longer names or there are too many of them, they overflow the page.
This can hide valuable information from the user and hinder navigation.
![overflowing tabs](https://github.com/ni/nimble/assets/33986780/28667a44-6c66-4cce-8bbf-3a53ec5d07d0)

## 💁 Proposed Solution

The tabs should be customizable so that they may be confined to the existing container width and the tab titles could be truncated.
![truncated tabs](https://github.com/ni/nimble/assets/33986780/81bdb1ab-4f3c-4513-af02-114f31e11fd5)

The easiest way this can be achieved is by providing a part name to the Tab component default slot so that it can be customized.
```HTML
<template slot="tab" role="tab" aria-disabled="${x => x.disabled}">
  ${startSlotTemplate(options)}
  <slot part="content"></slot>
  ${endSlotTemplate(options)}
</template>
```

## 📋 Tasks
- [x] #2337

m-akinc:
Dynamic form fields are another case where long, user-provided labels may have the same issue.

jattasNI:
The team that originally submitted this request mentioned that they plan to use a workaround of manually setting the tab titles to include ellipses and to add a tooltip with the full text.

fredvisser:
## 🧑‍🔬 Research

Scrollable:
- https://opensource.adobe.com/spectrum-web-components/components/tabs-overflow/
- https://js.devexpress.com/Angular/Demos/WidgetsGallery/Demo/Tabs/Overview/MaterialBlueLight/
- https://base.uber.com/6d2425e9f/p/2345e5-tabs
- https://polaris.shopify.com/components/navigation/tabs?example=tabs-default
- https://carbondesignsystem.com/components/tabs/style/
- https://shoelace.style/components/tab-group
- https://workbench.gusto.com/components/tabs/usage/#reach-skip-nav
- https://vivid.deno.dev/components/tabs/
- https://mesadesignsystem.lanl.gov/components/tabs/overview

Overflow menu:
- https://fluent2.microsoft.design/components/web/react/tablist/usage
- https://www.lightningdesignsystem.com/components/tabs/#With-Overflow
- https://style.monday.com/?path=/docs/navigation-tabs-tabs--docs
- https://clarity.design/documentation/tabs/code

Squish/wrap:
- https://dash.plotly.com/dash-core-components/tabs
[...]

PR #2397, opened 2024-09-16
Scrollable Tabs HLD

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1509

## 👩‍💻 Implementation

N/A

## ✅ Checklist


packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+Any decision/API around limiting the width of a particular tab. This is ultimately an orthogonal concern to the UX this HLD is presenting.
+
+## Links To Relevant Work Items and Reference Material
+
+-   [Issue 1509: Scrollable Tabs](https://github.com/ni/nimble/issues/1509)
+-   [Figma design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=11133-57220&node-type=instance&t=DlnrdRo7KWYlheFz-0)
+
+## Implementation / Design
+
+![Scrollable Tabs](./spec-images/scrollable-tabs.png)
```
fredvisser, line 18: The visual design shows the use of ghost buttons, and has 4px of padding between the button and the tab scroll area. Could I trouble you to update the image and make that clear?

image

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+
+-   After interacting with a scroll button via the keyboard a user would be required to use the mouse to select a tab that is currently in view (if a completely new set of tabs was scrolled into view).
+-   Possibly goes against conventions for composite components
+-   Possibly not in-line with other Nimble conventions such as not having the inc/dec buttons on a numeric behave as tab stops.
+
+Related:
+
+-   The [Shoelace design system](https://shoelace.style/components/tab-group#scrolling-tabs) has an example of this approach
+
+**Roving tab strategy**:
```
fredvisser, line 69: Could you give an example or otherwise describe the roving tab strategy?

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+-   Seems like it might be more in-line with ARIA conventions
+-   Scrolling behavior can still be achievable through other keyboard interactions like `PageUp/PageDown`, which ARIA doesn't provide any explicit guidance for or against.
+-   Tabbing through the page may seem nice without the two extra stops that may be present with the scroll buttons.
+
+Cons:
+
+-   More expensive to implement
+-   Getting to the scroll buttons could now be overly cumbersome when tab focus is on a tab.
+
+**Scroll buttons are unreachable via keyboard**:
```
fredvisser, line 82: I've come around to favoring this. I'm not convinced the con applies - especially if we have the pageup/pagedown support providing better and more intuitive keyboard navigation… and the rest keeps things as simple as possible.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+
+Pros:
+
+-   Keyboard-only users won't run into issues with interactions with scroll buttons leading them down an unexpected path (i.e. pressing arrow keys would undo their scrolling).
```
m-akinc, line 86: > i.e. pressing arrow keys would undo their scrolling

Can you clarify? I didn't understand this.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
m-akinc: +1
I think the scroll buttons are the mouse user's version of `PgUp/PgDwn` and `Home/End`. I can't see a reason for a keyboard user to interact with the scroll buttons unless they are just unaware of the other key bindings.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+Pros:
+
+-   Seems like it might be more in-line with ARIA conventions
+-   Scrolling behavior can still be achievable through other keyboard interactions like `PageUp/PageDown`, which ARIA doesn't provide any explicit guidance for or against.
+-   Tabbing through the page may seem nice without the two extra stops that may be present with the scroll buttons.
+
+Cons:
+
+-   More expensive to implement
+-   Getting to the scroll buttons could now be overly cumbersome when tab focus is on a tab.
```
m-akinc, line 80: Yes. If you're using the keyboard, the scroll buttons become essentially useless, since to get to the right scroll button (for example) from a tab or the left scroll button, you have to arrow through all the tabs to the right. At which point, you're on the right-most tab, so the scroll button does nothing.

All that to say, I think the roving tab strategy is essentially the same as having the scroll buttons unreachable by keyboard, but slightly sillier.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+
+Only the tabs themselves will be scrollable. The scroll buttons will only surround the tabs. If there is content being displayed in a `TabsToolbar` that has been slotted, and the tabs content is wide enough to merit showing the scroll buttons, the right scroll button will sit to the left of the `TabsToolbar` (example below).
+
+![Scrollable Tabs with Toolbar](./spec-images/scrollable-tabs-toolbar.png)
+
+### Keyboard interaction
+
+#### `PageUp/PageDown`
+
+Now that the tabs will be able to be scrolled, it makes sense to add a keyboard interaction that will perform this scroll without the need of using the buttons. When a tab in the tab list has focus, pressing `PageUp/PageDown` will perform the same scroll behavior as pressing the scroll buttons _and additionally_ jump the tab focus by the same page amount (or to the first/last tab). This is needed to make the scrolling of the tabs useful for a keyboard-only user.
```
m-akinc, line 40: When you say "jump the tab focus", it would be good to clarify that for anchor tabs, the tab focus moves, but it shouldn't cause the focused tab to be activated. (Same is true for `Home/End`.) I believe that is the only difference between regular tabs and anchor tabs that applies to this HLD.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,106 @@
...
+## Links To Relevant Work Items and Reference Material
+
+-   [Issue 1509: Scrollable Tabs](https://github.com/ni/nimble/issues/1509)
+-   [Figma design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=11133-57220&node-type=instance&t=DlnrdRo7KWYlheFz-0)
+
+## Implementation / Design
+
+![Scrollable Tabs](./spec-images/scrollable-tabs.png)
+
+The following decisions were agreed to in the comments section in the linked issue above:
```
m-akinc, line 20: Unless it's complicated, I think we should support scrolling the tabs via mouse scroll wheel (when cursor is over a tab).

m-akinc (approved):

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
atmgrifter00: I'm trying to speak to the point that if the scroll buttons were focusable and a user attempted to use the keyboard to perform a scroll through them, then they might be under the impression that pressing the arrow key would move them to the first visible tab button in the view, but I was asserting that it would not, and instead move the focus one to the right/left of the currently focused tab (which would then unroll all of the scrolling they might have done via the scroll buttons).

That said, I suppose we _could_ consider the idea that if a user did use the scroll buttons via the keyboard that pressing the arrow key would _either_ move them to the right/left of the currently focused tab _if_ it was still in view, _or_ to the first visible tab. However, this seems like an overly complicated implementation which doesn't even seem like that much of an improvement over just not making the scroll buttons accessible via the keyboard.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
atmgrifter00: I also favor this fwiw.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
atmgrifter00: Agreed.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
atmgrifter00: For starters I misnamed this section, which I will correct. It should be "Roving tab _index_ strategy".

If implemented for the `Tabs` it would simply mean that pressing `<Tab>` when the next tabbable component on the page is the `Tabs` component would focus either the first or most recently focused `Tab`. Then to move focus to the other interactive components you would use the arrow keys. I think the main distinction between what I am proposing in this HLD and a typical implementation of the roving tab index, is that the scroll buttons would still never be given a `tabindex` of 0 (which would allow them to receive tab focus if a user tabbed again to the `Tabs` component). They would only be reachable via the arrow keys.

An example in Nimble that has implemented this strategy is the `Toolbar`.

change/@ni-nimble-components-e1af2752-74da-4c10-853b-97ea1d1b03be.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "First draft of scrollable tabs PR.",
```
mollykreis, line 3: Can you update this changefile description to not call this a "first draft"?

mollykreis (approved):

fredvisser (approved):

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,110 @@
...
+
+-   No scrollbar will be shown
+-   Keyboard navigation will work as it does today, but when a tab that isn't fully in view gets focus it should be scrolled into view
+-   When a scroll button is pressed we will scroll either a full "page width" or the remainder of the available space, and tab focus is not changed.
+
+Additionally, the mouse wheel should allow scrolling through the tabs, but instead of scrolling a page at a time, a single scroll "tick" will scroll by one tab.
```
jattasNI, line 26: It looks like [shoelace doesn't scroll by an entire tab, it scrolls by a single pixel](https://shoelace.style/components/tab-group#scrolling-tabs). Are we sure we want to differ here? Scrolling can generate a lot of "ticks" so it seems like scrolling by entire tabs would make it hard to get fine-grained control over which tabs are visible.

jattasNI, line 26: Is it a requirement to support pan scrolling on mobile? Do we need to do anything to support this or do we get it for free?

atmgrifter00: I agree...will change.

atmgrifter00: I imagine this is _not_ free. Pan scrolling is usually accomplished by a "click-drag" operation, the behavior of which will not be provided by the tabs. Is it worth calling this out as out-of-scope?

atmgrifter00: So, I will say that I couldn't get the shoelace example to scroll via the mouse wheel (in Chrome), but I will say that a "single pixel" is _not_ what we want. Playing around with it, 20 pixels at a time feels pretty nice, but there's likely an aesthetic "sweet spot" we will need to have consensus on.

atmgrifter00: I've updated the text to be more vague in terms of precise amounts that will be scrolled, but instead describe the goal of the experience.

jattasNI: I added a video below showing what Shoelace looks like on macOS when I scroll with the trackpad and it does seem to go pixel-by-pixel when I move slowly. Maybe a trackpad sends more fine-grained events than a mouse scroll wheel? Regardless, I'm fine to keep it vague in the HLD.

For both this question and the one about mobile support, I'm surprised we need to think about the low level events much though. It looks like in the Shoelace shadow DOM there is just a really wide div that contains the tabs, contained in a div that restricts the width to the width of the tab container. The browser handles scrolling through its normal `overflow` and `scrollLeft` attributes. It's the only thing that worries about scrolling the correct amount depending on the platform and input device. Can we do the same?

https://github.com/user-attachments/assets/0c7ad99c-f424-42e5-a587-b35011c700b1

atmgrifter00: I think maybe just using the `deltaY` property of the `WheelEvent` instance will accomplish this (here's an [SO post(https://stackoverflow.com/questions/20110224/what-is-the-height-of-a-line-in-a-wheel-event-deltamode-dom-delta-line)] that dives a bit into the details of each browser and the mode they use for the wheel event).

My suspicion is that we'll see the `deltaY` property have different values dependent on the device being used (and possibly the user agent). For Chrome on Windows with a mouse, for instance, I see a `deltaY` of 100 pixels per scroll tick, which is a bit jumpy IMO, but that's what we would get without intervening.

jattasNI: Reopening this based on our conversation about using `overflow: scroll` instead of handling scroll events.

atmgrifter00: Updated.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,54 @@
...
+
+![Scrollable Tabs](./spec-images/scrollable-tabs.png)
+
+The following decisions were agreed to in the comments section in the linked issue above:
+
+-   No scrollbar will be shown
+-   Keyboard navigation will work as it does today, but when a tab that isn't fully in view gets focus it should be scrolled into view
+-   When a scroll button is pressed we will scroll either a full "page width" or the remainder of the available space, and tab focus is not changed.
+
+Additionally, the mouse wheel should allow scrolling through the tabs. This should be handled natively by the browser by simply applying CSS of `overflow: scroll;` to the div containing the tabs. _Note: Mouse users would have to hold `<Shift>` while using the scroll wheel in order to scroll the tabs. This is consistent with other horizontally scrollable content._
```
jattasNI, line 26: Could you broaden this description to include other scroll gestures like trackpad and mobile pan gesture? You don't need to call them all out, just mention that typical platform-specific scroll interactions will work. (Even if you make this more generic it's probably useful to keep the note about Shift-scroll-wheel since it doesn't seem as widely known).

jattasNI (approved):

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -23,6 +23,8 @@ The following decisions were agreed to in the comments section in the linked iss
 -   Keyboard navigation will work as it does today, but when a tab that isn't fully in view gets focus it should be scrolled into view
 -   When a scroll button is pressed we will scroll either a full "page width" or the remainder of the available space, and tab focus is not changed.
 
+Additionally, devices like mouse wheels and trackpads as well as pan gestures on mobile devices should allow scrolling through the tabs. This should be handled natively by the browser by simply applying CSS of `overflow: scroll;` to the div containing the tabs. _Note: Mouse users would have to hold `<Shift>` while using the scroll wheel in order to scroll the tabs. This is consistent with other horizontally scrollable content._
```
m-akinc, line 26: > This is consistent with other horizontally scrollable content

I haven't done any kind of exhaustive survey, but consider the following cases where tab scrolling or focus is moved horizontally without `Shift`:
1. VS Code tabs
2. Windows Explorer tabs
3. Notepad++ tabs
4. Outlook ribbon hubs (e.g. File, Home, etc)

When there is only one scrollable axis (vertical or horizontal), it should scroll using just the mouse wheel. Only when there are two scrollable dimensions (like in a Figma doc or Excel spreadsheet) does it make sense to use `Shift` to scroll the secondary dimension. Even then, it's not completely consistent, e.g. Excel uses `Ctrl`+`Shift`+wheel for horizontal scrolling.

atmgrifter00: I'm certainly not against this decision. I was just using the browser as an example when there was only a horizontal scrollbar, it required the `<Shift>` key, but this could reasonably be simply because there is opportunity for there to also be vertically scrollable content, and they want to have consistent expectations for any keyboard modifiers. In this case, there is no such opportunity, so I agree, it would be nice for the user to allow it without using `<Shift>`. Thoughts @jattasNI?

jattasNI: My goals are:
1. match the user expectations for the platform. 
2. avoid unnecessary work and implementation complexity if we can.

For 1, I'm iffy on Mert's examples because they are Windows apps not websites (granted some of them may be implemented in HTML). My suspicion is that browser users wouldn't expect this behavior, but I could be proven wrong with more examples from the web. What does the scrollwheel on Windows in a browser on horizontally scrollable content that isn't embedded in a vertically scrollable page? To test, you could open the codepen for the [shoelace scrollable tabs example](https://shoelace.style/components/tab-group#scrolling-tabs).

For 2, my understanding is that this would add an outsize amount of complexity because we'd have to handle scroll events and translate them to offsets manually versus just letting the browser handle the scroll. That opens the door to the scrolling not feeling right and would necessitate more code and test cases. But I could be wrong about that.

atmgrifter00: To answer your question about the shoelace scrollable tabs behavior on Windows, it does, indeed, need `<Shift>` to scroll.

fredvisser: > _Note: Mouse users would have to hold `<Shift>`

This meets my expectations - 👍

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### TabsToolbar
+
+Only the tabs themselves will be scrollable. The scroll buttons will only surround the tabs. If there is content being displayed in a `TabsToolbar` that has been slotted, and the tabs content is wide enough to merit showing the scroll buttons, the right scroll button will sit to the left of the `TabsToolbar` (example below).
+
+![Scrollable Tabs with Toolbar](./spec-images/scrollable-tabs-toolbar.png)
+
+#### Minimum scroll area width
+
+As the tab list view width is now resizable, we must consider the scenario where the view is shrunk considerably. Preventing the scrollable area from disappearing completely seems desirable, and so we will set the CSS `minimum-width` property on the container of the tabs to be `4 * ${controlHeight}` (wide enough for both scroll buttons and extra space to see a good portion of at least one tab).
```
jattasNI, line 40:
```suggestion
As the tab list view width is now resizable, we must consider the scenario where the view is shrunk considerably. Preventing the scrollable area from disappearing completely seems desirable, and so we will set the CSS `min-width` property on the container of the tabs to be `4 * ${controlHeight}` (wide enough for both scroll buttons and extra space to see a good portion of at least one tab).
```

jattasNI (approved):

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### TabsToolbar
+
+Only the tabs themselves will be scrollable. The scroll buttons will only surround the tabs. If there is content being displayed in a `TabsToolbar` that has been slotted, and the tabs content is wide enough to merit showing the scroll buttons, the right scroll button will sit to the left of the `TabsToolbar` (example below).
+
+![Scrollable Tabs with Toolbar](./spec-images/scrollable-tabs-toolbar.png)
+
+#### Minimum scroll area width
+
+As the tab list view width is now resizable, we must consider the scenario where the view is shrunk considerably. Preventing the scrollable area from disappearing completely seems desirable, and so we will set the CSS `min-width` property on the container of the tabs to be `4 * ${controlHeight}` (wide enough for both scroll buttons and extra space to see a good portion of at least one tab).
```
rajsite, line 40: May also be too much of an implementation detail but setting the width to be a function of control height seems wrong for two reasons:
1. We should avoid having sizing calculation be a function of control height. control height should be a default value for native css sizing and layout, i.e. height, etc. See point 2 in https://github.com/ni/nimble/issues/610#issue-1277404188
2. control height seems to be used as a size ramp token in this case. The context is not about the height of controls. In that case we could be using or adding to the size ramp tokens: https://nimble.ni.dev/storybook/?path=/docs/tokens-size-ramp--docs Those are called "padding" but are just fixed sizes without actual contextual meaning. In this case we might as well just have a fixed value since it is kinda arbitrary.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Alternative Implementations / Designs
+
+Another way to display tabs that would otherwise exceed the horizontal visible area is to provide multiple rows of tabs. If we ever wanted to provide this ability to a user, we could always provide an API that allows them to switch between a `scrollable` and a `stacked` mode.
+
+Additionally, we _could_ begin with an API that enables/disables this new behavior via an enum configuration that defaults to `none` (or the new behavior). However, we could always add it if there are clients that ask for the previous behavior in addition to the scrollable tabs.
```
rajsite, line 46: I think we'd be able to come up with an api that is backwards compatible / non-breaking pretty easily. Something like:
```
TabsLayout {
    scrollabe: undefined, // i.e. the default
    stacked: 'stacked'
}
```

So not overly concerned about beginning that way

rajsite (approved):

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
```diff
@@ -0,0 +1,58 @@
...
+
+The following decisions were agreed to in the comments section in the linked issue above:
+
+-   No scrollbar will be shown
+-   Keyboard navigation will work as it does today, but when a tab that isn't fully in view gets focus it should be scrolled into view
+-   When a scroll button is pressed we will scroll either a full "page width" or the remainder of the available space, and tab focus is not changed.
+
+Additionally, devices like mouse wheels and trackpads as well as pan gestures on mobile devices should allow scrolling through the tabs. This should be handled natively by the browser by simply applying CSS of `overflow: scroll;` to the div containing the tabs. _Note: Mouse users would have to hold `<Shift>` while using the scroll wheel in order to scroll the tabs. This is consistent with other horizontally scrollable content._
+
+To accomplish this we will fork FAST's template in order to add the scroll buttons. A `ResizeObserver` will be used to track when the scrollable area exceeds the viewable area in order to show/hide the scroll buttons.
```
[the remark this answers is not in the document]
atmgrifter00: I've removed mention of the `ResizeObserver` and simply described the behavior we will need.

packages/nimble-components/src/tabs/specs/scrollable-tabs.md
atmgrifter00: Limited the description to simply saying we will need to set the `min-width` on the `Tab` component, and that it should be of a sufficient size to accommodate both scroll buttons.

PR #2399, opened 2024-09-18
Text field error spacing

### LINKED ISSUE #2090: Error icon is incorrectly spaced in nimble-text-field
# 🐛 Bug Report

In the nimble-text-field, the error "!" icon should be 8px from the edge of the component. However, it is currently 8px from the border, which makes it 9px from the edge of the component.

We should evaluate whether this is the case with other components as well.

## 🤔 Expected Behavior

The error icon in components should be 8px from the edge of the component, not 8px from the border of the component.

## 😯 Current Behavior

The error icon is 8px from the border of the component.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2090

There are a few problems with the spacing in the text field:

1. The spacing between the error icon and the right edge of the component is 8px rather than 4px.
2. The spacing between the "parts" of the component is 8px rather than 4px. This is spacing such as the inner `input` to the error icon, the inner `input` to the `action` slot buttons, and the error icon to the `action` slot buttons.
3. The spacing between the right edge of the component and the content within the component (error icon, `action` slot buttons, or right edge of the inner `input`) is measured from the border rather than including the border.
4. The spacing between the left edge of the component and the `start` slot icon is 8px rather than 4px.
5. The spacing between the left edge of the component and the content within the component (`start` slot icon or left edge of the inner `input`).

This PR addresses problems 1 & 2 above, but it does not address problems 3-5 because these should be addressed across all components (select, combobox, number field, etc) in order to avoid creating misalignments of text across components. Therefore, I created #2400.

## 👩‍💻 Implementation

- Updated the `gap` used in the text-field to be small padding (4px) rather than medium padding (8px), which addressed problems 1 & 2.
    - In order to keep the input 8px from the left side of the text field, with the new 4px `gap`, I also added 4px to the left padding of the `.root` element.
    - Similarly, I added a 4px margin on the right of the slotted `start` icon to keep the icon 8px from the `input`.

## 🧪 Testing

- Inspected the chromatic diffs

## ✅ Checklist


mollykreis:

m-akinc (approved):

change/@ni-nimble-components-961874d3-9456-4e57-b9c6-f49f80c38843.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
mollykreis, line 2: Note: I made this a `patch` update because there aren't any API changes, and I don't expect most clients to be impacted by this change. However, we should update the DFF boolean input when this change is picked up by SLE because they designed their boolen input to have its error icon align with the previous state of the text field.

jattasNI (approved):

packages/nimble-components/src/text-field/styles.ts
```diff
@@ -104,6 +105,7 @@ export const styles = css`
         color: transparent;
         width: 0px;
         ${userSelectNone}
+        margin-left: -1px;
```
mollykreis, line 108: Note: If we decide not to fix #2400, then this negative left margin can be removed.

rajsite: How would you feel about splitting out any border spacing stuff from this PR to avoid blocking on that discussion? i.e. hold the current assumption that spacing from the border does not include the border for now.

mollykreis: That seems like a reasonable plan to me.

packages/nimble-components/src/text-field/styles.ts
```diff
@@ -168,6 +170,7 @@ export const styles = css`
     [part='end']::after {
         content: '';
         position: absolute;
+        margin-left: calc(-1 * (${borderWidth} + ${smallPadding}));
```
rajsite, line 173: Poking in devtools, the `::after` is an absolutely positioned element. Instead of the negative margin (which I'll admit don't fully understand where this calc is coming from) can we say this this is is just positioned `left: 0px`? Seems like the intention is to have the hover line flush to the left.

rajsite (approved):

packages/nimble-components/src/text-field/styles.ts
mollykreis: I agree that `left: 0px` would be more ideal styling, but it breaks the hover animation. The animation is intended to grow from the center of the control, but setting `left: 0px` makes it grow from the left edge.

The negative margin here is to offset the padding of the `.root` element, which had made this element (the hover border) incorrectly shifted to the right.

rajsite: makes sense!

Issue #2219, opened 2024-06-26
RichTextEditorMentionListbox test intermittent with error No selection range exists


rajsite:
# 🐛 Bug Report

The test `RichTextEditorMentionListbox Dynamically update mention popup items based on configuration changes should show mention popup for multiple mention configuration elements FAILED` is intermittent and fails with the error message:

```
InvalidStateError: Selection.collapseToEnd: No selection range exists thrown
```

Update: same error seen on test `Markdown serializer Serialize rich text editor content to its respective markdown Mention node` in [this build](https://github.com/ni/nimble/actions/runs/9897035682/job/27340606528?pr=2242). That test is now disabled as well.

## 💻 Repro or Code Sample

See build https://github.com/ni/nimble/actions/runs/9686770115/job/26729936486#step:17:411

## 💁 Possible Solution

Seems like it may be related to a similar sounding error from https://github.com/ni/nimble/pull/2200

rajsite:
@vivinkrishna-ni could you create a corresponding AzDo issue to track this intermittent test failure? It seems like it may be similar to the type of failure covered in https://github.com/ni/nimble/pull/2200.

vivinkrishna-ni:
Created an AzDo bug: [Bug 2787273](https://dev.azure.com/ni/DevCentral/_workitems/edit/2787273): Intermittent failures in Rich Text Editor for Mention support

Last time (in https://github.com/ni/nimble/pull/2200), it was evident that the editor was not focused at the time of committing the mention node and was failing in local with some conditions. However, this time, it is a rare occurrence where I tested like running the tests in a loop for about a thousand times, running the test that ran before in the failed pipeline, or keeping the breakpoint at the line where the error was thrown in Tiptap, which still not be able to reproducible.

rajsite:
@vivinkrishna-ni an additional test failed with the same error message and was disabled in https://github.com/ni/nimble/pull/2262

vivinkrishna-ni:
Closing the issue as the test failures are fixed as part of #2403. Please feel free to reopen the issue if the same tests are failing again.

PR #2396, opened 2024-09-13
Draft | Run mention intermittent failing test in loop

aagash-ni:
# Pull Request

## ✅ Checklist

- [ ] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2403, opened 2024-09-19
Re-enable mention intermittent failed cases

aagash-ni:
# Pull Request

## 🤨 Rationale

**Issue**: [#2219](https://github.com/ni/nimble/issues/2219)

We are re-enabling tests that were previously disabled due to intermittent failures and disabling only in the Firefox environment, as the issue occurred only once during four days of testing. Additionally, there were other cases that are validating the same mention workflow, which has been successfully running in Firefox.

After conducting tests in a loop environment on GitHub agents, the tests did not fail with the same error.

Please refer to the bug discussion for all the pipeline checks: [Bug 2787273](https://dev.azure.com/ni/DevCentral/_workitems/edit/2787273) - Intermittent failures in the Rich Text Editor for Mention support.

Following this, we will monitor for any related failures and have created a [TD](https://github.com/ni/nimble/issues/2404) to track the skipped Firefox browser tests.

## 👩‍💻 Implementation

Re-enabled the disabled test case.

Update:
1. For test `should show mention popup for multiple mention configuration elements`:
    1. Instead of committing the `@` mention node, remove the `@` and add `!` to verify other mention type list options and popup visibility.
    2. The test can now verify the list options for multiple mentions by opening the popup.
1. For test `Serialize rich text editor content to its respective markdown Mention node`:
    1. Update the `clickMentionListboxOption` to focus on the editor before committing the mention from the list options.
1. Update other methods that modify the content of the editor to also receive focus to the editor to ensure any other intermittencies.
1. Update the `isMentionListboxOpened` method to read from the `hidden` property and also wait before reading to resolve the intermittent failures on Firefox when running in loop.

See the discussion below for more details, https://github.com/ni/nimble/pull/2403#discussion_r1767401349

## 🧪 Testing

NA

## ✅ Checklist


jattasNI (approved):

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -1542,8 +1542,8 @@ describe('RichTextEditorMentionListbox', () => {
             expect(pageObject.isMentionListboxOpened()).toBeTrue();
         });
 
-        // Intermittent, see https://github.com/ni/nimble/issues/2219
-        xit('should show mention popup for multiple mention configuration elements', async () => {
+        // Firefox skipped, see https://github.com/ni/nimble/issues/2404
+        it('should show mention popup for multiple mention configuration elements #SkipFirefox', async () => {
```
rajsite, line 1546: I don't think I understand the plan. What is going to drive understanding the intermittency in Firefox? The test won't be running and no logging is being added to gather more information on failure.
It seems like this will stay disabled in Firefox indefinitely. @jattasNI @aagash-ni

rajsite: A concrete proposal would be:
1) Take time to consider what should happen in a failure case and what extra information would be useful to gather from the page to help debug.
2) Re-enable it under the idea that it will not fail and if it fails again, restart the triage process. Seems undesirable because their seems to be knowledge that it is expected to fail intermittently on Firefox.
3) In this PR: https://github.com/ni/nimble/pull/2396/files#diff-4b859665afd3d8b0eab8e029e088d38102ae9bda4f89964d8efa3d718f6c397aR1379 it looks like you only ran the test in a loop 30 times. You could try pushing it harder. Mark the test as `fdescribe` to ignore other tests and maybe try 100, 1000, 10000 times? Thirty seems like a pretty low and oddly specific number to try and say that it's not reproducible.

rajsite: @aagash-ni I haven't seen any follow-up to thee review, moving the PR back to draft until can make further progress

vivinkrishna-ni: Hi @rajsite, We have tried a few more ways to find the root cause of the issue today. However, the cause has not yet been identified.

However, I am sharing some observations based on the above proposals:
1. After numerous attempts, the test for `should show mention popup for multiple mention configuration elements` failed only once due to the reason specified in the bug report. This occurred after running the test in a loop hundreds of times. Here is the error log.
     image

      The error was caused by `clickMentionListboxOption(0)` in the line above. However, for this test, committing the node into the editor was not necessary, and we were also not asserting the committed mention node. Instead, after asserting mention 1, I removed the `@` from the editor using the `sliceEditorContent` method from the page object. Then, added `!` and asserted the list box options and the visibility of the popup for other mention.
1. For the other test in the serializer, the test never failed even after many retries. However, based on the previous issue fix #2200, before committing the node from the listbox, focus the editor again by updating a page object in the `clickMentionListboxOption`.
      1. Can we try re-enabling with these changes and monitoring for a few more days? Maybe I will try to figure out any helpful logs that can be added.
1. In the other pull request (#2396), we attempted to focus on fixing the failing test and running it multiple times.
     1. Run both the tests 1000 times - No failures: https://github.com/ni/nimble/actions/runs/11160728028/job/31021797869?pr=2396#step:17:17743
     3. Run both the tests 10000 times - Pipeline exited with an error code: https://github.com/ni/nimble/actions/runs/11164365077/job/31033523031?pr=2396#step:17:472
     4. Run the tests 1000 times again with the changes explained in the above points (`#1` & `#2`) - the test in the editor failed 27 times (Only on Firefox) but for a different reason: https://github.com/ni/nimble/actions/runs/11161515146/job/31024221277?pr=2396#step:17:12176
           1. The failed test can only be reproduced if the test is in a loop and not in the ideal case. After further debugging, it was found that `isMentionListboxOpened` returned `false` because the `nimble-anchored-region` intermittently has the `hidden` attribute. This issue is not seen in the actual Firefox browser, but it is reproducible when running `npm run test-firefox:debugger` locally. So, could this be related to the Firefox launcher? Is debugging in that direction the right approach?

Please share your thoughts if you have any on these observations and let us know if we can proceed with the changes in the PR and monitor the fix for some more time.

vivinkrishna-ni: > Maybe I will try to figure out any helpful logs that can be added.

The only helpful log I could think of for this error is to get the focus status of the editor. I would like to understand if that's the case the test in the serializer is failing intermittently on the Firefox browser. The reason for this log is that the error (`No selection range exists`) can be thrown from the Tiptap, only if there are no selections (the editor might not be focused).

Therefore, can we add a log to the test in the PR? Or can we double-check by focusing on the editor again when `clickMentionListboxOption` is executed?

vivinkrishna-ni: @rajsite, Kindly review the changes in the PR and share your thoughts on the proposals and observations mentioned above.

vivinkrishna-ni: @rajsite Kindly review the changes in the PR and let us know your thoughts on the above proposals and observations.

rajsite (commented):
Should poke at isMentionListboxOpened a bit more

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
```diff
@@ -1542,29 +1542,34 @@ describe('RichTextEditorMentionListbox', () => {
...
-            expect(pageObject.isMentionListboxOpened()).toBeTrue();
+
             expect(pageObject.getMentionListboxItemsName()).toEqual([
                 'username1',
                 'username2'
             ]);
-            await pageObject.clickMentionListboxOption(0);
+            expect(pageObject.isMentionListboxOpened()).toBeTrue();
+
+            await pageObject.sliceEditorContent(0, 2);
```
rajsite, line 1559: Another thought is `setEditorTextContent` has this wait check:

https://github.com/ni/nimble/blob/52a54083a9350199fa16a7344e83efd81d0549ce/packages/nimble-components/src/rich-text/editor/testing/rich-text-editor.pageobject.ts#L251-L255

Maybe `sliceEditorContent` should have that as well? Maybe others that manipulate the content in the page object like `replaceEditorContent`?

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
rajsite: @vivinkrishna-ni I see you pushed the brainstorming ideas to the branch. To emphasize, I didn't prototype or test those ideas to see if they are actually beneficial or just hurt test performance by adding unnecessary waits.

It sounded like you had found a way to cause the failures somewhat reliably via many executive so hopefully the brainstorming ideas were tested against that

vivinkrishna-ni: @rajsite I've tried both of the suggestions (mentioned above) in the other PR (#2396) where we run the tests in a loop. It looks like the tests fail for the same reason. However, the frequency of failure is reduced, from 53 failures out of 2000 attempts to 37 failures out of 2000 attempts. In terms of property vs attribute and when logging the `hidden` property and attribute value, both are the same. So, I think adding the `await waitForUpdatesAsync();` before reading the `hidden` property/attribute makes the frequency difference I guess.

But, based on the other comment (https://github.com/ni/nimble/pull/2403#discussion_r1793940205) below, it fixes the issue. By focusing on the editor in the `sliceEditorContent` method, the issue has been fixed. I have updated the changes in the other PR, and the tests did not fail for all three attempts when running them in a loop for 2000 times each. Pipeline action: https://github.com/ni/nimble/actions/runs/11272830491/job/31351196076 (Verified the failures from the additional logs).

Therefore, I focused on the editor for all the methods in the page object that modify the content of the editor to avoid any other intermittencies. There are some lint failures in the PR which I will fix soon but apart from that please let me know whether can we try to fix this long-standing intermittent test failure with these changes. We also plan to monitor the failures for a few days before we close the bug.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
vivinkrishna-ni: As mentioned in the above thread, focusing on the `sliceEditorContent` fixes the issue. Also, did the same for other methods that manipulate the content of the editor.

packages/nimble-components/src/rich-text/editor/tests/rich-text-editor-mention.spec.ts
rajsite: Thanks for testing in the other PR. Sounds very promising that the tests are passing more reliably! Generally looks good but can approve after the build passes 👍

vivinkrishna-ni: @rajsite, Please consider this a gentle reminder to review the PR and approve it if you don't have any comments.

rajsite (approved):

PR #2409, opened 2024-09-24
Update styling for number field when readonly

### LINKED ISSUE #2402: Number field shows hover styling even when `readonly`
# 🐛 Bug Report

A `nimble-number-field` with `readonly` set should not respond to mouse hover by showing hover styling.

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2402 by adding styling for the `readonly` number field. Those styles were based heavily on the `readonly` styling for the text-field.

## 👩‍💻 Implementation

- Add styling for the number-field when it is `readonly`
- Update storybook page to include `readonly`
- Made shared docs for `readonly`

## 🧪 Testing

- Manually tested combinations of `readonly`, `disabled`, and `error-visible` with the different appearance modes in storybook
- Updated matrix tests, including interactive matrix tests, to include `readonly`
- Verified that `readonly` is exposed in Angular & Blazor APIs even though it was overlooked in storybook, matrix tests, and styling

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/utilities/storybook.ts
+++ b/packages/storybook/src/utilities/storybook.ts
@@ -181,6 +181,9 @@ export const iconDescription = 'Set `slot="start"` to include an icon before the
 export const disabledDescription = (options: {
     componentName: string
 }): string => `Styles the ${options.componentName} as disabled and prevents focus and user interaction.`;
+export const readonlyDescription = (options: {
+    componentName: string
+}): string => `Disallows input on the ${options.componentName} while maintaining enabled appearance.`;
 export const slottedLabelDescription = (options: {
     componentName: string
 }): string => `Label text to display adjacent to the ${options.componentName} describing its purpose to the user.`;
--- a/packages/storybook/src/utilities/storybook.ts
+++ b/packages/storybook/src/utilities/storybook.ts
@@ -183,7 +183,7 @@ export const disabledDescription = (options: {
 }): string => `Styles the ${options.componentName} as disabled and prevents focus and user interaction.`;
 export const readonlyDescription = (options: {
     componentName: string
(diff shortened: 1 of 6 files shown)


mollykreis:

packages/storybook/src/utilities/storybook.ts
```diff
@@ -181,6 +181,9 @@ export const iconDescription = 'Set `slot="start"` to include an icon before the
 export const disabledDescription = (options: {
     componentName: string
 }): string => `Styles the ${options.componentName} as disabled and prevents focus and user interaction.`;
+export const readonlyDescription = (options: {
+    componentName: string
+}): string => `Disallows input on the ${options.componentName} while maintaining enabled appearance.`;
```
msmithNI, line 186: We say `while maintaining enabled appearance` but that doesn't seem accurate, since we change the border/underline color (same as for disabled styling). Minor, but figured I'd comment to see if others think we should remove that wording or not.

msmithNI (approved):

packages/storybook/src/utilities/storybook.ts
jattasNI: I think I might have written this originally in the hopes of distinguishing it from `disabled`, but I agree it's not really accurate. How about this, borrowing from `disabled` above?
```suggestion
}): string => `Styles the ${options.componentName} as readonly and prevents the user from changing the value.`;
```

Alternatively we could just copy the MDN definitions, but they're a bit wordier than we usually go with:
>The Boolean readonly attribute, when present, makes the element not mutable, meaning the user can not edit the control.
>The Boolean disabled attribute, when present, makes the element not mutable, focusable, or even submitted with the form. The user can neither edit nor focus on the control, nor its form control descendants.

jattasNI (approved):

rajsite (approved):

Issue #1667, opened 2023-11-15
Blazor .NET 8 support 


rajsite:
## 😯 Problem to Solve

We should come up with a plan for .NET 8 support. The Fluent folks made it a breaking change (might mostly be due to a namespace change) and added support for some. NET 8 specific features detailed in a [blog](https://baaijte.net/blog/microsoft-fluentui-aspnetcore.components-4/)

.NET 6 (our current version) goes out of active support in November 2024.

[Here is a Teams thread](https://teams.microsoft.com/l/message/19:5833c467dc794397881707f0c48dde03@thread.skype/1713202657398?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=908d6bb1-598b-4ad1-bb3a-4f7573b10dd0&parentMessageId=1713202657398&teamName=NI%20Web%20UI%20Chapter&channelName=Blazor&createdTime=1713202657398) discussing versions that existing clients use.

## 📋 Tasks
- [x] Create HLD - #2411
- #2454 :
  - [x] Update `TargetFramework` from `net6.0` to `net8.0` for all BlazorWorkspace.sln projects
  - [x] Update `Microsoft.AspNetCore.Components.Web` reference from `6.x` to `8.x` for all BlazorWorkspace.sln projects
  - [x] Unpin analyzer versions (`NI.CSharp.Analyzers`, from `[2.0.21]`), update to latest - see #2014 
  - [x] Update examples: Re-build Demo.Server example app off of new "Blazor Web App" template ("Blazor Server App" template is now deprecated / cannot be targeted to .NET 8). Evaluate if Demo.Client/Demo.Server should be merged. Ensure Demo.Hybrid still works as expected.
  - [x] Update NimbleBlazor and SprightBlazor README.md - location to add Nimble/Spright imports changed (no longer `_Layout.cshtml` for Blazor Server/ Web App variant, now `App.razor`)
  - [x] Other doc updates (including CONTRIBUTING, updating SDK version references). Document supported Blazor render modes.
  - [x] Evaluate that Playwright tests are written to sufficiently test .NET 8 apps. For example our existing tests didn't catch the issue above.
  - [x] #2355 
- [ ] Fix all errors/warnings: ~72 to resolve. Most common 3 are IDE0290 (Use primary constructor), IDE0028 (Collection initialization can be simplified), IDE0090 (new expression can be simplified). Alternatively suppress to start with, and create separate tech debt / PR to address fully. : Split off to #2453 
[...]

PR #2411, opened 2024-09-25
HLD: Blazor .NET 8 support

msmithNI:
# Pull Request

## 🤨 Rationale

HLD for #1667

### DIFF
--- a/specs/nimble-blazor/blazor-net8-upgrade-hld.md
+++ b/specs/nimble-blazor/blazor-net8-upgrade-hld.md
@@ -25,10 +25,11 @@ Related work items:
   - Update code analyzer package versions to latest versions
   - Resolve build warnings (mostly due to new code analysis rules)
   - Update/recreate demo projects based on newer Blazor .NET 8 templates
-  - Blazor [render modes](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0#render-modes):
+    - Evaluate if Demo.Server/ Demo.Client should be merged (based on new "Blazor Web App" template type)
+    - Ensure Blazor Hybrid example (Demo.Hybrid) still works as expected
+  - Blazor [render modes](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0#render-modes): Static Server, Interactive Server, Interactive WebAssembly, Interactive Auto
     - Test each render mode with our components
     - Evaluate how to represent each in example projects/ docs/ acceptance tests (if possible)
-    - Render modes: Static Server, Interactive Server, Interactive WebAssembly, Interactive Auto
   - Ensure adequate test coverage, especially of interactions with Blazor initialization / lifecycle code.
 - Update documentation and readme files
   - Ensure correct getting started instructions, required .NET SDK versions, etc.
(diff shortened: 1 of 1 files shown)


atmgrifter00 (approved):

specs/nimble-blazor/blazor-net8-upgrade-hld.md
```diff
@@ -0,0 +1,88 @@
...
+Design System work items:
+- [Feature 2812085: Nimble Blazor .NET 8 upgrade](https://dev.azure.com/ni/DevCentral/_workitems/edit/2812085)
+- [nimble#1667: Blazor .NET 8 support](https://github.com/ni/nimble/issues/1667)
+
+Related work items:
+- ASW: [Epic 2753078: ASW .NET 8 Migration](https://dev.azure.com/ni/DevCentral/_workitems/edit/2753078)
+
+## Implementation / Design
+
+### Planned Changes ###
```
jattasNI, line 21: There will be some doc updates needed too. If you search .md files for "net" you can see a few references to SDK versions in CONTRIBUTING, for example.

jattasNI (approved):

specs/nimble-blazor/blazor-net8-upgrade-hld.md
```diff
@@ -0,0 +1,88 @@
...
+- For all projects in the Nimble `BlazorWorkspace.sln` (including Nimble Blazor and Spright Blazor):
+  - Update `TargetFramework` from `net6.0` to `net8.0`
+  - Update `Microsoft.AspNetCore.Components.Web` references from `6.x` to `8.x`
+  - Update code analyzer package versions to latest versions
+  - Resolve build warnings (mostly due to new code analysis rules)
+  - Update/recreate demo projects based on newer Blazor .NET 8 templates
+  - Ensure adequate test coverage, especially of interactions with Blazor initialization / lifecycle code.
```
rajsite, line 28: My understanding is that a big change with .NET 8 is being able to mix and match [render modes](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0#render-modes) of SSR, Blazor Server, and Blazor WASM. And also that Razor Class Libraries in [ASP.NET Core MVC and Razor Pages can mix and match with Blazor](https://learn.microsoft.com/en-us/aspnet/core/tutorials/choose-web-ui?view=aspnetcore-8.0#combine-multiple-web-ui-solutions-aspnet-core-mvc-or-razor-pages-plus-blazor). And then there is also Blazor Hybrid.

Since NimbleBlazor really is a Razor Class Library fundamentally, I think it would be helpful to spell out more concretely the different modes we will support and test. My first reaction is that I'd like to support all of those modes.

mollykreis:
I'm declining to review this PR, so don't wait for a review from me.

specs/nimble-blazor/blazor-net8-upgrade-hld.md
msmithNI: Added a bullet point for doc updates.

specs/nimble-blazor/blazor-net8-upgrade-hld.md
msmithNI: Good point. Called out the Blazor render modes more specifically in the HLD. The example updates will be part of that, but we'll also document which modes we've tested / that we support.

msmithNI: As far as mixing&matching Blazor with other MVC/Razor pages, I'm not sure that there's really any additional testing to do for that, since the Blazor example projects are already ASP.NET Core MVC too.

rajsite (approved):

PR #2412, opened 2024-09-25
Updates for switch to eslint return-await to always

rajsite:
# Pull Request

## 🤨 Rationale

Adopting the soon to update `return-await` configuration in the `always` mode.

## 👩‍💻 Implementation

- Updated the eslint rule
- Ran the fixer (via `npm run format`)
- Made a couple minor tweaks pointed out below

## 🧪 Testing

Rely on CI

## ✅ Checklist


packages/nimble-components/src/drawer/tests/drawer.spec.ts
```diff
@@ -40,7 +40,7 @@ describe('Drawer', () => {
     async function completeAnimationAsync(
         nimbleDrawerElement: Drawer | Drawer<string>
     ): Promise<void> {
-        return new Promise(resolve => {
+        await new Promise<void>(resolve => {
```
rajsite, line 43: There were a couple spots that failed type checking after running the formatter like here. Because the return of this function is `Promise<void>` that means no return value value is needed anymore. The fixer changed this to just `await new Promise()` (which I think is a better representation of this function to not return anything since it's `Promise<void>`) 

But now that it is not part of the return type, the Promise return value cannot be inferred and had to be made explicit.

packages/nimble-components/src/utilities/wait-until-custom-elements-defined-async.ts
```diff
@@ -1,8 +1,10 @@
...
     elements: Element[]
 ): Promise<void> {
-    const definedElements = elements.map(async item => (item.matches(':not(:defined)')
-        ? customElements.whenDefined(item.localName)
-        : Promise.resolve()));
+    const definedElements = elements.map(async item => {
+        if (item.matches(':not(:defined)')) {
+            await customElements.whenDefined(item.localName);
+        }
+    });
```
rajsite, line 8: This was the only other interesting change, The fixer resulted in:

```ts
    const definedElements = elements.map(async item => (item.matches(':not(:defined)')
        ? await customElements.whenDefined(item.localName)
        : await Promise.resolve()));
```

And that's technically fine but it looked funny. We don't actually care about this function returning anything, it just needs to resolve when everything is defined. I tweaked the function and I think that is conveyed more clearly now.

rajsite:
@jattasNI gonna bypass merge mechanical lint rule adoption

packages/nimble-components/src/utilities/wait-until-custom-elements-defined-async.ts
rajsite: In retrospect doing a filter and only awaiting what needs to be awaited would have been even better, but it's a nit:

```ts
const definedElements = elements
    .filter(item => item.matches(':not(:defined)'))
    .map(async item => {
        await customElements.whenDefined(item.localName);
    });
```

PR #2413, opened 2024-09-25
Dependency update September 2024

rajsite:
# Pull Request

## 🤨 Rationale

A follow-up for https://github.com/ni/nimble/pull/2410 and https://github.com/ni/nimble/pull/2408 which were getting in weird states due to dependency updates around rollup.

## 👩‍💻 Implementation

Previously had found that the rollup-plugin-sourcemaps was [causing issues](https://github.com/ni/nimble/pull/2318#discussion_r1700858311) because it had not been updated in a long time and resulted in a transitive dependency on an old rollup version. This seems to mess with rennovate updates.

- Updated to [rollup-plugin-sourcemaps2](https://github.com/2wce/rollup-plugin-sourcemaps2) which now seems to be [used in at least a few projects](https://www.npmjs.com/package/rollup-plugin-sourcemaps2)
- Updated other dependencies that were pending
- Leveraged a new flag added to concurrently that causes the build logs to align nicer by making prefixes the same width. Also has a bug fix that handles stray new lines once in a while.
- Rebuilt the lockfile

## 🧪 Testing

- Verified manually that the bundled minified builds have working source maps
- Rely on CI for rest of the updates
- Trigger full storybook build

## ✅ Checklist


atmgrifter00 (approved):

rajsite:
@jattasNI gonna bypass for a pretty straightforward dependency upgrade

PR #2423, opened 2024-10-01
Radio group error

mollykreis:
# Pull Request

## 🤨 Rationale

Includes `nimble-component` implementation for #2019

## 👩‍💻 Implementation

- Adds `error-visible` and `error-text` attributes to the `nimble-radio-group`
- Updates the radio group template to include an error icon and error text
    - Note: This required forking FAST's template
- Updated radio group styling to match visual design

## 🧪 Testing

- Forked FAST's radio group tests since the template was forked
- Updated matrix tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-e5bfd621-8ece-4d9c-a6ea-eb128a70e8ac.json
+++ b/change/@ni-nimble-components-e5bfd621-8ece-4d9c-a6ea-eb128a70e8ac.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Implement error states on nimble-radio-group",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/radio-group/styles.ts
+++ b/packages/nimble-components/src/radio-group/styles.ts
@@ -4,6 +4,7 @@ import {
     controlLabelDisabledFontColor,
     controlLabelFont,
     controlLabelFontColor,
+    controlLabelFontLineHeight,
     smallPadding,
     standardPadding
 } from '../theme-provider/design-tokens';
@@ -29,6 +30,7 @@ export const styles = css`
 
     .label-container {
         display: flex;
+        height: ${controlLabelFontLineHeight};
         gap: ${smallPadding};
         margin-bottom: ${smallPadding};
     }
--- a/packages/nimble-components/src/radio-group/styles.ts
+++ b/packages/nimble-components/src/radio-group/styles.ts
@@ -41,7 +41,7 @@ export const styles = css`
     :host([disabled]) slot[name='label'] {
         color: ${controlLabelDisabledFontColor};
     }
-    
+
     .error-icon {
         margin-left: auto;
         margin-right: ${smallPadding};
--- a/packages/nimble-components/src/radio-group/styles.ts
+++ b/packages/nimble-components/src/radio-group/styles.ts
@@ -4,15 +4,19 @@ import {
     controlLabelDisabledFontColor,
     controlLabelFont,
     controlLabelFontColor,
+    smallPadding,
     standardPadding
 } from '../theme-provider/design-tokens';
+import { styles as errorStyles } from '../patterns/error/styles';
 
 export const styles = css`
     ${display('inline-block')}
+    ${errorStyles}
 
     .positioning-region {
         display: flex;
         gap: ${standardPadding};
+        position: relative;
     }
 
     :host([orientation='vertical']) .positioning-region {
@@ -23,6 +27,12 @@ export const styles = css`
         flex-direction: row;
     }
 
+    .label-container {
+        display: flex;
+        gap: ${smallPadding};
+        margin-bottom: ${smallPadding};
+    }
+
     slot[name='label'] {
         font: ${controlLabelFont};
         color: ${controlLabelFontColor};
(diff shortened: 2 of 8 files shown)


mollykreis:

packages/nimble-components/src/radio-group/styles.ts
```diff
@@ -23,6 +27,12 @@ export const styles = css`
         flex-direction: row;
     }
 
+    .label-container {
+        display: flex;
+        gap: ${smallPadding};
+        margin-bottom: ${smallPadding};
```
m-akinc, line 33: Looks like this introduces some spacing even when there is no label. I'm not sure how much we care about the no-label use case, but would it be better to apply the `margin-bottom` to `::slotted([slot="label"]), .error-icon`  instead?
image

packages/storybook/src/nimble/radio-group/radio-group.stories.ts
```diff
@@ -105,6 +114,16 @@ export const radioGroup: StoryObj<RadioGroupArgs> = {
...
             control: false
+        },
+        errorText: {
+            name: 'error-text',
+            description: errorTextDescription,
+            table: { category: apiCategory.attributes }
+        },
+        errorVisible: {
+            name: 'error-visible',
+            description: errorVisibleDescription,
```
m-akinc, line 125: `errorVisibleDescription` doesn't say anything about the exclamation point visual, but I think it should.

packages/storybook/src/nimble/radio-group/radio-group.stories.ts
m-akinc, line 1: There might be something to document regarding setting the width of the radio group. For example, if you hadn't set `min-width: 200px`, turning on `error-visible` results in this:
image
I don't know if we should be setting the width to 100% by default, or if we should recommend setting a `min-width` (or just `width`) when the control may go into an error state.

m-akinc (approved):

packages/nimble-components/src/radio-group/styles.ts
mollykreis: I debated this myself. The thing I didn't want to have happen, though, is for the control with no label to shift when `error-visible` was toggled, which is why I decide to leave the space there all the time.

I'll leave the conversation open for others to weigh in.

packages/storybook/src/nimble/radio-group/radio-group.stories.ts
mollykreis: My inclination is that we should leave the width styling up to the client and not try to set defaults here (such as `width: 100%` or `min-width: <some-value>px`). Otherwise, I think it tends to make a client's life harder because they not only need to set the values they want (such as `min-width`), but they also have to set values to override the styling we've set (such as `width: auto`).

I'll leave this conversation open so that others can weigh in with other thoughts of the best practice to follow here.

jattasNI: I suspect that this will have a similar recommendation to [the width discussion on button labels](https://github.com/ni/nimble/pull/2385#discussion_r1745837067) where we decided to not document guidance here and instead do so in pattern guidance.

jattasNI (approved):

packages/nimble-components/src/radio-group/styles.ts
mollykreis: I actually realized that as-is the control still shifts when the error icon comes & goes without a label. I'll fix that, but it'll ultimately mean more spaces is reserved even when there isn't an error or a label.

packages/nimble-components/src/radio-group/template.ts
```diff
@@ -0,0 +1,37 @@
...
+            class="positioning-region ${x => (x.orientation === Orientation.horizontal ? 'horizontal' : 'vertical')}"
+            part="positioning-region"
+        >
+            <slot
+                ${slotted({
+                    property: 'slottedRadioButtons',
+                    filter: elements('[role=radio]')
+                })}
+            ></slot>
+            ${errorTextTemplate}
```
rajsite, line 34: It's an older pattern from before we established the convention but I think the errorTextTemplate should behave like our other title patterns and only set title if the text is overflow hidden.

change/@ni-nimble-components-e5bfd621-8ece-4d9c-a6ea-eb128a70e8ac.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Seems like a pre-exisiting issue but the keyboard behavior when in the initial state of no selected value is broken. Can't tab to the control at all.

packages/nimble-components/src/radio-group/styles.ts
rajsite: +1 for avoiding shifts when transitioning in and out of error states

rajsite (approved):

change/@ni-nimble-components-e5bfd621-8ece-4d9c-a6ea-eb128a70e8ac.json
mollykreis: Turns out that this is an issue with our storybook example, not the component itself. We were setting the value to an invalid option (the string "none") rather than setting the value to `undefined`. I'll fix that as part of this PR.

packages/nimble-components/src/radio-group/template.ts
mollykreis: I'll create an issue for this and address it in a follow up PR so that it can be addressed holistically across nimble without causing churn in this PR.

mollykreis: #2431

PR #2425, opened 2024-10-01
Update HLD with clarification about Angular form validation

m-akinc:
# Pull Request

## 🤨 Rationale

Questions were raised at stand-up about whether setting the `required` attribute on native elements would trigger native HTML form validation that could interfere with Angular's own validation experience.

## 👩‍💻 Implementation

Added clarifying note to HLD.

### DIFF
--- a/specs/required-inputs/README.md
+++ b/specs/required-inputs/README.md
@@ -85,6 +85,8 @@ We will file a Chromium bug for the issue affecting the `nimble-combobox` and `n
 
 The `required` attribute only plays a role in validation for template-driven forms (with reactive forms, it is only used for accessibility purposes). Nimble inputs already support setting the `required` attribute via template, but we will add it to the directive wrappers for completeness.
 
+Note also that Angular's form validation support [automatically disables the native HTML form validation](https://v17.angular.io/api/forms/NgForm#native-dom-validation-ui) that would otherwise be triggered by setting the `required` attribute. This means that our Angular clients are not exposed to any of the native HTML validation bugs listed earlier in this document.
+
 #### Blazor
 
 Typically, form validation is handled via a `DataAnnotationsValidator`, and an input is treated as required if it is bound to a model property annotated with `RequiredAttribute` (i.e. `[Required]`). The Nimble component's `required` attribute is still needed to turn on the visual affordance, though, so we will add it to the Razor API.
--- a/specs/required-inputs/README.md
+++ b/specs/required-inputs/README.md
@@ -85,7 +85,7 @@ When the user attempts to submit a form and a `required` input is missing a valu
 
 ![Missing value indicator](missing-value.png)
 
-We would expect this to already work, since the FAST components we derive from provide forms support. Unfortunately, this support is broken or incomplete for multiple components:
+Because of FAST's form-associated polyfill implementation we end up leveraging a native form control with validation properties forwarded to it. This results in the unexpected behavior of the native form presentation (ARIA and popups) being presented / conflicting with Nimble's defined presentation. The following are specific issues encountered:
 
 - `nimble-combobox`: 
     - In Chrome/Edge, a missing required value properly blocks form submission and updates the control's `validity` flags, but the visual indicator of the validation error is not shown, and instead an error is printed to the console: "An invalid form control with name='' is not focusable." Presumably, something is trying to focus the host element, though it delegates focus to the shadow DOM. The error comes from a call to [`form.requestSubmit()` in the FAST button code](https://github.com/microsoft/fast/blob/913c27e7e8503de1f7cd50bdbc9388134f52ef5d/packages/web-components/fast-foundation/src/button/button.ts#L221), which we cannot debug into. Things work properly in Firefox.
--- a/specs/required-inputs/README.md
+++ b/specs/required-inputs/README.md
@@ -26,18 +26,14 @@ Any other controls that could possibly be considered inputs are out of scope.
 
 ### API
 
-The [native API](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) for marking an input required (which is also the [Angular API](https://v17.angular.io/guide/form-validation#validating-input-in-template-driven-forms)) is:
-
-- `required`: boolean attribute whose presence indicates that a value must be provided to submit
-
-There are behavioral and presentational aspects of making an input participate in validation. In an ideal world, validation-related attributes like `required`, `min`, `max`, etc. would control both behavior and presentation, and frameworks like Angular and Blazor would utilize native form association APIs rather than implementing their own systems. However, given that frameworks like Angular automatically opt out of native form validation ([by setting `novalidate` on the `form` element](https://v17.angular.io/api/forms/NgForm#native-dom-validation-ui)), we don't want our validation presentations (e.g. "required" asterisk, error text/icon, and associated accessibility) coupled to the native validation system. Instead, we introduce a new attribute that controls just the presentation:
+There are behavioral and presentational aspects of making an input participate in validation. In an ideal world, native, validation-related attributes like `required`, `min`, `max`, etc. would control both behavior and presentation, and frameworks like Angular and Blazor would utilize native form association APIs rather than implementing their own systems. However, given that frameworks like Angular automatically opt out of native form validation ([by setting `novalidate` on the `form` element](https://v17.angular.io/api/forms/NgForm#native-dom-validation-ui)), we don't want our validation presentations (e.g. "required" asterisk, error text/icon, and associated accessibility) coupled to the native validation system. Instead, we introduce a new attribute that controls just the presentation:
 
 - `required-visible`: boolean attribute whose presence turns on a visual affordance (red asterisk) and accessible indication (`aria-required="true"`) that an input is required
 
 This approach is consistent with our existing approach to validation error presentation. Rather than using the native validation state (e.g. via the `:invalid` CSS pseudo-selector) to trigger the display of error text or error icon (red exclamation mark), we have the `error-visible` attribute which gives explicit control over the error presentation. Today, `error-visible` does not manage ARIA properly (e.g. setting `aria-invalid` and `aria-errormessage`), which seems to be an oversight.
 
 #### Radio buttons
-`required-visible` will be exposed on radio button groups, but not radio buttons, since the visual only appears on the radio button group label. This is a departure from the the native API, where the `required` attribute is only exposed on radio buttons, but affects the entire radio button group. We are prioritizing an intuitive API over consistency with the native 
+`required-visible` will be exposed on radio button groups, but not radio buttons, since the visual only appears on the radio button group label. This is a departure from the the native API, where the `required` attribute is only exposed on radio buttons, but affects the entire radio button group. We are prioritizing an intuitive API over consistency with the native API.
 
 #### Readonly/disabled required inputs
 
--- a/specs/required-inputs/README.md
+++ b/specs/required-inputs/README.md
@@ -30,11 +30,18 @@ The [native API](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/re
 
 - `required`: boolean attribute whose presence indicates that a value must be provided to submit
 
-We could key off of this existing attribute to show our visual "required" indicator (i.e. red asterisk), but we will instead introduce a new, parallel `required-visible` attribute, for the following reasons:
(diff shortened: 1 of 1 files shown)


specs/required-inputs/README.md
```diff
@@ -85,6 +85,8 @@ We will file a Chromium bug for the issue affecting the `nimble-combobox` and `n
 
 The `required` attribute only plays a role in validation for template-driven forms (with reactive forms, it is only used for accessibility purposes). Nimble inputs already support setting the `required` attribute via template, but we will add it to the directive wrappers for completeness.
 
+Note also that Angular's form validation support [automatically disables the native HTML form validation](https://v17.angular.io/api/forms/NgForm#native-dom-validation-ui) that would otherwise be triggered by setting the `required` attribute. This means that our Angular clients are not exposed to any of the native HTML validation bugs listed earlier in this document.
```
rajsite, line 88: I'm trying to understand the reason to mix the patterns. i.e. we don't style against the `:valid` pseudoselectors for error state visuals but we do key off the required state for the required visual. Have any reasons that we shouldn't follow the same pattern (i.e. have a separate `show-required` attribute like we have `show-error`)?

m-akinc: - `:valid`/`:invalid` are based on native HTML validation. Since Angular validation disables this validation with the `novalidate` attribute, our error styling wouldn't work in Angular if it were based on `:invalid`.
- Despite not using native validation, Angular apps still use the `required` attribute to mark an input required. For template-driven forms, this causes the necessary validation to occur, but even for reactive forms, where validator functions are specified in code, it is expected that you still apply the `required` attribute in the template, for accessibility purposes (see the last sentence under [this heading](https://v17.angular.io/guide/form-validation#validating-input-in-reactive-forms)).
- Since both vanilla HTML and Angular clients already have to add the `required` attribute to their required input elements, it seems convenient and intuitive for our "required" visual (i.e. asterisk) to key off of that. I do not see any benefit to decoupling from that, especially since, from a consistency standpoint, all required inputs in Nimble-styled apps should use our visual affordance. It seems redundant and more error-prone to expect clients to set both `require` and `required-visible` attributes.

specs/required-inputs/README.md
```diff
@@ -26,15 +26,19 @@ Any other controls that could possibly be considered inputs are out of scope.
...
+- Separation of concerns: We want clients to be able to display our "required" visual without triggering everything that comes with setting the `required` attribute. Though vanilla HTML and Angular clients will generally be setting `required` anyway, there may be corner cases where a client wants to avoid doing so.
+
+For radio buttons and radio button groups, our `required-visible` attribute will follow the same rules as the native `required` attribute. Specifically, the attribute only exists on radio button elements, and if any of the radio buttons in a radio button group have the attribute, the radio button group displays the visual indicator. It does not matter if the radio button with `required-visible` is disabled, hidden, etc. It is effectively as if the attribute is on the radio button group, rather than a specific one.
```
rajsite, line 37: For implementation simplicity since we aren't following the `required` semantics for the `required-visible` visual, we could simplify this and just have `required-visible` only on the radio button group element.

specs/required-inputs/README.md
```diff
@@ -43,17 +47,36 @@ This will be implemented using a new `nimble-icon-asterisk` icon ([named consist
...
     </label>
-    ${when(x => x.required, html`
+    ${when(x => x.requiredVisible, html`
         <nimble-icon-asterisk severity="error"></nimble-icon-asterisk>
     `)}
 </div>
 ```
 
-Note that for text-field, text-area, and number-field, we will also remove the visual if the control has the `readonly` attribute set, since [validation is not enforced on a readonly input](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions). This doesn't apply to select, combobox, or radio buttons, because they do not support `readonly`.
+Note that for text-field, text-area, and number-field, the `readonly` attribute will prevent the display of the asterisk icon, since [validation is not enforced on a readonly input](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions). This doesn't apply to select, combobox, or radio buttons, because they do not support `readonly`.
```
rajsite, line 56: 1. It's not clear to me that the required visual should be coupled to the readonly state. Like if the controls are currently marked `disabled` or `readonly` because the user does not have permission to edit them I don't know why the `required-visible` visual would be forced off. 
    
    It could be useful for the user to know what fields are required fields even if they don't have permissions, i.e. they need to ping an admin with permission the details to fill out the form. I think the `required-visible` visual is information on how to fill out the form, not the current state of the form.
2. The mdn docs make the clarification that readonly is not supported by non-text controls because they are semantically equivalent to disabled controls. But I dunno if either of those should impact the `required-visible` visual.

I think this is more an interaction design question. Should the `required-visible` visual always be shown so users understand the details / requirements of the form, not the current state of the form (which is already communicated through different styles for readonly / disabled controls).

specs/required-inputs/README.md
```diff
@@ -43,17 +47,36 @@ This will be implemented using a new `nimble-icon-asterisk` icon ([named consist
...
+
+## Alternative Implementations / Designs
+
+None
+
+## Form validation
+
+We are opting to keep our API separate from the API that actually triggers form validation, but for completeness, an overview of Nimble's support of the `required` attribute follows.
 
 #### Vanilla HTML
```
rajsite, line 81: A comment that I made yesterday was that the `required-visual` state has a visual for sighted users that a field is required and that it should also map to the aria-required stated for aria users to get the same information that a field is required. The discussion below seems to indicate differently and primarily discusses association via the `required` attribute still. Curious how we resolve that.

specs/required-inputs/README.md
rajsite: Discussed more offline about the consistency of decoupling the visuals from the validation states which are currently reflected in the HLD.

specs/required-inputs/README.md
m-akinc: If an input is required, empty, and readonly/disabled, then validation isn't enforced, and it might be confusing to be able to submit a form where a "required" input is missing. That said, I think that's enough of a corner case that we can probably just live with that in exchange for the simplicity of not having unexpected interactions we have to document. I'll make the change.

specs/required-inputs/README.md
m-akinc: You may have said this yesterday and I just forgot, but do you think setting `error-visible` should also set `aria-invalid`? (I would think the answer is no.) My takeaway from our conversation was that we want to decouple control of the visuals from the semantics they represent. This is supported by the fact that the attributes are named `foo-visible`, which implies nothing about functionality/semantics, but only a visual cue. Accessibility of an element's required-ness is coupled to the functional attribute, i.e. `required` (or `aria-required`, which would be bound to the presence of `required` on a host element). I don't see how we can couple the presence of our visual indicator (asterisk) with the accessibile indicator (i.e. screen reader announcing "required") unless we are also coupled to the functional attribute (`required`).

rajsite: > Accessibility of an element's required-ness is coupled to the functional attribute, i.e. required

I don't believe that is the right mental model. The whole point of aria is being able to get the right aria to arbitrary controls, i.e. making a `<div>` aria correct does not involve the `required` attribute, it involves a handful of `aria-` attributes.

rajsite: Chatted offline and discussed the following:

The semantics of the `required` attribute include:
- behavior (form-association)
- presentation (visuals and aria)

In a very ideal world:
- The `required` attribute and others associated with validation, i.e. `min`, `max` would handle both behavior and presentation. 
- Frameworks would rely on native form association and not implement parallel form validity systems

Today as a compromise for frameworks:
- The `required` attribute and others associated with validation, i.e. `min`, `max` would handle just behavior
- The user handles presentation via `error-visible` and `required-visible` (and as a reminder presentation includes visuals and aria)
- An application (and ideally a pattern for consistency across applications) should determine how aligned we are with the semantics of `required`, i,e, when `required` + `disabled` should the presentation be required or not-required
- Frameworks like angular will set the `required` attribute but opt out of the behavior as forms will be marked `novalidate` due to Angular's design. The user will still need to opt-in to presentation via `required-visible`.

Today what will actually happen because of implementation limitations:
- We do not use Form-Association apis directly and rely on FAST's polyfills for forma association
- The polyfill for Form-Association will forward required attributes to internal input elements used by the polyfill which will surface to aria kinda outside our control.
- So `required` may or may not influence presentation aria (due to limitations of the polyfill design) but will not influence presentation visuals. But `required-visible` will influence presentation of both visual and aria.

m-akinc: I have tried to add the most important points related to our reasoning to the HLD.

specs/required-inputs/README.md
```diff
@@ -69,37 +97,23 @@ We would expect this to already work, since the FAST components we derive from p
...
     - Accessibility: the control is not marked/announced as required. The accessible control (i.e. the one defining the `role`) is the `nimble-select` itself. Since this is not a native input, the accessibility tree doesn't see/understand the `required` attribute on it. We need to additionally set `aria-required` on it (bound to `required`).
```
jattasNI, line 101: There are a few more mentions of `required` in this section that need to be updated.

jattasNI (approved):

specs/required-inputs/README.md
```diff
@@ -26,15 +26,26 @@ Any other controls that could possibly be considered inputs are out of scope.
...
+
+- `required-visible`: boolean attribute whose presence turns on a visual affordance (red asterisk) and accessible indication (`aria-required="true"`) that an input is required
+
+This approach is consistent with our existing approach to validation error presentation. Rather than using the native validation state (e.g. via the `:invalid` CSS pseudo-selector) to trigger the display of error text or error icon (red exclamation mark), we have the `error-visible` attribute which gives explicit control over the error presentation. Today, `error-visible` does not manage ARIA properly (e.g. setting `aria-invalid` and `aria-errormessage`), which seems to be an oversight.
```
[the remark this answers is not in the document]
m-akinc: I think just removing the first couple lines (about the native `required` attribute) does the trick.

specs/required-inputs/README.md
m-akinc: I think you're saying that this section is using `required` when it should be `required-visible`. Let me know if I misunderstood. This section is talking about existing support for the native `required` attribute. In truth, none of it is really necessary in this HLD, but I went through the trouble of collecting all this info, so I wanted to record it.

jattasNI: Ah, sorry, I had lost the context of this section in the PR diff view. That makes sense.

But that leads me to a different question. This is the only section I can see that mentions `aria-required`. If we are planning to set that attribute when the client sets `required-visible` then we should mention it outside of the Alternatives section.

rajsite (approved):

specs/required-inputs/README.md
```diff
@@ -26,15 +26,22 @@ Any other controls that could possibly be considered inputs are out of scope.
...
+- `required-visible`: boolean attribute whose presence turns on a visual affordance (red asterisk) and accessible indication (`aria-required="true"`) that an input is required
 
-For radio buttons and radio button groups, only the radio button elements expose the `required` attribute. If any of the radio buttons in a radio button group are marked required, the radio button group is treated as required. It does not matter if the radio button with `required` is disabled, hidden, etc. It is effectively as if the attribute is on the radio button group, rather than a specific one. This matches the behavior of the native `input` with `type="radio"`. 
+This approach is consistent with our existing approach to validation error presentation. Rather than using the native validation state (e.g. via the `:invalid` CSS pseudo-selector) to trigger the display of error text or error icon (red exclamation mark), we have the `error-visible` attribute which gives explicit control over the error presentation. Today, `error-visible` does not manage ARIA properly (e.g. setting `aria-invalid` and `aria-errormessage`), which seems to be an oversight.
```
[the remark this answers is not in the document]
m-akinc: Created #2428

specs/required-inputs/README.md
m-akinc: It is parenthetically mentioned in the description of the `required-visible` attribute.

m-akinc:
@atmgrifter00 and @mollykreis I'm submitting this HLD update without your reviews, because I think there's been enough discussion and scrutiny from Jesse and Milan. If you want to review, feel free to add comments, and I'll address them in a follow-up.

PR #2432, opened 2024-10-09
Error text overflow

### LINKED ISSUE #2431: Error text on components should only have title when content overflows
# 🐛 Bug Report

The error text on nimble components should only have a `title` when the error text content overflows. However, the `errorTextTemplate` currently adds a title all the time, which is both unnecessary and against the current patterns within nimble. The template should use the `overflow` directive to only add a `title` when the error text content doesn't fit.

## 🤔 Expected Behavior

The error text under a component should only have a `title` if the error text content doesn't fit in the allocated width of the control.

## 😯 Current Behavior

The error text under a component always has a `title`.

## 💁 Possible Solution

Use the `overflow` directive.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2431 

## 👩‍💻 Implementation

Updated the error pattern to be a mixin rather than an interface so that controls using the error pattern don't have to redefine the properties/attributes that the pattern uses. Additionally, logic has been added to the mixin (beyond what the interface supported) to use the `overflow` behavior to add a title to the error text when it overflows.

## 🧪 Testing

Manually tested in storybook that all updated components:
- don't have a `title` on the error text when the error text is short
- have a `title` on the error text when the error text overflows
- the overflow on the error text does not affect any existing overflow behavior on the component

New unit tests for the error pattern mixin

## ✅ Checklist


mollykreis:

packages/nimble-components/src/patterns/error/types.ts
m-akinc, line 1: Could/should we replace the `ErrorPattern` interface with a mixin? Would be nice to remove all the declarations of those three properties and automatically enforce the `@attr` and `@observable` expectations.

packages/nimble-components/src/patterns/error/template.ts
```diff
@@ -1,8 +1,14 @@
 import { html } from '@microsoft/fast-element';
 import type { ErrorPattern } from './types';
+import { overflow } from '../../utilities/directive/overflow';
 
 export const errorTextTemplate = html<ErrorPattern>`
-    <div class="error-text" title="${x => x.errorText}" aria-live="polite">
+    <div
+        class="error-text"
+        ${overflow('errorHasOverflow')}
+        title="${x => (x.errorHasOverflow && x.errorText ? x.errorText : null)}"
```
m-akinc, line 9: Probably worthwhile to write unit tests for this behavior against one (not all) of the components that use this pattern. I'm pretty sure we have some existing title overflow behavior tests for labels, etc.

m-akinc (approved):

PR #2435, opened 2024-10-10
Enable table column resizing on touch-based devices

### LINKED ISSUE #1362: Investigate using pointer events for column resizing.
## 📌 User Story

The [current implementation](https://github.com/ni/nimble/pull/1321) for column resizing uses the HTML mouse events for handling the resizing of a column. This has several drawbacks including:
- not being touch-friendly
- not properly dealing with cases where a users drags off-screen, does mouse clicks off-screen, and then moves the mouse back into the browser (without doing extra work).
- necessitating an overlay in the template to intercept mouse events so other interactive elements don't see them while doing a column resize

Using the HTML [pointer events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) seem like they can address all of the above issues without much effort.

m-akinc:
# Pull Request

## 🤨 Rationale

The table columns can't be interactively resized using touch-based interfaces (mobile).
Also fixes #1362

## 👩‍💻 Implementation

We handle dragging column dividers via the mouse events `mousedown`, `mousemove`, and `mouseup`. I replaced that with handling for the `pointerdown`, `pointermove`, and `pointerup` events. We also use `setPointerCapture` to force events to be handled by our divider, which allows us to remove the `glass-overlay` div.

## 🧪 Testing

- Tested manually in local Storybook using Chrome's dev tools: works.
- ... using Firefox dev tools: does not work. The issue is that the column dividers never appear, so they cannot be dragged.
- Tested manually in Storybook on iPhone (Safari): works.
- Updated unit tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
+++ b/change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Enable table column resizing on touch-based devices",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/table/models/table-layout-manager.ts
+++ b/packages/nimble-components/src/table/models/table-layout-manager.ts
@@ -78,7 +78,9 @@ export class TableLayoutManager<TData extends TableRecord> {
         this.initialColumnTotalWidth = this.getTotalColumnFixedWidth();
         this.isColumnBeingSized = true;
         document.addEventListener('mousemove', this.onDividerMouseMove);
-        document.addEventListener('mouseup', this.onDividerMouseUp);
+        document.addEventListener('touchmove', this.onDividerTouchMove);
+        document.addEventListener('mouseup', this.onDividerMoveEnd);
+        document.addEventListener('touchend', this.onDividerMoveEnd);
     }
 
     /**
@@ -96,12 +98,19 @@ export class TableLayoutManager<TData extends TableRecord> {
     }
 
     private readonly onDividerMouseMove = (event: Event): void => {
-        const mouseEvent = event as MouseEvent;
+        this.onDividerMove((event as MouseEvent).clientX);
+    };
+
+    private readonly onDividerTouchMove = (event: Event): void => {
+        this.onDividerMove((event as TouchEvent).targetTouches[0]!.clientX);
+    };
+
+    private onDividerMove(clientX: number): void {
         for (let i = 0; i < this.visibleColumns.length; i++) {
             this.visibleColumns[i]!.columnInternals.currentPixelWidth = this.initialColumnWidths[i]?.initialPixelWidth;
         }
         this.currentTotalDelta = this.getAllowedSizeDelta(
-            mouseEvent.clientX - this.dragStart
+            clientX - this.dragStart
         );
         this.performCascadeSizeLeft(
             this.leftColumnIndex!,
@@ -118,11 +127,13 @@ export class TableLayoutManager<TData extends TableRecord> {
         } else {
             this.table.tableScrollableMinWidth = this.initialTableScrollableMinWidth!;
         }
-    };
+    }
 
-    private readonly onDividerMouseUp = (): void => {
+    private readonly onDividerMoveEnd = (): void => {
         document.removeEventListener('mousemove', this.onDividerMouseMove);
-        document.removeEventListener('mouseup', this.onDividerMouseUp);
+        document.removeEventListener('touchmove', this.onDividerTouchMove);
+        document.removeEventListener('mouseup', this.onDividerMoveEnd);
+        document.removeEventListener('touchend', this.onDividerMoveEnd);
         this.resetGridSizedColumns();
         this.isColumnBeingSized = false;
         this.activeColumnIndex = undefined;
--- a/packages/nimble-components/src/table/models/table-layout-manager.ts
+++ b/packages/nimble-components/src/table/models/table-layout-manager.ts
@@ -57,19 +57,19 @@ export class TableLayoutManager<TData extends TableRecord> {
 
     /**
      * Sets up state related to interactively sizing a column.
-     * @param divider The divider element that was clicked on
+     * @param activeColumnDividerElement The divider element that was clicked on
      * @param pointerId The pointerId of the pointer that started the drag
      * @param dragStart The x-position from which a column size was started
      * @param activeColumnDivider The 1-based index of the divider that was clicked on
      */
     public beginColumnInteractiveSize(
-        divider: HTMLElement,
+        activeColumnDividerElement: HTMLElement,
         pointerId: number,
         dragStart: number,
         activeColumnDivider: number
     ): void {
         this.activeColumnDivider = activeColumnDivider;
-        this.activeColumnDividerElement = divider;
+        this.activeColumnDividerElement = activeColumnDividerElement;
         this.leftColumnIndex = this.getLeftColumnIndexFromDivider(
             this.activeColumnDivider
         );
@@ -85,10 +85,16 @@ export class TableLayoutManager<TData extends TableRecord> {
         this.isColumnBeingSized = true;
         // pointerId of -1 indicates source was synthetic PointerEvent: https://w3c.github.io/pointerevents/#dom-pointerevent-pointerid
         if (pointerId !== -1) {
-            divider.setPointerCapture(pointerId);
+            activeColumnDividerElement.setPointerCapture(pointerId);
         }
-        divider.addEventListener('pointermove', this.onDividerPointerMove);
-        divider.addEventListener('pointerup', this.onDividerPointerUp);
+        activeColumnDividerElement.addEventListener(
+            'pointermove',
+            this.onDividerPointerMove
+        );
+        activeColumnDividerElement.addEventListener(
+            'pointerup',
+            this.onDividerPointerUp
+        );
     }
 
     /**
@@ -142,6 +148,7 @@ export class TableLayoutManager<TData extends TableRecord> {
         this.isColumnBeingSized = false;
         this.activeColumnIndex = undefined;
         this.activeColumnDivider = undefined;
+        this.activeColumnDividerElement = undefined;
         this.visibleColumns = [];
     };
 
--- a/packages/nimble-components/src/table/models/table-layout-manager.ts
+++ b/packages/nimble-components/src/table/models/table-layout-manager.ts
@@ -22,6 +22,7 @@ export class TableLayoutManager<TData extends TableRecord> {
     private initialTableScrollableWidth?: number;
     private initialTableScrollableMinWidth?: number;
     private initialColumnTotalWidth?: number;
+    private activeColumnDividerElement?: HTMLElement;
     private currentTotalDelta = 0;
     private dragStart = 0;
     private leftColumnIndex?: number;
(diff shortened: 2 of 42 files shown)


change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: We have pretty thorough interactive column size tests. Can you work with @atmgrifter00 to buddy and decide appropriate test coverage? Will move back to draft since requesting a buddy review

atmgrifter00: I think it would be sufficient to update this [parameterized test](https://github.com/ni/nimble/blob/29b790b9e5ee15182ee9efed772abef3e3c9b15f/packages/nimble-components/src/table/tests/table-column-sizing.spec.ts#L531) to essentially run the same tests but to use a touch event instead. This would require just updating this page object method to take another argument to determine which type of event to use (and then double the set of tests).

atmgrifter00:
I'm not sure it's worth mentioning in the description, but I'll note that at least with my Firefox dev tools, the touch events did not work (they did in Chrome however as you noted). I think it's also worth a manual test on actual device browsers.

change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
m-akinc: I started to update the tests, but then realized it would be cleaner if we used pointer events (which are meant to unify handling of mouse, touch, and pen inputs) instead of mouse and touch events. Then we don't need additional test cases, as long as we update the page object to send pointer events instead of mouse/touch events.

rajsite: Ahh yea, I remember going down that path and looks like we captured an issue for switching to pointer events: https://github.com/ni/nimble/issues/1362

May need to double check if the issues were ran into with Firefox are still a problem, it's been a while

packages/nimble-components/src/table/styles.ts
```diff
@@ -150,6 +150,7 @@ export const styles = css`
         cursor: col-resize;
         position: absolute;
         z-index: ${ZIndexLevels.zIndex1};
+        touch-action: pan-y;
```
m-akinc, line 153: This makes it so that the column dividers only support panning up/down by touch-dragging. We need this to prevent the browser from hijacking the horizontal drag (sends us the `pointercancel` event) to do a horizontal pan.

change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
m-akinc: The [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1556240) mentioned in that issue is still open, but testing locally, I don't see any weird behavior like drag-release selecting rows. I think we can close our issue with this change.

atmgrifter00: I'd be curious if with this change the `glass-overlay` div can be removed (and the styling related to it) from the table template. It's existence was at least [called out](https://github.com/ni/nimble/issues/1362#issue-1812224551) in the issue for investigating pointer events as something that pointer events could remove the need for.

m-akinc: Once I started calling `setPointerCapture`, the `glass-overlay` div became unnecessary, so I removed it.

atmgrifter00 (approved):

change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
atmgrifter00: While the changes seem fine, obviously the failing tests are a concern.

m-akinc: Using `setPointerCapture` required some changes to the page object. I had missed running those locally, because they were under a second top-level `describe`.

rajsite (changes requested):
A significant number of tests are failing, please reset when PR is passing

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -77,8 +83,12 @@ export class TableLayoutManager<TData extends TableRecord> {
         this.initialTableScrollableMinWidth = this.table.tableScrollableMinWidth;
         this.initialColumnTotalWidth = this.getTotalColumnFixedWidth();
         this.isColumnBeingSized = true;
-        document.addEventListener('mousemove', this.onDividerMouseMove);
-        document.addEventListener('mouseup', this.onDividerMouseUp);
+        // pointerId of -1 indicates source was synthetic PointerEvent: https://w3c.github.io/pointerevents/#dom-pointerevent-pointerid
+        if (pointerId !== -1) {
```
m-akinc, line 87: In tests where we create synthetic `PointerEvent`s, we don't have a real, active pointer whose ID we can put in the event. Passing `1` avoided an error on Chrome, but not Firefox. The best solution I can come up with is to use `-1` and treat that as a signal to skip calling `setPointerCapture`.

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc, line 1: Refactored functions here to encapsulate the event creation involved in testing column sizing.

packages/nimble-components/tsconfig.json
```diff
@@ -13,7 +13,13 @@
         "resolveJsonModule": true,
         "sourceMap": true,
         "inlineSources": true,
-        "lib": ["DOM", "ES2015", "ES2016.Array.Include", "ES2017.Object"],
+        "lib": [
+            "DOM",
+            "DOM.Iterable",
```
m-akinc, line 18: Added `DOM.Iterable` since I wanted the `DOMTokenList` type to support iteration. I don't think there's any reason not to.

rajsite: The spright and storybook tsconfigs should be in sync

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -1051,29 +1062,26 @@ describe('Table Interactive Column Sizing', () => {
...
+                        (): void => {
+                            for (let i = 0; i < dividers.length; i++) {
+                                const classes = dividers[i]!.classList;
+                                if (classes.contains('divider-active')) {
+                                    dividerActiveDividers.push(i);
+                                }
+                                if (classes.contains('column-active')) {
+                                    columnActiveDividers.push(i);
+                                }
+                            }
```
rajsite, line 1082: This doesn't seem very ergonomic as a pageobject api. It does seem like the test had too many element specific details that were not well encapsulated in the page object, but this isn't encapsulating the details in the page object well either. Are there higher level page object apis we can use to encapsulate this better?

packages/nimble-components/src/table/tests/table-column-sizing.spec.ts
```diff
@@ -1013,110 +1013,25 @@ describe('Table Interactive Column Sizing', () => {
...
-            const leftDivider = pageObject.getColumnLeftDivider(0);
-
-            expect(rightDivider).not.toBeNull();
-            expect(leftDivider).toBeNull();
-        });
-
-        it('last column only has left divider', () => {
-            const rightDivider = pageObject.getColumnRightDivider(3);
-            const leftDivider = pageObject.getColumnLeftDivider(3);
+    describe('active divider styling', () => {
```
m-akinc, line 1016: I don't think it's necessary to test every divider in a multi-column table. Also, I think it's fair enough to just test that the styling for a given divider changes when pressed, then again when released.

m-akinc:
@atmgrifter00  Can I have you take another look? I've made enough changes (primarily in the page object and tests) that I think it's warranted.

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -526,39 +528,44 @@ export class TablePageObject<T extends TableRecord> {
...
+     * @param columnIndex The index of the column to the left of the divider to press.
+     */
+    public pressRightColumnDivider(columnIndex: number): void {
+        this.pressColumnDivider(this.getColumnDivider(columnIndex, false));
+    }
+
+    /**
+     * @param columnIndex The index of the column to the right of the divider to press.
+     */
+    public pressLeftColumnDivider(columnIndex: number): void {
```
atmgrifter00, line 541: Why aren't we using any of the left divider page object APIs?

m-akinc: Looks like we still have a couple tests using `dragSizeColumnByLeftDivider`, but you're right--we're not using the `press/releaseLeftColumnDivider` functions I added. The right/left distinction is really only useful from a API ease of use perspective, since there aren't actually separate "left" and "right" dividers. I.e. the left divider of column 2 is the same element as the right divider of column 1. Arguably, we don't need separate left/right versions of the drag-size-column functions either, but then you'd have to identify the divider by index, which is (maybe?) a little harder to understand than, say, "the left divider of column three."

I'll remove the unused functions.

atmgrifter00 (approved):

jattasNI (approved):

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -120,9 +129,15 @@ export class TableLayoutManager<TData extends TableRecord> {
...
+    private readonly onDividerPointerUp = (): void => {
+        this.activeColumnDividerElement!.removeEventListener(
+            'pointermove',
+            this.onDividerPointerMove
+        );
+        this.activeColumnDividerElement!.removeEventListener(
+            'pointerup',
+            this.onDividerPointerUp
+        );
         this.resetGridSizedColumns();
```
rajsite, line 141: Looks like we should clear the activeColumnDividerElement reference as well

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -56,14 +57,19 @@ export class TableLayoutManager<TData extends TableRecord> {
...
     /**
      * Sets up state related to interactively sizing a column.
+     * @param divider The divider element that was clicked on
+     * @param pointerId The pointerId of the pointer that started the drag
      * @param dragStart The x-position from which a column size was started
-     * @param activeColumnDivider The divider that was clicked on
+     * @param activeColumnDivider The 1-based index of the divider that was clicked on
      */
     public beginColumnInteractiveSize(
+        divider: HTMLElement,
```
rajsite, line 66: nit: can we call this input `activeColumnDividerElement` to keep a consistent name through (also better contrasts to the `activeColumnDivider`)

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -570,52 +563,31 @@ export class TablePageObject<T extends TableRecord> {
...
+    /**
+     * @param columnIndex The index of the column whose right divider we're checking.
+     */
+    public columnRightDividerHasActiveStyling(columnIndex: number): boolean {
+        const divider = this.getColumnDivider(columnIndex, false);
+        const currentColor = parseColor(
+            window.getComputedStyle(divider).borderColor
+        );
+        const hoverColor = parseColor(borderHoverColor.getValueFor(divider))!;
+        return currentColor?.equalValue(hoverColor) ?? false;
```
rajsite, line 578: Maybe this should be a non-null assertion. In tests I think it's better to fail faster and I'm guessing there is no circumstance we expect this to every be nulliish.


```suggestion
        return currentColor!.equalValue(hoverColor);
```

packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -570,52 +563,31 @@ export class TablePageObject<T extends TableRecord> {
...
+    /**
+     * @param columnIndex The index of the column whose left divider we're checking.
+     */
+    public columnLeftDividerHasActiveStyling(columnIndex: number): boolean {
+        const divider = this.getColumnDivider(columnIndex, true);
+        const currentColor = parseColor(
+            window.getComputedStyle(divider).borderColor
+        );
+        const hoverColor = parseColor(borderHoverColor.getValueFor(divider))!;
+        return currentColor?.equalValue(hoverColor) ?? false;
```
rajsite, line 590: same as above

rajsite (approved):

packages/nimble-components/src/table/testing/table.pageobject.ts
m-akinc: Actually I should have removed this function since it is unused. Removed now.

change/@ni-nimble-components-3680232d-ba1a-47f0-8cdc-dc6effd176ed.json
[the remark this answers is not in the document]
m-akinc: I think I'll submit this, and if we subsequently find some issue with a browser on Android, I'll file a bug.

PR #2440, opened 2024-10-16
Scrollable tabs

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #1509

## 👩‍💻 Implementation

Adding buttons to template which will show/hide based on width of the container for the tabs either exceeding visible area or not. Layout of tabs is now using `flex` instead of `grid` (the original use of `grid` wasn't clear to me).

#### `ResizeObserver` vs `IntersectionObserver`

I currently have opted to use a `ResizeObserver` for controlling when to show/hide the scroll buttons. The reason for this is two-fold:
1. The observer only fires when the `contentRect` for the element containing the tabs changes. This will occur if the `Tabs` component changes its width, when tabs or added or removed (if all tabs are visible), or when the content of a tab is changed (when all tabs are visible), all of which are valid conditions to re-evaluate whether we need to show or hide the scroll buttons.
2. The implementation is easy to reason about, and there is no extra state necessary.

Two different approaches were used/attempted with the `IntersectionObserver`. The first involved observing each `Tab` element. While this worked, the implementation was cumbersome and harder to reason about, as it involved extra state that the `Tabs` component had to maintain (a map that tracked which tabs were currently visible or not), and "calculating" the visible condition from the state kept in the map. This was required to handle cases such as removing the last tab, where it was only partially visible (and thus was currently showing the scroll buttons), which, when removed, the `IntersectionObserver` would only report a change for the removed tab, but the visibility should still be updated.

The other approach attempted was trying to simply observe the container element of the tabs (whose `root` was simply its  containing element), which wouldn't have required any extra state. However, this requires the observer to fire both when the container element becomes partially visible and when it becomes fully visible. Supplying thresholds of [0, 1] (one pixel and 100% visible), or only 0, did not result in the needed behavior. It would simply fire once, and then never again.

## 🧪 Testing

Added unit tests, and matrix tests.

## ✅ Checklist


packages/nimble-components/src/tabs/testing/tabs.pageobject.ts
m-akinc, line 1: Both these new page objects are almost identical. Could use a shared abstract base class, e.g.
```ts
export abstract class TabsBasePageObject<T> {
    public constructor(
        protected readonly tabsElement: T,
        protected readonly tabElementName: string,
        protected readonly tabPanelElementName?: string
    ) {}
    ...
   // if this.tabPanelElementName === undefined, skip adding/removing panel elements
}

export class TabsPageObject extends TabsBasePageObject<Tabs> {
    public constructor(tabsElement: Tabs) { super(tabsElement, 'nimble-tab', 'nimble-tab-panel'); }
}

export class AnchorTabsPageObject extends TabsBasePageObject<AnchorTabs> {
    public constructor(tabsElement: AnchorTabs) { super(tabsElement, 'nimble-anchor-tab'); }
}
```

packages/nimble-components/src/tabs/tests/tabs.spec.ts
```diff
@@ -8,4 +33,131 @@ describe('Tabs', () => {
...
+            await connect();
+            tabsPageObject = new TabsPageObject(element);
+        });
+
+        afterEach(async () => {
+            await disconnect();
+        });
+
+        it('should not show scroll buttons when the tabs fit within the container', () => {
+            expect(tabsPageObject.areScrollButtonsVisible()).toBe(false);
```
m-akinc, line 69: nit: Up to you, but I think the reporting is _marginally_ nicer when you use `toBeFalse()` and `toBeTrue()`.

packages/nimble-components/src/tabs/tests/tabs.spec.ts
```diff
@@ -8,4 +33,131 @@ describe('Tabs', () => {
...
+            await tabsPageObject.setTabsWidth(1000);
+            await waitForUpdatesAsync(); // wait for the resize observer to fire
+            expect(tabsPageObject.areScrollButtonsVisible()).toBe(false);
+        });
+
+        it('should scroll left when the left scroll button is clicked', async () => {
+            await tabsPageObject.setTabsWidth(300);
+            await waitForUpdatesAsync(); // wait for the resize observer to fire
+            element.activeid = '6'; // scrolls to the last tab
+            const currentScrollLeft = element.shadowRoot!.querySelector('.tablist')!.scrollLeft;
```
m-akinc, line 90: Might as well add a page object method to get the scroll offset instead of using element internal details.

packages/storybook/src/nimble/anchor-tabs/anchor-tabs-matrix.stories.ts
```diff
@@ -29,9 +35,10 @@ export default metadata;
 // prettier-ignore
 const component = (
     toolbar: TabsToolbarState,
-    [disabledName, disabled]: DisabledState
+    [disabledName, disabled]: DisabledState,
+    [widthValue, setWidth]: WidthState
 ): ViewTemplate => html`
-    <${anchorTabsTag} activeid="tab1" style="padding: 15px;">
+    <${anchorTabsTag} activeid="tab1" style="padding: 15px; width: ${() => (setWidth ? widthValue : '')};">
```
m-akinc, line 41: You don't really need the boolean. Could just do: `widthValue ?? ''`.

packages/storybook/src/nimble/anchor-tabs/anchor-tabs-matrix.stories.ts
m-akinc, line 1: For some reason, the cases with the tab toolbar where the tabs should have scroll buttons isn't rendering right:
image
Same for the other matrix story.

packages/nimble-components/src/tabs/template.ts
```diff
@@ -0,0 +1,75 @@
...
+            ${startSlotTemplate(context, definition)}
+            ${when(x => x.showScrollButtons, html<Tabs>`
+                <${buttonTag} 
+                    content-hidden
+                    class="scroll-button left"
+                    appearance="ghost"
+                    tabindex="-1"
+                    @click="${x => x.onScrollLeftClick()}"
+                    ${ref('leftScrollButton')}
+                >
```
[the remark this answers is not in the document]
atmgrifter00: I suppose there could be some debate on this. The buttons are not reachable via the keyboard, and arguably we could/should opt to simply set `aria-hidden="true"`. Having a tooltip on hover for these buttons doesn't seem that helpful, and accidentally pressing a button is ultimately innocuous.

packages/storybook/src/nimble/anchor-tabs/anchor-tabs-matrix.stories.ts
atmgrifter00: Needed a better check in the `ResizeObserver` implementation.

packages/nimble-components/src/tabs/index.ts
```diff
@@ -16,10 +17,76 @@ declare global {
...
+    public readonly rightScrollButton!: Element;
+
+    private readonly tabListResizeObserver: ResizeObserver;
+
     public constructor() {
         super();
         // We disable the built-in active indicator so that we can implement our own
         this.activeindicator = false;
+        this.tabListResizeObserver = new ResizeObserver(entries => {
+            let tabsContainerWidth = entries[0]?.contentRect.width;
```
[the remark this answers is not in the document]
atmgrifter00: Opted for `tabListVisibleWidth`.

packages/nimble-components/src/tabs/template.ts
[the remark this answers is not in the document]
m-akinc: I considered that. A focusable element must _not_ have `aria-hidden="true"`, but the converse is not true, i.e. just because an element isn't focusable doesn't mean it should be hidden from accessibility. From what I was reading, unless the element is purely decorative or would be redundant if announced by a screen reader, we shouldn't hide it. But I'm no expert, so I'm curious what others think, too.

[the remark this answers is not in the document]
atmgrifter00: At the very least I would think whatever decision we arrived at for these buttons would apply to things like the inc/dec buttons on the number field (which do _not_ provide text content).

packages/nimble-components/src/anchor-tabs/specs/README.md
```diff
@@ -57,7 +57,7 @@ We cannot extend FAST's tabs control because it has baked into it the idea that
 -   _CSS Classes and CSS Custom Properties that affect the component_ - None
 -   _Slots_
     -   `start` - content placed to the left of tab list
-    -   `anchortab` - slot for anchor tab elements
+    -   `tab` - slot for anchor tab elements
```
m-akinc, line 60: The one issue with changing this slot name from `"anchortab"` to `"tab"` is that it allows `nimble-tabs` to slot `nimble-anchor-tab` elements (and `nimble-anchor-tabs` to slot `nimble-tab` elements). It's certainly not expected to work, but it might work well enough to mask the client's mistake. One solution is to add a `tabSlotName` property to `Tabs` and `AnchorTabs`. Alternatively, we could check the slotted elements in the `tabsChanged` handler and throw an error, but that feels clunkier.

m-akinc (approved):

packages/nimble-components/src/tabs/template.ts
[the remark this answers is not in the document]
m-akinc: Sir, [I refute you](https://github.com/ni/nimble/blob/9aec468e4b2c8feabebf39a57632faaa28448f1e/packages/nimble-components/src/number-field/index.ts#L79).

packages/nimble-components/src/anchor-tabs/specs/README.md
atmgrifter00: I'll go ahead and implement your `tabSlotName` suggestion.

packages/nimble-components/src/anchor-tabs/index.ts
```diff
@@ -70,14 +76,41 @@ export class AnchorTabs extends FoundationElement {
...
+        });
+    }
+
     /**
      * @internal
      */
     public activeidChanged(_oldValue: string, _newValue: string): void {
         if (this.$fastController.isConnected) {
             this.setTabs();
+            this.activetab?.scrollIntoView({ block: 'nearest' });
```
mollykreis, line 113: If you have a long tab name, scrolling into view from the right (i.e. clicking the left arrow button) feels odd because only the end of the tab name becomes visible. I _think_ if you used `inline: 'start'` here, you might get more intuitive behavior.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
```diff
@@ -66,10 +67,10 @@ describe('AnchorTabs', () => {
             expect(element.tablist.getAttribute('role')).toBe('tablist');
         });
 
-        it('should have a slots named "start", "anchortab", and "end", in that order', () => {
+        it('should have a slots named "start", "tab", and "end", in that order', () => {
             const slots = element.shadowRoot?.querySelectorAll('slot');
             expect(slots![0]?.getAttribute('name')).toBe('start');
-            expect(slots![1]?.getAttribute('name')).toBe('anchortab');
+            expect(slots![1]?.getAttribute('name')).toBe('tab');
```
mollykreis, line 73: I'm probably missing something, but wouldn't this value come from `tabSlotName`, which has a value of `"anchortab"` for the anchor tabs?

packages/nimble-components/src/anchor-tabs/index.ts
```diff
@@ -70,14 +76,41 @@ export class AnchorTabs extends FoundationElement {
...
+                }
+                this.showScrollButtons = tabListVisibleWidth < this.tablist.scrollWidth;
+            }
+        });
+    }
+
     /**
      * @internal
      */
     public activeidChanged(_oldValue: string, _newValue: string): void {
```
mollykreis, line 110: For the anchor tabs, would it make more sense to scroll a tab into view when it is navigated to rather than waiting for the activeid to change? As is, you can't always see what tab you're going to activate when you press Enter.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
atmgrifter00: Nope, I just missed updating this.

packages/nimble-components/src/anchor-tabs/index.ts
atmgrifter00: So, navigating will already bring the tab into view (via arrow keys). All this implementation does is bring the tab into view via a programmatic change to `activeid`.

mollykreis: For the anchor tabs too? That wasn't the behavior I saw in Edge.

packages/nimble-components/src/patterns/tabs/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    .scroll-button.left {
+        margin-right: ${smallPadding};
+    }
+
+    .tablist {
+        display: flex;
+        width: max-content;
+        min-width:
```
rajsite, line 27: missing value

change/@ni-nimble-components-51de64d0-1537-4dce-89b5-e85857542ebb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Scrollable tabs for Tabs and AnchorTabs.",
```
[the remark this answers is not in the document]
atmgrifter00: The problem lies in the lack of the `ResizeObserver` firing when changing from "Tab Two" to "2" (so tabs are now "1", "2", and "Tab Three") . The reason for this is that the content rect for the `tablist` hasn't changed yet, because it has already been reduced because of the presence of the buttons.

Fixing this, I believe, would require a more heavy-handed approach using a `MutationObserver` on the content of each tab (note that I don't think an `IntersectionObserver` would resolve this either). At the moment, I'm not convinced that this is a big enough issue to resolve as this kind of asymmetric behavior isn't that bad.

packages/nimble-components/src/tabs/index.ts
```diff
@@ -1,10 +1,11 @@
+import { observable } from '@microsoft/fast-element';
```
rajsite, line 1: Could you update the tabs stories to show more options configurations to make it easier to experiment. Maybe something similar to select, i.e. simple tabs, many tabs, wide tabs.

packages/nimble-components/src/patterns/tabs/testing/tabsbase.pageobject.ts
```diff
@@ -0,0 +1,112 @@
+import type { Button } from '../../../button';
```
rajsite, line 1: nit: can you name this file `tabs-base.pageobject.ts`. File names are supposed to be kebab-case but realizing that was overlooked in `pageobject`.

packages/nimble-components/src/patterns/tabs/template.ts
```diff
@@ -0,0 +1,67 @@
...
+    FoundationElementTemplate,
+    TabsOptions
+} from '@microsoft/fast-foundation';
+import type { Tabs } from '../../tabs';
+import type { AnchorTabs } from '../../anchor-tabs';
+import { buttonTag } from '../../button';
+import { iconArrowExpanderLeftTag } from '../../icons/arrow-expander-left';
+import { iconArrowExpanderRightTag } from '../../icons/arrow-expander-right';
+
+type TabsOrAnchorTabs = Tabs | AnchorTabs;
```
rajsite, line 15: This doesn't really lend itself to being a reusable pattern. Should we create an interface instead that Tabs and AnchorTabs are implementing?

packages/nimble-components/src/patterns/tabs/styles.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    .tablist {
+        display: flex;
+        width: max-content;
+        align-self: end;
+        overflow-x: scroll;
+        scrollbar-width: none;
+        margin-left: var(--ni-private-tabs-container-padding);
+        margin-right: var(--ni-private-tabs-container-padding);
```
rajsite, line 31: I don't see this defined anywhere and they show up as `undefined` in devtools

packages/storybook/src/nimble/tabs/tabs-matrix.stories.ts
```diff
@@ -45,14 +50,16 @@ const component = (
             Tab Two ${() => disabledName}
         </${tabTag}>
         <${tabTag} hidden>Tab Three</${tabTag}>
```
rajsite, line 52: If I go in devtools and change the tab from hidden to visible then operate the left right arrows I get an error:
arrows-after-hidden

packages/nimble-components/src/tabs/template.ts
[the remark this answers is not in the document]
rajsite: Off the top without searching around I can imagine the low-visibility scenario wanting the labels read (moving the mouse cursor around on the screen with a screen reader). Things that are aria-hidden should not be tab focusable but I don't think the inverse is true, i.e. I don't believe something that is not tab focusable needs to be aria-hidden.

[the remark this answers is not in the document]
rajsite: My guesses might be wrong, the spin button patterns shows the next and previous buttons that are not tab focusable as aria-hidden: https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/datepicker-spinbuttons/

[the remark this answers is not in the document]
rajsite: Actually nope, I misread, the buttons in the patterns are not aria-hidden. They have an aria-label and are not tab focusable (tabindex="-1"). What was aria-hidden was helper text on the page showing what the previous and next values are (ie. current month is october and the spinner hides the preview text of september / november)

change/@ni-nimble-components-51de64d0-1537-4dce-89b5-e85857542ebb.json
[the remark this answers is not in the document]
mollykreis: I agree that this is a pretty low-impact issue that isn't very likely to occur. I'm fine not fixing it if the solution would be overly complex or inefficient.

packages/nimble-components/src/patterns/tabs/template.ts
```diff
@@ -0,0 +1,67 @@
...
+> = (context, definition) => html`
+    <div
+        class="tab-bar"
+    >
+        ${startSlotTemplate(context, definition)}
+        ${when(x => x.showScrollButtons, html<Tabs>`
+            <${buttonTag} 
+                content-hidden
+                class="scroll-button left"
+                appearance="ghost"
```
mollykreis, line 30: You can bind to the value of "ghost" here instead of hard-coding the string: `appearance="${ButtonAppearance.ghost}"`. Same for the scroll right button. Then

change/@ni-nimble-components-51de64d0-1537-4dce-89b5-e85857542ebb.json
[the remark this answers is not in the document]
rajsite: I didn't fully follow the discussion but just to confirm we aren't ending up in a scenario where sometimes the tabs scroll left and right are not visible when they are needed right? It's only that sometimes they are still show when they could instead be hidden?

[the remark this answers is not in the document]
atmgrifter00: Correct, there are scenarios where the scroll buttons aren't necessary but are shown anyway. There is no known scenario where we should be showing the buttons but do not.

mollykreis (approved):

packages/nimble-components/src/patterns/tabs/template.ts
atmgrifter00: Added missing APIs to existing interface I created for tests.

packages/nimble-components/src/tabs/template.ts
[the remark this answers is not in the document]
atmgrifter00: Went ahead and removed the `aria-hidden` settings, and added the labels.

packages/nimble-components/src/label-provider/core/index.ts
```diff
@@ -29,7 +31,9 @@ const supportedLabels = {
     popupIconInformation: popupIconInformationLabel,
     filterSearch: filterSearchLabel,
     filterNoResults: filterNoResultsLabel,
-    loading: loadingLabel
+    loading: loadingLabel,
+    scrollLeftLabel, // 'scrollLeft' is already reserved in the base class
+    scrollRightLabel
```
rajsite, line 36: Can we use an alternate name to avoid the discrepancy (i.e. the other attributes avoid label in the names)?

Left and Right might not be great long-term because of [ltr/rtl issues](https://github.com/openui/open-ui/issues/522) either. Some ideas, scrollForward/scrollBackward, scrollNext/scrollPrevious (the [tabs aria pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) seems to refer to the concepts as next and previous but not sure how much to read into that in terms of i18n)

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
     }
 };
 
+const simpleTabs: readonly TabArgs[] = [
```
rajsite, line 47: Sorry didn't state it more directly but can you make a similar change for anchor-tabs as well?

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
...
+    { title: 'Tab 1', id: '0', disabled: false },
+    { title: 'Tab 2', id: '1', disabled: true },
+    { title: 'Tab 3', id: '2', disabled: false }
```
rajsite, line 50: The id vs the index being off is really confusing. i.e. selecting tab 1 says "content of tab 0":
image

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
     }
 };
 
+const simpleTabs: readonly TabArgs[] = [
+    { title: 'Tab 1', id: '0', disabled: false },
```
rajsite, line 48: These ids are 0-indexed while the activeid choices are 1-indexed, can we make all these consistent one way or another, i.e. only show and use indexes or positions for id / label text?

rajsite:
Looks good overall, will take a last look after story updates and label tweaks

packages/nimble-components/src/label-provider/core/index.ts
rajsite: Just saw a push for the name of `scrollLeftButtonLabel` but I think I'd still suggest one of the other alternatives because:
1. Same rtl/ltr concerns
2. the labels shouldn't be coupled to control usage, i.e. button, in their naming

packages/storybook/src/nimble/tabs/tabs.stories.ts
atmgrifter00: image

packages/nimble-components/src/label-provider/core/index.ts
atmgrifter00: Went with 'scrollForward'/'scrollBackward'.

packages/nimble-components/src/label-provider/core/label-tokens.ts
```diff
@@ -45,3 +45,13 @@ export const loadingLabel = DesignToken.create<string>({
...
     cssCustomPropertyName: null
 }).withDefault(coreLabelDefaults.loadingLabel);
+
+export const scrollBackwardLabel = DesignToken.create<string>({
+    name: 'scroll-backward-label',
+    cssCustomPropertyName: null
+}).withDefault(coreLabelDefaults.scrollBackwardLabel);
+
+export const scrollForwardLabel = DesignToken.create<string>({
+    name: 'scroll-right-button-label',
```
rajsite, line 55:
```suggestion
    name: 'scroll-forward-label',
```

packages/nimble-components/src/label-provider/core/index.ts
```diff
@@ -65,6 +69,12 @@ export class LabelProviderCore
     @attr({ attribute: 'loading' })
     public loading: string | undefined;
 
+    @attr({ attribute: 'scroll-backward' })
+    public scrollBackward: string | undefined;
+
+    @attr({ attribute: 'scroll-forward' })
+    public scrollForward: string | undefined;
```
rajsite, line 76: Reminder, the label public api changes require angular and blazor updates. Can do in follow-on PR or in this one.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
```diff
@@ -380,4 +378,122 @@ describe('AnchorTabs', () => {
...
+            await disconnect();
+        });
+
+        it('should not show scroll buttons when the tabs fit within the container', () => {
+            expect(tabsPageObject.areScrollButtonsVisible()).toBeFalse();
+        });
+
+        it('should show scroll buttons when the tabs overflow the container', async () => {
+            await tabsPageObject.setTabsWidth(300);
+            await waitForUpdatesAsync(); // wait for the resize observer to fire
```
rajsite, line 414: Can we move the waitForUpdateAsync into the setTabsWidth page object? Looks like every instance of its call is followed by a manual waitForUpdatesAsync in the tests.

Also setTabsWidth already has a waitForUpdatesAsync call in it. So if a second one is required in there after width updates it's fine buut can document once there instead of in every usage below.

packages/nimble-components/src/label-provider/core/index.ts
rajsite: If  done in this PR I guess reset me. Maybe better to do in a follow-up

packages/storybook/src/nimble/tabs/tabs.stories.ts
rajsite, line 47: when using `as const` we should avoid overriding that more specific type with a manual weaker type
```suggestion
const simpleTabs = [
```

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
     }
 };
 
+const simpleTabs: readonly TabArgs[] = [
+    { title: 'Tab 1', id: '1', disabled: false },
+    { title: 'Tab 2', id: '2', disabled: true },
+    { title: 'Tab 3', id: '3', disabled: false }
+] as const;
+
+const wideTabs: readonly TabArgs[] = [
```
rajsite, line 53:
```suggestion
const wideTabs  = [
```

packages/storybook/src/nimble/anchor-tabs/anchor-tabs.stories.ts
```diff
@@ -27,6 +29,49 @@ interface ToolbarArgs {
     toolbar: boolean;
 }
 
+const simpleTabs: readonly AnchorTabArgs[] = [
```
rajsite, line 32: same as below, I won't tag all the spots but lets avoid manual weaker types

```suggestion
const simpleTabs = [
```

packages/storybook/src/nimble/anchor-tabs/anchor-tabs.stories.ts
```diff
@@ -27,6 +29,49 @@ interface ToolbarArgs {
...
+    { title: 'Short', id: '2', disabled: false, href: 'https://nimble.ni.dev' }
+] as const;
+
+const manyTabs: AnchorTabArgs[] = [];
+for (let i = 0; i < 100; i++) {
+    manyTabs.push({
+        title: `Tab ${i}`,
+        id: `${i}`,
+        disabled: false,
+        href: 'https://nimble.ni.dev'
```
rajsite, line 65: in each of these urls that target nimble.ni.dev can you add a query parameter so it's possible to see the url is updating quickly

```suggestion
        href: `https://nimble.ni.dev?tabid=${i}`
```

packages/storybook/src/nimble/tabs/tabs.stories.ts
rajsite: Re-opening because this does not seem addressed consistently, see the following:
image

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
...
     // prettier-ignore
     render: createUserSelectedThemeStory(html`
         <${tabsTag} activeid="${x => x.activeid}">
-            <${tabTag} id="1" disabled">Tab One</${tabTag}>
-            <${tabTag} id="2">Tab Two</${tabTag}>
-            <${tabTag} id="3">Tab Three</${tabTag}>
-            <${tabPanelTag}>Content of the first tab</${tabPanelTag}>
-            <${tabPanelTag}>Content of the second tab</${tabPanelTag}>
-            <${tabPanelTag}>Content of the third tab</${tabPanelTag}>
+        ${repeat(x => (tabSets[x.tabsType] as TabArgs[]), html<TabArgs>`
```
rajsite, line 88: Seems like there is a visual difference when activeid is set to a disabled tab in anchor tabs vs tabs. Looks like anchor tabs has a disabled state visual

atmgrifter00: We discussed this off-line, and this is an existing distinction between the two. Ultimately, for the `Tabs` component we want to display the underline on a disabled, active tab. We will capture his in a separate task, as this will require more significant changes.

packages/storybook/src/nimble/anchor-tabs/anchor-tabs.stories.ts
```diff
@@ -27,6 +29,49 @@ interface ToolbarArgs {
...
+
+const manyTabs: AnchorTabArgs[] = [];
+for (let i = 0; i < 100; i++) {
+    manyTabs.push({
+        title: `Tab ${i}`,
+        id: `${i}`,
+        disabled: false,
+        href: `https://nimble.ni.dev?tabid=${i}`
+    });
+}
```
rajsite, line 67: @atmgrifter00 it's still inconsistent Remember that the activeid parameter can be used with each of the configurations. It seems like 1-indexing is preeferred so lets do it here to be consistent

```suggestion
for (let i = 1; i <= 100; i++) {
    manyTabs.push({
        title: `Tab ${i}`,
        id: `${i}`,
        disabled: false,
        href: `https://nimble.ni.dev?tabid=${i}`
    });
}
```

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
...
+] as const;
+
+const manyTabs: TabArgs[] = [];
+for (let i = 0; i < 100; i++) {
+    manyTabs.push({
+        title: `Tab ${i}`,
+        id: `${i}`,
+        disabled: false
+    });
+}
```
rajsite, line 74:
```suggestion
for (let i = 1; i <= 100; i++) {
    manyTabs.push({
        title: `Tab ${i}`,
        id: `${i}`,
        disabled: false
    });
}
```

change/@ni-nimble-components-51de64d0-1537-4dce-89b5-e85857542ebb.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
atmgrifter00: I am. I would say that its behavior before wasn't exactly desirable anyway.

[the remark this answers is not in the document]
atmgrifter00: Yeah, I discussed this with @mollykreis, and we both felt that, while nice, this expanded the scope enough to not bother with for now.

packages/nimble-components/src/label-provider/core/index.ts
```diff
@@ -10,7 +10,9 @@ import {
     popupIconInformationLabel,
     filterSearchLabel,
     filterNoResultsLabel,
-    loadingLabel
+    loadingLabel,
+    scrollBackwardLabel,
```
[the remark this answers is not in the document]
rajsite: The quarterly maintenance task isn't something to rely on (it's a backstop to forgetting), should expect to do it in a follow-up PR soon so that the new needed localized strings propagate around sooner.

packages/storybook/src/nimble/anchor-tabs/anchor-tabs.stories.ts
```diff
@@ -27,6 +29,49 @@ interface ToolbarArgs {
...
+        disabled: false,
+        href: 'https://nimble.ni.dev'
+    },
+    {
+        title: 'Tab 2 that is also too long but disabled',
+        id: '2',
+        disabled: true,
+        href: 'https://nimble.ni.dev'
+    },
+    { title: 'Short', id: '2', disabled: false, href: 'https://nimble.ni.dev' }
```
rajsite, line 56:
```suggestion
    { title: 'Short', id: '3', disabled: false, href: 'https://nimble.ni.dev' }
```

packages/storybook/src/nimble/tabs/tabs.stories.ts
```diff
@@ -42,18 +44,58 @@ const metadata: Meta<TabsArgs> = {
...
 
+const simpleTabs = [
+    { title: 'Tab 1', id: '1', disabled: false },
+    { title: 'Tab 2', id: '2', disabled: true },
+    { title: 'Tab 3', id: '3', disabled: false }
+] as const;
+
+const wideTabs = [
+    {
+        title: 'Tab 1 that is too long to fit in the drop down width',
```
rajsite, line 55:
```suggestion
        title: 'Tab 1 that is too long and should probably be shorter but is not',
```

packages/storybook/src/nimble/tabs/tabs-matrix.stories.ts
```diff
@@ -28,12 +28,17 @@ const metadata: Meta = {
...
 // prettier-ignore
 const component = (
     toolbar: TabsToolbarState,
-    [disabledName, disabled]: DisabledState
+    [disabledName, disabled]: DisabledState,
+    widthValue: WidthState
+
 ): ViewTemplate => html`
-    <${tabsTag} style="padding: 15px;">
+    <${tabsTag} style="padding: 15px; width: ${() => (widthValue ?? '')};">
```
rajsite, line 41: This will render out as invalid / incomplete css, ie: `style="width: "` can we make it render out correctly for either case?

packages/storybook/src/nimble/anchor-tabs/anchor-tabs-matrix.stories.ts
```diff
@@ -29,9 +32,10 @@ export default metadata;
 // prettier-ignore
 const component = (
     toolbar: TabsToolbarState,
-    [disabledName, disabled]: DisabledState
+    [disabledName, disabled]: DisabledState,
+    widthValue: WidthState
 ): ViewTemplate => html`
-    <${anchorTabsTag} activeid="tab1" style="padding: 15px;">
+    <${anchorTabsTag} activeid="tab1" style="padding: 15px; width: ${() => (widthValue ?? '')};">
```
rajsite, line 38: This will render out as invalid / incomplete css, ie: `style="width: "` can we make it render out correctly for either case?

packages/nimble-components/src/utilities/tests/timeout.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Timeout for use in async tets.
+ */
+export async function timeout(ms = 0): Promise<void> {
+    await new Promise((resolve, _reject) => {
+        window.setTimeout(() => {
+            // eslint-disable-next-line no-void
+            resolve(void 0);
+        }, ms);
+    });
```
rajsite, line 10: Can we just do `resolve()`? I don't see why we need to pass `undefined`. If not for some weird reason can we pass `undefined` instead of `void 0`? That's a weird old hack we haven't needed since like ES3

```suggestion
    await new Promise((resolve) => {
        window.setTimeout(() => resolve(), ms);
    });
```

packages/nimble-components/src/utilities/tests/timeout.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Timeout for use in async tets.
+ */
+export async function timeout(ms = 0): Promise<void> {
```
rajsite, line 4: Since this is being used in a pageobject and pageobjects are in the public API, it cannot be in a tests folder since those are removed from build. This file should be moved to the `src/utilities/testing` folder:
https://github.com/ni/nimble/tree/main/packages/nimble-components/src/utilities/testing

rajsite (approved):

packages/nimble-components/src/utilities/tests/timeout.ts
rajsite: Actually maybe just move is to this file? It seems to be a catch all file and already has some timing functions init: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/utilities/testing/component.ts#L68

rajsite: to align with the other naming conventions (waitForEvent, and waitAnimationFrame) lets also change it to `waitTimeout`

packages/nimble-components/src/tabs/index.ts
```diff
@@ -15,11 +17,83 @@ declare global {
...
+     */
+    public override activeidChanged(oldValue: string, newValue: string): void {
+        super.activeidChanged(oldValue, newValue);
+        this.activetab?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
+    }
+
+    /**
+     * @internal
+     */
+    public onScrollLeftClick(): void {
```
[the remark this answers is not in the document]
jattasNI: Do we want to update the calls to `scrollIntoView()` too or do we prefer those happen instantaneously? Some of them look like programmatic APIs so I could see that being preferable, just double checking.

packages/nimble-components/src/anchor-tabs/tests/anchor-tabs.spec.ts
```diff
@@ -374,4 +372,111 @@ describe('AnchorTabs', () => {
             expect(document.activeElement).toBe(tab(0));
         });
     });
+
+    describe('scroll buttons', () => {
+        async function setup(): Promise<Fixture<AnchorTabs>> {
+            return await fixture<AnchorTabs>(
+                html`<nimble-anchor-tabs activeid="tab-two">
```
rajsite, line 379: Avoid hard coding the tag names

packages/nimble-components/src/utilities/tests/timeout.ts
atmgrifter00: Btw...you have to pass `undefined`.

packages/nimble-components/src/tabs/tests/tabs.spec.ts
```diff
@@ -1,7 +1,153 @@
+import { html } from '@microsoft/fast-element';
 import { Tabs, tabsTag } from '..';
+import { fixture, Fixture } from '../../utilities/tests/fixture';
+import { waitForUpdatesAsync } from '../../testing/async-helpers';
+import { TabsPageObject } from '../testing/tabs.pageobject';
+
+async function setup(): Promise<Fixture<Tabs>> {
+    const viewTemplate = html`
+        <nimble-tabs>
```
rajsite, line 9: Avoid hard coded tag names

packages/nimble-components/src/tabs/index.ts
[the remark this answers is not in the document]
atmgrifter00: That's my preference. When you are navigating with the arrow keys, having each press cause an animation seems distracting.

change/@ni-nimble-components-51de64d0-1537-4dce-89b5-e85857542ebb.json
jattasNI, line 1: I tested on macOS Firefox and Safari and everything looks great. The scrolling animation is a nice improvement. Trackpad scrolling works well too.

On iOS Safari and Firefox things mostly worked well too, including pan scrolling and momentum scrolling. I saw a minor bug where tapping the scroll buttons while a scroll animation was still in progress would allow you go past the end of the last tab, showing blank space while the tabs scrolled out of the viewport and are no longer visible. I don't think that's worth investigating: you can restore your state easily by scrolling back.

jattasNI (approved):

Issue #2351, opened 2024-08-20
Add new color token for Soliton Calendar Dark UI Text


NIbokeefe:
# 🧹 Tech Debt

This request is to add a new color token for Dark UI text in the Calendar (work for Soliton team).

Token name: PowerGreen-Text-100
Color: Hex #98F5CA

![Image](https://github.com/user-attachments/assets/917a638c-e53c-49f9-a657-32eaa5ad586c)
![Image](https://github.com/user-attachments/assets/65ed4eb7-89de-4dfd-b981-423d3f4481a5)

sandeep-ni:
The new color token has been added.

Token name: PowerGreenDark50
Color: Hex #98F5CA

PR #2442, opened 2024-10-18
Add new base token and update theme aware token for calendar event text

sandeep-ni:
# Pull Request

## 🤨 Rationale

Added a new base color token required to create calendar theme-aware token for the event text.

Nimble issue: https://github.com/ni/nimble/issues/2351

## 👩‍💻 Implementation

Added the new base token - PowerGreenText with hex #98F5CA
Updated the existing theme aware token - calendarEventStaticFontColor

## 🧪 Testing

NA

## ✅ Checklist

### DIFF
--- a/packages/nimble-tokens/source/styledictionary/properties/colors.json
+++ b/packages/nimble-tokens/source/styledictionary/properties/colors.json
@@ -152,6 +152,10 @@
       "value": "#356D54ff",
       "type": "color"
     },
+    "PowerGreenText": {
+      "value": "#98F5CAff",
+      "type": "color"
+    },
     "Black22": {
       "value": "#E6E6E6ff",
       "type": "color"
--- a/packages/nimble-tokens/source/styledictionary/properties/colors.json
+++ b/packages/nimble-tokens/source/styledictionary/properties/colors.json
@@ -152,7 +152,7 @@
       "value": "#356D54ff",
       "type": "color"
     },
-    "PowerGreenText": {
+    "PowerGreenDark50": {
       "value": "#98F5CAff",
       "type": "color"
     },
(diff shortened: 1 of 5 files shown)


packages/nimble-tokens/source/styledictionary/properties/colors.json
```diff
@@ -152,6 +152,10 @@
       "value": "#356D54ff",
       "type": "color"
     },
+    "PowerGreenText": {
```
vivinkrishna-ni, line 155: I'm not sure if we should use the word "Text" in the colors.json file since all the others seem to be the names of colors. 

So should it be like the below as I see it is a lighter version of `PowerGreen`?
```suggestion
    "PowerGreenLight": {
```

However, looking into the linked issue, it seems the suggested toke name is 
```suggestion
    "PowerGreenText100": {
```

vivinkrishna-ni (approved):

packages/nimble-tokens/source/styledictionary/properties/colors.json
sandeep-ni: As you've mentioned `PowerGreenText100` is the suggested name. But considering that we'll be appending the opacity at the end when we use a percent of the given color like 10, 30 or 110 and not for 100 percent, I deferred adding it.

Having this comment open to get thoughts from other reviewers as well.

rajsite: Chatted with @NIbokeefe and for the styledictionary token lets go with `PowerGreenDark50` (representing 50 opaque and no transparency as all tokens in this file are `xff` for alpha).

He shared this image which shows the token name concepts of Opaque (against a background, i.e. Light or Dark) and Transparency (the actual alpha used):

image

sandeep-ni: I've updated the name to PowerGreenDark50. Thanks for sharing that info on naming the base tokens, @rajsite.

rajsite (approved):

jattasNI (approved):

PR #2445, opened 2024-10-22
Consistent tags

rajsite:
# Pull Request

## 🤨 Rationale

- Stops using hard coded element tag names
- Removes unnecessary test asserting tag name string value
- A couple tests used a fixture with invalid tag names, i.e. `nimble-xmark-icon`

## 👩‍💻 Implementation

See above

## 🧪 Testing

Rely on CI.

## ✅ Checklist


jattasNI (approved):
I was gonna suggest documenting this policy but it turns out [we already do](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#compose-elements-into-a-new-component)! 🤓

PR #2455, opened 2024-10-28
Pin FAST foundation and element versions until new versions vetted

m-akinc:
# Pull Request

## 🤨 Rationale

Unit tests on Firefox seem to be sporadically timing out when we update to the latest exports of `fast-foundation` (2.50.0) and `fast-element` (1.14.0). We want to pin our dependency versions to the prior versions until we are confident that the new versions do not introduce real problems for clients.

## 👩‍💻 Implementation

Pinned `fast-element` to 1.13.0 and `fast-foundation` to 2.49.6 in the three packages that reference them.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/package.json
+++ b/packages/nimble-components/package.json
@@ -60,8 +60,8 @@
   "homepage": "https://github.com/ni/nimble#readme",
   "dependencies": {
     "@microsoft/fast-colors": "^5.3.1",
-    "@microsoft/fast-element": "^1.12.0",
-    "@microsoft/fast-foundation": "^2.49.6",
+    "@microsoft/fast-element": "1.12.0",
+    "@microsoft/fast-foundation": "2.49.6",
     "@microsoft/fast-web-utilities": "^6.0.0",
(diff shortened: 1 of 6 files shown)


m-akinc:
Forgot to update the lock file. Will push soon.

packages/nimble-components/package.json
```diff
@@ -60,8 +60,8 @@
   "homepage": "https://github.com/ni/nimble#readme",
   "dependencies": {
     "@microsoft/fast-colors": "^5.3.1",
-    "@microsoft/fast-element": "^1.12.0",
-    "@microsoft/fast-foundation": "^2.49.6",
+    "@microsoft/fast-element": "1.12.0",
```
rajsite, line 63: https://github.com/ni/nimble/blob/ecf525b763bd98eb8eae78a1e8460fe80c113b38/package-lock.json#L5468

rajsite (changes requested):
Packages were incorrectly pinned to versions not in the lock file

rajsite (approved):

PR #2457, opened 2024-10-29
Add a table page object function to get an icon's severity

m-akinc:
# Pull Request

## 🤨 Rationale

I'd like to be able to verify that icon severities are being set correctly in the Assets table.

## 👩‍💻 Implementation

Added a new page object function that gets the icon severity for a particular cell. It is very similar to the existing function for getting a cell's icon/spinner tag name.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/testing/table.pageobject.ts
+++ b/packages/nimble-components/src/table/testing/table.pageobject.ts
@@ -20,6 +20,7 @@ import type { Button } from '../../button';
 import { Icon } from '../../icon-base';
 import { Spinner, spinnerTag } from '../../spinner';
 import { borderHoverColor } from '../../theme-provider/design-tokens';
+import { IconSeverity } from '../../icon-base/types';
 
 /**
  * Summary information about a column that is sorted in the table for use in the `TablePageObject`.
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/table/testing/table.pageobject.ts
```diff
@@ -207,6 +208,18 @@ export class TablePageObject<T extends TableRecord> {
...
+    public getRenderedMappingColumnCellIconSeverity(
+        rowIndex: number,
+        columnIndex: number
+    ): IconSeverity {
+        const iconOrSpinner = this.getRenderedMappingColumnIconOrSpinner(
+            this.getRenderedCellView(rowIndex, columnIndex)
+        );
+        return iconOrSpinner instanceof Icon
+            ? iconOrSpinner.severity
+            : undefined;
```
rajsite, line 220: Other page object methods generally throw in scenarios where preconditions fail (i.e. this is supposed to be used on cells that have icons and instead non-icon cells are silently ignored).

Also this approach makes it not possible to distinguish between an existing icon of severity default (which is the value undefined) and an incorrect mapping (where this method returns undefined). Think it should be consistent in strictness and not ambiguous in return value.

m-akinc: The call to `getRenderedMappingColumnIconOrSpinner` throws if the cell isn't rendering an icon or spinner, but I agree with your point that we should also throw if it is a spinner.

rajsite (approved):

PR #2467, opened 2024-11-11
Checkbox error state HLD update

mollykreis:
# Pull Request

## 🤨 Rationale

During development of #2018, I realized that implementing the error state of the checkbox the way it was spec'ed out would be much more difficult than originally thought because making only the label and square checked indicator clickable would break a number of tests (inside & outside of nimble) and would be against ARIA guidelines. Therefore, we decided to keep the effort of the feature more in line with what our original estimates were by deferring the work to update which portion of the control was clickable.

## 👩‍💻 Implementation

Update HLD

## 🧪 Testing

None

## ✅ Checklist


m-akinc (approved):

atmgrifter00 (approved):

jattasNI (approved):
This is more of a UX / scoping decision than a technical one so I added @fredvisser as a reviewer and I think it's ok not to wait for @rajsite. I'm also ok with verbal signoff from Fred.

jattasNI:
> This is more of a UX / scoping decision than a technical one so I added @fredvisser as a reviewer and I think it's ok not to wait for @rajsite. I'm also ok with verbal signoff from Fred.

I chatted with @fredvisser offline and he said he is ok with this change.

PR #2470, opened 2024-11-13
Upgrade Storybook to 8.4

jattasNI:
# Pull Request

## 🤨 Rationale

I saw that Storybook released a new minor version which is supposed to have a bunch of improvements. An exciting sounding one: "We cut bundle size in half" though the package-lock has about as many additions as subtractions

## 👩‍💻 Implementation

Followed [their docs](https://storybook.js.org/blog/storybook-8-4/) which basically involved:
1. Run `npx storybook@latest upgrade`
2. Agree to the prompts to auto-update stuff
3. The tool also suggested we might be able to remove our React dependency but I didn't try this because I know we create some custom React components so it seems correct for us to depend on it

## 🧪 Testing

1. Manually inspected our stories including the ones that use custom React components
2. verified Chromatic did a full rebuild, no turbosnaps, and there are no diffs

## ✅ Checklist


packages/storybook/.storybook/main.js
```diff
@@ -67,9 +68,7 @@ export const framework = {
     name: getAbsolutePath('@storybook/html-webpack5')
 };
 
-export const docs = {
-    autodocs: false
-};
+export const docs = {};
```
rajsite, line 71: Guessing we don't need this export at all anymore?

packages/storybook/package.json
```diff
@@ -22,32 +22,33 @@
...
-    "@storybook/addon-interactions": "^8.2.4",
-    "@storybook/addon-links": "^8.2.4",
+    "@storybook/addon-a11y": "^8.4.3",
+    "@storybook/addon-actions": "^8.4.3",
+    "@storybook/addon-docs": "^8.4.3",
+    "@storybook/addon-essentials": "^8.4.3",
+    "@storybook/addon-interactions": "^8.4.3",
+    "@storybook/addon-links": "^8.4.3",
+    "@storybook/addon-mdx-gfm": "^8.4.3",
     "@storybook/addon-webpack5-compiler-swc": "^1.0.2",
```
rajsite, line 37: nit: may as well reflect what is resolved in the lock file since doing that to other deps.
```suggestion
    "@storybook/addon-webpack5-compiler-swc": "^1.05",
```

rajsite (approved):

packages/storybook/.storybook/main.js
```diff
@@ -67,9 +68,6 @@ export const framework = {
     name: getAbsolutePath('@storybook/html-webpack5')
 };
```
rajsite, line 70:
```suggestion
```

Issue #2469, opened 2024-11-13
Issues with two-way binding of NimbleNumberField


epetersoni:
# 🐛 Bug Report

Using a `NimbleNumberField` with `Value` and `ValueChanged` both set is sometimes causing the value displayed to the user to get out of sync with the C# value it should be bound to.

## 💻 Repro or Code Sample

Display a `NimbleNumberField` that has `Value` and `ValueChanged` set.
Have another control that updates the C# variable that `Value` points to, e.g. a NimbleButton.

At first if you click the button it updates the value displayed in the number field.
Once you use the mouse / keyboard to directly update the `NimbleNumberField` value, clicking the button no longer updates the displayed value, even though the associated C# value has been updated.

Note that if I replace the number field with a `<input type="numeric">` component and use `@bind:get` and `@bind:set` to bind the value, that fixes the issue.

## 🤔 Expected Behavior

Programmatic updates to the bound C# value should continue to update the number field display.

## 😯 Current Behavior

The number field stops updating on programmatic updates after making manual updates in the UI.

## 💁 Possible Solution

It sounds like updating how the NimbleNumberField binds its display value to work like other Nimble fields should fix it.

## 🔦 Context

We have an array editor that uses an HTML table filled with a NimbleNumberField control on each row. We provide buttons to copy the values in the array or paste values into the array.

On paste, only NimbleNumberFields where I have not manually changed the value get updated, even though the underlying C# array has the correct values.

## 🌍 Your Environment

* OS & Device: Windows PC
* Browser: WPF WebView2

atmgrifter00:
Fixed by [this PR](https://github.com/ni/nimble/pull/2471).

PR #2471, opened 2024-11-14
Fix blazor number field

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #2469

## 👩‍💻 Implementation

Most of our input controls bind their `current-value` attribute in the razor template. The `NumberField`, however, was binding its `value` attribute. There doesn't seem to be a valid reason for this, and I have switched it be in-line with the other input controls.

## 🧪 Testing

Wrote an acceptance test.

## ✅ Checklist


jattasNI (approved):

change/@ni-nimble-blazor-c81c50dc-52b2-4900-9f60-01674f5cec09.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Bind to current-value instead of value.",
```
msmithNI, line 3: Can you specify that this is for the number field here?

PR #2477, opened 2024-11-19
Editable table cells IxD spec

fredvisser:
# Pull Request

## 🤨 Rationale

Clients would like to use the Nimble table to allow their users to interactively edit tabular string and numeric data. This IxD spec proposes a set of interactions that builds upon the existing keyboard navigation work. 

## 👩‍💻 Implementation

TODO: Add discussion of save workflows

## ✅ Checklist


packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+#### Error State
+
+![Editable cell error states](./spec-images/editable-cell-error-data.png)
+
+### ARIA Considerations
+
+-   Maintain compatibility with existing keyboard navigation behavior
+-   Consider resolving [existing ARIA gaps](https://github.com/ni/nimble/issues/2285)
+
+### Mouse Interactions
```
jattasNI, line 51: It would be good to be explicit about mouse cursor states. Right now the table mostly just shows the arrow cursor but switches to the hand/pointer for interactive things like action menus and menu button columns. Should we indicate when a cell is editable somehow?

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+![Editable cell mouse navigation](./spec-images/editable-cell-mouse-navigation.png)
+
+-   Editable table cells show the same row hover state as non-editable cells on mouse hover.
+    -   I.e. When hovering over a row, the action menu button appears. Depending on the table's selection mode, the row may be highlighted.
+-   Single clicking editable cell shows the cell focus state.
```
jattasNI, line 57: Should this be the case for all tables and cells, or just editable ones? It differs from the current behavior for non-editable tables.

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+-   Double clicking the editable cell transforms the cell into an input control in the focus state.
+-   Clicking away from a focused editable cell sets the value and transforms it into the Cell focus state.
```
jattasNI, line 59: nit: throughout the document you switch between capitalizing the C "cell focus state" and "Cell focus state". I'd vote for the former but just prefer consistency.
```suggestion
-   Clicking away from a focused editable cell sets the value and transforms it into the cell focus state.
```

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+-   When a cell has Edit focus, pressing `ESCAPE` reverts any value change and transforms it into the Cell focus state.
+
+![Editable cell data error](./spec-images/editable-cell-error-data.png)
+
+-   When a cell has Edit focus, pressing `ENTER` validates the value, and if invalid, transforms the cell into the Cell error state.
+
+### Out of scope
+
+The client application is responsible for defining and implementing workflows for adding, deleting, and moving rows, as well as saving data.
```
jattasNI, line 82: We probably need to think through situations like if editing a value causes its sort order or grouping to change. Do we have a recommended experience? Are we ok with the value you just edited disappearing off screen? Should we scroll to it? Wait to apply changes until they finish a batch of edits?

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+![Editable table anatomy](./spec-images/editable-cell-anatomy.png)
+
+| Element               | Description                                                              |
+| --------------------- | ------------------------------------------------------------------------ |
+| Static cell value     | Non-editable display of cell value                                       |
+| Action button         | Button to open cell context menu. Appears on row hover or keyboard focus |
+| Editable cell control | Text or numeric control that supports entering or editing cell data      |
+
+## Behavior
```
jattasNI, line 36: What's the scope of "editability"? 

1. Do apps mark the entire table as editable? Specific rows? Specific columns? Specific cells?
2. Does every column type support editing? For pretty much everything except text columns that opens the door to questions about what the editor should look like and what kind of validation is available.

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+### States
+
+![Editable cell states](./spec-images/editable-cell-states.png)
+
+#### Error State
```
jattasNI, line 42: Do apps need to be able to do their own validation and explicitly put a cell in an error state? Or is the table completely in control of the error state?

jattasNI (commented):
We probably still need a discussion about the interactions for entering the edit state and navigating between cells, but capturing some other feedback in the meantime.

packages/nimble-components/src/table/specs/editable-table-ixd.md
atmgrifter00, line 57: How should this play with row selection? Is it similar to clicking on an action menu, where that also selects the row?

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+-   Maintain compatibility with existing keyboard navigation behavior
+-   Consider resolving [existing ARIA gaps](https://github.com/ni/nimble/issues/2285)
+
+### Mouse Interactions
+
+![Editable cell mouse navigation](./spec-images/editable-cell-mouse-navigation.png)
+
+-   Editable table cells show the same row hover state as non-editable cells on mouse hover.
+    -   I.e. When hovering over a row, the action menu button appears. Depending on the table's selection mode, the row may be highlighted.
```
atmgrifter00, line 56: Is there value in having hover over an interactive cell differ from that of a non-editable cell? What other indication would the user have that a cell/column was editable?

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+-   Single clicking editable cell shows the cell focus state.
+-   Double clicking the editable cell transforms the cell into an input control in the focus state.
```
msmithNI, line 58: In Meyer's prototype, once a cell has been clicked once (and focused), a 2nd single-click enters edit mode at that point. We should decide if we like that behavior, or if it should still take a double-click when a cell is focused. (Assuming we keep the cell focus state for mouse-clicked cells)

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: Requirements-wise: Client apps need to be able to specify their own cell validation rules (e.g., max/min/invalid characters/format) at the column level and they should get that feedback while editing the cell.

packages/nimble-components/src/table/specs/editable-table-ixd.md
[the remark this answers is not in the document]
fredvisser: I'd strongly prefer that we maintain the edit state even if the cell scrolls out of view, such that it can scroll back into view in the same state. While it's true the menu button dropdown closes on scroll, setting or reverting an edited cell value is a much more aggressive action in response to something as simple as scrolling.

Do we have any options for persisting this type of state as the cell is scrolled out of view?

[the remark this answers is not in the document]
fredvisser: If that proves too expensive, we should prevent the table from scrolling while a cell is in edit focus.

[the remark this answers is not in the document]
msmithNI: Discussed this a bit with Fred today. Maintaining an edited cell as 'still being edited' after a scroll will probably be difficult due to virtualization - as soon as you scroll more than 1 row's height, even if the row is still visible in the table viewport, the row elements have been re-bound to different data at that point. We can remember which row/cell was being edited (and the current edited cell value), but we'd need to reapply that to a new row visual again (since now a different row visual represents the data row that was being edited, after the scroll), and there might be visual artifacts/ visible flashing if we do that.

Assuming we can prevent mousewheel scrolling the table while editing a cell, that's probably the simplest approach (least work to implement).

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: The least disruptive option is keep the new edited value in the same relative location after completing the edit, even if the column sort would move the row.

If that's not possible, and the row needs to immediately sort to the correct position, some affordance to indicate that the row is moving would be helpful. E.g., an animation of the row moving to it's new location. We shouldn't scroll to the new position.

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Usage
+
+**When to use:**
+
+-   When users need to quickly and efficiently update data directly within the table.
+-   When inline editing can improve the user experience by reducing the need for separate forms or dialogs.
+-   When the table is used in scenarios where data accuracy can be ensured through validation and error handling mechanisms.
+
+**When not to use:**
```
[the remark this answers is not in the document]
fredvisser: Added to both "When not to use" and "Out of scope"

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: Specific columns and we'll need to start with text and numeric, and add select and checkbox later. We shouldn't need to support every column type (at least right away.)

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: Added to wireframes and the "Mouse Interactions" section

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: Current proposal is to immediately enter *edit focus* on single-click of an editable cell. 

This resolves some issues with nested focus states (action menu within a focused cell), and reduces the clicks required to edit the table, but we should discuss any other drawbacks.

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: Proposing to use text and hand mouse cursors to indicate that cells are editable. See the "Mouse Interactions" section and the associated wireframes.

packages/nimble-components/src/table/specs/editable-table-ixd.md
fredvisser: I'm proposing that clicking on an editable cell will _not_ select the row. The row hover behavior is consistent across static and editable cells, but you'll need to click on a static cell (or action menu or the checkbox) to select a row.

fredvisser: This _should_ be the case for all tables and cells… Please let me know what is inconsistent with the current behavior for non-editable tables.

packages/nimble-components/src/table/specs/editable-table-ixd.md
atmgrifter00: This seems like a call to essentially surface the [appropriate HTML `input` attributes](https://www.w3schools.com/html/html_form_attributes.asp) on the relevant column types. I'm concerned if we feel that there are validation APIs beyond those captured by the `input` attributes. The "format" you mention is one that may fall under that category. For such cases, I would prefer that clients manage their own custom validation logic, and simply rely  on the information provided via events that fire during input, and then call general APIs to set the cell in an error state.

packages/nimble-components/src/table/specs/editable-table-ixd.md
```diff
@@ -0,0 +1,129 @@
...
+
+![Editable cell mouse navigation](./spec-images/editable-cell-mouse-navigation.png)
+
+- Editable table cells show the same row hover state as non-editable cells on mouse hover.
+    - I.e. When hovering over a row, the action menu button appears. Depending on the table's selection mode, the row may be highlighted.
+- The cursor will change from an *arrow cursor* when hovering over editable cells
+    - To a *text cursor* when hovering over cells with editable text or numeric controls
+    - To a *hand cursor* when hovering over select or boolean controls
+- Single clicking editable cell shows the *edit focus* state.
+- Clicking away from a focused editable cell sets the value and removes *edit focus*.
```
atmgrifter00, line 82: To be crystal clear, when you click on a non-editable cell while an editable cell has focus, _and_ the row is selectable, do we expect the row to become selected? This is what I expect to happen with no extra work.

packages/nimble-components/src/table/specs/editable-table-ixd.md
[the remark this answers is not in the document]
jattasNI: After reading the [DFF tabular data spec](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/816389?_a=files&path=/docs/design-documents/Platform/Dynamic-Form-Fields/DynamicFormFields-field-type-table.md) I am more nervous that adding/deleting/reordering rows or columns is out of scope and expected to be implemented by clients. It's clear that they expect those operations to be possible and I think they'll seek our guidance about how to do them if we don't provide them. Re-opening in case we want to revisit this now, but I'm also fine to move it to another venue.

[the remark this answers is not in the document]
fredvisser: I'm still comfortable leaving interactive adding/deleting/reordering out of scope for this spec. Apps can do a reasonable job of handling these requirements on their own today, and I'm convinced this would blow up the scope of these table changes.

packages/nimble-components/src/table/specs/editable-table-ixd.md
jattasNI: Right now there is no visible mouse focus state for non-editable cells. The keyboard focus state (green border) only appears when you tab to a cell.

So I was interpreting this as saying that clicking on a cell should put it in the keyboard focus state, which feels like a big behavior change. I could be misunderstanding though.

packages/nimble-components/src/table/specs/spec-images/select-editable-cell-states.png
aangerhofer, line 1: IMO we should show the down error on hover to prime the user for what will happen when they click

packages/nimble-components/src/table/specs/spec-images/checkbox-editable-cell-states.png
aangerhofer, line 1: Show what this looks like in context when not checked

aangerhofer (changes requested):

rajsite:
Been idle for several months moving back to draft

PR #2478, opened 2024-11-19
Add error state to nimble-checkbox

### LINKED ISSUE #2091: nimble-checkbox has incorrect height
# 🐛 Bug Report

The nimble-checkbox height currently is driven by the line height of its text, which isn't the way the component should be. As a result, the checkbox without text is 16px tall, and the checkbox with text is 18px tall.

## 🤔 Expected Behavior

The checkbox should have a height that isn't controlled by the line height of the text. According to @NIbokeefe, we also need to make sure it is a multiple of 4, since that is standard for all nimble sizes. We may need to work with him to nail down the exact size it should be (20px?).

## 😯 Current Behavior

The height of the checkbox is the maximum of the checkbox square size (16px) and the text's line height (by default 18px).

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

This is part of #2018 and also resolves #2091

## 👩‍💻 Implementation

The `nimble-checkbox` now implements `ErrorPattern` to add `error-text` and `error-visible` to its API.

The bulk of the implementation is updating the template and styles of the checkbox. These changes include:
- Adding error icon/text to the template with appropriate styles
- Fix the wrapping behavior to keep the checkbox aligned with the top line of text rather than centered within the control
- Update the checkbox to have a default min-height of `controlHeight` (see #2091)

## 🧪 Testing

- Existing unit tests pass
- Extended matrix tests to cover more cases -- error states and wrapping text

## ✅ Checklist


msmithNI (approved):

change/@ni-nimble-components-d948f8c3-7e90-4ec6-b1f3-468c20ad234d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jattasNI, line 2: Am I correct that the styling changes will impact clients by adding additional space around their checkboxes? For example, in the example app it looks like the checkbox has some extra space on all sides so no longer left aligns with the label above it.
https://60e89457a987cf003efc0a5b-ndvckxqkhv.chromatic.com/example-client-app/#/customapp
https://nimble.ni.dev/storybook/example-client-app/#/customapp

Is that an intentional design change? If so then I think we should:
1. mention it in the change file comment
2. spot check a few SLE apps to ensure it looks ok (I think Fred has complained that a lot of slide outs already have bad checkbox layouts so maybe this is a low bar)

I don't think it merits being marked as a breaking change or doing additional notifications on Teams, but we could consider those too if it seems disruptive.

mollykreis: The padding is coming from the [visual design specs](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=8729-97569&node-type=frame&t=F4QOnnPRawZhNAV2-0). Perhaps we need to have a discussion with Brandon if we don't agree with the design.

rajsite: I think either way I'd vote to split it out into a separate PR and not make the change here. If we do end up wanting it, I'd argue implementation-wise for the checkbox (which is a standalone control) that the padding / margin would be responsibility of the thing that is laying it out.

The padding is for a grouped, labelled set of checkboxes, not for every possible composition of a checkbox. We'd have a separate PR to capture the layout guidance in storybook docs or create something like a checkbox-group to help layout and we would also want to update radio button group at the same time.

mollykreis: I'll pull that change out of this PR.

mollykreis: The left padding has now been removed from the checkbox. The square control is back to being flush with the left edge of the `nimble-checkbox`.

rajsite (approved):

change/@ni-nimble-components-d948f8c3-7e90-4ec6-b1f3-468c20ad234d.json
jattasNI: The Chromatic diffs still show some additional vertical padding. Is that a client-facing change too? If so, what makes us ok with the vertical padding change but not the left padding change?

rajsite: The vertical control size change makes the control height consistent with out other controls. The left padding addition makes the start of the control bounds inconsistent with out other controls (and even if the design is correct, I'm arguing that is not the implementation we should make with the discussion above)

jattasNI (approved):

change/@ni-nimble-components-d948f8c3-7e90-4ec6-b1f3-468c20ad234d.json
jattasNI: The updated change file comment looks good, I think that should give sufficient breadcrumbs if clients see this change and wonder what happened.

PR #2485, opened 2024-12-04
NPM Dependency Updates for December 2024

rajsite:
# Pull Request

## 🤨 Rationale

Updates dependencies manually to address build issues revealed by #2482 #2481 #2480.
Found that just updating dependencies by rebuilding the lock fails the build which means the [repo is in a bad state](https://github.com/ni/nimble/issues/1479#issuecomment-2237523576) and that should be addressed quickly.

## 👩‍💻 Implementation

- Update major deps and rebuild lock
- Found that the now [deprecated tiptap link validate](https://tiptap.dev/docs/editor/extensions/marks/link#validate-deprecated) function is expected to still function with [backwards compat](https://github.com/ueberdosis/tiptap/pull/5812) however there seems to be a regression as it [fails our build by allowing autolinking of unexpected protocols](https://github.com/ni/nimble/actions/runs/12110590706/job/33761233944#step:17:613).
- Switched from tiptap link validate to [shouldAutoLink](https://tiptap.dev/docs/editor/extensions/marks/link#shouldautolink) as described by docs. That seems to not fail the build.
- Pushed the min semver range version up for all the tiptap / prosemirror packages as otherwise it seems to[ fail build when type checking libraries](https://github.com/ueberdosis/tiptap/issues/5867).
   
   **NOTE**: It seems to be transitive dependency type issues that will fail builds, I think it's possible that minimal upgrades done by renovate may hit issues in app updates. If an app adopting latest these changes fails to build it may need to rebuild the lock files in a clean workspace to align transitive deps on latest versions and avoid library type check issues.
- Updates to [prettier 3.4](https://prettier.io/blog/2024/11/26/3.4.0.html) resulting in lots of changes.

## 🧪 Testing

Rely on CI.

## ✅ Checklist


vivinkrishna-ni (approved):

packages/nimble-components/src/switch/styles.ts
```diff
@@ -108,7 +108,8 @@ export const styles = css`
                     var(--ni-private-switch-height) - var(
                             --ni-private-switch-indicator-size
                         )
-                ) / 2
+                ) /
+                2
```
atmgrifter00, line 112: Do we prefer this formatting?

rajsite: I'm OOO and did not designate an owner to get the PR through. Since the repo is in a bad state (see PR description) and touching lots of files, I'll volunteer you to resolve the discussions and shepard through / own finding someone to delegate to.

jattasNI: @atmgrifter00 I agree it looks goofy but I don't think it's worthwhile to block the dependency update on this, particularly it only affects a tiny number of files. If we find it bugs us as we do development we can either mark sections to be ignored by prettier, see if this setting is configurable, or re-open the debate about prettier more generally.

jattasNI (approved):

PR #2486, opened 2024-12-04
Correct the position of error text on the nimble-checkbox

mollykreis:
# Pull Request

## 🤨 Rationale

The vertical spacing between the checkbox and its error text is incorrect, so this PR is fixing it. According to [the visual design spec](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-63171&node-type=frame&t=ODewc6hH8Cq1LC4n-0), the error text on the checkbox should be 2px below the checkbox's height (likely 32px or 24px based on the control height token). However, it was incorrectly implemented to put the error text 2px below the checkbox control (i.e. the square indicator).

## 👩‍💻 Implementation

Update the template & styling for the checkbox to position the error-text correctly.

## 🧪 Testing

Manually tested
Verified chromatic diffs look correct

## ✅ Checklist


jattasNI (approved):
@rajsite is OOO, no need for any further review

Issue #2488, opened 2024-12-09
[Rich text] Test failures in firefox


mollykreis:
# 🐛 Bug Report

There are test failures in Firefox associated with the rich text editor. [See this test run](https://github.com/ni/nimble/actions/runs/12242337641/job/34149426850?pr=2432)

Failure excerpt:
```
[test-concurrent:nimble-components] [test-firefox]   Markdown serializer
[test-concurrent:nimble-components] [test-firefox]     Serialize rich text editor content to its respective markdown
[test-concurrent:nimble-components] [test-firefox]       ✗ Mention node (51ms)
[test-concurrent:nimble-components] [test-firefox] 	InvalidStateError: Selection.collapseToEnd: No selection range exists thrown
[test-concurrent:nimble-components] [test-firefox] 
[test-concurrent:nimble-components] [test-firefox]       ✗ Multiple Mention node of same type (49ms)
[test-concurrent:nimble-components] [test-firefox] 	InvalidStateError: Selection.collapseToEnd: No selection range exists thrown
```

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

rajsite:
Not currently running tests in Firefox on CI, disabled tests will need to be reevaluated generally if enabled again. Closing till then.

PR #2489, opened 2024-12-09
Disable failing tests in Firefox

mollykreis:
# Pull Request

## 🤨 Rationale

Disabling tests that I saw failing on an unrelated pipeline

Issues have been created to track these failing tests: #2488, #2490, #2491

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b8f24320-d835-42a0-b1d8-1b183206df94.json
+++ b/change/@ni-nimble-components-b8f24320-d835-42a0-b1d8-1b183206df94.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Disable failing rich-text tests in Firefox",
+  "comment": "Disable failing tests in Firefox",
   "packageName": "@ni/nimble-components",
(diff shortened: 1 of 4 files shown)


change/@ni-nimble-components-b8f24320-d835-42a0-b1d8-1b183206df94.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Capturing the plan discussed at standup:
We have a ton of bandwidth available to investigate these failures before the end of the year, but @msmithNI might have time next week. If we see more Firefox failures after this PR goes in then we should consolidate the bugs to a single issue and look into FF-specific fixes rather than test-specific fixes. Maybe we need to shard our FF tests or file memory leak bugs to Mozilla, for example. But for now it seems worth disabling these tests in the hope that the pipeline becomes more stable. Either way we left the bugs with the triage tag so we look at them in the new year.

jattasNI (approved):

PR #2495, opened 2024-12-12
Fork number-field and text-field templates

mollykreis:
# Pull Request

## 🤨 Rationale

Fork the FAST templates for the number-field and the text-field in preparation for adding a required-visible icon to the label (see #2100) 

## 👩‍💻 Implementation

Forked the FAST templates for the number-field and the text-field with no functional changes.

## 🧪 Testing

Verified tests still pass and no visual changes were introduced to those components.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b40541e2-0d52-423d-a278-53d68df3aa8e.json
+++ b/change/@ni-nimble-components-b40541e2-0d52-423d-a278-53d68df3aa8e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fork number-field and text-field templates",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/number-field/tests/number-field.foundation.spec.ts
+++ b/packages/nimble-components/src/number-field/tests/number-field.foundation.spec.ts
@@ -1,0 +1,916 @@
+import { DOM } from '@microsoft/fast-element';
+import { NumberField } from '..';
+import { template } from '../template';
+import { fixture } from '../../utilities/tests/fixture';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const FASTNumberField = NumberField.compose({
+    baseName: 'number-field',
+    template,
+});
+
+async function setup(props?: Partial<NumberField>): Promise<{
+    element: NumberField,
+    connect: () => Promise<void>,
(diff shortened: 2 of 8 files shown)


packages/nimble-components/src/number-field/tests/number-field.foundation.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+}> {
+    const { element, connect, disconnect, parent } = await fixture(FASTNumberField());
+
+    if (props) {
+        // eslint-disable-next-line guard-for-in
+        for (const key in props) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any
+            (element as any)[key] = (props as any)[key].toString();
+        }
+    }
```
mollykreis, line 28: Note: This function broke quite a few lint rules we have in place for nimble. I chose to mostly disable those rules and keep the original code rather than completely re-write this function. Let me know if you think it would be better to find a different way to write this code.

change/@ni-nimble-components-b40541e2-0d52-423d-a278-53d68df3aa8e.json
```diff
@@ -0,0 +1,7 @@
+{
```
mollykreis, line 1: There is an unrelated chromatic diff present in the nimble-select (only in Chrome) that I'm curious if anyone has any ideas about. I'm not sure how this code change could have caused that diff or why the diff is only present in the Dark Theme in Chrome, but the diff has persisted across two runs of the chromatic tests.

mollykreis:

packages/nimble-components/src/number-field/tests/number-field.foundation.spec.ts
msmithNI: If you have ideas on another way to write it without disabling the lint rules, it's probably worth doing so (I think we try to fix 'small' lint issues rather than suppress, when we bring FAST tests over).

msmithNI (approved):

packages/nimble-components/src/number-field/tests/number-field.foundation.spec.ts
mollykreis: I fixed one of the lint rules, but the collection of issues around assigning an arbitrary property from `props` to `element` seems much more challenging to fix (at least without jumping through a bunch of typing hoops).

rajsite (approved):

PR #2496, opened 2024-12-12
Required indicator on radio-group, select, combobox, and text-area

mollykreis:
# Pull Request

## 🤨 Rationale

Add `required-visible` attributes to the radio-group, select, combobox, and text-area components as part of #2100.

There is still a need to add this attribute to the text-field and number-field. That will be done once the templates for those components have been forked (#2495).

## 👩‍💻 Implementation

Within the `patterns` directory, I created a mixin that can be used on elements that should have a `requiredVisible` property and `required-visible` attribute. Additionally, shared styling was added for the `requiredVisible` state and a function was created to generate the template for an element's label. This is a function rather than a constant template because each element already had an existing label structure that I didn't want to change as part of this PR.

The template of each element included in this PR was also updated to set `aria-required` based on the value of `requiredVisible`.

I also updated the styling of icons to solve a problem I encountered when shrinking the asterisk icon down to 5x5.

## 🧪 Testing

- Manually tested
- New matrix tests
- Unit tests for the new mixin
- New unit tests for the ARIA behavior of each element

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
+++ b/change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Add `required-visible` attribute to combobox, select, and text-area",
+  "comment": "Add `required-visible` attribute to radio-group, combobox, select, and text-area",
   "packageName": "@ni/nimble-components",
   "email": "20542556+mollykreis@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
+++ b/change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add `required-visible` attribute to combobox, select, and text-area",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/icon-base/styles.ts
+++ b/packages/nimble-components/src/icon-base/styles.ts
@@ -21,6 +21,7 @@ export const styles = css`
     }
 
     .icon {
+        display: contents;
         width: 100%;
         height: 100%;
     }
@@ -42,7 +43,7 @@ export const styles = css`
     }
 
     .icon svg {
-        display: block;
+        display: inline-flex;
         fill: ${iconColor};
         width: 100%;
         height: 100%;
--- a/packages/nimble-components/src/icon-base/styles.ts
+++ b/packages/nimble-components/src/icon-base/styles.ts
@@ -42,6 +42,7 @@ export const styles = css`
     }
 
     .icon svg {
+        display: block;
         fill: ${iconColor};
         width: 100%;
         height: 100%;
--- a/packages/nimble-components/src/icon-base/styles.ts
+++ b/packages/nimble-components/src/icon-base/styles.ts
@@ -22,8 +22,6 @@ export const styles = css`
 
     .icon {
         display: contents;
-        width: 100%;
-        height: 100%;
     }
 
     :host([severity='error']) {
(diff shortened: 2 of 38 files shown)


mollykreis:
@msmithNI - will you buddy this PR for me?

packages/storybook/src/nimble/radio-group/radio-group.stories.ts
```diff
@@ -124,6 +128,10 @@ export const radioGroup: StoryObj<RadioGroupArgs> = {
             name: 'error-visible',
             description: errorVisibleDescription,
             table: { category: apiCategory.attributes }
+        },
+        requiredVisible: {
```
msmithNI, line 132: Missing `name: 'required-visible'` here

msmithNI (approved):

fredvisser:
Looks like https://github.com/nodejs/node/issues/56127 highlights the build issue we have here.

packages/nimble-components/src/icon-base/styles.ts
```diff
@@ -21,6 +21,7 @@ export const styles = css`
     }
 
     .icon {
+        display: contents;
         width: 100%;
         height: 100%;
```
rajsite, line 26: Think we can remove these since it is now `display: contents` and doesn't have its own box layout

rajsite (approved):

packages/nimble-components/src/patterns/required-visible/template.ts
```diff
@@ -0,0 +1,20 @@
+import { ViewTemplate, html, when } from '@microsoft/fast-element';
+import { iconAsteriskTag } from '../../icons/asterisk';
+import type { RequiredVisiblePattern } from './types';
+
+/* eslint-disable @typescript-eslint/indent */
+export function createRequiredVisibleLabelTemplate(
```
jattasNI, line 6: We haven't done a good job of this in other patterns, but I think it'd be helpful to have a brief comment explaining what the pattern is for. Something like "Given the template for a control label, adds an icon next to the label to indicate that the input is required."

change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Just want to double check that we're good with the wrapping behavior and test coverage for it.

image

I suppose we could argue about whether the asterisk should go at the end of the last line but the current behavior matches the spec and looks good so I'd rather not re-open that discussion.

For test coverage, I noticed that the matrix tests for some controls have long enough labels to verify this behavior but some controls don't. Since it's a shared implementation I don't think every control needs to test it. But it would be ideal to have some guarantee that it will continue to be tested even if we change the layout or cases covered by the existing matrix stories. What do you think about adding a dedicated test story for the pattern? We sort of have precedence for this with the [anchor patterns story](https://60e89457a987cf003efc0a5b-ntsdlpsnpz.chromatic.com/?path=/story/tests-anchor-patterns--anchor-patterns).

jattasNI (approved):
No need to wait for @fredvisser to review; I think he might be OOO this week and we can address any feedback from him in follow up PRs.

change/@ni-nimble-components-b67fa692-e6d5-42b5-b494-b284fe5833fd.json
mollykreis: @rajsite commented on the chromatic diff with the same question about where the asterisk should go when the label wraps. I didn't see a design in the spec for the label wrapping. I don't mind taking a look at putting the asterisk at the end of the last line, but I think that will be significantly different styling & layout than what is currently implemented. It may also be a question for @fredvisser or @NIbokeefe as to whether the label should wrap versus have an ellipsis when the label is longer than the control.

I can add a dedicated test story for the pattern.

rajsite: Can't say I feel strongly enough about it to block the PR and try and design + solve it. Seems like overlong / wrapping labels are probably a pattern we would not encourage and aren't a design / pattern that is clearly defined yet (like you mentioned wrapping vs ellipses on labels, etc).

Under the assumption this is mostly an edge case (don't expect long labels, use a tooltip etc to store extra info) I'm okay with the PR as-is and saving additional label work for follow-up / clear use case and design for overlong labels.

mollykreis: I'm creating a story that includes the 4 components in this PR with both long and short labels with `required-visible` set.

I'll also create a nimble issue for updating component labels to not wrap.

mollykreis: Created #2505

PR #2497, opened 2024-12-17
Fix Storybook build

fredvisser:
# Pull Request

## 🤨 Rationale

A combination of node 22.12.0 and our Storybook configuration fails. See https://github.com/nodejs/node/issues/56127 for more information. 

A quick fix is to remove some [unnecessary](https://storybook.js.org/docs/faq#how-do-i-fix-module-resolution-in-special-environments) code that the Storybook migration tools added to our configuration a few releases ago. 

## 👩‍💻 Implementation

- Removed uses of the `getAbsolutePath()` macro
- Removed the @storybook/addon-mdx-gfm Storybook add-on which, per the build logs, isn't necessary to have installed to support GitHub flavored markdown

## 🧪 Testing

- Verified that markdown tables still render in Storybook (GFM feature)
- The build no longer breaks

## ✅ Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7307,23 +7307,6 @@
         }
       }
     },
-    "node_modules/@storybook/addon-mdx-gfm": {
-      "version": "8.4.6",
-      "resolved": "https://registry.npmjs.org/@storybook/addon-mdx-gfm/-/addon-mdx-gfm-8.4.6.tgz",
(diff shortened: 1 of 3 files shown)


rajsite (approved):
Validated build worked with node 22.7.0 and failed with node 22.12.0 on main.
Validated changes here build with node 22.12.0.

PR #2506, opened 2025-01-06
Required indicator on text-field and number-field

mollykreis:
# Pull Request

## 🤨 Rationale

Add required-visible attributes to the text-field and number-field as part of https://github.com/ni/nimble/issues/2100.

## 👩‍💻 Implementation

Used the newly established pattern for required-visible to add `required-visible` to the text-field and number-field.

## 🧪 Testing

- New unit tests for `aria-required` on both components
- Manually tested both components in storybook
- Storybook updates
- Matrix test updates
- Updated required-visible story

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
+++ b/packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
@@ -137,9 +137,9 @@ const component = (
     </${textFieldTag}>
 `;
 
-export const lightThemeEditableEnabledWithoutButtons: StoryFn = createFixedThemeStory(
+export const lightThemeNotRequiredEditableEnabledWithoutButtons: StoryFn = createFixedThemeStory(
     createMatrix(component, [
-        requiredVisibleStates,
+        [requiredVisibleStates[0]],
         [readOnlyStates[0]],
         [disabledStates[0]],
         [actionButtonStates[0]],
@@ -153,9 +153,9 @@ export const lightThemeEditableEnabledWithoutButtons: StoryFn = createFixedTheme
     lightThemeWhiteBackground
 );
 
-export const lightThemeEditableEnabledWithButtons: StoryFn = createFixedThemeStory(
+export const lightThemeNotRequiredEditableEnabledWithButtons: StoryFn = createFixedThemeStory(
     createMatrix(component, [
-        requiredVisibleStates,
+        [requiredVisibleStates[0]],
         [readOnlyStates[0]],
         [disabledStates[0]],
         [actionButtonStates[1]],
@@ -169,9 +169,9 @@ export const lightThemeEditableEnabledWithButtons: StoryFn = createFixedThemeSto
     lightThemeWhiteBackground
 );
 
-export const lightThemeEditableDisabledWithoutButtons: StoryFn = createFixedThemeStory(
+export const lightThemeNotRequiredEditableDisabledWithoutButtons: StoryFn = createFixedThemeStory(
     createMatrix(component, [
-        requiredVisibleStates,
+        [requiredVisibleStates[0]],
         [readOnlyStates[0]],
         [disabledStates[1]],
         [actionButtonStates[0]],
@@ -185,9 +185,9 @@ export const lightThemeEditableDisabledWithoutButtons: StoryFn = createFixedThem
     lightThemeWhiteBackground
 );
 
-export const lightThemeEditableDisabledWithButtons: StoryFn = createFixedThemeStory(
+export const lightThemeNotRequiredEditableDisabledWithButtons: StoryFn = createFixedThemeStory(
     createMatrix(component, [
-        requiredVisibleStates,
+        [requiredVisibleStates[0]],
         [readOnlyStates[0]],
         [disabledStates[1]],
         [actionButtonStates[1]],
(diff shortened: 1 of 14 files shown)


packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -479,6 +507,391 @@ export const colorThemeReadOnlyDisabledWithoutButtons: StoryFn = createFixedThem
 
 export const colorThemeReadOnlyDisabledWithButtons: StoryFn = createFixedThemeStory(
     createMatrix(component, [
+        [requiredVisibleStates[0]],
```
mollykreis, line 510: I've updated the text-field matrix tests to run all of them with `requiredVisible` set to `true` and `false`. However, this state in the label has no impact on much of the control (such as how action buttons are rendered). Are owners fine with the number of new stories this created, or would it be better to be more selective in what set of matrices we snapshot?

mollykreis:

msmithNI (approved):

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
jattasNI: Good question. The number of new tests does feel excessive to me. My main concern is that there are so many snapshots that we'd no longer take a close look at their diffs. Another lesser concern is that we'll end up exceeding our snapshot budget for relatively low value test coverage.

Now that you've added each supported control to the [Required Visible Pattern test story](https://60e89457a987cf003efc0a5b-xwhxswsyco.chromatic.com/?path=/story/tests-required-visible--required-visible-pattern), what value do we get from including it in matrix testing? I guess some of the states could be interesting: 

1. theme (seems worth testing since we have different * icon colors for different themes)
2. disabled (debatable since we recommend against setting required and disabled together)
3. error-visible (somewhat interesting on controls like radio group that put the ! icon near the label, not very interesting otherwise)

I propose we remove `requiredVisible` from most of the matrix stories. We could keep it for radio group if you think the error-visible case is worth testing. And we could change the Required Visible Pattern story to also test each theme. Thoughts?

rajsite: > would it be better to be more selective in what set of matrices we snapshot?

> I propose we remove requiredVisible from most of the matrix stories.

I don't want to encourage partial matrices in how the matrices are currently configured in this file. It would get difficult to see what portion of the matrix is fully covered. I don't think that's what is being proposed but just want to double check.

rajsite: Another option is to add a separate matrix test for the requuiredVisible if we don't think all the state permutations are valuable. A separate story of

```
export const textFieldThemeMatrix: StoryFn = createMatrixThemeStory(
    createMatrix(component, [
        requiredVisibleStates,
        disabledStates,
        errorStates
    ])
);
```

mollykreis: I opted for the pattern @rajsite proposed -- a separate matrix test for required visible. It includes the following states:
- requiredVisibleStates
- readOnlyStates
- disabledStates
- errorStates
- appearanceStates

Which means it leaves out the following states:
- actionButtonStates
- leftIconStates
- typeStates
- fullBleedStates
- valueStates

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -132,6 +134,31 @@ const component = (
     </${textFieldTag}>
 `;
 
+// prettier-ignore
+const reducedStateComponent = (
```
rajsite, line 138: Maybe requiredVisibleStatesComponent?
Just the stuff interesting for the requiredVisible feature

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -492,6 +519,16 @@ export const colorThemeReadOnlyDisabledWithButtons: StoryFn = createFixedThemeSt
     colorThemeDarkGreenBackground
 );
 
+export const textFieldRequiredVisibleThemeMatrix: StoryFn = createMatrixThemeStory(
```
rajsite, line 522: The most recent storybook tests run is showing a lot of new text field stories in addition to just the one new one. Maybe we need to update the guid in preview.js to get it to re-evaluate correctly or something (also a change for the variable rename was jusst pushed so maybe that new story will be right)

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -492,6 +519,16 @@ export const colorThemeReadOnlyDisabledWithButtons: StoryFn = createFixedThemeSt
...
+    createMatrix(requiredVisibleStatesComponent, [
+        requiredVisibleStates,
```
rajsite, line 524: nit: could you move uuup the appearanceStates to the second dimension? So the grouping is first split by requiredVisible and then by appearance mode? Think it's easier to visually scan when the same appearance modes are clustered together

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
mollykreis: The 24 stories are still showing up as being new. I'm still trying to figure out why and how to fix it.

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
jattasNI: I like the approach you went with.

mollykreis:
Closing in favor of #2507 because the chromatic diff got in a bad state.

PR #2512, opened 2025-01-10
Replace deprectated `importsNotUsedAsValues` option with `verbatimModuleSyntax`

m-akinc:
# Pull Request

## 🤨 Rationale

The [`importsNotUsedAsValues`](https://www.typescriptlang.org/tsconfig/#importsNotUsedAsValues) TypeScript option is deprectated and will stop working in TypeScript 5.5. It is replaced by [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax).

My motivation for this change is that while updating Nimble to Angular 18, I found that `importsNotUsedAsValues` was forcing me to do a type-only import of `Router` and `ActivatedRoute` [in `router_link.ts`](https://github.com/ni/nimble/blob/a7927039f1da96ac4672c5550d6b8f55071f125d/packages/angular-workspace/nimble-angular/src/thirdparty/directives/router_link.ts#L39). But doing so caused a different error for `NimbleAnchorRouterLinkWithHrefDirective`:

```
error NG2016: The directive NimbleAnchorRouterLinkWithHrefDirective inherits its constructor from RouterLink, but the latter has a constructor parameter that is not compatible with dependency injection. Either add an explicit constructor to NimbleAnchorRouterLinkWithHrefDirective or change RouterLink's constructor to use parameters that are valid for DI.
```

Both `Router` and `ActivatedRoute` are parameters to the `RouterLink` constructor. The problem is that if we do a type-only import of `Router` and `ActivatedRoute`, the compiler doesn't know that they are decorated to be injectable. Using `verbatimModuleSyntax` solves this by allowing us to import from the definition file rather than the type declaration file for those types, even though we're not directly using the definition.

This PR changes the option for the Angular projects, as well as `nimble-components`, `spright-components`, and `storybook`.

## 👩‍💻 Implementation

Had to update all imports of type symbols to explicitly say `type`.

## 🧪 Testing

Builds and existing tests pass.

## ✅ Checklist


m-akinc:
@mollykreis will you buddy this change?

packages/angular-workspace/tsconfig.json
```diff
@@ -36,7 +36,7 @@
       "es2020",
       "dom"
     ],
-    "importsNotUsedAsValues": "error",
+    "verbatimModuleSyntax": true,
     "ignoreDeprecations": "5.0",
```
rajsite, line 40: If I remember correctly the `ignoreDeprecations` is only here because we were using `importsNotUsedAsValues` so can now be removed.

rajsite, line 40: These two settings should be consistent across our tsconfigs. We could stage it in separate PRs essentially back-to-back or migrate everything together in a single PR.

packages/angular-workspace/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
```diff
@@ -22,22 +22,22 @@ import {
...
+export type { HourCycleFormat };
+export type { HourFormat };
+export type { LocaleMatcherAlgorithm };
+export type { MinuteFormat };
+export type { MonthFormat };
+export type { SecondFormat };
+export type { TimeStyle };
+export type { TimeZoneNameFormat };
+export type { WeekdayFormat };
+export type { YearFormat };
```
rajsite, line 40: This is reverting an explicit change from before, seems fishy: https://github.com/ni/nimble/pull/1438

rajsite: We should make sure to build tgzs against SystemLinkShared and an app to reduce the chance we are regressing against something we expect to be exporting the value of and not just type.

mollykreis: I agree that some more testing would be good. It looks like there are a number of places where something that had been exported now only has its type exported. Each of those instances is a possible place where a client could be broken depending on how they are using the exported entity.

m-akinc: Ran a SystemLinkShared pipeline build against these changes, and everything looks good: https://ni.visualstudio.com/DevCentral/_build/results?buildId=12189861

packages/angular-workspace/tsconfig.json
```diff
@@ -36,8 +36,7 @@
       "es2020",
       "dom"
     ],
-    "importsNotUsedAsValues": "error",
-    "ignoreDeprecations": "5.0",
+    "verbatimModuleSyntax": true,
```
m-akinc, line 39: Updating the angular-workspace tsconfig options.

packages/nimble-components/tsconfig.json
```diff
@@ -22,10 +22,7 @@
         ],
         // Whitelist imported @types instead of using them all
         "types": ["jasmine", "webpack-env", "offscreencanvas"],
-        // Require type only imports where possible
-        // so imports with side-effects are explicit
-        "importsNotUsedAsValues": "error",
-        "ignoreDeprecations": "5.0",
+        "verbatimModuleSyntax": true,
```
m-akinc, line 25: Update `nimble-components` tsconfig to use `verbatimModuleSyntax`.

packages/spright-components/tsconfig.json
```diff
@@ -22,10 +22,7 @@
         ],
         // Whitelist imported @types instead of using them all
         "types": ["jasmine", "webpack-env"],
-        // Require type only imports where possible
-        // so imports with side-effects are explicit
-        "importsNotUsedAsValues": "error",
-        "ignoreDeprecations": "5.0",
+        "verbatimModuleSyntax": true,
```
m-akinc, line 25: Update `spright-components` tsconfig to use `verbatimModuleSyntax`.

packages/storybook/tsconfig.json
```diff
@@ -24,8 +24,7 @@
         "types": ["jasmine", "webpack-env", "offscreencanvas"],
         // Require type only imports where possible
         // so imports with side-effects are explicit
-        "importsNotUsedAsValues": "error",
-        "ignoreDeprecations": "5.0",
+        "verbatimModuleSyntax": true,
```
m-akinc, line 27: Update `storybook` tsconfig to use `verbatimModuleSyntax`.

packages/angular-workspace/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
m-akinc: There is really only one interesting change other than the tsconfig files. I've added a comment to each of those.

packages/angular-workspace/tsconfig.json
m-akinc: Updated `nimble-components`, `spright-compnents`, and `storybook` tsconfigs (and source files), too.

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
rajsite: I don't think that's the right thing to do. We don't actually own those types. If additional members are added in intl.d.ts (or where ever those types are defined by TypeScript) we will have type conflicts. They happen to align now but they won't always and the previous typing better represented that by just passing them through.

Looking at https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table-column/date-text/types.ts
the only value is DataTextFormat defined by us and the rest are just types from interfaces defined by TypeScript. Looking at https://github.com/ni/nimble/pull/1438 again I think the value export it actually needed to do was DateTextFormat and the rest should have continued to be exported as types.

packages/angular-workspace/nimble-angular/table-column/date-text/nimble-table-column-date-text.directive.ts
```diff
@@ -22,22 +22,22 @@ import {
 } from '@ni/nimble-components/dist/esm/table-column/date-text/types';
 
 export type { TableColumnDateText };
-export { DateStyle };
+export type { DateStyle };
 export { DateTextFormat };
```
m-akinc, line 26: Still exporting `DateTextFormat` fully, but all other option "enums" are exported as types.

packages/nimble-components/src/table-column/date-text/types.ts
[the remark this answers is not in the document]
m-akinc: Reverted that change, except for the part where we re-export `DateTextFormat` fully from `nimble-table-column-date-text.directive.ts` (added a comment to that line).

rajsite (commented):
PR is in draft and build is failing but I was reset. I'm assuming still in dev and I'm not reviewing yet but FYI in-case a review was expected.

change/@ni-nimble-angular-535ed6e4-6372-4c96-844d-968e3c68ef76.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: The docs for `verbatimModuleSyntax` aren't clear to me. Does it only change the compiler output so that `type` imports are elided and value imports aren't? Or does it also change the compiler error behavior to throw an error if you only use an import as a type but import it as a value?

Asked another way, what error triggered you to make all the code changes and did it have an automated fixer?

Part of my confusion is also whether the [`consistent-type-imports` TypeScript eslint rule](https://typescript-eslint.io/rules/consistent-type-imports/) fits into the picture. The [NI styleguide disables it](https://github.com/ni/javascript-styleguide/blob/19f14277e85efbd7613428fcec8cf0fda3ed1b21/packages/eslint-config-typescript/index.js#L70).

rajsite (approved):

change/@ni-nimble-angular-535ed6e4-6372-4c96-844d-968e3c68ef76.json
rajsite: > Part of my confusion is also whether the [consistent-type-imports TypeScript eslint rule](https://typescript-eslint.io/rules/consistent-type-imports/) fits into the picture.

Using [both is not recommend](https://typescript-eslint.io/rules/consistent-type-imports/#comparison-with-importsnotusedasvalues--verbatimmodulesyntax): 

> Because there are some differences, using both this rule and verbatimModuleSyntax at the same time can lead to conflicting errors. As such we recommend that you only ever use one or the other -- never both.

m-akinc: Thanks, @rajsite. The part I was struggling to explain is why our previous setting of `"importsNotUsedAsValues": "error"` wasn't already complaining about all the imports that I had to modify in this PR. I guess "`importsNotUsedAsValues` was a little loose" means that for certain usages that are actually type-only (such as [this use of `AnchorOptions`](https://github.com/ni/nimble/blob/7e46f34cc5aef41ea09a103e719587b57d6bde83/packages/nimble-components/src/anchor-tree-item/index.ts#L121C53-L121C66)), it wouldn't report an error. `verbatimModuleSyntax` is apparently smarter/stricter and requires us to mark those type-only import use cases as such. It's a little confusing, because `verbatimModuleSyntax` feels like it should be giving you more freedom to decide whether to import something as a type or not, but really it still requires that types are imported as types, and it enforces that in more cases than `importsNotUsedAsValues` did.

> did it have an automated fixer?

It does have a fix available in VS Code when you hover over a violation, but I don't think there's any way to fix violations en masse.

jattasNI: Thanks, it all makes sense now!

jattasNI (approved):
I don't think you need to wait for any of the other code owners to submit this.

packages/angular-workspace/nimble-angular/src/directives/list-option-group/nimble-list-option-group.directive.ts
```diff
@@ -3,7 +3,7 @@ import { type BooleanValueOrAttribute, toBooleanProperty } from '@ni/nimble-angu
 import type { ListOptionGroup, listOptionGroupTag } from '@ni/nimble-components/dist/esm/list-option-group';
 
 export type { ListOptionGroup };
-export { listOptionGroupTag };
```
mollykreis, line 6: Any idea why this tag is exported with `type` when it doesn't look like many of the other component tags are?

m-akinc:
There were three tab scrolling tests failing in Webkit, but I suspect they must be intermittent. I've restarted the build, and I'll disable those tests in a separate PR, assuming my change isn't somehow responsible.

packages/angular-workspace/nimble-angular/src/directives/list-option-group/nimble-list-option-group.directive.ts
atmgrifter00: Well, the difference between this (and `NimbleListOptionDirective`, which also has this change), is that they are importing the members differently. These two both import the tag with the `type` keyword, where other directives do not. This still seems like a mistake in that the tag being exported with as a type seems wrong.

m-akinc: Great catch. I'll fix them up.

mollykreis (approved):

PR #2513, opened 2025-01-15
Spec for Spright chat components

jattasNI:
# Pull Request

## 🤨 Rationale

A client team is interested in contributing components to render a chat UI. They are less familiar with Nimble's architecture so I wrote the initial spec to help get them started.

## 👩‍💻 Implementation

Add spec detailing new components in Spright: chat message and chat conversation.

## 🧪 Testing

N/A, just a spec

## ✅ Checklist


mollykreis (approved):

packages/spright-components/src/chat/specs/README.md
```diff
@@ -0,0 +1,252 @@
...
+<template>
+    <slot></slot>
+</template>
+```
+
+### Native form integration
+
+Native form integration is not needed for these components.
+
+### Angular integration
```
jattasNI, line 167: My hunch is that the framework integration will be trivial since the API of these components is trivial but I didn't want to make that claim without digging into it. If any reviewers want to weigh in to confirm/deny that hunch, I'm happy to update these sections.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+The components will only provide the presentation layer, not logic for interacting with each other or any service to add messages to a conversation.
+
+The message component will allow slotting arbitrary content, but any efforts to add content types to Nimble are out of scope of this document. For example, adding capabilities to the rich text viewer or adding styling for specific content types.
+
+### Features
+
+#### Chat message
+
+1. Display arbitrary slotted content. For example: text, rich text, buttons, or a spinner.
```
joseahdz, line 38: We had a meeting with LV about their AI assistant and they require showing an image of the VI connector.  The same image LV shows in a VI's context help.  So, we require supporting images.

Another feature they want is the ability to add buttons at the top-right of a message respond.  The scenario UX showed was generating multiple VI descriptions and having '<' and '>' buttons to see them one at time.  There will also be a button or a dropdown to allow the user to ask the AI to try again with some options from the dropdown.  Alice is working on the UX design.  I will ask her to send me a link so you can see it.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -0,0 +1,252 @@
...
+Standard Storybook documentation. Since these are a Spright components we should ensure the documentation conveys the component status and gaps.
+
+There are parallel efforts to standardize and document other aspects of chat applications (for example the tone and language used by automated conversation participants) that are out of scope of this document.
+
+---
+
+## Open Issues
+
+1. Should we introduce an input toolbar component where a user can type messages and interact with related buttons? Or should applications construct this using the existing Nimble toolbar?
+    - Pros of dedicated component:
```
joseahdz, line 248: I prefer to have a dedicated component based on the pros.  I do agree it is better to have the toolbar in a single place so it can be updated in a single place.

joseahdz (approved):

packages/spright-components/src/chat/specs/README.md
rajsite: I don't think there are any complexities to be concerned about like router integration, form integration, or non-primitive property binding. Following the standard stateless directive pattern should be expected.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -0,0 +1,252 @@
...
+</spright-chat-message
+```
+
+### API
+
+#### Message
+
+- _Component Name_ `spright-chat-message`
+- _Props/Attrs_
+    - `status = "incoming" | "outgoing" | "system"`
```
rajsite, line 113: Is there a reason we went for `status` over `appearance` was the concern we might want block / ghost / outline versions or something? Or I guess we could have appearance `bubble` (sms-style) vs `threaded` (slack/discord-style), etc. in the future. Edit: actually I don't think that appearance would be per message but instead per conversation.

The name `status` doesn't feel quite right, feel like status is a transitory thing like "in progress", "delayed", etc. Like the status of a message could be `loading` or `is-typing` or `failed-to-send` or something.

Did we already think about `type` or `mode`? Other options didn't come to mind yet.

rajsite, line 113: Any preference of incoming/outgoing vs inbound/outbound? Seems like the latter is used more for tech / messaging (email inbox, sms, etc). Looks like the terms used in lightning as well (didn't search more broadly): https://www.lightningdesignsystem.com/components/chat/

packages/spright-components/src/chat/specs/README.md
jattasNI: I'll add images to the list of example content. With this design the message content is entirely up to the client application so supporting images wouldn't take any additional work. However at some point (outside the scope of this spec) we'll want to discuss how we plan to do that: if we want to use Nimble's rich text viewer we would need to enhance its capabilities.

I'll be interested to see those new designs. If they're within the body of the message then we can support it with the proposed approach. If they're separate content that are positioned by the message then we may want to add some new slots or attributes.

packages/spright-components/src/chat/specs/README.md
jattasNI: I think we may have different answers in the short term and medium/long term. My current leaning for the first pass is to just focus on the message and conversation and not create a toolbar in Spright. That gives the dev team time to get up to speed and the UX team time to solidify the toolbar UX more.

Then once we know more in a few weeks/months if we find we still want a toolbar it's easy to add it.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+---
+
+## Open Issues
+
+1. Should we introduce an input toolbar component where a user can type messages and interact with related buttons? Or should applications construct this using the existing Nimble toolbar?
+    - Pros of dedicated component:
+        1. consistent layout and reduced implementation effort across applications
+        1. single implementation to change if requirements change
+    - Pros of applications leveraging Nimble toolbar:
```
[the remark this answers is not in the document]
jattasNI: Good points, I'll add another bullet to capture that.

packages/spright-components/src/chat/specs/README.md
joseahdz: I like `type`.  It is easier to explain that there are three types of messages: incoming, outgoing and system.

joseahdz: After reading the section `Attribute naming convention` in the contributing document, we cannot use the name `type` for an attribute. So, we can use the name `mode`.

jattasNI: Now that I think about it I agree `status` doesn't feel right. I don't love `appearance` either since for existing Nimble components it has mostly referred to visual style but this feels like it has more semantic meaning, so the options will be quite different from things like "ghost" or "prominent".

Brain dump of other ideas:
- `mode`
- `message-direction` (don't want to conflict with native `direction` attribute
- `message-type` (don't want to conflict with native `type` attribute)

Or if we wanted to switch to describing who rather than where/how it came from:
- `sender = this-user | other-user | system`
- `sender = self | other | system`
- `sender = me | you | system`
- `entity = (same options as above)`

I kind of like `message-direction` but could also live with `mode` or `message-type`. Preference @rajsite @joseahdz ?

jattasNI: I'm good with inbound/outbound over incoming/outgoing. Now that I think of it the "going" options sound like it's still in the process of being sent. Some of the options in the other thread might change this so let's keep brainstorming there.

atmgrifter00 (approved):

packages/spright-components/src/chat/specs/README.md
rajsite: I like `message-direction` with the enum states `inbound/outbound` but `system` doesn't feel like it fits as a direction.

Seems like `message-type` is a bit more flexible and works for enum states `inboud/outbound/system`.

rajsite (approved):

packages/spright-components/src/chat/specs/README.md
jattasNI: I moved this to Out of Scope for now.

PR #2521, opened 2025-01-24
Allow an item in the toolbar to lose focus by clicking the toolbar itself

### LINKED ISSUE #2494: Toolbar Active control blur/focus-out events not happening when clicking within in Edge and Chrome
# 🐛 Bug Report

In Chrome and Edge, when a control (e.g., a button or text field) within a toolbar is focused or active, clicking anywhere else within the toolbar does not blur the currently active control or set its value. However, clicking outside the toolbar works as expected and blurs the active control. 

This behavior functions correctly in Firefox, where clicking within the toolbar properly blurs the active control.

focusing-toolbar-issue

## 💻 Repro or Code Sample

- Click on any element inside the toolbar to make it focused or active.  
- Then, click on any other area within the toolbar to blur or remove focus from the currently active control.

## 🤔 Expected Behavior

- clicking on any area within the toolbar should blur or remove focus from the currently active control, just as it does when clicking outside the toolbar.

## 😯 Current Behavior

- Clicking within the toolbar does not blur or focus out the currently active control

## 🔦 Context
Discussion link: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullRequest/850146#1733521296

## 🌍 Your Environment

* OS: Windows
* Browser: Microsoft Edge, Google Chrome,  Mozilla FireFox

mollykreis:
# Pull Request

## 🤨 Rationale

Resolves #2494

## 👩‍💻 Implementation

Removed `delgatesFocus: true` on the toolbar so that clicking on an empty space within the toolbar will cause the currently focused item to lose focus.

Note: The use case described in the referenced issue may still be a bit problematic given all the editable fields (date pickers, numerics, etc) because the user may expect to use the left/right arrow keys to navigate within that component, but the toolbar will handle the key event first and move focus to a different item in the keyboard. That behavior is not affected in any way by this change.

## 🧪 Testing

Manually tested with a numeric field and text field in the toolbar.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-66a1a299-08a4-495c-9750-85902fe5413f.json
+++ b/change/@ni-nimble-components-66a1a299-08a4-495c-9750-85902fe5413f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Allow an item in the toolbar to lose focus by clicking the toolbar itself",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/toolbar/index.ts
+++ b/packages/nimble-components/src/toolbar/index.ts
@@ -21,10 +21,7 @@ const nimbleToolbar = Toolbar.compose<ToolbarOptions>({
     baseName: 'toolbar',
     baseClass: FoundationToolbar,
     template,
-    styles,
-    shadowOptions: {
-        delegatesFocus: true
-    }
+    styles
 });
 
 DesignSystem.getOrCreate().withPrefix('nimble').register(nimbleToolbar());


rajsite (approved):

mollykreis:
@jattasNI, do you want to look at this change, or should I go ahead and merge it?

rajsite:
Think we can bypass @jattasNI for this minor change (seemed okay in storybook for mouse and keyboard interactions)

rajsite:
Disabled automerge because pending change on main

change/@ni-nimble-components-66a1a299-08a4-495c-9750-85902fe5413f.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: @mollykreis @rajsite Did we reach a conclusion on whether the original use case which spurred this bug should actually be using nimble-toolbar (rather than the custom implementation they switched to)? If we think nimble-toolbar is appropriate, could you ping the original bug filer to let them know it's fixed (if you haven't already)?

mollykreis: I don't think we reached an official conclusion, but my leaning is that they are probably better off with their custom implementation because the roving index of the nimble-toolbar won't work well with all the inputs they put in their toolbar. A user will probably have much better luck using the keyboard with their toolbar when each component is individually tabbable.

rajsite: Agreed, yea we probably need some usage guidance on nimble-toolbar for that and maybe higher-level patterns that discourage placing components with high interaction inside toolbars (framed around the aria guidance).

PR #2524, opened 2025-01-29
Disable Firefox in concurrent test runs for CI

rajsite:
# Pull Request

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-5ea84983-4985-4423-9b3c-0cec94b65700.json
+++ b/change/@ni-nimble-components-5ea84983-4985-4423-9b3c-0cec94b65700.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Remove firefox from concurrent testing",
+  "packageName": "@ni/nimble-components",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
--- a/packages/nimble-components/package.json
+++ b/packages/nimble-components/package.json
@@ -41,7 +41,7 @@
     "test-webkit:debugger": "karma start karma.conf.js --browsers=WebkitDebugging --skip-tags SkipWebkit",
     "test-webkit:watch": "karma start karma.conf.headless.js --browsers=WebkitHeadless --skip-tags SkipWebkit --watch-extensions js",
     "test-webkit": "karma start karma.conf.headless.js --browsers=WebkitHeadless --single-run --skip-tags SkipWebkit",
-    "test-concurrent": "concurrently --pad-prefix --timings --group \"npm:test-chrome\" \"npm:test-firefox\" \"npm:test-webkit\""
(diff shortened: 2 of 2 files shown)


change/@ni-nimble-components-5ea84983-4985-4423-9b3c-0cec94b65700.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Ironic that the PR to fix intermittent FF tests has a failing build due to an intermittent webkit test 🤦

packages/nimble-components/package.json
```diff
@@ -41,7 +41,7 @@
     "test-webkit:debugger": "karma start karma.conf.js --browsers=WebkitDebugging --skip-tags SkipWebkit",
```
jattasNI, line 41: With this change our Firefox coverage is just Chromatic tests, correct? That leaves us vulnerable to not catching FF differences in JS or browser API behaviors: an unlikely source of bugs but still possible. Want to brainstorm some mitigations?

1. Submit PR as-is, rely on ad-hoc usage from client devs who happen to use Firefox
2. Add a #1747 item to run the Firefox tests
3. Create a new pipeline (or pipeline step) for running Firefox tests which doesn't block builds. Monitor it occasionally for full failures
4. Devote effort to test execution strategies to work around the issue (e.g. re-run failed tests or parallelize tests across agents if we think memory pressure is the root cause)
5. Devote effort to fixing the tests.

I suspect 3 is likely to be ignored and 4-5 are too much work. But doing 2 seems feasible. Thoughts?

jattasNI (approved):

packages/nimble-components/package.json
rajsite: 2 sounds reasonable since we already have Safari as a case in the manual tasks. I'll add Firefox to that.

rajsite: > parallelize tests across agents if we think memory pressure is the root cause

I would bet that parallelizing across browser instances would be enough to improve. Seems like Firefox's perf issues were around GCs getting much longer over time so just splitting up the test suite could maybe help.

Looking quickly karma-parallel exists but hasn't been updated since 2018: https://github.com/joeljeske/karma-parallel (their package json shows karma 3 vs we are on karma 6). We also have a pretty complex config with the webpack build as part of it.

rajsite: Temporarily resolving to merge

jattasNI: I didn't find anything other than karma-parallel either. I guess we could also do clunky manual segmentation of our suite, like separate commands for table tests, rich text tests, and everything else. Then we could use `concurrently` to parallelize. But I'm ok to leave that as an innovation idea.

Issue #1271, opened 2023-05-25
Consider using scrollbar-gutter where useful when available


rajsite:
## 🧹 Tech Debt

The [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) CSS feature allows you to reserve space in a components should a scrollbar render. It's aware of system-specific details about scrollbars (size, whether overlay scrollbars are used or classix scrollbars).

Unfortunately currently blocked on Safari support.

When available we should be able to use it to remove the complexity of scrollbar tracking in the nimble-text-area:

https://github.com/ni/nimble/blob/14402b0d2e2a8a56454f79aa971324bf96dea5fe/packages/nimble-components/src/text-area/index.ts#L123-L126

msmithNI:
The table virtualizer is another place that has similar code that we might be able to revisit: https://github.com/ni/nimble/blob/14402b0d2e2a8a56454f79aa971324bf96dea5fe/packages/nimble-components/src/table/models/virtualizer.ts#L51

PR #2525, opened 2025-01-29
Table height does not account for horizontal scrollbar

### LINKED ISSUE #2520: Table 'ni-nimble-table-fit-rows-height' token doesn't account for horizontal scrollbar
# 🐛 Bug Report

When styling the nimble-table with using the `ni-nimble-table-fit-rows-height` token, the height of the table is correct unless the table also has a horizontal scrollbar. In that case, the token's value does not account for the scrollbar, and you end up with a vertical scrollbar as well (even if there are fewer than 40 rows). In the worst case, the table only has 1 row, and the styled height in addition to the horizontal scrollbar makes the single row of data virtually unreadable.

## 🤔 Expected Behavior

Whether or not a horizontal scrollbar is visible (along with its height) should be taken into account when setting the value of the `ni-nimble-table-fit-rows-height` token.

## 😯 Current Behavior

A table styled with a height of `ni-nimble-table-fit-rows-height` and fewer than 40 rows can still get a vertical scrollbar if the table also has a horizontal scrollbar. In cases where there are very few rows (e.g. 1 row), the table is very difficult to use because it is so short, and the data can't be fully read.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2520 

## 👩‍💻 Implementation

The issue is that when the horizontal scrollbar was present while styling the table with a height of `tableFitRowsHeight`, a vertical scrollbar would be present because the height token's value didn't account for the height of the horizontal scrollbar. Now, the virtualizer's ResizeObserver hander calculates the height of the horizontal scrollbar (which is 0 if there isn't a horizontal scrollbar), and exposes that value as an `@observable`, similar to `scrollHeight`. As a result, the table can use both values in its calculation of `tableFitRowsHeight`, resulting in no vertical scrollbar when using the token (unless the max-height of the table is reached).

## 🧪 Testing

Manually tested, including the following scenarios to ensure the auto height of the table adjusted correctly as a horizontal scrollbar was added/removed:
- toggle the value of `column-visible` on a column
- add/remove a column from the DOM
- increase/decrease the minimum width of a column
- resize browser window

New auto tests, including chromatic test that loads a table with auto height and a horizontal scrollbar

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/table/models/virtualizer.ts
+++ b/packages/nimble-components/src/table/models/virtualizer.ts
@@ -25,6 +25,9 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
     public scrollHeight = 0;
 
     @observable
+    public horizontalScrollbarHeight = 0;
+
+    @observable
     public isScrolling = false;
 
     @observable
@@ -58,6 +61,7 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
                 // by the same margin so the column headers align with the corresponding rendered cells
                 const viewportBoundingWidth = borderBoxSize.inlineSize;
                 this.headerContainerMarginRight = viewportBoundingWidth - this.table.viewport.clientWidth;
+                this.horizontalScrollbarHeight = borderBoxSize.blockSize - this.table.viewport.clientHeight;
             }
         });
     }
--- a/packages/nimble-components/src/table/tests/table.spec.ts
+++ b/packages/nimble-components/src/table/tests/table.spec.ts
@@ -2400,6 +2400,21 @@ describe('Table', () => {
                 expect(tokenValue).toBe(expectedHeight);
             });
 
+            it('adjusts height when horizontal scrollbar is shown', async () => {
+                await element.setData(simpleTableData);
+                await waitForUpdatesAsync();
+
+                await pageObject.sizeTableToGivenRowWidth(100, element);
+                await waitForUpdatesAsync();
+
+                const tokenValue = getTableHeight();
+                const heightWithoutScrollbar = getExpectedHeight(
+                    simpleTableData.length
+                );
+                expect(parseFloat(tokenValue)).toBeGreaterThanOrEqual(parseFloat(heightWithoutScrollbar));
+                expect(element.viewport.scrollHeight).toBe(element.viewport.clientHeight);
+            });
+
             it('has correct height when height changes because of setting data', async () => {
                 await element.setData(simpleTableData);
                 await waitForUpdatesAsync();
--- a/packages/nimble-components/src/table/tests/table.spec.ts
+++ b/packages/nimble-components/src/table/tests/table.spec.ts
@@ -2411,6 +2411,9 @@ describe('Table', () => {
                 const heightWithoutScrollbar = getExpectedHeight(
                     simpleTableData.length
                 );
+                // Use the `toBeGreaterThanOrEqual` comparison because in browsers with overlay scrollbars,
+                // the heights will match. In other browsers, the token height will be larger than the
+                // calculated height of the rows + header by the width of the scrollbar.
                 expect(parseFloat(tokenValue)).toBeGreaterThanOrEqual(
                     parseFloat(heightWithoutScrollbar)
                 );
--- a/packages/nimble-components/src/table/tests/table.spec.ts
+++ b/packages/nimble-components/src/table/tests/table.spec.ts
@@ -2411,8 +2411,12 @@ describe('Table', () => {
                 const heightWithoutScrollbar = getExpectedHeight(
                     simpleTableData.length
                 );
-                expect(parseFloat(tokenValue)).toBeGreaterThanOrEqual(parseFloat(heightWithoutScrollbar));
-                expect(element.viewport.scrollHeight).toBe(element.viewport.clientHeight);
+                expect(parseFloat(tokenValue)).toBeGreaterThanOrEqual(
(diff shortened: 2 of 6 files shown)


packages/nimble-components/src/table/models/virtualizer.ts
```diff
@@ -58,6 +61,7 @@ export class Virtualizer<TData extends TableRecord = TableRecord> {
                 // by the same margin so the column headers align with the corresponding rendered cells
                 const viewportBoundingWidth = borderBoxSize.inlineSize;
                 this.headerContainerMarginRight = viewportBoundingWidth - this.table.viewport.clientWidth;
+                this.horizontalScrollbarHeight = borderBoxSize.blockSize - this.table.viewport.clientHeight;
```
rajsite, line 64: Will this sufficiently catch other conditions besides the viewport changing size that could result in the scrollbar appearing? For example in the text area it ended up being a whack-a-mole of properties that could trigger a scrollbar to appear: https://github.com/ni/nimble/blob/52caaabfa9c595f1909d1bb374aae3d228c4d172/packages/nimble-components/src/text-area/index.ts#L69-L74

Related, maybe we could use [`scrollbar-gutter`](https://caniuse.com/mdn-css_properties_scrollbar-gutter) to always reserve the space and simplify? https://github.com/ni/nimble/issues/1271

rajsite: > Using scrollbar-gutter was one of the first paths I went down in my attempt to fix this issue, but from what I can tell, it doesn't apply to horizontal scrollbars.

Well dang, seems to be the case 🤦‍♂️: https://github.com/w3c/csswg-drafts/issues/5252

> I manually tested the following scenarios, and they all caused the value of the token to change appropriately:
> 
> - toggle the value of column-visible on a column
> - add/remove a column from the DOM
> - increase/decrease the minimum width of a column
> - resize browser window

Sold, thanks for looking into it.

packages/storybook/src/nimble/table/table-fit-rows-height-matrix.stories.ts
```diff
@@ -49,9 +56,9 @@ const component = (
         }
     </style>
     <${tableTag}>
```
rajsite, line 58: Think we need to set a width on the table for this to be reliable (my monitor was high enough resolution to happen to not show either case with a horizontal scrollbar). Maybe a fixed value like 600px or something?

packages/storybook/src/nimble/table/table-fit-rows-height-matrix.stories.ts
```diff
@@ -29,6 +29,12 @@ const groupingStates = [
 ] as const;
 type GroupingState = (typeof groupingStates)[number];
 
+const minColumnWidthStates = [
+    ['Small Minimum', 100],
+    ['Large Minimum', 500]
```
rajsite, line 34: Any spot to capture the scrollbar intention, maybe in the name?

```suggestion
    ['Large Minimum (show horizontal scrollbar)', 500]
```

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -2400,6 +2400,25 @@ describe('Table', () => {
                 expect(tokenValue).toBe(expectedHeight);
             });
 
+            it('adjusts height when horizontal scrollbar is shown', async () => {
```
rajsite, line 2403: It's hard for me to conceptualize, but do we expect this test to pass for operating systems that use overlay scrollbars which don't take-uup space (mac os, android / ios mobile)?

rajsite (approved):

packages/nimble-components/src/table/tests/table.spec.ts
mollykreis: Yes, I expect this to pass with overlay scrollbars. In fact, I believe that's the state our webkit tests run in.

The part of the test that makes it pass with overlay scrollbars is using `toBeGreaterThanOrEqual` rather than `toBeGreaterThan` when comparing the token value with the calculated expected height (line 2414 of the test). In chrome & FF, the token is greater than the calculated height because of the added scrollbar, but in webkit, the token has the same value as the calculated height because it is an overlay scrollbar.

rajsite: 🛳️-it!

atmgrifter00 (approved):

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -2400,6 +2400,25 @@ describe('Table', () => {
...
+                await waitForUpdatesAsync();
+
+                await pageObject.sizeTableToGivenRowWidth(100, element);
+                await waitForUpdatesAsync();
+
+                const tokenValue = getTableHeight();
+                const heightWithoutScrollbar = getExpectedHeight(
+                    simpleTableData.length
+                );
+                expect(parseFloat(tokenValue)).toBeGreaterThanOrEqual(
```
rajsite, line 2414: nit / optional: Maybe add a comment here discussing how this case handles both overlay scrollbars and sized scrollbars for future readers.

jattasNI (approved):
I verified in macOS Safari and Firefox (both of which show overlay scrollbars) that the table fit rows height stories behave correctly. It doesn't reserve extra space for the horizontal scrollbar and it doesn't cause an unnecessary vertical scrollbar to appear when the horizontal one is shown. 🚀

Issue #2531, opened 2025-02-04
Custom text & anchor support for Mapping Table Column


stefanc18:
# 🙋 Feature Request

## 😯 Problem to Solve

The Systems grid on SLE currently has 2 columns for rendering the job status of a system: an icon column and an anchor text column:

![Image](https://github.com/user-attachments/assets/59836cf1-20b8-4119-8cf8-278018b6461f)

- The icon column can have multiple states based on the job status; for example, it will be a spinner when the job is in progress, a green checkmark when the job is succeeded or a red xmark if the job failed.
- The icon column will always have as tooltip the status of the job. So even if there is a very limited set of icons, the text is customized to reflect the exact status of the job as queried from the Systems Management Service.
- The anchor column is a link to the job details.

The current implementation for the icon column is to use the existing mapping table column. A mapping is done for each system row, so the system id is essentially used as a key. This allows us to have a different tooltip for each system depending on its status. However, this approach has heavy performance issues when the table has 1000 rows and the entire column is populated:

- With the icon column added to the table:

![Image](https://github.com/user-attachments/assets/1dc764e8-dd9f-4e0c-8747-2bbf22f91be8)

- After removing it:

![Image](https://github.com/user-attachments/assets/1db793c3-5290-4037-a385-59f6ff262ec4)

Using the performance tools of Google Chrome we notice that the heavy operation is coming from the table, after the column map is changed:

![Image](https://github.com/user-attachments/assets/5be9c11c-f3f7-43fd-9cbc-fd7f4168b890)

## 💁 Proposed Solution

- Have a `textFieldName` (or a similar property) that would allow providing a custom text for the column. It would look something like this:

![Image](https://github.com/user-attachments/assets/2bb0c9e6-abef-4444-a353-3cee1005420a)

With this approach, the mapping icons would be fixed (the fixed set of possible icons) and only the tooltip text would be given dynamically by a different property of the row.

Moreover, the desire on the systems table is to unite those two columns into one. This would also require another property on the mapping column to allow an URL to be specified, which the Systems table would use to populate with the job details URL for each job status. This would make the text that is displayed next to the column to be clickable and link to the job details page.

rajsite:
In terms of sorting and grouping is this a mapping column with hyperlink text decoration or a hyperlink text column with mapping decoration?

I'm guessing it's more of the former, i.e. the state represented by the icon is the interesting thing to sort and group by and the hyperlink text is metadata.

stefanc18:
[...]

PR #2533, opened 2025-02-05
Mapping anchor column HLD

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #2531

### DIFF
--- a/packages/nimble-components/src/table/specs/table-column-specs/table-column-anchor-hld.md
+++ b/packages/nimble-components/src/table/specs/table-column-specs/table-column-anchor-hld.md
@@ -68,10 +68,6 @@ When grouped, the header item should not be a link.
 
 The column will also support having a custom sort order, as detailed in the [Custom Sort Order HLD](../table-column-custom-sort-field-hld.md).
 
-### Clearing state
-
-Because the anchor in the cell can have focus, we must override `focusedRecycleCallback()` in our cell view and have it call `blur()` on our anchor. If we don't, the focus can pass to other cells as you scroll.
-
 ### Sizing
 
 The column should support the same sizing modes as the text column, which is fractional widths plus minimum pixel widths.
@@ -84,9 +80,9 @@ Arrowing to a anchor table cell should focus the link (if it is actually a link 
 
 In the accessibility tree, the cells of an anchor column are instances of [`nimble-table-cell`](https://github.com/ni/nimble/blob/f663c38741e731bef91aa58e8fb2d1cec653b679/packages/nimble-components/src/table/components/cell/template.ts#L6) which has a `role` of [`cell`](https://w3c.github.io/aria/#cell). The cell then has a child `nimble-anchor`, which has a `role` of [`link`](https://w3c.github.io/aria/#link).
 
-### Angular RouterLink Support
+### Angular Routing support
 
-The real challenge of this column type is integrating with the Angular router. The `RouterLink`/`RouterLinkWithHref` directives are used to intercept clicks on anchors and replace the default navigation action with a call to `Router.navigateByUrl()`. As we have done in the past for other anchor components, we will have our own directive deriving from `RouterLinkWithHref`. Our directive will apply based on the presence of the `nimbleRouterLink` attribute. As we have done in the past, we will also have a directive that throws an error if `routerLink` is used instead. Normally, you would specify the `nimbleRouterLink` and related attributes (e.g. `queryParams`, `replaceUrl`, etc.) directly on the anchor element, but that's not possible for anchors in generated table cells. Our options are to put the directive on `nimble-table` or `nimble-table-column-anchor`. Because we would like to allow different anchor columns to be configured differently, and because it is a more intuitive API, we want to put our directive on the column element:
+Using the `TableColumnAnchor` in Angular requires the use of the [`NavigationGuard`](https://github.com/ni/nimble/blob/main/packages/angular-workspace/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts) in order to allow proper routing within the Angular application.
 
 ```html
 <nimble-table>
--- a/packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
+++ b/packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
@@ -1,0 +1,201 @@
+# Mapping Anchor Table Column
+
+## Overview
+
+The `nimble-table-column-mapping-anchor` is a component that supports rendering specific number, boolean, or string values as an icon or a spinner, with an accompanying text anchor element. The mappings are defined by child elements of `nimble-mapping-icon`, `nimble-mapping-spinner`, or `nimble-mapping-empty`.
+
+### Background
+
+[Icon with anchor column](https://github.com/ni/nimble/issues/2531)
+
+### Features
+
+- Supported input:
+    - string
+    - number
+    - boolean
+- Supported output:
+    - Icon with text link
+    - Spinner with text link
+    - text link
+
+### Non-goals
+
+- Non-Nimble icon support
+- Arbitrary icon colors
+- Non-icon, non-spinner Nimble components
+- Specifying different text for an icon/spinner's label than the overall text of the mapping
+
+---
+
+## Design
+
+Below is an example of how these elements would be used within a `nimble-table`:
+
+```HTML
+<nimble-table>
+    <nimble-table-column-mapping-anchor field-name="status" key-type="string" label-field-name="status" href-field-name="status-url">
+        Status
+        <nimble-mapping-icon key="fail" icon="nimble-icon-xmark" severity="error" text="Failed"></nimble-mapping-icon>
+        <nimble-mapping-icon key="error" icon="nimble-icon-xmark" severity="error" text="Errored"></nimble-mapping-icon>
+        <nimble-mapping-icon key="pass" icon="nimble-icon-check" severity="success" text="Passed"></nimble-mapping-icon>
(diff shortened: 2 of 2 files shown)


packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+    - Icon with text link
+    - Spinner with text link
+    - text link
+
+### Non-goals
+
+- Non-Nimble icon support
+- Arbitrary icon colors
+- Non-icon, non-spinner Nimble components
+- Specifying different text for an icon/spinner's label than the overall text of the mapping
```
rajsite, line 27: Lol, I think that is the exact goal of this PR, to specify text outside the mapping :P

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+
+N/A
+
+### Documentation
+
+Documented in Storybook
+
+---
+
+## Open Issues
```
rajsite, line 197: Do we think we will have a visual design pass? My biggest question was how an icon + icon text + link should look together @fredvisser 

image

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+- `href-field-name`: string
+- `appearance`: string
+- `underline-hidden`: string
+- `hreflang`
+- `ping`
+- `referrerpolicy`
+- `rel`
+- `target`
+- `type`
+- `download`
```
rajsite, line 78: Might too much detail for HLD but was thinking:
- a mixin for the native anchor attributes, i.e. `hreflang`, `ping`, `target`, etc.
- a mixin for AnchorColumns which calls the above mixin and adds `label-filed-name`, `href-field-name`, `appearance`, `underline`

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
rajsite: The icon with it's text description is our recommended configuration: https://nimble.ni.dev/storybook/index.html?path=/docs/components-table-column-mapping--docs#choosing-a-mapping-style

It is conceptual mapping + metadata. The mapping is ideally icon + text but can also be just icon. They together describe the state. The metadata is a hyperlink, plain text, or placeholder text.

> It would be nice to support navigating by clicking the icon.

I didn't think I agree, the metadata is a distinct thing from the icon. I think some concrete examples would help me understand your proposals

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
stefanc18: I believe Milan's right, that's the main reason we created that issue. The current requirement is to have both the text next to the icon, as well as the text displayed in the tooltip when hovered over display the same text. If a different decision is taken for this feature, I would like to get the confirmation of the PO first that this is acceptable.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+### Background
+
+[Icon with anchor column](https://github.com/ni/nimble/issues/2531)
+
+### Features
+
+- Supported input:
+    - string
+    - number
+    - boolean
```
mollykreis, line 16: This statement is a bit confusing (at least to me). Since there are multiple inputs to the column that support different types, it might be more helpful to split out the "Supported input" section into what is supported for a given input. Something like:

- Supported input:
    - string, number, or boolean mapping
    - string label
    - string href

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+When cellRecord.href is missing
+<span
+    @mouseover="${(x, c) => setTitleWhenOverflow(...)}"
+    @mouseout="${(_x, c) => removeTitle(...)}"
+>
+    ${cellState.cellRecord.label}
+</span>
+```
+
+### Sorting
```
mollykreis, line 118: Can you add a similar section for grouping behavior?

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
...
+
+In the accessibility tree, the cells of an anchor column are instances of [`nimble-table-cell`](https://github.com/ni/nimble/blob/f663c38741e731bef91aa58e8fb2d1cec653b679/packages/nimble-components/src/table/components/cell/template.ts#L6) which has a `role` of [`cell`](https://w3c.github.io/aria/#cell). The cell then has a child `nimble-anchor`, which has a `role` of [`link`](https://w3c.github.io/aria/#link).
+
+### Angular integration
+
+Angular directives will be created for the column component. No component has form association, so a `ControlValueAccessor` will not be created.
+
+#### Angular Routing support
+
+Using the `TableColumnAnchor` in Angular requires the use of the [`NavigationGuard`](https://github.com/ni/nimble/blob/main/packages/angular-workspace/nimble-angular/table-column/anchor/nimble-table-column-anchor-navigation-guard.directive.ts) in order to allow proper routing within the Angular application.
```
mollykreis, line 140: Should this say "Using the `TableColumnMappingAnchor`"? Or is this saying the `TableColumnAnchor` will somehow be reused?

rajsite: Think the intention was something along the lines of `TableColumnMappingAnchor` will follow the similar `NavigationGuard` directive pattern of `TableColumnAnchor`

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
atmgrifter00: Oops...that was just a copy-pasta goof.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
atmgrifter00: Cool. I think I'll leave that out of this HLD. May want to discuss details of this idea out-of-band.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
rajsite: Example of current systems table:
image

rajsite, line 197: Depending on which way we go on visual design, if we end up not wanting to show icon text for the icon mappings I propose we enforce that with validation. Require in mapping validation that all the icon mappings have "text-hidden" true (as opposed to ignoring the attribute on the mapping or creating a new icon mapping that does not have the unused attribute).

jattasNI: This could be stale information, but I thought one of the goals was to have a compact icon-only representation while still seeing per-row status information in a tooltip.

From the Proposed Solution on #2531:

>Have a textFieldName (or a similar property) that would allow providing a custom text for the column.... With this approach, the mapping icons would be fixed (the fixed set of possible icons) and only the tooltip text would be given dynamically by a different property of the row.

I assume in this case the desire would be to have the icon be clickable.

We could push back on that if we think it's not a good direction, but I can see it being useful for the same reasons that `width-mode=iconSize` is useful: it takes up less space.

If this is something we support, it'd be another visual design question about how to present the mapping text and the per-row label in the same tooltip.

packages/nimble-components/src/table/specs/table-column-specs/table-column-mapping-anchor.md
```diff
@@ -0,0 +1,197 @@
+# Mapping Anchor Table Column
```
jattasNI, line 1: Once this is closer to ready for approval, we should add Stefan as a required reviewer.

m-akinc:
@atmgrifter00 @fredvisser @rajsite Where did we end up on this? I remember us changing our minds about the need to implement this column type. We should update this PR and the related work item.

packages/nimble-components/src/table/specs/table-column-specs/table-column-anchor-hld.md
```diff
@@ -84,9 +80,9 @@ Arrowing to a anchor table cell should focus the link (if it is actually a link
 
 In the accessibility tree, the cells of an anchor column are instances of [`nimble-table-cell`](https://github.com/ni/nimble/blob/f663c38741e731bef91aa58e8fb2d1cec653b679/packages/nimble-components/src/table/components/cell/template.ts#L6) which has a `role` of [`cell`](https://w3c.github.io/aria/#cell). The cell then has a child `nimble-anchor`, which has a `role` of [`link`](https://w3c.github.io/aria/#link).
 
-### Angular RouterLink Support
+### Angular Routing support
```
rajsite, line 83: @atmgrifter00 @fredvisser @rajsite Where did we end up on this? I remember us changing our minds about the need to implement this column type. We should update this PR and the related work item.

rajsite:
Closing as the PR is stalled and there is an existing workaround to use two separate columns. Can re-open if becomes active again.

Issue #2281, opened 2024-07-17
Lighthouse performance metric assert dropped from 0.9 to 0.8


rajsite:
# 🧹 Tech Debt

Recent builds have been failing the lighthouse performance assertions for the Angular example-client-app. In a [commit pushed to main](https://github.com/ni/nimble/commit/235fef868edcdc11367d4c9ed8ea938a889e4f2b) the performance threshold was lowered to 0.8.

We should consider how to resolve the performance issue, some ideas:
- The metric is useful and we should listen. Modify the examples such that they have fewer elements by default (buttons to generate rows, etc) or that the perf warnings about unused JS are addressed
- We decide the metric is not useful and decide what we want to get out of the lighthouse review for the example angular app. Maybe we turn-off performance tracking (which might be unfavorable for early detection of drastic changes)

PR #2540, opened 2025-02-13
Drop lighthouse performance check to 0.7

rajsite:
# Pull Request

## 🤨 Rationale

Looks like the current lighthouse perf for the angular example app has lowered to high 0.7 / low 0.8 and is failing intermittently.

## 👩‍💻 Implementation

- Bump the current lighthouse performance check from 0.8 to 0.7
- Updated https://github.com/ni/nimble/issues/2281

## 🧪 Testing

Rely on CI

## ✅ Checklist

### DIFF
--- a/packages/angular-workspace/lighthouserc.js
+++ b/packages/angular-workspace/lighthouserc.js
@@ -16,7 +16,7 @@ module.exports = {
         },
         assert: {
             assertions: {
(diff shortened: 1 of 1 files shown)


rajsite:
Merging as it essentially continues the trend from the linked issue. @jattasNI is any follow-up the issue would be a good spot

PR #2542, opened 2025-02-21
Changed Rich Text Viewer markdown property to an attribute

joseahdz:
# Pull Request

## 🤨 Rationale

The markdown property cannot be set from Blazor since [Blazor does not currently support setting properties](https://github.com/dotnet/aspnetcore/issues/5506#issuecomment-485385419). 

Some options are:
- Use JS to programmatically change the property value. Potentially following a similar pattern as [FluentCheckbox for the indeterminate state](https://github.com/microsoft/fluentui-blazor/blob/f0f1bf8c93ef23c6425b30bd4accf217a1256a20/src/Core/Components/Checkbox/FluentCheckbox.razor.cs#L131)
- Change the `@observable` property to an `@attr` property + attribute

For the rich text viewer `markdown` property there does not seem to be a specific reason to avoid configuring `markdown` as an attribute. 

## 👩‍💻 Implementation

Changed markdown property from `observable` to `attr`.  Also, updated story book to show `markdown` as an attribute instead of a property.

The `fromView` mode will be used to avoid unnecessarily reflecting large strings set as properties back as attributes.

## 🧪 Testing

Fixed a couple of tests due to changing from a property to an attribute.  All other tests are passing.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-33b8266d-c20f-4ec8-92a9-a6ecf81fe328.json
+++ b/change/@ni-nimble-components-33b8266d-c20f-4ec8-92a9-a6ecf81fe328.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Changed Rich Text Viewer markdown property to an attribute",
+  "packageName": "@ni/nimble-components",
+  "email": "107654185+joseahdz@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/rich-text/viewer/index.ts
+++ b/packages/nimble-components/src/rich-text/viewer/index.ts
@@ -22,7 +22,7 @@ export class RichTextViewer extends RichText {
      * @remarks
      * HTML Attribute: markdown
      */
-    @attr
+    @attr({ mode: 'fromView' })
     public markdown = '';
 
     /**
--- a/packages/nimble-components/src/rich-text/viewer/index.ts
+++ b/packages/nimble-components/src/rich-text/viewer/index.ts
@@ -1,5 +1,5 @@
 import { DesignSystem } from '@microsoft/fast-foundation';
-import { observable } from '@microsoft/fast-element';
+import { attr } from '@microsoft/fast-element';
 import { template } from './template';
 import { styles } from './styles';
 import { RichTextMarkdownParser } from '../models/markdown-parser';
@@ -19,8 +19,10 @@ export class RichTextViewer extends RichText {
      *
      * @public
      * Markdown string to render its corresponding rich text content in the component.
+     * @remarks
+     * HTML Attribute: markdown
      */
-    @observable
+    @attr
     public markdown = '';
 
     /**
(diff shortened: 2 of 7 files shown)


packages/nimble-components/src/rich-text/viewer/index.ts
```diff
@@ -19,8 +19,10 @@ export class RichTextViewer extends RichText {
      *
      * @public
      * Markdown string to render its corresponding rich text content in the component.
+     * @remarks
+     * HTML Attribute: markdown
      */
-    @observable
+    @attr
```
rajsite, line 25: Let's use the [attr `fromView` configuration](https://fast.design/docs/getting-started/fast-element#attribute-bindings) instead of the default which is `reflect`. These are likely large strings and there isn't benefit to serializing the property assignments to attributes to reflect to the DOM. It's just extra overhead in well-behaved apps.

packages/nimble-components/src/rich-text/viewer/tests/rich-text-viewer.spec.ts
```diff
@@ -76,16 +76,16 @@ describe('RichTextViewer', () => {
...
-        expect(element.markdown).toBe('');
+        expect(element.markdown).toBe('**markdown string**');
     });
 
-    it('set the markdown property and ensure there is no markdown attribute', () => {
+    it('set the markdown property and ensure there is markdown attribute', () => {
         element.markdown = '**markdown string**';
 
-        expect(element.hasAttribute('markdown')).toBeFalse();
+        expect(element.hasAttribute('markdown')).toBeTrue();
```
rajsite, line 88: With the above change I expect this should change to false as we will not serialize the property assignments back to the attribute,

change/@ni-nimble-components-33b8266d-c20f-4ec8-92a9-a6ecf81fe328.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Changed Rich Text Viewer markdown property to an attribute",
+  "packageName": "@ni/nimble-components",
+  "email": "107654185+joseahdz@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
rajsite, line 7: - Can you update the chat conversation stories to use the rich-text-viewer attribute binding instead of property binding. I think that will make the `show code` view clearer. Don't need to update the matrix tests.

change/@ni-nimble-components-33b8266d-c20f-4ec8-92a9-a6ecf81fe328.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
joseahdz: I am not sure what alternatives you want me to add to the description.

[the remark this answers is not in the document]
rajsite: Was just saying that I updated your PR description to discuss the two alternatives we had discussed. i.e. using JS to update the property value or switching the rich text viewer to use an attribute. No action needed, it was just an FYI that I changed the PR description.

rajsite (approved):

rajsite:
Minor change that should be no observable difference to exisiting users setting the property. Bypassing owners @jattasNI @vivinkrishna-ni

PR #2544, opened 2025-02-25
Created Blazor wrapper for nimble-rich-text-viewer component

joseahdz:
# Pull Request

## 🤨 Rationale

There is no Blazor wrapper for the rich text viewer component.

## 👩‍💻 Implementation

Created Blazor wrapper.

## 🧪 Testing

Added auto tests and were passing.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-blazor-e360966f-afa8-482e-992c-c0b68976d3cf.json
+++ b/change/@ni-nimble-blazor-e360966f-afa8-482e-992c-c0b68976d3cf.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added Blazor wrapper for nimble-rich-text-viewer",
+  "packageName": "@ni/nimble-blazor",
+  "email": "107654185+joseahdz@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
+++ b/packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
@@ -41,10 +41,7 @@ public partial class ComponentsDemo
                                          "4. Bulleted lists\n" +
                                          "   * Option 1\n" +
                                          "   * Option 2\n" +
-                                         "5. Absolute link: <https://nimble.ni.dev/>\n" +
-                                         "6. @mention:\n" +
-                                         "    1. User pattern: <user:1>\n" +
-                                         "    2. HTTPS pattern: <https://user/2>";
+                                         "5. Absolute link: <https://nimble.ni.dev/>\n";
 
     public IEnumerable<DemoColor> PossibleColors
     {
--- a/packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
+++ b/packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
@@ -32,6 +32,20 @@ public partial class ComponentsDemo
     private readonly HashSet<string> _recordsLoadingChildren = new();
     private readonly HashSet<string> _recordsWithLoadedChildren = new();
 
+    public const string MarkdownString = "Supported rich text formatting options:\n" +
+                                         "1. **Bold**\n" +
+                                         "2. *Italics*\n" +
+                                         "3. Numbered lists\n" +
+                                         "   1. Option 1\n" +
+                                         "   2. Option 2\n" +
+                                         "4. Bulleted lists\n" +
+                                         "   * Option 1\n" +
+                                         "   * Option 2\n" +
+                                         "5. Absolute link: <https://nimble.ni.dev/>\n" +
+                                         "6. @mention:\n" +
+                                         "    1. User pattern: <user:1>\n" +
+                                         "    2. HTTPS pattern: <https://user/2>";
+
     public IEnumerable<DemoColor> PossibleColors
     {
         get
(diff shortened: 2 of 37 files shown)


change/@ni-nimble-blazor-e360966f-afa8-482e-992c-c0b68976d3cf.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: To fully support the capabilities of the rich text viewer, we also need the `nimble-rich-text-mention-users` and `nimble-mapping-user` elements. These are used to resolve `@mention` URLs within the viewer into actual user names. You can see the difference if you compare the rendered user in the [anbular build](https://60e89457a987cf003efc0a5b-duwvxprsgs.chromatic.com/example-client-app/index.html#/customapp) and [blazor build](https://60e89457a987cf003efc0a5b-duwvxprsgs.chromatic.com/blazor-client-app/wwwroot/index.html) for this PR.

I think our options are:
1. add wrappers for these two components in this PR. This shouldn't be much work, but they do each have some attributes so the work is greater than 0.
2. decide you don't need this capability yet. But then we shouldn't include that syntax in the example and we shouldn't change the component status.

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
```diff
@@ -32,6 +32,20 @@ public partial class ComponentsDemo
     private readonly HashSet<string> _recordsLoadingChildren = new();
     private readonly HashSet<string> _recordsWithLoadedChildren = new();
```
jattasNI, line 34: When possible we try to keep the Angular and Blazor examples in sync. One tiny difference is that the Angular example [adds a border around the rich text viewer](https://github.com/ni/nimble/blob/7c26dc3c89759676d4055e1b5809ceaec888e882/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.scss#L80). I can't recall why we did that but possibly just to make the bounds of the component more clear. Could you either add the same styling to the Blazor app or remove that styling from the Angular app? My very slight preference is to add it to the Blazor app too.

change/@ni-nimble-blazor-e360966f-afa8-482e-992c-c0b68976d3cf.json
joseahdz: I don't need @mention so I am going to remove that from the example.  I will also revert the component status and will update the Blazor example to match the angular one.

packages/blazor-workspace/NimbleBlazor/Components/NimbleRichTextViewer.razor.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public partial class NimbleRichTextViewer : NimbleInputBase<string?>
+{
+    [Parameter]
+    public string? Markdown { get; set; }
+
+    [Parameter]
+    public RenderFragment? ChildContent { get; set; }
+
+    protected override bool TryParseValueFromString(string? value, out string? result, [NotNullWhen(false)] out string? validationErrorMessage)
```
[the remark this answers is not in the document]
joseahdz: This is the pattern other Blazor wrappers have.  I am simply following it.  I can delete it.

change/@ni-nimble-blazor-e360966f-afa8-482e-992c-c0b68976d3cf.json
joseahdz: Implemented option 2.

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor.cs
joseahdz: Style container to match the Angular example.

packages/blazor-workspace/NimbleBlazor/Components/NimbleRichTextViewer.razor.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Components;
+
+namespace NimbleBlazor;
+
+public partial class NimbleRichTextViewer : NimbleInputBase<string?>
```
rajsite, line 6: I'll ultimately defer to the blazor owners @atmgrifter00 @msmithNI but it seems like we should not be extending NimbleInputBase. My understanding is the NimbleInputBase features are EditForm integration of a two-way bound Value property so this seems like it won't behave well if used as the Rich Text Viewer is one-way bound Markdown property. I'd advocate not extending NimbleInputBase and saving EditoForm integration as follow-on work if desired (and if implemented either way should be tested).

packages/blazor-workspace/NimbleBlazor/Components/NimbleRichTextViewer.razor.cs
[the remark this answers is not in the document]
atmgrifter00: Ultimately, I think we should follow @rajsite's observation that this shouldn't extent `NimbleInputBase` which should remove this implementation anyway.

packages/blazor-workspace/NimbleBlazor/Components/NimbleRichTextViewer.razor.cs
joseahdz: Changed base class to `ComponentBase` after talking to Meyer.

PR #2548, opened 2025-03-06
Add usage guidance about table column mapping performance

jattasNI:
# Pull Request

## 🤨 Rationale

The impetus for #2531 was performance issues due to creating a mapping element for each row of table data. We should explicitly suggest not doing this.

## 👩‍💻 Implementation

Added usage guidance to the table column mapping story.

## 🧪 Testing

Inspect storybook build

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
+++ b/packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
@@ -61,6 +61,13 @@ table.
 - Provide a mapping for every expected record value. Because grouping is
   performed on the record value, non-mapped record values can result in multiple
   groups without group labels.
+    - Do not create mappings that are are one-to-one for each record in the
+      table for large numbers of records (such as mappings with unique text per
+      record) or update a large number of mappings frequently (such as to show
+      progress in text). Mappings should represent clear groupings or
+      categorizations and large numbers of mappings or lots of mappings with
+      unique text cause poor grouping behaviors and potential performance
+      issues.
     - To improve grouping behavior of a value that shouldn't be rendered in
       table cells, use a <Tag name={mappingEmptyTag}/> element to provide text
       for that value's group row.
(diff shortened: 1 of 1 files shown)


packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
```diff
@@ -72,6 +72,11 @@ table.
   Choose whether to enable the `text-hidden` attribute for all mappings or none.
     - Exception: You can use a spinner-only mapping in the same column as an
       icon-only or icon and text mapping.
+- The number of mappings should be bounded and should not grow with the size of
+  the table data. For example, don't add a mapping for each record in order to
+  provide a custom string for each row. Too many mappings can hurt table update
+  performance. For more background, see
+  [issue 2531](https://github.com/ni/nimble/issues/2531).
```
rajsite, line 79: To me the issue doesn't have a clear resolution, I'm not really clear what I'm supposed to take away from that background and if implementing a table column mapping I don't really want to read all that.

I think what needs most clarification is the very first bullet which says:

> Provide a mapping for every expected record value.

We should add a clarification sub bullet to that along the lines of:
```
- Provide a mapping for every expected record value....
      - Do not create mappings that are are one-to-one for each record in the table for large numbers of records (such as mappings with unique text per record) or update a large number of mappings frequently (such as to show progress in text). Mappings should represent clear groupings or categorizations and large numbers of mappings or lots of mappings with unique text cause poor grouping behaviors and potential performance issues.
```

rajsite (approved):

mollykreis (approved):

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.mdx
jattasNI: Works for me! I applied your text pretty much as you wrote it.

PR #2549, opened 2025-03-06
Improve usage guidance for select filter mode

jattasNI:
# Pull Request

## 🤨 Rationale

At UI working group this week we heard feedback that our docs implied that select should change its filter mode in response to the number of items being returned from dynamic filtering. This is not the case so we should clarify our docs.

## 👩‍💻 Implementation

Update select story docs for `filter-mode` property.

## 🧪 Testing

Storybook build

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/select/select.stories.ts
+++ b/packages/storybook/src/nimble/select/select.stories.ts
@@ -122,7 +122,7 @@ const optionSets = {
 const filterModeDescription = `
 Controls the filtering behavior of the select. The default of \`none\` results in a dropdown with no input for filtering. A non-'none' setting results in a search input placed at the top or the bottom of the dropdown when opened (depending on where the dropdown is shown relative to the component). The \`standard\` setting will perform a case-insensitive and diacritic-insensitive filtering of the available options anywhere within the text of each option. The \`manual\` setting will provide the search input in the dropdown, but performs no filtering of the options. This allows for custom filtering to be implemented by the consuming application.
 
-It is recommended that if the select has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`. It is recommended that each select should keep the same filter mode for its lifetime.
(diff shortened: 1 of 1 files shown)


packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -122,7 +122,7 @@ const optionSets = {
 const filterModeDescription = `
 Controls the filtering behavior of the select. The default of \`none\` results in a dropdown with no input for filtering. A non-'none' setting results in a search input placed at the top or the bottom of the dropdown when opened (depending on where the dropdown is shown relative to the component). The \`standard\` setting will perform a case-insensitive and diacritic-insensitive filtering of the available options anywhere within the text of each option. The \`manual\` setting will provide the search input in the dropdown, but performs no filtering of the options. This allows for custom filtering to be implemented by the consuming application.
 
-It is recommended that if the select has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`.
+It is recommended that if the select has 15 or fewer options that you use the \`none\` setting for the \`filter-mode\`. It is recommended that each select should keep the same filter mode for its lifetime.
```
jattasNI, line 125: I'll admit this guidance feels a little weird: there are other properties where the same guidance might apply but we don't bother to mention it. I considered other guidance like:
"Since filter mode affects behavior without any obvious indication to users, it is recommended not to change it after the select is first populated"

We could also just leave as-is and see if any other clients misinterpret the existing guidance. I'm open to reviewer feedback.

rajsite, line 125: What about something like:

```suggestion
If it expected that the select will always have 15 or fewer options then use the \`none\` setting for the \`filter-mode\`, otherwise some form of filtering should be enabled.
```

jattasNI: I like that better, applied with some minor tweaks
```suggestion
If it is expected that the select will always have 15 or fewer options then use the \`none\` setting for \`filter-mode\`, otherwise some form of filtering should be enabled.
```

PR #2552, opened 2025-03-12
Added new slots for the spright-chat-message to allow adding action buttons and follow-up prompts

joseahdz:
# Pull Request

## 🤨 Rationale

UX design calls for adding action buttons to chat messages to allow user to edit prompts and provide feedback on AI responses.
Also, messages can have suggested follow-up button prompts and those should be below the action buttons.

## 👩‍💻 Implementation

1. Updated spec and implementation to add slots `footer-actions` and `end` to allow adding buttons below the message.
2. Updated styles (mostly padding) to more closely match the latest designs
3. Added storybook docs for the new slots

## 🧪 Testing

New matrix test

## ✅ Checklist


change/@ni-spright-components-095116e3-8f7b-478c-8c7f-4d72964c5c4d.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @joseahdz I see this PR has been in draft for a few days. Just want to make sure and mention that for a buddy review in draft make sure to create a comment and `@mention` the person to request a review. By default folks ignore draft PRs until feedback is requested.

joseahdz: I was not aware that I needed to notify someone.  Thank you for letting me know.  

@jattasNI Can you please review my draft PR?  Thanks.

jattasNI (commented):
Mostly looked at the spec and the docs, not the implementation yet

packages/spright-components/src/chat/specs/README.md
```diff
@@ -182,7 +212,7 @@ Angular integration has not yet been evaluated in detail, but is expected to be
 
 ### Blazor integration
 
-Blazor integration has not yet been evaluated, but is expected to be able to follow existing patterns. It is anticipated to be needed for initial clients so this section will be updated when Blazor development begins if anything interesting is discovered.
+The Blazor wrappers `SpringChatConversation` and `SprightChateMessage` have been created.
 
 ### Visual Appearance
```
jattasNI, line 218: If we don't end up implementing the keyboard-accessible pattern for interacting with the hover-only edit button, we should start keeping a list of known accessibility gaps in the Accessibility section below.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -28,13 +35,17 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
     },
     render: createUserSelectedThemeStory(html`
         <${chatConversationTag}>
```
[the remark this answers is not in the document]
joseahdz: I looked at other components and I did not see them document slots in the argTypes.  Do you have an example where a slot is documented in argTypes?

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,22 @@ We will not yet introduce an input toolbar component where a user can type and s
 
 #### Chat message
 
-1. Display arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+The `spright-chat-message` is divided into four sections.
```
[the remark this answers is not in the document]
joseahdz: Updated description to use "slot elements".  I will update the names after our meeting.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,22 @@ We will not yet introduce an input toolbar component where a user can type and s
...
+
```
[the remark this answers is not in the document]
joseahdz: We met offline and discuss this feedback.  This was the outcome
1. Removed the left slot.  We cannot keyboard navigate to the hidden edit button and we did not want to always show the button since it would add noise.  So, we decided to not add the edit button on the left slot for now which meant also removing the left slot.
2. We decided to renamed the left-bottom slot to footer-actions.
3. We decided to keep the followup-prompt slot and the name.
4. I tried using a nimble-toolbar for the footer-action slot, but the toolbar did not support adding slotted elements to the toolbar.  An issue will be created to support this use case.

packages/spright-components/src/chat/specs/README.md
joseahdz: Since we no longer have an edit button, there is nothing to do so resolving this comment.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,20 @@ We will not yet introduce an input toolbar component where a user can type and s
 
 #### Chat message
 
-1. Display arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+The `spright-chat-message` has the following slot elements.
+
+1. `default` slot displays arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+1. `footer-actions` slot element which is used to add actions buttons after the main content.
+1. `followup-prompt` slot element which is used to add followup prompt buttons. They are below any action buttons.
```
jattasNI, line 54: Small miscommunication from my previous comment, we prefer "slot" over "slot element". Also changing the language to be more precise, since "after" could mean "below" or "right of"

```suggestion
1. `default` slot displays arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
1. `footer-actions` slot which is used to add actions buttons below the main content.
1. `followup-prompt` slot which is used to add followup prompt buttons. They are below any action buttons.
```

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,20 @@ We will not yet introduce an input toolbar component where a user can type and s
...
+
+All action buttons must meet the following criteria
```
jattasNI, line 56: We should document similar criteria for the followup slot. I suspect we want them all to be block buttons with text rather than icons, but that's UX's call.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -14,9 +14,22 @@ import { imgBlobUrl, markdownExample } from '../conversation/story-helpers';
 import { SpinnerAppearance } from '../../../../../nimble-components/src/spinner/types';
 import { ButtonAppearance } from '../../../../../nimble-components/src/menu-button/types';
 import { isChromatic } from '../../../utilities/isChromatic';
+import { iconThumbUpTag } from '../../../../../nimble-components/src/icons/thumb-up';
+import { iconThumbDownTag } from '../../../../../nimble-components/src/icons/thumb-down';
+
+const footerActionsDescription = `Place a button at the bottom of the message to allow the user to invoke a custom action.
```
jattasNI, line 20:
```suggestion
const footerActionsDescription = `Place 0 or more buttons at the bottom of the message to allow the user to invoke a custom action.
```

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -14,9 +14,22 @@ import { imgBlobUrl, markdownExample } from '../conversation/story-helpers';
...
+
+The buttons should have the \`ghost\` appearance and \`content-hidden\`.
+
+Nimble will set the height of the buttons to \`$ni-nimble-control-slim-height\`.
+`;
+
+const followUpPromptsDescription = 'Place buttons with followup prompts. They are below any action buttons.';
```
jattasNI, line 27: Let's add the guidance about follow up button appearance and content here too.
```suggestion
const followUpPromptsDescription = 'Place 0 or more buttons with followup prompts. They appear below any action buttons.';
```

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -14,9 +14,22 @@ import { imgBlobUrl, markdownExample } from '../conversation/story-helpers';
 import { SpinnerAppearance } from '../../../../../nimble-components/src/spinner/types';
 import { ButtonAppearance } from '../../../../../nimble-components/src/menu-button/types';
 import { isChromatic } from '../../../utilities/isChromatic';
+import { iconThumbUpTag } from '../../../../../nimble-components/src/icons/thumb-up';
+import { iconThumbDownTag } from '../../../../../nimble-components/src/icons/thumb-down';
+
```
jattasNI, line 19: We should also add guidance that footer actions and follow up prompts should only be added to inbound messages (or do the refactor to create a different component for each message type).

packages/spright-components/src/chat/specs/README.md
jattasNI, line 54: Now that we're settled on the follow up prompt being in a slot rather than a separate message, we should agree on the name. I'd like to avoid "prompt" since it's AI-specific. Alternatives:

1. `slim-footer-actions` and `footer-actions`
2. `footer-actions` and `footer-2-actions`
3. `footer-actions` and `sub-footer-actions`

At the moment I like 1 best but I'll ping other owners for their opinions.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -182,7 +205,7 @@ Angular integration has not yet been evaluated in detail, but is expected to be
 
 ### Blazor integration
 
-Blazor integration has not yet been evaluated, but is expected to be able to follow existing patterns. It is anticipated to be needed for initial clients so this section will be updated when Blazor development begins if anything interesting is discovered.
+The Blazor wrappers `SpringChatConversation` and `SprightChateMessage` have been created.
```
jattasNI, line 208:
```suggestion
The Blazor wrappers `SpringChatConversation` and `SprightChatMessage` have been created.
```

packages/spright-components/src/chat/specs/README.md
```diff
@@ -182,7 +205,7 @@ Angular integration has not yet been evaluated in detail, but is expected to be
...
 
 ### Visual Appearance
```
jattasNI, line 211: Let's mention the toolbar alternative and link to the issue from the spec somewhere. You could do that in the Accessibility section or an Alternatives section.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -44,7 +50,34 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
                 ></${spinnerTag}>
             </${chatMessageTag}>
             <${chatMessageTag} message-type="${() => ChatMessageType.inbound}">
+                <${buttonTag} slot='footer-actions' appearance='ghost' content-hidden>
```
jattasNI, line 53: The [nimble-button accessibility guidance](https://nimble.ni.dev/storybook/?path=/docs/components-button--docs#accessibility) suggests also setting a `title` whenever you set content-hidden. I think that makes sense for this use case. Could you update the user-visible stories to do this for each of these buttons? (Just the ones that are content-hidden, not the ones with visible text content)

I considered also asking for you to document this suggestion in the slot's docs, but I think I'd rather not duplicate the guidance that we already have in the button docs.

packages/spright-components/src/chat/message/styles.ts
```diff
@@ -31,21 +34,34 @@ export const styles = css`
...
         border: ${borderWidth} solid ${borderHoverColor};
         border-radius: 8px 8px 0px 8px;
     }
 
-    :host([message-type='inbound']) div {
+    :host([message-type='inbound']) .message-content {
         border-radius: 8px 8px 8px 0px;
     }
+
+    .footer-actions::slotted(${buttonTag}) {
```
jattasNI, line 58: Can you double check the design spec and/or chat with Brandon about the spacing? To me it looks strange that the hover outline is right up against the message, I'd expect a bit of padding between them. But maybe it's OK if this will only be added to messages that don't have a visible background.

It's also rare for us to use 10px spacing, usually things are multiples of 4.

image

packages/spright-components/src/chat/specs/README.md
rajsite: For implementing the `end` slot we should follow the exisiting start-end pattern.

So for example, see how the combobox applies the `StartEnd` mixin to the class:
https://github.com/ni/nimble/blob/175f1ab564314a51cb6eb6446dc1eb5eff3805f1/packages/nimble-components/src/combobox/index.ts#L885-L886

It then uses the `startSlotTemplate` and `endSlotTemplate` helpers in the template:
https://github.com/ni/nimble/blob/175f1ab564314a51cb6eb6446dc1eb5eff3805f1/packages/nimble-components/src/combobox/template.ts#L40
https://github.com/ni/nimble/blob/175f1ab564314a51cb6eb6446dc1eb5eff3805f1/packages/nimble-components/src/combobox/template.ts#L69

And then you can style the slot using the part selector `[part="end"] { /* styles */ }`.

Currently we are only using the `end` template but we should still include both start and end, for example:

```js
    <div class="root">
        ${startSlotTemplate(context, definition)}
        <section class="message-content">
            <slot></slot>
        </section>
        <section>
           <slot class="footer-actions" name="footer-actions"></slot>
        </section>
        ${endSlotTemplate(context, definition)}
    </div>
```

But we should hide the start template with styles since it is unneeded currently:

```css
        [part='start'] {
            display: none;
        }

        [part='end'] {
             /* Styles */
        }

```

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -45,6 +68,28 @@ export const chatMessageText: StoryObj<ChatMessageTextArgs> = {
     render: createUserSelectedThemeStory(html`
```
rajsite, line 68: This indentation seems surprising in this section. You can start this section with prettier-ignore to manually control the indentation

```suggestion
    // prettier-ignore
    render: createUserSelectedThemeStory(html`
```

packages/spright-components/src/chat/specs/README.md
joseahdz: Made the suggested edits.

packages/spright-components/src/chat/specs/README.md
joseahdz: Added criteria for the end slot.  For now, I added that the appearance for a button should be `Outline` since that was what I was in many of the Figma explorations.  I asked Brandon for an official design for the action and follow up buttons.

I plan to submit the changes that I have now since I don't know how long it will take Brandon to give me the design. I will submit any updates on a different PR.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
joseahdz: Updated description and update constant name to be endButtonDescription.  The slot name is end.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
joseahdz: Updated description as suggested.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
joseahdz: Added to the description of the footer action and end buttons to only be added to inbound messages.

packages/spright-components/src/chat/specs/README.md
joseahdz: Changed slot name to `end` and started using the start-end pattern.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,20 @@ We will not yet introduce an input toolbar component where a user can type and s
...
-1. Display arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+The `spright-chat-message` has the following slot elements.
+
+1. `default` slot displays arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+1. `footer-actions` slot element which is used to add actions buttons after the main content.
+1. `followup-prompt` slot element which is used to add followup prompt buttons. They are below any action buttons.
+
+All action buttons must meet the following criteria
+
+1. They are `nimble-button`s
```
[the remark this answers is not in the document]
joseahdz: Added that `nimble-toggle-buttons` can also be used.

packages/spright-components/src/chat/specs/README.md
joseahdz: Fixed mispelling

packages/spright-components/src/chat/specs/README.md
joseahdz: Added to the Accessibility section and added a link to the toolbar issue.

packages/spright-components/src/chat/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: If that's the case, then doesn't the styling need to be updated to support toggle buttons? As is, it only supports `buttonTag`.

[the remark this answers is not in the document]
joseahdz: Update styling to include toggle buttons.

packages/spright-components/src/chat/message/styles.ts
joseahdz: I am going to meet with UX team and I will update this comment with the outcome.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -67,6 +113,30 @@ export const chatMessageRichText: StoryObj<ChatMessageRichTextArgs> = {
...
         <${chatMessageTag} message-type="${x => ChatMessageType[x.messageType]}">
             <${richTextViewerTag} markdown="${x => x.markdown}"></${richTextViewerTag}>
+            ${when(
+        x => x.footerActions,
+        html`
```
rajsite, line 118: Similar to the above example, this whitespace looks unusual. I won't comment on all instances as there are several more below but please update them as well

packages/spright-components/src/chat/message/styles.ts
joseahdz: After talking to the UX team, adding action buttons on an outbound message is not supported.  So, they want to leave it as is.  Also, the spacing between the action buttons should be 16 pixels.  Also, the gap between the action buttons the followup prompts should be 32 pixels.  I made those changes and resolving this comment.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
joseahdz: Not sure what happened.  I made those changes, but got undone.  I remade them.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
```diff
@@ -119,6 +144,15 @@ export const systemSizing: StoryFn = createStory(html`
     ])}
 `);
 
+export const messageInteractionsThemeMatrix: StoryFn = createMatrixThemeStory(
```
jattasNI, line 147: It's good that you're adding matrix test coverage for the new slots. But there are a few issues:
1. it's not necessary to include coverage for states combinations like hover and active. Those states will be covered by existing button tests and we don't need to verify that they still work inside a message slot
2. we should add coverage for layout and sizing. If there are multiple buttons, what's the default padding between them? If there are more buttons than the width of the message, how do they layout? If there are more buttons than fit in the width of the conversation, how do they layout?

For 2 I think the designers are still deciding but we could add the tests with whatever behavior we get today and then when we adopt the designers' suggestions we'll see a useful diff in the snapshots for the tests.

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -450,7 +450,13 @@
                     <NimbleSpinner Appearance="SpinnerAppearance.Accent"></NimbleSpinner>
                 </SprightChatMessage>
                 <SprightChatMessage message-type="inbound">
+                    <NimbleButton slot="footer-actions" Appearance=ButtonAppearance.Ghost ContentHidden="true">
```
jattasNI, line 453: We should make sure the example app follows our usage guidance. So if we're going to use the footer-actions slot it should be on an outbound message. (Or use the end slot with an inbound message, or both)

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,27 @@ We will not yet introduce an input toolbar component where a user can type and s
 
 #### Chat message
 
-1. Display arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+The `spright-chat-message` has the following slot elements.
+
+1. `default` slot displays arbitrary slotted content. For example: text, rich text, buttons, images, or a spinner.
+1. `footer-actions` slot which is used to add actions buttons below the main content.
```
jattasNI, line 53:
```suggestion
1. `footer-actions` slot which is used to add action buttons below the main content.
```

packages/spright-components/src/chat/specs/README.md
```diff
@@ -47,7 +47,27 @@ We will not yet introduce an input toolbar component where a user can type and s
...
+1. They are `nimble-button`s or `nimble-toggle-button`
+1. The `appearance` attribute is set to `ghost`
+1. The `content-hidden` attribute is set to `true`
+1. They contain only Nimble icons
+
+All end text buttons must meet the following criteria
+
+1. They are `nimble-button`s
+1. The `apperance` attribute is set to `block`
+1. The buttons only have text
```
jattasNI, line 67: nit: copy edits
```suggestion
Nimble will set the height of the action buttons to `$ni-nimble-control-slim-height`. All action buttons must meet the following criteria:

1. They are `nimble-button`s or `nimble-toggle-button`s
1. The `appearance` attribute is set to `ghost`
1. The `content-hidden` attribute is set to `true`
1. They contain only Nimble icons

All end text buttons must meet the following criteria:

1. They are `nimble-button`s
1. The `apperance` attribute is set to `block`
1. The buttons only have text
```

packages/spright-components/src/chat/specs/README.md
```diff
@@ -131,8 +151,12 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
     - A message will grow its height to fit its content, with no maximum height.
     - Clients could override this behavior but we don't anticipate use cases for doing so when the message is used within a conversation
 - _Slots_
-
-    - arbitrary content can be added to the default slot to be displayed within the message
+    - `footer-actions`
+        - Action buttons to display after the main content.
+    - `end`
+        - Buttons with text that are display at the bottom after any action buttons.
```
jattasNI, line 157:
```suggestion
        - Buttons with text that are displayed at the bottom after any action buttons.
```

packages/spright-components/src/chat/specs/README.md
```diff
@@ -198,7 +229,9 @@ None.
 
 ### Accessibility
 
-Accessibility has not yet been evaluated.
+Only keyboard navigation has been evaluated. This has led to the removal of a showing an action button only on hover. Users cannot navigate to a hidden button. We wil revisit this in the future if desired behavior is required.
+
+For the footer actions elements, they need to be in a `nimble-toolbar`, however, the `nimble-toolbar` does not support content slotted across a shadow root. [This](https://github.com/ni/nimble/issues/2571) issue has been opened to fix it.
```
jattasNI, line 234:
```suggestion
Only keyboard navigation has been evaluated. The desired behavior is for each message's action buttons to be a single tab stop, with navigation between a message's action buttons accomplished using arrow keys. This should be achieved by placing the items within a `nimble-toolbar`, however the `nimble-toolbar` does not yet detect content in nested slots (see #2571). For scoping reasons the content will not initially be placed in a toolbar and thus each button will be a tab stop. Once that issue is addressed it should be possible to add a toolbar and achieve the desired behavior without breaking clients.

Buttons placed in the `end` slot should each be their own tab stop and thus will not be placed in a toolbar.

The design spec includes a proposal to add another "Edit" button to the left of some messages when the user hovers over the message. The button would be a tab stop and would become visible when keyboard focused. This work is currently out of scope.

Other aspects of accessibility have not yet been evaluated.
```

packages/spright-components/src/chat/message/index.ts
```diff
@@ -10,6 +17,12 @@ declare global {
     }
 }
 
+/**
+ * SprightChatMessage configuration options
+ * @public
+ */
+export type ChatMessageOptions = FoundationElementDefinition & StartEndOptions;
```
jattasNI, line 24: I'm not familiar with the our/FAST's StartEndOptions pattern. It looks like this follows what the tab does by exporting this type and using it from the template. But I was expecting to see the pattern in more other components. @rajsite could you double check that it's being applied correctly?

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
```diff
@@ -55,6 +59,25 @@ const contentHeightStates = [
...
 
+const component = ([
+    _messageTypeLabel,
+    messageType
+]: MessageTypeStates): ViewTemplate => html`
+    <span>${() => `Message Type: ${_messageTypeLabel}`} </span>
+    <${chatConversationTag}>
+        <${chatMessageTag} message-type="${() => messageType}">
+            <${buttonTag} slot='footer-actions' appearance='ghost' ContentHidden='true'>
+                <${iconCopyTextTag} slot='start' />
```
rajsite, line 70: html does not support self-closing tags, make sure to specify a closing tag
```
<${iconCopyTextTag} slot='start'></${iconCopyTextTag}>
```

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -44,7 +51,34 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
                 ></${spinnerTag}>
             </${chatMessageTag}>
             <${chatMessageTag} message-type="${() => ChatMessageType.inbound}">
+                <${buttonTag} slot='footer-actions' appearance='ghost' title='Copy' content-hidden>
+                    <${iconCopyTextTag} slot='start' />
```
rajsite, line 55: Same as above, for all these icons make sure to avoid self closing tags.

packages/spright-components/src/chat/message/styles.ts
```diff
@@ -31,21 +35,40 @@ export const styles = css`
...
+        height: ${controlSlimHeight};
+        width: ${controlSlimHeight};
+        margin: 0px 16px 0px 0px;
+    }
+
+    [part='start'] {
+        display: none;
+    }
+
+    [part='end'] ::slotted(*) {
```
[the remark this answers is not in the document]
joseahdz: I will greatly appreciate if make those changes.  Thank you.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -14,9 +14,23 @@ import { imgBlobUrl, markdownExample } from '../conversation/story-helpers';
 import { SpinnerAppearance } from '../../../../../nimble-components/src/spinner/types';
 import { ButtonAppearance } from '../../../../../nimble-components/src/menu-button/types';
 import { isChromatic } from '../../../utilities/isChromatic';
+import { iconThumbUpTag } from '../../../../../nimble-components/src/icons/thumb-up';
+import { iconThumbDownTag } from '../../../../../nimble-components/src/icons/thumb-down';
+
+const footerActionsDescription = `Place 0 or more buttons at the bottom of the message to allow the user to invoke a custom action.
+Footer actions should only be added to inbound messages.
```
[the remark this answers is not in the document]
joseahdz: `inbound` is correct.  Outbound messages are the user prompt messages.  The inbound are the responses from the AI.

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
joseahdz: footer-actions should only be on inbound messages.  Outbound messages should not have any action buttons.  Unless I am missed something, they way it is now is correct.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
joseahdz: Are there examples I can look at for those types of tests?  If not, I am concern about how much time it will take to write them.  I am under a very tight schedule and I need to minimize the amount of time I spend on non-chat Blazor UI changes.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
```diff
@@ -55,6 +59,25 @@ const contentHeightStates = [
...
 type ContentHeightStates = (typeof contentHeightStates)[number];
 
+const component = ([
+    _messageTypeLabel,
+    messageType
+]: MessageTypeStates): ViewTemplate => html`
+    <span>${() => `Message Type: ${_messageTypeLabel}`} </span>
+    <${chatConversationTag}>
+        <${chatMessageTag} message-type="${() => messageType}">
+            <${buttonTag} slot='footer-actions' appearance='ghost' ContentHidden='true'>
```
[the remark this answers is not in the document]
joseahdz: updated

packages/spright-components/src/chat/specs/README.md
joseahdz: Updated

packages/spright-components/src/chat/specs/README.md
joseahdz: Added new text

packages/spright-components/src/chat/specs/README.md
joseahdz: Updated text as suggested.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
joseahdz: Fixed.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
joseahdz: Fixed all instances.

packages/spright-components/src/chat/specs/README.md
joseahdz: Updated

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
jattasNI: My bad, I got it switched around.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
jattasNI: [Your existing conversation sizing](https://60e89457a987cf003efc0a5b-xczzitaxvl.chromatic.com/?path=/story/tests-spright-chat-conversation--outbound-sizing) tests cover some similar cases. You could just add a few more cases to those:
1. a couple footer action buttons and end buttons on the same message
1. lots of footer action buttons 
    - wider than the message but narrower than the conversation
    - wider than the conversation
3. lots of end buttons 
    - wider than the message but narrower than the conversation
    - wider than the conversation

jattasNI: Milan and I discussed a couple more ideas. Let me take a stab at this along with the styling changes. I think we want some of this in a new matrix test and some of it in the existing sizing tests.

packages/spright-components/src/chat/message/styles.ts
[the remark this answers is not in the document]
jattasNI: I'm not sure I'll get to this today but tomorrow morning looks open.

packages/spright-components/src/chat/message/index.ts
jattasNI: I learned more about this pattern while I contributed to the PR and it looks good to me. Resolving.

packages/spright-components/src/chat/message/index.ts
```diff
@@ -22,7 +35,23 @@ export class ChatMessage extends FoundationElement {
...
+
+    /** @internal */
+    @observable
+    public readonly slottedFooterActionsElements?: HTMLElement[];
+
+    public slottedFooterActionsElementsChanged(
+        _prev: HTMLElement[] | undefined,
+        next: HTMLElement[] | undefined
+    ): void {
+        this.footerActionsIsEmpty = !next?.length;
```
jattasNI, line 51: This pattern (and the corresponding code in the template which updates the classes on the element based on this attribute) is similar to what the FAST `endSlotTemplate` does (it removes the `end` class from the element when nothing is slotted). We also do this in the [Nimble dialog footer](https://github.com/ni/nimble/blob/c37746a42e97c24a46588f7fd3ae6f9f730a2a2e/packages/nimble-components/src/dialog/index.ts#L60C1-L67C1).

jattasNI (commented):
I pushed updates to the styles and tests. I still need to re-verify the storybook and chromatic diffs; I'll reset owners tomorrow once it looks good.

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
jattasNI: I ended up putting all of these cases in a new matrix story.

packages/spright-components/src/chat/message/styles.ts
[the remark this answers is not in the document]
jattasNI: I pushed these updates. If you're interested in learning some new CSS skills I'm happy to discuss how it works now.

rajsite (approved):
Looks great!

packages/spright-components/src/chat/message/index.ts
jattasNI: When it has baseline support we should consider using `:has-slotted` instead. Filed this tech debt to track it: #2579

PR #2561, opened 2025-03-20
Fix color of placeholder text in disabled select

### LINKED ISSUE #2555: Disabled select uses wrong placeholder text color
# 🐛 Bug Report

The placeholder text on the `nimble-select` is not the correct color when the select is disabled. Particularly, this text color does not change based on whether or not the select is disabled. As a result, the placeholder text in a disabled select is more prominent than the placeholder text in any other disabled nimble control.

Image

## 🤔 Expected Behavior

The placeholder text for all disabled nimble controls should be the same color.

## 💁 Possible Solution

Update the CSS for the select to use the correct token for the placeholder text color when the select is disabled.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2555 

## 👩‍💻 Implementation

There are six Nimble components that have placeholder text:
- combobox
- number-field
- rich-text editor
- select
- text-area
- text-field

All of them except the select already dim the placeholder text when the component is disabled.

Because the combobox and select use different CSS selectors to control this styling (`:host([disabled]) .selected-value.placeholder` vs `.selected-value[disabled]::placeholder`), I added the missing styling to the select's stylesheet rather than the shared "dropdown" stylesheet .

## 🧪 Testing

Added visual test case for select placeholder text. Checked in local Storybook build.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-2f1afb28-1f86-43f7-9618-d0e54d7221e0.json
+++ b/change/@ni-nimble-components-2f1afb28-1f86-43f7-9618-d0e54d7221e0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix color of placeholder text in disabled select",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/select/styles.ts
+++ b/packages/nimble-components/src/select/styles.ts
@@ -4,6 +4,7 @@ import { styles as dropdownStyles } from '../patterns/dropdown/styles';
 import { styles as errorStyles } from '../patterns/error/styles';
 import { styles as requiredVisibleStyles } from '../patterns/required-visible/styles';
 import {
+    bodyDisabledFontColor,
     borderRgbPartialColor,
     borderWidth,
     controlHeight,
@@ -35,6 +36,10 @@ export const styles = css`
 
     .selected-value.placeholder {
         color: ${placeholderFontColor};
+    }
+
+    :host([disabled]) .selected-value.placeholder {
+        color: ${bodyDisabledFontColor};
     }
 
     .selected-value {
--- a/packages/storybook/src/nimble/select/select-matrix.stories.ts
+++ b/packages/storybook/src/nimble/select/select-matrix.stories.ts
@@ -18,8 +18,8 @@ import {
     type DisabledState,
     type ErrorState,
     errorStates,
-    errorStatesNoError,
-    errorStatesErrorNoMessage
+    requiredVisibleStates,
+    type RequiredVisibleState
 } from '../../utilities/states';
 import { hiddenWrapper } from '../../utilities/hidden';
 import { textCustomizationWrapper } from '../../utilities/text-customization';
@@ -33,8 +33,9 @@ const appearanceStates = [
 type AppearanceState = (typeof appearanceStates)[number];
 
 const valueStates = [
-    ['Short Value', 'Option 1'],
-    ['Long Value', loremIpsum]
+    ['Short Value', 1],
+    ['Long Value', 2],
+    ['Placeholder', undefined]
 ] as const;
 type ValueState = (typeof valueStates)[number];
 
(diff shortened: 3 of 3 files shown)


change/@ni-nimble-components-2f1afb28-1f86-43f7-9618-d0e54d7221e0.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Curious for your thoughts on auto testing this. The [select theme matrix](https://60e89457a987cf003efc0a5b-atkkxnkzpr.chromatic.com/?path=/story/tests-select--select-theme-matrix) doesn't include placeholders but we have [a story](https://60e89457a987cf003efc0a5b-atkkxnkzpr.chromatic.com/?path=/story/tests-select--select-below-open-no-filter-light-theme-white-background-with-placeholder) for each theme with the value set to a placeholder. Would it be easy to add a second select to each of those stories that's disabled?

m-akinc: Probably would rather add the case to the matrix story. I considered the fact that this had been a visual test gap (was kind of surprised that no visual tests had diffs), but decided it wasn't really that important. But since you brought it up, I'll go ahead and add a test case.

m-akinc: Actually, the select theme matrix story was taking a long time to load, so I moved the "required asterisk" testing into a separate story and created a separate story for placeholder text.

jattasNI: Reminder to update the PR description Testing section

jattasNI (approved):

change/@ni-nimble-components-2f1afb28-1f86-43f7-9618-d0e54d7221e0.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
+  "comment": "Fix color of placeholder text in disabled select",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
```
rajsite, line 7: Couple top-level concerns:
- @m-akinc @jattasNI I feel like this is a frequent piece of feedback, but here goes: please make sure to check for consistency. With this change requiredVisible + disabled changes label just for the select. But at least two other controls that I looked at, combobox and textarea are now inconsistent. I suspect more are. PR description doesn't mention anything about how this change will ensure consistency.
  - Additionally, I don't want to give this feedback anymore. I know process sucks but I feel like I give this feedback way too often. What process changes do we need to make sure I don't need to be the backstop for that feedback?
- This change is splitting the matrix tests without clear reasons. Don't split the matrix tests without clear reasons. We want few, thorough matrix tests that help us make sure we are covering the matrix of states.
  - We should not split the matrix without clear justification. Doing it offhand without consistency across our stories makes it really difficult to know if our test coverage is consistent.
  - A reason could be that we want to split the states that impact labels out into a separate matrix. That could be a fair reason. Most control states do not impact only the label so all the label states add unnecessary permutations to the main matrix story. If that is the reason then be clear and consistent. Let's add placeholder to the main matrix and have a separate matrix for labels that include relevant states (disabled, placeholder) and required visible. I don't see why placeholder has it's own matrix.

rajsite (changes requested):
Change adds inconsistency and breaks testing conventions

change/@ni-nimble-components-2f1afb28-1f86-43f7-9618-d0e54d7221e0.json
rajsite: Talked with mert offline, got mixed up and this PR is about placeholder + disabled and not requiredVisisble + disabled. General statements still apply, verifying across controls that placeholder text + disabled works and making test coverage consisent.

m-akinc: I updated the PR description with details about consistency across components. I also put the required asterisk and placeholder testing back into the select matrix test. It ends up being quite large (and includes combinations that I'm not sure are very important/distinct), but I guess those aren't really problems.

rajsite (approved):

Issue #2528, opened 2025-01-30
Storybook build breaks if Spright depends on Nimble


jattasNI:
# 🧹 Tech Debt

While developing the chat components I hit build issues in Storybook caused by a dependency I introduced from spright-components to nimble-components. For the chat components I worked around it by eliminating the dependency but that's not a viable long term strategy.

The problem is caused by the different ways that the storybook package and the spright-components package import content from nimble-components. All the storybook imports are relative paths to source (to keep turbosnap working) but all the components imports are package imports `@ni/nimble-components` (to ensure correct package dependencies in the build output). This results in 2 copies of nimble-components classes being pulled in to the storybook build, which leads to type conflicts.

@rajsite thought we might be able to use tsconfig type mappings similar to what we do in SystemLinkShared. We would apply this to storybook's config to map package paths to the corresponding relative paths. We should try this out and see if we can get the build and turbosnap working together.

rajsite:
[Found an existing chromatic issue](https://github.com/chromaui/chromatic-cli/issues/1149) where folks are having issues with monorepos and a description of the approach I was thinking of at stand-up is more spelled out (relying on package imports/exports and conditions). We might need to [switch over to package entry points](https://github.com/ni/nimble/issues/172) to even try it but guessing will hit issues as well.

rajsite:
The [`tsconfig-paths-webpack-plugin`](https://github.com/storybookjs/storybook/issues/14087#issuecomment-1339108022) may be useful for expressing paths in the tsconfig and having Webpack build be aware of it. Based on [discussion from the exisiting chromatic monorepo issue](https://github.com/chromaui/chromatic-cli/issues/1149#issuecomment-2744859166).

PR #2567, opened 2025-03-26
Avoid duplicate imports in storybook for spright components

rajsite:
# Pull Request

## 🤨 Rationale

Fixes https://github.com/ni/nimble/issues/2528

- Last year when introducing the spright package in the monorepo, storybook was split out into a separate package.
- Placing storybook in a separate package [broke turbosnap](https://github.com/ni/nimble/issues/2118) as import paths referenced package identifiers and not paths to files in git (build output in dist vs checked in src)
- A [workaround](https://github.com/ni/nimble/pull/2131) was introduced to reference source paths across package boundaries directly 🤒.
- The workaround [breaks if a spright component depends on a nimble component](https://github.com/ni/nimble/issues/2528) because storybook will have an absolute reference, i.e. text field story `../../../../nimble-components/src/text-field` and in the spright component `@ni/nimble-components/dist/esm/text-field`

## 👩‍💻 Implementation

- Found a [storybook issue](https://github.com/storybookjs/storybook/issues/14087#issuecomment-1339108022) that described the package [`tsconfig-paths-webpack-plugin`](https://github.com/dividab/tsconfig-paths-webpack-plugin) where we can define a tsconfig path that gets types to resolve correctly and the same paths are picked up by webpack so the resulting build also uses the same source path.
- Added tsconfig paths that redirect `@ni/nimble-components/dist/esm/` imports to `@ni/nimble-components/src`
- Added tsconfig paths that redirect `@ni/spright-components/dist/esm/` imports to `@ni/spright-components/src`

Alternatives / potential future work:
- We might be able to switch back to normal imports in storybook across all files and rely more heavily on the plugin to redirect `dist/esm` imports to `src`. I did not attempt this or validate that turbosnap would still function with that change.
- There is a corresponding chromatic issue where others have ran into problems with similar monorepo setups. Once resolved hopefully we can remove all the workarounds. See https://github.com/chromaui/chromatic-cli/issues/1149

## 🧪 Testing

- Created a [test commit](https://github.com/ni/nimble/pull/2567/commits/0eb68e5c161318f9e0bd8249b14c01239c1c91b9) where the [build failed](https://github.com/ni/nimble/actions/runs/14073130172/job/39411209846#step:12:221):
   ```
   Error: ../nimble-components/dist/esm/text-field/index.d.ts(5,9): error TS2717: Subsequent property declarations must have the same type.  Property ''nimble-text-field'' must be of type 'TextField', but here has type 'TextField'.
   ```
- [Applied the webpack](https://github.com/ni/nimble/pull/2567/commits/b8bc960d94237bd7911dc302dbe48b2172a0b086) and tsconfig path changes and the build passed

## ✅ Checklist

### DIFF
--- a/packages/storybook/.storybook/main.js
+++ b/packages/storybook/.storybook/main.js
@@ -1,6 +1,7 @@
 import remarkGfm from 'remark-gfm';
 import CircularDependencyPlugin from 'circular-dependency-plugin';
 import TerserPlugin from 'terser-webpack-plugin';
+import TsconfigPathsPlugin from 'tsconfig-paths-webpack-plugin';
 
 // All files participating in storybook should be in src
 // so that TypeScript and linters can track them correctly
@@ -42,6 +43,9 @@ export function webpackFinal(config) {
             }
         ]
     });
+    config.resolve.plugins = [
+        new TsconfigPathsPlugin(),
+    ];
     config.plugins.push(
         new CircularDependencyPlugin({
             exclude: /node_modules/,
--- a/packages/storybook/.storybook/main.js
+++ b/packages/storybook/.storybook/main.js
@@ -1,7 +1,6 @@
 import remarkGfm from 'remark-gfm';
 import CircularDependencyPlugin from 'circular-dependency-plugin';
 import TerserPlugin from 'terser-webpack-plugin';
-import TsconfigPathsPlugin from 'tsconfig-paths-webpack-plugin';
 
 // All files participating in storybook should be in src
 // so that TypeScript and linters can track them correctly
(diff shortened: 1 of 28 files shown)


fredvisser (approved):
Nice - 👍

packages/storybook/.storybook/main.js
```diff
@@ -1,6 +1,7 @@
 import remarkGfm from 'remark-gfm';
```
jattasNI, line 1: Thanks for tackling this! My only suggestion would be to double check that TurboSnap is still working as expected. Looks like it was disabled for this build because of the changes to `main.js`. I'd be fine with doing that verification after completing this PR in whatever next PR is created that makes a component change.

jattasNI (approved):

PR #2572, opened 2025-03-27
Document placeholder attribute for number-field

### LINKED ISSUE #2565: Number field's `placeholder` attribute is not documented
# 🐛 Bug Report

The number field supports setting placeholder text via a `placeholder` attribute (just like the text field), but it does not include this attribute in the Storybook doc (unlike the text field). Interestingly, we do set the placeholder attribute in the Storybook (set to same value as label), and you can see this if you click "Show code," but you cannot modify or disable it via the story controls.

## 💻 Repro or Code Sample

Image

## 🤔 Expected Behavior

Placeholder attribute should be documented just like in the text field story.

m-akinc:
# Pull Request

## 🤨 Rationale

Fixes #2565 

## 👩‍💻 Implementation

Documented in same way as text-field.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/number-field/number-field.stories.ts
+++ b/packages/storybook/src/nimble/number-field/number-field.stories.ts
@@ -21,12 +21,14 @@ import {
     errorTextDescription,
     errorVisibleDescription,
     slottedLabelDescription,
-    requiredVisibleDescription
+    requiredVisibleDescription,
+    placeholderDescription
 } from '../../utilities/storybook';
 
 interface NumberFieldArgs extends LabelUserArgs {
     label: string;
     value: number;
+    placeholder: string;
     step: number;
     hideStep: boolean;
     min: number;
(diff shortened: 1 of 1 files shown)


packages/storybook/src/nimble/number-field/number-field.stories.ts
```diff
@@ -78,6 +80,12 @@ const metadata: Meta<NumberFieldArgs> = {
                 'The number displayed in the number field. Note that the property value is not synced to an attribute.',
             table: { category: apiCategory.nonAttributeProperties }
         },
+        placeholder: {
```
rajsite, line 83: Can you make the label and placeholder independently configurable? Similar to the text field story

rajsite: oh you just pushed that, nice

rajsite (approved):

packages/storybook/src/nimble/number-field/number-field.stories.ts
m-akinc: Yep, was an oversight.

PR #2573, opened 2025-03-28
Update Actions to Ubuntu 24.04 and Playwright to 1.51.0

rajsite:
# Pull Request

## 🤨 Rationale

- Updates playwright to latest versions available in both npm and nuget
- Updates the GitHub Action runner to the latest ubuntu release

## 👩‍💻 Implementation

- Performed updates above but did not attempt to re-enable previously failing tests mentioned in Manual Tasks issue. Will attempt to re-enable those in a follow-up (or when Manual Tasks are re-evaluated, whichever is first).
- In the Angular workspace ran into error:
   ```
   Cannot start ChromeHeadless
   [test:sequential                  ] 	[9840:9840:0328/170111.541561:FATAL:zygote_host_impl_linux.cc(132)] No usable sandbox! If you are running on Ubuntu 23.10+ or another Linux distro that has disabled unprivileged user namespaces with AppArmor, see https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md. Otherwise see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information on developing with the (older) SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.
   ```
   Looking at other karma configurations we already had the flag `--no-sandbox` configured for chrome. I just updated the karma workspace to use the same set of chrome flags as all the others.

## 🧪 Testing

- Rely on CI

## ✅ Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         if: env.HAS_GITHUBPAGESDEPLOYTOKEN == 'true'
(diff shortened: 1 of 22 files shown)


rajsite:
@jattasNI gonna override merge to get runtime, can address any feedback in follow-ups

Issue #1855, opened 2024-02-21
Bring number-column formatting options to nimble-number-field 


fredvisser:
## 😯 Problem to Solve

Clients like DFF require locale-specific, decimal number formatting in the `nimble-number-field`.

## 💁 Proposed Solution

Adopt the decimal formatting options from the `nimble-table-column-number-text` component.
- Support for the `format` property including the `default` and `decimal` options
- Support for the `decimal-digits` and `decimal-maximum-digits` properties

Adopt the automatically locale specific number formatting from the `nimble-table-column-number-text` component.
- E.g. Should render `1,002.9` as `1.002,9` for German locales.

## 📋 Tasks
- [ ] update nimble-number-field custom element
- [ ] update nimble-number-field unit tests
- [ ] update nimble-number-field Storybook tests
- [ ] update nimble-number-field Storybook docs
- [ ] update nimble-number-field Angular wrapper
- [ ] update nimble-number-field Blazor wrapper
- [ ] Add to Angular example app
- [ ] Add to Blazor example app

rajsite:
@fredvisser is this for number inputs or number rendering as text? Number input support is likely to be challenging.

jattasNI:
Some of the challenges if we want to support formatting inputs:

1. localization. Right now we always use `.` for decimal separator and `,` for thousands separator, but some locales use the opposite or other characters like space. If the user types one of those characters, how do we interpret it? If we decide to allow inputting only in en-us, will it be weird to switch the format to a different locale after the user commits the value?
2. if a number is being formatted with a limited number of digits but the user inputs a higher precision number, what happens to the extra digits? Are they discarded completely? Or are they persisted in the value but truncated from the display?

ChrisDecombe:
Answer to point 1: The input format should be depending on the language set on the UI (Browser). YOu should save always in the same format independent to the display format.

Answer to point 2: If the field is configured with 2 digits after the comma. If you try to enter a third digit this digit should not be displayed in the input field and it shouldn't be saved.

rajsite:
Ran into the maskito library which looks potentially promising for building up some forms of formatted text input: https://maskito.dev/getting-started/what-is-maskito

jattasNI:
Very tentative proposal outline after 15 minutes of chatting with @fredvisser. Feedback welcome.
1. fork FAST template and base class, rip out their validation of allowed characters/min/max
2. integrate the maskito library linked above
[...]

PR #2586, opened 2025-04-08
Number Field IxD Update

fredvisser:
# Pull Request

## 🤨 Rationale

https://github.com/ni/nimble/issues/1855 asks for several new number field features. This IxD spec aims to define requirements that could satisfy them.

In short:

1. Support locale-specific formatting for decimal values (e.g., 10212,456 in German).
1. Allow specifying decimal precision (e.g., 2 places → 123.45).
1. Provide validation-only mode for max/min, supporting both coercion and non-coercion on blur.

## 👩‍💻 Implementation

Followed an existing IxD spec as a template.

## 🧪 Testing

Spell check

## ✅ Checklist


packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Design
+
+### Configuration
+
+Proposed attributes:
+
+| Attribute               | Description                                                                                                               |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| coerce-invalid          | Boolean - Whether to coerce invalid max/min values on blur.                                                               |
```
jattasNI, line 95: I don't think we need an attribute for this.
 - `coerce-invalid=true` is the existing `nimble-number-field` behavior when `min` and `max` are set
 - `coerce-invalid=false` is the existing `nimble-number-field` behavior when `min` and `max` are not set
In the latter case an application can contain logic to set `error-text` and `error-visible` and disable form submission.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+locales.
+
+Additionally, clients have requested the ability to define max/min values for
+the `nimble-number-field` to support form validation without automatically
+coercing user-entered values to fit within the specified range.
+
+In comparison, the HTML `<input type="number">` natively supports locale-aware
+_decimal separators_ and does not coerce values to the min/max limit.
+
+This IxD proposes updates to the Number Field to natively support localization,
```
jattasNI, line 29: As we discussed, some of these may not belong in the number field if we're not ready to do the API design to try for maximum consistency with other component APIs and minimum churn as new formatting requirements arise. If we're waiting for the HLD to decide that then we should make this requirement (and maybe other places like the document title) more general, not specify that the API goes in the existing number field.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+
+![mouse interactions](./spec-images/localized-number-input-cursor-click.png)
+
+#### Keyboard Interactions
+
+| Key           | Description                                                                                                                                      |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `ENTER`/`ESC` | No action.                                                                                                                                       |
+| `UP ARROW`    | (If `hide-step = false`) Increment the number by the step value. (If `hide-step = true`) Moves the caret to the beginning of the entered number. |
+| `HOME`        | Moves the caret to the beginning of the entered number.                                                                                          |
```
jattasNI, line 117: Worth specifying ⌘-arrow should behave like HOME/END on macOS?

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Usage
+
+Number fields are used in three distinct ways:
+
+1. **With Increment/Decrement Buttons (Steppers)**
+
+    - Suitable for small, controlled ranges of values where users adjust numbers
+      incrementally (e.g., specifying a slot number).
+    - Discrete integers with a predictable range.
```
atmgrifter00, line 40: I'm not sure we need to be so opinionated as to assume using the step buttons only is useful for integers.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+
+Users can type values outside the maximum range of allowed values.
+
+![demonstrate the difference between coerce-invalid true and false](./spec-images/coerce-number-input.png)
+
+Users can type more or fewer fractional digits than required.
+
+![demonstrate the behavior of the maximum-fraction-digits and minimum-fraction-digits attributes](./spec-images/fraction-digits-number-input.png)
+
+## Future Considerations
```
jattasNI, line 139: While we're here do you want to spec out the desired interactions for mobile? For example, what keyboard should appear? Or at least list it as out of scope?

jattasNI: On the iPhone I tested, the number field currently shows the number keypad, which shows 10 digits and delete but no decimal key. That seems potentially problematic for the Decimal Values use case above.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+| `UP ARROW`    | (If `hide-step = false`) Increment the number by the step value. (If `hide-step = true`) Moves the caret to the beginning of the entered number. |
+| `HOME`        | Moves the caret to the beginning of the entered number.                                                                                          |
+| `DOWN ARROW`  | (If `hide-step = false`) Decrement the number by the step value. (If `hide-step = true`) Moves the caret to the end of the entered number.       |
+| `END`         | Moves the caret to the end of the entered number.                                                                                                |
+| _Numbers_     | (If focused) Adds number at cursor.                                                                                                              |
+| `SPACE`       | No action.                                                                                                                                       |
+| _Alphabetic_  | No action.                                                                                                                                       |
```
jattasNI, line 122: Some numeric inputs allow entering `e` and `E` and then will expand inputs like `1e3` to `1000`. The nimble input today allows you to enter e/E but then rejects the input on blur if it's present. Should we fix that bug as part of this work? Or add the feature to expand inputs?

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `ENTER`/`ESC` | No action.                                                                                                                                       |
+| `UP ARROW`    | (If `hide-step = false`) Increment the number by the step value. (If `hide-step = true`) Moves the caret to the beginning of the entered number. |
+| `HOME`        | Moves the caret to the beginning of the entered number.                                                                                          |
+| `DOWN ARROW`  | (If `hide-step = false`) Decrement the number by the step value. (If `hide-step = true`) Moves the caret to the end of the entered number.       |
+| `END`         | Moves the caret to the end of the entered number.                                                                                                |
+| _Numbers_     | (If focused) Adds number at cursor.                                                                                                              |
```
jattasNI, line 120: For completeness we should add `-` and `+`. If you want you can specify the behavior if the user tries to enter it in an invalid location or just leave that unspecified. The current behavior seems to allow them to be typed anywhere and then they're either truncated or the entire input is rejected on blur, depending on where the invalid ones are placed.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,146 @@
...
+### Behavior
+
+#### Mouse Interactions
+
+Hovering over the field input indicates interactivity.
+
+Clicking on the field input focuses the input and places a caret. If no value
+exists, the caret is placed at the far left. Otherwise, it is placed where
+clicked.
+
```
jattasNI, line 108: Double click to select all? We get this for free, but might be worth specifying to ensure it doesn't break.

packages/nimble-components/src/number-field/specs/IxD.md
atmgrifter00: So, the `nimble-number-field` currently handles entering `e`, but not `E`, and expands the number as expected. That is, if you type `1e3` that will automatically be changed to 1000 if you have `max` either not set, or set to at least 1000.

packages/nimble-components/src/number-field/specs/IxD.md
atmgrifter00: It seems for us to have a better mobile experience we should be using `inputmode="decimal"`. It's unclear to me why we are hard-coding it to `inputmode="numeric"`.

packages/nimble-components/src/number-field/specs/IxD.md
jattasNI: Huh, must be browser specific. So far I've tried macOS Firefox and Windows Firefox and Edge, and typing "1e3" and then click/tab outside results in the value going back to what it was on focus.

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Not only integers, but a small, predictable range of integers!

Today it's not possible to make clear to a user what the step value is, nor is it possible to contrain the step increments, so I'm confident more users fail than are helped by an increment/decrement accelerator on decimal numbers. E.g., 99.25 + increment = 100.25 (not helpful)

Regardless, the existing number-text-field handles the small integer case well, so I'm specifically excluding steppers from these requirements.

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Yeah, `inputmode="decimal"` works for positive decimals, but the keyboard [doesn't](https://inputmodes.com) have a minus sign, so can't support negative numbers.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,159 @@
...
+| `SPACE`                   | No action.                                                                                                                                       |
+| _Alphabetic_              | No action.                                                                                                                                       |
+| `e`/`E`                   | No action. This does not need to support scientific notation.                                                                                    |
+| `COMMA`                   | (As thousands separator) No action. (As decimal separator) Adds once at cursor.                                                                  |
+| `PERIOD`                  | (As thousands separator) No action. (As decimal separator) Adds once at cursor.                                                                  |
+| `-`                       | (If cursor at beginning of field) Add once at cursor.                                                                                            |
+
+Users can type valid numbers or the correct decimal separator for their locale
```
mollykreis, line 134: Can the user type multiple decimal separators into the field to enter an invalid state, or is the expectation that the separator key will be ignored if the character is already present in the field?

For example, in English, would someone be able to enter the number `120.456.152`?

packages/nimble-components/src/number-field/specs/spec-images/localized-number-input-cursor-click.png
mollykreis, line 1: Is the swapping from `,` to `.` on the second row correct? We're not switching separators based on whether or not the characters are highlighted, right?

packages/nimble-components/src/number-field/specs/IxD.md
jattasNI: I didn't find a great answer for this in my searching:
- [MDN docs for `inputmode="decimal"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inputmode#decimal) confirm that showing the minus sign is device-specific
- [the webkit bug requesting a minus sign](https://bugs.webkit.org/show_bug.cgi?id=193825) for the decimal keyboard has been open with no activity for a while
- [a whatwg bug about this issue](https://github.com/whatwg/html/issues/3478) has also been open for a while and mostly suggests doc/spec fixes

It seems like entering negative numbers isn't a common use case from the point of view of browser vendors or [other design systems like UK.gov](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).

If we think it is a common use case for our users then some options include:
1. we use a different affordance for negating (e.g. inc/dec buttons or a dedicated negate button)
2. we pick a keyboard that supports negating (e.g. `text` or `url`), with the downside of not showing the number pad by default

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Changed to - "Adds period at cursor. No action if decimal separator is already present."

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Updated to propose a `min` value specific option.

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Added - & + for completeness

packages/nimble-components/src/number-field/specs/spec-images/localized-number-input-cursor-click.png
fredvisser: Correct - fixed

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Proposing that we do not support scientific notation. Added row to that effect.

packages/nimble-components/src/number-field/specs/IxD.md
fredvisser: Sounds good - removed.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,160 @@
+# Number Field Updates (IxD)
```
rajsite, line 1: Cross posting the [concern from teams:](https://teams.microsoft.com/l/message/19:3924daf38ac14f20b8a1c09d3768468a@thread.tacv2/1745600839572?tenantId=eb06985d-06ca-4a17-81da-629ab99f6505&groupId=180bf0c7-4ff2-405e-8330-fdbe8ab6eb52&parentMessageId=1745531352610&teamName=TM-Design%20System&channelName=Dev%20Squad&createdTime=1745600839572)

Milan:
> My understanding of the state of it is that there is a proposed IxD (which maybe is trying to leak some implementation detail into it) that still has open threads (I haven't looked closely) but no HLD yet. My concern is that if we are advocating for all the proposed changes to the nimble-text-field (not just Intl separator awareness but also starting to think about restricted input / formatting / precision changes) I believe we need to think harder about that API and how it would grow longer-term to not be overly fit to the current use-case. If the proposal is a minor change to numeric text field (just Intl seperator awareness) and a spright control tailored to the use cases of the contributor for numeric formatting (choosing number numberer of visible decimal digits), we still need HLDs but we can be looser on the spright IxD since it isn't trying to define the general control.

fredvisser: >If the proposal is a minor change to numeric text field (just Intl seperator awareness) and a spright control tailored to the use cases of the contributor for numeric formatting (choosing number numberer of visible decimal digits), we still need HLDs but we can be looser on the spright IxD since it isn't trying to define the general control.

Yes, let's start with a minor change to the nimble-number-field to support locale aware decimal separators. Created a work item to create the HLD in AzDO.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,160 @@
...
+      data) and coerces values to enforce it.
+    - Does not use increment/decrement buttons.
+    - Supports locale-specific formatting (e.g., 1.23 vs. 1,23).
+
+## Requirements
+
+1. Support locale-specific formatting for decimal values (e.g., 10212,456 in
+   German).
+2. Allow specifying decimal precision (e.g., 2 places → 123.45).
+3. Provide validation-only mode for max/min, supporting both coercion and
```
jattasNI, line 66: For the non-coercion mode, are there any features missing from the existing numeric? A client can set no `min` or `max`, subscribe to the value change event, and put it in an error state if the value is out of range.

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,160 @@
...
+- [Bring number-column formatting options to nimble-number-field](https://github.com/ni/nimble/issues/1855)
+- [Number IxD (Figma)](https://www.figma.com/design/r2yGNQNVFdE7cBO9CyHmQx/Nimble---IxD?node-id=1603-3464)
+
+The `nimble-number-field` leverages
+[fast-foundation](https://github.com/ni/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/number-field/number-field.ts)
+for its behavior, so we've inherited much of the behavior, including max/min
+validation and localization.
+
+For supported languages (English, French, German, Japanese, Chinese), the
+_decimal separator_ does not align with user expectations in German. For
```
m-akinc, line 16: > the _decimal separator_ does not align with user expectations in German.

French, too, right? (I believe Japanese and Chinese use a dot (`.`) like English.)

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,160 @@
...
+Hovering over the field input indicates interactivity.
+
+Clicking on the field input focuses the input and places a caret. If no value
+exists, the caret is placed at the far left. Otherwise, it is placed where
+clicked.
+
+Double-clicking within the field input selects the closest contiguous
+characters.
+
+Triple-clicking within the field input selects all characters.
```
m-akinc, line 113: It took me a while to figure out how you can have characters that aren't contiguous. Might be worth explicitly pointing out that `+`s, `-`s, and decimal separators (`.`/`,`) break contiguity. (`e`s are contiguous with digits).

packages/nimble-components/src/number-field/specs/IxD.md
```diff
@@ -0,0 +1,160 @@
...
+
+![mouse interactions](./spec-images/localized-number-input-cursor-click.png)
+
+#### Keyboard Interactions
+
+| Key                       | Description                                                                                                                                      |
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `ENTER`/`ESC`             | No action.                                                                                                                                       |
+| `UP ARROW`                | (If `hide-step = false`) Increment the number by the step value. (If `hide-step = true`) Moves the caret to the beginning of the entered number. |
```
m-akinc, line 122: This would be a change in behavior (same for down arrow). Currently, even if you `hide-step`, up/down increment/decrement the value.

rajsite:
Moving to draft because of inactivity

PR #2596, opened 2025-05-01
Text field appearance read only

mollykreis:
# Pull Request

## 🤨 Rationale

Add `appearance-readonly` attribute to the `nimble-text-field` to style the component as readonly when it is disabled.

## 👩‍💻 Implementation

- Add `appearance-readonly` attribute (tied to `appearanceReadOnly` property) on the text field
- Updated styles to make `disabled + appearanceReadonly` look like the `readonly` styles
- Updated storybook docs
- Updated matrix tests

## 🧪 Testing

- Manually tested
- Chromatic test

## ✅ Checklist


packages/nimble-components/src/text-field/styles.ts
```diff
@@ -24,6 +24,7 @@ import { styles as errorStyles } from '../patterns/error/styles';
 import { styles as requiredVisibleStyles } from '../patterns/required-visible/styles';
 import { userSelectNone } from '../utilities/style/user-select';
 
+// prettier-ignore
```
rajsite, line 27: Could we leverage CSS layers to avoid the reliance on `:not` selectors?

Like how the button has layers, maybe we can add a layer after `disabled` for the `appearsReadonly` state or something:
https://github.com/ni/nimble/blob/6cfd185413ab6e327b55d296813c88c639d545d8/packages/nimble-components/src/patterns/button/styles.ts#L29

That with the normal thought around don't try and share selectors for multiple states, share custom properties across multiple selectors.
Our [cascade guidance](https://github.com/ni/nimble/blob/6cfd185413ab6e327b55d296813c88c639d545d8/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes) may also need to be updated (after the pattern is settled) to put `[disabled][appearance-readonly]` after `[disabled]`

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -92,6 +94,7 @@ export default metadata;
 const component = (
     [_readOnlyName, readonly]: ReadOnlyState,
     [_disabledName, disabled]: DisabledState,
+    [_appearanceReadOnlyName, appearanceReadonly]: AppearanceReadOnlyState,
```
rajsite, line 97: (open ended / brainstorming) Maybe we should merge the readonly, disabled, and appearanceReadonly states?
const editableStates =  [normal, readonly, disabled, appearanceReadonly]

names for the set of states could be: editableStates / interactionStates (we already use for hover and stuff so maybe not) / disabledReadonlyStates (very explicit)

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -111,6 +114,7 @@ const component = (
         ?readonly="${() => readonly}"
         ?error-visible="${() => errorVisible}"
         error-text="${() => errorText}"
+        ?appearance-readonly="${() => appearanceReadonly}"
```
rajsite, line 117: It's a pain but lets only ever use `readonly` for the attribute and consistently use readOnly for JS properties

```suggestion
        ?appearance-readonly="${() => appearanceReadOnly}"
```

packages/nimble-components/src/text-field/styles.ts
rajsite: Possible the higher specificity of `[disabled][appearance-readonly]` over `[disabled]` means wee won't even need css layers added. Just might mean introducing more design tokens / private css custom properties for the readonly properties to share across the `[readonly]` and `[disabled][appearance-readonly]` selectors in cascade ordering

mollykreis: Sure, I can try out some different CSS approaches. I thought I'd start with `:not` since it was the smallest change and see if we thought it was an acceptable use of `:not` before I made a larger change.

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
mollykreis: The downside I see with that approach is that we lose our test coverage for things like:
- `disabled` and `readOnly` - making sure `disabled` styling wins over `readOnly` styling
- not `disabled` and `appearanceReadOnly` - making sure to `appearanceReadOnly` causes no styling changes

rajsite: yea that's true, am alternate thought is add all those state permutations to collapse in a single dimension just so it's easier to split / add new dimensions in the future (particularly for text field since it has so many permutations / story splits). But I don't feel strongly there

packages/nimble-components/src/text-field/styles.ts
```diff
@@ -165,6 +179,10 @@ export const styles = css`
         color: ${bodyDisabledFontColor};
     }
 
+    :host([appearance-readonly]) .control[disabled]::placeholder {
+        color: ${controlLabelFontColor};
+    }
```
rajsite, line 184: These aren't intentionally targetting the control vs host right? For consistency can we change this and the selector above to key off state from the host instead of the `.control`?

```css
    :host([disabled]) .control::placeholder {
    }

    :host([disabled][appearance-readonly]) .control::placeholder {
        color: ${controlLabelFontColor};
    }


```

packages/nimble-components/src/text-field/styles.ts
rajsite: I like how it turned out! Thanks for taking a look at it 👍

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -506,8 +807,21 @@ export const colorThemeReadOnlyDisabledWithoutButtons: StoryFn = createFixedThem
...
+        appearanceStates,
+        fullBleedStates,
+        valueStates
+    ]),
+    colorThemeDarkGreenBackground
+);
+
+export const colorThemeReadOnlyDisabledAppearanceReadOnlyWithButtons: StoryFn = createFixedThemeStory(
+    createMatrix(component, [
+        [disabledReadOnlyStates[7]],
```
rajsite, line 824: If we think we'll need to break out this dimension again maybe we should go ahead and make some const aliases for states?

```ts
const disabledReadOnlyState = {
    none: disabledReadOnlyStates[0],
    appearanceReadOnly: disabledReadOnlyStates[1],
    //...
    readOnlyDisabledAppearanceReadOnly: disabledReadOnlyStates[7],
} as const;
```

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
rajsite: That turned out really nice for the storybook splits. We could maybe even remerge the button dimension but that's probably not even worth the effort to attempt. Looks good 👍

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
mollykreis: I like that suggestion. I'm not 100% sure if this was your suggestion or not, but I created those aliases in `states.ts` so we can easily use them from our other component matrix tests. I think that'll also make it really easy to use them from components that don't support the full matrix (e.g. select & combobox that don't have `readonly`).

packages/storybook/src/utilities/storybook.ts
```diff
@@ -184,6 +184,9 @@ export const disabledDescription = (options: {
 export const readonlyDescription = (options: {
     componentName: string
 }): string => `Styles the ${options.componentName} as readonly and prevents the user from changing the value.`;
+export const appearanceReadOnlyDescription = (options: {
```
jattasNI, line 187: Thoughts on the best way to convey why a client would use this setting? I suspect that someone coming to these docs without having read #2592 might be confused/curious about when to use `readonly` vs `disabled` and `appearance-readonly`. Some possibilities:

1. Add some more usage guidance in this description string. e.g. "This is useful for applications that use a forms library that sets `disabled` on components but which don't want those components to have a disabled appearance"
2. Add some shared usage guidance that provides more context / examples about how to use this. This could either be a few sentences that get embedded in every relevant story under the Usage section or it could be a separate page. Either way we'd link to it from this description.

A related question is whether we should mention this attribute or link to this usage guidance from the docs for similar attributes (`disabled` and `readonly`). I'm leaning not: seems like it'd just be noisy and not really help people discover the feature any better than them scrolling down the docs page and seeing this attribute. But I'm open to other opinions.

jattasNI (commented):
Some feedback from playing with storybook. Still need to look at the code and the matrix tests.

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -24,7 +24,9 @@ import {
     errorStates,
```
jattasNI, line 24: This might be a lost cause but I'm struggling to parse the names of these stories. I think my challenge is that there aren't any delimiters between the states, and now that some of the states include identical words I can't tell what's what. For example:
"Light Theme Read Only Disabled Appearance Read Only Without Buttons"

Maybe we should start putting commas or dashes between the states?
"Light Theme, Read Only, Disabled Appearance Read Only, Without Buttons"

Or turn each state into one hyphenated word?
"Light-Theme Read-Only Disabled-Appearance-Read-Only Without-Buttons"

Or switch to a more code-like syntax?
"theme=light readonly=true disabled=true appearance-readonly=true buttons=false"

Feel free to tackle in a follow up to reduce churn or to argue that this is a pre-existing problem not worth solving now.

jattasNI (commented):
Finished review. Looking good, just want to see where you land with a couple of my comment threads.

change/@ni-nimble-components-a4d0f3b2-8a61-4e8c-b81a-0f10c03c185b.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
mollykreis: From my testing (on Windows only), the behavior in Edge, Chrome, and FF of the `nimble-text-field` matches that of the native `input`. Therefore, my inclination is not to do anything to be "smarter" than the browser or deviate from native behavior.

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
mollykreis: I don't have a strong opinion for or against giving these stories more readable names using the `storyName` property. I agree that the names are hard to parse, but I'm also not sure it's extremely important since these are in the "Tests" section, which isn't part of our published storybook.

That being said, I'd be happy to give them names since it isn't hard to do. Let's agree on the convention first, though, so I don't have to make multiple passes. I'd vote for either of the following, with a slight preference to comma-delimited:

comma-delimited names: _Light Theme, Read Only, Disabled Appearance Read Only, Without Buttons_
or
hyphenated names: _Light-Theme Read-Only Disabled-Appearance-Read-Only Without-Buttons_

@rajsite - do you have any opinions on this topic?

packages/storybook/src/utilities/storybook.ts
mollykreis: I went with your first suggestion of updating the in-line docs for the attribute. I think that's probably sufficient for now, and we can add/update docs in the future if we decide it isn't.

change/@ni-nimble-components-a4d0f3b2-8a61-4e8c-b81a-0f10c03c185b.json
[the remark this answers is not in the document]
jattasNI: Good call, I see the same on macOS FF with the native input so I'm ok to leave as-is.

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
jattasNI: My use case for wanting a parseable name was when inspecting the new stories to figure out if they were correctly rendering the states they were supposed to. I agree that's a much less common workflow than clients reading our public-facing docs, but I think either of these options would be a nice developer experience improvement for me.

rajsite: I can't think of a way to do this that doesn't make a lot of duplicated crazy names:
```ts
export const lightThemeEditableDisabledWithoutButtons ...
lightThemeEditableDisabledWithoutButtons.storyName = 'Light Theme, ...'
```
seems like maintaining it would be a big pain, i.e. we need to add more dimensions. I'd prefer to not go that route even if reading the names is a bit easier.

Because these are just story tests, what about a hackier and lighter weight touch:
```
export const lightTheme$EditableEnabled$WithButtons: 
```
image

i.e. `$` iis a valid character in JS identifiers and looks like storybook doesn't filter it out. Could also use emoji in identifer if we find a good one.

jattasNI (approved):

packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
mollykreis: I couldn't find any emoji that would work as part of a JS identifier, so perhaps `$` is the best option.

@jattasNI - what do you think about the `$` approach?

jattasNI: I hadn't thought of the maintenance issue that Milan pointed out. I agree it's not worth doing if we have to duplicate like that.

I can't say I love the aesthetics of the `$` approach but it solves the problem. So I think it's an improvement over where we are today if Molly doesn't mind the busywork of updating our stories to use that.

One last low-value thing to check is how Storybook renders `lightTheme$editableEnabled` vs `lightTheme$EditableEnabled`. If the former gets rid of the space in the rendered name ("Light Theme$Editable Enabled") then I think that would look a tiny bit better (but don't care much).

mollykreis: Using lower case does get rid of the space, but at least in my opinion, it looks quite a bit worse without a space.

Here's a screenshot of what it looks like. The top story uses capital (`lightTheme$EditableEnabled$WithoutButtons`) and the lower story uses lower case (`lightTheme$editableEnabled$appearanceReadOnly$withoutButtons`).
image

mollykreis: I'll make a pass in the coming days to update our matrix stories to use this naming convention.

PR #2597, opened 2025-05-05
Align positioning behavior of rich-text mention listbox, combobox, select, and menu button

### LINKED ISSUE #2246: Mention listbox doesn't open on right position based on the view available
# 🐛 Bug Report

The mention listbox in nimbl-rich-text-editor does not open in the right position where the view is visible. For example, if there is no space at the bottom of the webpage, still the popup opens below.

## 💻 Repro or Code Sample

1. In the storybook, position the rich text editor in such a way that a part of the editor is not visible
2. Add `@` character
3. The popup opens

## 🤔 Expected Behavior

Based on the available view, the popup should open there, similar to how the combobox or select works. For example, if the bottom view is hidden, the popup should open at the top.

## 😯 Current Behavior

In the below case, it is expected to open the popup at the top as the bottom view is hidden. 
mention-listbox-issue

## 💁 Possible Solution

Maybe, similar to the implementation of combobox or [select](https://github.com/ni/nimble/blob/bc79e9bc7c3f21da124cc3aebd493bfba5c01e4b/packages/nimble-components/src/select/template.ts#L105) in the anchored region template.

## 🌍 Your Environment

* Version [e.g. 1.8.0]

m-akinc:
# Pull Request

## 🤨 Rationale

The mention listbox only opens down, this change allows the mention listbox to open based on available viewport space. Fixes #2246

## 👩‍💻 Implementation

Of the controls that implement anchored region, the select and combobox are most aligned to the rich text mention listbox as they are all Listbox implementations with sized listboxes
- menu button uses a [simple pattern for positioning](https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/nimble-components/src/menu-button/template.ts#L40-L41) that is unaware of popover sizing
- tooltip has [more detailed positioning](https://github.com/ni/fast/blob/archives/fast-element-1/packages/web-components/fast-foundation/src/tooltip/tooltip.ts#L446-L511) that is unaware of popover sizing

This PR updates the richtext mention listbox to align with the select and combobox sizing behavior but does not bring forward the configurable positioning behavior.
 - rich text mention listbox will use a fixed `dynamic` positioning because it does not have a [user configurable positioning](https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/nimble-components/src/select/template.ts#L114-L115) like select and combobox.

PR also includes:
- minor update to align on the shared `DropdownPosition` enum consistently
- minimum dropdown height of 1.5 options (for dropdowns that are confined to the window height)

## 🧪 Testing

- Manual testing in Storybook
- New visual tests for default positioning behavior of combobox, select, and mention listbox

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-f4992b1c-b7f0-46e1-a684-b5afff8683f6.json
+++ b/change/@ni-nimble-components-f4992b1c-b7f0-46e1-a684-b5afff8683f6.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Align positioning behavior of rich-text mention listbox, combobox, select, and menu button",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/rich-text/mention-listbox/index.ts
+++ b/packages/nimble-components/src/rich-text/mention-listbox/index.ts
@@ -36,8 +36,6 @@ export class RichTextMentionListbox extends FoundationListbox {
     public region?: AnchoredRegion;
 
     /**
-     * The space available in the viewport for the listbox when opened.
-     *
      * @internal
      */
     @observable
@@ -108,10 +106,7 @@ export class RichTextMentionListbox extends FoundationListbox {
      * @public
      */
     public show(options: MentionListboxShowOptions): void {
-        const listboxTop = options.anchorNode.getBoundingClientRect().bottom;
-        this.availableViewportHeight = Math.trunc(
-            window.innerHeight - listboxTop
-        );
+        this.updateAvailableViewportHeight();
         this.filter = options.filter;
         this.anchorElement = options.anchorNode;
         this.setOpen(true);
@@ -286,6 +281,19 @@ export class RichTextMentionListbox extends FoundationListbox {
 
     private setOpen(value: boolean): void {
         this.open = value;
+    }
+
+    private updateAvailableViewportHeight(): void {
+        const currentBox = this.getBoundingClientRect();
+        const viewportHeight = document.documentElement.getBoundingClientRect().height;
+        const availableSpaceAbove = Math.trunc(currentBox.top);
+        const availableSpaceBelow = Math.trunc(
+            viewportHeight - currentBox.bottom
+        );
+        this.availableViewportHeight = Math.max(
+            availableSpaceAbove,
+            availableSpaceBelow
+        );
     }
 }
 
--- a/packages/storybook/src/nimble/combobox/combobox.stories.ts
+++ b/packages/storybook/src/nimble/combobox/combobox.stories.ts
@@ -141,8 +141,14 @@ const metadata: Meta<ComboboxArgs> = {
         },
         dropDownPosition: {
             name: 'position',
-            options: [DropdownPosition.above, DropdownPosition.below],
-            control: { type: 'select' },
+            options: [undefined, ...Object.values(DropdownPosition)],
+            type: 'string',
+            control: {
+                type: 'radio',
+                labels: {
+                    undefined: 'default'
+                }
+            },
             description: dropdownPositionDescription({
                 componentName: 'combobox'
             }),
@@ -221,7 +227,7 @@ const metadata: Meta<ComboboxArgs> = {
     args: {
         label: 'Combobox',
         disabled: false,
-        dropDownPosition: 'below',
+        dropDownPosition: undefined,
         autocomplete: ComboboxAutocomplete.both,
         errorVisible: false,
         errorText: 'Value is invalid',
--- a/packages/storybook/src/nimble/select/select.stories.ts
+++ b/packages/storybook/src/nimble/select/select.stories.ts
@@ -5,7 +5,10 @@ import { listOptionTag } from '@ni/nimble-components/dist/esm/list-option';
 import { listOptionGroupTag } from '@ni/nimble-components/dist/esm/list-option-group';
 import { selectTag } from '@ni/nimble-components/dist/esm/select';
 import { FilterMode } from '@ni/nimble-components/dist/esm/select/types';
-import { DropdownAppearance } from '@ni/nimble-components/dist/esm/patterns/dropdown/types';
+import {
+    DropdownAppearance,
+    DropdownPosition
+} from '@ni/nimble-components/dist/esm/patterns/dropdown/types';
 
 import {
     apiCategory,
@@ -190,8 +193,14 @@ const metadata: Meta<SelectArgs> = {
         },
         dropDownPosition: {
             name: 'position',
-            options: ['above', 'below'],
-            control: { type: 'select' },
+            options: [undefined, ...Object.values(DropdownPosition)],
+            type: 'string',
+            control: {
+                type: 'radio',
+                labels: {
+                    undefined: 'default'
+                }
+            },
             description: dropdownPositionDescription({
                 componentName: 'select'
             }),
@@ -295,7 +304,7 @@ const metadata: Meta<SelectArgs> = {
         errorVisible: false,
         errorText: 'Value is invalid',
         filterMode: 'none',
-        dropDownPosition: 'below',
+        dropDownPosition: undefined,
         appearance: DropdownAppearance.underline,
         optionsType: ExampleOptionsType.simpleOptions,
         clearable: false,
--- a/packages/nimble-components/src/anchored-region/index.ts
+++ b/packages/nimble-components/src/anchored-region/index.ts
@@ -1,7 +1,9 @@
+import { observable } from '@ni/fast-element';
 import {
     DesignSystem,
     AnchoredRegion as FoundationAnchoredRegion,
-    anchoredRegionTemplate as template
+    anchoredRegionTemplate as template,
+    type AnchoredRegionPositionLabel
 } from '@ni/fast-foundation';
 import { styles } from './styles';
 
@@ -22,7 +24,11 @@ declare global {
 /**
  * A nimble-styled anchored region control.
  */
-export class AnchoredRegion extends FoundationAnchoredRegion {}
+export class AnchoredRegion extends FoundationAnchoredRegion {
+    /* @internal */
+    @observable
+    public override verticalPosition: AnchoredRegionPositionLabel | undefined;
+}
 
 const nimbleAnchoredRegion = AnchoredRegion.compose({
     baseName: 'anchored-region',
--- a/packages/nimble-components/src/combobox/index.ts
+++ b/packages/nimble-components/src/combobox/index.ts
@@ -1,9 +1,16 @@
-import { DOM, Observable, attr, html, observable, ref } from '@ni/fast-element';
+import {
+    attr,
+    DOM,
+    html,
+    observable,
+    Observable,
+    ref,
+    type Notifier
+} from '@ni/fast-element';
 import {
     DesignSystem,
     type ComboboxOptions,
     ComboboxAutocomplete,
-    SelectPosition,
     ListboxOption,
     DelegatesARIACombobox,
     applyMixins,
@@ -28,8 +35,10 @@ import { styles } from './styles';
 import { mixinErrorPattern } from '../patterns/error/types';
 import {
     DropdownAppearance,
+    DropdownPosition,
     type DropdownPattern
 } from '../patterns/dropdown/types';
+import { anchoredRegionPositionToDropdownPosition } from '../patterns/dropdown/utility';
 import type { AnchoredRegion } from '../anchored-region';
 import { template } from './template';
 import { FormAssociatedCombobox } from './models/combobox-form-associated';
@@ -69,7 +78,7 @@ export class Combobox
      * The placement for the listbox when the combobox is open.
      */
     @attr({ attribute: 'position' })
-    public positionAttribute?: SelectPosition;
+    public positionAttribute?: DropdownPosition;
 
     /**
      * The open attribute.
@@ -92,7 +101,7 @@ export class Combobox
      * @public
      */
     @observable
-    public position?: SelectPosition;
+    public position?: DropdownPosition;
 
     /**
      * @internal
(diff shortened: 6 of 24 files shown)


packages/nimble-components/src/rich-text/mention-listbox/index.ts
```diff
@@ -108,9 +108,11 @@ export class RichTextMentionListbox extends FoundationListbox {
      * @public
      */
     public show(options: MentionListboxShowOptions): void {
-        const listboxTop = options.anchorNode.getBoundingClientRect().bottom;
+        const anchorRect = options.anchorNode.getBoundingClientRect();
+        const availableSpaceBelow = window.innerHeight - anchorRect.bottom;
+        const availableSpaceAbove = anchorRect.top;
         this.availableViewportHeight = Math.trunc(
-            window.innerHeight - listboxTop
+            Math.max(availableSpaceAbove, availableSpaceBelow)
```
rajsite, line 115: Talked offline but biggest concern is making sure discussion is captured about how these changes here align with out other listbox / anchored region implementations. Want alignment to increase generally or have clear reasons why they are different

m-akinc: Updated the PR description. Essentially, we're making the mention listbox more consistent with the other dropdown controls with this change.

rajsite: That description is not conducive to quick review 😅. Could you break it into bullets for the controls referenced and then link to relevant source. Goal is to be able to quickly see that this is aligning or misaligned with how our other controls behave.

m-akinc: What, you don't trust me? 😅 Fine, fine.... Reorganized description and added all the relevant source links.

packages/nimble-components/src/rich-text/mention-listbox/index.ts
```diff
@@ -287,6 +285,25 @@ export class RichTextMentionListbox extends FoundationListbox {
     private setOpen(value: boolean): void {
         this.open = value;
     }
+
+    // Aligns with select / combobox
+    // Modified to remove forced position concept
+    private setPositioning(options: MentionListboxShowOptions): void {
```
rajsite, line 291: Aligned to select setPositioning implementation without the forcedPositioning conceept

 https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/nimble-components/src/select/index.ts#L1064-L1089

change/@ni-nimble-components-f4992b1c-b7f0-46e1-a684-b5afff8683f6.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: In testing in storybook, you can see on the canvas pages that in both the original proposed change and in the updates I pushed the popover opens upward surprisingly when there is plenty of vertical space below.

Couldn't find a reference quickly but I'm pretty sure that's why [select](https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/storybook/src/nimble/select/select.stories.ts#L279) and [combobox](https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/storybook/src/nimble/combobox/combobox.stories.ts#L205) are configured to `below` by default.

- There is a [storybook test coverage](https://github.com/ni/nimble/blob/aa21afb2ee252fa3333de6d0b209a7f54634d66b/packages/storybook/src/nimble/rich-text/mention-listbox/mention-listbox-matrix.stories.ts#L30) for listbox mention being shown, my concern is if it's possible that not setting the position explicitly could make those unstable.
- I'm not super concerned about the storybook canvas behavior personally, but it could be misleading about the expected behavior.

image

rajsite: @jattasNI @m-akinc thoughts on whether we want to make the position popovers open user configurable for testing purposes like our other popover controls for storybook consistency and potentially also snapshot consistency? Or accept the storybook behavior and hope / find a way to convince ourselves the snapshots will be reliable.

m-akinc: I'm looking into this now. I had noticed this before, but mistakenly attributed it to the fact that the root `html` element of the iframe was only slightly larger than the control. The actual problem is that the anchored region is using the bounds of the rich-text editor (rather than the window/viewport) to determine whether to open above or below. See how changing the scroll offset by just a pixel or two makes the difference (see gap between blue line and `@` on the right):
image

rajsite: When I was playing with it it seemed to be the bounds of the `html` element, which in the canvas view happens to be sized pretty much to the size of the containing element.

jattasNI (commented):
I've been holding off reviewing this since it looks like there are active discussion threads which I don't have much of an opinion on. Please let me know if you're expecting my input on any of them, otherwise I'll take a look once they're resolved and my vote's reset.

change/@ni-nimble-components-f4992b1c-b7f0-46e1-a684-b5afff8683f6.json
m-akinc: Sorry, I wasn't clear. The anchor node is the `@` symbol, but it seems like it's using the height of the rich-text editor and the location of the `@` symbol within it to determine whether there is more room above or below. But I'll double check that that's the case vs it using the height of the `html` element.

m-akinc: Ugh, this is so stupid. I figured out why the mention listbox was behaving differently from the select/combobox. Essentially, the select/combobox _never_ actually use the dynamic positioning mode, and [we end up doing our own determination based on `window.innerHeight`](https://github.com/ni/nimble/blob/5837f010b916ca93aa634c23f26eba5d35144c8c/packages/nimble-components/src/select/index.ts#L1071). `setPositioning` figures out whether we want the listbox above or below, and then [we end up setting `position` and `positionAttribute`](https://github.com/ni/nimble/blob/5837f010b916ca93aa634c23f26eba5d35144c8c/packages/nimble-components/src/select/index.ts#L1084). If `positionAttribute` is set (which it always is, _because we set it_), [we use `locktodefault` vertical positioning and we set the default to the corresponding position](https://github.com/ni/nimble/blob/5837f010b916ca93aa634c23f26eba5d35144c8c/packages/nimble-components/src/select/template.ts#L114). I kind of hate this, but it lets us base the position on `window.innerHeight` as opposed to the rect of the "viewport element", which [FAST defaults to `document.documentElement`](https://github.com/microsoft/fast/blob/fd9068b94e4aa8d2282f0cce613f58436fae955d/packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts#L608) (i.e. the `html` element).

I've updated the mention listbox's template and added an internal `position` property, so that we match what we're doing in select/combobox. 🤦‍♂️🙄

jattasNI, line 1: The [rich text theme matrix story snapshots](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=681e99069066451a7d917d88) look incorrect. The Dark theme story is rendering overlapping with the color theme.

image

jattasNI, line 1: Could you comment why you went with only manual testing rather than adding auto test coverage for the new behavior? If there are technical challenges or precedent from similar components we should capture that in the PR description. Even better would be to add test coverage, for example via Chromatic snapshots, but if we're limited by the actual browser's size I could see that being challenging.

m-akinc: Didn't notice this locally, because it depends on the window size. 😑 The behavior now is that it opens to whichever side of the anchor has more space. This is consistent with the default behavior of the select and combobox, but we haven't run into this problem in their visual tests because all of their "Open" tests set an explicit attribute value for `position` (i.e. "above" or "below"). The mention listbox doesn't have a `position` property to explicitly force it to open above or below.

This issue exposes a gap in our testing coverage for the combobox and select. I did not find any tests--functional or visual--for them that validate the default behavior (i.e. when no explicit `position` is set). The menu button at least has unit tests that validate the expected configuration of the anchored region.

Moreover, the default behavior of the combobox, select, and now mention listbox is inconsistent with the default behavior (`position = "auto"`) of the menu button. The menu button defaults to configuring the anchored region to use `"dynamic"` `vertical-positioning-mode` with a `vertical-default-position` of `"bottom"`. What this means is that it will open below, as long as there is enough space, otherwise it opens above. Combobox/select/mention listbox, on the other hand, default to opening to whichever side has more space.

I propose to:
- change the default behavior of the combobox, select, and mention listbox to match the menu button (i.e. only open above if there isn't enough room to fit the full listbox below)
- add visual test cases for the combobox and select for the default (unspecified) `position` case

If you feel we should leave the existing behavior as it is, I can instead add a `position` property to the mention listbox that we only set (to "below") in our visual test.

m-akinc: I don't really have a good excuse. I think I felt like we already have pretty scattershot coverage of our components' dynamic behaviors, so I wasn't that compelled to go beyond manual testing. But we should, as there's no significant technical barrier that I'm aware of. Based on your response to my comment (about testing and behavior) in the other thread, I'll add the appropriate test cases.

jattasNI: I think your proposal matches my intuition of what should happen.

I'm debating whether we need any more input / buy-in / testing to validate it as the right direction. This seems like a detail that should be low risk to change, but I could imagine follow-on effects like poorly written tests that fail because things move around or specific corner cases that behave differently than people expect.

I'm landing on a couple risk mitigation ideas:
1. run it by Fred 
2. validate the behavior change in a couple SLE apps and make sure it feels ok. For example, a couple slide outs with select/combobox near the bottom of the screen. Or some comment threads with the comment editor at various positions on screen

rajsite: Seems like we should create a tech debt issue to cover two aspects of this:
- The setPositioning logic in select, combobox, and mention listbox do not make it clear that they always override the vertical-positioning-mode of their owned anchored regions. Maybe instead that could be updated in this PR and all three could be updated to `locktodefault` since as you describe the `dynamic` mode effectively is never used
- The other anchored region based elements (menu-button and tooltip) are not aligned in implementation in that they calculate viewport based on the anchored region implementation targetting `html` and not the override using `window.innerHeight`

rajsite: I think the PR is primarily waiting on this thread? @m-akinc are we actively trying to resolve this? If not we can move the PR to draft

m-akinc: I've run it by Fred, who was on board with my proposal. Aligning combobox/select/mention-listbox with menu button turns out to be more than a trivial change, and I've just been moving this forward during my "spare cycles." I currently have the implementations updated and am just adding visual test cases, but it might still be a while, so I'll move the PR to draft for now.

m-akinc: Based on the proposal (which seems to have support) in the other thread, I'm going the other way and aligning on using the anchored region's `dynamic` positioning mode for all the controls.

My only pending concern is whether the `dynamic` mode's use of the `html` element's height rather than `window.innerHeight` will have any negative consequences (other than in Storybook, which I can deal with). My expectation is that it will not, i.e. that in "real world" use cases the `html` element fills the viewport. If we're not comfortable with that assumption, we could consider alternatives (one of which being to change the anchored region in `@ni/fast-foundation` to use `window.innerHeight`).

m-akinc: Finally got around to pushing this forward. Moving out of draft.

packages/storybook/src/nimble/combobox/combobox.stories.ts
```diff
@@ -142,8 +142,14 @@ const metadata: Meta<ComboboxArgs> = {
         },
         dropDownPosition: {
             name: 'position',
-            options: [DropdownPosition.above, DropdownPosition.below],
-            control: { type: 'select' },
+            options: [undefined, ...Object.values(DropdownPosition)],
```
m-akinc, line 145: We weren't providing a way to demo the default (unset) for `dropDownPosition`.

packages/storybook/src/nimble/select/select.stories.ts
```diff
@@ -191,8 +194,14 @@ const metadata: Meta<SelectArgs> = {
         },
         dropDownPosition: {
             name: 'position',
-            options: ['above', 'below'],
-            control: { type: 'select' },
+            options: [undefined, ...Object.values(DropdownPosition)],
```
m-akinc, line 197: Like with the combobox stories, we weren't providing a way to demo the default (unset) `dropDownPosition` behavior.

change/@ni-nimble-components-f4992b1c-b7f0-46e1-a684-b5afff8683f6.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "patch",
```
rajsite, line 2: Putting a comment as a placeholder as still needs review. I think we should hold off merging this until after the release branch for 2025-07 is created.

m-akinc:
@jattasNI @rajsite Obviously not urgent, since this hasn't been active for several months, but I don't think there's anything blocking this. Review when convenient.

rajsite:
closing for now pending other desired refactors, see [azdo discussion](https://dev.azure.com/ni/DevCentral/_workitems/edit/3951459#11437144)

PR #2598, opened 2025-05-07
Rename matrix tests

mollykreis:
# Pull Request

## 🤨 Rationale

This is a follow up change from #2596 to rename the matrix tests to use `$` as a delimiter when the name includes which dimension a test is using.

## 👩‍💻 Implementation

Rename matrix tests using the following guidelines:
- Don't include the name of the component in the test name
- If the theme is part of the name of the test, put it first without the background color (e.g. without `WhiteBackground`)
- Don't include theme in the name if it isn't part of the test dimension. For example, if a test is run only for one theme to test something unrelated to theme, it doesn't need to be included in the name.
- Use `$` as a delimiter in the test name when the name includes multiple constant dimension values

In addition, I split the test for a hidden radio item into its own test file under `Tests/Radio`. It had been a story under `Tests/Radio Group`.

## 🧪 Testing

Tested in storybook locally

## ✅ Checklist

### DIFF
--- a/packages/storybook/CONTRIBUTING.md
+++ b/packages/storybook/CONTRIBUTING.md
@@ -104,6 +104,32 @@ All other Markdown formatting is supported. See any
 [Markdown Cheatsheet](https://www.markdownguide.org/cheat-sheet/) for more
 information.
 
+### Matrix story naming
+
+The names of stories within matrix tests should conform to the following guidelines:
+
+1. Do not include component name in the story name.
+    - Do: `Hidden`, `Theme Matrix`
+    - Don't: `Hidden Text Field`, `Select Theme Matrix`
+1. Do not include tests for multiple components under one storybook group.
+    - Do: `Radio Group >> Hidden`, `Radio >> Hidden`
+    - Don't: `Radio Group >> Hidden Radio Group` and `Radio Group >> Hidden Radio`
+1. Use `$` as a delimiter in test names that have a constant value for a matrix dimension.
+    - Do: `LightTheme$ Open$ No Filter`
+    - Don't: `Light Theme Open No Filter`
+1. Consider including a "false" attribute in a story name when there is a story for the value being both "false" and "true". This name should be in the form of "BooleanPropertyName" and "BooleanPropertyNameAbsent".
+    - Do: `Read Only Absent$ Disabled`, `Read Only$ Disabled Absent`
+    - Don't: `Editable$ Disabled`, `ReadOnly$ Enabled`
+1. Specify the theme as the first segment of the story name when it is a fixed value.
+    - Do: `Light Theme$ Open $No Filter`
+    - Don't: `Open$ No Filter$ Light Theme`
+1. Do not include the theme in the story name if it isn't part of the test dimension. If a test is run only for a single theme to test something unrelated to theme, the theme should not be included in the story name.
+    - Do: `Light Theme$ Open$ No Filter` (if there are also  `Dark Theme$ Open$ No Filter` and  `Color Theme$ Open$ No Filter`)
+    - Don't: `Light Theme$ Open$ No Filter` (if there are not also  `Dark Theme$ Open$ No Filter` and  `Color Theme$ Open$ No Filter`)
+1. Do not include the background color in the story name.
+    - Do: `Light Theme$ Open$ No Filter`
+    - Don't: `Light Theme White Background$ Open$ No Filter`, `Light Theme$ White Background$ Open$ No Filter`
+
 ### Testing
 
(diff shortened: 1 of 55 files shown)


packages/storybook/src/nimble/text-field/text-field-matrix.stories.ts
```diff
@@ -161,7 +161,7 @@ const requiredVisibleStatesComponent = (
     </${textFieldTag}>
 `;
 
-export const lightThemeEditableEnabledWithoutButtons: StoryFn = createFixedThemeStory(
+export const lightTheme$Editable$Enabled$WithoutButtons: StoryFn = createFixedThemeStory(
```
rajsite, line 164: I feel like having `editable` and `enabled` in the names end up more confusing. If the state is unset maybe we should just not have that part in the name:

`lightTheme$Editable$Enabled$WithoutButtons` vs `lightTheme$WithoutButtons`

We kinda did that with appearanceReadOnly already. When unset it's just not in the name, we don't have a named value for it like AppearanceReadOnlyAbsent. So it would be doing that consistently and leaving unset parts out of the name

rajsite: Hmm looking again I guess `without buttons` would be removed from my example (we treat it binary and it's literally an absence of buttons).

Alternatively maybe one could argue we should have a name to represent `AppearanceReadOnlyAbsent` so that way adding a dimension is easier, count the number of `$` to know what part of the state you are editing consistently. 

I don't think I feel strongly either way, it's only coming up because the text field is extra complex / has so many states

mollykreis: My leaning is to rename these stories in the text-field-matrix to include all parts of the state. At least in this case where we systematically go through all permutations of readonly on/off, disabled on/off, appearanceReadOnly on/off, and buttons on/off, keeping each name explicit about a state being on or off makes sense.

rajsite (approved):

packages/storybook/src/nimble/anchor-button/anchor-button-matrix.stories.ts
```diff
@@ -57,7 +57,7 @@ const component = (
     </${anchorButtonTag}>
```
jattasNI, line 57: Could you document the naming conventions in your PR description somewhere in the repo? The [storybook package CONTRIBUTING](https://github.com/ni/nimble/blob/main/packages/storybook/CONTRIBUTING.md) is probably my first choice. Inline docs on the matrix utility methods could also work.

packages/storybook/src/nimble/anchor-button/anchor-button-matrix.stories.ts
```diff
@@ -57,7 +57,7 @@ const component = (
...
 `;
```
jattasNI, line 58: Could you apply the guidance to Spright tests too? For example I noticed that Rectangle is still including its component name in the Theme Matrix story.

jattasNI (approved):

packages/storybook/src/nimble/anchor-button/anchor-button-matrix.stories.ts
mollykreis: I added a section to the CONTRIBUTING docs. Let me know if you think I missed anything or if anything could be clearer.

packages/storybook/CONTRIBUTING.md
```diff
@@ -104,6 +104,43 @@ All other Markdown formatting is supported. See any
...
+1. Do not include component name in the story name.
+    - Do: `Hidden`, `Theme Matrix`
+    - Don't: `Hidden Text Field`, `Select Theme Matrix`
+1. Do not include tests for multiple components under one storybook group.
+    - Do: `Radio Group >> Hidden`, `Radio >> Hidden`
+    - Don't: `Radio Group >> Hidden Radio Group` and
+      `Radio Group >> Hidden Radio`
+1. Use `$` as a delimiter in test names that have a constant value for a matrix
+   dimension.
+    - Do: `LightTheme$ Open$ No Filter`
```
jattasNI, line 121:
```suggestion
    - Do: `Light Theme$ Open$ No Filter`
```

packages/storybook/src/nimble/anchor-button/anchor-button-matrix.stories.ts
jattasNI: Just left a 1-character suggestion but otherwise looks good. The "Do" and "Don't" examples are really helpful!

jattasNI (approved):

PR #2600, opened 2025-05-12
Add appearance-readonly to nimble-number-field

mollykreis:
# Pull Request

## 🤨 Rationale

Add `appearance-readonly` attribute to the `nimble-number-field` to style the component as readonly when it is disabled.

I also realized that the cursor was wrong for the `disabled + appearanceReadOnly` state on the `nimble-text-field`, so I fixed it.

## 👩‍💻 Implementation

- Add `appearance-readonly` attribute (tied to `appearanceReadOnly` property) on the number field
- Updated styles to make `disabled + appearanceReadOnly` look like the `readonly` styles
- Updated storybook docs
- Updated matrix tests
    - As part of the matrix test updates, I split the number field matrix tests into 3 different tests (one for each theme) because the canvas of the story got too tall for chromatic otherwise.

## 🧪 Testing

- Manually tested
- Chromatic test

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-3b2df972-f5ae-4415-be0f-8e4268970423.json
+++ b/change/@ni-nimble-components-3b2df972-f5ae-4415-be0f-8e4268970423.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add appearance-readonly to nimble-number-field",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
+++ b/packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
@@ -2,13 +2,13 @@ import type { StoryFn, Meta } from '@storybook/html';
 import { html, ViewTemplate } from '@ni/fast-element';
 import { numberFieldTag } from '../../../../nimble-components/src/number-field';
 import { NumberFieldAppearance } from '../../../../nimble-components/src/number-field/types';
-import { createStory } from '../../utilities/storybook';
+import { createFixedThemeStory, createStory } from '../../utilities/storybook';
 import {
     createMatrixThemeStory,
-    createMatrix,
     sharedMatrixParameters,
     cartesianProduct,
-    createMatrixInteractionsFromStates
+    createMatrixInteractionsFromStates,
+    createMatrix
 } from '../../utilities/matrix';
 import {
     errorStates,
@@ -19,7 +19,8 @@ import {
     requiredVisibleStates,
     type DisabledReadOnlyState,
     disabledReadOnlyStates,
-    disabledReadOnlyState
+    disabledReadOnlyState,
+    backgroundStates
 } from '../../utilities/states';
 import { hiddenWrapper } from '../../utilities/hidden';
 import { textCustomizationWrapper } from '../../utilities/text-customization';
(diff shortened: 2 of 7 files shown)


jattasNI (approved):

packages/storybook/src/utilities/states.ts
```diff
@@ -78,5 +78,8 @@ export const disabledReadOnlyState = {
     readOnly: disabledReadOnlyStates[4],
     readOnlyAppearanceReadOnly: disabledReadOnlyStates[5],
     readOnlyDisabled: disabledReadOnlyStates[6],
-    readOnlyDisabledAppearanceReadOnly: disabledReadOnlyStates[7]
+    readOnlyDisabledAppearanceReadOnly: disabledReadOnlyStates[7],
+    allDisabledAbsentStates: disabledReadOnlyStates.filter(
+        (x: readonly (boolean | string)[]) => x[2] === false
```
rajsite, line 83: Could we keep this as a separate state array? I'm thinking of `disabledReadOnlyState` as names of individual states and then we have different state arrays filtered to a subset of the permutations

```ts
export const disabledReadOnlyStates = [/* unchanged */] as const;
export type DisabledReadOnlyState = (typeof disabledReadOnlyStates)[number];

export const disabledAbsentStates = disabledReadOnlyStates.filter(
    (state: readonly [name: string, readOnly: boolean, disabled: boolean, appearanceReadonly: boolean]) => state[2] === false
);
export type DisableAbsentState = (typeof disabledAbsentStates)[number];

export const disabledReadOnlyState = { /* allDisabledAbsentStates entry removed*/ } as const;
```

nit: In that example I also show naming the tuples on the states during filter which I think makes it a bit easier to understand but is purely documentation / a convenience / not type enforced.

rajsite (approved):

packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
```diff
@@ -73,34 +76,69 @@ const component = (
...
+] = backgroundStates;
+
+if (remaining.length > 0) {
+    throw new Error('New backgrounds need to be supported');
+}
+
+export const lightTheme: StoryFn = createFixedThemeStory(
+    createMatrix(component, [
+        requiredVisibleStates,
+        disabledReadOnlyStates,
```
rajsite, line 104: nit / pref, could we move up the appearance states so those controls are grouped together?
```
disabledReadOnlyStates,
appearanceStates,
requiredVisibleStates,
hideStepStates,
valueStates,
errorStates
```

Guess a concept from that could be that larger dimensions should be first so that you have bigger groupings spread out (like 8 different categories for disabled states) and then binary dimensions last which keeps the smaller sets of variations closer together.

change/@ni-nimble-components-3b2df972-f5ae-4415-be0f-8e4268970423.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Think you mentioned it but the snapshots looked snipped off, are these stories too large now?
image

packages/storybook/src/utilities/states.ts
mollykreis: My only concern with the suggested approach is that the name `disabledAbsentStates` loses the context that it is the `disabledReadOnlyStates` where `disabled === false`. Perhaps the type is sufficient?

rajsite: Maybe a name like `onlyDisabledAbsentStates`  gives a better hint that it's associated with more context?

> Perhaps the type is sufficient?

I don't think I follow the implications of the idea.

change/@ni-nimble-components-3b2df972-f5ae-4415-be0f-8e4268970423.json
mollykreis: This is actually a different problem that the bottom margin of the control wasn't large enough to account for error text possibly being below the bottom row of controls. This isn't a problem any more with the way I'm planning to reorder the matrix stories (i.e. the bottom row shouldn't have error text anymore), but I've changed the margin to be more robust.

PR #2602, opened 2025-05-13
Add appearance-readonly to nimble-select and nimble-combobox

mollykreis:
# Pull Request

## 🤨 Rationale

Add `appearance-readonly` attribute to the `nimble-select` and `nimble-combobox` to style the components as readonly when they are disabled.

## 👩‍💻 Implementation

- Add `appearance-readonly` attribute (tied to `appearanceReadOnly` property) on the select and combobox
- Added styles to make `disabled + appearanceReadOnly` look like other readonly controls (neither select nor combobox has a `readonly` property or styles today)
- Updated storybook docs
- Updated matrix tests
    - As part of the matrix test updates, I split the select matrix story into 3 stories (one for each theme) because chromatic was timing out loading the story

## 🧪 Testing

- Manually tested
- Chromatic test

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/select/styles.ts
+++ b/packages/nimble-components/src/select/styles.ts
@@ -37,6 +37,18 @@ export const styles = css`
     .selected-value {
         order: 1;
     }
+    
+    .selected-value.placeholder {
+        color: ${placeholderFontColor};
+    }
+
+    :host([disabled]) .selected-value.placeholder {
+        color: ${bodyDisabledFontColor};
+    }
+
+    :host([disabled][appearance-readonly]) .selected-value.placeholder {
+        color: ${placeholderFontColor};
+    }
 
     .clear-button {
         order: 3;
--- a/packages/nimble-components/src/select/styles.ts
+++ b/packages/nimble-components/src/select/styles.ts
@@ -34,14 +34,6 @@ export const styles = css`
          error icon, and dropdown arrow because they are not "interactive" i.e. part of the tab order */ ''
     }
 
-    .selected-value.placeholder {
-        color: ${placeholderFontColor};
-    }
-
-    :host([disabled]) .selected-value.placeholder {
-        color: ${bodyDisabledFontColor};
-    }
-
     .selected-value {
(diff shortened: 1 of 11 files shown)


packages/nimble-components/src/select/styles.ts
```diff
@@ -42,6 +42,10 @@ export const styles = css`
         color: ${bodyDisabledFontColor};
     }
 
+    :host([disabled][appearance-readonly]) .selected-value.placeholder {
```
jattasNI, line 45: I should have thought of this in one of your previous PRs so it doesn't need to be addressed in this one. Feel free to close this thread and take the conversation elsewhere, then address in a separate PR across all controls if needed.

Do we want to be displaying the placeholder text when a component is disabled and appearance-readonly? It will typically instruct the user to enter a value, which they can't do. It seems more correct to show a blank value in this case. Or maybe for clients to set different placeholder text that says something like "No value set".

We could write usage guidance instructing clients to clear the placehoder text when disabled or when both disabled and appearance-readonly. Or to set different placeholder text. But if we think the placeholder should always hide in those cases we could do it automatically.

mollykreis: I'm hesitant to do anything automatically when the control is disabled since we don't do that with most of our other attributes (e.g. `required-visible` or `error-visible`) that generally also wouldn't be used on a disabled control. We tend to be non-opinionated about how some of these configurations should be used together, so I think it makes sense to follow the same here.

I'm on the fence about writing guidance about the placeholder. The guidance we're following is the SLE, but I'm not sure how generally we want to state that as the best practices for all clients of nimble controls.

jattasNI: Good points, I'm fine with us not doing anything in Nimble. Maybe an update to the [Placeholder section of the Stratus Control Text pattern doc](https://stratus-storybook.ni.dev/?path=/docs/patterns-control-text--docs#placeholder) would be useful though.

rajsite (approved):

PR #2606, opened 2025-05-14
Update nimble-number-field to support frameless and full-bleed styling

mollykreis:
# Pull Request

## 🤨 Rationale

Updated the `nimble-number-field` appearance modes to include `frameless` and to also have the option to be `full-bleed` (which only applies when the appearance is `frameless`). This is part of the effort to be able to render more nimble controls in a "read-only" manner where the value is aligned with the label.

See #2592 

## 👩‍💻 Implementation

- Add `frameless` appearance for the number field
- Add `full-bleed` attribute to the number field

## 🧪 Testing

- Manual testing in storybook
- Updated chromatic tests
    - Because the theme tests got too big for the number field, I split each story into 2 -- one with the step buttons visible and one with the step buttons hidden.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-66b0fc49-9117-4442-8063-c9c8baf59521.json
+++ b/change/@ni-nimble-components-66b0fc49-9117-4442-8063-c9c8baf59521.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Update nimble-number-field to support frameless and full-bleed styling",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
+++ b/packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
@@ -19,7 +19,9 @@ import {
     requiredVisibleStates,
     type DisabledReadOnlyState,
     disabledReadOnlyStates,
-    backgroundStates
+    backgroundStates,
+    fullBleedStates,
+    type FullBleedState
 } from '../../utilities/states';
 import { hiddenWrapper } from '../../utilities/hidden';
 import { textCustomizationWrapper } from '../../utilities/text-customization';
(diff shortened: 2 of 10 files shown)


packages/storybook/src/nimble/number-field/number-field-matrix.stories.ts
```diff
@@ -143,15 +194,17 @@ const notRequiredState = requiredVisibleStates[0];
...
+    fullBleedStates,
     [notRequiredState],
     [hideStepStateStepVisible],
     [valueStatesHasValue],
     [errorStatesNoError, errorStatesErrorWithMessage]
 ] as const);
 
 const interactionStates = cartesianProduct([
-    disabledReadOnlyStates,
+    onlyDisabledAbsentStates,
```
mollykreis, line 205: Note: This is fixing an oversight in my previous change (#2600). These interaction states should not be run on disabled controls, so I've updated the states that get passed here.

change/@ni-nimble-components-66b0fc49-9117-4442-8063-c9c8baf59521.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Do we want to provide any usage guidance about how `frameless` and `full-bleed` should be used alongside `readonly` or `disabled`? For example (not sure if these are 100% true):
1. always use frameless and full bleed if it's readonly
2. don't use frameless and full bleed if it's not readonly
3. only use full bleed if nearby left-aligned controls are also full bleed

If you think it's worth adding, let's add something similar to text field. But I could also see an argument that we either don't have concrete guidance yet or that this is like other invalid combinations that we don't document, so it's not worth documenting this one.

jattasNI (approved):

change/@ni-nimble-components-66b0fc49-9117-4442-8063-c9c8baf59521.json
mollykreis: I'm not sure. The first item in the list is not true, so we definitely wouldn't want to document that. However, I certainly could see a case for documenting (2). I'm not aware of a use case where we'd want to have a frameless + full-bleed control that is editable (since it doesn't look editable until you hover).

While it isn't a great argument, I'm leaning towards not documenting anything since we don't have concrete guidelines. We also didn't document anything for the frameless + full-bleed text-field (though that's not really a good argument for why we shouldn't document something now).

rajsite (approved):

Issue #2610, opened 2025-05-20
Spright chat feature backlog


jattasNI:
# 🙋 Feature Request

## 😯 Problem to Solve

This is a parent feature to collect smaller tasks related to chat components in Spright. A high level goal is to evaluate [these Blazor-only changes](https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FMeasurementServices%2FAiAssistants%2FControls%2FComponents&version=GBfeatures%2FAIAssistants&_a=history) and [additional features described in Figma](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1529-41777&p=f&t=m4UM0H0k6Na8NERl-0) for inclusion in Spright.

## 💁 Proposed Solution

See task list.

### Chat input

- [ ] Basic chat input component for SLE demo (input and button)
- [ ] Chat input component parity with Blazor implementation (file attachments etc)

### Chat conversation

- [ ] #2607 
- [ ] gradient background for chat conversation
- [ ] make space and scroll when new messages are added

### Components around conversation

- [ ] [Footer with disclaimer message](https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FMeasurementServices%2FAiAssistants%2FControls%2FComponents%2FChatbotViewFooter.razor&version=GBmain&_a=contents)
- [ ] toolbars?

### Chat messages

- [ ] separate components for different message types
- [ ] link to sources

### Chat message content

- [ ] markdown renderer that supports syntaxes returned by Nigel

PR #2611, opened 2025-05-20
Update Spright chat spec to add input component

jattasNI:
# Pull Request

## 🤨 Rationale

This is the design doc for the first task in #2610 

## 👩‍💻 Implementation

Update existing spec to propose a new component, `spright-chat-input`.

## 🧪 Testing

N/A

## ✅ Checklist


packages/spright-components/src/chat/specs/README.md
```diff
@@ -78,15 +71,40 @@ The component also contains the following features:
...
+    - pressing Enter while the text area has focus will behave the same as clicking "Send"
+    - pressing Shift-Enter will create a newline
+    - the button is disabled and Enter has no effect if the input text is empty
+1. Styling for default, focus, and rollover states
+
+##### Future work
+
+This is included for informational purposes but doesn't yet have a proposed implementation in the Design section below.
+
+1. Includes a "Stop" button for the user to abort an in-progress sent message
```
atmgrifter00, line 92: Is there a practical known use-case where we expect this could be useful? I'm guessing this is for use-cases where some pre-processing is done on the message before actually creating a `spright-chat-message` with its contents (I wasn't sure if this was a common/expected scenario)? Or would this be related to the other future work for things like attaching files to cancel an in-progress upload (in which case, I think the verbiage would change to reflect that)?

packages/spright-components/src/chat/specs/README.md
```diff
@@ -78,15 +71,40 @@ The component also contains the following features:
...
+    - the text area height is a single line initially but grows its height to fit the entered text up to its max-height
+    - the text area has configurable placeholder text
+1. Includes a "Send" button for the user to submit the current input text
+    - fires an event containing the current input content and then clears the content and sets keyboard focus back to the input
+    - pressing Enter while the text area has focus will behave the same as clicking "Send"
+    - pressing Shift-Enter will create a newline
+    - the button is disabled and Enter has no effect if the input text is empty
+1. Styling for default, focus, and rollover states
+
+##### Future work
```
atmgrifter00, line 88: Has there been any reasoning behind _not_ creating a component that wraps both the `spright-chat-input` and `spright-chat-conversation` components/aspects? It seems like all of the designs call for there to be an input along with the messages themselves, and requiring developers to lay out two sibling components in the DOM in likely _always_ the same way, seems strange and unnecessary.

packages/spright-components/src/chat/specs/README.md
jattasNI: Good question. I don't know for sure, but in the Figma design it looks like it's more about canceling a response to a sent message. I'll update this to be more vague for now. We can revisit if / when we move this from "Future work" to "In scope".

packages/spright-components/src/chat/specs/README.md
jattasNI: Excellent question. I hadn't really thought about that approach much. I agree that we shouldn't let the existing Blazor implementations stop us from going in that direction if we think it's better. As an alternative to a hideable CSS `part`, an `input-hidden` attribute or something could work too.

I can think of two cons / things that would need to be worked out:
1. The API of this component could grow to be somewhat large. Eventually it would contain:
    - existing conversation slot for adding messages
    - this spec's proposed input attributes for configuring the input and event for the send button
    - whatever new API is needed to support the input "Future work" section (likely a couple more slots and some more attributes and events)
    - whatever new API is needed to support the conversation features in #2610 (likely some methods for scrolling the conversation)
2. this would open us to some debate about where the responsibilities of that component would end. Is it purely presentational? Or should it:
   - create an outgoing message when the user hits Send
   - automatically scroll to make space when the user hits Send
   - manage other interactions like drag-drop to attach files

An Alternative 3 would be something like `spright-chat-window` which is a container/layout component that contains slots where you're expected to insert your own `spright-chat-conversation` and `spright-chat-input`. It would do nothing other than lay them out correctly. In future if we add toolbars at the top/side/bottom, those could go in slots too without bloating the API of the container component.

I'm still leaning to Alternative 1 in the spec already or Alternative 3 which would be follow-up work, but open to other opinions or rationale.

rajsite: I think chat conversation already is like option 3. Currently it defines layout of chat-messages via the default slot. I think we could extend it to define layout of an "input" component if slotted as well.

Buy I'd be hesitant to encode the input api on the conversation layout element though.

Gives the flexibility of defining different "input" contents like we have already done for "message" contents in the default slot, i.e. you can really put any element that handles the layout of being placed in a block flex.

As an aside I really want us to split out the messages into separate `message-inbound`, `message-outbound`, etc elements and mark the current `message` element as deprecated. Gives more room for flexibility without bloating existing apis like happened on that element.

jattasNI: I updated the proposal to include slotting the input into a dedicated slot in the conversation.

>I really want us to split out the messages into separate message-inbound, message-outbound, etc 

I tracked this in #2610. Still figuring out the priority between all that work, but I agree it's something we should do.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -129,11 +149,22 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
...
+</spright-chat-message>
+```
+
+<!-- prettier-ignore-end -->
+
+#### Input example
+
+```html
+<spright-chat-input placeholder="Ask Nigel" send-button-label="Send">
+</spright-chat-input>
```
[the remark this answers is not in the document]
jattasNI: I did that in the conversation example above. The input can technically be used on its own so I didn't also do it in this example, similar to the messages examples above. I'll leave this as-is but if you really want that change feel free to re-open this thread.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -76,17 +69,43 @@ The component also contains the following features:
...
+    - pressing Enter while the text area has focus will behave the same as clicking "Send"
+    - pressing Shift-Enter will create a newline
+    - the button is disabled and Enter has no effect if the input text is empty
+1. Styling for default, focus, and rollover states
+
+##### Future work
+
+This is included for informational purposes but doesn't yet have a proposed implementation in the Design section below.
+
+1. Includes a "Stop" button for the user to abort an action that was triggered in response to a sent message (e.g. a file upload or incoming response)
```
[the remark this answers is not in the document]
jattasNI: Good questions. 

The typical flow would be that after the user sent their text a client would call `resetInput()` and optionally set the state to `processing`. With the current proposal, `resetInput()` would set focus back to the text area regardless of whether the user sent the text using Enter in the text area, clicked the Send button, or tabbed to the Send button and pressed Enter. That seems sane to me since the user likely wants to type another message more than they want to stop the current one from being processed.

If a client doesn't call `resetInput()` at the same time they set the state to `processing` then the entered text would remain in the text area. That definitely seems weird to me so I don't think we'd encourage clients to do that. We're not currently offering a way to clear the text but maintain the focus on whatever had it before. If we need to do that we could potentially pass some config options to `resetInput()` to configure whether it should set focus. I agree that in that case you'd probably want the Stop button to get focus if Send had it before (and vice versa). But in the interest of scope I'd advocate to skip doing that for now.

Similarly I'd vote to punt on the question of announcing button changes (this is in Spright so we can be looser with our accessibility behavior). I don't know if users would find those announcements useful or annoying if they happened on every sent message.

I added a couple sentences to the Accessibility section to capture those proposals. Let me know if you disagree.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -171,6 +198,44 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
...
+- _How native CSS Properties (height, width, etc.) affect the component_
+    - Clients can set the input width using normal CSS rules. The input will have a default minimum width that clients are discouraged from overriding.
+    - The input will have a default height to fit one line of text and will grow its height to fit more lines, up to a max-height. After that limit it will show a vertical scrollbar.
+- _Slots_
+    - `footer-actions`
+        - Action buttons to display left of the send/stop button.
+    - `attachments`
+        - An area to slot arbitrary content adjacent to the text input area. Intended to be used for adding chips that represent attached files.
+
+##### Input API alternatives
```
jattasNI, line 229: @atmgrifter00 here's my rationale for the `value` API we discussed this morning

jattasNI (commented):
@rajsite @mollykreis heads up that I'm starting implementation on the first vertical slice of this. If you have feedback on the parts that configure the text area and send button, it'd be good to get agreement before I write tests. 

The stop button, error state, and slots are out of scope for my first slice of implementation.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -76,17 +69,39 @@ The component also contains the following features:
...
+#### Chat input
+
+1. Accepts text input in a text area
+    - the text area height is a single line initially but grows its height to fit the entered text up to its max-height
+    - the text area has configurable placeholder text
+1. Includes a "Send" button for the user to submit the current input text
+    - fires an event containing the current input content
+    - pressing Enter while the text area has focus will behave the same as clicking "Send"
+    - pressing Shift-Enter will create a newline
+    - the button is disabled and Enter has no effect if the input text and attachments slot are both empty or if the client has set an attribute to explicitly disable it
```
rajsite, line 84: Quickly tested Teams mobile vs Teams desktop for multiline and the behavior was:
On desktop pressing enter sent immediately, had to use shift+enter for multiline
On mobile pressing enter added a new line, had to use the send button to send a message

Dunno if we have UX guidance around that, should just do what Teams does, but should capture that somewhere otherwise mobile won't really have a way to do multiline I can think of.

jattasNI: Good point, I'll add a note to the mobile section and a task to the issue tracking this work.

Any ideas how to detect that the input came from a mobile keyboard? The only approaches I've found online are to look at user agent or to detect using a heuristic like whether the viewport size has recently changed (presumably due to an onscreen keyboard). Both of those seem like heuristics with less than perfect reliability. Some day maybe [virtual keyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API) could help, but it's not widely supported.

rajsite: > Any ideas how to detect that the input came from a mobile keyboard? The only approaches I've found online are to look at user agent or to detect using a heuristic like whether the viewport size has recently changed (presumably due to an onscreen keyboard).

Yea not sure off the top. What if they were more aligned? Enter always adds a newline and the button submits, desktop gets an extra shortcut for the submit button (shift + enter).

packages/spright-components/src/chat/specs/README.md
```diff
@@ -171,6 +198,46 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
...
+
+#### Input
+
+- _Component Name_ `spright-chat-input`
+- _Props/Attrs_
+    - `value` - string attribute to get or set the current text in the text area. See [Input API alternatives](#input-api-alternatives) for more info.
+    - `processing` - boolean attribute that causes the "Stop" button to be shown rather than the "Send" button
+    - `placeholder` - text to display in the text area when no text has been entered
+    - `send-button-label` - text to use for a `title` and ARIA attributes on the "Send" button. See Accessibility section for more info.
+    - `stop-button-label` - text to use for a `title` and ARIA attributes on the "Stop" button. See Accessibility section for more info.
```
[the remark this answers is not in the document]
rajsite: +1 good catch. And for scoping of that label provider seems like `spright-chat-label-provider` is a good scope for all the `spright-chat-*` elements.

packages/spright-components/src/chat/specs/README.md
rajsite: Having stewed on it more I agree with the idea that short quick messages is probably the general workflow. If that's the case I dunno why we would have a different default on mobile. Maybe there is a a well-established UX trend that enter is not commonly used on mobile to submit?

The heuristics based on viewport are also not super promising for mobile detection given our current blazor usage patterns of small / narrow viewports.

Maybe we can move the newline mobile behavior to open questions with the suggestion of aligning with Teams behavior instead of a decision? We might also make it a configuration of the chat-input and let applications decide instead of trying to guess a heuristic in nimble (or have a default auto and make it overridable or something).

packages/spright-components/src/chat/specs/README.md
```diff
@@ -171,6 +198,46 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
...
+    - `stop` - emitted when the user clicks the "Stop" button. Not cancelable.
+- _CSS Classes and CSS Custom Properties that affect the component_
+- _How native CSS Properties (height, width, etc.) affect the component_
+    - Clients can set the input width using normal CSS rules. The input will have a default minimum width that clients are discouraged from overriding.
+    - The input will have a default height to fit one line of text and will grow its height to fit more lines, up to a max-height. After that limit it will show a vertical scrollbar.
+- _Slots_
+    - `footer-actions`
+        - Action buttons to display left of the send/stop button.
+    - `attachments`
+        - An area to slot arbitrary content adjacent to the text input area. Intended to be used for adding chips that represent attached files.
```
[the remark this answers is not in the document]
jattasNI: Good point, I deleted the word "arbitrary". The next sentence describes the intended content in more detail and I agree we'll want usage guidance / testing that limits clients to what is designed to go there.

packages/spright-components/src/chat/specs/README.md
jattasNI: I updated the description in the Mobile section to describe the challenge and say that we won't solve it now. Also captured a summary of this discussion. I didn't capture it as an open issue in the document since we have a few other similar conscious scoping gaps that aren't listed there either. But it is captured in #2610.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -76,17 +69,43 @@ The component also contains the following features:
...
 
+#### Chat input
+
+##### In scope
+
+1. Accepts text input in a text area
+    - the text area height is a single line initially but grows its height to fit the entered text up to its max-height
+    - the text area has configurable placeholder text
+1. Includes a "Send" button for the user to submit the current input text
+    - fires an event containing the current input content and then clears the content and sets keyboard focus back to the input
```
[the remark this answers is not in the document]
rajsite: Out of office, so just some passing feedback. Could we make that behavior opt-in instead of the default? I.e. a `reset-on-send` attribute or something.

The designs today don't include error-text and validation patterns which I'm concerned auto clearing will fight against, but if it's opt-in that seems less of a concern.

PR #2613, opened 2025-05-22
Update select and combobox to support frameless and full-bleed styling

mollykreis:
# Pull Request

## 🤨 Rationale

Updated the `nimble-select` and `nimble-combobox` appearance modes to include `frameless` and to also have the option to be `full-bleed` (which only applies when the appearance is `frameless`). This is part of the effort to be able to render more nimble controls in a "read-only" manner where the value is aligned with the label.

See #2592 

## 👩‍💻 Implementation

- Add `frameless` appearance for the drop-down appearance modes
- Add `full-bleed` attribute to the select and combobox

## 🧪 Testing

- Manual testing in storybook
- Updated chromatic tests

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-b44508aa-2a65-4419-8047-367e63603ea2.json
+++ b/change/@ni-nimble-components-b44508aa-2a65-4419-8047-367e63603ea2.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Update select and combobox to support frameless and full-bleed styling",
+  "packageName": "@ni/nimble-components",
+  "email": "20542556+mollykreis@users.noreply.github.com",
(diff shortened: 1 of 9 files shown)


jattasNI (approved):

change/@ni-nimble-components-b44508aa-2a65-4419-8047-367e63603ea2.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @mollykreis looks like some of the stories are clipped on the edge like we saw before for error text in the last row.
Looks good otherwise, feel free to approve stories and merge after addressed.

PR #2636, opened 2025-06-19
Dotnet clean

rajsite:
# Pull Request

## ✅ Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,6 +43,10 @@ jobs:
           fetch-depth: 0
 
       # Install dependencies
+      - run: sudo apt remove --purge dotnet-sdk-8.0
+        working-directory: packages/blazor-workspace/
+      - run: dotnet --list-sdks
+        working-directory: packages/blazor-workspace/
       - uses: actions/setup-dotnet@v4
         with:
           global-json-file: packages/blazor-workspace/global.json
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,9 +43,8 @@ jobs:
           fetch-depth: 0
 
       # Install dependencies
-      - run: sudo apt remove --purge dotnet-sdk-8.0
-        working-directory: packages/blazor-workspace/
-      - run: dotnet --list-sdks
+      - name: Remove pre-existing dotnet sdk 8 installs
+        run: sudo apt remove --purge dotnet-sdk-8.0
         working-directory: packages/blazor-workspace/
       - uses: actions/setup-dotnet@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -43,6 +43,9 @@ jobs:
           fetch-depth: 0
 
       # Install dependencies
+      - name: Remove pre-existing dotnet sdk 8 installs
+        run: sudo apt remove --purge dotnet-sdk-8.0
+        working-directory: packages/blazor-workspace/
       - uses: actions/setup-dotnet@v4
```
jattasNI, line 49: >If the image version of the sdk is newer, then even if we install a locked older version of the SDK, our lockfile checks will fail

Just to double check, what mechanism did you use to install a locked older version? Was it [specifying an exact version using the `dotnet-version` parameter](https://github.com/actions/setup-dotnet?tab=readme-ov-file#supported-version-syntax)? If you haven't tried that already then it might be worth a go.

.github/workflows/main.yml
```diff
@@ -43,6 +43,9 @@ jobs:
...
         with:
```
jattasNI, line 50: I'm kind of surprised the build will attempt to use a newer SDK if it's available. In [global.json we specify `"rollForward": "disable"`](https://github.com/ni/nimble/blob/14562d199351070b2479a28652dbabdee918c52e/packages/blazor-workspace/global.json#L5), which is [supposed to force it to use the exact version we ask for](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward) and throw an error if that's not available.

I don't know why that wouldn't behave the way it's documented, but .NET versioning seems rather complicated so I doubt it's worth digging into. Although one thing I'll note is that we have an unrelated typo in our global.json: 

`"allowPrelease": "false",` should be
`"allowPrerelease": "false",`

As a shot in the dark it might be worth fixing that and seeing if the version behavior changes.

jattasNI (approved):
I left a couple notes about other ideas that came up while I was trying to understand how this works. Up to you how far to pursue them, I think the current approach is acceptable.

.github/workflows/main.yml
rajsite: We are using the [`global-json-file`](https://github.com/actions/setup-dotnet?tab=readme-ov-file#using-the-global-json-file-input) option.
I'm skeptical that using `dotnet-version` would behave differently (I don't know why it would try and uninstall old versions). It would also lead to duplicating the version used in two spots which is a different trade-off I'm not a huge fan of. If interested in that approach I can investigate it in a follow-on.

PR #2638, opened 2025-06-20
Spright chat input component and Angular wrapper

jattasNI:
# Pull Request

## 🤨 Rationale

This is the first vertical slice of the component in #2611. Part of #2610.

This adds the `spright-chat-input` component and Angular wrapper. 

The component supports: 
- user input (including multi-line via Shift-Enter)
- send button (and Enter key to send)
- resetInput() method to clear input after send
- placeholder text

Coming in future PRs:
- Blazor wrapper
- styling for states like hover/focus
- error state
- stop button and "processing" state
- slots for additional content like attachments

## 👩‍💻 Implementation

Pretty standard component and Angular integration, as described in the HLD. I'll mention some interesting notes in comments.

## 🧪 Testing

1. Unit tests for component and Angular directive
2. Matrix tests for component
3. Added to Angular example app and Storybook

## ✅ Checklist


packages/spright-components/src/chat/input/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        font: ${bodyFont};
+        color: ${bodyFontColor};
+        background-color: transparent;
+
+        width: 100%;
+        resize: none;
+        height: auto;
+        max-height: calc(6lh + 2 * ${mediumPadding});
+        field-sizing: content;
```
jattasNI, line 46: As mentioned in the spec this isn't yet supported in Firefox and Safari but that's OK for our current use cases.

packages/spright-components/src/chat/input/testing/chat-input.pageobject.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.element.textArea.focus();
+        await this.sendEnterKeyEvents(true);
+    }
+
+    private getSendButton(): Button {
+        const sendButton = this.element.shadowRoot!.querySelector<Button>('.send-button')!;
+        return sendButton;
+    }
+
+    private async sendEnterKeyEvents(shiftKey: boolean): Promise<void> {
```
jattasNI, line 73: Meyer and I spent a while figuring out the right way to test the enter key events. As we do elsewhere, this ends up dispatching events and modifying underlying values manually which the browser will do for us. But it's still useful to test the prevent default implementation of the real component.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -86,14 +109,45 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
...
+            inputRef.resetInput();
+            const message = document.createElement(chatMessageTag);
+            message.messageType = ChatMessageType.outbound;
+            const span = document.createElement('span');
+            span.textContent = event.detail.text;
+            // Preserves new lines and trailing spaces that the user entered
+            span.style.whiteSpace = 'pre-wrap';
+            message.appendChild(span);
+            conversationRef.appendChild(message);
+            message.scrollIntoView({
```
jattasNI, line 146: Eventually we probably want the conversation to handle scrolling to show new messages, but that's out of scope for now.

packages/spright-components/src/chat/message/index.ts
```diff
@@ -34,7 +34,7 @@ export class ChatMessage extends FoundationElement {
      * HTML Attribute: message-type
      */
     @attr({ attribute: 'message-type' })
-    public readonly messageType: ChatMessageType = ChatMessageType.system;
+    public messageType: ChatMessageType = ChatMessageType.system;
```
jattasNI, line 37: This was just an oversight. It's necessary to be able to change a message type after construction, for example in `chat-conversation.stories.ts` where we construct a message from TS code when the user presses Send.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
rajsite: The current implementation shown here isn't behaving very well for me on windows chrome:

scroll-lines

As an aside this looks like an interesting way to get pinned scrolling: https://css-tricks.com/books/greatest-css-tricks/pin-scrolling-to-bottom/

jattasNI: I'd like to propose we punt on this for now. It's tracked in #2610 ("make space and scroll when new messages are added") and it's going to get even more complicated since the UX request is to make enough space that the new message scrolls all the way to the top (ie adding extra whitespace below the message).

I did play around with a couple approaches:
1. changing the settings on `scrollIntoView()` to scroll to the bottom of the message rather than the top. But this ends up scrolling the entire document (Jose had similar problems when he implemented this in his Blazor app)
2. adding a method to the conversation like `scrollToEnd()` and calling that when a new message is added. It could either call `scrollTo({top: Number.MAX_SAFE_INTEGER})` on the scrollable element in the shadow DOM or manage a hidden item after the last message and call `scrollIntoView()` on it. But that would require an API change (and HLD debate) and wouldn't solve scrolling as the input grows.
3. using the approach in the blog you linked I was able to scroll to bottom as new messages were added but I couldn't get it to animate (I tried setting `scroll-behavior: smooth` on the scrollable container) and it also didn't solve scrolling as the input grows.

atmgrifter00 (approved):

packages/react-workspace/spright-react/src/chat/input/index.ts
```diff
@@ -0,0 +1,4 @@
+import { ChatInput } from '@ni/spright-components/dist/esm/chat/input';
+import { wrap } from '../../utilities/react-wrapper';
+
+export const SprightChatInput = wrap(ChatInput);
```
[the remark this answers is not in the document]
rajsite: Don't mind leaving the behavior as-is, if your up for tweaking the style to align with the other apps I'd appreciate that: https://github.com/ni/nimble/blob/5c3b45b4c6b189ef2f40c5064ef1dd87bdee90be/packages/react-workspace/react-client-app/src/components/App.scss#L99

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+                element.resetInput();
+                processUpdates();
+
+                expect(element.value).toEqual('');
+                expect(page.getRenderedText()).toEqual('');
+                expect(page.textAreaHasFocus()).toBeTrue();
+                expect(page.isSendButtonEnabled()).toBeFalse();
+            });
+
+            element.addEventListener('send', spy);
```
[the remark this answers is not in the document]
jattasNI: I was able to use `waitForEvent` without duplicating it; the folder it's in is exported from nimble-components so it can be used from page objects. It's not intended for "public" use but using it internally within the repo seems ok to me.
https://github.com/ni/nimble/blob/c7b9daabf88cfa42e088fac0fc80604ee3b301cc/packages/nimble-components/src/utilities/testing/README.md?plain=1#L3

I like how `waitForEvent` cleaned up the two tests that assert the event is fired. I didn't see a good way to use it in the tests that assert an event is not fired. I kept those as they are and added a new test following the same pattern that asserts the event is fired synchronously. This is necessary to ensure that the other tests are actually detecting the absence of a callback, not just finishing before the callback happens. That was previously validated by the two tests that now use `waitForEvent` but that's async so they're no longer validating that.

Hopefully that makes sense, let me know if you like the old way better or can think of other ways to modify the "triggers no send event" tests.

Issue #2579, opened 2025-04-03
Investigate using :has-slotted when it's available in all browsers


jattasNI:
# 🧹 Tech Debt

We have some places where we apply styles based on whether a slot contains content. Our current pattern for doing this involves using the FAST `slotted` directive to run JS code that tracks content being added and removed, then set classes on the component based on that state.

The [native :has-slotted CSS pseudoclass](https://developer.mozilla.org/en-US/docs/Web/CSS/:has-slotted) is meant to provide the same capability using pure CSS. It's not currently available in Safari but once it is we should investigate whether we can use it instead to simplify our code and reduce execution overhead.

A non-exhaustive list of places that do this:
 - [chat message `footer-actions` slot](https://github.com/ni/nimble/blob/58f200e3e63687194296075c7990595add1c8371/packages/spright-components/src/chat/message/index.ts#L47)
 - [dialog `footer` slot](https://github.com/ni/nimble/blob/58f200e3e63687194296075c7990595add1c8371/packages/nimble-components/src/dialog/index.ts#L66)
 - [ni/FAST start/end slot pattern](https://github.com/ni/fast/blob/4a6152c50851e47bb654c7bfaf2ced55aa5f553e/packages/web-components/fast-foundation/src/patterns/start-end.ts#L45)

Lots more context on this CSSWG issue, including a tangent about why `:has` isn't appropriate since it doesn't operate across the shadow root boundary: https://github.com/w3c/csswg-drafts/issues/6867

PR #2639, opened 2025-06-25
Update Spright chat conversation styling

jattasNI:
# Pull Request

## 🤨 Rationale

The [chat component visual design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=16245-43453&t=gHQn9RfFbEU5cLaB-0) includes some style differences to what I implemented originally. It should have:
- a gradient background on the chat conversation
- a 1px border around the conversation
- different padding:
    - better align the left/right edges of the input with the messages
    - conversation provides padding that apps were having to provide (will remove in [SLE PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/1003566))

While discussing this with Brandon he also gave guidance on some new token values that apps should use to set dividers between panes.

This is one of the tasks in #2610: "gradient background for chat conversation". 

## 👩‍💻 Implementation

1. new / updated divider tokens in nimble-components. I'll use these between the chat pane and the application in SLE. Brandon is building a list of other places in SLE that should be updated too.
1. new background gradient token in nimble-components
    - it might be a bit unexpected to include Spright-related tokens within Nimble but we don't want to duplicate all that infrastructure in Spright
    - [naming scheme](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#naming) is:
        - element=`section`
        - part=`background`
        - type=`image`
     - included adding support to storybook and token system for rendering CSS image as token
3. use tokens in styles in spright chat conversation to make a gradient
4. updates to chat borders and padding. Includes adopting the pattern from #2579 to ensure an empty input slot doesn't take up space with its padding

## 🧪 Testing

Storybook inspection and demo to Brandon.

I also noticed one of the matrix stories wasn't rendering correctly because I'd forgotten to set the slot on the input within the conversation and fixed that oversight.

## ✅ Checklist


packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -18,8 +23,22 @@ export const styles = css`
...
         row-gap: 32px;
-        padding: ${mediumPadding};
-        background: ${applicationBackgroundColor};
+        padding: ${mediumPadding} ${standardPadding} ${mediumPadding}
+            ${standardPadding};
+        border: ${borderWidth} solid ${applicationBackgroundColor};
+        background: linear-gradient(
+            ${chatConversationBackgroundGradientStartColor},
+            ${chatConversationBackgroundGradientEndColor}
+        );
```
rajsite, line 32: What if we added a [gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient) or even [image](https://developer.mozilla.org/en-US/docs/Web/CSS/image) as a new [type suffix](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/theme-provider/design-token-names.ts#L287)? (think I'd lean more to `image`)

Can still design the token to be responsive to theme and just avoid the intermediary tokens (which don't seem valuable on their own?).

packages/nimble-components/src/theme-provider/design-token-comments.ts
```diff
@@ -21,6 +21,10 @@ export const comments: { readonly [key in TokenName]: string } = {
         'Font color for "accent" appearance-variant outline buttons',
     buttonBorderAccentOutlineColor:
         'Border color for "accent" appearance-variant outline buttons',
+    chatConversationBackgroundGradientStartColor:
```
rajsite, line 24: > element=chat-conversation
> part=background-gradient-start/background-gradient-end (you could argue that gradient-start is part of the states or the type, but the outcome is the same)
> type=color

For element, it feels like chat-conversation is unnecessarily specific. Maaaaybe it would also be used in comment threads or we'd have multiple widgets oriented around chat that would leverage the same background. What do you think about `element = chat`? I also don't feel strongly about this point.

packages/spright-components/src/chat/conversation/template.ts
```diff
@@ -1,9 +1,13 @@
-import { html } from '@ni/fast-element';
+import { html, slotted } from '@ni/fast-element';
 import type { ChatConversation } from '.';
 
 /* eslint-disable @typescript-eslint/indent */
 // prettier-ignore
 export const template = html<ChatConversation>`
 <div class="messages"><slot></slot></div>
-<div class="input"><slot name="input"></slot></div>`;
+<div class="input ${x => (x.inputIsEmpty ? '' : 'has-content')}">
```
rajsite, line 8: Would be nice if there wasn't a name change + logic inversion between these, seems like unnecessary mental hoops
```suggestion
<div class="input ${x => (x.inputEmpty ? 'input-empty': '')}">
```

packages/spright-components/src/chat/conversation/styles.ts
jattasNI: Nice idea, I went with "image" as the name for the new type. It was pretty easy to plumb it through and render it in the theme-aware tokens story (though it's hard to see the gradient at such a small size).

I ended up using `string` as the type. There are [some native types](https://developer.mozilla.org/en-US/docs/Web/API/CSSImageValue) to represent this concept but it didn't look like I could create an instance of a linear gradient object in that type system so we pass around CSS strings like we do on some of our other token types.

packages/nimble-components/src/theme-provider/design-token-comments.ts
jattasNI: I spoke with Brandon and he suspects we will use the gradient in other places but didn't have specific examples in mind. His best prediction for another use case was "IDE backgrounds" but I don't think that terminology aligns with Nimble component names.

I debated for a while and landed on `sectionBackgroundImage`. This is aligned with the existing `sectionBackgroundColor`. I wish we could offer better guidance about "what is a section" and "when should I use the section background image vs the section background color" but realistically the answer for these (and many similar token questions) is "use as directed by a visual designer". Since that's very generic advice I decided not to include it in the docs.

My vote is to stick with this (or a similarly very generic name). My second choice is to go back to "chat-conversation" and wait for more use cases.

change/@ni-nimble-components-c52231dc-0a44-4eed-8cba-f64e61b0ae15.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: nit: PR description needs an update

packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -18,8 +22,19 @@ export const styles = css`
         flex-direction: column;
         justify-content: flex-start;
         row-gap: 32px;
-        padding: ${mediumPadding};
-        background: ${applicationBackgroundColor};
+        padding: ${mediumPadding} ${standardPadding} ${mediumPadding}
+            ${standardPadding};
+        border: ${borderWidth} solid ${applicationBackgroundColor};
```
rajsite, line 27: playing in devtools it seems odd to me that the border is just around the the top half of the control (messages) instead of the full control. Should this be on `:host` instead?

packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -18,8 +22,19 @@ export const styles = css`
...
+        background: ${sectionBackgroundImage};
         overflow-y: auto;
     }
+
+    .input {
+        padding: ${mediumPadding} ${standardPadding} ${standardPadding}
```
rajsite, line 33: mentioned in sb comment but the top padding here ends up clipping messages in a weird way / in mid air

jattasNI: I chatted with Brandon and we decided to shrink that padding down to 1px for now.

image

change/@ni-nimble-components-c52231dc-0a44-4eed-8cba-f64e61b0ae15.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
jattasNI, line 2: Placeholder comment to block completion. These styling changes will sort of mess up the chat appearance in SLE so I want to wait to merge this until I can block the Renovate update and submit the PR to fix that, probably next week.

[Pull Request 1003566](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/1003566): Remove Nigel styling that Spright now provides

PR #2642, opened 2025-06-30
Delegate focus to chat input when other area is clicked

jattasNI:
# Pull Request

## 🤨 Rationale

On the spright chat input component I found myself clicking the area below the text area in an attempt to give it focus. But that had no effect because only the text area could receive focus.

## 👩‍💻 Implementation

Configure the chat input to delegate focus to its first focusable element (the text area). [Several other components](https://github.com/search?q=repo%3Ani%2Fnimble%20delegatesfocus&type=code) do the same thing.

## 🧪 Testing

I didn't see any test coverage for this codepath for other components so I propose not writing an auto test for this one.

I'll manually confirm on several browsers that I can click anywhere in the chat input to give it focus.

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/chat/input/index.ts
+++ b/packages/spright-components/src/chat/input/index.ts
@@ -107,7 +107,10 @@ export class ChatInput extends FoundationElement {
 const sprightChatInput = ChatInput.compose({
     baseName: 'chat-input',
     template,
-    styles
+    styles,
+    shadowOptions: {
+        delegatesFocus: true
+    }
 });
 
 DesignSystem.getOrCreate().withPrefix('spright').register(sprightChatInput());
--- a/change/@ni-spright-components-50182611-73fb-4bf7-86a9-153c03644f22.json
+++ b/change/@ni-spright-components-50182611-73fb-4bf7-86a9-153c03644f22.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Delegate focus to input when chat is clicked",
+  "packageName": "@ni/spright-components",
+  "email": "jattasNI@users.noreply.github.com",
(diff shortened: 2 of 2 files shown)


packages/spright-components/src/chat/input/index.ts
```diff
@@ -107,7 +107,10 @@ export class ChatInput extends FoundationElement {
 const sprightChatInput = ChatInput.compose({
```
rajsite, line 107: The [focusDelegates docs](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#focus-delegation) have some discussion around tabindex behavior when using `focusDelegates: true`. Maybe we should include that as well?

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -268,4 +268,25 @@ describe('ChatInput', () => {
             expect(page.getSendButtonTextContent()).toEqual('Send it!');
         });
     });
+
+    describe('tabindex', () => {
```
rajsite, line 272: Our other tests on tabindex are covering three cases (default, set to value, set and then clear value): https://github.com/ni/nimble/blob/55a4ce35d7064b9bae809825b4f72aaacab05fdf/packages/nimble-components/src/button/tests/button.spec.ts#L15-L51

packages/spright-components/src/chat/input/index.ts
rajsite: Think the docs should have an update to describe these manual test cases? I think the current wording of "will work properly" vs "won't" isn't clear enough to validate implementation.

rajsite (approved):

PR #2645, opened 2025-07-01
Initial working graph prototype.

atmgrifter00:
# Pull Request

## 🤨 Rationale

This is just a dummy draft PR to allow early comments.

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/fast-graph/index.ts
+++ b/packages/spright-components/src/fast-graph/index.ts
@@ -1,0 +1,108 @@
+/* eslint-disable import/extensions */
+/* eslint-disable import/no-extraneous-dependencies */
+import { observable } from '@ni/fast-element';
+import { DesignSystem, FoundationElement } from '@ni/fast-foundation';
+import 'flot/source/jquery';
+import 'flot/lib/jquery.event.drag.js';
+import 'flot/lib/jquery.mousewheel.js';
+import 'flot/source/jquery.canvaswrapper.js';
+import 'flot/source/jquery.colorhelpers.js';
+import 'flot/source/jquery.flot.js';
+import 'flot/source/jquery.flot.saturated.js';
+import 'flot/source/jquery.flot.browser.js';
+import 'flot/source/jquery.flot.drawSeries.js';
+import 'flot/source/jquery.flot.uiConstants.js';
+import 'flot/source/jquery.flot.logaxis.js';
+import 'flot/source/jquery.flot.symbol.js';
+import 'flot/source/jquery.flot.flatdata.js';
+import 'flot/source/jquery.flot.navigate.js';
+import 'flot/source/jquery.flot.fillbetween.js';
+import 'flot/source/jquery.flot.stack.js';
+import 'flot/source/jquery.flot.touchNavigate.js';
+import 'flot/source/jquery.flot.hover.js';
+import 'flot/source/jquery.flot.touch.js';
+import 'flot/source/jquery.flot.time.js';
+import 'flot/source/jquery.flot.axislabels.js';
+import 'flot/source/jquery.flot.selection.js';
+import { styles } from './styles';
+import { template } from './template';
+import type { AxisAutoScale } from './types';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'spright-fast-graph': FastGraph;
+    }
+}
+
+interface Flot {
+    setData: (data: unknown) => void;
+    setupGrid: (show: boolean) => void;
+    draw: () => void;
+    clearTextCache: () => void;
+    resize: () => void;
+}
+
+interface FlotConfig {
+    yaxis: { min?: number, max?: number, autoScale: AxisAutoScale, color?: string };
+    xaxis: { autoScale: AxisAutoScale, color?: string };
+    zoom: { interactive: boolean };
+    pan: { interactive: boolean };
+}
+
+interface FlotFactory {
+    plot(placeholder: HTMLElement, data: unknown[], config: FlotConfig): Flot;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
+const $ = ((window as any).$ || (window as any).jQuery) as FlotFactory;
+
+/**
+ * A Spright demo component (not for production use)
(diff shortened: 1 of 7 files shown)


packages/spright-components/src/fast-graph/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import 'flot/source/jquery.flot.flatdata.js';
+import 'flot/source/jquery.flot.navigate.js';
+import 'flot/source/jquery.flot.fillbetween.js';
+import 'flot/source/jquery.flot.stack.js';
+import 'flot/source/jquery.flot.touchNavigate.js';
+import 'flot/source/jquery.flot.hover.js';
+import 'flot/source/jquery.flot.touch.js';
+import 'flot/source/jquery.flot.time.js';
+import 'flot/source/jquery.flot.axislabels.js';
+import 'flot/source/jquery.flot.selection.js';
```
rajsite, line 26: Can we move these to a separate file like `flot-imports.js`.

And then similar organization but it exports a [noConflict](https://api.jquery.com/jQuery.noConflict/) version of jQuery that does not pollute the global scope [like we did in WebVIs](https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FVI%2FHtmlControls.Design%2FWeb%2FFramework%2FniGlobals.jQuery.js&_a=contents&version=GBmain).

So in the end would look like:
```js
import 'flot/source/jquery';
import 'flot/lib/jquery.event.drag.js';
import 'flot/lib/jquery.mousewheel.js';
import 'flot/source/jquery.canvaswrapper.js';
// ...
import 'flot/source/jquery.flot.selection.js';
export const jQueryWithPlugins = $.noConflict(true);
```

Having a separate file is at a minimum less noisy here and can encapsulate types for jquery in the that file. We also get to avoid dealing with the global variables outside the file.

packages/spright-components/src/fast-graph/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    @observable
+    public flotElement!: HTMLElement;
+
+    private flot!: Flot;
+    private resizeObserver?: ResizeObserver;
+    private data: unknown[] = [{ data: [], flatdata: true }];
+
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        this.flot = $.plot(this.flotElement, [], { yaxis: { autoScale: 'loose' }, xaxis: { autoScale: 'exact' }, zoom: { interactive: true }, pan: { interactive: true } });
```
rajsite, line 72: Let's save off a jquery reference bound to this element [like we did in the webvi days](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/VI/HtmlControls.Design/Web/Elements/ni-tree.js&version=GBmain&_a=contents&line=112&lineStyle=plain&lineEnd=113&lineStartColumn=1&lineEndColumn=1) and then save off a flot instantiation if helpful:

```js
/* using the new jQueryWithPlugins import from above */
this.jqref = jQueryWithPlugins(this.flotElement);
this.flot = this.jqref.plot(/**/)

```

packages/spright-components/src/fast-graph/index.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable import/extensions */
+/* eslint-disable import/no-extraneous-dependencies */
+import { observable } from '@ni/fast-element';
+import { DesignSystem, FoundationElement } from '@ni/fast-foundation';
+import 'flot/source/jquery';
```
rajsite, line 5: This [jQuery version is crazy old](https://github.com/flot/flot/blob/84f3e5d5e997f2f7c0180314ccbc6175ad40cab7/source/jquery.js#L2) at jQuery v1.8.3. 

If we are being conservative, WebVIs were on version [jQuery 3.5.1](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/NodeApps/node_modules/%40ni-private/webvi-deps/package.json&version=GBmain&line=12&lineEnd=13&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), but the latest is [jQuery 3.7.1](https://www.npmjs.com/package/jquery) and they are very good about breaking changes so I suspect that would work fine.

Immediate future:

It may not be possible yet, but I think we should try and get flot to support the [jquery slim builds](https://blog.jquery.com/2023/08/28/jquery-3-7-1-released-reliable-table-row-dimensions/#:~:text=install%20jquery%403.7.1-,Slim%20build,-Sometimes%20you%20don%E2%80%99t) to rely on nicer modern apis instead and trim some of the bundle size we almost certainly don't need.

Further future:

The [jQuery v4 is in beta](https://blog.jquery.com/2024/07/17/second-beta-of-jquery-4-0-0/) and will support es modules but that's not out yet.

packages/spright-components/src/fast-graph/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    zoom: { interactive: boolean };
+    pan: { interactive: boolean };
+}
+
+interface FlotFactory {
+    plot(placeholder: HTMLElement, data: unknown[], config: FlotConfig): Flot;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
+const $ = ((window as any).$ || (window as any).jQuery) as FlotFactory;
```
rajsite, line 57: There are types for jquery we may be able to extend: https://www.npmjs.com/package/@types/jquery

packages/spright-components/src/fast-graph/index.ts
```diff
@@ -0,0 +1,134 @@
+/* eslint-disable import/extensions */
+/* eslint-disable import/no-extraneous-dependencies */
+import { observable } from '@ni/fast-element';
+import { DesignSystem, FoundationElement } from '@ni/fast-foundation';
+import 'flot/source/jquery';
```
rajsite, line 5: Another thought is if webpack is treating  jquery as an import (from discussion earlier) you may be able to do:
```
import jQuery from 'jquery';
window.jQuery = jQuery;
```

rajsite:
Closing the PR to move off the PR list, doesn't prevent comments, etc so can keep using for that

Issue #2599, opened 2025-05-08
Lighthouse command error during execution


rajsite:
# 🐛 Bug Report

See https://dev.azure.com/ni/DevCentral/_workitems/edit/3120123/

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

rajsite:
@fredvisser re-adding triage tag as this is also filled as an azdo bug and hit the 30 day mark, should decide the priority

rajsite:
Looks like `@lhci/cli` had a recent release with updated dependencies, likely can update and re-enable: https://www.npmjs.com/package/@lhci/cli

PR #2653, opened 2025-07-10
Dependency update July 2025

rajsite:
# Pull Request

## 🤨 Rationale

Dependency updates for July 2025, some interesting notes:
- Tiptap introduced an accidental breaking change by adding a new field to a type as `blah | null` instead of optional as `blah?`: https://github.com/ueberdosis/tiptap/pull/6300
  - This PR bumps the minimum supported version of tiptap to make sure the breaking type change and our adjustment is included.
- We have historically just aligned the playwright js and nuget package version numbers but turns out they don't really correlate and we have been mostly lucky.
  - With this upgrade when trying to upgrade to playwright js to v1.53.0 it would download chrome playwright build v1178 ([see log](https://github.com/ni/nimble/actions/runs/16226058033/job/45818119192#step:9:1228)):
    > Downloading Chromium 138.0.7204.15 (playwright build v1178) from https://cdn.playwright.dev/dbazure/download/playwright/builds/chromium/1178/chromium-linux.zip
  - However v1.53.0 of the nuget package expected chrome playwright build v1179 to be installed ([see log](https://github.com/ni/nimble/actions/runs/16226058033/job/45818119192#step:17:353)):
    > Microsoft.Playwright.PlaywrightException : Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium_headless_shell-1179/chrome-linux/headless_shell
  - Turns out the [playwright nuget package bundles a specific version of playwright-core js](https://github.com/microsoft/playwright-dotnet/blob/v1.53.0/src/Common/Version.props#L5) as the `DriverVersion` variable so we should be aligning those versions: 
    > `<DriverVersion>1.53.1</DriverVersion>`
- The apache arrow js library moved out of the apache arrow monorepo and is now versioned independently. They [gave great feedback](https://github.com/apache/arrow-js/discussions/204) to my questions about how versioning works and I feel mostly confident we don't actually need to pin the version anywhere, we can just use the latest versions of the libraries.
  - This PR does not unpin the apache arrow dependencies, it only allows them to differ between dotnet and js.
- A new version of lighthouse was released that seems to work. I had to bump down our accessibility test threshold to re-enable lighthouse though. This PR fixes https://github.com/ni/nimble/issues/2599.

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Rely on CI.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated manual tasks docs.

### DIFF
--- a/change/@ni-nimble-react-16326706-0ee6-4ff7-a7e8-8b67bc9fdc40.json
+++ b/change/@ni-nimble-react-16326706-0ee6-4ff7-a7e8-8b67bc9fdc40.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update dependencies",
+  "packageName": "@ni/nimble-react",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@ni-spright-react-36120f9e-a5b3-46cf-9c29-4fae8b390c08.json
+++ b/change/@ni-spright-react-36120f9e-a5b3-46cf-9c29-4fae8b390c08.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update dependencies",
+  "packageName": "@ni/spright-react",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}


packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
```diff
@@ -149,7 +149,7 @@ describe('TableColumnNumberText', () => {
         );
     });
 
-    it('changing data from null to value displays value', async () => {
+    it('changing data from null to value displays value #SkipWebkit', async () => {
```
rajsite, line 152: Todo understand these failures better / capture as tech debt

.github/workflows/main.yml
```diff
@@ -43,9 +43,13 @@ jobs:
           fetch-depth: 0
 
       # Install dependencies
+      - name: List all current packages
+        run: sudo apt --installed list
```
rajsite, line 47: todo remove these debugging logs here and a couple lines down

.github/workflows/main.yml
```diff
@@ -43,9 +43,13 @@ jobs:
...
       - name: Remove pre-existing dotnet sdk 8 installs
-        run: sudo apt remove --purge dotnet-sdk-8.0
+        run: sudo apt remove --purge '^dotnet-.*' '^aspnetcore-.*'
```
rajsite, line 49: This change wasn't ultimately necessary but is more complete. Should share on the MS github issue around this stuff

.github/workflows/main.yml
```diff
@@ -100,10 +104,7 @@ jobs:
 
       # Run Lighthouse audit (from any push)
       - name: Performance Audit with Lighthouse CI
-        # if: env.HAS_LHCI_GITHUB_APP_TOKEN == 'true' && github.event_name == 'push'
-        # Disabled due to errors during execution on CI
-        # See: https://github.com/ni/nimble/issues/2599
-        if: false
+        if: env.HAS_LHCI_GITHUB_APP_TOKEN == 'true' && github.event_name == 'push'
```
rajsite, line 107: should mark the related issue fixed with this pr

packages/nimble-components/src/table-column/number-text/tests/table-column-number-text.spec.ts
rajsite: Captured as tech debt

.github/workflows/main.yml
```diff
@@ -43,8 +43,8 @@ jobs:
           fetch-depth: 0
 
       # Install dependencies
```
rajsite, line 45: Todo: bump the fast packages, see: https://github.com/ni/nimble/pull/2659

rajsite: done

rajsite:
@jattasNI besides one trivial source change the rest of the PR is finagling with dependency versions. Gonna bypass merge but (someone 🌴) can address changes in a follow-up :P

PR #2666, opened 2025-07-24
Spright `ChatInput` remove `resetInput`

atmgrifter00:
# Pull Request

## 🤨 Rationale

The design for the chat input calls to _always_ clear the text in the input once a user sends the message. There does not seem to be a compelling enough reason to make the client call a method in response to the `send` event to make this happen ([additional context](https://github.com/ni/nimble/pull/2611#discussion_r2226714997)).

## 👩‍💻 Implementation

Make `resetInput` private and call it from the `sendButtonClickHandler`.

## 🧪 Testing

Unit test.

## ✅ Checklist

### DIFF
--- a/change/@ni-spright-components-4795f510-5b35-4ec4-b04d-6935fca41947.json
+++ b/change/@ni-spright-components-4795f510-5b35-4ec4-b04d-6935fca41947.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Remove resetInput event and auto-clear text on send for chat input.",
+  "packageName": "@ni/spright-components",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/spright-components/src/chat/input/tests/chat-input.spec.ts
+++ b/packages/spright-components/src/chat/input/tests/chat-input.spec.ts
@@ -205,53 +205,6 @@ describe('ChatInput', () => {
         });
     });
 
-    describe('resetInput method', () => {
-        beforeEach(async () => {
-            await connect();
-        });
-
-        it('clears the input contents, sets focus, and disables the button', () => {
-            page.setText('new value');
-            element.resetInput();
-            processUpdates();
-
-            expect(element.value).toEqual('');
-            expect(page.getRenderedText()).toEqual('');
-            expect(page.textAreaHasFocus()).toBeTrue();
-            expect(page.isSendButtonEnabled()).toBeFalse();
-        });
-
-        it('can be called from send event handler', async () => {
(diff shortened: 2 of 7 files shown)


atmgrifter00:
> @atmgrifter00 some related tests are failing, gonna wait for you to resolve before reviewing

Done.

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -176,6 +170,17 @@ describe('ChatInput', () => {
             expect(spy).toHaveBeenCalled();
         });
 
+        it('clears the input contents, sets focus, and disables the button', () => {
```
jattasNI, line 173: Could you add a test that calls `focus()` on the chat-input and verifies that the text area receives focus? I think it will be common for apps to want to start out in that state. Previously they could call `resetInput()` to guarantee it but now calling `focus()` would be the way to do it. (They might have tried calling `focus()` before too and it would work, but we didn't have test coverage for it).

change/@ni-spright-components-4795f510-5b35-4ec4-b04d-6935fca41947.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
jattasNI, line 2: Will you update this usage or do you want me to?

https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/nigel/src/components/sl-nigel/sl-nigel.component.ts&version=GBmaster&line=31&lineEnd=32&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -205,53 +210,6 @@ describe('ChatInput', () => {
...
-            element.resetInput();
-            processUpdates();
-
-            expect(element.value).toEqual('');
-            expect(page.getRenderedText()).toEqual('');
-            expect(page.textAreaHasFocus()).toBeTrue();
-            expect(page.isSendButtonEnabled()).toBeFalse();
-        });
-
-        it('can be called from send event handler', async () => {
```
jattasNI, line 224: Think it would be worth adding an assert to one or more of the `send` event tests to assert the state of `element.value` when the event is fired? Basically asserting that we reset it before firing the event, not after.

jattasNI (approved):

jattasNI:
@rajsite OOO so bypassing review requirement to merge. Not sure why my review was insufficient to meet the requirement.

Issue #413, opened 2022-03-07
nimble-chip or nimble-pill component


jattasNI:
## 😯 Problem to Solve

The [Argon visual design](https://xd.adobe.com/view/ee317991-74e3-4c8f-62aa-dd7512dc7951-9800/screen/31177479-494b-4bdb-be24-4e4b1e78047a/) shows chips/pills that display a short snippet of text within a colored region with rounded corners.
![image](https://user-images.githubusercontent.com/10500124/157135311-c0c76d50-e69e-4d47-81f2-3fc042cde0b5.png)

Argon is [planning to implement this imminently](https://dev.azure.com/ni/DevCentral/_git/srd-semi-ses-software/pullrequest/274367?_a=files&path=/Argon/Docs/Design/Workflows/Dataspaces/QueryResult.md) so we might not be able to deliver it in time for their use case but I'm creating this issue to gather requirements and use cases.

SystemLink also shows chips in the [sl-details-panel.component](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/src/sl-details-panel/sl-details-panel.component.html&version=GBmaster&line=47&lineEnd=48&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

Web Portal also shows tags using the [DevExtreme Tag Box](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxTagBox/), which lets you select tags from a drop down and then shows the selected items in a list of chips.
![image](https://user-images.githubusercontent.com/10500124/158455254-97dc6bb1-9ecb-4d37-b7c2-beb3afee484f.png)

## 💁 Proposed Solution

We might want to also provide a component that stores a collection of chips. It could handle deleting a chip when the `x` icon is clicked and could provide a default affordance if there are too many chips to fit in available space.

We might want to provide a slot to display an icon to the left of the chip [as Material does](https://material.io/components/chips#usage).
![image](https://user-images.githubusercontent.com/10500124/157135934-754373d5-e64e-4502-aec5-3808c22557af.png)

We might want to allow different colored chips as GitHub labels do
![image](https://user-images.githubusercontent.com/10500124/157135999-049cbfb3-667d-4b90-83f1-36969b15be9a.png)

Longer term there are lots of potential interactions with chips. Some examples from [Material](https://material.io/components/chips):
- dragging chips to move them around
[...]

PR #2667, opened 2025-07-24
Nimble Chip HLD

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #413

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A


packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+- [Spright chat feature backlog](https://github.com/ni/nimble/issues/2610) (component for attachment chips)
+
+### Containing Library
+
+The intent is for this to be put in Nimble, as 1) this seems to be a fairly constrained component, and thus easily ported to any needed UI platform, and 2) it seems to have broad use possibilities.
+
+### Non-goals
+
+- Arbitrary HTML content
+- Implementing a "chip container" component that would be responsible for laying out the chips in a particular fashion, as well as possibly managing their removal if the "remove" button on a chip is pressed. While this may be useful at some point it is not necessary for the `Chip` component to have value.
```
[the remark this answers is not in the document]
atmgrifter00: Are answers to these meant to be captured in this HLD, or just as part of the documentation for the component in Storybook?

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+---
+
+## Design
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
```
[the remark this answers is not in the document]
atmgrifter00: [Infragistics also calls it a chip](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/chip#usage).

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+N/A
+
+### Documentation
+
+Storybook stories will be created.
+
+---
+
+## Open Issues
```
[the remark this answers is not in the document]
atmgrifter00: I've added some open issues.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+    - `prevent-remove` - set to hide the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `error-visible`
+- _Methods_
+- _Events_
+    - `remove` - fired when the chip remove button is pressed (or \<Del\> or \<Backspace\> is pressed), and `prevent-remove` hasn't been set.
+- _Slots_
+    - `start` - content placed to the left of the chip text
+    - (default) - for the primary message text
+- _CSS Classes and CSS Custom Properties that affect the component_
```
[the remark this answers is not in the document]
atmgrifter00: No, I'm not expecting to provide an API to allows clients to change chip presentation in these ways. Just explicitly saying it's "not supported" now.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `prevent-remove` - set to hide the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `error-visible`
+- _Methods_
+- _Events_
+    - `remove` - fired when the chip remove button is pressed (or \<Del\> or \<Backspace\> is pressed), and `prevent-remove` hasn't been set.
```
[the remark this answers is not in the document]
atmgrifter00: Correct. I've added that statement to the HLD.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+    "Option Label"
+</nimble-chip>
+```
+
+![](./spec-images/Chip-checked.PNG)
+
+### Anatomy
+
+```html
+<template aria-disabled="${x => x.disabled}">
```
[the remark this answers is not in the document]
atmgrifter00: Added to API.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Containing Library
+
+The intent is for this to be put in Nimble, for the following reasons:
+
+- This is a fairly constrained component, and thus easily ported to any needed UI platform.
+- It has broad use possibilities.
+- Designs for the component have been completed (see above link)
+
+However, there is a potential for there to be some accessibility concerns, notably around the remove button, that could persuade us to put this in Spright.
```
mollykreis, line 21: I'd expect those accessibility concerns to be hashed out within this HLD rather. That would allow this to be a more concrete statement of "this will be put in nimble". I think it'll cause a lot of churn for us to move into development without knowing whether or not there are too many accessibility problems for this component to live in nimble.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+- Arbitrary HTML content
+- Implementing a "chip container" component that would be responsible for laying out the chips in a particular fashion, as well as possibly managing their removal if the "remove" button on a chip is pressed. While this may be useful at some point it is not necessary for the `Chip` component to have value.
+
+### Features
+
+- Optionally removable
+    - Note that the `Chip` won't remove itself, and it will be up to the client to remove it from the DOM.
+- Allows slotted content to be placed to left of text content. Slotted content will be styled automatically to be appropriately sized for the chip (i.e. given a height of `controlSlimHeight`). Types of slotted content that will be documented as acceptible are:
+    - icons
+    - buttons with `content-hidden` set and `appearance` set to `ghost`.
```
mollykreis, line 34: What's the use-case for having buttons slotted in the `start` slot? Are there designs? I can't think of any other control (though may be forgetting one) that allows putting a button in the `start` slot.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Design
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `prevent-remove` - set to hide the remove button
```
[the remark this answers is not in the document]
mollykreis: FWIW, I like `removable` as well.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+            <nimble-button content-hidden @click="${x => x.handleRemoveClick()}">
+                <nimble-icon-times>
+                </nimble-icon-times>
+            </nimble-button>
+        `)}
+    </div>
+</template>
+```
+
+While the template will have a named `end` slot, this will not be documented, and only be populated with the error icon as needed, following existing Nimble patterns.
```
mollykreis, line 122: I don't think the statement about the error icon is needed anymore since you've gotten rid of `error-visible` on the chip. I think it would also be reasonable to change this statement to say we may consider using the `end` slot in the future if/when we add an error state to the chip.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
+# Nimble Chip
+
+## Overview
+
+The `nimble-chip` is a component that presents text, an optional "remove" button, optional slotted content to the left of the text, and an optional error icon.
```
mollykreis, line 5: I think this overview section shouldn't mention the optional error icon since that's being considered out of scope for the initial creation of the component.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+        ${startSlotTemplate(context, definition)}
+        <span class="content" part="content">
+            <slot></slot>
+        </span>
+        ${endSlotTemplate(context, definition)}
+        ${when(x => !x.preventRemove, html<Chip>`
+            <nimble-button content-hidden @click="${x => x.handleRemoveClick()}">
+                <nimble-icon-times>
+                </nimble-icon-times>
+            </nimble-button>
```
mollykreis, line 116: nit: This could be improved a little

```suggestion
            <nimble-button content-hidden @click="${x => x.handleRemoveClick()}" title="Remove">
                <nimble-icon-times slot="start">
                </nimble-icon-times>
                Remove
            </nimble-button>
```

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+_Consider the accessibility of the component, including:_
+
+- _Keyboard Navigation and Focus_
+    - the chip component is focusable, as is the remove button (when present).
+- _Form Input_
+    - N/A
+- _Use with Assistive Technology_
+    - a `chip`'s accessible name comes from the element's contents by default
+    - `aria-disabled` will be set whenever the `disabled` attribute is set
+    - the remove button will have an `aria-label` of "remove", which we will provide a label provider token for.
```
mollykreis, line 155: I'm not sure the button needs an `aria-label` explicitly set on it. Instead, the content of the button should be set to "remove" (from the label provider), as should the `title` of the button.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Accessibility
+
+_Consider the accessibility of the component, including:_
+
+- _Keyboard Navigation and Focus_
+    - the chip component is focusable, as is the remove button (when present).
+- _Form Input_
+    - N/A
+- _Use with Assistive Technology_
```
mollykreis, line 152: Do we need to consider the `role` of the chip? It would be good to make a statement about what it will be (even if we don't need to explicitly set it in the template).

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
jattasNI: I'm OK to leave the details to the Storybook docs. For the HLD you can just mention that we'll do it. And reminder to start the conversation with VxD if those questions aren't already answered in the Figma spec.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
jattasNI: `removable` aligns well with `clearable` on the select. And I don't have a good sense for whether showing or hiding it by default is more common so I'm ok to pick one based on what others do. Let's go with `removable`.

The only other precedence I could find is `prevent-dismiss` on banner/dialog/drawer. That would lead to `prevent-remove` or `allow-remove`. I'd be fine with those too but they sound a bit clunky.

packages/nimble-components/src/chip/specs/README.md
jattasNI: In Material they appear to append the chip label to the remove button label. e.g. "remove myAttachment". I can see a benefit to that if there are lots of chips and the user wants to quickly distinguish between their remove buttons. It could be tedious to tab through them all and figure out whether the "remove" applies to the previous label tab stop or the next one.

packages/nimble-components/src/chip/specs/README.md
atmgrifter00: Yeah, I don't think there is a use-case for this. Ultimately, if anything, we will make the chip itself interactive (and potentially an actual button). Updated.

packages/nimble-components/src/chip/specs/README.md
atmgrifter00: Added a statement asserting that we don't need to set the `role`. I'm not going to die on this hill though. It seems if we do set it it would be to `status` (until we make the chip component interactive).

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `prevent-remove` - set to hide the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `disabled`
```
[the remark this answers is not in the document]
atmgrifter00: Updated.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: Updated to use `removable`.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: Updated.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: Oops...sorry. Updated (spoiler...the chip handles it).

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+- [Nimble Chip component issue](https://github.com/ni/nimble/issues/413)
+- [Chip design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=9269-229459&m=dev)
+- [Spright chat feature backlog](https://github.com/ni/nimble/issues/2610) (component for attachment chips)
+
+### Containing Library
+
+The intent is for this to be put in Nimble, for the following reasons:
+
+- This is a fairly constrained component, and thus easily ported to any needed UI platform.
```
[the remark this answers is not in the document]
atmgrifter00: Yes, and I updated the language to say basically that.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+- This is a fairly constrained component, and thus easily ported to either Angular or Blazor.
+- It has broad use possibilities.
+- Designs for the component have been completed (see above link)
+- While the remove button introduces a new interactive pattern for our controls from an accessibility perspective, the described plan seems in-line with ARIA expectations.
+
+### Non-goals
+
+- Arbitrary HTML content
+- Implementing a "chip container" component that would be responsible for laying out the chips in a particular fashion, as well as possibly managing their removal if the "remove" button on a chip is pressed. While this may be useful at some point it is not necessary for the `Chip` component to have value.
+- Usage guidance will be provided in the Storybook.
```
mollykreis, line 26: > Usage guidance will be provided in the Storybook.

Is this a comment associated with the bullet point directly above? If so, perhaps the sentence should be added to the above bullet point or as a sub-bullet.

As-is, it's unclear how saying you'll add usage documentation in storybook is considered a non-goal.

packages/nimble-components/src/chip/specs/README.md
mollykreis: @atmgrifter00 - can you remove the `aria-label` from the button in the template within the "Anatomy" section?

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+### States
+
+The CSS `hover` and `focusVisible` states will have specific stylings (as seen in the UX designs), in addition to styling capturing the `disabled` attribute state.
+
+### Accessibility
+
+_Consider the accessibility of the component, including:_
+
+- _Keyboard Navigation and Focus_
+    - the chip component is focusable, as is the remove button (when present).
```
mollykreis, line 148: Why does the chip need to be focusable? Would it make more sense for the chip not to be focusable (and not have keyboard interactions associated with it) and just have the remove button be focusable? That would be more aligned with the nimble-banner.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Documentation
+
+Storybook stories will be created.
+
+---
+
+## Open Issues
+
+- Should we have a `removable` attribute instead of `prevent-remove` to invert the API state? This is the approach Infragistics takes with their component.
```
mollykreis, line 200: Should this open issue be removed?

packages/nimble-components/src/chip/specs/README.md
atmgrifter00: Yeah, I've just removed this line.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: Stating explicitly that the `start` slot is meant for an icon.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+### Non-goals
+
+- Arbitrary HTML content
+- Implementing a "chip container" component that would be responsible for laying out the chips in a particular fashion, as well as possibly managing their removal if the "remove" button on a chip is pressed. While this may be useful at some point it is not necessary for the `Chip` component to have value.
+
+### Features
+
+- Optionally removable
+    - Note that the `Chip` won't remove itself, and it will be up to the client to remove it from the DOM.
+- Allows icons to be slotted left of text content. Slotted content will be styled automatically to be appropriately sized for the chip (i.e. given a height of `controlSlimHeight`)
```
mollykreis, line 31: I don't think the comment about giving slotted content a height of `controlSlimHeight` makes sense here since it sounds like we're really only supporting icons in that slot (which will likely be smaller than `controlSlimHeight`).

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `removable` - set to show the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `disabled` - styles the chip in the typical Nimble manner, including any slotted content, and prevents all interaction with it including being able to focus it
+- _Methods_
```
mollykreis, line 67: It would probably make sense to remove the `Methods` bullet point since there aren't any methods (or explicitly put "none" under it)

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+- _Events_
+    - `remove` - fired when the chip remove button is pressed (or \<Del\> or \<Backspace\> is pressed), and `removable` has been set.
```
mollykreis, line 69: Is the plan to have Delete & Backspace remove the chip if the chip isn't focusable? Is this remaining from when the chip was going to be focusable, or is the idea that we'd handle those keys on the remove button?

atmgrifter00: Yeah, I think this was from when the chip was also meant to be focusable. I'm not sure having the \<Del\> and \<Backspace\> buttons do anything is really needed if the user can only perform this action when focusing the delete button (where \<Enter\> is the expected action).

Will remove.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Documentation
+
+Storybook stories will be created.
+
+---
+
+## Open Issues
+
+Designs call for there to be a max width of 284 pixels for the text portion of the chip. Do we need to create a token for this?
```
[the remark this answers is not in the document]
atmgrifter00: @NIbokeefe - I think your input here would be useful.

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Design
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `removable` - set to show the remove button
```
rajsite, line 64: We use `prevent-dismiss` on both the dialog (prevents esc key behavior) and banner (hides the `x` to close the banner). Can we use the same here?

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+    - `appearance` - supports `outline` and `block` appearances
+    - `disabled` - styles the chip in the typical Nimble manner, including any slotted content, and prevents being able to focus and/or interact with the remove button.
+- _Methods_
+    - _None_
+- _Events_
+    - `remove` - fired when the chip remove button is pressed.
```
rajsite, line 70: If we go with `prevent-dismiss` then I guess here we could change the name to `dismiss`

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `removable` - set to show the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `disabled` - styles the chip in the typical Nimble manner, including any slotted content, and prevents being able to focus and/or interact with the remove button.
```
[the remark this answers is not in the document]
atmgrifter00: I guess the corollary to this is that the remove button should _not_ be shown when disabled (just like the 'clearable' button for the Select)?

packages/nimble-components/src/chip/specs/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+### API
+
+_The key elements of the component's public API surface:_
+
+- _Component Name_: `nimble-chip`
+- _Props/Attrs_:
+    - `removable` - set to show the remove button
+    - `appearance` - supports `outline` and `block` appearances
+    - `disabled` - styles the chip in the typical Nimble manner, including any slotted content, and prevents all interaction with it including being able to focus it
```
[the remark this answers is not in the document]
atmgrifter00: Based on some of the conversation it seems that `disabled` should also hide the remove button. I've updated this sentence to reflect that.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
atmgrifter00: @jattasNI, my reaction to this conversation is to not bother with a token for this at the moment.

packages/nimble-components/src/chip/specs/README.md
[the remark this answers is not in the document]
rajsite: Yea, the control can manipulate the components it owns on disable (chips own the remove button in the shadowroot), it's just the weirdness of manipulating components the control doesn't own that are slotted (user owns configuration of slotted stuff).
We can communicate things to slotted content by things like custom properties (which is how icon color is communicated).

PR #2668, opened 2025-07-28
Collapse breadcrumbs to dropdown when too narrow

m-akinc:
# Pull Request

## 🤨 Rationale

TODO

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/breadcrumb/template.ts
+++ b/packages/nimble-components/src/breadcrumb/template.ts
@@ -1,0 +1,50 @@
+import {
+    elements,
+    html,
+    ref,
+    repeat,
+    slotted,
+    when,
+    type ViewTemplate
+} from '@ni/fast-element';
+import type { FoundationElementTemplate } from '@ni/fast-foundation';
+import type { Breadcrumb } from '.';
+import { breadcrumbItemTag, type BreadcrumbItem } from '../breadcrumb-item';
+import { menuButtonTag } from '../menu-button';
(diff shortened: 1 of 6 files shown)


packages/nimble-components/src/breadcrumb/template.ts
```diff
@@ -0,0 +1,50 @@
...
+            ${when(x => x.collapseState !== CollapseState.none, html<Breadcrumb>`
+                <${breadcrumbItemTag}>
+                    <${menuButtonTag}
+                        appearance="ghost"
+                        content-hidden
+                        class="collapsed-items-button"
+                    >
+                        <${iconThreeDotsLineTag} slot="start"></${iconThreeDotsLineTag}>
+                        <${menuTag} slot="menu">
+                            ${repeat(x => x.collapsedItems, html<BreadcrumbItem>`
```
jattasNI, line 39: I believe the main technical challenge with this approach is the idea of copying contents from slotted elements into new elements. This could cause problems if clients are expecting the elements they slotted to be present for various reasons: styles that target them, tests that expect them, metadata they added to the element, etc. It gets even more risky if its a framework that did one of those things outside the client's control.

As we mentioned at standup, it's probably worth a brainstorming meeting on this if you want to move forward with this approach. I agree that it is a better UX than #2669 (if we can figure out the flickering).

One idea to seed for that meeting: treat breadcrumb items as pure configuration elements (similar to table columns), not presentation elements. Then the actual presentation could swap between visible links or menu items.

PR #2689, opened 2025-08-11
Nimble chip

atmgrifter00:
# Pull Request

## 🤨 Rationale

- #413 
- #2610 

Creating the `nimble-chip` component for the immediate requirement of adding attachments to the `spright-chat-input` component.

## 👩‍💻 Implementation

Typical Nimble implementation. Adding a new label provider for the 'Remove' text for the remove button.

## 🧪 Testing

Unit and chromatic tests.

## ✅ Checklist


packages/nimble-components/src/label-provider/core/index.ts
```diff
@@ -75,6 +77,9 @@ export class LabelProviderCore
     @attr({ attribute: 'scroll-forward' })
     public scrollForward: string | undefined;
 
+    @attr({ attribute: 'chip-remove' })
+    public chipRemove: string | undefined;
```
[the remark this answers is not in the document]
jattasNI: Never mind, I think I just missed it when I first looked.

packages/nimble-components/src/chip/template.ts
```diff
@@ -0,0 +1,48 @@
...
+        >
+            <slot
+                ${ref('contentSlot')}
+                ${slotted({ property: 'content' })}
+            ></slot>
+        </span>
+        ${when(
+        x => x.removable,
+        html<Chip>`
+                <${buttonTag}
```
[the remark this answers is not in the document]
atmgrifter00: Fixed. I could swear I had made all these changes and committed them, but I think I'm getting mixed up switching between machines (meaning I didn't actually commit the changes from the other machine).

packages/storybook/src/nimble/chip/chip.mdx
```diff
@@ -0,0 +1,26 @@
...
+<Canvas of={chipStories.chip} />
+
+## API
+
+<Controls of={chipStories.chip} />
+<ComponentApisLink />
+
+{/* ## Styling */}
+
+{/* ## Usage */}
```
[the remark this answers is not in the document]
atmgrifter00: I'll check with Brandon.

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+    ${display('inline-flex')}
+
+    :host {
+        height: ${controlHeight};
+        width: fit-content;
+        max-width: 300px;
+        color: ${bodyFontColor};
+        font: ${bodyFont};
+        padding: 0 ${mediumPadding};
+        gap: 4px;
```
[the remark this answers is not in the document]
atmgrifter00: @NIbokeefe, just bringing your attention to this comment. In the design spec, the spacing between the icon and the test, and the text and the remove button currently has this spacing disparity (which I think is just due to the nature of the button, which is 16x16):
image

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+        background-color: transparent;
+        border: ${borderWidth} solid rgba(${actionRgbPartialColor}, 0.3);
```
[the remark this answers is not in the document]
atmgrifter00: This is what the button uses for its [outline appearance](https://github.com/ni/nimble/blob/f8dd3dc7f440aaf6d6a6560f723cace8f4cfdb9d/packages/nimble-components/src/patterns/button/styles.ts#L205).

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,110 @@
...
+        border-radius: 4px;
+        justify-content: center;
+        align-items: center;
+        background-repeat: no-repeat;
+        ${
+            /*
```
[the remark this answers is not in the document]
atmgrifter00: Yeah, I've now removed this. I don't think the use-case this was addressing for buttons applies to the chips. If there are issues related to this style not being applied it would likely be pathological.

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    .content {
+        text-overflow: ellipsis;
+        overflow: hidden;
+        white-space: nowrap;
+    }
+
+    .remove-button {
+        height: 16px;
```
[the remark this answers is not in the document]
mollykreis: We don't currently have a way for the touchpoint to be larger than the button, so we'd need to be fine with 16x16 touch point for now.

Also, we never added conditional styling to the button to change the focus-visible styling of the button when it got smaller (shouldn't have the double focus rings when small). Are we fine having the double ring on the 16x16 button until if/when that button styling gets updated?
image

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,98 @@
...
+        background-image: none;
+        background-size: 100% 100%;
+    }
+
+    :host([disabled])::before {
+        box-shadow: none;
+    }
+
+    :host([disabled]) slot[name='start']::slotted(*),
+    :host([disabled]) .remove-button {
```
mollykreis, line 52: Why do you need to set disabled styling on the remove button? Isn't the remove button hidden when the chip is disabled?

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,98 @@
...
+        height: ${iconSize};
+        width: ${iconSize};
+        margin-right: calc(-1 * ${smallPadding});
+    }
+`.withBehaviors(
+    appearanceBehavior(
+        ChipAppearance.block,
+        css`
+            :host {
+                background-image: linear-gradient(
```
mollykreis, line 82: Why did you need to set a background-image to achieve a solid background color?

packages/nimble-components/src/chip/tests/chip.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+import { Chip, chipTag } from '..';
+import { fixture, type Fixture } from '../../utilities/tests/fixture';
+import { ChipPageObject } from '../testing/chip.pageobject';
+import { waitForUpdatesAsync } from '../../testing/async-helpers';
+
+async function setup(): Promise<Fixture<Chip>> {
+    return await fixture<Chip>(html`<${chipTag}></${chipTag}>`);
+}
+
+describe('Chip', () => {
```
mollykreis, line 11: Can you add some tests for the tab index behavior?

packages/nimble-components/src/chip/template.ts
```diff
@@ -0,0 +1,44 @@
...
+            <slot
+                ${ref('contentSlot')}
+                ${slotted({ property: 'content' })}
+            ></slot>
+        </span>
+        ${when(x => x.removable && !x.disabled, html<Chip>`
+            <${buttonTag}
+                class="remove-button"
+                content-hidden
+                appearance="ghost"
```
mollykreis, line 35: Can you use `ButtonAppearance.ghost` here?

packages/storybook/src/nimble/chip/chip-matrix.stories.ts
```diff
@@ -0,0 +1,127 @@
...
+        'Long Label',
+        'A very long label that will likely overflow the chip and require truncation with set width'
+    ]
+] as const;
+type LabelState = (typeof labelStates)[number];
+const labelStatesOnlyShort = labelStates[0];
+
+const widthStates = [
+    ['Default Width', ''],
+    ['Custom Width', 'width: 250px;']
```
mollykreis, line 51: If you haven't already, can you confirm with @NIbokeefe that when the chip is wider than the default "fit-text" width, that the remove button should stick to the text rather than the edge of the chip?

image

FWIW - in my opinion, the remove button should always be right-aligned in the chip rather than floating in the middle next text. I also wonder if the text & icon should always be left aligned rather than centered.

packages/storybook/src/nimble/chip/chip-matrix.stories.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export const interactionsThemeMatrix: StoryFn = createMatrixThemeStory(
+    createMatrixInteractionsFromStates(component, {
+        hover: interactionStatesHover,
+        hoverActive: [],
+        focus: interactionStates,
+        active: []
+    })
+);
+export const textCustomized: StoryFn = createMatrixThemeStory(
```
mollykreis, line 125: The italic text looks clipped in the Text Customized story.

packages/storybook/src/nimble/chip/chip.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Usage
+
+When displaying multiple related chips together, the recommended configuration
+is:
+
+- chips are arranged horizontally and wrap to new lines when there is
+  insufficient horizontal space
+- all chips have the same appearance
+- use the medium-padding token to create space between chips
```
mollykreis, line 30: Assuming medium padding token is intended to be used for the horizontal and vertical spacing between chips, can you explicitly state that to avoid confusion?

packages/nimble-components/src/chip/styles.ts
atmgrifter00: This is what the button does. From what I can this is done to remove the border?

With `background-image`:
image

Using `background-color = `:
image

Design:
image

I'll try a different route, however, and just have a transparent border, so we don't see that fuzziness at the corners.

atmgrifter00: Fixed.

packages/nimble-components/src/chip/styles.ts
atmgrifter00: Yup. Removed.

packages/storybook/src/nimble/chip/chip-matrix.stories.ts
atmgrifter00: Yeah, it is, and I talked with @jattasNI about this, and it's unclear to me what's causing it. The `button` doesn't exhibit this behavior, and even if I align the styles for the chip and the button (for the content part), I can't get rid of the cut-off. Interested in suggestions here.

rajsite (approved):

packages/storybook/src/nimble/chip/chip-matrix.stories.ts
NIbokeefe: @atmgrifter00 @mollykreis,

I don't see the use case for wanting to expand a chip's width.  We do allow that with buttons. However, with chips anywhere in the short or long term they should be nested within a control and not a stand alone selection element that would require modified margins to fill spaces.

If it makes sense to build that functionality into the control regardless, I agree with @mollykreis, the text and icon should left align, and the close button should right align as shown in example "A."  We should prevent this from happening until we have a use case that warrants it.

Screenshot 2025-08-25 at 12 11 07 PM

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        justify-content: center;
+        align-items: center;
+        background-repeat: no-repeat;
+    }
+
+    :host([disabled]) {
+        cursor: default;
+        color: ${bodyDisabledFontColor};
+        border-color: rgba(${borderRgbPartialColor}, 0.3);
+        box-shadow: none;
```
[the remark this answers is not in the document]
NIbokeefe: @rajsite I can't currently think of a reason we would use a box shadow (elevation) on a chip to date.  Maybe a future iteration we use elevate a control with a box-shadow on rollover, but this would become a universal rule not just for chips alone.

packages/nimble-components/src/chip/styles.ts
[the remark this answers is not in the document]
NIbokeefe: @mollykreis In my spec I drew it up with one 2px ring just like a Rollover state.  Due to its small size and likelihood of covering the icon, I skipped the inner ring.  But, I don't mind you mockup with the two rings, I could go either way.

Screenshot 2025-08-25 at 1 32 43 PM

change/@ni-nimble-components-fa1ae624-6b7b-4035-b6bd-1418d8451b11.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @atmgrifter00 still intend to get this pushed through?

jattasNI: I pushed some changes to resolve many of the open threads.

packages/nimble-components/src/chip/styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        background-image: none;
+        background-size: 100% 100%;
+    }
+
+    :host([disabled])::before {
+        box-shadow: none;
+    }
```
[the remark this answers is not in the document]
jattasNI: Deleted and don't see any difference in how it renders.

packages/nimble-components/src/label-provider/core/index.ts
```diff
@@ -33,7 +34,8 @@ const supportedLabels = {
     filterNoResults: filterNoResultsLabel,
     loading: loadingLabel,
     scrollBackward: scrollBackwardLabel,
-    scrollForward: scrollForwardLabel
+    scrollForward: scrollForwardLabel,
+    chipRemove: chipRemoveLabel
```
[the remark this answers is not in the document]
jattasNI: The name `remove` was conflicting with a built in field in a base class so I went with `itemRemove` to disambiguate, but the user-visible string is still the same.

packages/nimble-components/src/chip/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    public hasOverflow = false;
+
+    /** @internal */
+    public get elementTextContent(): string {
+        return slotTextContent(this.contentSlot);
+    }
+
+    /** @internal */
+    public get removeButtonContent(): string {
+        return `${chipRemoveLabel.getValueFor(this)} ${this.elementTextContent}`;
```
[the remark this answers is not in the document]
rajsite: Yea it's interesting. I was wondering if there was a better way to describe the relationship of this action to the chip itself. I think we have a similar problem with banner dismiss buttons as well. Didn't find a great solution and this issue seems to talk about that problem in particular with action buttons on interactive components / components with existing role behaviors: https://github.com/w3c/aria/issues/1440

PR #2691, opened 2025-08-19
Add chat input hover and focus styling.

atmgrifter00:
# Pull Request

## 🤨 Rationale

Adding additional styling that is in the design spec (for hover & focus) [see design spec[](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=12342-81782&p=f&t=hPUto4UbWGKauq0C-0)).

## 👩‍💻 Implementation

Styling changes.

## 🧪 Testing

Manual

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/chat/input/styles.ts
+++ b/packages/spright-components/src/chat/input/styles.ts
@@ -7,7 +7,9 @@ import {
     controlLabelFontColor,
     elevation2BoxShadow,
     mediumPadding,
-    popupBorderColor
+    popupBorderColor,
+    borderHoverColor,
+    smallDelay
 } from '@ni/nimble-components/dist/esm/theme-provider/design-tokens';
 import { display } from '../../utilities/style/display';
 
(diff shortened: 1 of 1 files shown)


packages/spright-components/src/chat/input/styles.ts
```diff
@@ -31,6 +37,51 @@ export const styles = css`
         box-shadow: ${elevation2BoxShadow};
     }
```
jattasNI, line 39: The behavior I'm seeing in Storybook doesn't match the behavior I'm seeing speced out in [Nimble_Components Figma](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=16003-87883&m=dev).

image


1. The double line is supposed to be on key focus but I'm seeing it on click focus too.
2. The bottom line is supposed to be thicker on rollover than on focus, but I'm seeing it's thick in both cases.

atmgrifter00: Hmmm...I'm wondering if there is even a good way to do this. The only other component that shows double lines on keyboard focus is the Select. I'm not sure there is a CSS-only way of distinguishing that textarea has focus, but the component never received keyboard focus. I realize the intend of the `:focus-visible` selector is intended to help with this, but I think there are issues keeping us from getting precisely the behavior spec'd out here. Anecdotally, Copilot failed tragically trying to get the behavior described (and I'm pretty convinced it understood the goals, and various means to try things), but couldn't quite get it. Once it got to the point of "let's fix this using Javascript" or "let's try to use animation keyframes to signal stuff", I put it to bed.

NIbokeefe: 1.  Yes, Double underline should only be on Key Focus.
2. Standard focus is 1px line, Rollover is 2px line, Key focus is double line combo 2px and 1px.

The majority of controls utilize the keyboard focus with some version of double underline/outline treatment.
Specifically looking at double underline controls (disregard my Figma select/combo which has a bug):
Select
Combobox
Text Input
Numeric input
Search
Text Area
Links
Breadcrumb Links
Tab to text

All others are a variant of the double underline, more of a double outline.


Screenshot 2025-08-20 at 3 08 13 PM

atmgrifter00: I'll just note that at the moment, _none_ of the Nimble components that contain an input element (i.e. TextField, TextArea, NumberField, etc...) have an implemented double-underline styling. My concern is that there isn't a good way for us to actually do this in a straightforward declarative means in CSS. If that's the case, should we review the style choices for keyboard focus?

atmgrifter00: For now, the spright-chat-input will align with nimble-textarea as discussed above.

packages/spright-components/src/chat/input/styles.ts
```diff
@@ -7,7 +7,9 @@ import {
     controlLabelFontColor,
```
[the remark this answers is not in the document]
atmgrifter00: Added.

PR #2692, opened 2025-08-20
Storybook docs formatting update

fredvisser:
# Pull Request

## 🤨 Rationale

Initial implementation of an updated visual design for Storybook docs.

## 👩‍💻 Implementation

- Created a handful of React layout components:
  - Leveraged new Figma component designs and copilot to largely autogenerate components
  - For the TOC component – based the implementation on the OOTB TOC component, but updated to match the new design
- Updated the `Button` story as an example

## 🧪 Testing

- Manual review of the Button story

## ✅ Checklist


packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
-form, opening a dialog, canceling an action, or performing a delete operation.
+## Overview
 
-Consider <Tag name={anchorButtonTag} /> if you want a button that triggers
-navigation to a URL.
+A button is a widget that enables users to trigger an action or event, such as
+submitting a form, opening a dialog, canceling an action, or performing a delete
+operation. _Source: [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/button/)_
 
-<Canvas of={buttonStories.outlineButton} />
```
jattasNI, line 19: I don't like that we're getting rid of the canvas. My primary workflow with storybook is putting the control in a certain state and looking at the rendered control or generated code. To be able to do that without crazy scrolling I need the canvas and controls to be near each other. Right now this randomly controls the Outline button canvas below which is too far away to be noticeable.

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
+<TableOfContents
```
jattasNI, line 23: I was a bit confused by the visual design of the ToC initially. Is it a breadcrumb with arrows instead of slashes between the items? Is it links with arrows next to them? What does the arrow signify again? (My first guess was external link but I guess not). Why isn't there an arrow next to the last link?

image

I sort of figured out answers to those by playing with it but I'm not sure I see how this is better than the ToC we had before.

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
+/>
+
+<img
+    alt="Button anatomy"
+    src={ButtonAnatomy}
+    style={{ maxWidth: '936px', width: '100%', display: 'block' }}
+/>
 
-## API
+## Implementation & API
```
jattasNI, line 35: Do you know the rationale behind adding the "Implementation" heading? I don't think we generally want to be adding implementation details in storybook.

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
 
-## Styling
+## Usage
+
+### When to Use
+
+- Primary Actions: Use buttons for primary actions that users need to take, such
+  as submitting a form, saving changes, or proceeding to the next step.
+- Call to Action (CTA): Buttons are ideal for CTAs that encourage users to take
+  a specific action, like “Sign Up,” “Buy Now,” or “Learn More.”
```
jattasNI, line 48: This list seems incomplete. You use buttons for secondary and tertiary actions too (by my definition, "OK" is a primary action while "Cancel" would be a secondary action). And in toolbars and many other situations where there is no hierarchy/precedence. 

I was also expecting the list to direct me to how to configure the button for each of these use cases but "CTA" isn't used elsewhere in the document and this use of "Primary" doesn't tie back to our `appearance-variant='primary'`.

After reading a bit more would it make more sense for this to correspond to the numbered sub-headings under Appearance & Styling? Or just remove this list and let users discover the info below?

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
+### Keyboard Usage
+
+Keyboard accessibility is a critical component of our design system, ensuring
+that all users, including those with disabilities, can navigate and interact
+with our products efficiently. By adhering to best practices for keyboard usage,
+we create a more inclusive and user-friendly experience.
+
+## Alternatives & Related
+
+### Alternative to Consider
```
jattasNI, line 237: I personally liked having the alternatives near the top because it quickly redirects users to the right component before they spend time reading the whole document.

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
+#### The Stand Out
+
+The Block button is more visually prominent and functionally direct than the
+Outline or Ghost options. Use it when higher visibility is a priority or in busy
+environments where important controls might be missed. This appearance helps
+ensure critical actions aren’t overlooked.
+
+### Button Variants
+
+There are two available values for appearanceVariant — primary and accent —
```
jattasNI, line 91: appearanceVariant should either use code font and the attribute-style name (`appearance-variant`) or plain English (appearance variant). Since the rest of this section is using English names (e.g. "Outline and Block appearances") I'd vote for the latter.

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
...
+### 4. Primary Button
 
-<StylingDocs components={{ Button: buttonTag }} />
+<ExampleContainer>
+    <div style={{ display: 'flex', alignItems: 'center', gap: '48px' }}>
+        <NimbleButton appearance="outline" appearanceVariant="primary">
+            Primary outline
+        </NimbleButton>
+        <NimbleButton appearance="block" appearanceVariant="primary">
+            Primary outline
```
jattasNI, line 107: I'm not sure what's going on with the styling here but I don't think black text on a dark grey background is correct.

image

packages/storybook/src/nimble/button/button.mdx
```diff
@@ -1,45 +1,269 @@
 import { Canvas, Meta, Controls, Title } from '@storybook/addon-docs/blocks';
```
jattasNI, line 1: In my browser (macOS Firefox) the page load time seems not great. 

Warm refresh of the production storybook is <1 second: https://nimble.ni.dev/storybook/?path=/docs/components-button--docs

But warm refresh of this page is 6-8 seconds: https://60e89457a987cf003efc0a5b-mgbjyxprll.chromatic.com/?path=/docs/components-button--docs

It's possible this is due to the chromatic URL / pre-deployment build but I'm worried the page might be heavier than before.

jattasNI (changes requested):
Just reviewed the rendered button story and its content, not the new blocks. Might be worthwhile to meet to discuss the vision some more: I like some of the changes but a lot of it feels harder to use to me (possibly due to unfamiliarity or due to me not being the target user).

PR #2694, opened 2025-08-31
Add background api for chat conversation

atmgrifter00:
# Pull Request

## 🤨 Rationale

We have use cases for the chat UI that require the background to be transparent in order for a larger app background to appear through the chat conversation area.

## 👩‍💻 Implementation

Add a simple boolean attribute to control whether or not to show the conversation background or not.

## 🧪 Testing

Chromatic tests.

## ✅ Checklist

### DIFF
--- a/change/@ni-spright-components-ccb78f2f-5f4d-4606-909d-df44623238b8.json
+++ b/change/@ni-spright-components-ccb78f2f-5f4d-4606-909d-df44623238b8.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add API to hide background visual for chat conversation.",
+  "packageName": "@ni/spright-components",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/spright-components/src/chat/conversation/styles.ts
+++ b/packages/spright-components/src/chat/conversation/styles.ts
@@ -17,6 +17,10 @@ export const styles = css`
         border: ${borderWidth} solid ${applicationBackgroundColor};
     }
 
+    :host([hide-background]) {
+        background: transparent;
+    }
+
     .messages {
         flex: 1;
         display: flex;
@@ -27,6 +31,10 @@ export const styles = css`
             ${standardPadding};
         background: ${sectionBackgroundImage};
         overflow-y: auto;
+    }
+
+    :host([hide-background]) .messages {
+        background: transparent;
     }
 
     .input {
--- a/packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
+++ b/packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
@@ -57,6 +57,12 @@ const contentHeightStates = [
 ] as const;
 type ContentHeightStates = (typeof contentHeightStates)[number];
 
+const hideBackgroundStates = [
+    ['hide-background:true', true],
+    ['hide-background:false', false]
+] as const;
+type HideBackgroundStates = (typeof hideBackgroundStates)[number];
+
 const componentSizing = (
     [_messageTypeLabel, messageType]: MessageTypeStates,
     [viewportLabel, viewportWidth, viewportHeight]: ViewportStates,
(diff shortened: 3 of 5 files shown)


change/@ni-spright-components-ccb78f2f-5f4d-4606-909d-df44623238b8.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Could you update the chat HLD/spec to mention the new attribute too?

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -115,6 +116,11 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
         </${chatConversationTag}>
     `),
     argTypes: {
+        hideBackground: {
+            description:
```
jattasNI, line 120: We use the attribute names rather than the property names in storybook
```suggestion
            name: 'hide-background',
            description:
```

packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -17,6 +17,10 @@ export const styles = css`
         border: ${borderWidth} solid ${applicationBackgroundColor};
```
jattasNI, line 17: Do you want to hide the border too? It's subtle but I suspect we'd rather not delineate the conversation from its parent at all.
image

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
```diff
@@ -218,6 +218,31 @@ export const conversationWithInputSizing: StoryFn = createMatrixThemeStory(html`
     ${createMatrix(conversationWithInput, [heightStates])}
 `);
 
+export const conversationBackground: StoryFn = createStory(
```
jattasNI, line 221: Could you make this a matrix test that includes the theme and the hide-background state?

It'd also be nice to copy the `<p>` styling from above.

```
    <p 
        style="
        color: var(${bodyFontColor.cssCustomProperty});
        margin-bottom: 0px;
        "
    >
```

packages/spright-components/src/chat/conversation/styles.ts
jattasNI: I could see an argument that this would mean it would be better to call this setting `appearance='frameless'`. But since this is a spright component I don't think we should get bogged down in naming, I'm ok with hide-background for now and we can change it if we get new designs with different variations.

jattasNI (approved):

packages/spright-components/src/chat/conversation/styles.ts
atmgrifter00: Is the other available value `default`?

rajsite: > I kind of like "overlay" as it conveys intent for usage.

Yea `overlay` works here since the usage is still visual oriented (wouldn't want an enum over coupled to use-cases like appearance="for-header" or something). Also wouldn't want `frameless` because that isn't really aligned to our other uses of frameless. I think `overlay` works okay for this new kind of visual state 👍

I'd also follow the newer attribute pattern we have been using where `default`  is `undefined` (I'd like to go back and fix our existing appearance attrs at some point).

```
"default": undefined,
"overlay": "overlay"
```

packages/storybook/src/spright/chat/conversation/chat-conversation-matrix.stories.ts
atmgrifter00: I think I have it the way you expect?

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
atmgrifter00: Attribute name has changed.

rajsite (approved):

rajsite:
Bypass merging, GitHub Actions is being super slow (40 min on `npm ci`) and the [most recent build failures were table virtualization related](https://github.com/ni/nimble/actions/runs/17586837551/job/49957023952) so this PR seems very unlikely to be related.

PR #2699, opened 2025-09-09
Add nimble-select Angular guidance for compareWith function

msmithNI:
# Pull Request

## 🤨 Rationale

I recently investigated a [bug](https://ni.visualstudio.com/DevCentral/_workitems/edit/3244716) filed for the `nimble-select` about incorrect behavior when using `filter-mode` of `manual`:

> When using `nimble-select` in an Angular reactive form with `filter-mode="manual"`, the component's visual display value is cleared if the list of `nimble-list-option` elements is dynamically replaced.
This occurs even if the new list contains an option that corresponds to the selected value and the underlying `formControl` value remains correct.
> 
> **Steps to Reproduce**
> Create an Angular component with a formControl.
> Bind this formControl to a <nimble-select> with filter-mode="manual".
> Populate the select with <nimble-list-option> elements using *ngFor and bind objects to the options using [ngValue].
> Implement a filter function (e.g., on the (filter-input) event) that creates a new array of options.
> Run the application and select an option from the dropdown. The select correctly displays the name of the selected object.
> Type into the filter. The list of options is rebuilt.
> Expected Behavior
> If the currently selected object is present in the new, filtered list of options, the nimble-select should continue to display its name/label.
>  **Actual Behavior**
> The nimble-select's display value is cleared and it reverts to showing the placeholder, even though formControl.value still holds the correct selected object.

Based on my investigation, the "binding to options with [ngValue]" was the important part. I could reproduce the behavior described (`nimble-select` reverting to showing the placeholder after updating the options). However specifying a `compareWith` function on the `nimble-select` was sufficient to fix the problem.

## 👩‍💻 Implementation

Update `nimble-select` Storybook docs with guidance about specifying a `compareWith` function if binding to options with `ngValue`.

## 🧪 Testing

PR build.

## ✅ Checklist

### DIFF
--- a/packages/storybook/src/nimble/select/select.mdx
+++ b/packages/storybook/src/nimble/select/select.mdx
@@ -37,19 +37,20 @@ in addition to selecting from the list.
 <Description of={listOptionStories} />
 <Controls of={listOptionStories.listOption} />
 
+#### Angular
+
+In Angular, a <Tag name={listOptionTag} />
+[can be created](https://angular.dev/api/forms/NgSelectOption) with a string
+`value` or an object `ngValue`. When using `ngValue`, provide a
+[`compareWith` implementation](https://angular.dev/api/forms/SelectControlValueAccessor#customizing-option-selection)
+on the Select control. Without this, the Select may display unexpected behavior
+(like reverting to the placeholder value when updating the list of filtered
+options).
+
 ### List Option Group
 
 <Description of={listOptionGroupStories} />
 <Controls of={listOptionGroupStories.listOptionGroup} />
-
-### Angular
-
-In Angular, an <Tag name={listOptionTag} /> can be created with an `ngValue`. In
-that scenario, especially when using a `filter-mode` of `manual`, a
-[`compareWith` implementation](https://angular.dev/api/forms/SelectControlValueAccessor#customizing-option-selection)
-should be provided on the Select control. Without this, the Select may display
-unexpected behavior (like reverting to the placeholder value when updating the
-list of filtered options).
 
 ## Styling
 
--- a/packages/storybook/src/nimble/select/select.mdx
+++ b/packages/storybook/src/nimble/select/select.mdx
@@ -42,6 +42,15 @@ in addition to selecting from the list.
 <Description of={listOptionGroupStories} />
 <Controls of={listOptionGroupStories.listOptionGroup} />
 
+### Angular
+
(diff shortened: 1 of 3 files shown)


packages/storybook/src/nimble/select/select.mdx
```diff
@@ -42,6 +42,15 @@ in addition to selecting from the list.
...
 <Controls of={listOptionGroupStories.listOptionGroup} />
 
+### Angular
+
+In Angular, an <Tag name={listOptionTag} /> can be created with an `ngValue`. In
+that scenario, especially when using a `filter-mode` of `manual`, a
+[`compareWith` implementation](https://angular.dev/api/forms/SelectControlValueAccessor#customizing-option-selection)
+should be provided on the Select control. Without this, the Select may display
+unexpected behavior (like reverting to the placeholder value when updating the
+list of filtered options).
```
jattasNI, line 52: Based on [these docs](https://angular.dev/api/forms/NgSelectOption#ngValue) the difference between `value` and `ngValue` is that the latter should be used when the value is an object. I think it'd be helpful to add that context here, either inline or with a link. Maybe something like this?

I'd also suggest moving this inside the `### List Option` section.

```suggestion
### Angular

In Angular, a <Tag name={listOptionTag} /> [can be created](https://angular.dev/api/forms/NgSelectOption) with a string `value` or an object `ngValue`. When using `ngValue`, provide a
[`compareWith` implementation](https://angular.dev/api/forms/SelectControlValueAccessor#customizing-option-selection) on the Select control. Without this, the Select may display
unexpected behavior (like reverting to the placeholder value when updating the
list of filtered options).
```

packages/storybook/src/nimble/select/select.mdx
```diff
@@ -42,6 +42,15 @@ in addition to selecting from the list.
 <Description of={listOptionGroupStories} />
 <Controls of={listOptionGroupStories.listOptionGroup} />
 
+### Angular
```
jattasNI, line 45: Looks like the example app follows this pattern. Should we update it to obey the new guidance? Or update the guidance to explain whatever subtlety makes it so the bug doesn't occur in that case?

https://github.com/ni/nimble/blob/3b0f6f6a149e769fb040b5b1331154d9709327fb/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html#L199

jattasNI (approved):

rajsite (approved):

PR #2709, opened 2025-09-30
[Draft] Upgrade to eslint v9

gokulprasanth-ni:
# Pull Request

## 🤨 Rationale

- Update nimble packages to support flat eslint configuration format

## 👩‍💻 Implementation

- converted the legacy eslint configs to flat config format.
- Updated the dependencies version to be compatible with eslint v9. 

## ✅ Checklist

- [] I have updated the project documentation to reflect my changes or determined no changes are needed.


packages/angular-workspace/example-client-app/eslint.config.js
```diff
@@ -0,0 +1,48 @@
+import { fileURLToPath } from 'url';
```
rajsite, line 1: These are too many changes in one shot. One thing to note is that packages do not need to migrate to `"type": "module"` in order to adopt eslint v9. Those can be decoupled, see the discussion here: https://github.com/ni/javascript-styleguide/pull/170#discussion_r2392464487

If the changes are minor outside the nimble repo then it may be fine but we should not switch the packages in nimble.

rajsite: Also incase it is a consideration, we should be updating the AzDo pipelines first

gokulprasanth-ni: Yeah thanks for the suggestion Milan. I will plan to create PRs as suggested.

Issue #2708, opened 2025-09-24
Configurable table selection mode for action menus


rajsite:
# 🙋 Feature Request

## 😯 Problem to Solve

When a nimble table action menu is opened, the selection state of the table is updated to reflect that:
- in single selection mode the record the action menu is opened for becomes selected
- in multiselect mode if the record activated is inside the selected set, the selected set doesn't change
- in multiselect mode if the record activated is outside the selected set, the selected set is cleared and just the activated record becomes selected

This is problematic for a couple scenarios:
- In a master-detail view it can be jarring / unexpected for an action menu that opens on a record to change selection
- In a scenario where the selection context is significant, opening the action menu causing selection to change may be unexpected

Example use cases:
- Master-detail view in [specs management app](https://www.figma.com/design/Vvs0tyCQ4SsbUM8UG0iiKD/Specs-Management?node-id=2596-8904&p=f&t=wfvvXNVdHHufEhuo-0)
- Side panel opened for nigel where selected items are context for the AI
- Expected workflows in chat history UIs (effectively same pattern as master-detail)

### Alternative: Add a configurable option to the table for selection behavior from action menus
- Changing the behavior has two related impacts:
  - Opening the action menu impacts selection state. New behavior would be to not change selection state on open.
  - Opening the action menu fires an event that includes a list of `records`. New behavior:
    - Sub alternate: The event `records` represents what the action menu is operating on so with the default setting it happens to represent the updated selection state and with the new setting it always represents a single record that the menu is opened for
    - Sub alternate: The event `records` always represents the current selection state of the table for default setting and new setting but we add a new field like `operatingRecord` that represents the exact record the action menu was opened on. This is potentially useful in both the default and new mode, it's context we didn't expose before.
- Pros
   - Allows apps that migrate to master-detail UIs to opt-in to the behavior independently
- Cons
   - Unknown / FUD: how large a change it would be or what the api should be

### Alternative: Add a configurable option to each table column for selection behavior from action menus
- Pros
  - Potentially more scoped than a table wide change
  - Could maybe be useful to have i.e. a "main" column (like the first column in SystemLink apps) that is conceptually coupled to the selection state and other columns where one is conceptually operating on that column and would not expect operations to be coupled to full table selection state. (no known use-cases, closest was analysis table where action menus are used to show / hide columns)
- Cons
  - More chances for inconsistencies between columns in tables and between tables in UIs?
[...]

PR #2732, opened 2025-10-22
[HLD] Table Selection Behavior with Action Button 

hellovolcano:
# Pull Request

## 🤨 Rationale

For some tables, there is a need to open the action menu without changing the selection. This HLD proposes a way to support multiple selection modes in a table when opening an action menu. This is the design for #2708 .

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
```diff
@@ -0,0 +1,234 @@
+# Table Selection Mode for Action Menus HLD
```
jattasNI: The PR pipeline is failing in the lint step (see the "Checks" section at the bottom of the Conversation tab). This is an automated check that makes sure our documents have consistent formatting. To fix it you can follow [these steps in CONTRIBUTING.md](https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#linting). If you haven't already, you'll first have to do the [getting started steps](https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#getting-started) (install Node.js and run `npm install`).

packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Links To Relevant Work Items and Reference Material
+
+[Configurable table selection mode for action menus](https://github.com/ni/nimble/issues/2708)
+
+## Implementation / Design
+
+### Overview
+
+This design adds a new boolean attribute `action-menu-changes-selection` to the `nimble-table` component with the following characteristics:
```
jattasNI: We document our [attribute naming conventions in the nimble-components CONTRIBUTING](https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#represent-control-states-as-attributes). The proposal is mostly following them, but one thing to fix is that attributes must default to false (not present). So to preserve existing behavior for clients we'll need to invert the name and meaning of this attribute.

Another small point is that a table can have multiple action menus. This proposal would affect all of them so I'd vote to use "action menus".

Something like `action-menus-preserve-selection` (or "maintain" or "keep" or similar) would make sense to me.

One other alternative would be to change it to an enum like `action-menus-selection-behavior` with values like `default` and `"preserve-selection"`. That would give us room to add new modes in the future.

I'm struggling to imagine other modes so tentatively vote for the boolean attribute, but would like to hear your and other owners' opinions too.

packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
```diff
@@ -0,0 +1,234 @@
...
+
+- **Default value**: `true` (action menus change selection - preserves existing behavior)
+- **Backward compatibility**: Existing tables continue to work unchanged
+- **Opt-in new behavior**: Setting to `false` enables non-intrusive action menu behavior
+- **Enhanced event context**: New `operatingRecord` field provides trigger record information in both modes
```
jattasNI: Can you clarify the name of the event this is referring to?

packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
hellovolcano: I can't really think of any other modes either, so I've renamed to `action-menus-preserve-selection`

packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
hellovolcano: I took this section out, and specifically list the new field under `TableActionMenuToggleEventDetail` in the API changes section

packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md
jattasNI: I'll re-open this thread for now just to give other owners a place to comment if they have opinions.

rajsite:
Didn't have a chance to review today, expect to review tomorrow

rajsite:
Sorry for taking so long to review, but looks great! Thanks for the patience :)

PR #2737, opened 2025-10-31
Checkbox directive fix

### LINKED ISSUE #2734: Angular form control needs to be reset twice on nimble-checkbox
# 🐛 Bug Report

In some cases, a form control backing a nimble-checkbox in Angular needs to be reset twice to restore it back to its initial state. Specifically, this is a problem when using a tri-state checkbox where the form control can take the following values:
- true (checkbox checked)
- false (checkbox unchecked)
- null (checkbox indeterminate)

Additionally, for this to be a problem, the initial value needs to be null (i.e. indeterminate checkbox).

After some debugging the issue is the following:
- Resetting form control changes its value to `null`
- nimble-checkbox sees the value of `null` and changes the value of the checkbox to `false`
- [Root problem] nimble-checkbox triggers `change` event for the programmatic change of setting the value to `false`
- Nimble checkbox's CVA handles `change` event by setting the value on the form => **form value is incorrectly set to `false` rather than being reset to `null`**

## 💻 Repro or Code Sample

Template code:
```html
<nimble-checkbox
    [formControl]="booleanFormControl"
    [indeterminate]="booleanFormControl.value === null || booleanFormControl.value === ''">
        Test checkbox
    </nimble-checkbox>

<nimble-button (click)="onResetFormControl()">
    Reset Form Control
</nimble-button>
```

App code:
```ts
 public booleanFormControl = new FormControl({ value: '', disabled: false });

public onResetFormControl(): void {
    this.booleanFormControl.reset();
}
```

Steps to reproduce:
- Use the above code
- Notice the checkbox is in an indeterminate state by default
- Put the checkbox in a checked state
- Press the "Reset Form Control" button
- [Incorrect behavior] Notice the checkbox is in an unchecked state when it should be indeterminate
- Press the "Reset Form Control" button again
- Notice the checkbox now enters the correct state of being indeterminate

## 🤔 Expected Behavior

Resetting a form control to `null` should make its value `null`

## 😯 Current Behavior

Resetting a form control to `null` makes its value `false`

## 💁 Possible Solution

The nimble checkbox CVA could be updated to have code that ignores change events on the checkbox when the value is being set programmatically.

Something like this:
```ts
public override writeValue(value: unknown): void {
    this.ignoreChangeEvents = true;
    super.writeValue(value);
    this.ignoreChangeEvents = false;
}

@HostListener('change', ['$event'])
public nimbleOnChange(event: Event): void {
    if (this.ignoreChangeEvents) {
        return;
    }
    this.onChange(event);
}
```

Note: this is a similar strategy that is used in the nimble-table for some selection-related behavior that ran into a similar problem.

## 🔦 Context

This is affecting SLE DFF: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/1087594

## 🌍 Your Environment

* Version [e.g. 1.8.0]

mollykreis:
# Pull Request

## 🤨 Rationale

Fixes #2734

## 👩‍💻 Implementation

Resetting the form control sets the value to `null` (unless `nonNullable` is configured on the form control). However, the `change` event fired from the checkbox due to the programmatic value change causes the value to be set again to `false`. Therefore, to solve the problem, I added code to ignore change events fired from the checkbox while the form was programmatically setting the value.

Ultimately, it would be nice to solve the root cause of the problem which is that the checkbox fires a `change` event when the value is programmatically set. However, this is a larger problem to tackle -- not only for the checkbox but across all/most FAST-based components. Therefore, I think this is an acceptable solution for now, and it is aligned with a change that was put into the table to resolve a similar problem related to row selection (see https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/components/group-row/index.ts#L75-L79)

## 🧪 Testing

- Manually tested in an Angular app
- Added new unit tests for the checkbox CVA that creates a form control (previous checkbox CVA tests only exercised `ngModel` binding code paths)

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-ee4f4590-982e-4443-b9c6-4d05df22f7bb.json
+++ b/change/@ni-nimble-components-ee4f4590-982e-4443-b9c6-4d05df22f7bb.json
@@ -1,7 +1,0 @@
-{
-  "type": "minor",
-  "comment": "Add new icons for camera, scanner gun, DIAdem, calipers",
-  "packageName": "@ni/nimble-components",
-  "email": "jattasNI@users.noreply.github.com",
-  "dependentChangeType": "minor"
-}
--- a/change/@ni-nimble-tokens-08e4a54a-5d0b-472f-ad6a-31da92005744.json
+++ b/change/@ni-nimble-tokens-08e4a54a-5d0b-472f-ad6a-31da92005744.json
@@ -1,7 +1,0 @@
-{
-  "type": "minor",
-  "comment": "Add new icons for camera, scanner gun, DIAdem, calipers",
-  "packageName": "@ni/nimble-tokens",
-  "email": "jattasNI@users.noreply.github.com",
-  "dependentChangeType": "patch"
-}
--- a/package-lock.json
+++ b/package-lock.json
@@ -29541,7 +29541,7 @@
     },
     "packages/angular-workspace/nimble-angular": {
       "name": "@ni/nimble-angular",
-      "version": "30.8.0",
+      "version": "30.9.0",
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.2.0"
@@ -29552,12 +29552,12 @@
         "@angular/forms": "^18.2.13",
         "@angular/localize": "^18.2.13",
         "@angular/router": "^18.2.13",
-        "@ni/nimble-components": "^34.4.0"
+        "@ni/nimble-components": "^34.5.0"
       }
     },
     "packages/angular-workspace/ok-angular": {
       "name": "@ni/ok-angular",
-      "version": "0.0.6",
+      "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.2.0"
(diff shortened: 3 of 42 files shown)


packages/angular-workspace/nimble-angular/src/directives/checkbox/nimble-checkbox-control-value-accessor.directive.ts
```diff
@@ -10,14 +11,30 @@ import { CheckboxControlValueAccessor } from '../../thirdparty/directives/checkb
 @Directive({
     selector:
       'nimble-checkbox[formControlName],nimble-checkbox[formControl],nimble-checkbox[ngModel]',
-    // The following host metadata is duplicated from CheckboxControlValueAccessor
```
rajsite: This doesn't need to necessarily be removed. On Angular upgrades we should be aligning to the Angular CheckboxControlValueAccessor changes / bugfixes

packages/angular-workspace/nimble-angular/src/directives/checkbox/nimble-checkbox-control-value-accessor.directive.ts
```diff
@@ -10,14 +11,30 @@ import { CheckboxControlValueAccessor } from '../../thirdparty/directives/checkb
...
-    host: { '(change)': 'onChange($event.target.checked)', '(blur)': 'onTouched()' },
+    host: { '(blur)': 'onTouched()' },
     providers: [{
         provide: NG_VALUE_ACCESSOR,
         useExisting: forwardRef(() => NimbleCheckboxControlValueAccessorDirective),
         multi: true
     }]
 })
 export class NimbleCheckboxControlValueAccessorDirective extends CheckboxControlValueAccessor {
+    private ignoreChangeEvents = false;
```
rajsite: Can we create / link to an existing bug around checkbox change events firing on programmatic changes?
Sounds very familiar so may already exist. Can reference that issue here to remove the workaround when fixed.

rajsite (approved):

PR #2746, opened 2025-11-07
Fix NumberField value coercion to zero in non-English locales

fredvisser:
# Pull Request

## 🤨 Rationale

This PR fixes a [bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/3430038) where `NimbleNumberField` would render incorrect decimal separators (e.g., `1,5` instead of `1.5`) when running in locales that use commas for decimals (like German). The underlying web component requires a period `.` as the decimal separator regardless of the user's locale.

Additionally, this PR improves the quality of the `NimbleNumberField` tests by addressing resource leaks and code style issues.

## 👩‍💻 Implementation

- **Fix**: Updated `NimbleNumberField.razor` to use `FormatValueAsString` (which enforces `CultureInfo.InvariantCulture`) for `step`, `min`, and `max` attributes. This ensures they are always rendered with a period separator.
- **Refactoring**:
    - **TestContext Disposal**: Switched to `using var context = ...` in `NimbleNumberFieldTests.cs`. This ensures the Bunit `TestContext` is properly disposed, aligning with Bunit documentation. *Note: While this pattern should ideally be applied globally, this PR focuses only on the files touched by the fix. Global adoption should be a separate effort.*
    - **Removed `RenderWithPropertySet`**: This helper method was removed because it encapsulated `TestContext` creation without returning it, making proper disposal impossible. Inlining the logic (and using the new `CreateTestContext` helper) makes the lifecycle explicit and allows for the `using` pattern.
    - **Clean Code**: Introduced `CultureScope` and `CreateTestContext` helpers to reduce duplication in the tests.

## 🧪 Testing

- Added specific regression tests in `NimbleNumberFieldTests.cs` that force the thread culture to `de-DE` (German).
- **Verification**: Verified that these new tests **failed** before the fix was implemented (rendering `1,5`) and **passed** after the fix (rendering `1.5`).
- Ran all unit tests to ensure no regressions.

## ✅ Checklist

### DIFF
--- a/packages/blazor-workspace/NimbleBlazor/Components/NimbleNumberField.razor
+++ b/packages/blazor-workspace/NimbleBlazor/Components/NimbleNumberField.razor
@@ -1,5 +1,4 @@
 ﻿@namespace NimbleBlazor
-@using System.Globalization
 @inherits NimbleInputBase<double?>
 <nimble-number-field
     appearance="@Appearance.ToAttributeValue()"
@@ -8,10 +7,10 @@
     readonly="@ReadOnly"
     disabled="@Disabled"
     placeholder="@Placeholder"
-    step="@Step.ToString(CultureInfo.InvariantCulture)"
+    step="@FormatValueAsString(Step)"
     hide-step="@HideStep"
-    min="@Min.ToString(CultureInfo.InvariantCulture)"
-    max="@Max.ToString(CultureInfo.InvariantCulture)"
+    min="@FormatValueAsString(Min)"
+    max="@FormatValueAsString(Max)"
     error-text="@ErrorText"
     error-visible="@ErrorVisible"
     required-visible="@RequiredVisible"
--- a/packages/blazor-workspace/Tests/NimbleBlazor.Tests/Unit/Components/NimbleNumberFieldTests.cs
+++ b/packages/blazor-workspace/Tests/NimbleBlazor.Tests/Unit/Components/NimbleNumberFieldTests.cs
@@ -7,7 +7,7 @@ using Xunit;
 namespace NimbleBlazor.Tests.Unit.Components;
 
 /// <summary>
-/// Tests for <see cref="NimbleTextField"/>
+/// Tests for <see cref="NimbleNumberField"/>
 /// </summary>
 public class NimbleNumberFieldTests
 {
@@ -16,8 +16,7 @@ public class NimbleNumberFieldTests
     [Fact]
     public void NimbleNumberField_Rendered_HasNumberFieldMarkup()
     {
-        using var context = new TestContext();
-        context.JSInterop.Mode = JSRuntimeMode.Loose;
+        using var context = CreateTestContext();
 
         var textField = context.RenderComponent<NimbleNumberField>();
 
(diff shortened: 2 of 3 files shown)


packages/blazor-workspace/Tests/NimbleBlazor.Tests/Unit/Components/NimbleNumberFieldTests.cs
```diff
@@ -56,111 +57,225 @@ public void NimbleNumberField_SupportsAdditionalAttributes()
...
+            // Verify that the value attribute uses period (.) not comma (,)
+            // This is required because the web component expects period regardless of locale
+            Assert.Contains("current-value=\"1.5\"", numberField.Markup);
+            Assert.DoesNotContain("current-value=\"1,5\"", numberField.Markup);
+        }
+        finally
+        {
+            // Restore the original culture
+            CultureInfo.CurrentCulture = originalCulture;
+        }
```
jattasNI:
```suggestion
        var originalCulture = CultureInfo.CurrentCulture;
        try
        {
            CultureInfo.CurrentCulture = new CultureInfo("de-DE");

            using var context = new TestContext();
            context.JSInterop.Mode = JSRuntimeMode.Loose;

            var numberField = context.RenderComponent<NimbleNumberField>(
                p => p.Add(x => x.Value, 1.5));

            Assert.Contains("current-value=\"1.5\"", numberField.Markup);
            Assert.DoesNotContain("current-value=\"1,5\"", numberField.Markup);
        }
        finally
        {
            CultureInfo.CurrentCulture = originalCulture;
        }
```

packages/blazor-workspace/Tests/NimbleBlazor.Tests/Unit/Components/NimbleNumberFieldTests.cs
```diff
@@ -15,7 +16,7 @@ public class NimbleNumberFieldTests
     [Fact]
     public void NimbleNumberField_Rendered_HasNumberFieldMarkup()
     {
-        var context = new TestContext();
+        using var context = new TestContext();
```
jattasNI: It would be helpful if the PR description mentioned the rationale for the `using` change.

From some digging, [it appears to match the Bunit docs](https://bunit.dev/docs/getting-started/writing-tests.html?tabs=xunit#instantiate-bunitcontext-in-each-test) (TestContext was renamed to BunitContext) so I think this is a positive change. But if it is then we should apply it to [everywhere that creates a TestContext](https://github.com/search?q=repo%3Ani%2Fnimble%20testcontext&type=code) in a separate PR.

packages/blazor-workspace/NimbleBlazor/Components/NimbleNumberField.razor
```diff
@@ -1,16 +1,17 @@
...
 <nimble-number-field
     appearance="@Appearance.ToAttributeValue()"
-    current-value="@CurrentValue"
+    current-value="@CurrentValueAsString"
     @onchange="@(EventCallback.Factory.CreateBinder<string?>(this, __value => CurrentValueAsString = __value, CurrentValueAsString))"
     readonly="@ReadOnly"
     disabled="@Disabled"
     placeholder="@Placeholder"
-    step="@Step"
+    step="@Step.ToString(CultureInfo.InvariantCulture)"
```
jattasNI: We should use the new `FormatValueAsString` method everywhere that we format a value as a string, not duplicate its logic.

PR #2751, opened 2025-11-11
Add configurable table selection mode for action menus

### LINKED ISSUE #2708: Configurable table selection mode for action menus

hellovolcano:
# Pull Request

## 🤨 Rationale

Adds a new boolean attribute to control whether action menus preserve the current table selection or change selection to the clicked row, as detailed in [Table Selection Mode for Action Menus HLD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table/specs/table-action-menu-selection-hld.md).

Fixes #2708 

## 👩‍💻 Implementation

- Added `action-menus-preserve-selection` boolean attribute to table with default value `false`.
- Added `operatingRecordId` to capture the row from which the action menu was clicked.
- Passed `actionMenusPreseveSelection` attribute to selection managers
   - added handleActionMenusPreserveSelectionChanged method to InterfactiveSelectionManager
   - modified `SingleSelectionManager` and `MultiSelectionManager` to use preservation flag
   - added additional checks to determine if we trigger a selection change event
- Added new attribute to Angular and Blazor wrappers

## 🧪 Testing

- Manual testing in Storybook
- Added tests for both selection modes

## ✅ Checklist


packages/angular-workspace/nimble-angular/table/nimble-table.directive.ts
```diff
@@ -66,6 +66,14 @@ export class NimbleTableDirective<TData extends TableRecord = TableRecord> imple
         this.renderer.setProperty(this.elementRef.nativeElement, 'selectionMode', value);
     }
 
+    public get actionMenusPreserveSelection(): boolean {
```
jattasNI: We have a testing pattern for Angular directives for each of our attributes and properties. You can see it in [nimble-table.directive.spec.ts](https://github.com/ni/nimble/blob/main/packages/angular-workspace/nimble-angular/table/tests/nimble-table.directive.spec.ts)

For each property / attribute we test it in 4 or 5 different configurations (default values, setting via property, setting via attribute, etc). If the table directive doesn't have examples of that pattern for a boolean attribute then you can check what it looks like on another component.

packages/blazor-workspace/NimbleBlazor/Components/NimbleTable.razor.cs
```diff
@@ -33,6 +33,9 @@ public partial class NimbleTable<TData> : ComponentBase
     [Parameter]
     public TableRowSelectionMode? SelectionMode { get; set; }
 
+    [Parameter]
+    public bool ActionMenusPreserveSelection { get; set; }
```
mollykreis: I believe this should be `bool?`.


```suggestion
    public bool? ActionMenusPreserveSelection { get; set; }
```

rajsite (approved):
Looks great! 🎉

Issue #2744, opened 2025-11-06
Console error and broken page after opening multiple action menus in table [Firefox only]


jattasNI:
# 🐛 Bug Report

If I click around to open and close multiple action items in the Nimble table, after a few clicks I see a JS error in the browser console and the page stops working.

## 💻 Repro or Code Sample

1. Open the [table story in storybook](https://nimble.ni.dev/storybook/?path=/docs/components-table--docs)
2. Open the action menu on one row
3. Close it either by clicking the menu button or clicking elsewhere in the table
4. Open the action menu on a different row and close it
5. Try to open the first action menu again

## 🤔 Expected Behavior

The action menu opens

## 😯 Current Behavior

The action menu doesn't open, some row content disappears, and an error appears in the console.

https://github.com/user-attachments/assets/8519b03a-09ec-41a9-94a2-2e88fbd2d25d

```
Uncaught TypeError: can't access property "insertBefore", node.parentNode is null
    insertBefore https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:82139
    updateContentTarget https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81790
    handleChange https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81934
    notify https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81312
    call https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81493
    tryRunTask https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81049
    process2 https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81059
    enqueue https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81073
    tryReflectToAttribute https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:82636
    setValue https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:82606
    set https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81574
    checkedChanged https://nimble.ni.dev/storybook/assets/form-associated-Br3tIWZA.js:345
    checkedChanged https://nimble.ni.dev/storybook/assets/switch-eaWEFoHN.js:48
    setValue https://nimble.ni.dev/storybook/assets/iframe-BsQQIFs0.js:81420
[...]

PR #2760, opened 2025-11-21
Add work around to table to fix action-menu bug in Firefox

mollykreis:
# Pull Request

## 🤨 Rationale

Provides a work around for the action menu issue reported in #2744

## 👩‍💻 Implementation

This doesn't solve the root cause of the issue, but it does avoid it (in a way that I don't fully understand).

The change in this PR sets a row's `currentActionMenuColumn` to `undefined` when the action menu closes, which removes the `slot` element for the menu within row.

## 🧪 Testing

Manually verified that the action menu can be opened on multiple rows in the table without an exception being thrown and without data vanishing from the table.

## ✅ Checklist


rajsite:
Brainstormed with Molly offline. Some things we looked into:

- Doesn't seem related to slot forwarding attach / detach timings (tried to see what happens if the slots are never removed which is what clearing `currentActionMenuColumn` does but doesn't seem to help)
- Columns on rows end up disappearing during the interactions, couldn't understand why
- Didn't successfully root cause a reason why the workaround helps in the time allotted
- Wouldn't be surprised if it's a Firefox browser regression, pretty sure Meyers was a daily driver in Firefox so probably worked historically, but have not tried to bisect / reproduce in a reportable issue.

Decided having a patch that we don't understand but seems to work around the behavior and seems probably innocuous maybe is probably better than a core workflow being broken. Doesn't close the issue but works around until we can investigate more. Will approve but wait for @jattasNI sign-off to merge.

rajsite (approved):

jattasNI (approved):
I tested the storybook build in Firefox and verified that I can no longer reproduce.

PR #2765, opened 2025-11-26
Update react example app and associated component changes

venkat040906:
# Pull Request

## 🤨 Rationale

- Update the example app to enable remaining missing functionality to align with Angular / Blazor
- Add a keyValue property to mapping elements as an alias to avoid conflicting with React reserved key prop.
- Added an entry to the nimble.dev landing page linking to the react app
- In react wrapper expose UserDismissed as aliases DialogUserDismissed and DrawerUserDismissed 
- Enable react eslint config recommended 

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Manually tested behavior in example app.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated mapping key topic in storybook and documented select placeholder workaround

### DIFF
--- a/change/@ni-nimble-components-a19e51ab-0944-4825-a59d-cd80f926c4d7.json
+++ b/change/@ni-nimble-components-a19e51ab-0944-4825-a59d-cd80f926c4d7.json
@@ -2,6 +2,6 @@
   "type": "minor",
   "comment": "Add keyValue accessor to Mapping base class for React compatibility",
   "packageName": "@ni/nimble-components",
-  "email": "venkatnarayana.akula@emerson.com",
+  "email": "1588923+rajsite@users.noreply.github.com",
   "dependentChangeType": "patch"
 }
--- a/change/@ni-nimble-react-e9d4584a-e10b-4d4b-93ec-5285737aa256.json
+++ b/change/@ni-nimble-react-e9d4584a-e10b-4d4b-93ec-5285737aa256.json
@@ -2,6 +2,6 @@
   "type": "patch",
   "comment": "Export missing types from drawer and table wrappers",
   "packageName": "@ni/nimble-react",
-  "email": "venkatnarayana.akula@emerson.com",
+  "email": "1588923+rajsite@users.noreply.github.com",
   "dependentChangeType": "patch"
 }


package.json
```diff
@@ -67,6 +67,6 @@
     "beachball": "^2.45.0",
     "concurrently": "^9.0.0",
     "patch-package": "^8.0.0",
-    "playwright": "1.54.1"
+    "playwright": "^1.57.0"
```
rajsite: There seems to be a bunch of unrelated changes outside of the react workspace changing dependencies, etc. Can you revert these unnecessary changes?

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -97,6 +597,7 @@ export function App(): JSX.Element {
                     <div className="sub-container">
                         <div className="container-label">Anchor</div>
                         <div><NimbleAnchor href="#" appearance="prominent">Site root</NimbleAnchor></div>
+                        Let's try it <NimbleAnchor href="#" underline-hidden={false}>using href</NimbleAnchor>.
```
jattasNI: nit: looks like the Angular example app has slightly different configuration for this link. Let's try to keep them in sync across frameworks.

https://github.com/ni/nimble/blob/4e353887ac1607dd2a05c0544e2b6189dd90c03c/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html#L12

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -299,7 +801,7 @@ export function App(): JSX.Element {
                         <div className="container-label">Select</div>
                         <NimbleSelect filter-mode="standard" appearance="underline">
                             Underline Select
-                            <NimbleListOption hidden selected disabled>Select an option</NimbleListOption>
+                            <NimbleListOption hidden disabled>Select an option</NimbleListOption>
```
jattasNI: What's the rationale for removing `selected` from these list options? The Angular app keeps it and this state of hidden + selected + disabled is [documented as the best approach for placeholder values](https://nimble.ni.dev/storybook/?path=/docs/components-select--docs#placeholder) so ideally we'd keep using that approach here too.

https://github.com/ni/nimble/blob/4e353887ac1607dd2a05c0544e2b6189dd90c03c/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html#L161

packages/react-workspace/react-client-app/src/components/App.tsx
rajsite: routerLinks only apply to angular, we can follow blazor and just not have this in the example

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -84,6 +393,197 @@ export function App(): JSX.Element {
...
+    function onTabToolbarButtonClick(): void {
+        alert('Tab toolbar button clicked');
+    }
+
+    function onChatInputSend(event: CustomEvent<{ text: string }>): void {
+        const text = event.detail.text;
+        setChatUserMessages(prevMessages => [...prevMessages, text]);
+    }
+
+    // Set attributes on mapping user elements imperatively since React doesn't pass 'key' prop to DOM
```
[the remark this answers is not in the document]
rajsite: we don't need it!

packages/react-workspace/react-client-app/src/components/App.tsx
rajsite: Also added a note in storybook

packages/nimble-components/CONTRIBUTING.md
```diff
@@ -195,6 +195,7 @@ It is common in web development to represent variations of control states using
     - Avoid any names in the [MDN HTML attribute reference list](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#attribute_list) (unless the attribute is trying to match that behavior exactly).
     - Do a best effort search in relevant working groups for new attributes that may be coming to avoid, i.e. https://github.com/openui and https://github.com/whatwg.
     - Avoid any names that are [reserved words](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) in JavaScript.
+    - Avoid any names that are [reserved props](https://legacy.reactjs.org/warnings/special-props.html) in React.
```
rajsite: Best doc I could find, didn't see an equivalent topic in the latest site

rajsite:
reset @jattasNI for another lookover

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -84,6 +385,174 @@ export function App(): JSX.Element {
...
 
+    function openDrawer(): void {
+        void (async (): Promise<void> => {
+            const closeReason = await drawerRef.current?.show();
+            setDrawerCloseReason((closeReason === DrawerUserDismissed) ? 'escape pressed' : String(closeReason ?? 'unknown'));
+        })();
+    }
+
+    function closeDrawer(reason: string): void {
+        drawerRef.current?.close(reason as never);
```
rajsite: Should double check what's going on, not sure why the type is casted to `never`

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -299,7 +769,7 @@ export function App(): JSX.Element {
                         <div className="container-label">Select</div>
                         <NimbleSelect filter-mode="standard" appearance="underline">
                             Underline Select
-                            <NimbleListOption hidden selected disabled>Select an option</NimbleListOption>
+                            <NimbleListOption hidden disabled>Select an option</NimbleListOption>
```
rajsite: Maybe should comment and add a link to the placeholder issue

packages/react-workspace/react-client-app/src/components/App.tsx
```diff
@@ -66,13 +68,312 @@ import { SprightRectangle } from '@ni/spright-react/dist/esm/rectangle';
...
+    age: number;
+    hasChildren?: boolean;
+}
 
 export function App(): JSX.Element {
     const [bannerOpen, setBannerOpen] = useState(true);
     const [selectedRadio, setSelectedRadio] = useState('mango');
     const dialogRef = useRef<Dialog<string>>(null);
     const [dialogCloseReason, setDialogCloseReason] = useState('');
+    const drawerRef = useRef<Drawer>(null);
```
rajsite: think we can get bettter typing out of drawerRef like dailogRef

PR #2773, opened 2025-12-01
Eslint 9 Upgrade

rajsite:
# Pull Request

## 🤨 Rationale

Adopts eslint v9 🎉

## 👩‍💻 Implementation

- Switches to ni eslint config packages using flat config
- Removes `prettier-eslint` which is in [alpha builds](https://github.com/prettier/prettier-eslint/releases?q=v17&expanded=true) for eslint v9 support, significantly hurt lint performance, and had [major conflicts with eslint](https://github.com/ni/nimble/pull/2718#discussion_r2528336617). Now that eslint libraries like [stylistic](https://eslint.style/) are dedicated to style rules and eslint supports more formats such as [markdown syntax](https://github.com/eslint/markdown) and [markdown style](https://github.com/ota-meshi/eslint-plugin-markdown-preferences?tab=readme-ov-file) it's possible we don't need to rely on prettier long term. Either way we remove prettier for now to unblock eslint upgrades and can reconsider in the future
- Linting performance was significantly improved so simplified the `validate` concurrency scripts to run all lint sequentially in parallel to tests. Now `validate` completes ~100s faster. It's also possible we will get another speed bump enabling [multithreaded eslint](https://eslint.org/blog/2025/08/multithread-linting/) but will save for a follow-up PR.
- In the nimble private `eslint-config-nimble` package:
    - Created a shared `lintNimbleConfig` configuration that is used in each eslint config to make sure the eslint config file itself is linted and to give a spot for very general shared config in follow-up PRs (like markdown or json linting)
    - Created an angular lint similar to our typescript / component configs as shared config for the angular projects. This resolves issues where extending from a base config in the angular workspace is technically crossing a library boundary and triggered lint errors about reaching across package boundaries. Now we explicitly depend on a shared package.
- Angular workspace changes:
    -  The angular workspace root config is now for the config file in root, not the angular projects.
    -  Each angular project uses the shared `angularTypescriptNimbleConfig` and `angularTemplateNimbleConfig`
- Root workspace changes:
   - Added a top-level eslint config file for the config in root (the config file itself and beachball) but could also be used for linting markdown, etc. in the future.
   - Having a root eslint config also prevents the eslint vscode plugin from getting confused (errors in extension output logs) searching for the eslint config for files in root.

## 🧪 Testing

Note, see #2774 for CI results and source changes resulting from this config change.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. **A follow-up PR will remove prettier references and docs**

### DIFF
--- a/packages/angular-workspace/nimble-angular/label-provider/table/nimble-label-provider-table.directive.ts
+++ b/packages/angular-workspace/nimble-angular/label-provider/table/nimble-label-provider-table.directive.ts
@@ -51,7 +51,6 @@ export class NimbleLabelProviderTableDirective {
     }
 
     // Renaming because property should have camel casing, but attribute should not
-    // eslint-disable-next-line @angular-eslint/no-input-rename
     @Input('collapse-all') public set collapseAll(value: string | undefined) {
         this.renderer.setProperty(this.elementRef.nativeElement, 'collapseAll', value);
     }
--- a/packages/angular-workspace/nimble-angular/src/directives/dialog/nimble-dialog.directive.ts
+++ b/packages/angular-workspace/nimble-angular/src/directives/dialog/nimble-dialog.directive.ts
@@ -17,7 +17,6 @@ export { dialogTag };
 @Directive({
     selector: 'nimble-dialog'
 })
-// eslint-disable-next-line @typescript-eslint/no-invalid-void-type
 export class NimbleDialogDirective<CloseReason = void> {
     public get preventDismiss(): boolean {
         return this.elementRef.nativeElement.preventDismiss;
--- a/packages/angular-workspace/nimble-angular/src/directives/drawer/nimble-drawer.directive.ts
+++ b/packages/angular-workspace/nimble-angular/src/directives/drawer/nimble-drawer.directive.ts
@@ -19,7 +19,6 @@ export { DrawerLocation };
 @Directive({
     selector: 'nimble-drawer'
 })
-// eslint-disable-next-line @typescript-eslint/no-invalid-void-type
 export class NimbleDrawerDirective<CloseReason = void> {
     public get location(): DrawerLocation {
         return this.elementRef.nativeElement.location;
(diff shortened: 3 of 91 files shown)


packages/eslint-config-nimble/lint.js
```diff
@@ -0,0 +1,22 @@
...
+            '**/eslint.config.mts',
+            '**/eslint.config.ts'
+        ],
+        extends: [javascriptConfig, importNodeEsmConfig],
+        // Ignoring imports in each specific project eslint config
+        // so each package.json does not need explicit `eslint`
+        // and `eslint-config-nimble` dependencies.
+        rules: {
+            'import/no-extraneous-dependencies': 'off'
+        }
```
rajsite: I'm on the fence about this rule. What's its doing is ignoring the `package.json` dependency check specifically just for `eslint.config.*` files. Otherwise each package with eslint would need specific dependencies on `eslint` and the configs they depend on, i.e. `@ni/eslint-config-javascript` because each file imports from them.

As I write this I'm more convinced we should just do the dependency definitions correctly since packages have various purpose specific dependencies that should be tracked (react specific / storybook specific, etc). I'll keep it as a personal list background task follow-up PR unless someone feels strongly to do it in an immediate follow-on.

packages/eslint-config-nimble/components.js
```diff
@@ -27,6 +27,11 @@ export const componentsNimbleConfig = defineConfig([
                     paths: [...restrictedFastImports()],
                 },
             ],
+
+            // Strict rules components have not adopted yet
```
jattasNI: Am I reading this correctly that these are things that the NI styleguide recently started enforcing but we're disabling them temporarily due to time constraints / scoping? If so, could you create and link to a tech debt issue to track fixing the violations and removing this config.

jattasNI (approved):

rajsite:
Bypassing other owners for lint only changes

PR #2775, opened 2025-12-03
Fix link to Example Client App in README

### LINKED ISSUE #2776: Example Client App broken on public NPM page
# 🐛 Bug Report

The link to the Example Client App is broken on the NPM page.

## 💻 Repro or Code Sample

See link in first paragraph at https://www.npmjs.com/package/@ni/nimble-angular

## 🤔 Expected Behavior

It should open this page when clicked: https://github.com/ni/nimble/tree/main/packages/angular-workspace/example-client-app

## 😯 Current Behavior

It currently goes to this page and returns a 404: https://github.com/ni/nimble/blob/HEAD/packages/angular-workspace/nimble-angular/packages/angular-workspace/example-client-app 

## 💁 Possible Solution

Update the link

## 🔦 Context

I'm trying to figure out how to use everything appropriately. I knew enough to poke around and find it, but someone unfamiliar with the repo might not have.

## 🌍 Your Environment

Overcast and rather cold

fredvisser:
# Pull Request

## 🤨 Rationale

Fixes #2776 - A recent Renovate update added a directory field to the package.json for most components (a few already had it specified). 

While NPM [suggests](https://docs.npmjs.com/cli/v11/configuring-npm/package-json) that monorepos should use this field "if the package.json for your package is not in the root directory (for example if it is part of a monorepo), you can specify the directory in which it lives…", it breaks our absolute relative paths strategy.

## 👩‍💻 Implementation

Removed directory path from all package.json files

## 🧪 Testing

- [x] Confirmed that existing links continue to work in Github. 
- [ ] Confirm that NPM readme links are correct

## ✅ Checklist

### DIFF
--- a/change/@ni-jasmine-parameterized-91fbedc8-bb17-4b83-8981-d76b9d27474c.json
+++ b/change/@ni-jasmine-parameterized-91fbedc8-bb17-4b83-8981-d76b9d27474c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Remove directory path from package.json to resolve README linking issue",
+  "packageName": "@ni/jasmine-parameterized",
+  "email": "1458528+fredvisser@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@ni-nimble-angular-fd69ece0-50ac-4ed0-b744-8beadc14ed3c.json
+++ b/change/@ni-nimble-angular-fd69ece0-50ac-4ed0-b744-8beadc14ed3c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Remove directory path from package.json to resolve README linking issue",
+  "packageName": "@ni/nimble-angular",
+  "email": "1458528+fredvisser@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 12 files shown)


packages/angular-workspace/nimble-angular/README.md
```diff
@@ -10,7 +10,7 @@ NI-styled UI components for [Angular](https://angular.io) applications
 
 ## Getting started
 
-The steps to use components from Nimble Angular are similar to using components from any other Angular library. You can see the [Example Client App](/packages/angular-workspace/example-client-app) project for an example.
+The steps to use components from Nimble Angular are similar to using components from any other Angular library. You can see the [Example Client App](../example-client-app) project for an example.
```
rajsite: READMEs should use absolute relative paths: https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#documentation-policies

Can see that it worked in a slightly older package: https://www.npmjs.com/package/@ni/nimble-angular/v/30.9.1

Issue is that `directory` was added to the package.json but it doesn't serve a real purpose: https://github.com/ni/nimble/blob/main/packages/angular-workspace/nimble-angular/package.json#L12

Think we should just remove "directory" from all the `package.json` files so that links behave correctly on npmjs and github uis for rendering markdown.

PR #2780, opened 2025-12-05
Minimal turborepo integration

rajsite:
# Pull Request

## 🤨 Rationale

Before this change the top-level nimble commands behaved as follows:

- build: ran serially based on order of packages listed in the top-level workspace
- pack: ran serially based on order of packages listed in the top-level workspace
- validate (lint and test): hand crafted to run the following in parallel:
    - all packages running lint serially
    - nimble components test
    - all other packages running test serially

This PR does a bare minimum integration of turborepo to enable running more of those commands in parallel while staying closely aligned with normal npm workspace patterns, see: https://turborepo.com/docs/crafting-your-repository/configuring-tasks

## 👩‍💻 Implementation

This change does not change the result behavior of the top-level commands, just how they run:

- build: packages run build in parallel based on the dependency tree
    - It is defined in the top-level `package.json` as relying on two commands, `build` and `build-bundle` (split the component library `build` commands into `build` + `build-bundle`)
    - The only package / script that relies on the `build-bundle` output is blazor so specify that as a dependency of blazor
    - Now generating the bundle builds does not block the rest of the package builds
    - Splitting the build command is being done as an incremental step towards [splitting the bundle builds to their own package](https://github.com/ni/nimble/pull/827/files#diff-0967e47790c3f683cc13aff0bdc76ecdcf42ebe20f0d9aadcb89b4fe7bf2ed5aR36)
- validate / lint and test: packages all run in parallel as fast as possible (no interdependent relationships, still need to have run build first)

This change does add the following top-level helper commands:

- `npm run build-storybook`
- `npm run build-angular`
- `npm run build-blazor`
- `npm run build-react`

The commands will build a subset of the nimble monorepo just for the defined target. For example, `npm run build-storybook` will build all the dependencies needed to start storybook but not angular, blazor, etc. After storybook is built it can be started like before with `npm run storybook`.

The change also does the following:
- Packages that list in tree dependencies as `peerDependencies` also list them as `devDependencies`. This does not impact published packages and seems to be a direction turborepo folks want to stick to: https://github.com/vercel/turborepo/issues/1601
- Meta packages like the site package were updated to express their dependencies correctly, that's just a win!
- Scripts that are called from the top-level commands (`build`, `lint`, `test`) need to be in the same package as the `build` script for dependency tracking. It's subtle and normal packages won't have an issue, only came up in the blazor workspace where the main `pack` command needed to co-locate with the blazor `build` command. Most node packages won't have an issue with this so don't think it's worth documenting.

## 🧪 Testing

- Rely on CI, seems to shave 2ish minutes off the build as-is. The potential for [caching](https://turborepo.com/docs/guides/ci-vendors/github-actions#remote-caching-with-github-actionscache) in the future could do much more
- Take note of the build graphs added in the summary of the build job, super cool

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated the top-level contributing to callout the new commands for building a subset of the repo.


.github/workflows/main.yml
```diff
@@ -137,6 +138,9 @@ jobs:
           git config --global user.email "${{ env.GITHUB_SERVICE_EMAIL }}"
           npm run invoke-publish
 
+      - uses: charpeni/turborepo-summary-action@v1
+        if: always()
```
rajsite: Checkout the cool graphs this adds to build summary!: https://github.com/ni/nimble/actions/runs/19976693844

change/@ni-nimble-blazor-2c03062d-afdf-4ad5-8aa9-08fb6714071f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
rajsite: Doing patch builds to test: 
- nothing breaks with deployment
- beachball will handle the duplicated peer / dev deps introduced

rajsite (commented):
make sure each of lint/format/validate does top-level lint

turbo.json
```diff
@@ -0,0 +1,28 @@
...
+    "build-bundle": {
+      "dependsOn": ["build"],
+      "cache": false
+    },
+    "test": {
+      "cache": false
+    },
+    "lint": {
+      "cache": false
+    },
```
jattasNI: The test and lint commands implicitly depend on the build command of the same package, correct? What enforces that? Based on [Depending on tasks in the same package](https://turborepo.com/docs/crafting-your-repository/configuring-tasks#depending-on-tasks-in-the-same-package) I was expecting to see `"dependsOn": ["build"]` on these.

packages/nimble-components/package.json
```diff
@@ -37,7 +38,7 @@
     "test-webkit:debugger": "karma start karma.conf.js --browsers=WebkitDebugging --skip-tags SkipWebkit",
     "test-webkit:watch": "karma start karma.conf.headless.js --browsers=WebkitHeadless --skip-tags SkipWebkit --watch-extensions js",
     "test-webkit": "karma start karma.conf.headless.js --browsers=WebkitHeadless --single-run --skip-tags SkipWebkit",
-    "test-concurrent": "concurrently --pad-prefix --timings --group \"npm:test-chrome\" \"npm:test-webkit\""
+    "test": "concurrently --pad-prefix --timings --group \"npm:test-chrome\" \"npm:test-webkit\""
```
jattasNI: Longer term are `concurrently` and `turbo` complementary or will we replace instances of `concurrently` with `turbo`? Would this be equivalent to `turbo test-chrome test-webkit` or something like that?

jattasNI (approved):
Did some local sanity testing with your branch and everything worked fine. I didn't do a before and after compare but it felt super fast!

rajsite:
bypassing blazor owners (though we may want to consider reevaluating all our c# disables at some point fyi @msmithNI )

rajsite:
After merging to main and build failing realized I missed updating the `invoke-publish` command in the blazor packages to point to the newly renamed `invoke-pack` commands. Fixed that in a push to main + a bechball sync via `npm run sync`: https://github.com/ni/nimble/commit/a9c71d8b7757f732f74501d6422fcd591911a5bc

PR #2789, opened 2025-12-12
Switch to karma-vite

rajsite:
# Pull Request

## 🤨 Rationale

Uses vite instead of webpack for running karma tests. Has the following benefits:
- Fast enough to target `.ts` source files so can avoid needing to build manually first
- Starts running tests almost immediately (2-3 seconds) instead of waiting for tdd, i.e.  a components build followed by a full webpack build (30-40 seconds)
- Improves workflow of focus a specific set of tests and run tests
- Reduces times on `nimble-component:test` on CI by about 10-15s (net impact on full CI is minimal due to concurrency)

## 👩‍💻 Implementation

- Switch to karma-vite in karma config files
- Remove webpack deps from components
- Had to add overrides to use newer vite versions until it's addressed in karma-vite, see: https://github.com/credred/karma-vite/pull/17
  - Looks like overrides only have an impact on root package.json:  https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides
- Also added a `build-components` top-level helper command
- Also aligned the browsers we test against across packages

## 🧪 Testing

Try it yourself, pretty amazing.
- Sync repo and npm install
- `npm run build-components`
- Focus a test (i.e. `fdescribe`), for example in `nimble-components` (note that means source has changed since build)
- `npm run test-chrome`, should run against the latest source quickly and finish

Local run and CI

## ✅ Checklist

### DIFF
--- a/packages/storybook/package.json
+++ b/packages/storybook/package.json
@@ -30,7 +30,6 @@
     "@storybook/builder-vite": "^10.1.0",
     "@storybook/cli": "^10.1.0",
     "@storybook/html-vite": "^10.1.0",
-    "concurrently": "^9.2.1",
     "js-beautify": "^1.15.1",
     "react-syntax-highlighter": "^16.1.0",
     "remark-gfm": "^4.0.0",


packages/nimble-components/karma.conf.headless.js
```diff
@@ -2,12 +2,14 @@
  * A karma config file that sets up reporting for headless test runs
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
rajsite: After https://github.com/ni/javascript-styleguide/pull/184 this will be caught and removed as unneeded by eslint

change/@ni-nimble-components-4381b5e1-bbcc-48b4-9751-38f3ac8f2aa8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
rajsite: I'd also be okay marking all these as none since they are test changes. The tsconfig changes shouldn't have generated js impacts (but I didn't check)
Technically nimble components has a README change which is package contents so could be patch and dependent change type none.
Our just leave em as patch 🤷

PR #2799, opened 2025-12-18
HLD updates for new Spright chat capabilities: welcome and disclaimer

jattasNI:
# Pull Request

## 🤨 Rationale

We would like to unify as much of the chat implementation as practical across Angular and Blazor. These are two components that have Blazor implementations in an application codebase that we want in an Angular codebase too, so we want to introduce them to Spright: a welcome screen and a disclaimer at the bottom of the conversation.

(Older Figma designs for illustration only)

**Welcome**

chat-welcome-login

**Disclaimer**

chat-disclaimer

## 👩‍💻 Implementation

Update the chat component HLD with an implementation proposal for these components.

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/chat/specs/README.md
+++ b/packages/spright-components/src/chat/specs/README.md
@@ -184,11 +184,11 @@ richText.markdown = 'Welcome **Homer**, how can I help?';
 
 ```html
 <spright-chat-conversation>
-    <spright-chat-welcome-message title="Welcome to Nigel AI" subtitle="Log in to get started">
+    <spright-chat-message-welcome title="Welcome to Nigel AI" subtitle="Log in to get started">
             <nimble-anchor-button appearance="block" appearance-variant="primary" href="/login">
                 Login
             </nimble-anchor-button>
-    </spright-chat-welcome-message>
+    </spright-chat-message-welcome>
 </spright-chat-conversation>
 ```
 


jattasNI:
@mollykreis @rajsite heads up that I moved this out of draft and it's ready for your review. No intense time pressure, but I hope we'll have devs starting work on these components this month so I don't want to let it sit for too long.

rajsite:
FYI @jattasNI didn't get to it today and I'll be OOO rest of week, I don't expect to get to review till early next week

packages/spright-components/src/chat/specs/README.md
```diff
@@ -247,6 +305,18 @@ In the case of the auto-clearing being undesirable, a `resetInput()` method was
 
 [Some concern](https://github.com/ni/nimble/pull/2611#discussion_r2110130708) was raised with having the 'Send' button (or pressing \<Enter\>) automatically clearing the text, however there is enough basis to do so both in that this is designed behavior for the control (there is no expectation that the text field should maintain any text after the send event occurs), and we already have a similar UX semantic with the clear button for the `Select`.
 
+#### Disclaimer
+
+- _Component Name_ `spright-chat-disclaimer`
+- _Props/Attrs_
+    - `target` - an attribute which is forwarded to the link's `target` attribute. Defaults to `_self`.
```
mollykreis: This is likely a controversial comment, but I'm curious why you've picked a default of `_self`. I know that the general guidance is that links should open in the same tab, but it seems like viewing the terms and conditions of the AI from within a product would be a case where switching to ni.com (assuming that's where the terms and conditions are) would be pretty disruptive to what someone was trying to do (e.g. editing a resource in SystemLink). Would that be a case where opening by default in a new tab would be acceptable?

jattasNI: Interesting thought. I didn't put much consideration into that choice, just matching the [default value of `target` for `a` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#target).

Desktop apps would definitely [continue to want `target="_blank"`](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/MeasurementServices/AiAssistants/Controls/Components/ChatbotViewFooter.razor&version=GBmain&line=3&lineEnd=4&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) to avoid navigating within the embedded browser hosting Nigel. So the question is really what SystemLink would want.

I suspect you're right that most users SystemLink that click the link would prefer it in a new tab. But my first thought is that I don't believe that suspicion is enough to justify us ignoring / changing [the SystemLink pattern on opening links in new tabs](https://stratus-storybook.ni.dev/?path=/docs/patterns-links--docs#open-in-new-tab), which says to do that only if interrupting a complex workflow where data loss would be likely. I don't think the common case would be to click this while editing a resource. (To be honest, the common case will be never to click the link at all :-D)

So my vote is to keep the default as `_self` and for SystemLink to use that default. But I'll keep the thread open for Milan to weigh in.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -268,6 +338,16 @@ A message is simply a `div` which will styled with background / border / rounded
 </template>
 ```
 
+#### Message welcome content
+
+The template will include an `svg` element to render the image. The image requires different svg contents for dark and light themes (they use different gradient parameters). The gradient content will be specified in a new design token, `messageWelcomeContentGradient`, and the template will read the correct gradient value for the current theme using `messageWelcomeContentGradient.getValueFor()`.
+
+A `slot` element will be used to host the login button.
+
+Text content will be placed in `div` elements, conditionally shown if there is no slotted login button. We will use our standard pattern to detect whether there is slotted content (see #2579).
```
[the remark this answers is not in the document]
rajsite: Kinda similar I wonder if the post login content being kinda hard coded in nimble is too rigid. could imagine more instructions "type something or drop a test report to begin", etc being app specific. May even be buttons, etc to choose a file or something.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -146,6 +169,20 @@ const richText = document.querySelector('#welcome');
...
+
+```html
+<spright-chat-conversation>
+    <spright-chat-message message-type="system">
+        <spright-chat-welcome-message-content>
+            <nimble-anchor-button appearance="block" appearance-variant="primary" href="/login">
+                Login
+            </nimble-anchor-button>
+        </spright-chat-welcome-message-content>
+    </spright-chat-message>
```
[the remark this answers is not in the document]
jattasNI: I agree and addressed that tech debt for existing messages in #2835. I've updated this proposal to put welcome messages in a new message component instead of as content of an existing message component.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -247,6 +305,18 @@ In the case of the auto-clearing being undesirable, a `resetInput()` method was
 
 [Some concern](https://github.com/ni/nimble/pull/2611#discussion_r2110130708) was raised with having the 'Send' button (or pressing \<Enter\>) automatically clearing the text, however there is enough basis to do so both in that this is designed behavior for the control (there is no expectation that the text field should maintain any text after the send event occurs), and we already have a similar UX semantic with the clear button for the `Select`.
 
+#### Disclaimer
+
+- _Component Name_ `spright-chat-disclaimer`
```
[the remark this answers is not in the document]
jattasNI: Good thought. Updated the proposal to allow the disclaimer content to be slotted, with Spright providing the styles.

packages/spright-components/src/chat/specs/README.md
```diff
@@ -74,12 +75,22 @@ All end text buttons must meet the following criteria
 1. The `apperance` attribute is set to `block`
 1. The buttons only have text
 
+##### Welcome message features
+
+1. Centered horizontally within a conversation, similar to system messages
+1. Displays a fixed image to brand the chat experience
```
OliverCosma: could it be a simple icon as well?

OliverCosma: Yep, I was just curious if 1 could be sometime needed/possible, it isn't a "requirement" or specific use case :)

PR #2800, opened 2025-12-19
Added spec for Spright chat conversation scroll behavior

aangerhofer:
# Pull Request

## 🤨 Rationale

We need to define scrolling behavior for the chat for [this feature](https://ni.visualstudio.com/DevCentral/_workitems/edit/3649742)

## 👩‍💻 Implementation
Added a MD Spec

## 🧪 Testing
Reviewed what other chats do

## ✅ Checklist


hellovolcano (approved):

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Types of Scrolling in AI Chat
+
+**User-initiated scrolling:**
+User initiated scrolling is done by the user to view previous chats in the message area. User scrolling can be initiated with the mouse, keyboard, or clicking on the scrollbar.
+-**Manual interruption:** A specific type of user scrolling that takes place when the user scrolls while AI is generating response to stop auto-scroll.
+
+**System-initiated scrolling:**
+- **Post send auto-scroll:** Automatic scrolling to the user's next message when the user sends a message
+- **AI response auto-scroll:** Scrolling as AI generates and streams response content
```
jattasNI: nit: some of these lists aren't rendering correctly in the markdown preview view ("Display the rich diff" button in the top right of this file on the PR). Here's my stab at fixing but I didn't validate if it works.

```suggestion
**User-initiated scrolling:**

User initiated scrolling is done by the user to view previous chats in the message area. User scrolling can be initiated with the mouse, keyboard, or clicking on the scrollbar.
- **Manual interruption:** A specific type of user scrolling that takes place when the user scrolls while AI is generating response to stop auto-scroll.

**System-initiated scrolling:**

- **Post send auto-scroll:** Automatic scrolling to the user's next message when the user sends a message
- **AI response auto-scroll:** Scrolling as AI generates and streams response content
```

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Auto-scrolling behavior
+
+**Post send auto-scroll**
+Post send auto-scrolling should appear any time the user sends a new message (regardless of where they are in the chat history at the time)
```
jattasNI:
```suggestion
Post send auto-scrolling should occur any time the user sends a new message (regardless of where they are in the chat history at the time)
```

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+
+**AI response auto-scroll**
+AI response auto-scroll should occur only if the AI streams responses. Long responses should auto scroll because streaming should be happening at a human readable pace and therefore start removing the old (assumed read) content off screen. If we are not streaming the response and instead returning a block of text, we should not autoscroll because this would take unread content off the screen.
```
jattasNI, line 29: From an accessibility perspective, I suspect it's problematic to assume a reading pace. And for SystemLink at least, we're currently trying to stream as fast as the responses come back since we felt that artificially limiting the response rate might be frustrating, especially if someone reads quickly or cares most about the end of the response.

I'm not sure what we might do to resolve that. Our mitigation of stopping auto-scroll when a manual scroll happens is probably sufficient to start. I guess maybe we could provide the user with a way to control the streaming rate, though that feels low priority. Or we could not auto-scroll for long messages, just keep the top of the message visible and let the user scroll to bottom.

Just pointing out in case you have ideas. I'm fine with the current approach.

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Non-Mouse Interactions
+
+### Keyboard Navigation
+- **Page Up/Down in message area:** Navigate through message history, disables auto-scroll
+- **Tab navigation:** Focus management doesn't affect scroll position for on screen elements but if we begin to tab upwards in the chat to nonvisible content we should scroll to it
+
+### Accessibility Considerations
+- **Screen readers:** Announce new messages without forcing scroll position changes
+- **Focus management:** Maintain logical tab order regardless of scroll position
```
jattasNI: nit: I'm not sure what this is trying to convey. What's the alternative to maintaining a logical tab order, randomly rearranging items as you scroll? 😆 I suspect this came from an AI that really wanted more than one bullet point and thought that tab order was a good thing to talk about. I'd vote to either clarify what it means or delete it.

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+- The Nigel chat needs defined scrolling behavior to balance automatic updates with user control
+
+### Types of Scrolling in AI Chat
+
+**User-initiated scrolling:**
+User initiated scrolling is done by the user to view previous chats in the message area. User scrolling can be initiated with the mouse, keyboard, or clicking on the scrollbar.
+-**Manual interruption:** A specific type of user scrolling that takes place when the user scrolls while AI is generating response to stop auto-scroll.
+
+**System-initiated scrolling:**
+- **Post send auto-scroll:** Automatic scrolling to the user's next message when the user sends a message
```
jattasNI, line 18: It'd be good to have the next level of detail about this behavior. I believe the proposal is something like:
1. scroll so that the top of the user's message is at the top of the visible conversation, adding whitespace below as necessary
2. if the user's message is taller than the height of the visible conversation, show the end of the message (last few lines? as much as possible?)

packages/spright-components/src/chat/specs/scrolling-behavior.md
```diff
@@ -0,0 +1,52 @@
...
+- **AI response auto-scroll:** Scrolling as AI generates and streams response content
+
+## Auto-scrolling behavior
```
jattasNI, line 21: Might be worth mentioning that scrolling should animate when possible (["smooth" rather than "instant"](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll#behavior), to use MDN's terminology)

jattasNI (approved):

packages/spright-components/src/chat/specs/scrolling-behavior.md
aangerhofer: I'm gonna say show the last 3 lines.

packages/spright-components/src/chat/specs/scrolling-behavior.md
aangerhofer: I don't have ideas I feel strongly about. I think we still want to autoscroll for large messages, at least for things like code generation because people want to get to the summary.

Probably a larger UX thing though where if we people constantly just wanting to be at the bottom, we should focus on shorter messages?

I'm leaving as is for now

packages/spright-components/src/chat/specs/scrolling-behavior.md
aangerhofer: I'll clarify, the idea was if the user was already focused on something, not to lose that just because of scrolling

PR #2810, opened 2026-01-01
Switch to package exports for React packages

rajsite:
# Pull Request

## 🤨 Rationale

Switches to package exports in the react packages so  `dist/esm` can be removed from the package names.

## 👩‍💻 Implementation

- Updated package.json exports config and switched to package type module
- Unrelated refactor: since we are on a modern node version which can import cjs from esm was able to remove the build step around the icon generation script, this can be done in the other packages but didn't do in this PR
- Also included the styles scripts in the built packages 😅

## 🧪 Testing

- The normal build exercises `moduleResolution: bundler` in an example app and added a test `tsconfig.legacy.json` to exercise `moduleResolution: node`.
- Checked the storybook `tsconfig.tsbuildinfo` and `preview-stats.json` for regressions on imports resolving to `dist` instead of `src` for turbopack

## ✅ Checklist

### DIFF
--- a/packages/react-workspace/react-client-app/package.json
+++ b/packages/react-workspace/react-client-app/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "start": "vite",
     "build": "tsc -b && vite build",
+    "test": "tsc -p ./tsconfig.legacy.json",
     "lint": "eslint .",
     "format": "eslint . --fix",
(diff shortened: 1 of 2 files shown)


rajsite:
@jattasNI bypassing for nights and weekends update to react package exports, can address feedback in follow-up

PR #2813, opened 2026-01-08
Support Non-Nimble icons in the Nimble table

hellovolcano:
# Pull Request

## 🤨 Rationale

Clients need a way to display non-nimble icons, specifically SVG and PNG, in a Nimble table. This HLD proposes adding additional mappings for an SVG and PNG so they can be displayed.

## 👩‍💻 Implementation
N/A

## 🧪 Testing

N/A

## ✅ Checklist


packages/ok-components/src/mapping/custom-image/specs/table-column-mapping-custom-icons-hld.md
```diff
@@ -0,0 +1,126 @@
...
+- `src`: string - URL for custom image
+- `severity`: IconSeverity - Controls color theming (default, error, warning, success, information)
+- `text`: string - Tooltip and display text
+- `text-hidden`: boolean - Controls text visibility
+
+**Supporting Infrastructure**
+
+- Add **`MappingCustomImageConfig`** following existing `MappingIconConfig` pattern
+- Custom ViewTemplates for rendering images with proper styling and accessibility
+- Extend existing `TableColumnMappingCellView` and `TableColumnMappingGroupHeaderView` to handle new mapping type
```
rajsite: This implies a dependency from nimble-components down towards spright-components/ok-components which is a circular dependency pattern we won't support.

For the direction of customizing the set of mappings, options are:

- refactor the mapping column to allow for arbitrary mappings. This would be a non-trivial refactor.
- create a new context specific mapping column duplicated from nimble mapping column that supports additional mapping types. This is more feasible (mapping column is already designed with the intention of multiple implementations built on the EnumBase class) but is quite a bit to maintain.

Another direction is make it possible to register new types of icons and then re-use the existing mapping column as is. That's quickly prototyped here: https://github.com/ni/nimble/pull/2814

hellovolcano: Updated the HLD to use the prototype

change/@ni-nimble-components-e24ccafe-f3e9-4a34-bd19-7696fb0d00d1.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI: You can delete this change file since you're no longer changing nimble-components.

PR #2826, opened 2026-01-23
HLD Add Breakpoint table column to Ok

hellovolcano:
# Pull Request

## 🤨 Rationale

Clients need a way to provide a breakpoint in the nimble table. This is the implementation of the breakpoint [IxD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table-column/breakpoint/IxD.md)

## 👩‍💻 Implementation

N/A

## 🧪 Testing

N/A

## ✅ Checklist


packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Column Element
+
+_Element Name_
+
+- `spright-table-column-breakpoint`
+
+_Props/Attrs_
+
+- `breakpoint-field-name`: string - The field name in the data record that contains the breakpoint state
```
jattasNI: For other table columns, we typically just call the attribute `field-name` unless we know the column requires data from multiple fields (like the anchor column with `href-field-name` and `label-field-name`. Unless you're aware of something coming from the breakpoint column that will require another field, I'd vote to rename this to `field-name` too.

packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,161 @@
...
+- Breakpoint indicators will have proper ARIA labels describing current state
+- Keyboard navigation support with Tab/Shift+Tab
+- Space/Enter key activation for toggling breakpoints
+- Tooltips will be announced by screen readers
+- High contrast support for different breakpoint states
+- Minimum 24x24 pixel hit targets for web accessibility
+
+### Angular Integration
+
+- Add new table column and events to Angular and Blazor wrappers.
```
jattasNI: nit: feels weird to mention a wrapper for the event but not the rest of the API.

Also, if building in Ok you're welcome to skip the Angular wrapper if you only need Blazor. If you want to go fully complete then you should add a React wrapper too.

```suggestion
- Standard Angular and Blazor wrappers.
```

packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,161 @@
...
+### Background
+
+[IxD](https://github.com/ni/nimble/blob/main/packages/nimble-components/src/table-column/breakpoint/IxD.md)
+
+### Features
+
+- Visual breakpoint indicators with multiple states (Off, Enabled, Disabled, Hit)
+- Click-to-toggle functionality for setting/removing breakpoints
+- Callback mechanisms for client-implemented breakpoint operations (like row highlighting and context menu, as specified in IxD)
+- Keyboard shortcuts for breakpoint management (Ctrl+B, F9)
```
jattasNI: Heads up that keyboard shortcuts might get complicated. For example I'm not sure what would be responsible for delivering the key events to that specific column if the entire row is focused. If this is an ok component we could figure it out in implementation and consider cutting that capability. If this is a spright component then we probably want a little more detail about how this will work in the spec.

packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Placeholder
+
+When the breakpoint state is `undefined` or `null`, the cell will render in the "Off" state (no visible indicator until hover). 
+
+### Delegated Events
+
+The following events from the cell view will be delegated to the column:
+
+- `click` events from the breakpoint indicator will be delegated as `breakpoint-toggle` events
```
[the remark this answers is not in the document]
hellovolcano: renamed event

packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,161 @@
...
+- Verify tooltip display on hover/focus
+- Verify proper event delegation with correct recordId values
+- Test that column appears leftmost in table layout
+- Verify non-resizable behavior
+- Test with various data states (undefined, null, invalid values)
+- Verify callback events provide proper context for client implementations
+
+### Internationalization
+
+Tooltips will use the standard HTML `title` attribute, with built-in default text ("Add breakpoint" / "Remove breakpoint"). If applications need custom tooltip text, they can override the `title` attribute on the column element. This follows the same pattern as other nimble components.
```
jattasNI: Instead of using the title attribute we should use a [Label Provider](https://nimble.ni.dev/storybook/?path=/docs/tokens-label-providers--docs). That's the standard pattern for text that Nimble provides a default value but allows clients to override.

packages/spright-components/src/table-column/breakpoint/specs/table-column-breakpoint-hld.md
```diff
@@ -0,0 +1,160 @@
...
+- Verify tooltip display on hover/focus
+- Verify proper event delegation with correct recordId values
+- Test that column appears leftmost in table layout
+- Verify non-resizable behavior
+- Test with various data states (undefined, null, invalid values)
+- Verify callback events provide proper context for client implementations
+
+### Internationalization
+
+Tooltips will use label provider strings for the default text ("Add breakpoint" / "Remove breakpoint").
```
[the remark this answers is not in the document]
jattasNI: Correct, all of our existing label providers are in Nimble. I was envisioning that we could add a new one to OK without too much trouble but haven't verified that to be true. If there are major barriers, we could add a new one to Nimble that contains all the OK labels. That would be conceptually similar to how other tokens like icons and colors are defined in nimble-tokens even if they're only used by Spright/OK.

Assuming the new label provider goes in OK we'd have to decide on its name / organization structure. I think I'd advocate for having one per group of OK components like we do for Nimble, so naming this one something like LabelProviderTableBreakpoint.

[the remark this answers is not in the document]
hellovolcano: > Just to double check, are we expecting this in spright components or ok components (

For some reason my title change didn't save, but targeting Ok. The spec is still in spright for now because I wanted to be able to diff it, but I will move to Ok once I'm done making changes to the doc.

PR #2827, opened 2026-01-23
Add Icon dynamic

rajsite:
# Pull Request

## 🤨 Rationale

Clients need a supported way to use custom icons in Nimble components. This change introduces the dynamic icon component so icons can be registered at runtime.

## 👩‍💻 Implementation

- Added the dynamic icon component and registration.
- Added Blazor helper/interop support for registration.
- Added Storybook story and docs for usage.

## 🧪 Testing

- Added unit tests to verify javascript helper method was called
- Added ok component unit tests

## ✅ Checklist


packages/ok-components/src/icon-dynamic/index.ts
```diff
@@ -0,0 +1,63 @@
...
+ * <nimble-mapping-icon icon="ok-icon-dynamic-awesome"></nimble-mapping-icon>
+ * ```
+ */
+export class IconDynamic extends Icon {
+    public constructor(/** @internal */ public readonly url: string) {
+        super();
+    }
+
+    public static registerIconDynamic(name: string, url: string): void {
+        if (!/^[a-z][a-z]+$/.test(name)) {
```
rajsite: This would be a minor tweak but I'd like to suggest it to make more consistency in the API surface.

In the spots we consume the icon we use the full tag name:
```html
<NimbleMappingIcon Icon="ok-icon-dynamic-awesome"></NimbleMappingIcon>
<!-- in the future, something like -->
<NimbleButton><OkIconViewer  Icon="ok-icon-dynamic-awesome" Slot="start"></NimbleButton>
```

So lets require that the input name match that format, i.e.:
```cs
 await OkIconDynamic.RegisterIconDynamicAsync(JSRuntime!, "ok-icon-dynamic-awesome", "https://myimage");
```

So we add a check that the string tagName starts with `ok-icon-dynamic-`, throw an error if it doesn't, take the last part of the string as the name and continue.

This way at the app level one only has to think about the one tag name string and not different possible ones.

rajsite (commented):
@hellovolcano looks good! I pushed some storybook doc tweaks and minor request to update the status table to include dynamic icon.
Ready to merge after @jattasNI gives it a lookover

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
```diff
@@ -273,7 +273,7 @@
                     <NimbleMappingIcon
                         Key=@("success")
                         Text="Success"
-                        Icon="nimble-icon-check"
+                        Icon="ok-icon-dynamic-awesome"
```
jattasNI: It's good that you tested locally that the dynamic icon works in a mapping column, but I don't think we should make this change to the example app in main. It causes this app to diverge from the Angular / React ones and it adds a dependency from this Nimble example onto Ok.

I'd be fine with either:
1. reverting this change
2. reverting this change and adding a new dynamic icon example down below in the Ok section. That example could be as simple or complex as you like, eg using the icon in a table like this or just slotted in a button

jattasNI (approved):

packages/blazor-workspace/Examples/Demo.Shared/Pages/ComponentsDemo.razor
hellovolcano: Reverted the change

PR #2830, opened 2026-01-26
Nimble Stepper HLD

rajsite:
# Pull Request

## 🤨 Rationale

- HLD for stepper and associated child components
- Also minor tweaks to templates to remove `<hr>` sectioning (can lean into headers / TOCs if want more sections) and added an "Interactions" section as an analog to the "Visual Appearance" section

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/stepper/specs/README.md
+++ b/packages/nimble-components/src/stepper/specs/README.md
@@ -106,8 +106,8 @@ The `step` elements will primarily render a provided nimble icon (and new nimble
     - disabled: boolean (standard, i.e. visually disabled and interactions prevented)
     - readonly: boolean (standard, i.e. visually not disabled, interactions other than tab focus prevented)
     - selected: boolean (visual change only, no behavior) (aligned with card button)
-    - severity: default / error / warning / success / information (aligned with icon)
-    - severity-text: string (if provided, only renders during warning, error, and information severity)
+    - severity: default / error / warning / success (similar to icon and banner but different subset)
+    - severity-text: string (always shown but styled based on severity)
   <!--
   - Properties (not attribute reflected)
   - Methods
@@ -259,7 +259,7 @@ No additional support beyond discussion in API section on CSS sizing.
 - *Localization* -->
 
 Label providers for visibly hidden step states:
-- Reuse `popupIconError`, `popupIconInformation`, `popupIconWarning`
+- Reuse `popupIconError`, `popupIconWarning`
 - Add (for consistency but maybe unexpected naming): `popupIconComplete`, `popupIconCurrent` 
 
 ### Security


packages/nimble-components/src/stepper/specs/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+- `nimble-stepper`
+  - Attributes
+    - orientation: vertical / horizontal (aligned with radio group and wafer)
+  <!-- - Properties (not attribute reflected)
+  - Methods
+  - Events
+  - CSS custom properties -->
+  - CSS native properties
+    - Will respond to width / height sizing (on the axis corresponding to orientation) and show overflow scroll buttons following the pattern of breadcrumb
```
rajsite: fyi, fred this concept of overflow scrollers like breadcrumb / tabs is not captured in the visual design. curious if brandon would have any concerns. In particular the vertical orientation having the scroll buttons would be new to this component

packages/nimble-components/src/stepper/specs/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+The `nimble-stepper` acts a a progress indicator for a wizard / step-by-step workflow. It behaves conceptually as either a collection of card buttons (i.e. collection of `nimble-step`) or as breadcrumbs (i.e. collection of `nimble-anchor-step`). Each item has a standard `severity` state associated with it and the ability to express that a state is the "current" `selected` step.
+
+The `nimble-stepper` is just for layout, placing steps either horizontal or vertical orientation and communicating internal state to child steps as needed (ideally just via style but implementation TBD).
+
+The `nimble-step` and `nimble-anchor-step` are elements representing individual steps with `nimble-step` behaving as a card button (i.e. a button with a concept of a `selected` visual appearance that does not change behavior) and a `nimble-anchor-step` looking visually identical but with link behaviors.
+
+The `step` elements will primarily render a provided nimble icon (and new nimble icons for the visuals of digits 0 - 9 will be added). When a non-default severity is provided the provided icon will be replaced with an icon representing the severity.
```
rajsite: fyi @fredvisser we would need these digit icons as requests to brandon

packages/nimble-components/src/stepper/specs/README.md
```diff
@@ -0,0 +1,293 @@
...
+  <!-- - Parts
+  - Localizable labels -->
+
+- `nimble-step`
+  - Attributes
+    - disabled: boolean (standard, i.e. visually disabled and interactions prevented)
+    - readonly: boolean (standard, i.e. visually not disabled, interactions other than tab focus prevented)
+    - selected: boolean (visual change only, no behavior) (aligned with card button)
+    - severity: default / error / warning / success / information (aligned with icon)
+    - severity-text: string (if provided, only renders during warning, error, and information severity)
```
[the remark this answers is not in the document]
mollykreis: Off the cuff, I could see us adding a parallel attribute to `severity-text` named `helper-text`. Alternatively, we could rename `severity-text` to something more generic and have the client update the text based on the value of `severity`.

packages/nimble-components/src/stepper/specs/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+No particularly interesting implementation concerns. Follows existing patterns around buttons and anchors.
+
+### States
+
+<!-- *Key component states, valid state transitions, and how interactions trigger a state transition.* -->
+
+Some specific usage notes:
+- Only one step should be marked selected at a time
+- Disabled steps represent a step that doesn't apply for the current workflow (previous step configuration disabled a future step which is now skipped)
```
jattasNI: Several of the linked examples from other design systems use a pattern where later steps aren't clickable until the current step has been completed. Can you clarify if those later steps would be represented as disabled or readonly? I suspect the answer is disabled and I suspect that will be the most common use case for disabled, but want to double check.

packages/nimble-components/src/stepper/specs/README.md
```diff
@@ -0,0 +1,293 @@
...
+- Readonly steps are steps that don't have associated views for them (it's a step that just indicates something like disk formatted or software installed)
+
+### Accessibility
```
jattasNI: The [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) attribute looks relevant. From my reading of the docs we could set `aria-current=step` when the client sets a given step to `selected` (and clear it when they select a different step).

packages/nimble-components/src/stepper/specs/README.md
rajsite: Moved to other thread: https://github.com/ni/nimble/pull/2830#discussion_r2761165770

PR #2832, opened 2026-01-27
Updated Styling of Switch to Match Visual Design Spec

aangerhofer:
# Pull Request

## 🤨 Rationale

The switch didn't look like our figma file. We are going to use it in the AI chat so I wanted to update it.

## 👩‍💻 Implementation

Looked at Brandon's styling & made the colors match

## 🧪 Testing

Ran a linter. 

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. purely visual changes, no function

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6739,18 +6739,18 @@
       "license": "MIT"
     },
     "node_modules/@ni/fast-element": {
-      "version": "10.0.5",
-      "resolved": "https://registry.npmjs.org/@ni/fast-element/-/fast-element-10.0.5.tgz",
-      "integrity": "sha512-fT4sFiWUcE2VgHhLYOijXhTBS1r+tq7IHj+9c5cE16iWfPhHP1GLzRGm/PrjIL+WZsFOQJSu5zKrDt1ysfDGzw==",
+      "version": "10.1.0",
+      "resolved": "https://registry.npmjs.org/@ni/fast-element/-/fast-element-10.1.0.tgz",
+      "integrity": "sha512-gFqkNPOJFEGFo05V1mKbVw/3n1eU2b1hAVO3HNczdnqAderVlxW+kaEI5NLiH8pdyk3OFbyAjEfURpDWGlr5Pg==",
       "license": "MIT"
     },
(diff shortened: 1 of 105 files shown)


packages/nimble-components/src/switch/styles.ts
```diff
@@ -86,6 +88,21 @@ export const styles = css`
...
+                    0.3
+                )};
+    }
+
+    :host([aria-checked='true']:not([disabled])) .switch {
+        background-color: var(--ni-private-switch-background-selected-color);
+    }
+
+    :host([aria-checked='true']:hover:not([disabled])) .switch {
+        background-color: var(--ni-private-switch-background-selected-rollover-color);
```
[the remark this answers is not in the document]
aangerhofer: Ok I met with Brandon. We are going to leave the changes as is. We explored some other options but as is made the most sense with the visual direction he wanted for these controls. If you want the longer explanation feel free to set up a call with me

packages/nimble-components/src/switch/styles.ts
```diff
@@ -165,64 +214,117 @@ export const styles = css`
         Theme.light,
         css`
             :host {
+                --ni-private-switch-background-color: ${hexToRgbaCssColor(Black91, 0.1)};
```
jattasNI: note for other reviewers/posterity: I poked around at our existing token values and didn't see any that matched these colors. Since they're all specific to the switch, I think it makes sense for now to define them as private variables within this file rather than try to expose public tokens. If we find other controls that want these colors then maybe we can refactor to new public tokens.

packages/nimble-components/src/switch/styles.ts
```diff
@@ -103,51 +121,82 @@ export const styles = css`
...
         border-radius: calc(var(--ni-private-switch-indicator-size) / 2);
-        margin: calc(
-            calc(
-                    var(--ni-private-switch-height) - var(
-                            --ni-private-switch-indicator-size
-                        )
-                ) /
-                2
-        );
+        margin: -2px;
```
rajsite: Can we represent this value with a named token so we know what it represents?

PR #2838, opened 2026-01-30
Add character limit to chat input

Alexia-Claudia-Micu:
# Pull Request
https://ni.visualstudio.com/DevCentral/_workitems/edit/3724380

## 🤨 Rationale

Nigel chat crashes when an input that is too large is introduced. We should silently truncate the input to a configurable number.

## 👩‍💻 Implementation

Add `maxlength` to `<textarea>` and bind it to `ChatInput`.

## 🧪 Testing

Manual testing on the storybook and added/updated unit tests.

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/chat/input/index.ts
+++ b/packages/spright-components/src/chat/input/index.ts
@@ -28,7 +28,7 @@ export class ChatInput extends FoundationElement {
     public override tabIndex!: number;
 
     @attr({ attribute: 'maxlength', converter: nullableNumberConverter })
-    public maxlength?: number = -1;
+    public maxLength?: number = -1;
 
     /**
      * @internal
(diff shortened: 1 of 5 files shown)


packages/spright-components/src/chat/input/index.ts
```diff
@@ -27,6 +27,9 @@ export class ChatInput extends FoundationElement {
     @attr({ attribute: 'tabindex', converter: nullableNumberConverter })
     public override tabIndex!: number;
 
+    @attr({ attribute: 'maxlength', converter: nullableNumberConverter })
+    public maxlength?: number = -1;
```
jattasNI: It turns out that the [native HTML attribute is called `maxlength`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/maxlength) but [the native class property is called `maxLength` (capital "L")](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/maxLength). I'm going to be annoying and say we should match that capitalization. Sorry for not catching this sooner! I'll leave comments on the classes impacted but could you update the tests to match?

I don't think the story file needs to change: our standard is to document the attribute names there.

```suggestion
    public maxLength?: number = -1;
```

Alexia-Claudia-Micu: Renamed it.

jattasNI (approved):

PR #2858, opened 2026-02-09
Add visible error for chat input

Alexia-Claudia-Micu:
# Pull Request
https://dev.azure.com/ni/DevCentral/_workitems/edit/3739259

## 🤨 Rationale

Chat input should have an option to show an error.

## 👩‍💻 Implementation

Modify `ChatInput` to extend `mixinErrorPattern`.
Add getters and setters for `errorText` and `errorVisible` in `spright-angular`.

Added the error pattern similarly to the one in `TextArea` nimble component https://github.com/ni/nimble/blob/main/packages/nimble-components/src/text-area/index.ts.

## 🧪 Testing

Added unit tests for `spright-angular` and updated matrix tests.

## ✅ Checklist


packages/angular-workspace/spright-angular/chat/input/spright-chat-input.directive.ts
```diff
@@ -46,5 +46,21 @@ export class SprightChatInputDirective {
...
+
+    @Input('error-text') public set errorText(value: string | undefined) {
+        this.renderer.setProperty(this.elementRef.nativeElement, 'errorText', value);
+    }
+
+    public get errorVisible(): boolean {
+        return this.elementRef.nativeElement.errorVisible;
+    }
+
+    @Input('error-visible') public set errorVisible(value: boolean | '' | null) {
```
OliverCosma: could we use `value: BooleanValueOrAttribute` ?

packages/spright-components/src/chat/input/styles.ts
```diff
@@ -23,6 +28,10 @@ export const styles = css`
         --ni-private-hover-indicator-width: calc(${borderWidth} + 1px);
     }
 
+    :host([error-visible]) {
+        margin-bottom: calc(${errorTextFontLineHeight} + 2px);
```
OliverCosma: do we need this ?

OliverCosma (changes requested):
can you also add an implementation example to the custom app ?

codrutaserban (approved):

packages/spright-components/src/chat/input/template.ts
```diff
@@ -1,19 +1,28 @@
...
+        <textarea
+            ${ref('textArea')}
+            placeholder="${x => x.placeholder}"
+            rows="1"
+            tabindex="${x => x.tabIndex}"
+            maxlength="${x => x.maxLength}"
+            @keydown="${(x, c) => x.textAreaKeydownHandler(c.event as KeyboardEvent)}"
+            @input="${x => x.textAreaInputHandler()}"
+        ></textarea>
+        <${iconExclamationMarkTag}
```
jattasNI: The way we add the exclamation mark icon doesn't look quite right:
1. we're adding it in both the template and the `end` property in `index.ts`, but all other components that implement the error pattern do it in only one of those places
2. we're adding the `scrollbar-width-calculated` class unconditionally but not setting any styles related to it

The goal of this code is to shift the icon to the left when the scrollbar appears so that it doesn't get covered by the scrollbar appears. I wasn't able to validate that because I'm on a mac which has a virtual scrollbar whose width is always 0.

I'd suggest trying to copy the text area implementation more closely: include its styles for `.scrollbar-width-calculated` and only add the icon in the template, not index.ts. And then you can check if the way the icon shifts on Windows for the chat input matches the way it shifts in the nimble text area story

change/@ni-spright-angular-dd14a85b-18a2-4830-af7d-297d213803c4.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI: We talked in our sync today about how a client app would use the new error attributes to reflect that the character count exceeds the limit. We agreed that it shouldn't be tied to `maxlength`. But I think it would be useful for the chat input to expose a getter that contains the current character count. That would ensure that clients use a consistent counting method (it can [get very complicated](https://docs.x.com/fundamentals/counting-characters)). 

@OliverCosma do you agree we should do that? If so, I'd be fine if it happened in a follow up PR. Once we have that we could consider removing `maxlength` too.

If we do this we could consider using the browser's [Intl.Segmenter object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) to count the "graphemes" (like letters/characters but account for emojis and other wide characters).

jattasNI:
Once you've resolved the last comments I can complete the PR for you, I don't think we need @mollykreis to review this one as it's pretty trivial Angular integration.

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.ts
```diff
@@ -298,7 +301,15 @@ export class CustomAppComponent implements AfterViewInit {
 
     public onChatInputSend(e: Event): void {
         const chatInputSendEvent = (e as CustomEvent<ChatInputSendEventDetail>);
-        this.chatUserMessages.push(chatInputSendEvent.detail.text);
+        const text = chatInputSendEvent.detail.text;
+        if (text === 'error') {
```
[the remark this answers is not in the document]
Alexia-Claudia-Micu: Removed it.

jattasNI:
Bypassing @rajsite and @mollykreis for trivial Angular changes

change/@ni-spright-angular-dd14a85b-18a2-4830-af7d-297d213803c4.json
rajsite: Another interesting example I ran across, bluesky uses a character limit algorithm of `min(3k bytes, 300 graphemes)` for the length of posts and someone made a calculator for it: https://data.runhello.com/bs-limits/
The single grapheme (i.e. a single "character" selectable with a mouse): 👨‍👩‍👧‍👦
Is 7 unicode codepoints, a string length 11 in JS, or 25 bytes lol

image

So +1 for exposing utilities to let apps decided what they are actually looking for and generic error-text system to report their restrictions back to the user

PR #2861, opened 2026-02-11
Add spright custom colored icons for calendar work item types

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Nimble icons currently do not have a way to change colors based on the requirements, especially when used within Nimble table components. While Nimble provides color override mechanisms, these cannot be applied when icons are used in table cells through the mapping-icon component.

This is an application-specific requirement, so utilizing the `Spright` to register the icons for different types of work items.

## 👩‍💻 Implementation

1. Extends `IconSvg` for Nimble table compatibility
2. Imports SVG data from `nimble-tokens` to reuse existing assets
3. Applies custom styles with specific calendar event color design tokens
4. Registers with spright prefix via FAST DesignSystem
5. Exports class and tag constant for type-safe usage
6. Add an icon-metadata file for the storybook similar to the Nimble
8. Create 5 custom colored event icons that are specific to work items:
    1. work-item-calendar-week
    2. work-item-calipers
    3. work-item-forklift
    4. work-item-rectangle-check-lines
    5. work-item-wrench-hammer
9. Add Angular directives for all the above icons to integrate with the Angular application.
    1. Export the standalone directives from the secondary entry points
10. Add to the example Angular app

## 🧪 Testing

Verified from the storybook build.

## ✅ Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31502,7 +31502,7 @@
     },
     "packages/angular-workspace/ok-angular": {
       "name": "@ni/ok-angular",
-      "version": "1.1.20",
+      "version": "1.1.21",
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.2.0"
@@ -31511,12 +31511,12 @@
         "@angular/common": "^19.2.17",
         "@angular/core": "^19.2.17",
         "@angular/localize": "^19.2.17",
-        "@ni/ok-components": "^0.2.19"
+        "@ni/ok-components": "^0.2.20"
       }
     },
     "packages/angular-workspace/spright-angular": {
       "name": "@ni/spright-angular",
-      "version": "8.6.0",
+      "version": "8.6.1",
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.2.0"
@@ -31525,7 +31525,7 @@
         "@angular/common": "^19.2.17",
         "@angular/core": "^19.2.17",
         "@angular/localize": "^19.2.17",
-        "@ni/spright-components": "^6.11.0"
+        "@ni/spright-components": "^6.12.0"
       }
     },
     "packages/blazor-workspace": {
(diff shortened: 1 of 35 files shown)


packages/spright-components/src/all-components.ts
```diff
@@ -12,4 +12,9 @@ import './chat/message';
 import './chat/message/inbound';
 import './chat/message/outbound';
 import './chat/message/system';
+import './icons/calendar-calibration';
+import './icons/calendar-maintenance';
+import './icons/calendar-reservation';
+import './icons/calendar-test-plan';
+import './icons/calendar-transport-order';
```
rajsite: After the other rename can just do:


```suggestion
import './icons/all-icons';
```

packages/spright-components/src/icons/calendar-calibration/index.ts
```diff
@@ -0,0 +1,32 @@
+import { calipers16X16 } from '@ni/nimble-tokens/dist/icons/js';
+import { IconSvg } from '@ni/nimble-components/dist/esm/icon-svg';
+import { template } from '@ni/nimble-components/dist/esm/icon-svg/template';
+import { DesignSystem } from '@ni/fast-foundation';
+import { styles } from './styles';
+
+declare global {
+    interface HTMLElementTagNameMap {
+        'spright-icon-calendar-calibration': IconCalendarCalibration;
```
jattasNI: I can't remember, did we have a discussion about naming? Based on yesterday's UI WG presentation, my understanding is that these aren't only going to be used within the calendar: they'll also be used to represent work items in other places like tables and summary panes. Is that correct? If so then I don't think we should include `calendar` in the name.

And regardless, we need to decide if [the Nimble icon naming guidelines](https://github.com/ni/nimble/blob/main/packages/nimble-tokens/CONTRIBUTING.md#icon-naming) should apply to Spright too. If they do then we would name these more aligned to their Nimble equivalents. So `nimble-icon-calipers` would map to something like `spright-icon-calipers-pink` or `spright-icon-calipers-color` if we want to avoid encoding specific colors that may change in the future.

Curious for @rajsite's opinion but I could see an argument that we should allow domain-specific names in Spright, so maybe something like `spright-icon-work-item-calibration`.

rajsite:
@mollykreis accidentally reset you but could be a good fyi for the new usage of standalone components in the spright icons that might be interesting

PR #2870, opened 2026-02-24
Update Blazor Wrapper with latest Chat Input Changes

hellovolcano:
# Pull Request

## 🤨 Rationale

Latest changes are to the Chat Conversation and Input components are needed in Blazor for use with the desktop app

## 👩‍💻 Implementation

- Exposed all public parameters
- Added event handler for stop event

## 🧪 Testing

- Added new attribute visible tests to wrapper 
- Verified new properties/events worked as expected in the Blazor example app.

## ✅ Checklist


packages/blazor-workspace/SprightBlazor/Components/SprightChatConversation.razor
```diff
@@ -1,5 +1,6 @@
 ﻿@namespace SprightBlazor
 <spright-chat-conversation
     @attributes="AdditionalAttributes">
+    @Toolbar
```
jattasNI: It doesn't look like any other components include a RenderFragment or template content for non-default slots. Even components like NimbleButton (which supports a start slot, end slot, and default slot) just use `@ChildContent` for all of them. I'll defer to Blazor owners but I think we can remove this and the change to `SprightChatConversation.razor.cs`

packages/blazor-workspace/SprightBlazor/Components/SprightChatInput.razor
```diff
@@ -2,7 +2,14 @@
 <spright-chat-input
```
jattasNI: Heads up about a couple of upcoming changes:
1. we noticed we're missing the `send-disabled` attribute as a way to disable the send button, for example when the component is in an error state (it already disables itself automatically when the input is empty)
2. we want to migrate from attributes for button labels to using a label provider

I think the SL team is going to add both of those (hopefully this week for 1, not sure about timing for 2). I'll ask them to consider the Blazor wrappers but if they don't have bandwidth / expertise we may ask you to make another quick update for each of those.

jattasNI (approved):

packages/blazor-workspace/SprightBlazor/Components/SprightChatConversation.razor
rajsite: Yep! Good catch, we should rely on the native element slot behavior so correct that we shouldn't use a render fragment here (which is Blazor's emulation of native slot behavior).

rajsite (approved):

PR #2873, opened 2026-02-25
Make test retries configurable and disable when debugging

m-akinc:
# Pull Request

## 🤨 Rationale

When debugging a test failure locally, it isn't helpful to retry the test. We really should only be running retries for official validation runs, and arguably for non-debugging, local, test runs (as we likely want the behavior to mirror what happens on build agents).

## 👩‍💻 Implementation

- `jasmine-extensions/src/index.ts` now checks a global `TEST_RETRY_COUNT` variable that is set by Vite (based on `--retries=N` argument is passed to `karma start`). If a value has been provided, it uses that. Otherwise, we fall back to the standard default of 4 retries (+ the initial run = 5 total runs)
- added `--retries=0` to the the debugging test commands (in `package.json`) for `nimble/ok/spright-components`. (The Angular test runner does not use Vite, so we can't do the same for those projects.)
- minor tweaks to retry code (e.g. moved `beforeEach` calls to _after_ the wait, in case the timing between setup and test run matters)
- tweaked the output printed when retrying tests. Now looks like this:
```
> karma start karma.conf.cjs --browsers=ChromeDebugging --retries=2 --skip-tags SkipChrome

LOG: 'FAILED: Table keyboard navigation in scrollable viewport: when column header is scrolled out of view, tabbing to the table (focusing the table without any pointer events) focuses the header and scrolls it into view
        Expected 0 to be true.
        Expected undefined to be true.

Retrying... (1/2)'
LOG: 'FAILED: Table keyboard navigation in scrollable viewport: when column header is scrolled out of view, tabbing to the table (focusing the table without any pointer events) focuses the header and scrolls it into view
        Expected 0 to be true.
        Expected undefined to be true.

Retrying... (2/2)'
Chrome 139.0.0.0 (Windows 10) Table keyboard navigation in scrollable viewport: when column header is scrolled out of view, tabbing to the table (focusing the table without any pointer events) focuses the header and scrolls it into view FAILED
        Expected 0 to be true.
            at <Jasmine>
            at UserContext.<anonymous> (src/table/tests/table-keyboard-navigation.spec.ts:1213:17)
            at run (http://localhost:9876/@fs/C:/dev/nimble/packages/jasmine-extensions/dist/esm/retry-failed-tests.js:8:23)
            at spec.queueableFn.fn (http://localhost:9876/@fs/C:/dev/nimble/packages/jasmine-extensions/dist/esm/retry-failed-tests.js:53:41)
        Expected undefined to be true.
            at <Jasmine>
            at UserContext.<anonymous> (src/table/tests/table-keyboard-navigation.spec.ts:1214:22)
            at run (http://localhost:9876/@fs/C:/dev/nimble/packages/jasmine-extensions/dist/esm/retry-failed-tests.js:8:23)
            at spec.queueableFn.fn (http://localhost:9876/@fs/C:/dev/nimble/packages/jasmine-extensions/dist/esm/retry-failed-tests.js:53:41)
TOTAL: 1 FAILED, 0 SUCCESS
```

## 🧪 Testing

Manual testing

### DIFF
--- a/packages/jasmine-extensions/src/index.ts
+++ b/packages/jasmine-extensions/src/index.ts
@@ -3,7 +3,8 @@ import { retryFailedTests } from './retry-failed-tests.js';
 declare const TEST_RETRY_COUNT: string | undefined;
 const defaultRetryCount = 4;
 
-const retriesString = TEST_RETRY_COUNT ?? '-1';
+// Need to check type (rather than just the value), in case the global doesn't exist at all
+const retriesString = typeof TEST_RETRY_COUNT !== 'undefined' ? TEST_RETRY_COUNT : '-1';
 const retries = parseInt(retriesString, 10);
 
 export function applyExtensions(): void {


packages/nimble-components/karma.conf.cjs
```diff
@@ -126,6 +126,9 @@ module.exports = config => {
                     alias: {
                         '/base': '',
                     }
+                },
+                define: {
+                    TEST_RETRY_COUNT: config.retries === undefined ? undefined : JSON.stringify(config.retries)
```
rajsite: I don't want to rely on a pattern that's vite specific / coupled to build configuration. What I've had on my personal backlog is to turn this into a karma jasmine plugin and configure it that was like any other plug-in. Would have been helpful to chat about it before spending time but I'd like to push back on this implementation direction

m-akinc: Sounds like we'll backburner this until we can do it as a proper plugin. At that point, we can consider which other changes from this branch (output formatting, refactor/wait ordering, etc) we want to pull in.

PR #2874, opened 2026-02-26
Jasmine extensions karma plugin

rajsite:
# Pull Request

## 🤨 Rationale

- Update jasmine-extensions to behave like a karma plugin (to be used from karma config) and library (to be used from a jasmine node helper).
- Updated the component library configs to only load the plugin in the headless karma configs

## 👩‍💻 Implementation

- Changed the library main import to be a karma config and moved the `applyExtensions` export to a secondary entrypoint
- As part of the karma plugin added a rollup build step for the browser js file that is loaded in the browser by the plugin

## 🧪 Testing

- Rely on existing tests
- Manually verified locally that reties happen with test commands but not test debugger commands in components libraries. Angular libraries not updated yet.

## ✅ Checklist

### DIFF
--- a/packages/jasmine-extensions/package.json
+++ b/packages/jasmine-extensions/package.json
@@ -69,7 +69,7 @@
     "karma-spec-reporter": "^0.0.36",
     "karma-vite": "^1.0.5",
     "playwright": "1.54.1",
-		"rollup": "^4.12.0",
+    "rollup": "^4.12.0",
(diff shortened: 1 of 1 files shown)


rajsite:
Bypassing owners for nights and weekends change to internal tool but can address feedback in follow-on @msmithNI @mollykreis @jattasNI fyi @m-akinc

PR #2892, opened 2026-03-09
Blazor: Reorganize files and prepend NationalInstruments namespace

### LINKED ISSUE #2865: Blazor source file layout improvements
# 🧹 Tech Debt

In most packages we have a folder per component. In the Blazor packages, the source for all components is in a flat structure within a single folder. We should consider aligning Blazor's structure to match other packages for general consistency and to make it easier to do things like changing codeowners per component. We don't think this would be a breaking change as long as the namespaces remain the same.

From a comment thread: https://github.com/ni/nimble/pull/2864#discussion_r2814302910

m-akinc:
# Pull Request

## 🤨 Rationale

For consistency with other NI-authored binaries, we want the Blazor assemblies from this repo to be namespaced under `NationalInstruments`.

Also, we want to reorganize the Blazor projects to better match the structure of the "components" packages. Fixes #2865

## 👩‍💻 Implementation

- changed namespaces and assembly names from `FooBlazor` to `NationalInstruments.FooBlazor`
- moved components `.razor` and `.razor.cs` files into component-specific subdirectories
- moved component-specific type files into said directories
- renamed `Components` folder to `Source`
- moved shared files from the root into a `Source/Patterns` directory

## 🧪 Testing

Build/tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36018,7 +36018,7 @@
     },
     "packages/angular-workspace/ok-angular": {
       "name": "@ni/ok-angular",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.2.0"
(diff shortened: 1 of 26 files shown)


change/@ni-nimble-blazor-5d43c97c-a899-4d8d-93d2-b818794e368a.json
rajsite: Spright as well:

`Chat\MessageInbound` -> `Chat\Message\Inboud`
`Chat\MessageOutbound` -> `Chat\Message\Outboud`
`Chat\MessageSystem` -> `Chat\Message\System`

PR #2907, opened 2026-03-27
Fix step icon background

rajsite:
# Pull Request

## 🤨 Rationale

Fixes two visual issues: 
- radial gradients used for the background image of the icon snaps to pixel boundaries and when you have odd scaling (like windows 150% zoom on high res displays) the background doesn't look centered and jitters as it resizes and snaps on pixel boundaries
- border sizes are on pixel boundaries so you see a snapping behavior at the animation end

Screenshot_20260327-192908

stepper-before

## 👩‍💻 Implementation

To address:
- Switched the icon background to a div with a bg color and radius that is transform sized
- Switched the icon border to a box shadow where the sizes can scale continuously

stepper-after

## 🧪 Testing

Manual tests and chromatic results 

## ✅ Checklist


jattasNI (approved):
I verified that both pinch zoom and the Zoom In menu look good on macOS Firefox and Safari

PR #2909, opened 2026-03-30
Step behavior tests

rajsite:
# Pull Request

## 🤨 Rationale

Adds test to cover behavior testing for step and anchor step.

## 👩‍💻 Implementation

- For step:
  - As step extends from FoudationButton but with a new template, follows the convention of duplicating the foundation tests. Omits a test associated with replicating a [very specific native button behavior](https://github.com/microsoft/fast/pull/5436/changes#diff-973aea4243da254b01dc55ebc076ddcaeeb1860f033a8207c233119042e061e4R590) for child spans when disabled.
  - Pulls in button tabindex tests
  - adds types test for StepSeverity
- For anchor step:
  - Pulls in anchor-button tests
  - Pulls in tabindex tests (from button)
  - adds types test for AnchorStepSeverity

Doesn't cover:
- Stepper behavior tests (already complete)
- readonly behavior (not implemented yet, coming soon)

## 🧪 Testing

See above.

## ✅ Checklist


packages/nimble-components/src/step/tests/step.foundation.spec.ts
```diff
@@ -0,0 +1,641 @@
...
+        ).toEqual(formaction);
+
+        await disconnect();
+    });
+
+    it('should set the `formenctype` attribute on the internal button when provided', async () => {
+        const { element, connect, disconnect } = await setup();
+        const formenctype = 'text/plain';
+
+        element.formenctype = formenctype;
```
jattasNI, line 95: I may be missing something, but isn't the property `formEnctype` rather than `formenctype` like the attribute? Same for `formMethod` and others. 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement/formEnctype
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/formEnctype

jattasNI (approved):

packages/nimble-components/src/step/tests/step.foundation.spec.ts
rajsite: Sure looks that way, looks like it was noticed but not clear why wasn't aligned: https://github.com/microsoft/fast/pull/2905#discussion_r406372567

The test is covering how the properties are defined in the Foundation button base class:
https://github.com/ni/fast/blob/b6d033d38166955c0544173346056475721504fe/packages/web-components/fast-foundation/src/button/button.ts#L65-L78

And agree that it's wrong. I'll create a tech debt for it to improve native form support (although technically a bug I'm not aware of anyone relying on native form support).

rajsite: created https://github.com/ni/nimble/issues/2910

Issue #2879, opened 2026-02-27
Read-only / readonly text field and number field do not have focus states


mollykreis:
# 🐛 Bug Report

The read-only state of the nimble-text-field and the nimble-number-field are focusable (which is correct), but there is a clear focus-visible state. When there is text in the input, the text becomes highlighted when the control gets focus, but when the value is empty, there is no visual indication that the control is focused.

## 🤔 Expected Behavior

When the nimble-text-field or nimble-number-field is read-only and focused, there should be a clear visual indication that the control is focused.

## 😯 Current Behavior

The only indication that a read-only nimble-text-field or a read-only nimble-number-field has focus is that the text within it gets highlighted. However, this is not enough of an indication that the field is focused, particularly when the text field is empty.

## 🔦 Context

Currently tabbing to a read-only text field makes it looks like focus becomes lost on the page.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #2913, opened 2026-04-01
Step and anchor step readonly behavior

rajsite:
# Pull Request

## 🤨 Rationale

Implement readonly state for the step and anchor step. Implements the new readonly styling discussed in #2879 of showing focus state on keyboard interactions (and not active / hover states).

## 👩‍💻 Implementation

- Implement readonly visual, i.e. the only interaction states are default and focusVisible.
  - Leveraged the [revert-layer](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/revert-layer) css feature within active and hover layers to undo them for the readonly state
- Implement readonly behavior by catching click event from the underlying button / anchor and stopping propagation & preventing default action

## 🧪 Testing

- Added to matrix tests and interactions matrix tests. Required splitting the matrix and followed patterns elsewhere of splitting on theme and manipulation states
- Added tests for verifying click event in disabled and readonly

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Enabled readonly docs in storybook.


rajsite:
@mollykreis could you review the changes for adding step readonly behavior in Jesse's absense?

packages/storybook/src/nimble/stepper/stepper.stories.ts
```diff
@@ -63,6 +66,10 @@ export const anchorStep: StoryObj<AnchorStepArgs> = {
             }),
             table: { category: apiCategory.attributes }
         },
+        readonly: {
+            description: readonlyDescription({ componentName: 'anchor step' }),
```
mollykreis, line 70: The existing readonly description doesn't seem quite right for the step/anchor-step since it doesn't have a value that can ever be changed:

>`Styles the ${options.componentName} as readonly and prevents the user from changing the value.`

mollykreis (approved):

packages/storybook/src/nimble/stepper/stepper.stories.ts
rajsite: Updated it to contrast with disabled:

```
Styles the ${options.componentName} as disabled and prevents focus and user interactions.
Styles the ${options.componentName} as readonly and allows focus but prevents other interactions.
```

think it's a bit clearer and general

PR #2758, opened 2025-11-19
Add prompts for AI coding agents

fredvisser:
## 🤨 Rationale

Provides GitHub Copilot and other AI coding agents with essential context about the Nimble design system architecture, workflows, and conventions to enable more productive assistance.

## 👩‍💻 Implementation

Added  covering:
- Multi-framework architecture (web components + Angular/Blazor/React wrappers)
- Critical workflows (build, test, Beachball change files, Chromatic visual testing)
- Component development patterns (FAST Foundation integration, design tokens, framework-specific patterns)
- Project-specific conventions (attributes over CSS classes, testing requirements, common pitfalls)

## 🧪 Testing

N/A - Documentation only

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2917, opened 2026-04-04
Add OK accordion-item

fredvisser:
# Pull Request

## 🤨 Rationale

This PR adds the OK accordion-item component.

The component was generated with Copilot from the corresponding Stratus Storybook accordion as the source reference, then adapted to fit the OK component package structure and repository conventions.

## 👩‍💻 Implementation

The implementation adds the OK accordion-item web component in the OK components package, including its template, styles, appearance type, and unit tests.

It adds the matching Angular wrapper entry point for ok-angular/accordion-item, Storybook stories and MDX docs for the OK component, and a small example-client-app integration.

Important files:
- packages/ok-components/src/accordion-item/index.ts
- packages/ok-components/src/accordion-item/template.ts
- packages/ok-components/src/accordion-item/styles.ts
- packages/angular-workspace/ok-angular/accordion-item/ok-accordion-item.directive.ts
- packages/storybook/src/ok/accordion-item/accordion-item.stories.ts

## 🧪 Testing

Automated validation run locally:
- npm run build:components -w @ni/ok-components
- npm run lint -w @ni/ok-components
- npm run test-chrome -w @ni/ok-components
- npm run build:ok -w @ni-private/angular-workspace
- npx ng lint @ni/ok-angular
- npm run test-ok -w @ni-private/angular-workspace
- npm run build:application -w @ni-private/angular-workspace
- npx ng lint example-client-app
- npm run build:ts -w @ni-private/storybook
- npm run lint -w @ni-private/storybook

## ✅ Checklist


packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
```diff
@@ -519,5 +519,14 @@
...
         </div>
+        <div class="sub-container">
+            <div class="container-label">Accordion Item (Ok)</div>
+            <ok-accordion-item i18n-header header="Asset details" [expanded]="true">
+                Calibration assets can expose operator-facing status, location, and ownership details.
+            </ok-accordion-item>
+            <ok-accordion-item i18n-header header="Maintenance history">
+                This section starts collapsed to show the default interaction state.
+            </ok-accordion-item>
+        </div>
```
mollykreis: We don't localize the example apps, so you don't need to mark any of these strings with `i18n`.


```suggestion
        <div class="sub-container">
            <div class="container-label">Accordion Item (Ok)</div>
            <ok-accordion-item header="Asset details" [expanded]="true">
                Calibration assets can expose operator-facing status, location, and ownership details.
            </ok-accordion-item>
            <ok-accordion-item header="Maintenance history">
                This section starts collapsed to show the default interaction state.
            </ok-accordion-item>
        </div>
```

packages/angular-workspace/ok-angular/accordion-item/ok-accordion-item.directive.ts
```diff
@@ -0,0 +1,48 @@
+import { booleanAttribute, Directive, ElementRef, Input, Renderer2 } from '@angular/core';
+import type { AccordionItem } from '@ni/ok-components/dist/esm/accordion-item';
+import { accordionItemTag } from '@ni/ok-components/dist/esm/accordion-item';
+import type { AccordionItemAppearance } from '@ni/ok-components/dist/esm/accordion-item/types';
+
+export type { AccordionItem };
+export { accordionItemTag };
```
mollykreis: Shouldn't this also export `AccordionItemAppearance`? Is that being exported somewhere else?

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+import { display } from '../utilities/style/display';
+import { AccordionItemAppearance } from './types';
+
+export const styles = css`
+    ${display('block')}
+
+    .accordion-item-details > .accordion-item-summary {
+        display: flex;
+        box-sizing: border-box;
+        height: 36px;
```
mollykreis: Ideally, we wouldn't hard-code a height like this. Can it be determined from other tokens? For example, is this really equal to `control-height` with a border (i.e. 2 * `divider-width`)?

packages/ok-components/src/accordion-item/template.ts
```diff
@@ -0,0 +1,31 @@
...
+const arrowExpanderRightTag = iconArrowExpanderRightTag;
+
+export const template = html<AccordionItem>`
+    <details
+        class="accordion-item-details"
+        ?open="${x => x.expanded}"
+        @toggle="${(x, c) => {
+        const details = c.event.target as HTMLDetailsElement;
+        x.expanded = details.open;
+    }}"
```
mollykreis: If you have more than a single line of an event handler, then I would recommend putting it in a function that lives in the `index.ts` rather than putting multiple lines of typescript code in the template.

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+        align-items: center;
+        margin-left: calc(-1 * ${mediumPadding});
+        border: ${dividerWidth} solid transparent;
+        list-style: none;
+        cursor: pointer;
+        user-select: none;
```
mollykreis: You should use `userSelectNone` from `utilities/style/user-select.ts` instead of setting `user-select` directly. It'll handle the styling for all browsers.

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+    .accordion-item-details[open] > .accordion-item-summary > .accordion-item-icon {
+        transform: rotate(90deg);
+    }
+
+    .accordion-item-title {
+        flex: 1;
+        width: 100%;
+        position: relative;
+        font: ${bodyPlus1EmphasizedFont};
+        line-height: 20px;
```
mollykreis: I wouldn't expect you to need to set the line-height explicitly. Is there a reason this is required?

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+        text-overflow: ellipsis;
+        overflow: hidden;
+        white-space: nowrap;
+        overflow-wrap: normal;
+    }
+
+    .accordion-item-content {
+        display: flex;
+        flex-direction: column;
+        gap: ${mediumPadding};
```
mollykreis: This `gap` styling isn't correct. We either (a) shouldn't set it because this ultimately styles content slotted by the client, or (b) should set it to a recommended value of `standardPadding` rather than `mediumPadding`.

As a point of reference, we when with option (b) in the dialog, so I could see the argument of being consistent here.

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+                margin-left: 0;
+                padding-left: 0;
+                border: none;
+            }
+
+            .accordion-item-details > .accordion-item-summary:hover {
+                border: none;
+                outline: ${dividerWidth} solid ${borderHoverColor};
+                outline-offset: -1px;
+            }
```
mollykreis: When I hover over the block accordion header, the outline flickers. I think it's because the `:hover` styling above has a transition associated with it, but this does not.

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+
+            .accordion-item-content {
+                margin-left: calc(${iconSize} + (2 * ${mediumPadding}));
+            }
```
mollykreis: I don't think we want the layout of a block accordion to be different than outline/ghost. Therefore, I wouldn't expect this margin to need to change based on the appearance.

packages/ok-components/src/accordion-item/styles.ts
```diff
@@ -0,0 +1,118 @@
...
+                border-bottom: ${dividerWidth} solid rgba(${borderRgbPartialColor}, 0.2);
+            }
+        `
+    ),
+    appearanceBehavior(
+        AccordionItemAppearance.block,
+        css`
+            :host {
+                margin-bottom: 2px;
+            }
```
mollykreis: Is there a reason not to have a 2px gap between accordion items in all appearances? That seems better than having inconsistent spacing.

In the outline appearance, that 2px gap would then be filled with the divider, but for ghost and block, it would be empty (perhaps a transparent divider if that made things easier?).

packages/ok-components/src/fv-accordion-item/template.ts
```diff
@@ -0,0 +1,31 @@
...
+    }}"
+    >
+        <summary
+            class="accordion-item-summary"
+            aria-expanded="${x => x.expanded}"
+        >
+            <${arrowExpanderRightTag}
+                class="accordion-item-icon"
+            ></${arrowExpanderRightTag}>
+            <span class="accordion-item-title" title="${x => x.header}">
```
mollykreis, line 20: Ideally, we'd use our `overflow` behavior here to conditionally set the `title` only when it didn't fully fit in the width of the control. However, always setting it is probably fine for an `ok` component.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+No borders or background. Use for nested accordions or minimal visual chrome.
+
+<Canvas of={accordionItemStories.ghost} />
+
+### Outline
+
+A bottom border divider between items. This is the default appearance.
+
+<Canvas of={accordionItemStories.outline} />
```
mollykreis: I'm curious how this was chosen as the default appearance. Just my opinion, but the outline appearance is my least favorite mode because it makes the stack of items look cluttered, and it makes it harder to visually scan the controls with extra horizontal lines next to the horizontal underlines of other nimble controls.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
```diff
@@ -0,0 +1,213 @@
...
+
+export const block: StoryObj<AccordionItemArgs> = {
+    args: {
+        header: 'Block Accordion',
+        expanded: true,
+        appearance: AccordionItemAppearance.block
+    }
+};
+
+export const outlineGroup: StoryObj<AccordionItemArgs> = {
```
mollykreis: What is the rationale of having `outlineGroup` and a `blockGroup` stories but not a `ghostGroup` story?

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
```diff
@@ -0,0 +1,213 @@
...
+        expanded: { control: false, table: { disable: true } }
+    }
+};
+
+export const nestedAccordion: StoryObj<AccordionItemArgs> = {
+    render: createUserSelectedThemeStory(html`
+        <div style="width: 400px;">
+            <${accordionItemTag} header="Versioning" appearance="${x => x.appearance}" expanded>
+                <${accordionItemTag} header="GET / API information" appearance="ghost" expanded>
+                    <p style="margin: 0;">Returns information about API versions and available operations.</p>
```
mollykreis, line 131: Can you make the text in the examples use nimble body font/color?

mollykreis (dismissed):

packages/angular-workspace/ok-angular/accordion-item/ok-accordion-item.directive.ts
fredvisser: Exported AccordionItemAppearance from the Angular wrapper so consumers can bind against the public appearance API the same way they do with other wrappers.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Replaced the hard-coded header height with token math based on controlHeight and dividerWidth.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Removed the explicit line-height. It did not appear necessary after aligning the header height with the token-based control geometry.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Added matching border and outline color transitions on the summary so the block hover state does not flicker between hover styles.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Changed the content gap to standardPadding so the spacing matches the recommended value instead of mediumPadding.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Switched from raw user-select styling to the shared userSelectNone helper.

packages/ok-components/src/accordion-item/template.ts
fredvisser: Moved the toggle logic into the component class so the FAST template stays declarative. Also made the handler return true to align with FAST event binding behavior.

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Removed the appearance-specific content indent override so block does not have a different layout than ghost and outline.

packages/ok-components/src/fv-accordion-item/template.ts
fredvisser: Left the unconditional title behavior in place. The simpler OK-component tradeoff seemed acceptable here.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.mdx
fredvisser: Changed the default appearance to ghost and updated the stories, docs, and tests to match.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
fredvisser: Updated the examples to use Nimble body font and color tokens for raw text content.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
fredvisser: None - fixed

packages/ok-components/src/accordion-item/styles.ts
fredvisser: Fixed - Added a consistent 2px divider space for all appearances by giving the host a shared bottom divider. Outline fills it, while ghost and block leave it visually empty.

packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
fredvisser: Turns out linting requires it, so added it back:

>@ni-private/angular-workspace:lint: /home/runner/work/nimble/nimble/packages/angular-workspace/example-client-app/src/app/customapp/customapp.component.html
@ni-private/angular-workspace:lint:   524:36  error  Attribute "header" has no corresponding i18n attribute. See more at https://angular.dev/guide/i18n-common-prepare#mark-element-attributes-for-translations  @angular-eslint/template/i18n
@ni-private/angular-workspace:lint:   527:36  error  Attribute "header" has no corresponding i18n attribute. See more at https://angular.dev/guide/i18n-common-prepare#mark-element-attributes-for-translations  @angular-eslint/template/i18n

mollykreis: Any idea why it was required here but doesn't seem to have been required elsewhere within this file?

packages/ok-components/src/accordion-item/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    interface HTMLElementTagNameMap {
+        'ok-accordion-item': AccordionItem;
+    }
+}
+
+/**
+ * An accordion item component that can be expanded or collapsed to
+ * show or hide its content.
+ */
+export class AccordionItem extends FoundationElement {
```
rajsite: Chatted with @fredvisser and to avoid potential clashes between libraries we should add a name prefix to these like we ended up doing with the spright icons., i.e. to avoid conflicts with `nimble-icon-calendar-week` we added a  `work-item` name prefix to be `spright-icon-work-item-calendar-week`

Following that here, thinking a name prefix of `fv` so:

```
/ok-components/src/fv-accordion-item/
/storybook/src/ok/fv-accordion-item/
    fv-accordion-item-matrix.stories.ts
    fv-accordion-item.mdx
    fv-accordion-item.stories.ts
/angular-workspace/ok-angular/fv-accordion-item/
    fv-accordion-item.directive.ts
    fv-accordion-item.module.ts
...
```

And then for element names / class names:
```
ok-fv-accordion-item
FvAccordionItem
...
```

Should also make it easier to define CODEOWNERS for the group of components:
```
/**/fv-*/**
```
fyi @jattasNI for thoughts on the naming scheme / direction?

change/@ni-ok-angular-09482ac4-a62c-47fc-be43-c52b823e870e.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Things that are in new component dev in nimble:
- Having the tests create and use page objects
  - Goal is to have those available from their `/testing` entrypoint for use in clients. The internal details and timing is encapsulated in the page object for tests
  - See folder structure: https://github.com/ni/nimble/blob/main/packages/nimble-components/CONTRIBUTING.md#folder-structure
  - See common issue notes: https://github.com/ni/nimble/issues/1479#issuecomment-1703311244
  - See LLM generated pageobject for stepper: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/stepper/testing/stepper.pageobject.ts
  - See usage of pageobject in LLM generated tests: https://github.com/ni/nimble/blob/main/packages/nimble-components/src/stepper/tests/stepper.spec.ts#L206
- Have styles in control all organized into the CSS layers stack:
  - See docs: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#organize-styles-with-css-cascade-layers
  - See step example:  https://github.com/ni/nimble/blob/629480505b98404f37f1c3542c2acdfb8c4a7af6/packages/nimble-components/src/patterns/step/styles.ts#L33

fredvisser:
Hey @NIbokeefe - If you have feedback on this component, feel free to leave it here, or push changes - thanks!

FYI - I believe I've fixed the layout since you saw it last

Screenshot 2026-04-10 at 5 52 20 PM

change/@ni-ok-angular-09482ac4-a62c-47fc-be43-c52b823e870e.json
rajsite: - Also consider wrappers for all the supported frameworks, i.e. angular, blazor, react and corresponding example app updates

.github/CODEOWNERS
```diff
@@ -18,6 +18,9 @@
 /packages/nimble-components/src/rich-text-mention @rajsite @jattasNI @vivinkrishna-ni
 /**/chat/** @jattasNI
 /**/breakpoint/** @jattasNI
+/packages/ok-components/src/fv-* @rajsite @jattasNI @fredvisser
+/packages/angular-workspace/ok-angular/fv-* @rajsite @mollykreis @fredvisser
+/packages/storybook/src/ok/fv-* @rajsite @jattasNI @fredvisser
```
rajsite: similar to chat above this can just be:

```suggestion
/**/fv-*/** @jattasNI @fredvisser
```
(assuming jesse has the bandwidth for basic infrastructure tracking in the repo for the changes)

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+};
+
+export const block: StoryObj<AccordionItemArgs> = {
+    args: {
+        header: 'Block Accordion',
+        expanded: true,
+        appearance: FvAccordionItemAppearance.block
+    }
+};
```
rajsite: We have removed the trivial appearance stories from other components. Consider removing the trivial appearance stories and combining the nested stories into one with a configurable parameter:

Image

```
- FV Accordion Item
  - Docs
  - FV Accordion Item (primary canvas story with most options exposed)
  - Grouped Items
  - Nested Items
  - Mixed Content
```

change/@ni-ok-angular-09482ac4-a62c-47fc-be43-c52b823e870e.json
fredvisser: I've added this and other review feedback to a `component-review` skill that I'm maintaining locally - any interest in including this in the repo?

rajsite: Not interested in growing the scope of this PR. There are open questions about the scope and maintenance of such files that were never addressed in #2758 and those discussions should happen outside this PR.

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
```diff
@@ -0,0 +1,188 @@
+import type { Meta, StoryObj } from '@storybook/html-vite';
```
jattasNI, line 1: I think we should add a disclaimer to every "ok" story that's similar to our incubating disclaimer. I know we document the standards of "ok" in various places but it's more likely to be seen if it's on the individual component docs. Borrowing from the [status table](https://nimble.ni.dev/storybook/index.html?path=/docs/component-status--docs#ok-components), I'd propose something like:

"Warning: the accordion is an ok component. These components have varying levels of quality and stability. Clients should consult with component-specific owners before leveraging in new applications."

packages/storybook/src/ok/fv-accordion-item/fv-accordion-item.stories.ts
```diff
@@ -0,0 +1,188 @@
...
+import { html } from '@ni/fast-element';
```
jattasNI, line 2: The new components should probably be added to the component status table.

jattasNI (approved):

packages/ok-components/src/accordion-item/index.ts
jattasNI: I'm ok with that direction.

PR #2920, opened 2026-04-07
Add welcome message to chat

Alexia-Claudia-Micu:
# Pull Request

https://dev.azure.com/ni/DevCentral/_workitems/edit/3786413
Continuation of https://github.com/ni/nimble/pull/2891

## 🤨 Rationale

Adding welcome message for Spright Chat Component, as defined in [Chat components]https://github.com/ni/nimble/blob/main/packages/spright-components/src/chat/specs/README.md)

## 👩‍💻 Implementation

- Added new spright-chat-message-welcome component and framework wrapper.
- Added new story for the welcome message.

## 🧪 Testing

Manual testing and added tests

## ✅ Checklist


change/@ni-spright-angular-3dd80b72-fa3a-4f62-9749-f852ef6a2b91.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add welcome message for spright chat conversation",
+  "packageName": "@ni/spright-angular",
+  "email": "5265744+hellovolcano@users.noreply.github.com",
```
OliverCosma: reminder to change this with your configured email - in all the change files

packages/storybook/src/spright/chat/conversation/chat-conversation.mdx
```diff
@@ -69,6 +70,16 @@ Use the <Tag name={chatMessageSystemTag} /> element to display content in a syst
...
 
+#### Welcome message
+
+Use the <Tag name={chatMessageWelcomeTag} /> element to display welcome content in the chat window.
+
+##### Welcome message content
+
+<Canvas of={messageStories.chatMessageWelcome} />
+<Controls of={messageStories.chatMessageWelcome} />
+
```
OliverCosma:
```suggestion
```

packages/angular-workspace/spright-angular/chat/message/welcome/tests/spright-chat-message-welcome.directive.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+                declarations: [TestHostComponent],
+                imports: [SprightChatMessageWelcomeModule]
+            });
+            fixture = TestBed.createComponent(TestHostComponent);
+            fixture.detectChanges();
+            directive = fixture.componentInstance.directive;
+            nativeElement = fixture.componentInstance.elementRef.nativeElement;
+        });
+
+        it('can be configured with property binding for welcomeTitle and subtitle', () => {
```
OliverCosma, line 126: Looking at `spright-chat-input.directive.spec` I see that each property gets its own separate test for property binding, can we split this into two separate tests? 

If one property breaks, you'd know exactly which one

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    );
+}
+
+describe('ChatMessageWelcome', () => {
+    let element: ChatMessageWelcome;
+    let connect: () => Promise<void>;
+    let disconnect: () => Promise<void>;
+
+    afterEach(async () => {
+        if (typeof disconnect === 'function') {
```
OliverCosma: I don't think I understand this check, can you verify if it is necessary?

packages/spright-components/src/chat/message/welcome/styles.ts
```diff
@@ -0,0 +1,65 @@
...
+    .message-content {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        width: fit-content;
+        height: fit-content;
+        overflow-x: auto;
+        gap: ${mediumPadding};
+    }
+
```
OliverCosma:
```suggestion
```

packages/spright-components/src/chat/message/welcome/template.ts
```diff
@@ -0,0 +1,27 @@
+import { html, ViewTemplate, when } from '@ni/fast-element';
+import type {
+    FoundationElementTemplate
+} from '@ni/fast-foundation';
+import type { ChatMessageWelcome } from '.';
+import { iconNigelChatTag } from '../../../icons/nigel-chat';
+
+export const template: FoundationElementTemplate<
+ViewTemplate<ChatMessageWelcome>
+> = (_context, _definition) => html<ChatMessageWelcome>`
```
OliverCosma: if context and definition are not used, maybe we could have the same pattern as in the other message components with `> = () => html<ChatMessageWelcome>`

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -83,6 +85,12 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
                     This is the message text of this banner. It tells you something interesting.
                 </${bannerTag}>
             `)}
+            ${when(x => x.welcome, html<ChatConversationArgs>`
```
OliverCosma: Looks like in this way we have the welcome messages along with the other messages 

Image

My suggestion would be to have a separate story with the welcome message, given that in practice we will have login and then display the rest of the chat conversation

packages/storybook/src/spright/chat/message/chat-message-matrix.stories.ts
```diff
@@ -68,3 +69,15 @@ export const messageSystemTextCustomized: StoryFn = createMatrixThemeStory(
         html`<${chatMessageSystemTag}>System Message</${chatMessageSystemTag}>`
     )
 );
+
+export const messageWelcomeHidden: StoryFn = createStory(
+    hiddenWrapper(
+        html`<${chatMessageWelcomeTag} hidden title="Welcome" subtitle="Get started">Hidden Chat Welcome Message</${chatMessageWelcomeTag}>`
```
OliverCosma:
```suggestion
        html`<${chatMessageWelcomeTag} hidden welcome-title="Welcome" subtitle="Get started">Hidden Chat Welcome Message</${chatMessageWelcomeTag}>`
```

Image

packages/storybook/src/spright/chat/message/chat-message-matrix.stories.ts
```diff
@@ -68,3 +69,15 @@ export const messageSystemTextCustomized: StoryFn = createMatrixThemeStory(
...
+    )
+);
+
+export const messageWelcomeTextCustomized: StoryFn = createMatrixThemeStory(
+    textCustomizationWrapper(
+        html`<${chatMessageWelcomeTag} title="Welcome" subtitle="Get started">Welcome Message</${chatMessageWelcomeTag}>`
```
OliverCosma:
```suggestion
        html`<${chatMessageWelcomeTag} welcome-title="Welcome" subtitle="Get started">Welcome Message</${chatMessageWelcomeTag}>`
```

OliverCosma (approved):

change/@ni-spright-angular-3dd80b72-fa3a-4f62-9749-f852ef6a2b91.json
Alexia-Claudia-Micu: Changed them.

packages/storybook/src/spright/chat/conversation/chat-conversation.mdx
Alexia-Claudia-Micu: Removed line.

packages/spright-components/src/chat/message/welcome/styles.ts
Alexia-Claudia-Micu: Removed line.

packages/spright-components/src/chat/message/welcome/template.ts
Alexia-Claudia-Micu: Changed it.

packages/storybook/src/spright/chat/message/chat-message-matrix.stories.ts
Alexia-Claudia-Micu: Changed it.

packages/angular-workspace/spright-angular/chat/message/welcome/tests/spright-chat-message-welcome.directive.spec.ts
Alexia-Claudia-Micu: Split them into `should display the subtitle when both title and subtitle are set` and `should display the welcome title when both title and subtitle are set`

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
Alexia-Claudia-Micu: Removed it.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
OliverCosma: @jattasNI  - there is just one story with the entire "chat conversation". The system/inbound/other types of messages appear in the subsection `chat messages`. However, to me it feels weird to have the welcome messages in the chat conversation story - would it make sense to have a separate story just with the chat conversation - welcome message - disclaimer - and the login button ?

packages/storybook/src/spright/chat/message/chat-message-matrix.stories.ts
Alexia-Claudia-Micu: Added suggestion

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
jattasNI: I like having the welcome message in a separate story: it doesn't feel right to put it in a conversation next to the other messages. But I think it does make sense to include it in the conversation MDX so that it is documented next to other messages and chat components.

If we stick with that approach then I don't think we need the new `welcomeLogin` story in `chat-conversation.stories.ts`. The `messageStories.chatMessageWelcome` that's being embedded in the MDX and [included in the Chromatic diff](https://www.chromatic.com/test?appId=60e89457a987cf003efc0a5b&id=69d7a1df2c9a27b97046971e) should be sufficient.

packages/storybook/src/spright/chat/message/chat-message.stories.ts
```diff
@@ -156,3 +159,70 @@ export const chatMessageImage: StoryObj<ChatMessageInboundArgs> = {
...
+        loginButton: {
+            name: 'Show login button',
+            description: 'Slot a login anchor button below the title and subtitle.',
+            table: { category: apiCategory.slots }
+        },
+        suggestions: {
+            name: 'Show suggestions',
+            description: 'Slot suggested outbound messages buttons below the title and subtitle.',
+            table: { category: apiCategory.slots }
+        }
```
jattasNI: We try to keep a 1-1 mapping between parameter names and actual parameters. In this case I'd like to show a single slot parameter with `name: 'default'` rather than multiple slots with names "Show login button" and "Show suggestions".

The way we've achieved that in other places is to change the type of the slot to be an enum that lets you pick between different types of higher-level content. In this case you could show all or some of: 
- "None" - equivalent to both login button and suggestions being false
- "Login button"
- "Suggestions"
- "Login button and suggestions"

You could skip any combinations that don't make sense (e.g. maybe we shouldn't include the last one because you shouldn't be able to receive suggestions if you're not logged in).

You can see an example of this in the breadcrumb story's docs for its default slot. https://nimble.ni.dev/storybook/?path=/docs/components-breadcrumb--docs

packages/spright-components/src/chat/message/welcome/styles.ts
```diff
@@ -0,0 +1,64 @@
...
+    .container {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        max-width: calc(90%);
+        gap: ${mediumPadding};
+    }
+
+    .brand-icon {
```
jattasNI, line 37: From screenshots of desktop and/or wireframes I thought I recalled this icon being big and multicolored rather than small and only black. I can confirm what's right at the UX meeting this Friday, but to start I'd suggest matching what desktop is doing:
- [images](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/MeasurementServices/AiAssistants/NigelLocalService/wwwroot/Images/two-chat-sparkle_green_LightUI_48x48.svg) (there's also a dark version)
- [image sizing](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/MeasurementServices/AiAssistants/Controls/Components/EmptyConversation.razor.css&version=GBmain&line=11&lineEnd=12&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

**From this PR** 

Image

**From Figma**

Image

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        expect(titleDiv?.textContent?.trim()).toBe('Welcome to Nigel AI');
+    });
+
+    it('should display the subtitle when both title and subtitle are set', async () => {
+        ({ element, connect, disconnect } = await setup(
+            'Welcome to Nigel AI',
+            'Chat below to get started'
+        ));
+        await connect();
+        const subtitleDiv = element.shadowRoot?.querySelector('.subtitle');
```
jattasNI: Could you put all the DOM manipulation code into a page object? Should include querySelector, textContent, trim, etc.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    it('should not render subtitle element when subtitle is not set', async () => {
+        ({ element, connect, disconnect } = await setup());
+        await connect();
+        const subtitleDiv = element.shadowRoot?.querySelector('.subtitle');
+        expect(subtitleDiv).toBeNull();
+    });
+
+    it('should have a brand-icon slot', async () => {
```
jattasNI: I think we're missing a test about the default brand icon.

packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
Alexia-Claudia-Micu: Removed it from where it was and added a new story on the same level as Docs, Chat Conversation: `Chat Message Welcome`

packages/storybook/src/spright/chat/message/chat-message.stories.ts
Alexia-Claudia-Micu: Added the `default` parameter with options: "None", "Login Button", "Suggestions"

packages/spright-components/src/chat/message/welcome/styles.ts
Alexia-Claudia-Micu: Added light and dark mode icons to `icon-data.ts‎` for now.

@jattasNI Do you think it would be better if I added them to nimble-tokens instead?

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
Alexia-Claudia-Micu: Added `chat-message-welcome.pageobject.ts` and moved the DOM manipulation code there.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
Alexia-Claudia-Micu: Added test `should display the default Nigel chat icon when no brand-icon slot content is provided`

packages/storybook/src/spright/chat/conversation/chat-conversation.mdx
```diff
@@ -69,7 +69,7 @@ Use the <Tag name={chatMessageSystemTag} /> element to display content in a syst
 <Canvas of={messageStories.chatMessageSpinner} />
 <Controls of={messageStories.chatMessageSpinner} />
 
-### Chat input
+chat-message.stories.ts### Chat input
```
jattasNI: Something got messed up with this change. We should add the `chatMessageWelcome` canvas and controls here.

packages/spright-components/src/icons/nigel-chat/styles.ts
```diff
@@ -0,0 +1,27 @@
+import { css } from '@ni/fast-element';
+import { styles as iconSvgStyles } from '@ni/nimble-components/dist/esm/icon-svg/styles';
+import { Theme } from '@ni/nimble-components/dist/esm/theme-provider/types';
+import { themeBehavior } from '@ni/nimble-components/dist/esm/utilities/style/theme';
+
+export const styles = css`
+    ${iconSvgStyles}
+
+    :host {
```
jattasNI: I think we should make some changes to icon sizing.

Right now the new `nigel-chat` icon has an intrinsic size of 48px while other icons are 16px. This results in [it not fitting in the icon table in storybook](https://60e89457a987cf003efc0a5b-jjrzhrkrcd.chromatic.com/?path=/docs/spright-icons--docs&args=brandIcon:!true;defaultSlot:none). On the plus side, clients can still set its size to something else.

Also, right now the size of the icon within the message is just whatever the clients sets. It's good to offer that flexibility but the result is that in storybook the new brand icon is 48px but the custom slotted icon is only 16px, which looks small.

I'd propose the following instead:
1. set the default size of the new `nigel-chat` icon to 16px to match other icons (I think `iconSvgStyles` already does this via the `iconSize` token, so you'd just delete this `:host` style)
2. change the welcome message to style the icon slot contents to 48px by default. This will apply to both the default icon and client-provided icons.
3. allow clients to override that styling

I think you can achieve 2 and 3 by using the `slot[name='brand-icon']::slotted(*)` selector within the welcome message styles.

packages/spright-components/src/chat/message/welcome/styles.ts
jattasNI: The way you added the icon to spright looks correct to me! (With some minor sizing changes suggested in another thread)

jattasNI (commented):
Partial review of the spright-components changes, will return to the other changes later today.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+        it('should have a brand-icon slot', () => {
+            expect(pageObject.getBrandIconSlot()).not.toBeNull();
+        });
+
+        it('should display the default Nigel chat icon when no brand-icon slot content is provided', () => {
+            expect(pageObject.getDefaultBrandIcon()).not.toBeNull();
+        });
+
+        it('should display slotted content in the default slot', () => {
+            expect(pageObject.getInnerText().includes('Welcome content')).toBeTrue();
```
jattasNI: Instead of exposing a generic method `getInnerText()` we should be more specific and add a page object method that queries for the contents of the slot inside the `.message-content` section.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+        ? `welcome-title="${welcomeTitle}"`
+        : '';
+    const subtitleAttr = subtitle ? `subtitle="${subtitle}"` : '';
+    return await fixture<ChatMessageWelcome>(
+        // prettier-ignore
+        html`<${chatMessageWelcomeTag} ${titleAttr} ${subtitleAttr}>Welcome content</${chatMessageWelcomeTag}>`
+    );
+}
+
+describe('ChatMessageWelcome', () => {
```
jattasNI, line 30: We're missing test coverage for slotting non-default `brand-icon` content.

packages/spright-components/src/chat/message/welcome/testing/chat-message-welcome.pageobject.ts
```diff
@@ -0,0 +1,50 @@
...
+        return this.element.shadowRoot?.querySelector('slot:not([name])');
+    }
+
+    public getBrandIconSlot(): Element | null | undefined {
+        return this.element.shadowRoot?.querySelector('slot[name="brand-icon"]');
+    }
+
+    public getDefaultBrandIcon(): Element | null | undefined {
+        return this.element.shadowRoot?.querySelector(iconNigelChatTag);
+    }
```
jattasNI: We should try to avoid exposing methods that return elements in the shadow DOM since that breaks encapsulation. Instead you could either (or both):
1. expose boolean methods that return whether the slot exists
2. expose methods that return the client-provided contents of the slot via [`assignedElements()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement#instance_methods)

If both are needed for tests then you could have a private helper that gets the slot element.

Personally I don't think it's too useful to explicitly assert that the slot exists. If we have a test that asserts that the slot contents are correct then that implies the slot exists.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
Alexia-Claudia-Micu: Added `should display the slotted brand-icon content`

packages/spright-components/src/chat/message/welcome/styles.ts
Alexia-Claudia-Micu: Made sizing changes.

packages/storybook/src/spright/chat/conversation/chat-conversation.mdx
Alexia-Claudia-Micu: Fixed it and added `chatMessageWelcome`

packages/spright-components/src/icons/nigel-chat/styles.ts
Alexia-Claudia-Micu: Implemented suggestions, used `slot[name='brand-icon']::slotted(*), slot[name='brand-icon'] > * ` for the welcome message style.

packages/spright-components/src/chat/message/welcome/testing/chat-message-welcome.pageobject.ts
Alexia-Claudia-Micu: Made methods private and removed the test that asserted that the slot exists.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
Alexia-Claudia-Micu: Replaced with page object method.

packages/spright-components/src/chat/message/welcome/tests/chat-message-welcome.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+            expect(pageObject.isTitleRendered()).toBeFalse();
+        });
+
+        it('should not render subtitle element when subtitle is not set', () => {
+            expect(pageObject.isSubtitleRendered()).toBeFalse();
+        });
+
+        it('should display the default Nigel chat icon when no brand-icon slot content is provided', () => {
+            expect(pageObject.getSlottedBrandIconNodes().length).toBe(0);
+            expect(pageObject.hasBrandIconDefaultContent()).toBeTrue();
```
jattasNI, line 63: FYI I pushed a small change to the test to add an assertion about the brand icon being present when there's no client-provided content.

jattasNI (approved):

jattasNI:
Bypassing framework owners @rajsite @mollykreis @msmithNI as changes only affect chat components

PR #2695, opened 2025-09-05
Scrollable breadcrumb

rajsite:
# Pull Request

## 🤨 Rationale

This PR resolves https://github.com/ni/nimble/issues/2186 by implementing overflow behavior similar to the tab control.
Notably this PR does not implement the design discussed in https://github.com/ni/nimble/issues/2213 and is instead a shorter term workaround.

## 👩‍💻 Implementation

- Fork the foundation breadcrumb template and tests
- Duplicate the tab scroll implementation and tests tweaked for the breadcrumb
  - Notable behavior difference is that for breadcrumb each breadcrumb-item is a tab stop instead of having a single tab stop like tabs and using roving tabs between each tab.
- Corresponding SystemLink changes can be seen in [AZDO 1063821](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/1063821)
- Opting to duplicate scrollable tabs / breadcrumb-items pattern for now, can consider de-duplicating it if another usage is found

## 🧪 Testing

- Duplicates test patterns of scrollable tabs
- Manual validation in SystemLinkShared sl-header storybook and TestInsights app
  - Example [stratus-dev sl-header storybook](https://62431f379ac921003aaca05a-vmhuqvbrqg.chromatic.com/?path=/story/components-header--default&args=itemsType:many) build

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2921, opened 2026-04-08
Scrollable stepper

rajsite:
# Pull Request

## 🤨 Rationale

Implements scrollable stepper pattern similar to breadcrumbs and tabs.

## 👩‍💻 Implementation

Follows the pattern established in #2695 reasonably closely with the following differences:
- Implements both horizontal and vertical scrollable behavior based on orientation
- Uses an interaction observer instead of a resize observer (comments in source go into more detail)

## 🧪 Testing

- Duplicates the same test patterns and matrix tests
- Similar pattern for storybook tests

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/stepper/index.ts
+++ b/packages/nimble-components/src/stepper/index.ts
@@ -18,9 +18,7 @@ export class Stepper extends FoundationElement {
     @attr
     public orientation: StepperOrientation = StepperOrientation.horizontal;
 
-    /**
-     * @internal
-     */
+    /** @internal */
     @observable
     public showScrollButtons = false;
 
@@ -28,23 +26,17 @@ export class Stepper extends FoundationElement {
     @observable
     public steps?: (StepPattern)[];
 
-    /**
-     * @internal
-     */
+    /** @internal */
     public list!: HTMLElement;
 
-    /**
-     * @internal
-     */
+    /** @internal */
     public readonly startScrollButton?: HTMLElement;
 
     private listIntersectionObserver?: IntersectionObserver;
 
-    /**
-     * @internal
-     */
+    /** @internal */
     public onScrollStartClick(): void {
-        if (this.orientation === StepperOrientation.horizontal) {
+        if (this.isHorizontal()) {
             this.list.scrollBy({
                 left: -this.list.clientWidth,
                 behavior: 'smooth'
@@ -57,11 +49,9 @@ export class Stepper extends FoundationElement {
         }
     }
 
-    /**
-     * @internal
-     */
+    /** @internal */
     public onScrollEndClick(): void {
-        if (this.orientation === StepperOrientation.horizontal) {
+        if (this.isHorizontal()) {
             this.list.scrollBy({
                 left: this.list.clientWidth,
                 behavior: 'smooth'
@@ -74,6 +64,7 @@ export class Stepper extends FoundationElement {
         }
     }
 
+    /** @internal */
     public override connectedCallback(): void {
         super.connectedCallback();
         // Steps fill parent container so can't rely on a resize observer to track their space usage.
@@ -87,9 +78,15 @@ export class Stepper extends FoundationElement {
         });
     }
 
+    /** @internal */
     public override disconnectedCallback(): void {
         super.disconnectedCallback();
         this.listIntersectionObserver?.disconnect();
+    }
+
+    /** @internal */
+    public isHorizontal(): boolean {
+        return this.orientation !== StepperOrientation.vertical;
     }
 
     private orientationChanged(): void {
@@ -110,26 +107,27 @@ export class Stepper extends FoundationElement {
         }
         const lastIndex = this.steps.length - 1;
         this.steps.forEach((step, index) => {
-            step.stepInternals.orientation = this.orientation;
+            step.stepInternals.orientation = this.isHorizontal()
+                ? StepperOrientation.horizontal
+                : StepperOrientation.vertical;
             step.stepInternals.last = index === lastIndex;
             step.stepInternals.position = index + 1;
         });
     }
 
     private handleListOverflow(): void {
-        const isHorizontal = this.orientation === StepperOrientation.horizontal;
-        let listVisibleSize = isHorizontal
+        let listVisibleSize = this.isHorizontal()
             ? this.list.clientWidth
             : this.list.clientHeight;
         if (listVisibleSize !== undefined) {
-            const buttonSize = isHorizontal
+            const buttonSize = this.isHorizontal()
                 ? this.startScrollButton?.clientWidth ?? 0
                 : this.startScrollButton?.clientHeight ?? 0;
             listVisibleSize = Math.ceil(listVisibleSize);
             if (this.showScrollButtons) {
                 listVisibleSize += buttonSize * 2;
             }
-            const listScrollSize = isHorizontal
+            const listScrollSize = this.isHorizontal()
                 ? this.list.scrollWidth
                 : this.list.scrollHeight;
             this.showScrollButtons = listVisibleSize < listScrollSize;
--- a/packages/nimble-components/src/stepper/index.ts
+++ b/packages/nimble-components/src/stepper/index.ts
@@ -39,14 +39,7 @@ export class Stepper extends FoundationElement {
      */
     public readonly startScrollButton?: HTMLElement;
 
-    private readonly listResizeObserver: ResizeObserver;
-
-    public constructor() {
-        super();
-        this.listResizeObserver = new ResizeObserver(_ => {
-            this.handleListOverflow();
-        });
-    }
+    private listIntersectionObserver?: IntersectionObserver;
 
     /**
      * @internal
@@ -84,12 +77,25 @@ export class Stepper extends FoundationElement {
 
     public override connectedCallback(): void {
         super.connectedCallback();
-        this.listResizeObserver.observe(this.list);
+        // Unlike other scrollable implementations, the steps fill the space
+        // of the container instead of the container sized to the items.
+        // We can't rely on the resizing of the container so instead
+        // track intersection of all steps inside the container.
+        // Found in manual testing that can get in states of partial occlusion
+        // before tabbing / using arrow keys triggers an intersection change.
+        // Hopefully the reliable solution is container scroll state queries when available:
+        // https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries
+        this.listIntersectionObserver = new IntersectionObserver(_ => {
+            this.handleListOverflow();
+        }, {
+            root: this.list,
+            threshold: 1.0
+        });
     }
 
     public override disconnectedCallback(): void {
         super.disconnectedCallback();
-        this.listResizeObserver.disconnect();
+        this.listIntersectionObserver?.disconnect();
     }
 
     private orientationChanged(): void {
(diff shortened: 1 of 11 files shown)


rajsite:
@mollykreis could you take a look while jesse is OOO?

packages/nimble-components/src/stepper/template.ts
```diff
@@ -1,7 +1,44 @@
...
+import { iconArrowExpanderUpTag } from '../icons/arrow-expander-up';
+import { iconArrowExpanderDownTag } from '../icons/arrow-expander-down';
 
-export const template = html<Stepper>`<ol><slot
-        name="step"
-        ${slotted('steps')}
-    ></slot></ol>`;
+export const template = html<Stepper>`
+    ${when(x => x.showScrollButtons, html<Stepper>`
+        <${buttonTag}
```
jattasNI, line 16: On the tabs pattern we also set `tabindex="-1"`.
https://github.com/ni/nimble/blob/d5f4ee7abdc7c280d27a9e822fbb3a659fcaedee/packages/nimble-components/src/patterns/tabs/template.ts#L33

On the breadcrumb we don't do that.
https://github.com/ni/nimble/blob/d5f4ee7abdc7c280d27a9e822fbb3a659fcaedee/packages/nimble-components/src/breadcrumb/template.ts#L21-L27

The end result is that the scroll buttons are keyboard focusable for the breadcrumb but not the tab. I can't recall for certain whether that was an intentional choice. If it was, then maybe we justified it because the tabs allow using arrow keys to navigate focus and don't allow tab focus on individual tabs? If that's the justification then I guess it makes sense for stepper to align with breadcrumb, as it's doing right now. Though in playing with it, I didn't get a lot of value from tabbing to the > button after I'd already tabbed through all the steps, which had the side effect of doing the scrolling for me already.

Not a super consequential decision so don't spend too much time thinking about it. But if you have a preference and rationale for it, it'd be good to write down in this thread for the next time we face this choice.

packages/storybook/src/nimble/stepper/stepper.stories.ts
```diff
@@ -226,9 +226,31 @@ const severity: readonly StepSetItem[] = [
     { title: 'Share results', subtitle: 'Reveal who is responsible', selected: true },
 ];
```
jattasNI, line 228: I noticed some Chromatic diffs in step stories in Firefox but couldn't see a reason why. Any explanation for that or just Chromatic being noisy?

packages/nimble-components/src/stepper/styles.ts
```diff
@@ -1,21 +1,67 @@
...
     }
 
     :host([orientation="vertical"]) {
         flex-direction: column;
     }
 
-    ol {
-        display: contents;
+    .scroll-button.start {
+        width: ${controlHeight};
```
jattasNI, line 17: What's the rationale for explicitly sizing the buttons? Are we (perhaps in an undocumented way) supporting overriding controlHeight to get a "slim" mode? If so, worth adding a Chromatic state for it?

packages/nimble-components/src/stepper/index.ts
```diff
@@ -18,16 +18,96 @@ export class Stepper extends FoundationElement {
...
+     */
+    @observable
+    public showScrollButtons = false;
+
     /** @internal */
     @observable
     public steps?: (StepPattern)[];
 
+    /**
+     * A reference to the list element
```
jattasNI: super nit: seems like an obvious comment (obvious enough that we don't need it for the other element refs)

```suggestion
```

packages/nimble-components/src/stepper/index.ts
```diff
@@ -18,16 +18,96 @@ export class Stepper extends FoundationElement {
...
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        // Unlike other scrollable implementations, the steps fill the space
+        // of the container instead of the container sized to the items.
+        // We can't rely on the resizing of the container so instead
+        // track intersection of all steps inside the container.
+        // Found in manual testing that can get in states of partial occlusion
+        // before tabbing / using arrow keys triggers an intersection change.
+        // Hopefully the reliable solution is container scroll state queries when available:
+        // https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries
```
jattasNI: Worth filing a tech debt backlog item to track adopting this approach when it's available in baseline, similar to #2579? I recognize we may never get to it, but it's nice to imagine we will.

packages/nimble-components/src/stepper/tests/stepper.spec.ts
```diff
@@ -201,6 +203,112 @@ describe('Stepper', () => {
...
+
+                        afterEach(async () => {
+                            await disconnect();
+                        });
+
+                        it('should not show scroll buttons when the steps fit within the container', () => {
+                            expect(stepperPageObject.areScrollButtonsVisible()).toBeFalse();
+                        });
+
+                        it('should show scroll buttons when the steps overflow the container', async () => {
```
jattasNI, line 247: You could consider adding a case where the stepper is in horizontal orientation with an explicit width set, and you change the length of a step title, causing the scroll buttons to appear. Though I guess we already have coverage for the individual components of that: Chromatic coverage of the stepper growing to fit its steps and of setting an explicit width, unit test coverage that buttons are shown when steps exceed width. Your call.

jattasNI (approved):
I'm OOO but partially working remote this week, so found time to review this. @mollykreis is welcome to review it if you want or call it good with just my review.

packages/nimble-components/src/stepper/template.ts
rajsite: Yep, you got it and was intentionally inline with breadcrumb and as noted on the breadcrumb PR #2695 

```
- Duplicate the tab scroll implementation and tests tweaked for the breadcrumb
  - Notable behavior difference is that for breadcrumb each breadcrumb-item is a tab stop instead of having a single tab stop like tabs and using roving tabs between each tab.
```

Which wanted a review of the dev UX but haven't gotten yet: https://github.com/ni/nimble/pull/2695#discussion_r2370372690

For both breadcrumbs and here it's a dev design without UX alignment so think we can keep them aligned until find otherwise.

packages/storybook/src/nimble/stepper/stepper.stories.ts
rajsite: I couldn't find a reason why just ffx had a small change to the line length and only for the last line. Guessing maybe some aliasing around the icon messing with aliasing around the line but dunno. If it continues to be unstable would need to look into but leaving it for now.

packages/nimble-components/src/stepper/styles.ts
rajsite: No slim mode on the step, has too much funky layout to be that easy unfortunately. This is to align with with step icon size so the tabbing between them is continuous. Can add a comment saying that's the intention of the sizing.

> Are we (perhaps in an undocumented way) supporting overriding controlHeight to get a "slim" mode?

Definitely not, I really don't want to go down a path of overriding custom properties we invent for sizing. Size custom properties are just for defaults (and in the future their values could map to a theme property like density). Users should use normal CSS sizing / layout props on the control or on parts exposed by the control but leverage our css prop values. We can expose neat things like custom values as css props to set on standard CSS props like we did on the table.

packages/nimble-components/src/stepper/tests/stepper.spec.ts
rajsite: I'm afraid of intermittency there. That's the exact scenario where in manual testing partial occlusions weren't getting hit 100% reliably. Messing with slotted content generally but not always got caught by the intersection observer. Sometimes after hitting tab (and not causing a scroll, just tabbing between things in view) would kick it and then make it realize the intersection is partially occluded. And intermittency on these overflow tests and not figuring it out was why the whole jasmine retry infrastructure was added lol.
Despite all that will probably add a test tomo.

packages/nimble-components/src/stepper/index.ts
rajsite: The llm worked so hard on that comment, will nuke

packages/nimble-components/src/stepper/styles.ts
jattasNI: >This is to align with with step icon size so the tabbing between them is continuous.

The part I'm still missing is that the default size of a content-hidden button is already 32px x 32px and `controlHeight` is also 32px. So is this doesn't actually change anything, right? I'm fine to keep it if you prefer to make the sizing explicit, just wanting to understand if it has an impact I'm not seeing.

rajsite: Ah, yea it does have an impact. So we specify height and an aspect-ratio:1/1. The aspect ratio sets the auto behavior, in this case we only have a height defined so width is auto.

In flow layout that makes it a 32x32 box when `content-hidden` is set otherwise it sizes based on content size and aspect-ratio behaves kinda like a min.

In flex layout auto behavior behaves like [align-self: stretch](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-self#stretch) and tries to fill the container

image

The intrinsic content size of the button is only guided by the icon it seems, so none of the other cross-axis sizing properties, i.e. `align-self` does what you would expect. They shrink the button down to the icon size 16x16. Edit: looks like if button had a min-width / min-height configured then the `align-self` options would behave as expected as the min-content size would be 32x32 may be worth doing at some point.

But yea setting width gives it an actual intrinsic size now and can position correctly. Technically we don't need height, but eh, just clearly defining the size here. P.S. wasn't 100% clear on the above before implementing and the sizing of the button stumped me for a bit. Still only 85% sure of the explanation.

packages/nimble-components/src/stepper/tests/stepper.spec.ts
rajsite: Gah, the others (tabs, breadcrumb) I based the PR on had that kind of test and at some point llm removed them. I don't know when, should have re-reviewed all the tests were still there. It also put them in a different order than breadcrumb so it's harder to review.

Put them in order and aligned them. Made the title content change tests only apply in horizontal using a sketchy but maybe reasonable pattern of an if condition around those `it` statements. They are consistently evaluated based on the parameterization so seems okay. Essentially like a feature flag I guess.

packages/nimble-components/src/stepper/index.ts
rajsite: Simplified the comment and added https://github.com/ni/nimble/issues/2922

packages/nimble-components/src/stepper/styles.ts
rajsite: Resolving, not planning on capturing the button min-width / min-height discussion elsewhere but lemme know if you think it's worth it.

rajsite:
@jattasNI thanks for the review, removed @mollykreis since we have a review (but welcome to give feedback if interested!).

Also pushed another tweak to make the stepper a little more robust to invalid orientation assignment (i.e js users manipulating directly without types) by checking orientation explicitly for 'vertical' and otherwise falling back to default of horizontal. But that's not something we have checked / tested for from what I've seen so not adding explicit coverage for it.

rajsite:
@jattasNI ahh one thing I missed was putting the styles into the CSS layers stack for stepper (which I did do for step): https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#organize-styles-with-css-cascade-layers

Something to double check for in new component dev. Should really do that for all the new chat components, imagine it's a straightforward retrofit. I'll keep doing it for the stepper on my personal backlog, not worth a tech debt item.

PR #2923, opened 2026-04-10
 Fix Firefox crash in anchored-region by replacing when directive with CSS-based slot visibility

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

#2744 

In Firefox, using `@mention` in the `rich-text-editor` causes an error in the FAST element, breaking the page.

The root cause is the foundation's [`anchoredRegionTemplate` using a `when` directive](https://github.com/ni/fast/blob/b6d033d38166955c0544173346056475721504fe/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts#L14) to conditionally add/remove the `<slot>` from the shadow DOM based on `initialLayoutComplete`. When `initialLayoutComplete` transitions, `updateContentTarget` calls `view.remove()` on the view containing the `<slot>`. Removing the `<slot>` triggers a slotchange microtask (per the DOM spec). In Firefox, this microtask runs between iterations of FAST's `process()` loop, causing interleaved binding updates that call `view.remove()` a second time on the same view, crashing because the view's nodes have already been moved into the fragment.

Chrome processes these microtasks in a different order that avoids the double-remove.

## 👩‍💻 Implementation

Replaced the foundation's `anchoredRegionTemplate` (which uses a `when` directive) with a local nimble template that keeps the `<slot>` always present in the DOM:

- Always renders `<slot></slot>` without a `when` directive. The `loaded` class is still applied via the `initialLayoutComplete` binding on the host.
- Added slot display toggle styles to control visibility through CSS instead of DOM manipulation.
- Switched from importing `anchoredRegionTemplate` from `@ni/fast-foundation` to the local template.
- This eliminates the `updateContentTarget` / `view.remove()` cycle entirely. No DOM mutations on the `<slot>`, no slotchange events, no race condition.

## 🧪 Testing

- Added tests for the newly updated anchored region template and style updates.
- Manually tested in the storybook, where no console error is thrown, and the app is not frozen. It is also related to the fix in the FAST library: https://github.com/ni/fast/pull/39

### Known Issue

When entering characters after entering `@` that do not match any user in the list, it shows the `No items found` option in the list for the first time. However, when reproducing the same behavior, it shows nothing when finding no options. This is again reproducible only in the Firefox browser.

no-result-issue

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-01bb6111-fa96-46dd-8351-53d8d2e79be4.json
+++ b/change/@ni-nimble-components-01bb6111-fa96-46dd-8351-53d8d2e79be4.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix Firefox crash in rich test mention list box by having a custom anchored region template",
+  "packageName": "@ni/nimble-components",
+  "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/anchored-region/index.ts
+++ b/packages/nimble-components/src/anchored-region/index.ts
@@ -1,8 +1,8 @@
 import {
     DesignSystem,
-    AnchoredRegion as FoundationAnchoredRegion,
-    anchoredRegionTemplate as template
+    AnchoredRegion as FoundationAnchoredRegion
 } from '@ni/fast-foundation';
+import { template } from './template';
 import { styles } from './styles';
 
 declare global {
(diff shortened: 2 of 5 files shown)


change/@ni-nimble-components-01bb6111-fa96-46dd-8351-53d8d2e79be4.json
```diff
@@ -0,0 +1,7 @@
+{
```
vivinkrishna-ni, line 1: Would it make sense to track this known issue under #2744 since it is related and currently reproducible only in Firefox? If so, I can update the issue description to reflect this behavior; otherwise, I can create a new issue/work item to track it separately.

rajsite:
Closing in favor of https://github.com/ni/fast/pull/40

PR #2896, opened 2026-03-14
Document CSS-based sizing for nimble-chip

fredvisser:
# Pull Request

## Rationale

The `nimble-chip` component needed documentation for how to achieve the 24px compact size variant. Per [design guidance](https://github.com/ni/nimble/issues/610), sizing should be done via CSS rather than component attributes, following the same pattern as the spinner component.

## Implementation

Added CSS-based sizing documentation and a Storybook size control:

- **chip.stories.ts**: Added a `size` control under the **Styles** category with "Default (32px)" and "Small (24px)" radio options. Selecting "Small" applies `height: var(--ni-nimble-control-slim-height)` via inline CSS, matching the spinner approach.
- **chip.mdx**: Added a "Sizing" subsection under "Styling" explaining the 32px default and how to achieve 24px using the `control-slim-height` design token.

No component code changes — sizing is purely a consumer-side CSS override using existing design tokens.

## Testing

No new component tests needed since no component code was changed. The sizing control is visible in Storybook.

## Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

PR #2924, opened 2026-04-10
Add Attachments and Footer Actions Slots

hellovolcano:
# Pull Request

## 🤨 Rationale

Adding footer actions and attachments slots in accordance with [the spec](https://github.com/ni/nimble/blob/main/packages/spright-components/src/chat/specs/README.md).

## 👩‍💻 Implementation

Added new sections for attachments and footer-actions and adjusted styling and spacing to accommodate the new sections.

## 🧪 Testing

Added tests

## ✅ Checklist


hellovolcano:
I have not updated the example apps yet because there doesn't appear to be a nimble chip wrapper in react or blazor yet.

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -458,4 +460,58 @@ describe('ChatInput', () => {
             expect(page.getTextAreaTabIndex()).toBeNull();
         });
     });
+
+    describe('footer-actions slot', () => {
+        it('should have a footer-actions slot element in the shadow DOM', async () => {
+            await connect();
+            const footerActionsSlot = element.shadowRoot?.querySelector('slot[name="footer-actions"]');
```
jattasNI: Let's move the `footer-actions` queries into the page object too. The goal should be no `querySelector` calls from the tests themselves.

packages/spright-components/src/chat/input/tests/chat-input.spec.ts
```diff
@@ -458,4 +460,58 @@ describe('ChatInput', () => {
...
+            await connect();
+            const footerActionsSlot: HTMLSlotElement = element.shadowRoot!.querySelector('slot[name="footer-actions"]')!;
+            expect(footerActionsSlot.assignedElements().length).toBe(1);
+        });
+    });
+
+    describe('attachments slot', () => {
+        it('should have an attachments slot element in the shadow DOM', async () => {
+            await connect();
+            const attachmentsSlot = element.shadowRoot?.querySelector('slot[name="attachments"]');
```
jattasNI: Use the page object method here too.

packages/spright-components/src/chat/input/testing/chat-input.pageobject.ts
```diff
@@ -86,6 +86,11 @@ export class ChatInputPageObject {
         return this.element.textArea!.value;
     }
 
+    public getAttachmentsSlotElements(): Element[] {
```
jattasNI: We try to avoid exposing `Element` return types from public page object methods. Instead add methods like `hasAttachmentSlot(): boolean` and/or `getAttachmentsSlotElementCount(): number`.

packages/spright-components/src/chat/input/styles.ts
```diff
@@ -117,4 +122,42 @@ export const styles = css`
...
+        flex-wrap: wrap;
+        column-gap: ${mediumPadding};
+        margin: ${mediumPadding};
+    }
+
+    .footer-actions ::slotted(${buttonTag}),
+    .footer-actions ::slotted(${toggleButtonTag}),
+    .footer-actions ::slotted(${anchorButtonTag}),
+    .footer-actions ::slotted(${menuButtonTag}) {
+        height: ${controlSlimHeight};
```
jattasNI, line 160: Do we have guidance from UX about what other control types should look like when in the attachments slot? To me it looks kind of weird to have slim buttons but not a slim select.

Though if we do want a slim select we'll probably want to add chromatic tests to make sure it looks ok, so that could happen in a separate PR.

packages/spright-components/src/chat/input/styles.ts
```diff
@@ -117,4 +122,42 @@ export const styles = css`
...
         margin: ${mediumPadding};
     }
+
+    .attachments {
+        display: none;
+    }
+
+    .attachments.has-content {
+        display: flex;
+        flex-wrap: wrap;
```
jattasNI, line 128: Do we have guidance from UX on overflow behavior? Wrapping looks fine to me for now, though for other controls like tab and breadcrumb and stepper we've also made them scrollable and added arrow buttons to do the scrolling. This can be a follow up conversation, just want to make sure we're thinking about it. 

But if we're happy with the current behavior, let's add a matrix test state that adds many slotted items so that we get test coverage for that configuration.

Image

packages/storybook/src/spright/chat/input/chat-input.stories.ts
```diff
@@ -19,10 +24,26 @@ interface ChatInputArgs {
...
+
+Spright will set the height of the buttons to \`$ni-nimble-control-slim-height\`.
+`;
+
+const attachmentsDescription = `An area to slot content adjacent to the text input area. Intended to be used for adding chips that represent attached files.
+
+When using removable chips, the chip's \`remove\` event bubbles through the chat input. Listen for the \`remove\` event and remove the chip from the DOM to clear the attachment.
+
+On send button click or Enter key press, the \`send\` event includes the text input value but does not include information about attachments. Implementers will need to track attachments separately and correlate them with the send event as needed.
+`;
```
jattasNI, line 40: This is more detail than we usually include in our parameter docs. My vote would be to remove the last two paragraphs since it's mostly talking about standard DOM behavior and hinting at non-existent APIs. But if you think it's valuable then we could put it in a collapsible summary ([example](https://github.com/ni/nimble/blob/19ccfe6cfad8de187f4eb0d79d7d9178c1812377/packages/storybook/src/nimble/table/table.stories.ts#L238)).

packages/storybook/src/spright/chat/input/chat-input.stories.ts
```diff
@@ -19,10 +24,26 @@ interface ChatInputArgs {
     sendDisabled: boolean;
     errorText: string;
     errorVisible: boolean;
+    footerActions: boolean;
+    attachments: boolean;
     send: undefined;
     stop: undefined;
 }
 
+const footerActionsDescription = `Place 0 or more buttons to the left of the send/stop button.
```
jattasNI: If selects are supported/encouraged we should mention them here. If not, we should remove the select from the example.

packages/storybook/src/spright/chat/input/chat-input.stories.ts
```diff
@@ -46,6 +67,20 @@ export const chatInput: StoryObj<ChatInputArgs> = {
...
+            ${when(x => x.attachments, html`
+                <${chipTag} slot="attachments" removable>Placeholder.txt</${chipTag}>
+                <${chipTag} slot="attachments" removable>Image.png</${chipTag}>
+            `)}
+            ${when(x => x.footerActions, html`
+                <${buttonTag} slot="footer-actions" appearance="ghost" title="Attach" content-hidden>
+                    <${iconAddTag} slot="start"></${iconAddTag}>
+                    Attach
+                </${buttonTag}>
+                <${selectTag} slot="footer-actions" title="Worker of the Week">
```
jattasNI: Nice Simpsons reference! ☢️

packages/spright-components/src/chat/input/styles.ts
jattasNI: The chip also looks pretty bulky to me personally. If UX agrees and wants to slim it down, we can follow the approach in #2896

packages/spright-components/src/chat/input/styles.ts
hellovolcano: UX guidance will be to limit the height of the attachments section to three rows. Anything that overflows should be bumped into a little overflow menu (opened with a button + three dots icon).  Also will be applying the slim control height per the suggestions in that other PR.

Open to suggestions on how to implement that in the component. First thought was maybe a bool in the API that we could set to true on overflow so clients can know when to bump additional attachments into that overflow menu?

packages/spright-components/src/chat/input/styles.ts
hellovolcano: Yes, we are going to go with that approach since the normal chips look pretty bulky.

For now, attachments slots would only have chips + a button

Only known thing we need in the footer-action button + select.

I would assume that list is complete, but I made a change to just apply that height to all slotted items.

packages/spright-components/src/chat/input/styles.ts
hellovolcano: Per our conversation: for initial version, we will leave it up to clients to implement overflow behavior.

packages/storybook/src/spright/chat/input/chat-input.stories.ts
```diff
@@ -19,10 +24,23 @@ interface ChatInputArgs {
...
+    attachments: boolean;
     send: undefined;
     stop: undefined;
 }
 
+const footerActionsDescription = `Place 0 or more controls to the left of the send/stop button.
+
+The buttons should have the \`ghost\` appearance and \`content-hidden\`.
+
+Select controls should have the \`outline\` appearance.
```
jattasNI: Right now we're not obeying this guidance in this story (the select's default appearance is underline and we're not setting it to anything else). We should update the guidance or the actual control so that they match.

packages/spright-components/src/chat/input/styles.ts
jattasNI: I believe we settled on starting with that overflow behavior being a vertical scrollbar with 3 rows of chips visible. I don't think clients have what they need to achieve that: if they want to style the chip container to have a max-height and overflow-y: scroll, they need to be able to target the chip container but it's internal to the shadow DOM right now.

The options I see are:
1. expose a [shadow part](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Shadow_parts) for clients to style
2. do the styling in Spright

I'd prefer the latter so that all clients get it and it's simpler / less API surface. But if you want to allow more experimentation then the former is fine too. I'm ok with you doing one or both of those in a follow up PR if you prefer.

packages/spright-components/src/chat/input/styles.ts
jattasNI: Ideally we'd add Chromatic test coverage for the controls that we know could render with slim height. As a first step, could you add the select to the new matrix story (in whatever appearance it's supposed to be configured with)?

Longer term we should probably add coverage to the button and select's matrix stories, but that can be tracked as part of #610.

jattasNI (approved):

packages/storybook/src/spright/chat/input/chat-input.stories.ts
hellovolcano: Oh yep, that was just a typo on my part. Changed guidance to underline.

packages/spright-components/src/chat/input/styles.ts
hellovolcano: added the overflow and max height

PR #2926, opened 2026-04-13
Componentize example apps

rajsite:
# Pull Request

## 🤨 Rationale

Breaks down the example apps into components based on the existing sections.
Helps make examples more standalone / self-contained for reference.
Makes it possible to have more constrained ownership in the example app.

## 👩‍💻 Implementation

See above.

## 🧪 Testing

Manual test.

## ✅ Checklist


rajsite:
@jattasNI bypassing for non-library facing nights-and-weekends refactor of example app into a smaller set of components. Can address feedback in a follow-on.

Issue #504, opened 2022-04-18
 Tier 1 support for React


samkristoff:
## 😯 Problem to Solve

Tier 1 support for React.

Digilent / MCC is working with a firm to build some of our web infrastructure and leveraging their existing stack which is built with Django and React.  It would be nice to adopt nimble to make it easier to leverage NI's design guidelines.   We're still in definition so I'll add more details about which components we'd use and use case details as we progress.  

## 💁 Proposed Solution

Provide a dedicated library for React, similar to nimble-angular but for React.

fredvisser:
Closing as low ROI, but folks are still welcome to add use cases to help change the priority.

jattasNI:
Not a new use case, but the React app I linked to a year ago [appears to be close to shipping](https://nio365.sharepoint.com/sites/SoftwareDiscipline/SitePages/The-Technologist-2024-C2.aspx?CT=1721702206614&OR=OWA-NT-Mail&CID=f9075473-d57b-fbfc-c8b4-a54626433f3b#codesigner-for-schematic-reuse-in-hw...-coming-soon!) (without Nimble, based on those screenshots)

jattasNI:
There is some talk of using a Nimble query builder (once it's available from #506) in React-based Grafana plugins [in this Asset Calibration HLD PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/757640). If we go that direction we may need to revive this issue.

jattasNI:
I experimented with nimble-react on an innovation day. You can see the changes in the [nimble-react branch](https://github.com/ni/nimble/compare/nimble-react?expand=1).

Notes:
1. I extracted wrappers from storybook into their own package. Doing this required changing storybook's imports from that package and from `nimble-components` to be package dependencies instead of relative path dependencies. This was necessary because we want the `nimble-react` package to have a package dependency on `nimble-components` so our clients can install them like normal (and storybook can't mix package and relative dependencies on `nimble-components` because that results in 2 copies being pulled in). A consequence of this is that storybook would no longer live detect changes to components.
2. I added a next example app but couldn't get its dependency on nimble react to work because I hit [this error](https://stackoverflow.com/questions/31815633/what-does-the-error-jsx-element-type-does-not-have-any-construct-or-call). I also tried the create-react-app example app but it was even more broken.
[...]

PR #2927, opened 2026-04-13
Stepper component wrappers

rajsite:
# Pull Request

## 🤨 Rationale

Adds Stepper, Step, and Anchor Step component wrappers for Angular, Blazor, and React

## 👩‍💻 Implementation

- Adds component wrappers following existing conventions
- As new components in angular, added as secondary entrypoints. To leverage existing anchor base directives had to move them to a secondary entrypoint (`@ni/nimble-angular/anchor-base`)
  -  Implementing routerLink behaviors will require a similar refactor and split out of PR
- Addressed some Blazor build lint errors that only happen locally and not on CI. Not sure why but seemed innocuous.

## 🧪 Testing

For blazor and angular aligned with wrapper test patterns. React does not have an equivalent.

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed. Updated example apps to use simple Stepper example.


packages/blazor-workspace/NimbleBlazor/Components/NimbleListOption.razor.cs
```diff
@@ -34,12 +34,8 @@ public partial class NimbleListOption : ComponentBase
     /// <inheritdoc />
     protected override void OnParametersSet()
     {
-        Context = string.IsNullOrEmpty(Name) ? CascadedContext : CascadedContext?.FindContextInAncestors(Name);
-
-        if (Context == null)
-        {
-            throw new InvalidOperationException($"{GetType()} must have an ancestor {typeof(NimbleSelect)} " +
+        Context = (string.IsNullOrEmpty(Name) ? CascadedContext : CascadedContext?.FindContextInAncestors(Name))
+            ?? throw new InvalidOperationException($"{GetType()} must have an ancestor {typeof(NimbleSelect)} " +
```
rajsite, line 38: Lint really wanted this change. Not personally a huge fan but went with it
```
@ni-private/blazor-workspace:build: C:\dev\github\ni\nimble\packages\blazor-workspace\NimbleBlazor\Components\NimbleListOption.razor.cs(39,9): error IDE0270: Null check can be simplified (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0270) [C:\dev\github\ni\nimble\packages\blazor-workspace\NimbleBlazor\NimbleBlazor.csproj]
```

packages/blazor-workspace/NimbleBlazor/NimbleInputBase.cs
```diff
@@ -49,7 +49,9 @@ public abstract class NimbleInputBase<TValue> : ComponentBase, IDisposable
     /// Gets the associated <see cref="AspNetCore.Components.Forms.EditContext"/>.
     /// This property is uninitialized if the input does not have a parent <see cref="EditForm"/>.
     /// </summary>
+#pragma warning disable IDE0370 // Keep suppression which conflicts with cli
     protected EditContext EditContext { get; set; } = default!;
+#pragma warning restore IDE0370 // Keep suppression which conflicts with cli
```
rajsite, line 54: If you remove the none-null assertion like lint wants then you start getting cascading issues and changes. Didn't spend time looking more closely and opted to suppress the lint for now. It's also not clear to me why local builds on windows have different lints reported than CI. We go to great lengths to lock all deps so something isn't being caught there.

packages/angular-workspace/lighthouserc.cjs
```diff
@@ -17,7 +17,7 @@ module.exports = {
         assert: {
             assertions: {
                 'categories:performance': ['error', { minScore: 0.7 }],
-                'categories:accessibility': ['error', { minScore: 0.8 }]
+                'categories:accessibility': ['error', { minScore: 0.7 }]
```
rajsite, line 20: Lighthouse reports the following new accessibility issues with the stepper added to the angular example app:

Image

Complaining about lists and list items not having direct parent child relationships but if you look at the actual accessibility tree you see of the accessibility tree impacting elements they are direct parent-child relationships:

Image

I'm not sure how to reconcile this, think it's probably an issue to report to axe-core. For now just dropping the threshold.

rajsite: Created the following to track: https://github.com/ni/nimble/issues/2928

packages/angular-workspace/nimble-angular/anchor-step/nimble-anchor-step.directive.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    @Input() public set disabled(value: BooleanValueOrAttribute) {
+        this.renderer.setProperty(this.elementRef.nativeElement, 'disabled', toBooleanProperty(value));
+    }
+
+    public get readOnly(): boolean {
+        return this.elementRef.nativeElement.readOnly;
+    }
+
+    @Input() public set readOnly(value: BooleanValueOrAttribute) {
```
jattasNI, line 47:
```suggestion
    @Input('readonly') public set readOnly(value: BooleanValueOrAttribute) {
```

You could also copy the comment that we've copy-pasted on all of our other directives that implement readonly, but personally I think it's not super useful.
https://github.com/ni/nimble/blob/19ccfe6cfad8de187f4eb0d79d7d9178c1812377/packages/angular-workspace/nimble-angular/src/directives/number-field/nimble-number-field.directive.ts#L32

packages/angular-workspace/nimble-angular/step/nimble-step.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    @Input() public set disabled(value: BooleanValueOrAttribute) {
+        this.renderer.setProperty(this.elementRef.nativeElement, 'disabled', toBooleanProperty(value));
+    }
+
+    public get readOnly(): boolean {
+        return this.elementRef.nativeElement.readOnly;
+    }
+
+    @Input() public set readOnly(value: BooleanValueOrAttribute) {
```
jattasNI, line 46:
```suggestion
    @Input('readonly') public set readOnly(value: BooleanValueOrAttribute) {
```

packages/react-workspace/nimble-react/src/anchor-step/index.ts
```diff
@@ -0,0 +1,6 @@
+import { AnchorStep, anchorStepTag } from '@ni/nimble-components/dist/esm/anchor-step';
```
jattasNI, line 1: When reviewing another PR, copilot noticed that we don't re-export our enums like severity from our React wrappers like we do in our Angular wrappers. I don't think it's come up yet because the react example app typically uses literals for enum values.

Should we start? You don't have to fix all of them in this PR unless you're motivated to. And if you don't even want to think about it now, we could just add it to #504.

jattasNI (approved):

packages/react-workspace/nimble-react/src/anchor-step/index.ts
rajsite: Added a task to the React v1 issue

packages/angular-workspace/nimble-angular/anchor-step/nimble-anchor-step.directive.ts
rajsite: Since stepper is still marked incubating and actively working on follow-up directive changes for routerLink, will capture there.

packages/angular-workspace/nimble-angular/step/nimble-step.directive.ts
rajsite: Since stepper is still marked incubating and actively working on follow-up directive changes for routerLink, will capture there.

rajsite:
@msmithNI @mollykreis  bypassing blazor owners review for pretty straightforward blazor component wrappers. When you have a chance to review can address concerns in a follow-up.

PR #2934, opened 2026-04-18
Re-organize Blazor file layout

### LINKED ISSUE #2865: Blazor source file layout improvements

rajsite:
# Pull Request

## 🤨 Rationale

Splits the file re-organization from #2892 since that was backed out for now.
Resolves #2865

## 👩‍💻 Implementation

Applies the same folder layout approved previously

## 🧪 Testing

Rely on CI.

## ✅ Checklist


rajsite:
@msmithNI bypassing for file layout change that was already approved but if you have feedback can address in a follow up

Issue #2047, opened 2024-04-29
`SprightBlazor` should have a dependency on `NimbleBlazor`


m-akinc:
# 🧹 Tech Debt

At some point, it is likely that the `SprightBlazor` API might include types (e.g. enums) that are defined in `NimbleBlazor`. At that point, we will need `SprightBlazor` to depend on `NimbleBlazor`. One issue to figure out is how to create a project dependency (`ProjectReference` in `csproj`) that results in the nuget having a dependency with the correct version range. `ProjectReference` doesn't support specifying a version range.

rajsite:
Some investigation notes, [creating a ProjectReference](https://markheath.net/post/multiple-nuget-single-repo) from SprightBlazor to NimbleBlazor:
- After running build the lock files show a transitive dependency on NimbleBlazor (makes sense) but they all say version 1.0.0, like `"NimbleBlazor": "[1.0.0, )"`
- Version issue is probably because versions are not specified in the csproj, they are specified only during publish via the `npm run pack` command: `cross-env-shell dotnet pack -c Release -p:PackageVersion=$npm_package_version --output ../dist`
- Maybe we can [specify version in a props file](https://stackoverflow.com/a/67055157)?
- Problem then is lock file versions being out of date after publish. May need a custom beachball plugin to make sure those updates are submitted after publish (like we used to do manually for npm lock files previously).

PR #2940, opened 2026-04-22
Add react and blazor wrapper for nigel chat icon

Alexia-Claudia-Micu:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/3835544

## 👩‍💻 Implementation

Add react and blazor wrapper for nigel chat icon

## ✅ Checklist


packages/blazor-workspace/SprightBlazor/Source/Icons/SprightIconNigelChat.razor.cs
```diff
@@ -0,0 +1,12 @@
+using Microsoft.AspNetCore.Components;
+
+namespace SprightBlazor;
+
+public partial class SprightIconNigelChat : ComponentBase
```
jattasNI: I would vote to derive from [NimbleIconBase](https://github.com/ni/nimble/blob/1cd470c5a3398356f76ba725ff9ae85a0d054719/packages/blazor-workspace/NimbleBlazor/NimbleIconBase.cs) and remove the AdditionalAttributes from this class because the base class has it. Then you could also change the template to match what NimbleBlazor generated icons do (keeping your icon name and replacing Nimble with Spright, obviously):

```
@namespace NimbleBlazor
@inherits NimbleIconBase
<nimble-icon-arrow-down
    severity="@Severity.ToAttributeValue()"
    @attributes="AdditionalAttributes">
</nimble-icon-arrow-down>
 ```

This has a minor downside of exposing the severity property which this icon doesn't respect, but we [have precedent for that](https://nimble.ni.dev/storybook/index.html?path=/docs/ok-ts-icon-dynamic--docs). And I think the benefit of consistency with other icons outweighs that.

**Review by jattasNI — APPROVED** [review]

File: packages/react-workspace/spright-react/src/icons/nigel-chat/index.ts [hunk 2]
```diff
@@ -0,0 +1,4 @@
+import { IconNigelChat } from '@ni/spright-components/dist/esm/icons/nigel-chat';
```
rajsite: Can you also export the class as a type and the tagname like other icons do?

```
// AUTO-GENERATED FILE - DO NOT EDIT DIRECTLY
// See generation source in nimble-react/build/generate-icons

import { IconArrowDown, iconArrowDownTag } from '@ni/nimble-components/dist/esm/icons/arrow-down';
import { wrap } from '../../utilities/react-wrapper';

export type { IconArrowDown };
export { iconArrowDownTag };
export const NimbleIconArrowDown = wrap(IconArrowDown);
```

rajsite (approved):

packages/blazor-workspace/SprightBlazor/Source/Icons/SprightIconNigelChat.razor.cs
rajsite: Ah, so actually there are challenges we haven't resolved with cross assembly packages in the workspace. https://github.com/ni/nimble/issues/2047

Should revert the project reference change and lock file updates. Can either leave how you had it before or copy NimbleIconBase into SprightBlazor to extend from. I'd vote copying it in since there are other spright icons that may use it eventually as well.

packages/react-workspace/spright-react/src/icons/nigel-chat/index.ts
Alexia-Claudia-Micu: Implemented changes.

packages/blazor-workspace/SprightBlazor/Source/Icons/SprightIconNigelChat.razor.cs
Alexia-Claudia-Micu: Removed the lock files and added `SprightIconBase`.

jattasNI (approved):

PR #2942, opened 2026-04-27
Add containerBackgroundColor token

fredvisser:
# Pull Request

## 🤨 Rationale

Block appearance on the [container/card design](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=5069-8503&m=dev) calls for a container background color that isn't currently represented by a token.

## 👩‍💻 Implementation

Pulled colors from the Figma design and mapped to our base tokens.

Used `container` naming vs `card` naming, preferring the affinity to the `applicationBackgroundColor` and `sectionBackgroundColor` vs calling it a `cardBackgroundColor` or `cardFillColor`.

Opinions welcome

## 🧪 Testing

Manual Storybook token review

## ✅ Checklist

### DIFF
--- a/packages/nimble-components/src/theme-provider/design-token-comments.ts
+++ b/packages/nimble-components/src/theme-provider/design-token-comments.ts
@@ -6,6 +6,7 @@ export const comments: { readonly [key in TokenName]: string } = {
     actionRgbPartialColor:
         'DEPRECATED: *-partial tokens are used with rgba() to set color transparency in component stylesheets',
     applicationBackgroundColor: 'Primary background color for the application',
+    containerBackgroundColor: 'Background color for container surfaces layered above the application background',
     dividerBackgroundColor: 'Background color for static / non-movable dividers',
     dividerBackgroundDynamicColor: 'Background color for dynamic / movable dividers',
     headerBackgroundColor: 'Background color for application headers',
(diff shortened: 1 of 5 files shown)


packages/nimble-components/src/theme-provider/design-token-comments.ts
jattasNI: I don't think I could tell the difference between a container and a section from this description. I don't know what VxD's vision is, but here's a stab at something that sounds logical to me.

`containerBackgroundColor: 'Background color for containers. Containers are smaller groupings than sections.',`

`sectionBackgroundColor: 'Background color for sections. Sections are groupings within an application.',`

jattasNI (approved):

rajsite (approved):

packages/nimble-components/src/theme-provider/design-token-comments.ts
rajsite: Alternatively if this is just a variation on section then maybe we could do `sectionBlockBackgroundColor`.

Whichever aligns with the design intention, agree I can't tell just looking at the figma

fredvisser: How about "Background color for block-appearance component container surfaces, such as cards"?

PR #2943, opened 2026-04-27
Mark all react components as 'use client'

rajsite:
# Pull Request

## 🤨 Rationale

Nimble components rely on running client side so mark all the Nimble / Spright / Ok components with [`'use client';`](https://react.dev/reference/rsc/use-client)

## 👩‍💻 Implementation

- Update each file and the icon build script to be marked as `use client`

## 🧪 Testing

- Rely on CI to avoid regressions
- Manual testing from React Server Components (RSC) and Next.JS users

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-react-89ce99f4-0fd5-42de-8eff-3a1ef4a7dfdf.json
+++ b/change/@ni-nimble-react-89ce99f4-0fd5-42de-8eff-3a1ef4a7dfdf.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Mark all react components as 'use client'",
+  "packageName": "@ni/nimble-react",
+  "email": "1588923+rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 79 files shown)


rajsite:
@jattasNI bypassing for trivial change to react components. Can address any concerns in follow-up

rajsite:
Just capturing some notes, looks like for a next.js hello world one also needs to disable SSR for the Nimble components: https://blog.bitsrc.io/using-non-ssr-friendly-components-with-next-js-916f38e8992c

One way is to do that on import using a next.js specific pattern, not clear to me yet if there is a more generic react server component pattern:

image

```
'use client'
 
import dynamic from 'next/dynamic'
 
export const NimbleButton = dynamic(() => import('@ni/nimble-react/button').then(mod => mod.NimbleButton), { ssr: false,});
export const NimbleThemeProvider = dynamic(() => import('@ni/nimble-react/theme-provider').then(mod => mod.NimbleThemeProvider), { ssr: false,});
 ```

But with the above does seem to work: https://stackblitz.com/edit/nextjs-nimble-nossrimport?file=app%2Fpage.tsx&file=app%2Fnimble.ts

PR #2963, opened 2026-05-21
Add Ok Breakpoint Column

hellovolcano:
# Pull Request

## 🤨 Rationale

A client needs a breakpoint column for debugging workflows.
[HLD](https://github.com/ni/nimble/blob/main/packages/ok-components/src/ts/table-column/breakpoint/specs/HLD.md) + [IxD](https://github.com/ni/nimble/blob/main/packages/ok-components/src/ts/table-column/breakpoint/specs/IxD.md)

## 👩‍💻 Implementation

This component mimics the nimble menu button component, but in addition to the toggle events on click, also includes a context menu event to allow clients to set additional states.

Added:
- A fixed width table column
- toggle events to switch between enabling/adding a breakpoint and removing
- oncontextmenu events to track right-click/context menu actions
- BreakpointStateEnum to track states and visual styles for breakpoint states
- Blazor wrappers and examples

## 🧪 Testing

manual testing
added unit tests

## ✅ Checklist


jattasNI:
@hellovolcano, Here's @msmithNI 's innovation day branch for an ok custom column type (aka a square breakpoint column 🟥): https://github.com/ni/nimble/compare/main...users/masmith/innovationDay-okTableColumn1

I haven't looked at either implementation but it could be useful to scan it for differences with your approach and ping him with any questions. (Note when we talked I misspoke about who had done this project, it is definitely Malcolm)

hellovolcano:
FYI @jattasNI -- I know we originally talked about adding a label provider for ok. I'm open to adding that later, but for this first review I just used stringsin the component itself.

change/@ni-ok-components-744c8788-8d43-48b6-9bff-7f72757113b5.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Can you update the component status table to mention this component and show its blazor support?

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
```diff
@@ -0,0 +1,172 @@
...
+            <${tableColumnTextTag} field-name="code">
+                Code
+            </${tableColumnTextTag}>
+        </${tableTag}>
+    `),
+    argTypes: {
+        fieldName: {
+            name: 'field-name',
+            description:
+                'Set this attribute to identify which field in the data record contains the breakpoint state value for each row.',
```
jattasNI:
```suggestion
                'Set this attribute to identify which field in the data record contains the breakpoint state value for each row. See the **Usage** section below for valid breakpoint states.',
```

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
```diff
@@ -0,0 +1,172 @@
...
+                    const event = c.event as CustomEvent<BreakpointToggleEventDetail>;
+                    const detail = event.detail;
+                    x.currentData = x.currentData.map(record => (record.id === detail.recordId
+                        ? { ...record, breakpointState: detail.newState }
+                        : record));
+                    void x.tableRef.setData(x.currentData);
+                }}"
+            >
+            </${tsTableColumnBreakpointTag}>
+            <${tableColumnTextTag} field-name="code">
```
jattasNI: nit: you might disable sorting on this column too. It was handy to test that the breakpoint column does reorder when another column is sorted, but that's not a use case we want to encourage.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
```diff
@@ -0,0 +1,172 @@
...
+        code: '}',
+        breakpointState: BreakpointState.off
+    }
+];
+
+const metadata: Meta<SharedTableArgs> = {
+    title: 'Ok/Ts Table Column: Breakpoint',
+    decorators: [withActions<HtmlRenderer>],
+    parameters: {
+        actions: {
```
jattasNI, line 90: If I click or right-click on the column in [the storybook sub-page that shows the Actions tab](https://60e89457a987cf003efc0a5b-ajigwensnm.chromatic.com/?path=/story/ok-ts-table-column-breakpoint--breakpoint-column), I'm seeing two events per click. I'm not sure if that's a Storybook bug (we've seen things like that in the past where if you navigate around you end up with multiple copies of the component being tracked by Storybook). But it might be worth verifying that the events are behaving correctly in a real app.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint-matrix.stories.ts
```diff
@@ -0,0 +1,76 @@
...
+    <${tableTag} id-field-name="id" style="height: 320px">
+        <${tsTableColumnBreakpointTag}
+            field-name="breakpointState"
+        >
+            BP
+        </${tsTableColumnBreakpointTag}>
+    </${tableTag}>
+`;
+
+export const themeMatrix: StoryFn = createMatrixThemeStory(
```
jattasNI, line 66: I'm guessing the client app only supports light theme? [The storybook](https://60e89457a987cf003efc0a5b-ajigwensnm.chromatic.com/?path=/story/tests-ok-ts-table-column-breakpoint--theme-matrix) shows that some of the breakpoint icons are hard to see in other themes. For an ok component that's perfectly acceptable, just pointing it out in case it's not expected.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
```diff
@@ -0,0 +1,172 @@
+import { html, ref } from '@ni/fast-element';
```
jattasNI, line 1: For the 'fv' ok components we started adding a disclaimer explaining how they are special purpose and of varying quality. Could you add that to this component and the ts icon dynamic? There's a shared helper to make the disclaimer consistent.

packages/blazor-workspace/Examples/Demo.Shared/Pages/Sections/Ts/TsBreakpointTableSection.razor
```diff
@@ -0,0 +1,43 @@
...
+            <NimbleTableColumnText FieldName="Name" FractionalWidth="2">Name</NimbleTableColumnText>
+            <OkTsTableColumnBreakpoint
+                FieldName="BreakpointState"
+                BreakpointToggle="OnBreakpointToggle"
+                BreakpointContextMenu="OnBreakpointContextMenu">
+            </OkTsTableColumnBreakpoint>
+            <NimbleTableColumnNumberText FieldName="LineNumber">Line</NimbleTableColumnNumberText>
+        </NimbleTable>
+
+        @if (_contextMenuOpen)
```
jattasNI: Up to you how much effort you want to put into the UX of the example app, but I noticed that if you right click on a breakpoint while an existing menu is open, the menu remains visible and the right click causes the browser context menu to appear rather than the breakpoint menu. I was expecting the previous menu to be dismissed and a new one to appear at the new click location.

packages/blazor-workspace/Examples/Demo.Shared/Pages/Sections/Ts/TsBreakpointTableSection.razor
```diff
@@ -0,0 +1,43 @@
...
+                    }
+                    @if (_contextMenuRecordState == OkBlazor.BreakpointState.Disabled
+                        || _contextMenuRecordState == OkBlazor.BreakpointState.HitDisabled)
+                    {
+                        <NimbleMenuItem @onchange="@(_ => OnEnableBreakpoint())">Enable breakpoint</NimbleMenuItem>
+                    }
+                }
+            </NimbleMenu>
+        }
+        <p><strong>Last event:</strong> @_lastEvent</p>
```
jattasNI: nit: there's a lot of white space between the table and the "Last event"

packages/ok-components/src/ts/table-column/breakpoint/styles.ts
```diff
@@ -0,0 +1,12 @@
+import { css } from '@ni/fast-element';
+import { display } from '../../../utilities/style/display';
+
+export const styles = css`
```
jattasNI: This looks similar to what's in table-column base styles. Can you just use those instead of duplicating or is there a behavior difference or some other complication that I'm missing?

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
```diff
@@ -0,0 +1,637 @@
...
+        return await fixture<Table<SimpleTableRecord>>(
+            html`<${tableTag} ${ref('table')} id-field-name="id" style="width: 700px">
+                    <${tsTableColumnBreakpointTag} ${ref('column')} field-name="breakpointState">
+                    </${tsTableColumnBreakpointTag}>
+                </${tableTag}>`,
+            { source }
+        );
+    }
+
+    function getBreakpointButton(
```
jattasNI: Our typical code organization is to put helpers like this in a page object which helps keep the test file clean and enables them to be used by clients too. Your call whether that's valuable for this component.

If you do this then our preference is to keep methods that return element types as `private` and instead expose public methods like `clickBreakpointButton()` `rightClickBreakpointButton()` or `getBreakpointButtonIconTag()` which match user actions and return primitive types or void.

jattasNI: If you do that then it might also be useful to add a method that returns `(getRenderedCellView() as TsTableColumnBreakpointCellView).currentState` because it looks like a large percentage of this file is repeating that code across several lines.

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
```diff
@@ -0,0 +1,637 @@
...
+        expect(elementReferences.column.checkValidity()).toBeTrue();
+    });
+
+    it('uses fixed icon width and disables resizing', () => {
+        expect(elementReferences.column.columnInternals.resizingDisabled).toBeTrue();
+        expect(elementReferences.column.columnInternals.pixelWidth).toBe(singleIconColumnWidth);
+        expect(elementReferences.column.columnInternals.minPixelWidth).toBe(singleIconColumnWidth);
+    });
+
+    describe('rendering breakpoint states', () => {
```
jattasNI, line 101: You could consider using `paramaterizeSpec` to avoid duplication in these tests.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.mdx
```diff
@@ -0,0 +1,36 @@
...
+import * as breakpointColumnStories from './ts-table-column-breakpoint.stories';
+import ComponentApisLink from '../../../docs/component-apis-link.mdx';
+import { tsTableColumnBreakpointTag } from '@ni/ok-components/dist/esm/ts/table-column/breakpoint';
+import { Tag } from '../../../utilities/story-layout';
+
+<Meta of={breakpointColumnStories} />
+<Title of={breakpointColumnStories} />
+
+The <Tag name={tsTableColumnBreakpointTag}/> renders a breakpoint indicator in each row that
+can be toggled between breakpoint states by click or keyboard interaction.
```
jattasNI, line 11: It could be useful to document what the keyboard interaction is.

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
```diff
@@ -0,0 +1,637 @@
...
+            const cellView = tablePageObject.getRenderedCellView(
+                0,
+                0
+            ) as TsTableColumnBreakpointCellView;
+            expect(cellView.currentState).toBe(BreakpointState.off);
+        });
+    });
+
+    describe('click-to-toggle', () => {
+        it('emits toggle event from off to enabled on click', async () => {
```
jattasNI: Some of these toggle tests could be combined with `parameterizeSpec` too. The hierarchical data one might be different enough that it stays a separate case.

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
```diff
@@ -0,0 +1,148 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        'ok-ts-table-column-breakpoint-cell-view': TsTableColumnBreakpointCellView;
+    }
+}
+
+/**
+ * Cell view for the breakpoint column that renders a clickable breakpoint indicator.
+ */
+export class TsTableColumnBreakpointCellView extends TableCellView<
```
jattasNI, line 27: This column seems to break the tab behavior of the table:
1. If I tab until a breakpoint has focus, I can't find any way to get out of that focus
2. The styling of the focused button doesn't match any other Nimble focus styling (blue box in my browser instead of green box)

2 is up to you to decide what you want but 1 feels consequential enough that we should find a fix to avoid breaking existing table UX workflows. Malcolm might be a good person to talk to if you need help as he built a lot of the table focus code.

Image

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
```diff
@@ -0,0 +1,637 @@
...
+            );
+            await waitForUpdatesAsync();
+
+            expect(contextMenuSpy).toHaveBeenCalledTimes(1);
+            const eventDetail = getContextMenuEventDetail(contextMenuSpy);
+            expect(eventDetail.recordId).toBe('1');
+            expect(eventDetail.currentState).toBe(BreakpointState.enabled);
+        });
+
+        it('emits breakpoint-column-context-menu on Menu key', async () => {
```
jattasNI: There are a couple of possibly platform-specific behaviors here. On macOS there isn't a Menu key and the typical shortcut modifier is Command rather than Control.

In terms of user behavior, it's up to you whether to support mac-equivalent shortcuts.

In terms of test behavior, the tests might actually pass on macOS if they only rely on simulated events and not system behavior. If they do fail on macOS we have a way to disable them via a tag. You're welcome to try running them if you have access to a mac or you can submit them as-is and wait for #1747 to happen, at which point someone will try running them on mac for you.

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
```diff
@@ -0,0 +1,148 @@
...
+        if (value && Object.values(BreakpointState).includes(value as BreakpointState)) {
+            return value as BreakpointState;
+        }
+        return BreakpointState.off;
+    }
+
+    /** @internal */
+    public get tooltipText(): string {
+        if (this.currentState === BreakpointState.off) {
+            return 'Add breakpoint';
```
jattasNI: The downside of hardcoding strings rather than adding a label provider is that there isn't any way for clients to localize them. If that's not a requirement then this is fine.

jattasNI (commented):
Reviewed the ok-components changes with some rigor (not quite normal Nimble level but more than normal Ok level). Only just skimmed the Blazor changes.

Most of the feedback is nice-to-have for the Ok level with the exception of the tab focus getting stuck on the breakpoint.

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
rajsite: At the very least lets make them `@observable` private / protected string variables on the element which positions well for those observable strings to be updated by a label provider in a follow-up. Also allows for testing updates to those strings today.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
hellovolcano: I confirmed in the blazor app with a console log that the event is only firing once.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint-matrix.stories.ts
hellovolcano: I noticed that too and reached out to Brandon. Thanks for pointing it out.

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
hellovolcano: looks like my problem was not returning a value with my keydown event on the breakpoint column. Once I returned a bool value the keydown event for escape/tab was properly handled by the keyboard navigation manager

packages/ok-components/src/ts/table-column/breakpoint/styles.ts
hellovolcano: removed

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint-matrix.stories.ts
hellovolcano: will follow up with the change needed to make these look better in dark theme

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
hellovolcano: I guess as long as it's okay, I'll just wait until someone runs them on mac (I don't have access to a work mac right now), and we can disable them then if needed.

packages/blazor-workspace/Examples/Demo.Shared/Pages/Sections/Ts/TsBreakpointTableSection.razor
hellovolcano: I refactored this a bit because the behavior in the example didn't have good UX. The pattern now more closely matches the pattern of the menu button/column, which provides for significantly better client experience + UX.

packages/blazor-workspace/Examples/Demo.Shared/Pages/Sections/Ts/TsBreakpointTableSection.razor
hellovolcano: removed that

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
jattasNI: I confirmed that you can now tab to a breakpoint and toggle it via space/enter/arrow keys to show menu. Or you can press Esc to move focus out to the cell level and then move focus around between rows with the up and down arrows. That seems aligned with other table interactions now.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
```diff
@@ -0,0 +1,197 @@
...
+            control: false,
+            table: { category: apiCategory.events }
+        },
+        contextMenuEvent: {
+            name: 'breakpoint-column-context-menu',
+            description:
+                'Emitted when the breakpoint context menu is requested. The event detail includes `recordId` and `currentState`.',
+            control: false,
+            table: { category: apiCategory.events }
+        },
```
jattasNI, line 198: With the new approach where the menu is provided in a slot I'm less clear about the purpose of the context menu event callback. It might be helpful to clarify what you can / must do in the callback. Do you need to show the menu? Use the opportunity to populate the menu contents?

packages/ok-components/src/ts/table-column/breakpoint/testing/ts-table-column-breakpoint.pageobject.ts
```diff
@@ -0,0 +1,87 @@
+import type { TablePageObject } from '@ni/nimble-components/dist/esm/table/testing/table.pageobject';
```
jattasNI, line 1: Page objects typically go in /testing rather than /tests which allows them to be exported and used by clients

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
```diff
@@ -0,0 +1,470 @@
...
+            await table.setData([
+                { id: '1', breakpointState: BreakpointState.off }
+            ]);
+            await waitForUpdatesAsync();
+
+            expect(breakpointPageObject.getTabbableChildrenCount(0, 0)).toBe(1);
+        });
+    });
+
+    describe('context menu', () => {
```
jattasNI, line 285: I'm only seeing asserts about the events getting fired, not the state of the context menu itself. It's possible I'm missing something and it's up to you how much test coverage you want.

jattasNI (approved):
On this pass I mostly reviewed for Storybook and "ok" level code/testing.

packages/storybook/src/ok/ts/table-column-breakpoint/ts-table-column-breakpoint.stories.ts
hellovolcano: good catch -- refactoring this to a change requested event and clarifying docs

packages/ok-components/src/ts/table-column/breakpoint/tests/ts-table-column-breakpoint.spec.ts
hellovolcano: added some test cases for the state of the context menu

jattasNI:
@rajsite @msmithNI bypassing you on this ok PR with a couple mechanical changes to non-ok Blazor files.

Issue #848, opened 2022-12-02
Pinned columns


nate-ni:
## 😯 Problem to Solve
This is a development configuration freezing the first/last column (or first/last **_n_** columns) in a table. This is helpful when a table is very wide and the first column should remain visible when horizontally scrolling.

From table spec this is phrased, "Pin columns: columns remain fixed during horizontal scrolling"

## 💁 Proposed Solution
- [x] #892

## 🤔 Open Questions
1. Have we enabled horizontal scrolling?
2. Do we need to be able to enable pinning for **_last_** columns?
   - It's not unheard of, but is this something we need?
   - https://material.angular.io/components/table/overview#sticky-rows-and-columns
4. Should there be an interactive version of this feature (so the end user can "pin/freeze" a column)?

## 🖼️ References/Prior Art
https://user-images.githubusercontent.com/109235103/205920352-456669aa-0586-466c-9792-9088c1d9a2dd.mov

atmgrifter00:
Rough prototype from innovation day: [https://github.com/ni/nimble/tree/table-pinned-columns](https://github.com/ni/nimble/tree/table-pinned-columns) ([Storybook](https://60e89457a987cf003efc0a5b-cohvfnnpop.chromatic.com/?path=/story/incubating-table-column-configuration--grouping)).

Things of note:
- Using 'sticky' for the frozen entities seems like the only viable implementation route (as opposed to something like separate containers for pinned and unpinned elements).
- Using display of `inline-grid` seems necessary to get sticky behavior to work properly (at least for containers that were using `grid` before).
- Offsetting header row using CSS `left` for scrolling purposes interferes with setting CSS `left` for making header content sticky (the latter of which seems necessary). Instead of setting the CSS `left` property, we can instead set the `scrollLeft` of the element containing all of the header content (requiring implementation in code-behind). This seems like it also might require its CSS `overflow` to be explicitly be set to `hidden`.
- Cell elements will likely require having their background explicitly set so that appropriate visual clipping occurs when a scrollable column is dragged passed a pinned column.
- Assumption is that all pinned columns will be rendered in order before unpinned columns (not implemented in prototype)

Visual things to finalize:
- Should there be any explicit visual indicator denoting the border between pinned and unpinned columns?
- Perhaps things like the group rows and collapse-all/selection checkboxes should _always_ be sticky/pinned. Only the cell content can actually be scrolled.

Remaining unprototyped work:
[...]

PR #2964, opened 2026-05-21
Pinning Fixed Width Columns

hellovolcano:
# Pull Request

## 🤨 Rationale

This PR adds support for pinning fixed-width table columns to the left side of the Nimble table, with primary support for icon-sized mapping columns.

Works toward addressing #848.

## 👩‍💻 Implementation

Added a sticky pinned column. When a column is pinned, it is the only thing that is sticky. Multiple fixed width columns can be pinned.

- Implemented pinned-column rendering/layout in the table so pinned columns are separated from scrollable columns, rendered as sticky, and sized from pixel widths.
- Updated row/group-row behavior so indentation is applied at the first non-pinned column and group rows account for pinned-column offset.
- Added table validation for pinned-column configuration (invalidPinnedColumnConfiguration) to catch pinned columns without a resolvable fixed width.
- Updated table update tracking so pinned state changes trigger required layout/focus recalculations.
- Updated Storybook mapping-column docs/examples to expose pinned property.
- Added angular and blazor wrappers

## 🧪 Testing

Added unit tests for pinned columns to table test spec
Added matrix tests to catch multiple combinations of table configuration
Manually verified:
- Pinned column appears correctly when grouping, multi-select, and hierarchy are enabled
- First non-pinned column has hierarchy indentation applied correctly
- Pinned column is the only sticky column

## ✅ Checklist


change/@ni-nimble-components-817561ce-db7b-4dc2-866f-c794a24b0428.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Looks like there are still some styling opacity issues with group rows on hover when there is a pinned column. (Note the arrow button lighter grey on the hovered row)

Image

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.stories.ts
```diff
@@ -142,6 +143,11 @@ export const mappingColumn: StoryObj<MappingColumnTableArgs> = {
             description: widthModeDescription,
             table: { category: apiCategory.attributes }
         },
+        pinned: {
```
jattasNI: I think the right place for these docs is in [Table Column Configuration](https://nimble.ni.dev/storybook/?path=/docs/components-table-column-configuration--docs). That's where we describe the common column API and provide examples of each capability in actual use.

jattasNI (commented):
I ran out of time to do a full review today. Just pointing out a couple things I noticed when playing with storybook.

mollykreis:
I also ran out of time to review this today, but an additional observation from looking at storybook is that I think the "Collapse All" button in the table header should be pinned so it stays aligned with the "collapse" button on each group row (or at least the rows with the outer-most level of grouping)

image

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.stories.ts
```diff
@@ -73,6 +73,7 @@ interface MappingColumnTableArgs extends SharedTableArgs {
     fieldName: string;
```
jattasNI, line 74: We should try to beef up our visual test coverage to validate as many of the corner cases that require custom styling as we can. 

Ideally that would involve a matrix test that shows a pinned column along with various states for selection, grouping, hierarchy, hover, and themes. Ideally the content would be scrolled horizontally to make sure the clipping behind the pinned column is working.

There might be some gaps in our infrastructure to achieve all of that. For example: matrix tests may not yet support a Storybook `play()` function which would be the natural way to perform the horizontal scroll. And it might not be possible to put shadow dom parts like rows in the hover state. If you hit roadblocks like that feel free to reach out to one of us to brainstorm.

packages/nimble-components/src/table/styles.ts
```diff
@@ -121,6 +121,15 @@ export const styles = css`
         visibility: visible;
     }
 
+    .pinned-columns-header-container {
```
jattasNI: Molly, Milan, and shared an opinion that the way unpinned columns are clipped behind the pinned column looks a bit abrupt right now. You mentioned you're working with VxD on styling. Could you check if they have ideas how to delineate pinned and unpinned columns with some sort of shadow or divider to avoid sharply clipped content?

Image

packages/nimble-components/src/table/template.ts
```diff
@@ -49,9 +49,26 @@ export const template = html<Table>`
             --ni-private-table-row-grid-columns: ${x => (x.rowGridColumns ? x.rowGridColumns : '')};
             --ni-private-table-cursor-override: ${x => (x.layoutManager.isColumnBeingSized ? 'col-resize' : 'default')};
             --ni-private-table-scrollable-min-width: ${x => x.tableScrollableMinWidth}px;
+            --ni-private-table-pinned-columns-row-grid-columns: ${x => x.pinnedColumnsGridTemplateColumns};
             ">
             <div role="rowgroup" class="header-row-container">
                 <div class="header-row" role="row">
+                    <div class="pinned-columns-header-container">
```
jattasNI: In Firefox (I saw it on macOS, Molly saw it on Windows), we noticed that the header of the pinned column isn't laying out correctly. It scrolls with the unpinned headers initially. Then if you hover over it, its position refreshes to correctly be pinned.

https://github.com/user-attachments/assets/2817e51a-8b9b-4a6a-9e2f-cec7b0c58e18

change/@ni-nimble-angular-3b488452-3822-410a-a684-15d5f131dd98.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: FYI I left some API design feedback in the linked isue: https://github.com/ni/nimble/issues/848#issuecomment-4595571069

Some of it requires changes to the approach in the PR, hopefully not too big though. The main thing is changing the attribute from a boolean to a string enum.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.stories.ts
hellovolcano: removed and added section to table column configuration

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.stories.ts
hellovolcano: added matrix tests to for all states; left out the scrolling states for now

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
```diff
@@ -902,3 +961,61 @@ export const width: StoryObj<ColumnWidthTableArgs> = {
         }
     }
 };
+
+const pinnedColumnsDescription = 'Set `pin-location` to `left` on one or more fixed-width columns to pin them to the left side of the table. Pinned columns remain visible as the user scrolls the table horizontally. Only columns with a fixed pixel width (i.e. those that do not use `fractional-width`) support pinning.';
```
jattasNI: The relationship between `fractional-width` and fixed width is kind of confusing so I suggest we don't get into it here.

The backstory: for most column types, `fractional-width` defaults to 1 when not provided, so even if a column doesn't set it, it will still behave with fractional width. Certain column types (icon only mapping, breakpoint) opt out of fractional width entirely. Even setting `fractional-width` explicitly on those columns has no effect.

```suggestion
const pinnedColumnsDescription = 'Set `pin-location` to `left` on one or more fixed-width columns to pin them to the left side of the table. Pinned columns remain visible as the user scrolls the table horizontally. Only columns with a fixed pixel width support pinning.';
```

packages/nimble-components/src/table/components/group-row/template.ts
```diff
@@ -15,10 +15,15 @@ export const template = html<TableGroupRow>`
         role="row"
         @click=${x => x.onGroupExpandToggle()}
         aria-expanded=${x => x.expanded}
```
jattasNI, line 17: I'm seeing an issue where group row selection checkboxes aren't horizontally scrolling properly. This is in Firefox using [the Angular example app](https://60e89457a987cf003efc0a5b-fmdyjbpeqa.chromatic.com/example-client-app/index.html#/customapp) and using dev tools to set the icon column to be `pin-location="left"`.

https://github.com/user-attachments/assets/676d35a7-65fb-40f8-8d70-98e548db4ad0

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
```diff
@@ -0,0 +1,224 @@
...
+                field-name="quote"
+            >
+                Quote
+            </${tableColumnTextTag}>
+        </${tableTag}>
+    `;
+};
+
+// Play function to set data on all tables in the matrix
+const playFunction = async (): Promise<void> => {
```
jattasNI, line 236: Could you look into updating the play function to:
1. on tables that support selection, select at least one visible row
2. grow the width of a column so that the table shows a horizontal scrollbar and then scroll horizontally so that a column is clipped behind the pinned column

Both would be really nice to validate those styles. Then the only styling test gap left I can think of would be hover state, but that's harder to simulate reliably.

Feel free to comment back if these end up being difficult to make reliable or completely unachievable, but I think it's at least worth a try.

packages/blazor-workspace/NimbleBlazor/Source/TableColumn/Mapping/NimbleTableColumnMapping.razor.cs
```diff
@@ -17,6 +17,12 @@ public partial class NimbleTableColumnMapping<TKey> : NimbleTableColumnEnumBase<
     [Parameter]
     public double? MinPixelWidth { get; set; }
 
+    /// <summary>
+    /// The pin location for the column.
+    /// </summary>
+    [Parameter]
+    public TableColumnPinLocation? PinLocation { get; set; }
```
jattasNI, line 24: It's surprising to me that individual column types are adding the `PinLocation` property. Even though it *should* only be used on these column types, the property exists on every column type. Should we add it to the blazor table column base class instead? 

@msmithNI might know our patterns better than I do here. From a quick scan I saw a mix of approaches. Some properties like [SortIndex and GroupIndex](https://github.com/ni/nimble/blob/965106734da71d466f030ec68440aa6c7663799b/packages/blazor-workspace/NimbleBlazor/Source/TableColumn/NumberText/NimbleTableColumnNumberText.razor.cs#L30) are added on individual component classes but others like [ColumnHidden](https://github.com/ni/nimble/blob/965106734da71d466f030ec68440aa6c7663799b/packages/blazor-workspace/NimbleBlazor/Source/TableColumn/Base/NimbleTableColumn.cs#L31) are added on the base class and [bound on individual component razor templates](https://github.com/ni/nimble/blob/965106734da71d466f030ec68440aa6c7663799b/packages/blazor-workspace/NimbleBlazor/Source/TableColumn/NumberText/NimbleTableColumnNumberText.razor#L10).

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -54,6 +56,24 @@ export const styles = css`
         background-color: ${fillHoverSelectedColor};
     }
 
+    :host([selectable]:not([selected])[allow-hover]:hover) .pinned-cell-container {
```
jattasNI: I'm seeing some use of `not` in the styles which [our css guidelines advise against](https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes). From a quick scan it looks like you might simply be able to remove the `not` part of the selectors since the styles immediately after are targeting the same element without the `not`.

packages/nimble-components/src/table/components/row/template.ts
```diff
@@ -27,6 +28,42 @@ export const template = html<TableRow>`
         aria-expanded=${x => x.expanded}
         style="--ni-private-table-row-indent-level: ${x => x.nestingLevel};"
     >
+        <span class="pinned-cell-container">
+            ${repeat(x => x.columns, html<TableColumn, TableRow>`
```
jattasNI, line 64: Is this block identical to the existing table cell block around line 115, just with a different set of columns included? If so, I'm curious if @mollykreis or @rajsite have ideas about ways to reduce duplication. One idea, not sure how well this would turn out: a new "column container" component that takes a list of columns and renders the cells. Then this template would contain one instance of that component that iterates over the pinned columns and one instance that iterates over the non-pinned columns.

The "column container" might not even need to be a full blown component, it might be just a template constant similar to [labelTemplate](https://github.com/ni/nimble/blob/965106734da71d466f030ec68440aa6c7663799b/packages/nimble-components/src/radio-group/template.ts#L8).

packages/nimble-components/src/table/models/table-layout-manager.ts
```diff
@@ -362,7 +363,7 @@ export class TableLayoutManager<TData extends TableRecord> {
     }
 
     private getVisibleColumns(): TableColumn[] {
```
jattasNI: Could you rename this to something like `getUnpinnedVisibleColumns()` to make it clear that it's not all visible columns anymore?

packages/nimble-components/src/table/template.ts
```diff
@@ -49,8 +49,25 @@ export const template = html<Table>`
             --ni-private-table-row-grid-columns: ${x => (x.rowGridColumns ? x.rowGridColumns : '')};
             --ni-private-table-cursor-override: ${x => (x.layoutManager.isColumnBeingSized ? 'col-resize' : 'default')};
             --ni-private-table-scrollable-min-width: ${x => x.tableScrollableMinWidth}px;
+            --ni-private-table-pinned-columns-row-grid-columns: ${x => x.pinnedColumnsGridTemplateColumns};
             ">
             <div role="rowgroup" class="header-row-container">
+                <div class="pinned-columns-header-container">
+                    ${repeat(x => x.pinnedColumns, html<TableColumn, Table>`
```
jattasNI, line 73: If we agree on a good pattern to avoid template duplication (see other thread) we could apply it here too.

packages/nimble-components/src/table/components/group-row/template.ts
hellovolcano: fixed--a remnant sticky position was.. sticking around. (sorry not sorry)

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
hellovolcano: @jattasNI I've added both of the two test cases you listed, but haven't tackled the hover state.  Are there any other tests I can reference for that pattern to look into it more? And do you think that's required for this first push, or something we could consider adding later?

jattasNI: The closest we have for hover state are the matrix interactions tests. For example, [toggle button interactions theme matrix](https://60e89457a987cf003efc0a5b-mwtpvfukap.chromatic.com/?path=/story/tests-toggle-button--interactions-theme-matrix) and [the helper code that implements that](https://github.com/ni/nimble/blob/965106734da71d466f030ec68440aa6c7663799b/packages/storybook/src/utilities/matrix.ts#L143). 

But those are different because they simulate hover state across the entire component rather than a sub-part like an individual row. And simulating hover by applying styles to a component is somewhat unreliable vs actually hovering the mouse.

Because of those differences I am ok with not doing automated hover testing for this PR.

packages/nimble-components/src/table-column/base/index.ts
```diff
@@ -7,6 +7,9 @@ import {
 import type { TableColumnValidity } from './types';
 import type { ColumnValidator } from './models/column-validator';
 import { slotTextContent } from '../../utilities/models/slot-text-content';
+import { TableColumnPinLocation } from '../../table/types';
+
+export { TableColumnPinLocation };
```
mollykreis: You shouldn't need to export `TableColumnPinLocation` from here. It looks like it's already exported from `table\types.ts`, so you should be able to use it from there (similar to the way `TableColumnSortDirection` or `TableRowSelectionMode` is used throughout the code).

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -56,11 +60,31 @@ export const styles = css`
...
+
+    .pinned-column-spacer.has-pinned-columns {
+        position: sticky;
+        left: 0;
+        background: ${tableRowBorderColor};
+        z-index: ${ZIndexLevels.zIndex1000};
+    }
+
+    .expand-collapse-button-container.selectable {
+    }
```
mollykreis: Can this empty style be removed?

packages/nimble-components/src/table-column/base/index.ts
```diff
@@ -37,6 +40,9 @@ export abstract class TableColumn<
     @attr({ attribute: 'column-hidden', mode: 'boolean' })
     public columnHidden = false;
 
+    @attr({ attribute: 'pin-location' })
```
mollykreis: When a column gets pinned, a vertical gray line becomes visible just to the right of the pinned column. This isn't intention, is it?

Image

Note: the above screenshot was taken by setting `pinLocation = 'left'` on the breakpoint column in the "Ts Table Column: Breakpoint" storybook story.

packages/nimble-components/src/table-column/base/index.ts
```diff
@@ -37,6 +40,9 @@ export abstract class TableColumn<
...
+    public pinLocation: TableColumnPinLocation;
```
mollykreis: The focus outline for rows in the table looks like it got broken with this change:

https://github.com/user-attachments/assets/aaab6a48-c406-49f6-9754-88c10ae6db42

packages/nimble-components/src/table/components/row/template.ts
mollykreis: I'd expect you to be able to do something similar to the labelTemplate that @jattasNI referenced.

I didn't do thorough testing, but this seems to be working:
```html
const rowCellTemplate = html<TableColumn, TableRow>`
    <${tableCellTag}
        class="cell"
        :cellState="${(_, c) => c.parent.cellStates[c.index]}"
        :cellViewTemplate="${x => x.columnInternals.cellViewTemplate}"
        :column="${x => x}"
        column-id="${x => x.columnId}"
        ?has-action-menu="${x => !!x.actionMenuSlot}"
        action-menu-label="${x => x.actionMenuLabel}"
        @cell-action-menu-beforetoggle="${(x, c) => c.parent.onCellActionMenuBeforeToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>, x)}"
        @cell-action-menu-toggle="${(x, c) => c.parent.onCellActionMenuToggle(c.event as CustomEvent<MenuButtonToggleEventDetail>, x)}"
        @cell-view-slots-request="${(x, c) => c.parent.onCellViewSlotsRequest(x, c.event as CustomEvent<CellViewSlotRequestEventDetail>)}"
        :nestingLevel="${(_, c) => c.parent.cellIndentLevels[c.index]}"
    >

        ${when((x, c) => ((c.parent as TableRow).currentActionMenuColumn === x) && x.actionMenuSlot, html<TableColumn, TableRow>`
            <slot
                name="${x => `row-action-menu-${x.actionMenuSlot!}`}"
                slot="cellActionMenu"
            ></slot>
        `)}

        ${repeat(x => x.columnInternals.slotNames, html<string, TableColumn>`
            <slot
                name="${(x, c) => uniquifySlotNameForColumn(c.parent, x)}"
                slot="${(x, c) => uniquifySlotNameForColumn(c.parent, x)}"
            ></slot>
        `)}
    </${tableCellTag}>
`;

export const template = html<TableRow>`
    <!-- existing template code -->

    <span class="pinned-cell-container">
        ${repeat(x => x.columns, html<TableColumn, TableRow>`
            ${when(x => !x.columnHidden && x.pinLocation === TableColumnPinLocation.left, html<TableColumn, TableRow>`
                ${rowCellTemplate}
        `)}
        `, { recycle: false, positioning: true })}
    </span>

    <!-- more existing template code -->

    <span ${ref('cellContainer')} 
        class="cell-container ${x => (x.isNestedParent ? 'nested-parent' : '')}"
    >
        ${repeat(x => x.columns, html<TableColumn, TableRow>`
            ${when(x => !x.columnHidden && x.pinLocation !== TableColumnPinLocation.left, html<TableColumn, TableRow>`
                ${rowCellTemplate}
            `)}
        `, { recycle: false, positioning: true })}
    </span>
```

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -152,9 +155,19 @@ export class TableValidator<TData extends TableRecord> {
         this.invalidColumnConfiguration = columns.some(
             x => !x.columnInternals.validator.isColumnValid
         );
+
```
mollykreis: nit: can you revert this whitespace change?


```suggestion
```

packages/nimble-components/src/table/index.ts
```diff
@@ -207,6 +208,51 @@ export class Table<
...
+                minPixelWidth
+            } = column.columnInternals;
+            const resolvedPixelWidth = this.getPinnedColumnResolvedPixelWidth(
+                column
+            );
+            if (resolvedPixelWidth !== undefined) {
+                const coercedPixelWidth = Math.max(
+                    minPixelWidth,
+                    resolvedPixelWidth
+                );
```
mollykreis: nit: I think it would be good to write this consistently with line 223 above rather than assigning to `minPixelWidth` locally, which is only used in this one place.


```suggestion
            const resolvedPixelWidth = this.getPinnedColumnResolvedPixelWidth(
                column
            );
            if (resolvedPixelWidth !== undefined) {
                const coercedPixelWidth = Math.max(
                    column.columnInternals.minPixelWidth,
                    resolvedPixelWidth
                );
```

packages/nimble-components/src/table/index.ts
```diff
@@ -699,7 +751,10 @@ export class Table<
         if (this.tableUpdateTracker.updateColumnWidths) {
             this.rowGridColumns = this.layoutManager.getGridTemplateColumns();
             this.visibleColumns = this.columns.filter(
-                column => !column.columnHidden
+                column => !column.columnHidden && column.pinLocation !== TableColumnPinLocation.left
+            );
+            this.pinnedColumns = this.columns.filter(
+                column => column.pinLocation === TableColumnPinLocation.left && !column.columnHidden
             );
```
mollykreis, line 757: Can you make the order of the conditions in the filter consistent between these two lines? For example:

```ts
            this.visibleColumns = this.columns.filter(
                column => !column.columnHidden && column.pinLocation !== TableColumnPinLocation.left
            );
            this.pinnedColumns = this.columns.filter(
                column => !column.columnHidden && column.pinLocation === TableColumnPinLocation.left 
            );
```

mollykreis (changes requested):

packages/blazor-workspace/NimbleBlazor/Source/TableColumn/Mapping/NimbleTableColumnMapping.razor.cs
mollykreis: I think this brings up a bigger question than just the Blazor component implementation. We should consider whether or not we want to put `pin-location` on the base table column class in nimble-components. Most properties/attributes on table columns get added to specific columns (e.g. `group-index` and `sort-index`) through mixins (e.g. `groupable-column.ts` contains a `mixinGroupableColumnAPI()` function) to make it easy for columns to opt in to grouping without making grouping available on every column.

My thought is that we should follow that pattern to create a "pinnableColumn" mixin that we use wherever it makes sense rather than adding `pin-location` to every column.

packages/nimble-components/src/table-column/base/index.ts
hellovolcano: that was intentional to provide a visual separation. I gut checked that with one of the designers

packages/nimble-components/src/table/components/row/template.ts
hellovolcano: tested Molly's suggestion and applied

packages/nimble-components/src/table/template.ts
hellovolcano: applied similar pattern as rowCellTemplate

packages/blazor-workspace/NimbleBlazor/Source/TableColumn/Mapping/NimbleTableColumnMapping.razor.cs
hellovolcano: I added a mixinPinnableColumnAPI.

packages/nimble-components/src/table-column/mixins/pinnable-column.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type TableColumnWithPinLocation = TableColumn & {
+    pinLocation?: TableColumnPinLocation
+};
+
+export function getColumnPinLocation(
+    column: TableColumn
+): TableColumnPinLocation {
+    return (column as TableColumnWithPinLocation).pinLocation;
+}
```
mollykreis: I think this function is an indication that something is wrong with the way the pinLocation is being added to the column API.

Because the table itself needs access to the pinLocation and it isn't being added directly to the base `TableColumn` class, we should add it to `ColumnInternals` --  following the same pattern as grouping and sorting.

packages/nimble-components/src/table/template.ts
```diff
@@ -49,8 +68,14 @@ export const template = html<Table>`
...
+                        ${tableHeaderTemplate(false)}
```
mollykreis: Why do the headers of pinned columns not include grouping indicators? The only time I would've expected the grouping indicator to be hidden is when a column has specified that all indicators should be hidden (configured via `columnInternals.hideHeaderIndicators`).

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
```diff
@@ -0,0 +1,256 @@
...
+                    await waitForUpdatesAsync();
+                }
+            }
+        })
+    );
+};
+
+export const pinnedColumnsNoSelectionThemeMatrix: StoryFn = createMatrixThemeStory(
+    createMatrix(component, [
+        pinLocationStates,
```
mollykreis: Should all the "pinnedColumns__SelectionThememMatrix" stories specify `pinnedOnlyStates`? The two below do, but this one uses all `pinLocationStates`. Not sure if that was intentional or not.

packages/nimble-components/src/table/template.ts
hellovolcano: that was an oversight on my part. updated.

packages/nimble-components/src/table/template.ts
```diff
@@ -32,6 +33,24 @@ import {
     tableSelectAllLabel
 } from '../label-provider/table/label-tokens';
 
+const tableHeaderTemplate = (includeGrouping: boolean): ViewTemplate<TableColumn, Table> => html<TableColumn, Table>`
```
mollykreis: Since both callers of this function pass `true` for `includeGrouping` now, the behavior doesn't need to be configurable.

packages/nimble-components/src/table/models/table-update-tracker.ts
```diff
@@ -7,7 +7,7 @@ import { isColumnInternalsProperty } from '../../table-column/base/models/column
 
 const isColumnProperty = (
     changedProperty: string,
-    ...args: (keyof TableColumn)[]
+    ...args: (keyof TableColumn | 'pinLocation')[]
```
mollykreis: I think some of the changes to the TableUpdateTracker (such as this line) need to be revisited to account for 'pinLocation' moving to 'ColumnInternals'.

packages/nimble-components/src/table/template.ts
```diff
@@ -33,7 +33,7 @@ import {
     tableSelectAllLabel
 } from '../label-provider/table/label-tokens';
 
-const tableHeaderTemplate = (includeGrouping: boolean): ViewTemplate<TableColumn, Table> => html<TableColumn, Table>`
+const tableHeaderTemplate = (): ViewTemplate<TableColumn, Table> => html<TableColumn, Table>`
```
mollykreis: This doesn't need to be a function, right?

mollykreis (approved):

rajsite:
@hellovolcano I wasn't able to complete a review today, I do expect to review Monday

msmithNI (approved):

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -56,11 +60,32 @@ export const styles = css`
         background-color: ${fillHoverColor};
     }
 
+    :host([allow-hover]:hover) .pinned-column-spacer.has-pinned-columns,
+    :host([allow-hover]:hover) .checkbox-container.has-pinned-columns {
```
rajsite: Reminder to carefully follow: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

Emphasizing the following note:
> Note that the groups of selectors are organized by target. For example the target of the selector :host[disabled] .end {} is .end and not :host so it is grouped with the other .end selectors.

Here you have a single large selector targeting two different elements `pinned-column-spacer` and `checkbox-container`. Make sure to split these by the targeted element so that checkbox-container styles are together, etc.

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -56,11 +60,32 @@ export const styles = css`
...
+        background: linear-gradient(${fillHoverColor}, ${fillHoverColor}),
+            ${tableRowBorderColor};
+    }
+
     :host(${focusVisible}) {
```
rajsite, line 63: Note how this selector targets :host but is now split away from the other host targetted styles. follow dom order styles doc

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -56,11 +60,32 @@ export const styles = css`
...
         outline-offset: calc(-2 * ${borderWidth});
     }
 
+    :host(${focusVisible}) .pinned-column-spacer.has-pinned-columns {
+        box-shadow: inset 2px -2px 0 ${borderHoverColor};
+    }
+
+    .pinned-column-spacer {
+        height: 100%;
+    }
```
rajsite, line 88: Note the order of these. the default style target of just `pinned-column-spacer` should be before the overrides `:host(${focusVisible}) .pinned-column-spacer.has-pinned-columns`.

Maybe a useful point to clarify. The selector target is in the right most part of the selector. Selectors read right to left. i.e. "all pinned-column-spacer" vs "all pinned-column-spacer with class has-pinned-columns with parent host(:focus-visible)". The generic selector should be first followed by state specific overrides, see: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -104,6 +133,28 @@ export const styles = css`
...
+            }
+
+            .pinned-column-spacer.has-pinned-columns,
+            .checkbox-container.has-pinned-columns {
+                background: linear-gradient(
+                        ${hexToRgbaCssColor(White, 0.1)},
+                        ${hexToRgbaCssColor(White, 0.1)}
+                    ),
+                    ${tableRowBorderColor};
+            }
```
rajsite: Follow the css dom order rules for these as well. To redduce duplication do not group selectors that target different elements together. Instead make private variables for duplicated properties. Consider a well named private variable for one or both of the linear gradient definitions.

See: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#prefer-cascade-for-state-changes

emphasis: 
> Each selector should represent a single, clear state of the control and define a minimal set of properties that override the base configuration of the control. Duplicated property values are avoided by using well-named design tokens or private css custom properties.

See private css custom properties: https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#css-custom-properties

packages/nimble-components/src/table/components/group-row/template.ts
```diff
@@ -15,10 +15,15 @@ export const template = html<TableGroupRow>`
...
+        style="
+            --ni-private-table-group-row-indent-level: ${x => x.nestingLevel};
+            --ni-private-table-group-row-pinned-column-offset: ${x => x.pinnedColumnOffset}px;
+        "
     >
+        <span class="pinned-column-spacer ${x => (x.pinnedColumnOffset > 0 ? 'has-pinned-columns' : '')}"></span>
+
         ${when(x => x.selectable, html<TableGroupRow>`
-            <span role="gridcell" class="checkbox-container">
+            <span role="gridcell" class="checkbox-container ${x => (x.pinnedColumnOffset > 0 ? 'has-pinned-columns' : '')}">
```
rajsite: Instead of placing this class on multiple children of group-row can we instead make `has-pinned-column` an internal attribute on the group-row (similar to selectable, etc)?
then styles can leverage it with, i.e. `:host([has-pinned-columns]) .pinned-column-spacer`
We don't need to duplicate the state across all the elements that are owned by the group row.

packages/nimble-components/src/table/components/group-row/template.ts
```diff
@@ -31,7 +36,7 @@ export const template = html<TableGroupRow>`
             </span>
         `)}
 
-        <span role="gridcell">
+        <span role="gridcell" class="expand-collapse-button-container ${x => (x.selectable ? 'selectable' : '')}">
```
rajsite: Similar to above, `selectable` is already sate on the group row. I don't think we need this extra attribute. Can just target the host: `:host([selectable])`
Edit: Am I missing it, this class `expand-collapse-button-container` doesn't seem to be used anywhere? is it still needed?

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -54,6 +56,24 @@ export const styles = css`
...
+            ${applicationBackgroundColor};
+    }
+
+    :host([selected][allow-hover]:hover) .pinned-cell-container {
+        background: linear-gradient(
+                ${fillHoverSelectedColor},
+                ${fillHoverSelectedColor}
+            ),
+            ${applicationBackgroundColor};
+    }
```
rajsite: All the same discussion from above applies to this file. Make sure styles are placed in DOM order by the selector target. This chunk of selectors was plopped in the middle of selectors that target :host
https://github.com/ni/nimble/blob/main/packages/nimble-components/docs/css-guidelines.md#group-selectors-by-target-and-document-order

**I'm not calling it out further instances in each style file, please review all the styles updated in the PR for consistency**

packages/nimble-components/src/table/models/table-validator.ts
```diff
@@ -155,6 +158,16 @@ export class TableValidator<TData extends TableRecord> {
         return !this.invalidColumnConfiguration;
     }
 
+    public validatePinnedColumnConfigurations(
+        columns: readonly TableColumn[]
+    ): boolean {
+        this.invalidPinnedColumnConfiguration = columns.some(
+            x => x.columnInternals.pinLocation === TableColumnPinLocation.left
+                && x.columnInternals.currentPixelWidth === undefined
```
rajsite: @mollykreis should we be checking resizingDisabled instead of currentPixelWidth?
currentPixelWidth is internal and owned by the layout manager and seems like the wrong value for validation where instead we want something directly corresponding to user config. 
seems like resizingDisabled is the signal we are looking for right now.
That would mean that only mapping and breakpoint columns participate today but maybe that makes sense for now? Edit: see also comments below where lack of api to express fixed width intent, tests, and documentation coverage for text columns makes me think we should not support them with the current PR

packages/nimble-components/src/table/index.ts
```diff
@@ -821,6 +875,16 @@ export class Table<
...
     }
 
+    private getPinnedColumnResolvedPixelWidth(
+        column: TableColumn
+    ): number | undefined {
+        const {
+            currentPixelWidth,
+            pixelWidth
+        } = column.columnInternals;
+        return currentPixelWidth ?? pixelWidth;
```
rajsite: I think I find this check surprising. What is the scenario where currentPixelWidth is undefined while pixelWidth is defined?
Any assignment to pixelWidth should already result in an update to currentPixelWidth: https://github.com/ni/nimble/blob/7e82e7453b8cba71110fb6474cc11ea9921eb54c/packages/nimble-components/src/table-column/base/models/column-internals.ts#L230-L233

packages/nimble-components/src/table/tests/table.spec.ts
```diff
@@ -859,6 +860,88 @@ describe('Table', () => {
             });
         });
 
+        describe('pinned columns', () => {
```
rajsite: I think we need some more tests around a column being promoted from unpinned to pinned and tests for multiple pinned columns (consider updates such as 1->2 pinned columns, 2->1, change of order, for mapping, change of "icon-size" directly corresponding to resizingDisabled). edit: based on discussion further down I'd consider moving the tests to the mapping column / a standalone test file that can be updated to support multiple column types.

packages/nimble-components/src/table-column/text-base/index.ts
```diff
@@ -38,7 +39,9 @@ export function mixinTextBase<
 >(base: TBase) {
     return mixinGroupableColumnAPI(
         mixinFractionalWidthColumnAPI(
-            mixinColumnWithPlaceholderAPI(mixinSortableColumnAPI(base))
+            mixinPinnableColumnAPI(
```
rajsite: I don't think we should update text-base to be pinnable. This touches lots of column types and there are no js tests / chromatic tests / storybook docs. And based on the other comment, the validator signal we choose would not apply to text columns either. We also don't have a good way to express non-resizeable columns for text columns right now and I don't think the current check used is correct (see above)

packages/ok-components/src/ts/table-column/breakpoint/index.ts
```diff
@@ -33,7 +34,9 @@ declare global {
 /**
  * A table column that displays a breakpoint indicator with toggle functionality.
  */
-export class TsTableColumnBreakpoint extends TableColumn<TsTableColumnBreakpointColumnConfig> {
+export class TsTableColumnBreakpoint extends mixinPinnableColumnAPI(
```
rajsite, line 37: based on the library requirements doc: https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#choosing-a-component-library

I'm going to strongly suggest that storybook api docs are updated for the breakpoint column to include pin-location and weakly / non-blocking suggest having some matrix tests / js tests

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
```diff
@@ -902,3 +952,61 @@ export const width: StoryObj<ColumnWidthTableArgs> = {
         }
     }
 };
+
+const pinnedColumnsDescription = 'Set `pin-location` to `left` on one or more fixed-width columns to pin them to the left side of the table. Pinned columns remain visible as the user scrolls the table horizontally. Only columns with a fixed pixel width support pinning.';
```
rajsite: The intention of the Table Column Configuration story is to document properties that apply to all columns. pin-location prior to review did not apply to menu button columns and review currently suggests not applying to any of the text columns either. I think it's more appropriate for individual column stories that support the feature to document it and have a toggle to conditionally pin a column

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
```diff
@@ -0,0 +1,256 @@
...
+type SelectionModeState = (typeof selectionModeStates)[number];
+
+const pinnedOnlyStates = [pinLocationStates[1]] as const;
+
+// Component that renders a table with the specified configurations
+const component = (
+    [pinLocationName, pinLocation]: PinLocationState,
+    [groupedStateName, groupedState]: GroupedState,
+    [hierarchyStateName, hierarchyState]: HierarchyState,
+    selectionMode: SelectionModeState,
```
rajsite, line 151: Note: I updated the stories to have a dimension for the the story being scrolled off the page

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
```diff
@@ -0,0 +1,256 @@
...
+    TableRowSelectionMode.single,
+    TableRowSelectionMode.multiple
+] as const;
+type SelectionModeState = (typeof selectionModeStates)[number];
+
+const pinnedOnlyStates = [pinLocationStates[1]] as const;
+
+// Component that renders a table with the specified configurations
+const component = (
+    [pinLocationName, pinLocation]: PinLocationState,
```
rajsite: We should have multiple pinned columns as a possible configuration. Instead of adding an additional dimension I'd recommend making for example: pinConfigurationStates: None pinned, Single left pinned, Multi left pinned

rajsite (changes requested):

packages/nimble-components/src/table-column/text-base/index.ts
mollykreis: I agree. I didn't pay close enough attention to which file this change was in. I think the only nimble column we want to support pinning on is the mapping column.

packages/nimble-components/src/table/models/table-validator.ts
mollykreis: With the current implementation, there are two things that must be true about pinned columns to not break assumptions in the layout manager:
1. They aren't resizable
2. They can't be sized via fractional width

Therefore, it is more than not being resizable (unless we want to revisit item 2 above).

Perhaps, we should explicitly check for both of those conditions using `columnInternals.pixelWidth` and `columnInternals.resizingDisabled`.

packages/ok-components/src/ts/table-column/breakpoint/index.ts
hellovolcano: I have to make some changes the breakpoint column. I will add more tests as part of that PR. for this one I did expose the pin-location in the storybook.

packages/storybook/src/nimble/table-column/base/table-column.stories.ts
hellovolcano: ah, I originally had it just on the mapping column but changed after initial feedback. makes sense it should be there given the direction this has gone.

packages/nimble-components/src/table-column/text-base/index.ts
hellovolcano: removed

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -115,6 +117,46 @@ export const styles = css`
...
+
+    :host(${focusVisible}) .pinned-cell-container {
+        box-shadow:
+            inset calc(2 * ${borderWidth}) 0 0 ${borderHoverColor},
+            inset 0 calc(2 * ${borderWidth}) 0 ${borderHoverColor},
+            inset 0 calc(-2 * ${borderWidth}) 0 ${borderHoverColor},
+            inset -2px 0 0 0 ${tableRowBorderColor};
+    }
+
+    :host([menu-open]) .pinned-cell-container {
```
hellovolcano, line 156: @rajsite I added this so I could use it for the breakpoint column to ensure its menu so I can ensure its menu renders above the other pinned columns. I tried to keep it generic/use something that was already there, but let me know if you want to talk about it/figure out a better solution

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
hellovolcano: added multiple columns and cleaned up the configuration a bit

packages/nimble-components/src/table/components/row/styles.ts
rajsite: Can you share a commit / draft pr that demonstrates the intended usage I can reference?

hellovolcano: @rajsite #2990 has the intended usage in the ok breakpoint column.

packages/nimble-components/src/table/tests/table.pinned-column.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+        expect(element.checkValidity()).toBeFalse();
+        expect(element.validity.invalidPinnedColumnConfiguration).toBeTrue();
+
+        column1.columnInternals.pixelWidth = 120;
+        column1.columnInternals.resizingDisabled = true;
+        await waitForUpdatesAsync();
+
+        expect(element.checkValidity()).toBeTrue();
+        expect(element.validity.invalidPinnedColumnConfiguration).toBeFalse();
+    });
```
rajsite: for these three we don't want to exercise setting pixelWidth directly as that is an internal api. In this case we would be leveraging the mapping column's api so setting width-mode (which indirectly sets pixelWidth).

we can assert the mapping column is behaving as expected by asserting on columnInternals.pixelWidth (in case mapping column changes behavior in the future) but we should not exercise manipulating a private api is achieving the validation result we want.

the ideal scenario is that we have a custom Test column for the core table feature that exercises the pinned column feature standalone of other column types and that the mapping column has it's own separate tests to exercise it's participation as a column opting in to the mapping feature. We don't need to make that change, but something to consider in the future.

packages/storybook/src/nimble/table-column/mapping/table-column-mapping.stories.ts
```diff
@@ -73,6 +74,7 @@ interface MappingColumnTableArgs extends SharedTableArgs {
...
     keyType: string;
     widthMode: keyof typeof TableColumnMappingWidthMode;
+    pinLocationEnabled: boolean;
```
rajsite: we should use the enum values using width-mode as an example, this configuration is misleading as it makes it look like a boolean when the value is type string:

Image

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
```diff
@@ -234,9 +248,9 @@ const playFunction = async (): Promise<void> => {
 
 export const pinnedColumns$NoSelectionThemeMatrix: StoryFn = createMatrixThemeStory(
     createMatrix(component, [
-        pinnedOnlyStates,
+        pinConfigurationStates,
         groupedStates,
-        hierarchyStates,
+        [hierarchyStates[0]],
```
rajsite: when we split dimensions we don't want to lose coverage, have an idea how to improve it without causing more manual dimension splits, i'll push that change

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -61,6 +65,39 @@ export const styles = css`
...
+    .pinned-column-spacer {
+        height: 100%;
+    }
+
+    :host([has-pinned-columns]) .pinned-column-spacer {
+        position: sticky;
+        left: 0;
+        background: ${tableRowBorderColor};
+        z-index: ${ZIndexLevels.zIndex1};
+    }
```
rajsite: Lets remove the pinned column spacer from the display tree when not needed

```suggestion
    .pinned-column-spacer {
        display: none;
        height: 100%;
    }

    :host([has-pinned-columns]) .pinned-column-spacer {
        display: block;
        position: sticky;
        left: 0;
        background: ${tableRowBorderColor};
        z-index: ${ZIndexLevels.zIndex1};
    }
```

rajsite (approved):

packages/nimble-components/src/table/components/row/styles.ts
```diff
@@ -115,6 +117,46 @@ export const styles = css`
...
+    }
+
+    :host(${focusVisible}) .pinned-cell-container {
+        box-shadow:
+            inset calc(2 * ${borderWidth}) 0 0 ${borderHoverColor},
+            inset 0 calc(2 * ${borderWidth}) 0 ${borderHoverColor},
+            inset 0 calc(-2 * ${borderWidth}) 0 ${borderHoverColor},
+            inset -2px 0 0 0 ${tableRowBorderColor};
+    }
+
```
rajsite:
```suggestion
    ${'' /* Pushing the pinned-cell-container to a higher z-index for breakpoint menu behavior (not required by table directly) */}
```

packages/nimble-components/src/table/tests/table.pinned-column.spec.ts
rajsite: can track as a tech debt to decouple the table core test from a specific column

change/@ni-nimble-angular-3b488452-3822-410a-a684-15d5f131dd98.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: noticed that the breakpoint column does not work with the pinned column changes in this RP but that will be addressed in the followup https://github.com/ni/nimble/pull/2990

rajsite (approved):

packages/storybook/src/nimble/table/table-pinned-columns.matrix.stories.ts
rajsite: fyi @hellovolcano pushed a change to use grid layout for the matrix tests 👍

change/@ni-nimble-angular-3b488452-3822-410a-a684-15d5f131dd98.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "add support for pinning table columns",
```
rajsite, line 3: Let me know when ready and will do a quick double check of the changes before merge

packages/nimble-components/src/table/tests/table.pinned-column.spec.ts
hellovolcano: tracked in #2991

packages/nimble-components/src/table/components/group-row/styles.ts
```diff
@@ -61,6 +65,37 @@ export const styles = css`
         outline-offset: calc(-2 * ${borderWidth});
     }
```
hellovolcano, line 67: @rajsite when I added display: none, it to the standalone .pinned-column-spacer, it was causing test failures, I think because of [this](https://github.com/ni/nimble/pull/2964/changes#diff-9ae2525eb1b62079e236cdb03ac1f63f9ec06b50fee05e1efb3eb84e86420a4aR31).

I moved the height declaration to the style below and just removed this one. let me know your thoughts on that approach

change/@ni-nimble-angular-3b488452-3822-410a-a684-15d5f131dd98.json
hellovolcano: @rajsite -- just left one comment for you on the group row style.

packages/nimble-components/src/table/components/group-row/styles.ts
rajsite: ah, guess sit's needed for the grid layout? that's fine 👍

change/@ni-nimble-angular-3b488452-3822-410a-a684-15d5f131dd98.json
hellovolcano: @rajsite this is ready now

rajsite:
bypassing Jesse OOO

PR #2975, opened 2026-06-05
Add scrolling behavior to chat

Alexia-Claudia-Micu:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/3897966

## 👩‍💻 Implementation

Adds auto-scroll as an option to the Spright chat-conversation so the view follows the conversation as messages arrive. Has the following behavior:
- Auto-scroll from the bottom: while the user is at the bottom, the conversation follows newly added/streamed content; it disengages when the user scrolls away and re-engages when they return.
- Anchor-on-insert: when a new turn's anchor message is inserted, it's pinned to the top of the viewport (messages before it move to a history slot, and the anchored region reserves a viewport via min-height: 100% so the response streams in below it).
- New AutoScrollManager encapsulates the slot partitioning, anchoring, and scroll-follow logic.

## 🧪 Testing

Added tests and manual testing in the example app.

## ✅ Checklist


packages/storybook/src/spright/chat/conversation/chat-conversation.stories.ts
```diff
@@ -155,6 +156,12 @@ export const chatConversation: StoryObj<ChatConversationArgs> = {
             description: 'The appearance of the chat conversation.',
             table: { category: apiCategory.attributes }
         },
+        autoScroll: {
```
jattasNI, line 155: FYI I pushed some trivial changes to this documentation and to mention the auto-scroll attribute in the chat design spec.

change/@ni-spright-components-98a40075-e551-4376-aa9a-f3a3fb1d4c4a.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I looked into the Chromatic diffs and they all seem to be caused by the conversation now loading scrolled to the end rather than the beginning. I think that behavior makes sense so I'll approve the diffs. I also pushed a change to add a note about that behavior to the scrolling-behavior.md spec.

packages/angular-workspace/example-client-app/src/app/customapp/chat-conversation-section.component.ts
```diff
@@ -1,61 +1,193 @@
-import { Component } from '@angular/core';
+import { Component, type OnDestroy } from '@angular/core';
 import type { ChatInputSendEventDetail } from '@ni/spright-angular/chat/input';
 
+interface ChatEntry {
```
jattasNI, line 4: This new example is really helpful. To keep up with our goal of having equivalent functionality in all 3 example apps I asked copilot to make equivalent changes to Blazor and React and pushed those to the branch.

packages/spright-components/src/chat/conversation/index.ts
```diff
@@ -17,6 +18,9 @@ export class ChatConversation extends FoundationElement {
     @attr
     public appearance = ChatConversationAppearance.default;
 
+    @attr({ attribute: 'auto-scroll', mode: 'boolean' })
+    public autoScroll = true;
```
jattasNI: Boolean attributes should always default to false to since that aligns the behavior when the attribute isn't present (which is treated as false). This works well for this one because it will leave existing clients with unchanged behavior. I pushed a change to fix this.

packages/angular-workspace/spright-angular/chat/conversation/spright-chat-conversation.directive.ts
```diff
@@ -11,4 +11,14 @@ export { chatConversationTag };
     selector: 'spright-chat-conversation',
     standalone: false
 })
-export class SprightChatConversationDirective { }
+export class SprightChatConversationDirective {
+    public get autoScroll(): boolean | undefined {
+        return this.elementRef.nativeElement.autoScroll;
+    }
+
+    @Input('auto-scroll') public set autoScroll(value: boolean | undefined) {
```
jattasNI: We have a helper to handle boolean attributes. I pushed a change to use it and also add some directive tests.

packages/spright-components/src/chat/conversation/index.ts
```diff
@@ -49,6 +53,30 @@ export class ChatConversation extends FoundationElement {
     @observable
     public readonly slottedEndElements?: HTMLElement[];
 
+    /** @internal */
+    public messagesContainer: HTMLElement | null = null;
+    private scrollManager: ChatConversationScrollManager | null = null;
+
+    public override connectedCallback(): void {
+        super.connectedCallback();
+        const defaultSlot = this.shadowRoot?.querySelector('slot:not([name])') as HTMLSlotElement | null;
```
jattasNI: Instead of querying dynamically we should use a `ref` in the template to refer to it. I pushed a change to do this.

packages/spright-components/src/chat/conversation/index.ts
```diff
@@ -49,6 +53,30 @@ export class ChatConversation extends FoundationElement {
...
+        if (this.messagesContainer && defaultSlot) {
+            this.scrollManager = new ChatConversationScrollManager(
+                this.messagesContainer,
+                this,
+                defaultSlot,
+                () => this.autoScroll
```
jattasNI: It's better for the conversation component to tell the scroll manager when its autoscroll state changes. In fact, we don't even need a scroll manager if autoscroll is false. I pushed some changes to do this.

jattasNI (commented):
I pushed some clean ups and left comments describing them. I still have some more to review.

change/@ni-spright-blazor-08d16ce6-2034-41f4-9374-2a2fed2ee178.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: I pushed a change to add a Blazor wrapper for the new conversation attribute.

packages/spright-components/src/chat/conversation/scroll-manager.ts
```diff
@@ -0,0 +1,224 @@
+import { mediumPadding } from '@ni/nimble-components/dist/esm/theme-provider/design-tokens';
```
jattasNI: @Alexia-Claudia-Micu After the changes I pushed, the biggest remaining open issue I have is about the architecture of the scroll manager class. There are a few things I would like to improve about it. I ran out of time today to try these refactors myself / with copilot so I'm not sure how achievable they are.

1. we should look at using [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to detect when scrolling is necessary
2. we should try to find a cleaner API contract between the scroll manager and the conversation component. Instead of the scroll manager gaining access to the conversation's internal elements, ideally the conversation would expose an API that the scroll manager could use. This would contain things like:
    - fire events when new messages are added or the user scrolls
    - properties to get the sizes of certain children
    - methods to set the scroll offset and the bottom padding
3. the unit tests seem to be mocking a lot of the functionality of the conversation which makes it harder to ensure that the actual conversation is behaving as we want it to. It would be good to explore whether we can write reliable tests that use the real conversation. Alternatively, if we do 2 above then we could write tests for the new conversation API and also write unit tests for the scroll manager which mock out the conversation via a clean contract

I recognize that this is a big refactor so I'm happy to have a discussion about ROI or about me/others helping do it if you find that you + copilot can't get anywhere with it.

packages/blazor-workspace/Tests/SprightBlazor.Tests/Unit/Components/SprightChatConversationTests.cs
```diff
@@ -41,4 +43,19 @@ public void SprightChatConversation_WithToolbar_RendersContent()
         Assert.Contains("slot=\"toolbar\"", component.Markup);
         Assert.Contains("Toolbar Button", component.Markup);
     }
+
+    [Fact]
+    public void SprightChatConversation_AutoScroll_AttributeIsSet()
```
jattasNI, line 39: Heads up that #2980 may require some changes to these tests. That introduces some new helpers that do more robust DOM matching rather than the string matching we had been using before.

Alexia-Claudia-Micu: Fixed merge conflicts for this.

packages/spright-components/src/chat/conversation/scroll-manager.ts
Alexia-Claudia-Micu: Implemented suggestions

jattasNI (commented):
I ran out of time to review this today and am going OOO. @rajsite will take over.

rajsite:
@Alexia-Claudia-Micu unfortunately I did not get a chance to review this today but I expect to be able to take a look tomorrow

rajsite:
@Alexia-Claudia-Micu sorry again, I was not able to finish reviewing today. I do expect to be able to review Monday.

rajsite:
@Alexia-Claudia-Micu very sorry but (hopefully) last time should be able to review tomorrow 🙏

rajsite:
@Alexia-Claudia-Micu I was able to start reviewing today and have a draft PR that demonstrates some proposed changes to this branch (note the PR targets this branch): https://github.com/ni/nimble/pull/2989

I have not finished the changes yet and plan to tomorrow. The PR tries to leverage some more patterns we have used elsewhere:

- Similar to the table keyboard manager creating a separate object that holds observable state that the template can bind to and that can plugin to the element lifecycle (conenct, disconnect, observable property changes, etc)
- Use an internals concept like column internals / step internals to communicate / store state from the parent to children
- Observe sizing by placing an element around content that grows with the content (messages) and watch the resize observer (this is similar to tabs track sizing)
- Also did a mechanical refactor to add page objects for the tests.

I still plan to finish the changes but have a couple questions:
- did y'all decide explicitly one way or another if auto-scroll should also function for the legacy "chat-message" element, i.e. `spright-chat-message[message-type="outbound"]`?
- if you don't mind, could you explore the storybook of the proposed branch and see if it preserves the expected behavior?

Alexia-Claudia-Micu:
> * did y'all decide explicitly one way or another if auto-scroll should also function for the legacy "chat-message" element, i.e. `spright-chat-message[message-type="outbound"]`?
 
@rajsite To my knowledge, there was no decision taken for this. I've checked and could not find any usage of the legacy element in systemLink Nigel.

> * if you don't mind, could you explore the storybook of the proposed branch and see if it preserves the expected behavior?

I looked over the draft and the behavior looks good to me. I left a comment with a few minor things I noticed.

hellovolcano:
> > * did y'all decide explicitly one way or another if auto-scroll should also function for the legacy "chat-message" element, i.e. `spright-chat-message[message-type="outbound"]`?
> 
> @rajsite To my knowledge, there was no decision taken for this. I've checked and could not find any usage of the legacy element in systemLink Nigel.

@rajsite @Alexia-Claudia-Micu fyi our desktop nigel app also does not use the legacy message types.

rajsite:
@Alexia-Claudia-Micu I came up with a pattern in https://github.com/ni/nimble/pull/2989 that I think addressed the two behaviors you pointed out but overall simplifies a lot of the behavior.

Conceptually the idea is that I updated the conversation to have two slots, a message "history" slot and the default slot (for "anchored" messages). When a new message is added we determine if there is an anchored message (the most recent outbound message) and filter all the messages into one of the two slots: everything before the anchored message into history and everything including and after the anchored message (i.e. new inbound messages) in the anchored slot. Then all we need to do for the reserved space is set the size of the container for the anchored message slot to be min-height: 100%.

The benefit is that as content is streamed in, etc or if any messages heights are changing we aren't needing to perform any sizing or layout. if the anchored messages content grows css just sizes things for us. The work that remains is behavior for the scroll position. Got the llm to create a summary image of the design:

auto-scroll-architecture

I think this helps reduce some edge cases by relying on CSS layout more and will be a little more understandable for future maintainers. Could you give it a look to see if I'm missing any behaviors in the new proposal and exercise the storybook/ examples for issues? IF it looks good I can do some finishing touchups and tests and merge tomorrow.

Alexia-Claudia-Micu:
@rajsite Thank you, the new pattern looks good, and it solves the problems I mentioned. I got to test it out and left a comment for a small issue I noticed (the first message doesn't auto-scroll)

change/@ni-spright-angular-2188570c-9255-47e7-8f52-a34d062b7b81.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: @Alexia-Claudia-Micu I merged the updated implementation proposal and I think the last issues were addressed. If you can do another sanity check then I can push some test and example app tweaks and merge tomorrow

change/@ni-spright-angular-2188570c-9255-47e7-8f52-a34d062b7b81.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "minor",
```
rajsite, line 2: Note to self: enable scrollbar gutters

change/@ni-spright-angular-2188570c-9255-47e7-8f52-a34d062b7b81.json
Alexia-Claudia-Micu: I tested it out and it looks all good.

change/@ni-spright-angular-2188570c-9255-47e7-8f52-a34d062b7b81.json
rajsite: didn't do now to avoid scope increase

rajsite (approved):

rajsite:
@Alexia-Claudia-Micu re-reviewing the new pattern, a potential issue could be that it relies on saving references to the previous messages / the most recent anchor on insert message to use a signal if it should scroll or not. If one of the template systems is recycling elements then holding onto references is an incorrect assumption and might have unreliable behavior. I'm not sure if it will show up in practice but is something to be aware of.

PR #2976, opened 2026-06-05
Update Blazor packages to dual .NET 10 and .NET 8

rajsite:
# Pull Request

## 🤨 Rationale

- Update Blazor libraries to support .NET 10 in addition to .NET 8 (NimbleBlazor / SprightBlazor / OkBlazor).
- Standalone Blazor WebAssembly apps such as the Nimble Blazor example app may need to enable the following features:
  - Appears to be required (at least the Nimble Blazor example app seems to require it to be enabled to run): [Fingerprint client-side static assets in standalone Blazor WebAssembly apps](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-10.0#fingerprint-client-side-static-assets-in-standalone-blazor-webassembly-apps)
  - Optional but worth enabling if you have already enabled fingerprinting: [Preloaded Blazor framework static assets](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-10.0#preloaded-blazor-framework-static-assets)

## 👩‍💻 Implementation

- Update to a dotnet 10 sdk (10.0.300)
- Enable multiple `TargetFrameworks` for build / test / package deployment to support .net 10 and 8

## 🧪 Testing

- Verify in CI build logs that dotnet 10 and 8 are running
- Verified nuget package locally in Config App
- Verified blazor web app in storybook is still functional

## ✅ Checklist


.github/workflows/main.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       - uses: actions/setup-dotnet@v5
         with:
           global-json-file: packages/blazor-workspace/global.json
+      - uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '8.0.x'
```
rajsite: Need to pin this to a specific version otherwise we probably will get build issues on next release when this resolves to a new version

change/@ni-nimble-blazor-00ba2550-5ec2-4e46-bc64-083acdc72b44.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add .NET 10 support while retaining .NET 8 (multi-targeting net8.0;net10.0)",
```
rajsite: I'll add the Blazor Standalone client fingerprint note to this, but won't mark it as a breaking change since it's not a breaking change in Nimble (it's possible something in Nimble fights with the fingerprinting but I haven't figured it out).

packages/blazor-workspace/Examples/Demo.Server/packages.lock.json
```diff
@@ -1,18 +1,18 @@
 {
```
rajsite, line 1: I'll rebuild the lock files JIC there is an issue

packages/blazor-workspace/NimbleBlazor/NimbleBlazor.csproj
```diff
@@ -55,7 +62,7 @@
 
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="[23.0.0]" />
-    <PackageReference Include="Microsoft.AspNetCore.Components.Web" Version="8.0.27" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.Web" Version="$(AspNetCoreVersion)" />
```
rajsite: actually, I have a hunch renovate won't figure this out. think I'll just use a conditional instead

.github/workflows/main.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       - uses: actions/setup-dotnet@v5
         with:
           global-json-file: packages/blazor-workspace/global.json
+      - uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '8.0.422'
```
jattasNI, line 64: What's the plan for removing dual support and updating the version used by the workflow to 10? Could we find a way to track this work after 8 is EOL or after all clients have moved to 10?

.github/workflows/main.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       - uses: actions/setup-dotnet@v5
         with:
           global-json-file: packages/blazor-workspace/global.json
+      - uses: actions/setup-dotnet@v5
```
jattasNI, line 62: Reminder to post back on the Blazor teams channel to follow up with clients about where we landed.

packages/blazor-workspace/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.421",
+    "version": "10.0.301",
```
jattasNI, line 3: Does this change the version that Nimble devs should install? Looks like the [root CONTRIBUTING.md](https://github.com/ni/nimble/blob/main/CONTRIBUTING.md#getting-started) mentions this version so it needs an update. Also: reminder to post on the Nimble dev channel when this goes in.

.github/workflows/main.yml
jattasNI, line 62: I skimmed a few Skyline .NET 10 upgrade PRs to see if there were any other common tasks we've done. (You can find these by looking through the history of `global.json` in some microservices).

One suggestion I found is to [adopt the cleaner .slnx file format](https://teams.microsoft.com/l/message/19:aa36fb23db964a1d8931052d35f3b81a@thread.tacv2/1776195164803?tenantId=eb06985d-06ca-4a17-81da-629ab99f6505&groupId=1e74afef-f47b-4137-ad1b-e550ec9c6bef&parentMessageId=1776195164803&teamName=SystemLink%20R%26D&channelName=SLE%20Development&createdTime=1776195164803) for the BlazorWorkspace.sln. But you can consider that optional / follow up if you want, not all services did it.

You might also ask some of the devs who did these upgrades (Paul, Darren, Aaron, etc) if there's anything else we should be thinking about.

jattasNI (approved):

packages/blazor-workspace/package.json
```diff
@@ -24,7 +24,7 @@
         "copy-resources-ok": "node build/copy-resources.mjs ok",
         "update-lock-files": "dotnet restore --force-evaluate",
         "dotnet-local": "npm run dotnet-local:download && npm run dotnet-local:extract && npm run dotnet-local:check",
-        "dotnet-local:download": "curl https://builds.dotnet.microsoft.com/dotnet/Sdk/8.0.421/dotnet-sdk-8.0.421-linux-x64.tar.gz --output dotnet-sdk.tar.gz",
+        "dotnet-local:download": "curl https://builds.dotnet.microsoft.com/dotnet/Sdk/10.0.300/dotnet-sdk-10.0.300-linux-x64.tar.gz --output dotnet-sdk.tar.gz",
```
msmithNI: Should this be `10.0.301` now?
(The PR description also still refers to `10.0.300`)

packages/blazor-workspace/Examples/Demo.Hybrid/Demo.Hybrid.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net8.0-windows7.0</TargetFramework>
+    <TargetFramework>net10.0-windows7.0</TargetFramework>
```
msmithNI: If I try to launch Demo.Hybrid, I get an exception (in `WebView2CompositionControl` `OnApplyTemplate()` at `TryInitializeD3DImage()`).

We hit the same in ConfigUI and the fix ended up being we had to update the `TargetFramework` to be Windows 10 specific. `net10.0-windows10.0.19041.0` is what we're using for ConfigUI.

Reference: https://github.com/dotnet/maui/issues/32002

packages/blazor-workspace/global.json
msmithNI: `NimbleBlazor/README.md` also needs some updates, since it still recommends Visual Studio 2022. I think Visual Studio 2026 is basically required now.

msmithNI (approved):

packages/blazor-workspace/global.json
rajsite: Updated to vs 2026 and removed most direct references to .net v8, i.e. "since .net 8", etc

.github/workflows/main.yml
rajsite: Added slnx, and will ping previous .NET updaters

packages/blazor-workspace/Examples/Demo.Hybrid/Demo.Hybrid.csproj
rajsite: weird, I swear a ran each example app but yea I'm hitting it too, thanks for checking

.github/workflows/main.yml
rajsite: Done!

.github/workflows/main.yml
rajsite: Captured an issue: https://github.com/ni/nimble/issues/2978

.github/workflows/main.yml
rajsite: Suggestions from other .NET 10 upgrade devs:
- Update dependencies
  - Using `dotnet list package --outdated` only non-sdk dependency that is out-of-date is bunit tracked in [azdo 3911416](https://dev.azure.com/ni/DevCentral/_workitems/edit/3911416)
- Updates for code analysis rules
  -  tracked in [azdo 3921962](https://dev.azure.com/ni/DevCentral/_workitems/edit/3921962)
- Updates to improve developer experience
  - PR includes changes to preserve after initial build that rebuilds within IDE behave well

PR #2985, opened 2026-06-18
Fix chat send button not re-enabling

Alexia-Claudia-Micu:
# Pull Request

## 🤨 Rationale

The send button is disabled when either `sendDisabled` is true (consumer side) or the text box is empty. The bug was that `shouldDisableSendButton` combined both conditions and its result was held in `isInputEmpty` (which only refreshed on text events).

Since `sendDisabled` is an attribute, its state change never caused `IsInputEmpty` to refresh.

## 👩‍💻 Implementation

`IsInputEmpty` now only tracks `this.textArea!.value.length === 0`  leaving the button state to be calculated in the template:  `            ?disabled=${x => (x.processing ? false : (x.sendDisabled || x.isInputEmpty))}` 

## 🧪 Testing

Added unit test.

## ✅ Checklist

### DIFF
--- a/change/@ni-spright-components-1812c098-b28e-45a8-9766-edeb9c7555cc.json
+++ b/change/@ni-spright-components-1812c098-b28e-45a8-9766-edeb9c7555cc.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix send button not being re-enabled",
+  "packageName": "@ni/spright-components",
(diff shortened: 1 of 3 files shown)


Alexia-Claudia-Micu:
@rajsite Could you review this PR when you have the time as well? It is a fix for: https://github.com/ni/nimble/pull/2981

rajsite (approved):

PR #2986, opened 2026-06-23
Fix WebKit test failure for the Rich text mention

### LINKED ISSUE #1938: Disabled rich text editor tests on Webkit
# 🧹 Tech Debt

Several rich text editor and markdown serializer tests fail on webkit with the error:
 `        TypeError: undefined is not an object (evaluating 'found.startContainer') thrown`

The tests have been disabled on Webkit.

todo @jattasNI mark tests with this bug ID and link to a search identifying them

This still needs to be debugged and reported to an upstream dependency (probably Prosemirror)

vivinkrishna-ni:
# Pull Request

## 🤨 Rationale

Fixes #1938 

## 👩‍💻 Implementation

1. The underlying issue was fixed in the `Prosemirror-view` package, where it previously caused a console error and made the cursor disappear. As mentioned in the [comment](https://github.com/ProseMirror/prosemirror/issues/1454#issuecomment-2722857281), the console log error has been fixed, which in turn fixes the test failure in WebKit. 
    1. Although it was mentioned that the cursor disappearing issue is not yet fixed, when I tried locally in WebKit, the issue appears to be fixed.
        hard-enter-issue-webkit
2. Another reason the test was failing is that it uses the `innerText`, where WebKit preserves whitespace/newlines that exist in the DOM. Therefore, the actual label name returned by the `innerText` matches `at mention\n  `, while the expected value does not include the whitespace and newline characters. To fix that, added a `.trim()` to the button label page object.
     1. Also, the `innerText.trim()` is used in [several places](https://github.com/search?q=repo%3Ani%2Fnimble%20innerText.trim()&type=code) in the nimble.

## 🧪 Testing

Test passes

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-9d9a3ea4-0474-49a6-a345-6900ca379e08.json
+++ b/change/@ni-nimble-components-9d9a3ea4-0474-49a6-a345-6900ca379e08.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Fix webkit test failure that occurred due to the underlying prosemirror issue is now resolved",
+  "comment": "Fix the WebKit test failure caused by an underlying ProseMirror issue that has now been resolved",
   "packageName": "@ni/nimble-components",
   "email": "123377523+vivinkrishna-ni@users.noreply.github.com",
   "dependentChangeType": "none"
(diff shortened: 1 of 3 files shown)


change/@ni-nimble-components-9d9a3ea4-0474-49a6-a345-6900ca379e08.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
rajsite: Page objects are public api / a source change
```suggestion
  "type": "patch",
```

rajsite (approved):

PR #2989, opened 2026-06-24
Autoscroll pattern updates WIP

rajsite:
# Pull Request

This PR targets #2975 for a refactor in scrolling implementation

## 🤨 Rationale
See https://github.com/ni/nimble/pull/2975#issuecomment-4784896445

## ✅ Checklist


Alexia-Claudia-Micu:
> * if you don't mind, could you explore the storybook of the proposed branch and see if it preserves the expected behavior?
 
@rajsite I ran and tested your branch locally and it maintains the expected behavior.
 
A few things I noticed while testing with the example app, I'm not sure how important they are but I will leave them here:
 
1. For a shorter conversation, we get a slight pixel shift/scroll while the inbound message is being generated (despite it having enough room in the viewport). It's pretty hard to notice and disappears for longer conversations:
ScrollShiftDemo

2. If I send a message and scroll up, the auto-scroll will still catch when the inbound message reaches the viewport bottom. I'm not sure if this should happen as we would be in the "auto-scroll cancels when user scrolls up" case. This might be the expected behavior, but I thought it was worth being aware of.
ScrollCatchDemo

Alexia-Claudia-Micu:
@rajsite I took a look over the new implementation and the previous problems have both been fixed. 

The only small thing I noticed when re-testing is that the auto-scroll doesn't catch for the first message:
FirstMessageDemo

However, if the user manually scrolls to the bottom, it will catch for the first message as well.
This issue is resolved once a second message is sent, I'm thinking that is because we have the "pin message to top" behavior for all but the first message.

packages/spright-components/src/chat/conversation/models/auto-scroll-manager.ts
```diff
@@ -0,0 +1,286 @@
+import {
```
rajsite, line 1: Previous fix (onScroll reachedBottom): A smooth scroll does move, scroll events do fire, but the achievable max scroll settles a few pixels short of the requested target (layout rounding, etc.). In that case reachedTarget (scrollTop within 1px of target) is never satisfied, so the guard would stay set forever. The reachedBottom check clears it once we're at the bottom even if we never hit the exact target.

Current fix (smoothScrollTo no-op): The smooth scroll doesn't move at all (target within 1px of current scrollTop, e.g. first message where max scroll is 0). No scroll event fires, so onScroll never even runs — meaning the reachedBottom check can't help here. This fix avoids arming the guard in the first place.

They're complementary, at different points in the scroll lifecycle:

smoothScrollTo guards the start (don't arm the guard if no movement will occur).
onScroll reachedBottom guards the end (movement happened but settled short of target).
Each has its own regression test in chat-conversation-autoscroll.spec.ts:96-131 — 'keeps following after a programmatic scroll settles short of its target' fails without the reachedBottom check, and 'follows streamed content for the first message when the anchor scroll is a no-op' fails without the smoothScrollTo no-op guard (I just confirmed both fail-without/pass-with). So neither is redundant.

PR #2990, opened 2026-06-24
Update Ok Breakpoint Column to fix menu stacking

hellovolcano:
# Pull Request

## 🤨 Rationale

Pinning the breakpoint column was causing the menu to be clipped behind the sibling pinned cells.

## 👩‍💻 Implementation

- Set the `menu-open` attribute on menu open and remove on close so the current menu can be displayed above the other pinned cells

note: this is likely a temporary fix that should be addressed later via a refactor to how we toggle the breakpoint's menu

## 🧪 Testing

Verified in storybook
Added some matrix tests

## ✅ Checklist

### DIFF
--- a/packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
+++ b/packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
@@ -51,6 +51,8 @@ export class TsTableColumnBreakpointCellView extends TableCellView<
 
     private focusLastItemWhenOpened = false;
 
+    private restoreRowMenuOpenOnClose = false;
+
     /** @internal */
     @observable
     private readonly breakpointEnabledString = 'Breakpoint enabled';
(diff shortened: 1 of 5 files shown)


packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
```diff
@@ -332,6 +352,30 @@ export class TsTableColumnBreakpointCellView extends TableCellView<
         this.button?.focus();
     };
 
+    private getOwningRow(): (HTMLElement & { menuOpen: boolean }) | undefined {
```
rajsite, line 355: Can you create an issue to improve this? I think I see the issue more clearly. The position sticky for the pinned cell created a new stacking context above the anchored region it can't fight. The llm actually broke it down nicely:

Image

I think this will be an excellent use case for the popover api in a small refactor. We don't need any change in the table and the breakpoint popover will get promoted to a top layer. We do want to adopt it ubiquitously in nimble (eventually, likely via a change in anchored region). But it would be a real solution to this behavior today in the breakpoint column: https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

rajsite (approved):

packages/ok-components/src/ts/table-column/breakpoint/cell-view/index.ts
hellovolcano: created #2992

PR #2993, opened 2026-06-26
Support CSS `field-sizing` on text-area, text-field, number-field, and combobox

### LINKED ISSUE #1228: Make textarea height optionally auto-resizable

m-akinc:
# Pull Request

## 🤨 Rationale

I wanted to size a `nimble-text-field` to the width of its content. This isn't straightforward unless you use the [`field-sizing` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/field-sizing). The property has been supported in Chrome/Edge for a couple years, but only in Safari (12/2025) and Firefox (6/2026) fairly recently. It would be very useful if you could apply the style to our custom elements and see the effect on the the native elements in the shadow DOM.

Also, apparently resolves #1228

Found one edge case (field-sizing: content, and a fixed width) captured in matrix tests that does not block this PR. Created an issue: https://github.com/ni/nimble/issues/2999

## 👩‍💻 Implementation

`field-sizing` is only supported on `input` and `textarea` elements, so it only really applies to `nimble-text-field`, `nimble-number-field`, `nimble-text-area`, and `nimble-combo-box`.

Updated stylesheets to appy `field-sizing: inherit` to the native `input` and `textarea` elements, as well as their containing `div`s and/or `slot`s. This allows a user to style the host element and have that style "cascade".

In the case of `nimble-combobox` and `nimble-text-area`, there is a `min-width` by default, so if you want it to size smaller than that, you need to explicitly change that property, too.

Additionally, added a 4-px padding to the right side of the `nimble-number-field` when the stepper buttons are hidden to better match the design.

## 🧪 Testing

Added visual test cases for `nimble-text-field`, `nimble-number-field`, `nimble-text-area`, and `nimble-combobox`.

## ✅ Checklist

### DIFF
--- a/change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
+++ b/change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Support CSS field-sizing on text-area and text-field",
+  "comment": "Support CSS field-sizing on text-area, text-field, number-field, and combobox. Also minor padding adjustment for number-field",
   "packageName": "@ni/nimble-components",
   "email": "7282195+m-akinc@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
+++ b/change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Support CSS field-sizing on text-area and text-field",
+  "packageName": "@ni/nimble-components",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/nimble-components/src/number-field/styles.ts
+++ b/packages/nimble-components/src/number-field/styles.ts
@@ -142,7 +142,7 @@ export const styles = css`
     }
     
     :host([hide-step]) .control {
-        padding-right: ${mediumPadding};
+        padding-right: ${smallPadding};
     }
 
     .control:hover,
--- a/packages/nimble-components/src/number-field/styles.ts
+++ b/packages/nimble-components/src/number-field/styles.ts
@@ -140,6 +140,10 @@ export const styles = css`
         padding: 0px;
         padding-left: ${mediumPadding};
     }
+    
+    :host([hide-step]) .control {
+        padding-right: ${mediumPadding};
+    }
 
     .control:hover,
     .control:focus,
@@ -253,6 +257,7 @@ export const styles = css`
 
             :host([full-bleed]) .control {
                 padding-left: 0px;
+                padding-right: 0px;
             }
         `
     )
--- a/packages/nimble-components/src/number-field/styles.ts
+++ b/packages/nimble-components/src/number-field/styles.ts
@@ -32,6 +32,7 @@ export const styles = css`
         outline: none;
         ${userSelectNone}
         color: ${bodyFontColor};
+        field-sizing: fixed;
         --ni-private-hover-indicator-width: calc(${borderWidth} + 1px);
         --ni-private-height-within-border: calc(
             ${controlHeight} - 2 * ${borderWidth}
--- a/packages/nimble-components/src/number-field/styles.ts
+++ b/packages/nimble-components/src/number-field/styles.ts
@@ -64,6 +64,7 @@ export const styles = css`
         position: relative;
         display: flex;
         flex-direction: row;
+        field-sizing: inherit;
         justify-content: center;
         align-items: center;
         border-radius: 0px;
@@ -131,6 +132,7 @@ export const styles = css`
         font: inherit;
         background: transparent;
         color: inherit;
+        field-sizing: inherit;
         height: var(--ni-private-height-within-border);
         width: 100%;
         border: none;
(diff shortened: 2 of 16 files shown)


packages/nimble-components/src/number-field/styles.ts
rajsite, line 30: Can you set a default for field-sizing on the host to make it clearer to follow the inherit path. It's a bit subtle to follow the intended behavior otherwise.

change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: combobox is also an input text style control

m-akinc: Added support to combobox, too, though we give combobox a `min-width` by default, so if you want it to go below that size, you have to also change `min-width`.

packages/nimble-components/src/number-field/styles.ts
m-akinc: Done

change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
rajsite, line 1: talked offline, high-level concerns:
- Want to avoid any changes to default sizing behvior, i.e. in some iterations of the PR the size attr on input or width of parts of the control via css are changing. Don't want to make changes that are observable to clients.
- During demos point was brought up about controls that render additional content inside them, i.e. text-field action buttons, error visible states, those should be exercised in stories
- This behavior is an explicit feature request, we should capture styling docs in the control's respective storybook pages bout enabling this style and close the related issue https://github.com/ni/nimble/issues/1228

rajsite, line 1: @m-akinc  As an fyi, you can target the part directly today to enable field-sizing: https://github.com/ni/nimble/issues/1228#issuecomment-2705169585

But the changes in this PR would certainly be nice (just saying in-case it's blocking something to shore-up these questions)

m-akinc: I can wait, I think.

m-akinc: Done

packages/storybook/src/nimble/text-area/text-area.mdx
```diff
@@ -8,12 +8,16 @@ import ComponentApisLink from '../../docs/component-apis-link.mdx';
...
+If you configure your text area to be user-resizable (with the `resize`
+attribute) in a certain dimension, do not set an explicit size for that dimension
+(via `height` and/or `width` `style` properties), or you may experience unexpected
 resize behavior. If you want to set the initial size of a resizable text area,
 use the `rows` and/or `cols` attribute(s).
 
+To make the text area grow automatically to fit its content, style the element
+with `field-sizing: content`. This will allow it to grow both horizontally and
+vertically. Set a fixed `width` if you only want it to grow vertically.
+
```
rajsite: I think we should document this for each component being changed, not just textarea. Can you make a shared doc component to use across each of the elements we are adding support for?

Can make a shared styling doc in the pattern of button-styling-docs.mdx: https://github.com/ni/nimble/blame/main/packages/storybook/src/nimble/patterns/button/button-styling-docs.mdx

But don't have to go that overkill with examples and everything. Takeaways are a shared mdx for that doc snippet that ends up under the styling section of each of the component docs.

m-akinc: Done

packages/nimble-components/src/number-field/styles.ts
```diff
@@ -131,12 +133,17 @@ export const styles = css`
...
         height: var(--ni-private-height-within-border);
         width: 100%;
         border: none;
         padding: 0px;
         padding-left: ${mediumPadding};
     }
+    
+    :host([hide-step]) .control {
+        padding-right: ${mediumPadding};
+    }
```
rajsite, line 146: This seems like a completely unrelated change

It looks like the following in this PR (medium right padding):

Image

This way on main (no right padding):

Image

Niether of which align with the design:
image

change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
rajsite, line 1: PR feedback:
- The padding change should be documented in PR and change since it is an unrelated change and differs from main (no right padding)
- The padding with hide-step should be small padding to match the text field in figma, i.e. a numeric text field with hide-step and text field in line with each other with long values should fill the same space:
    Image
    [figma ref](https://www.figma.com/design/PO9mFOu5BCl8aJvFchEeuN/Nimble_Components?node-id=1295-51200&m=dev)
- lets make these as matrix tests instead of hand rolling the dimensions
- text field should have a long text dimension added like the others (it jsut has short)

packages/nimble-components/src/number-field/styles.ts
rajsite: merged with other comment

change/@ni-nimble-components-6d58234e-1827-4414-a92d-2c3b47400e14.json
m-akinc: Done

rajsite (approved):

PR #2994, opened 2026-06-30
Fix components not centering

Alexia-Claudia-Micu:
# Pull Request

## 🤨 Rationale

[Azdo issue](https://dev.azure.com/ni/DevCentral/_workitems/edit/3946267)

The auto-scroll refactor wrapped the default slot in a `.messages-anchored` div with `[flex: none]`. The previously used min-height: 100% approach does not work on flex items whose parent has no explicit height (only a computed one from flex: 1, which browsers resolve to 0). This caused `.messages-anchored` to collapse to its content height, leaving no free space for the welcome message to center within.

## 👩‍💻 Implementation

- Changed messages-anchored to always be min-height: 100%
- Changed welcome message sizing to fill available space in the flex layout.

## 🧪 Testing

Storybook and example app testing.

## ✅ Checklist

### DIFF
--- a/packages/spright-components/src/chat/conversation/styles.ts
+++ b/packages/spright-components/src/chat/conversation/styles.ts
@@ -89,7 +89,8 @@ export const styles = css`
     }
 
     .messages-anchored {
-        flex: 1;
+        flex: none;
+        min-height: 100%;
     }
 
     :host([appearance='overlay']) .messages {
--- a/packages/spright-components/src/chat/conversation/styles.ts
+++ b/packages/spright-components/src/chat/conversation/styles.ts
@@ -85,7 +85,7 @@ export const styles = css`
         display: none;
     }
 
-    .messages-anchored.anchor-active {
+    .messages-anchored {
         min-height: 100%;
     }
 
--- a/packages/spright-components/src/chat/conversation/styles.ts
+++ b/packages/spright-components/src/chat/conversation/styles.ts
@@ -75,18 +75,21 @@ export const styles = css`
 
     .messages-history,
     .messages-anchored {
-        flex: none;
         display: flex;
         flex-direction: column;
         row-gap: 32px;
+    }
+
+    .messages-history {
+        flex: none;
     }
 
     .messages-history.region-empty {
         display: none;
     }
 
-    .messages-anchored.anchor-active {
-        min-height: 100%;
+    .messages-anchored {
+        flex: 1;
     }
 
     :host([appearance='overlay']) .messages {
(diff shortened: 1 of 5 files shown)


Alexia-Claudia-Micu:
@rajsite I noticed a centering issue for the welcome and login messages and made this PR to fix it. Could you take a look over it?. Here's the bug: https://dev.azure.com/ni/DevCentral/_workitems/edit/3946267

packages/spright-components/src/chat/message/welcome/styles.ts
```diff
@@ -22,6 +22,7 @@ export const styles = css`
         flex-direction: row;
         justify-content: center;
         flex-shrink: 0;
+        margin: auto 0;
```
rajsite: Relying on margin for layout is a bit of a hack. We can leverage modern layout features. Can you do something like the following:
```
height: 100%;
align-items: center;
```
that states the intended behavior more explicitly, we want the welcome message to fill the height of the conversation and we want to use flex align-items center to position the content

packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -75,17 +75,21 @@ export const styles = css`
 
     .messages-history,
     .messages-anchored {
-        flex: none;
```
rajsite, line 78: I don't understand the change in this file, doesn't flex always stay none for message-histroy and message-anchored? What is the intention of the changes?

packages/spright-components/src/chat/conversation/styles.ts
```diff
@@ -75,17 +75,21 @@ export const styles = css`
...
+    .messages-history {
+        flex: none;
+    }
+
     .messages-history.region-empty {
         display: none;
     }
 
-    .messages-anchored.anchor-active {
+    .messages-anchored {
```
rajsite: The only effective change is that messages-anchored always stays min-height: 100% with this change. Is that actually correct?

rajsite:
Talked to @Alexia-Claudia-Micu and I'll address the feedback and complete the PR

packages/spright-components/src/chat/message/welcome/styles.ts
rajsite: even better, just set `flex: 1` so welcome fills the space available in the conversation

packages/spright-components/src/chat/conversation/styles.ts
rajsite: reverted

packages/spright-components/src/chat/conversation/styles.ts
rajsite: New behavior is that we want messages-anchored, i.e. the default slot to always be min-height 100% (i.e. equal to prent height). When then welcome message is slotted it'll fill all available flex layout space.

rajsite (approved):

PR #2995, opened 2026-06-30
Add scrolling case for longer user message

Alexia-Claudia-Micu:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/3897966
https://dev.azure.com/ni/DevCentral/_workitems/edit/3943833

Current behavior: a long user message is sent, scrolled to the bottom and will take up the entire viewport. This leaves no space for the ai answer to be generated in.
Desired behavior: a long user message is sent, scrolled to leave the bottom 20% visible. This leaves the rest of the viewport space for the ai message to be generated in.

## 👩‍💻 Implementation

Added `anchoredMessageMaxViewportFraction` = 0.2 to keep the user message within that height in the viewport.

## 🧪 Testing

Example app testing.

## ✅ Checklist


packages/spright-components/src/chat/conversation/models/auto-scroll-manager.ts
```diff
@@ -206,6 +224,19 @@ export class AutoScrollManager implements Subscriber {
...
+        const messageRect = message.getBoundingClientRect();
+        const messageTopInContainer = messageRect.top - containerRect.top + container.scrollTop;
+        const { clientHeight } = container;
+        const maxMessageCoverage = clientHeight * anchoredMessageMaxViewportFraction;
+        if (messageRect.height > maxMessageCoverage) {
+            return messageTopInContainer + messageRect.height - maxMessageCoverage;
+        }
+        return messageTopInContainer;
+    }
+
```
rajsite, line 260: A couple of concerns with this PR:
- If you notice from the previous refactor, a significant effort was put in to avoid manual calculations like this to measure the dom and instead rely on native CSS browser features. Please take some time to brainstorm and consider how this can be done in CSS with the browser native layout system instead of manually calculating layout. Things like this should be a last resort with comments describing approaches considered.
- This PR adds no tests. Notice the previous refactor includes tests to cover behavior. Please review and add tests
- This change is not covered by storybook matrix tests. Make sure this case is covered in storybook matrix snapshot testing to prevent regressions.

rajsite:
Talked to @Alexia-Claudia-Micu and they'll be OOO for a few weeks and can address feedback when back. Moving to draft for now.

packages/spright-components/src/chat/conversation/models/auto-scroll-manager.ts
Alexia-Claudia-Micu: I did some digging and the only CSS solution I found was a sentinel element at the message bottom with scroll-margin-top: 20cqh, but it has the problem that short messages end up with empty space above them instead of being flush with the viewport top (because scroll-margin can't reference the element's own height). So I opted to add a comment explaining the need for these manual calculations.

I also added unit tests, but had some trouble with the storybook matrix tests as they render the component and take a screenshot immediately. The scroll behavior only triggers after a user sends a message, so the matrix story only shows the un-scrolled state.  I could try to look for some workarounds for this such as waiting long enough for the scrolling animation to complete. However, I first wanted to ask if you think this is something worth pursuing.

rajsite: I'm OOO through 7/20, I can either take a look after I'm back in office or if need to look sooner @mollykreis may be able to review the approach and suggest alternatives / give the approval to merge.

In terms of alternatives to what is being proposed (concerns of lots of manual sizing calculations and concerns about timing assumptions / responsiveness to dynamic content) I'd have to look more closely when in office.

In terms of concerns with just how the implementation is done as-is, a concern I have is having dynamic style attributes being mutated directly on elements. Instead set custom property values that can be consumed via css ([table examples](https://github.com/ni/nimble/blob/bcc8bd28637fc7362d3b934a37ba392b8b04d1dc/packages/nimble-components/src/table/template.ts#L62-L70)). Want the definition of CSS styles to be static and can do things like `min-height: calc(100% + var(--ni-private-conversation-shortfall-height));`;

rajsite: Back in office but was asked to prioritize other tasks through this week so the earliest I can take a look myself is next week. Can work with @jattasNI to review the current changes to the chat conversation components.

jattasNI: I've finally cleared most of my backlog from being OOO and should have bandwidth to review this PR next week.

Before I do, @Alexia-Claudia-Micu could you clarify some questions?

1. Could you update the PR description to explain the current behavior and the desired behavior? The PR description links to a couple issues but those aren't enough to tell what this PR is doing.
2. From my storybook experimentation, it looks like the current behavior is that new user messages get scrolled so that their top aligns with the top of the conversation. For tall messages this means the top is visible but the bottom isn't. The new behavior is that, for messages greater than 2 lines, the top is scrolled offscreen, the bottom 2.5 lines are visible, and space is available for the response. Is this confirmed as the desired behavior from UX or is it currently a dev proposal?

I haven't looked into the implementation yet but my gut feel is that it might take some brainstorming to meet Milan's architectural concerns while achieving the desired UX behavior. So I want to make sure it's worth investing in that before we do. An alternative would be to keep the behavior as-is for now and wait for user feedback. I learned today that the desktop teams plan to uptake this soon, so maybe their feedback will help us decide what behavior we want.

Alexia-Claudia-Micu: 1. I added the current and desired behaviors to the PR description.
2. Yes, that is the behavior we tried to achieve. I chose 20% of the viewport by looking at other chatbots. I could not confirm this with an UX designer, but I can change that number through `anchoredMessageMaxViewportFraction`.

Thanks for taking the time to look over this. Seeing how this is a visual change, we could leave it as-is and wait for user/desktop team feedback. I will close this PR until then.

PR #3010, opened 2026-07-09
Add padding between switch and its label

m-akinc:
# Pull Request

## 🤨 Rationale

The `nimble-switch` design dictates that there should be a two-pixel gap/padding between the label and control:
image

Needed for AzDO bug: https://dev.azure.com/ni/DevCentral/_workitems/edit/3957974

## 👩‍💻 Implementation

There is not an existing 2-pixel padding/spacing token, and I'm not sure we need one. I just added a hard-coded 2px bottom padding to the label element.

## 🧪 Testing

Verified in Storybook.

### DIFF
--- a/packages/nimble-components/src/switch/styles.ts
+++ b/packages/nimble-components/src/switch/styles.ts
@@ -62,7 +62,7 @@ export const styles = css`
         display: flex;
         align-items: center;
         ${'' /*
-            Reserve space around the control to fill a 32px height until we have a switch 32 design.
+            Reserve space around the 24px control to fill a 32px height until we have a switch 32 design.
             See: https://github.com/ni/nimble/issues/3013
         */}
         padding-top: 2px;
--- a/packages/nimble-components/src/switch/styles.ts
+++ b/packages/nimble-components/src/switch/styles.ts
@@ -51,6 +51,7 @@ export const styles = css`
     .label {
         color: ${controlLabelFontColor};
         font: ${controlLabelFont};
+        padding-bottom: 2px;
     }
 
     :host([disabled]) .label {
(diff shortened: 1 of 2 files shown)


packages/nimble-components/src/switch/styles.ts
rajsite, line 70: Noticed a couple weird things. This control should be the layout for a 24px control but is faking the layout of a 32px control with padding. Root issue is we don't have designs for a 32px switch, created an issue for that:
https://github.com/ni/nimble/issues/3013

Looking at normal 32px controls they are 16px labels and 32px controls, so we should keep the label 16px and adjust the control / switch part.

The problem then is the control part is not sized correctly, moved the padding-bottom into the control instead of the top-level host and changed the size to 32px (2px top, 24px, 6px bottom). This is dev designed until we have a proper switch 32 design.


```suggestion
   :host {
       outline: none;
       font: ${bodyFont};
       color: ${buttonLabelFontColor};
       flex-direction: column;
       cursor: pointer;
       --ni-private-switch-height: 24px;
       --ni-private-switch-indicator-size: 24px;
       --ni-private-switch-indicator-inner-size: 18px;
       --ni-private-switch-indicator-margin: -2px;
   }

   :host([disabled]) {
       cursor: default;
       color: ${buttonLabelDisabledFontColor};
   }

   .label {
       color: ${controlLabelFontColor};
       font: ${controlLabelFont};
   }

   :host([disabled]) .label {
       color: ${controlLabelDisabledFontColor};
   }

   .label__hidden {
       display: none;
       visibility: hidden;
   }

   .switch-container {
       display: flex;
       align-items: center;
       ${'' /*
           Reserve space around the control to fill a 32px height until we have a switch 32 design.
           See: https://github.com/ni/nimble/issues/3013
       */}
       padding-top: 2px;
       padding-bottom: 6px;
   }
```

rajsite: wow... github completely messed up that patch commit. I'll do it manually

rajsite (approved):

PR #3029, opened 2026-08-05
Remove forked `RouterLink` implementation

m-akinc:
# Pull Request

## 🤨 Rationale

As of Angular 20, `RouterLink` supports custom elements. We should drop our vendored/modified version in favor of the official Angular one.

## 👩‍💻 Implementation
- Removed our forked copy of `router_link.ts`, as well as related, forked tests/utilities.

## 🧪 Testing
Existing tests pass


rajsite (changes requested):
> I created DisableableNimbleAnchorBaseDirective, which, after every change detection pass, overwrites RouterLink.target's value with "_disabled" if the host is disabled and RouterLink.target is "_self"/undefined. If the host is not disabled, and RouterLink.target is "_disabled", that means we replaced it, so we restore it to undefined.

That sounds like a huge hack that will result in timing and state inconsistencies and we should actually discuss changes like this before putting up large PRs. 
Can consult with Molly to consider possible alternatives or can discuss when back in office

m-akinc:
> > I created DisableableNimbleAnchorBaseDirective, which, after every change detection pass, overwrites RouterLink.target's value with "_disabled" if the host is disabled and RouterLink.target is "_self"/undefined. If the host is not disabled, and RouterLink.target is "_disabled", that means we replaced it, so we restore it to undefined.
> 
> That sounds like a huge hack that will result in timing and state inconsistencies and we should actually discuss changes like this before putting up large PRs. Can consult with Molly to consider possible alternatives or can discuss when back in office

I'm not convinced this would result in any actual (timing or other) issues the way it is implemented, but I concede it's a hack. Just yesterday, I thought of another possible approach: when the control becomes `disabled`, maybe we could temporarily remove `routerLink` (and cache the value for restoration), so that the `RouterLink` directive no longer applies. This feels a lot like the current approach, but has the advantages of being (1) agnostic about how `RouterLink` works (i.e. caring about `target`), and (2) more direct about intent (i.e. "turn off router link" rather than "change `target`").

packages/angular-workspace/nimble-angular/src/directives/breadcrumb-item/nimble-breadcrumb-item-router-link-with-href.directive.ts
```diff
@@ -1,59 +0,0 @@
...
-        // can be in control of when the click handler is called. This allows RouterLinkWithHref's onClick to only be called
-        // when the link within the element is clicked rather than any part of the element.
-        return true;
-    }
-
-    @HostListener(
-        'click',
-        ['$event', '$event.button', '$event.ctrlKey', '$event.shiftKey', '$event.altKey', '$event.metaKey']
-    )
-    public nimbleOnClick(event: MouseEvent): boolean {
```
mollykreis, line 51: Did this logic get lost, or does it still exist somewhere? (I didn't see it moved to the `NimbleBreadcrumbItemDirective`).

packages/angular-workspace/nimble-angular/src/directives/anchor-base/disableable-nimble-anchor-base.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { AnchorBase } from '@ni/nimble-components/dist/esm/anchor-base';
+import { type BooleanValueOrAttribute, toBooleanProperty } from '@ni/nimble-angular/internal-utilities';
+import { NimbleAnchorBaseDirective } from './nimble-anchor-base.directive';
+
+/**
+ * Base class for directives providing link attributes that include 'disabled'.
+ * @internal
+ */
+@Directive()
+export class DisableableNimbleAnchorBaseDirective<T extends (AnchorBase & { disabled: boolean })>
```
mollykreis: > > >I created DisableableNimbleAnchorBaseDirective, which, after every change detection pass, overwrites RouterLink.target's value with "_disabled" if the host is disabled and RouterLink.target is "_self"/undefined. If the host is not disabled, and RouterLink.target is "_disabled", that means we replaced it, so we restore it to undefined.
> > 
> > That sounds like a huge hack that will result in timing and state inconsistencies and we should actually discuss changes like this before putting up large PRs. Can consult with Molly to consider possible alternatives or can discuss when back in office
> 
> I'm not convinced this would result in any actual (timing or other) issues the way it is implemented, but I concede it's a hack. Just yesterday, I thought of another possible approach: when the control becomes disabled, maybe we could temporarily remove routerLink (and cache the value for restoration), so that the RouterLink directive no longer applies. This feels a lot like the current approach, but has the advantages of being (1) agnostic about how RouterLink works (i.e. caring about target), and (2) more direct about intent (i.e. "turn off router link" rather than "change target").

Unless I'm missing something, the current approach of replacing the target causes the target to get reset to `undefined` even if the client code had it configured to some other value. Is that correct?

It looks like the current nimble code handles disabled anchor components by overriding the `onClick` event handler. Is that not an option any more?

m-akinc: > Unless I'm missing something, the current approach of replacing the target causes the target to get reset to undefined even if the client code had it configured to some other value. Is that correct?

We will only replace the `target` value with `'_disabled'` if it is `'_self'` or `undefined` (which are equivalent). So, when restoring, we are always setting it back to (functionally) the same value we replaced. And if `target` is changed to something else while the element is disabled (and the directive has the `'_disabled'` value), we won't end up changing it once the element is re-enabled (because we only restore over a `'_disabled'` value).

> It looks like the current nimble code handles disabled anchor components by overriding the onClick event handler. Is that not an option any more?

That was possible because we were providing an _alternative_ version of the router link directive for users to use. If the goal is to stop doing that and have them use the standard, Angular router link directive, then I don't see how we can override its `onClick` handler--unless we maybe monkey-patch it, but that doesn't seem like a better solution to me.

packages/angular-workspace/nimble-angular/src/directives/breadcrumb-item/nimble-breadcrumb-item-router-link-with-href.directive.ts
m-akinc: It got lost. ☹️ This is another case of needing to gate router navigation on some custom condition, but, unfortunately, this is even hairier than the "disabled" case. I'm brainstorming some alternate ideas.

m-akinc: I've backed out the part of this change that removed our component-specific `NimbleAnchor***RouterLinkWithHrefDirective` and `NimbleAnchor***RouterLinkDirective` files, so now this is moot.

packages/angular-workspace/nimble-angular/src/directives/anchor-base/disableable-nimble-anchor-base.directive.ts
m-akinc: Moot now that I've narrowed the scope of this PR to only removing the forked Angular files.

mollykreis (approved):

rajsite (approved):

rajsite:
Looks great! Turned out to be a very clean improvement!

PR #3030, opened 2026-08-13
Angular 21: Manually enable provideZoneChangeDetection and directive type binding updates

m-akinc:
# Pull Request

## 🤨 Rationale

The Angular 21 update will:
- make zoneless change detection the default (we need to re-enable zone based change detection)
- default to type-checking host bindings

We can pre-emptively make the changes that will be necessary, so the actual upgrade change is smaller and quicker to review.

## 👩‍💻 Implementation

- Added `provideZoneChangeDetection()` to the TestBed configuration providers lists (where necessary)
- Added `$any()` in bindings to effectively suppress false-positive type errors where the declared types are not specific enough

## 🧪 Testing

Tests pass

### DIFF
--- a/packages/angular-workspace/example-client-app/src/app/app.component.spec.ts
+++ b/packages/angular-workspace/example-client-app/src/app/app.component.spec.ts
@@ -1,17 +1,13 @@
 import { provideZoneChangeDetection } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
-import { NimbleNumberFieldModule, NimbleTextFieldModule } from '@ni/nimble-angular';
+import { AppModule } from './app.module';
 import { AppComponent } from './app.component';
 
 describe('AppComponent', () => {
     beforeEach(() => {
         TestBed.configureTestingModule({
-            declarations: [
-                AppComponent
-            ],
             imports: [
-                NimbleTextFieldModule,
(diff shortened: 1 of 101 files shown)


packages/angular-workspace/example-client-app/src/app/app.component.spec.ts
```diff
@@ -13,6 +13,7 @@ describe('AppComponent', () => {
                 NimbleTextFieldModule,
                 NimbleNumberFieldModule
             ],
+            providers: [provideZoneChangeDetection()],
             schemas: [CUSTOM_ELEMENTS_SCHEMA]
```
rajsite: It's unrelated, could do in a follow-up or here, but we shouldn't need CUSTOM_ELEMENTS_SCHEMA

rajsite (approved):

packages/angular-workspace/example-client-app/src/app/app.component.spec.ts
m-akinc: Done

rajsite: fyi build failed

jattasNI (approved):

